PR #1529, opened 2022-11-29
Remove developer information from the homepage

benwerner01:

This PR addresses a `@todo` to remove developer information from the homepage, and adds a redirect to the `/login` page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203451531168809/f) _(internal)_

No.

## 🛡 What tests cover this?

None.

## ❓ How to test this?

1. Run the developer environment using `yarn dev`
2. Go to `localhost:3000` and ensure you are redirected to `localhost:3000/login`
3. Log in and ensure you are redirected to the workspace
4. Go to `localhost:3000` and ensure you are redirected to the workspace

## 📹 Demo

https://user-images.githubusercontent.com/42802102/204553968-1b4f4d19-61dd-420c-b525-c34c77b7e0b4.mov


packages/hash/frontend/src/pages/index.page.tsx
```diff
@@ -1,44 +1,23 @@
...
-          : "You don't have a kratos session"}
-      </Typography>
-      <Button onClick={logout}>Log out</Button>
-      <Typography gutterBottom>
-        This is what your kratos session looks like:
-      </Typography>
-      {kratosSession ? (
-        <Box component="pre">{JSON.stringify(kratosSession, null, 2)}</Box>
-      ) : null}
+      <Typography variant="h1">Homepage</Typography>
```
Alfred-Mountfield, line 20: My understanding is this would be a bug to get stuck on. Will user's see it in between React renders? 

If so, I think just seeing a Homepage H1 tag might not be a great experience, perhaps we could use a `Suspense` or something component?

Alfred-Mountfield (commented):
Thanks Ben, I think it looks good, I just generally have a question about what we want to replace the page contents with.

nonparibus:
Is it possible a user might get stuck on, or otherwise land on the homepage somehow, and not be redirected? Obviously various people were landing on this page previously, and seeing the Kratos message.

This still looks weird to me given it now flashes up and redirects away.

packages/hash/frontend/src/pages/index.page.tsx
benwerner01: > Will user's see it in between React renders?

Yes, when they go to the homepage (because the redirect happens client side).

> If so, I think just seeing a Homepage H1 tag might not be a great experience, perhaps we could use a Suspense or something component?

What do you mean? When would the suspense component render, and what would it display? If it's going to wait on the result of `authenticatedUser`, then we might as well make this a blank page because users will never see the contents of the suspense component.

I think the ideal solution would be for the redirect to happen server-side, which should be possible - but would require us to introduce a bunch of server-side things that let the server-side code make authenticated requests to the GQL API (for the `me` query), and I'm not sure that's worth doing right now. Especially given that I assume we will eventually have a homepage with things to show to logged out users (correct me if I'm wrong @nonparibus), and won't just automatically redirect them to the login page.

benwerner01:
> Is it possible a user might get stuck on, or otherwise land on the homepage somehow, and not be redirected? Obviously various people were landing on this page previously, and seeing the Kratos message.

On the homepage, if the user is logged in they are redirected to the workspace. If they're logged out, they're redirected to the `/login` page. I'm aware people were getting stuck on the homepage previously, which I'm hoping to have addressed in this PR with the following bug fixes:
1. [refreshing the kratos session if it is defined and `authenticatedUser` is undefined](https://github.com/hashintel/hash/pull/1529/commits/8e30da387c95ab6e60cb24a1cf8f3fcfa060f277), and
2. [fixing the `useAuthenticatedUser` hook in `_app`, which was previously not working as it was called outside of the `SessionProvider`](https://github.com/hashintel/hash/pull/1529/commits/5f4edef55b94029ec0f0fcf4246382ee59134483).

If people continue getting stuck on the homepage, that is a bug and should be reported (whilst testing locally I was not able to accomplish this).

> This still looks weird to me given it now flashes up and redirects away.

I've made a slight improvement so that the login page redirects the user straight to the workspace, rather than redirecting to the homepage and relying on the homepage to redirect the user to the workspace.

https://user-images.githubusercontent.com/42802102/204810762-3283bfa9-f338-4c47-9dd9-4a12ab95f471.mov

Regarding the flashing behaviour on the homepage, I'm not really sure what can be done as part of this PR, as we are currently checking if a user is logged in client side. This is an asynchronous operation, so there will be a short amount of time the user sees the homepage whilst the homepage figures out whether to send the user to the workspace, or the login page.

As I explained [in response to Alfie](https://github.com/hashintel/hash/pull/1529#discussion_r1035870794), we could make this a server side redirect so that the homepage is never rendered on the frontend. However this would require the server-side NextJS code to make authenticated requests to the GraphQL API. Although this should be possible, I'm not sure it's worth setting up if we end up implementing a homepage at some point anyway. If you feel strongly that we should address this now, I think it should be a standalone PR that's addressed in follow-up.

packages/hash/frontend/src/pages/index.page.tsx
benwerner01, line 20:
```suggestion
      <Typography variant="h1">Homepage</Typography>
```
@nonparibus given that users shouldn't ever be on this page, would you be okay with just making it a blank page (rather than having "Homepage" be briefly displayed)?

nonparibus: Yes. I think there was some miscommunication around the circumstances in which this would appear, originally.

nonparibus:
I don't understand why the "Home" tab is considered "the workspace" and not just _the homepage_ for logged-in users.

packages/hash/frontend/src/pages/index.page.tsx
benwerner01: I've removed it in https://github.com/hashintel/hash/pull/1529/commits/8796d88f8ac77842469ccf9054d8c8074d648909

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -252,7 +252,7 @@ export const updateEntity: ResolverFn<
 ) => {
   // The user needs to be signed up if they aren't updating their own user entity
   if (
-    entityId !== userModel.getEntityUuid() &&
+    entityId !== userModel.getBaseId() &&
```
thehabbos007, line 255: For anyone viewing along, this fixes an issue with non-signed up users not being able to sign up properly

packages/hash/frontend/src/lib/user.ts
```diff
@@ -138,14 +141,20 @@ export const constructAuthenticatedUser = (params: {
...
-  if (!properties || !metadata) {
+  const latestUserEntity = getEntityAtTimestamp(
+    subgraph,
+    userEntityEditionId.baseId,
+    new Date(),
+  );
+
+  if (!latestUserEntity) {
     throw new Error(
       `Could not find entity edition with ID ${userEntityEditionId} in subgraph`,
```
thehabbos007, line 146:
```suggestion
      `Could not find entity with ID ${userEntityEditionId.baseId} in subgraph`,
```

packages/hash/frontend/src/pages/index.page.tsx
```diff
@@ -1,46 +1,27 @@
...
+     * redirect to the workspace or login page is done before rendering
+     * this empty homepage.
+     *
+     * @see https://app.asana.com/0/1203179076056209/1203451531168818/f
+     */
     if (authenticatedUser) {
       void router.push(`/${authenticatedUser.userAccountId}`);
     }
-  }, [router, authenticatedUser]);
+  }, [router, authenticatedUser, kratosSession, loading]);
```
thehabbos007, line 22: May be more of a dependency list question, but would these (`kratosSession, loading`) be required here?

packages/hash/frontend/src/pages/login.page.tsx
```diff
@@ -108,13 +111,28 @@ const LoginPage: NextPageWithLayout = () => {
             identifier: email,
             password,
           })
-          // We logged in successfully! Let's bring the user home.
-          .then(() => {
+          // We logged in successfully! Let's redirect the user.
+          .then(async () => {
```
thehabbos007, line 115: Would the result of the push at line 105 ever be able to cancel the executiong of the code here? Or can the code be assumed to run no matter than route the user is in?

packages/hash/frontend/src/pages/index.page.tsx
benwerner01: Fixed in https://github.com/hashintel/hash/pull/1529/commits/add5b75801745614266c96b36b39019e1549227f

packages/hash/frontend/src/pages/login.page.tsx
benwerner01: > Would the result of the push at line 105 ever be able to cancel the executing of the code here? 

My understanding is no it wouldn't, as it's a shallow redirect.

benwerner01:
> I don't understand why the "Home" tab is considered "the workspace" and not just _the homepage_ for logged-in users.

We've discussed this internally, and created [this follow-up task](https://app.asana.com/0/1203179076056209/1203443872976476/f)

thehabbos007 (approved):
Thank you, Ben! Very welcome changes here for the signup. We should address any comments as follow-ups, as this would be nice to get in now :)

PR #1530, opened 2022-11-29
Fix edges not generated for root vertices

TimDiekmann:

The edges based on the root vertices are not added to the subgraph, this fixes that bug.

## 🔍 What does this change?

- Don't insert the vertices before the resolving logic (results in an `DependencyStatus::DependenciesUnresolved` status instead of `DependencyStatus::Unknown`)
- Rename functions to better reflect this
- Make `GraphResolveDepths` required for dependency maps


Alfred-Mountfield (approved):
Nice! It even simplifies it a bit.

Maybe we want to rename `with_dependency` to `with_dependencies`, or something with "traverse" in it (which is more conventional anyway for graph.. traversal)

TimDiekmann:
Good point: a46f445721a8cf4b4a65a6e8babf8567d6a1cc4e

Alfred-Mountfield (approved):
🎉

PR #1532, opened 2022-11-29
Simplify Subgraph lib function

Alfred-Mountfield:

There was a TODO in the code to get around a temporary solution that was added to ensure correctness. This is now unblocked and we can have a consistent (and faster) implementation to look for incoming links using their edge kinds by directly looking in the edge set, rather than looking through the complete edge set.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203448398358926/f) _(internal)_

## 🔍 What does this change?

See diff, is small

No

## ⚠️ Known issues

None

## 🛡 What tests cover this?

- No automated tests to my knowledge, but the function is used inside `lib/org.ts`, and as demonstrated below, the code using that hook continues to work.

## ❓ How to test this?

1.  Try X
1.  Confirm that Y

## 📹 Demo

image
image

### DIFF
--- a/packages/hash/subgraph/src/stdlib/edge/link.ts
+++ b/packages/hash/subgraph/src/stdlib/edge/link.ts
@@ -6,9 +6,9 @@ import {
   isOutwardLinkEdge,
   isHasRightEntityEdge,
   isHasLeftEntityEdge,
+  isIncomingLinkEdge,
 } from "../../types/edge/outward-edge-alias";
 import { mustBeDefined } from "../../shared/invariant";
-import { OutwardEdge } from "../../types/edge";
 
 const getUniqueEntitiesFilter = () => {
   const set = new Set();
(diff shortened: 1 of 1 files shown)


packages/hash/subgraph/src/stdlib/edge/link.ts
```diff
@@ -98,44 +98,45 @@ export const getIncomingLinksForEntityAtMoment = (
...
+      // Only look at edges that were created before or at the timestamp
+      .filter(([edgeTimestamp, _]) => edgeTimestamp <= timestampString)
+      // Extract the link `EntityEditionId`s from the endpoints of the link edges
+      .flatMap(([_, outwardEdges]) => {
+        return outwardEdges.filter(isIncomingLinkEdge).map((edge) => {
+          return edge.rightEndpoint;
+        });
+      })
+      .map(({ baseId: linkEntityId, timestamp: _firstEditionTimestamp }) => {
+        const linkEntity = mustBeDefined(
```
thehabbos007, line 120: It may be useful to have a message for the `mustBeDefined` here, it is a little cryptic when the throw message doesn't contain anything

packages/hash/subgraph/src/stdlib/edge/link.ts
```diff
@@ -98,44 +98,45 @@ export const getIncomingLinksForEntityAtMoment = (
...
-          (outwardEdge: OutwardEdge) =>
-            isHasRightEntityEdge(outwardEdge) &&
-            outwardEdge.rightEndpoint.baseId === entityId,
+  return (
+    Object.entries(entityEdges)
+      // Only look at edges that were created before or at the timestamp
+      .filter(([edgeTimestamp, _]) => edgeTimestamp <= timestampString)
+      // Extract the link `EntityEditionId`s from the endpoints of the link edges
+      .flatMap(([_, outwardEdges]) => {
+        return outwardEdges.filter(isIncomingLinkEdge).map((edge) => {
```
thehabbos007, line 115: nit: not sure if we want to call this `outwardEdges` in this method

thehabbos007 (approved):
Thanks, Alfie. Tested the app locally, and looked at members of workspace switcher, which seems to be fine. I have some nits, but they don't block merging :)

PR #1492, opened 2022-11-24
Add new entity (from scratch) creation flow

yusufkinatas:
## 🌟 What is the purpose of this PR?

This PR implements the entity creation flow, using this url: `/@{shortname}/entities/new`, and makes `create entity` button at entity-type-editor functional.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203203788735441) _(internal)_
- [Design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH-App?node-id=6067%3A248022&t=OX8sb5WsLKVk6joK-4) _(internal)_

## 🔍 What does this change?

- Improve `generateEntityLabel`
- Add `isPropertyValueArray` & `isPropertyNested` type guards
- Implement create new entity page
- Implement create new entity with pre-selected type
- Extract empty states of property editor sections to separate files (to reuse them)

## ❓ How to test this?

1.  Checkout the branch / view the deployment
1.  Open `http://localhost:3000/@{shortname}/entities/new`
1. Select an entity type from dropdown, and confirm entity creation is working
2. Open the entity type editor for an entity type
3. Click on `create new entity` button on the top right
4. Confirm creating an entity with a pre-selected type works

## 📹 Demo

https://user-images.githubusercontent.com/39553853/203777019-7a9f6c7d-6353-4a3a-90f9-3492bfb2ca9e.mov

PR #1533, opened 2022-11-30
New entity type flow improvements (follow up)

nathggns:

Follow up to https://github.com/hashintel/hash/pull/1492 to improve some of the flow around the new entity type stuff (and updating routes) 

## 🔗 Related links

- N/A

## 🚫 Blocked by

- [x] https://github.com/hashintel/hash/pull/1492

## 🔍 What does this change?

- `useAuthenticatedUser`:
	- Removed `onUserAuthenticated` as no way to implement it properly with current authentication state management
	- Kratos fetching is updated to not refetch it for every use of the hook (this fixes a bug)
- Routes for new pages updated – see https://hashintel.slack.com/archives/C022217GAHF/p1669774923256599 (internal)
- `useCreateNewEntityAndRedirect` is substantially updated
  - Instead of calling `useAuthenticatedUser` itself, the callback it returns now takes the user. This is because it's hard to reason about stale closures:
    - The caller of `useCreateNewEntityAndRedirect` isn't supposed to use its callback until the user has authenticated, so it's likely the parent component is also calling `useAuthenticatedUser`
    - Right now, it's possible that the order of `useCreateNewEntityAndRedirect` vs `useAuthenticatedUser` being called matters (and that the callback will be called with a stale closure). This is kinda hard to explain but happy to try on Tandem, but the approach of passing the user in directly avoids this problem
  - The redirect is now cancellable with an abort signal
  - Now able to use `router.replace` for the redirect – which makes the back button from "create new entity" button on entity type editor work
- The different state management on new entity page is updated to make it easier to follow 
- Home redirect to users dashboard now uses `router.replace` to ensure back button isn't broken

N/A

## ⚠️ Known issues

We're having to use an effect for something that is not strictly an effect, which is frustrating. 

## 🐾 Next steps

Eventually we'll redo state management of authentication, but it's not a priority

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Confirm behaviour is as before

### DIFF
--- a/packages/hash/frontend/src/components/hooks/useAuthenticatedUser.ts
+++ b/packages/hash/frontend/src/components/hooks/useAuthenticatedUser.ts
@@ -128,8 +128,8 @@ export const useAuthenticatedUser = (
   return {
     authenticatedUser,
     kratosSession,
-    refetch: () => {
-      return Promise.all([
+    refetch: () =>
+      Promise.all([
         // Until we change the GraphQL query scalars to return constrained Subgraphs, we need to (safely) cast
         // ourselves
         (
@@ -138,8 +138,7 @@ export const useAuthenticatedUser = (
           >
         )(),
         fetchKratosIdentity(forceLogin, true),
-      ]);
-    },
+      ]),
     loading: loadingUser || loadingKratosSession,
   };
 };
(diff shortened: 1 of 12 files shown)


yusufkinatas (dismissed):
Good work!

CiaranMn (approved):

PR #1534, opened 2022-11-30
Update TypeScript to 4.9

kachkaev:

This PR updates TypeScript from 4.8 to 4.9 in all workspaces except engine. I had to make only one small change for CI to pass – see diff comment.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203429129383886/f) _(internal)_
- [TS 4.9 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/)

## ❓ How to test this?

1.  Check CI output

### DIFF
--- a/packages/hash/frontend/src/pages/entity-editor.page.tsx
+++ b/packages/hash/frontend/src/pages/entity-editor.page.tsx
@@ -16,7 +16,10 @@ import { useBlockProtocolFunctionsWithOntology } from "./type-editor/blockprotoc
  * Helper type-guard for determining if a `ValueOrArray` definition is an array or a value.
  */
 const isArrayDefinition = <T,>(input: ValueOrArray<T>): input is Array<T> =>
-  "type" in input && input.type === "array";
+  input &&
+  typeof input === "object" &&
+  "type" in input &&
+  input?.type === "array";
 
 /**
  * This component is an example usage of the `getEntity` BP function.
--- a/packages/hash/frontend/src/pages/entity-editor.page.tsx
+++ b/packages/hash/frontend/src/pages/entity-editor.page.tsx
@@ -19,7 +19,7 @@ const isArrayDefinition = <T,>(input: ValueOrArray<T>): input is Array<T> =>
   input &&
   typeof input === "object" &&
   "type" in input &&
-  input?.type === "array";
+  input.type === "array";
 
 /**
  * This component is an example usage of the `getEntity` BP function.
--- a/blocks/calculation/package.json
+++ b/blocks/calculation/package.json
@@ -32,7 +32,7 @@
     "mock-block-dock": "0.0.38",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "typescript": "4.8.4"
+    "typescript": "4.9.3"
   },
   "peerDependencies": {
     "react": "^18.2.0",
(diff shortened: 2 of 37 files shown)


packages/hash/frontend/src/pages/entity-editor.page.tsx
```diff
@@ -16,7 +16,10 @@ import { useBlockProtocolFunctionsWithOntology } from "./type-editor/blockprotoc
  * Helper type-guard for determining if a `ValueOrArray` definition is an array or a value.
  */
 const isArrayDefinition = <T,>(input: ValueOrArray<T>): input is Array<T> =>
-  "type" in input && input.type === "array";
+  input &&
+  typeof input === "object" &&
+  "type" in input &&
+  input?.type === "array";
```
kachkaev, line 22: @Alfred-Mountfield does this change make sense (saw you in `git blame`)? I was getting:

```
error TS2322: Type 'ValueOrArray<T>' is not assignable to type 'object'. Type 'T' is not assignable to type 'object'.
```

kachkaev:
> Can we update the files that have TODOs to do with TS 4.9?

I suggest doing this in a separate PR to unblock `satisfies` earlier.

packages/hash/frontend/src/pages/entity-editor.page.tsx
Alfred-Mountfield: Hm I wonder if we want to say `T extends object`, perhaps we don't

Alfred-Mountfield: Anyway I don't believe you need the `?` after `input?` as you already asserted it (can't suggest changes here for some reason)

kachkaev:
@Alfred-Mountfield I removed comments from `packages/hash/subgraph/src/types/identifier.ts`. Do recall any more instances of them? `ts-ignore-expect` is not a valid TS directive, it's `ts-ignore` or `ts-expect`. ESLint was not failing because we don't use `--report-unused-disable-directives` in CLI command. Unused directives are only a warning. I am planning to fix this in an upcoming sprint ([internal Asana task](https://app.asana.com/0/1203363157432081/1203452982053233/f)).

Alfred-Mountfield:
> ts-ignore-expect is not a valid TS directive

Then how did it correctly disable ts compilation warnings? 🤔

kachkaev:
Could those warnings be ESLint-related?

Alfred-Mountfield:
No they were actively failing tsc

packages/hash/frontend/src/pages/entity-editor.page.tsx
kachkaev: I removed `?` in https://github.com/hashintel/hash/pull/1534/commits/5209ef46dd69f52fd7c6cdce494af672cc6469f7, thanks for spotting! This should be flagged as an autofixable ESLint error when we get to enabling `@typescript-eslint/no-unnecessary-condition` ([internal Asana task](https://app.asana.com/0/1201095311341924/1202736501900834/f)).

Happy for you to switch to `T extends object` in a separate PR if you want. I assumed we could have ‘arrays’ of scalars (e.g. numbers or booleans) so did not change the base type.

Alfred-Mountfield: Enable alllll the lints :grin:

This isn't an array like `T[]`, it's an array _definition_ in JSON schema, so something like:
```
{ 
  "type": "array",
  "items":  {
    "$ref": "...",
  }
}

Alfred-Mountfield (approved):

PR #1536, opened 2022-11-30
Fix VersionedUri generation

Alfred-Mountfield:

`VersionedUri` is defined as
```ts
type VersionedUri = `${string}/v/${number}`
```

And BaseUri as 
```ts
type BaseUri = string
```

A valid BaseUri _must have) a trailing slash (but this isn’t encapsulated in a template string literal type, maybe we should do that)

So Typescript was saying that
`${BaseUri}v/${number}` was incorrect, even though it is correct, and was guiding people towards writing `${BaseUri}/v/${number` even though that would result in a type with a double slash `//v/`.

## 🔗 Related links

## 🛡 What tests cover this?

None to my knowledge

## ❓ How to test this?

1.  Please run and test the app manually as usual

### DIFF
--- a/packages/hash/subgraph/src/shared/type-system-patch.ts
+++ b/packages/hash/subgraph/src/shared/type-system-patch.ts
@@ -17,7 +17,7 @@ const versionedUriGroupsPattern = /(.+\/)v\/(\d+)(.*)/;
 
 export const splitVersionedUri = (id: VersionedUri): [BaseUri, number] => {
   const [baseUri, version] = id.match(versionedUriGroupsPattern)!;
-  return [baseUri!, Number(version!)];
+  return [baseUri, Number(version)];
 };
 
 export const versionedUriFromComponents = (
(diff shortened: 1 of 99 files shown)


packages/hash/subgraph/src/shared/type-system-patch.ts
```diff
@@ -15,12 +15,24 @@ import {
 
 const versionedUriGroupsPattern = /(.+\/)v\/(\d+)(.*)/;
 
+export const splitVersionedUri = (id: VersionedUri): [BaseUri, number] => {
+  const [baseUri, version] = id.match(versionedUriGroupsPattern)!;
+  return [baseUri, Number(version)];
```
thehabbos007, line 20: This indexing seems wrong, we would have to have an empty placeholder first to select the capture groups
image
```suggestion
  const [_, baseUri, version] = id.match(versionedUriGroupsPattern)!;
  return [baseUri, Number(version)];
```

thehabbos007: Also I am getting a TSC error for `Type 'string | undefined' is not assignable to type 'string'.` with `baseUri`. We may need an assertion `!`

thehabbos007 (changes requested):
Thanks, Alfie! Seems to work fine for most parts on my end. There is some splitting which we should fix (regex capture group)

thehabbos007 (approved):
Thanks!

PR #1538, opened 2022-11-30
Fix incorrect default value in the GQL schema, make GQL playground accessible in prod

benwerner01:

This PR fixes an incorrect default value in the GQL schema. In addition, it makes the GQL playground available in a production environment.


thehabbos007 (approved):

thehabbos007 (approved):
Thanks, playground page still works locally :)

PR #1544, opened 2022-11-30
Move `Entity::EntityMetadata::LinkEntityMetadata` to `Entity::LinkData`

TimDiekmann:

Similar to `CreatedById`, the link data has to be returned from the `Update` method of entities, but the left/right entity id is immutable across updates, so this requires an additional query, which is not needed at all.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203465545813509/f) _(internal)_

## 🚫 Blocked by

- #1537 

## 🔍 What does this change?

- Rename `LinkEntityMetadata` to `LinkData`
- Move `LinkData` from `Entity::EntityMetadata` to `Entity`


packages/graph/hash_graph/lib/graph/src/lib.rs
```diff
@@ -55,7 +55,6 @@
               usage"
 )]
 #![expect(
-    clippy::cast_sign_loss,
```
Alfred-Mountfield, line 58: How did this disappear?

Alfred-Mountfield (changes requested):
Generally looks good, there are some linting failures, (I assume due to updating the branch). Give me a ping when it's ready again :)

packages/graph/hash_graph/lib/graph/src/lib.rs
TimDiekmann: We don't return the entity metadata from the database when moving an entity to the historic table anymore, so the version of the VersionedUri does not need to be retrieved.

Alfred-Mountfield: Can you point me to the specific line? I tried looking for it but didn't find it

TimDiekmann: In [packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs](https://github.com/hashintel/hash/pull/1544/files#diff-8f613375ad243c59cb8d23f81c2d7602cc3264296def6961c83d976a689e0f84), I removed line 841
```diff
- let entity_type_id = VersionedUri::new(base_uri, historic_entity.get::<_, i64>(4) as u32);
```

This line is only called in `update_entity()`. This function is going to be changed soon btw.

Alfred-Mountfield: Cheers

PR #1555, opened 2022-11-30
Persist the current workspace in local storage

benwerner01:

This PR persists the currently active workspace (the workspace that's currently selected in the workspace switcher dropdown) in local storage, and decouples it's value from what the current account is in the URL.

This means entities/pages/entity types can be viewed in a workspace, wether or not they are owned by the corresponding user/org.

There are several implications to the changes introduced as part of this PR:
- when creating instances of an entity type, they are owned by the active account - not the account that owns the entity type
- when creating a new property type on an entity type, it is owned by the active account - not the account that owns the entity type
- creating a new page/entity type in the navbar creates a new page/entity type in the currently active workspace

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203443872976450/f) _(internal)_

## 🔍 What does this change?

See commits.

Not that I'm aware of.

## ⚠️ Known issues

- I don't understand the distinction between `useRouteNamespace` and `useRouteAccountInfo`, and what the longterm plan for these hooks is. My instinct is these should be merged into a single hook, as they both serve to parse the `/[account-slug]/...` from the current URL

## 🛡 What tests cover this?

Unfortunately, none. Playwright is currently enabled, and is not fully covering the pages that are touched in this PR.

I've done my best to test things locally, but it's possible I may have missed something.

## ❓ How to test this?

1. Log in to the frontend
2. Try switching workspaces
3. Create pages/entities/entity types/property types and ensure the are being created in the currently active workspace

## 📹 Demo

https://user-images.githubusercontent.com/42802102/204879807-7412f4cb-2866-4429-8e99-1895604dbf0a.mov


packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx
```diff
@@ -167,14 +169,17 @@ const Page: NextPageWithLayout = () => {
       console.warn(
         `Error: Couldn't find entity type with id '${router.query["entity-type-id"]}'.`,
       );
-      void router.replace(`/@${namespace?.shortname}/types/new/entity-type`);
+      void router.replace(
+        `/@${activeWorkspace?.shortname}/types/new/entity-type`,
+      );
```
Comment (benwerner01) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx:174]: @nathggns I'm not sure I've understood this correctly, why would we want to redirect to the create entity type page here if the `namespace` couldn't be found from the URL? Wouldn't we want to display a 404 instead?

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx
```diff
@@ -92,21 +92,22 @@ export const PropertyTypeForm = ({
...
       throw new Error("User shortname must exist");
     }
 
     return generateBaseTypeId({
       domain: frontendUrl,
-      namespace: routeNamespace.shortname,
+      namespace: activeWorkspace.shortname,
       kind: "property-type",
       title: value,
     });
```
Comment (benwerner01) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: @nonparibus @CiaranMn @nathggns the implications of this are that if the user creates a property type on an entity type, it's no longer necessarily owned by the same account. It will be owned by whatever is the currently active workspace. This seemed most consistent with what the task outlines, but want to flag this implication nonetheless.

Entity types having property types that are owned by different accounts to my knowledge should not cause any issues on the backend (cc @hashintel/backend).

Comment (nonparibus) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: Great, and that is definitely a thing we want to support.

Thanks for flagging this.

We will give thought to the best UX for managing actions (such as creation) across multiple workspaces as we will want to introduce specific switchers -- or at least notices -- throughout the UI (besides the sidebar switcher) eventually. But we want to be consistent in how we do this and have not yet settled on best practice.

packages/hash/frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
```diff
@@ -59,7 +60,7 @@ export const PageSidebar: FunctionComponent = () => {
       <TopNavLink
         icon={faHome}
         title="Home"
-        href="/"
+        href={`/${activeWorkspaceAccountId}`}
```
benwerner01, line 63: Currently the workspace homepage is hosted at `/<workpace-account-id`, not `/`. This is probably something we want to change, but for consistency I've reflected the current state here.

packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx
```diff
@@ -43,6 +43,13 @@ export const WhiteCard = ({
           : {},
       ]}
     >
+      {/**
+       * @todo: refactor this to use `next/link` when a relative URL is passed
+       * into as the `href`, to avoid a flashing white screen when the user
+       * clicks on the entity's type.
+       *
+       * @see https://app.asana.com/0/1203179076056209/1203468350364504/f
+       */}
```
Comment (benwerner01) [inline packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx:52]: @yusufkinatas not sure if you were planning to address this in existing work, but just wanted to flag that I've created [a task](https://app.asana.com/0/1203179076056209/1203468350364504/f) for this in the meantime.

Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx:52]: Thanks Ben, converted it to a subtask of "entity editor" task.

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: I agree we want to support this but it seems like a strange default. If I'm editing an entity type in a particular workspace (which I must be in order to create a property type inline and add it to the type), I think it's definitely inintuitive that taking any action in editing the entity type is going to create things in another workspace.

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx]: https://github.com/hashintel/hash/pull/1527 changes this behavior

CiaranMn (dismissed):
Looking at this PR made me think of a few cleanup things we can do to the codebase, but none of them need to block this PR. I've made one non-blocking suggestion.

For the record the cleanup ideas are:
- figure out terminology between 'account' vs 'workspace' vs 'namespace' which we use interchangably in places
- consolidate `useRouteNamespace` and `useAccountInfo` hooks as you have suggested separately
- migrate `[account-slug]/index.page.tsx` and `[account-slug]/[page-slug].page.tsx` to use an account slug, not id, in the first URL segment

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: To add a property type to an entity type, the user must already have write access to the entity type's owner. Therefore I can easily revert this to default to the workspace the entity type belongs to. Let me know what your thoughts are @nonparibus.

packages/hash/frontend/src/pages/[account-slug]/index.page.tsx
```diff
@@ -10,7 +10,12 @@ import { useRouteAccountInfo } from "../../shared/routing";
 const Page: NextPageWithLayout = () => {
   const { loading, authenticatedUser } = useAuthenticatedUser();
   const { orgs } = useOrgs();
-  const { accountId } = useRouteAccountInfo();
+
+  /**
+   * @todo: this will need to be reworked once pages can't rely on the `accountId` being
+   * in the URL.
+   */
+  const { routeAccountSlug } = useRouteAccountInfo();
```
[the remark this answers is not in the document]
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/index.page.tsx]: Fair enough, I've addressed both in https://github.com/hashintel/hash/pull/1555/commits/9bbfef6109fe54d3e224136218677f43ecdf39c3

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx
Comment (nonparibus) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: Thinking about this, a user is never going to be able to edit an entity type unless they have the workspace it is part of selected in the sidebar.

If they are browsing as any other workspace (including their own user account) they will just see a view page when they visit the entity type.

Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: > If they are browsing as any other workspace (including their own user account) they will just see a view page when they visit the entity type.

To be clear that's not currently the case, but if it will be I suppose `activeWorkspace` in this code snippet will always be the entity type's owner's workspace.

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page.tsx]: Ah great, thanks for flagging Ciaran :)

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx:113]: Can we just make it the entity type owner's ten? 

We might decide that it's annoying to have to switch workspaces to edit things that you own (and we haven't implemented it anyway), but I don't think we will ever want the default in this case to be anything other than the entity type owner's. Just fixing it at that avoids having the odd behavior (a) while we haven't implemented 'must switch workspace to edit' and (b) if we ever decide to revert that / not to introduce it.

Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/property-type-form.tsx]: I think that makes sense, have reverted the changes in https://github.com/hashintel/hash/pull/1555/commits/01bc0e08e1e0b342b00399bf13afb32dad3d1a59

CiaranMn (approved):

nathggns:
As flagged on Slack, @benwerner01 only wanted one frontend review on this, so removing further requests as its been approved by @CiaranMn

PR #1399, opened 2022-11-15
new `append-vec` crate

indietyp:
## 🌟 What is the purpose of this PR?

This PR introduces a new crate `append-vec`, a read optimized concurrently accessible (for reads) vector that can only be pushed to.

### Reasoning

This crate originates from the needs of `deer` for a thoroughly tested and verified append only vector that is available on no-std systems. Previous crates like `scc`, `appendlist`, `intrusive-collections`, `sento`, or `append-only-vec` do not fit those requirements as they are only available on `std`, are not concurrently accessible or are poorly tested. This crate tries to remedy all those problems.

## 🔗 Related links

- [Asana task](link) _(internal)_

## 🔍 What does this change?

* introduce a new crate, `append-vec`

## 📜 Does this require a change to the docs?

Documentation is still needed, as this is a completely new crate.

## ⚠️ Known issues

Currently `append-vec` has no `std` option and uses spinlock either way, one could remove the spinlock and replace it with a fully blown `Mutex` on `std`, tho benchmarks have shown that due to speed of push, impact due to spinning is small.

## 🐾 Next steps

* Documentation
* Tests
* Infrastructure (CI)

> The commits are cherry-picked from #1396 via `git format-patch -k --stdout bm/deer/error-visitor...bm/deer/error-hooks -- append-vec | git am -3 -k`
> See https://stackoverflow.com/a/63745988/9077988 for more information.

## 📹 Demo

Coming Soon™

PR #1556, opened 2022-11-30
`error-stack` support `Debug` hooks in `no-std` contexts

indietyp:

While exploring #1399 we found the [spin crate](https://crates.io/crates/spin), which implements a `no-std` equivalent for `std::sync` types, like `RwLock` based on spin locks. This PR adds a new feature, `hooks`, which - if enabled - will replace the `RwLock` with the spin variant.

Note: while we expect a majority of calls to be read-only, there's still a chance of priority inversion, this is why if both `std` and `hooks` features are enabled, we will instead use the `std` `RwLock`.

> Must read for spin locks and their application: https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html

I consider using a spinlock okay for `error-stack` hooks, as we expect hooks to only be initialized very, very infrequently, and most operations are reads. A read is simply a `fetch_add` + `fetch_sub` without any spinning.

An alternative approach would be using [`critical-section`](https://crates.io/crates/critical-section) instead, which acts like a `Mutex`, but because we mostly only read, we do not need to hold the hooks exclusively, and `critical-section` is only available on single thread embedded microprocessors.

> **Note:** this increases our test permutation count to 81!

## 🔗 Related links

- https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html

## 🔍 What does this change?

- new feature `hook`

## 📋 TODO

- [x] Documentation
- [x] Changelog

We need to mention the new feature in our table and the consequences.

### DIFF
--- a/packages/libs/error-stack/Cargo.toml
+++ b/packages/libs/error-stack/Cargo.toml
@@ -22,7 +22,7 @@ anyhow = { version = "1.0.65", default-features = false, optional = true }
 eyre = { version = "0.6", default-features = false, optional = true }
 owo-colors = { version = "3", default-features = false, optional = true, features = ['supports-colors'] }
 serde = { version = "1", default-features = false, optional = true }
-spin = { version = "0.9.4", default-features = false, optional = true, features = ['rwlock', 'once'] }
+spin = { version = "0.9", default-features = false, optional = true, features = ['rwlock', 'once'] }
 
 [dev-dependencies]
 serde = { version = "1.0.148", features = ["derive"] }
--- a/packages/libs/error-stack/Cargo.toml
+++ b/packages/libs/error-stack/Cargo.toml
@@ -22,6 +22,7 @@ anyhow = { version = "1.0.65", default-features = false, optional = true }
 eyre = { version = "0.6", default-features = false, optional = true }
 owo-colors = { version = "3", default-features = false, optional = true, features = ['supports-colors'] }
 serde = { version = "1", default-features = false, optional = true }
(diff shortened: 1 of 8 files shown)


packages/libs/error-stack/Cargo.toml
```diff
@@ -22,6 +22,7 @@ anyhow = { version = "1.0.65", default-features = false, optional = true }
 eyre = { version = "0.6", default-features = false, optional = true }
 owo-colors = { version = "3", default-features = false, optional = true, features = ['supports-colors'] }
 serde = { version = "1", default-features = false, optional = true }
+spin = { version = "0.9.4", default-features = false, optional = true, features = ['rwlock', 'once'] }
```
TimDiekmann, line 25: Does `0.9` also work for our use cases?

TimDiekmann (commented):
LGTM, thank you!

Do we need to adjust the Makefile to also test `hooks` on `insta` checks?

One other question, then I'm happy to approve

indietyp:
Thanks! I completely forgot about the snapshot tests, I have adjusted the Makefile and test cfg guard

packages/libs/error-stack/Cargo.toml
indietyp: requirement relaxed in [122a119](https://github.com/hashintel/hash/pull/1556/commits/122a119a12021b17f17782443d0422b47826f166)

TimDiekmann: Great, thanks!

TimDiekmann (approved):
Excellent, thank you, and also for providing the link to the blog post.

PR #1558, opened 2022-11-30
GEN-35: `error-stack` implement `serde` hooks

indietyp:

This implements the equivalent of `Debug` hooks, but for the serde output, we're unable to know if a type implements `Serialize directly`; therefore, we need to register them via hooks.

## 🔍 What does this change?

Report gains 2 new functions:

```
impl Report<()> {
	pub fn install_serde_hook<T: serde::Serialize>();
	pub(crate) fn invoke_serde_hook<T>(closure: impl FnOnce(&Hooks) -> T) -> T;
}
```

(potentially, we could also provide a way to change how a type is serialized via: `fn install_serde_custom_hook<T: serde::Serialize>(hook: impl Fn(&T) -> Box<dyn erased_serde::Serialize>)` or something similar).

Yes, we need to update the docs to reflect the new capabilities.

## 🏃 Next Steps

1. The fallback format hooks currently violate the contract described in the documentation when used in serde hooks. The `fmt::HookContext` is currently constructed for every separate invocation. This should instead be persisted (like how we currently carry forward the `serde::HookContet`)
2. Hooks are not invoked on `Context` objects because the current representation does not allow multiple values for a `Context`. Hooks can return more than a single serialized object. Further investigation 🕵️  is needed.
3. Default hooks are currently not invoked if `std` or `hooks` features are missing, behaviour of `fmt` hooks should be mirrored
4. Documentation is largely missing
5. (maybe, if possible): `DynamicFn` is overly restrictive regarding lifetime. Closures need to have the lifetime `for<'a> Fn(&'a T &mut HookContext<T>) -> U + 'a`. This is fine for functions but routinely fails for closures, as we cannot explicitly specify lifetimes on closures. Another possibility would be to relax further the lifetime to: `for<'a: 'b, 'b> Fn(&'a T &mut HookContext<T>) -> U + 'b`, but currently, this isn't possible because one cannot set lifetime bounds in HRTB, another approach might - if possible - be necessary.

## 📹 Demo

https://github.com/hashintel/hash/blob/aaca87f865b299d7b33c55beef1eb6f92b0ae29a/packages/libs/error-stack/tests/snapshots/test_serialize__sources_nested%40spantrace.snap#L5-L363

## 💡 Additional Information

`DynamicFn` is overly restrictive. This issue can be addressed using the nightly feature [`closure_lifetime_binder`](https://github.com/rust-lang/rust/issues/97362), which enables us to specify all lifetimes explicitly. It is more verbose than regular closures (and CLion hates it), but it solves the problem. I provided a test case (showcased below) for more context. This resolves our issue because taking `Fn(&'a T): U`, Rust infers (for regular closures) that, if references are taken from `T`, `U` might have a lifetime _less_ than `'a`, using HRTBs we're unable to create a trait that reflects this behavior as far as I know (this is because HRTBs cannot define lifetime bounds). The experimental feature lets us circumvent the issue by explicitly stating that `U` will have a lifetime of `'a`. I have attached a screenshot of the error.

https://github.com/hashintel/hash/blob/43c600b2e5b127ad4c2bbc663c1e9b224041cac7/packages/libs/error-stack/tests/test_serialize.rs#L136-L145

image


packages/libs/error-stack/src/hook/mod.rs
```diff
@@ -16,6 +16,8 @@ type RwLock<T> = std::sync::RwLock<T>;
 type RwLock<T> = spin::rwlock::RwLock<T>;
 
 static FMT_HOOK: RwLock<Hooks> = RwLock::new(Hooks { inner: Vec::new() });
```
TimDiekmann, line 18: We probably also fully qualify the path to `Hooks` here as well or import them as `FmtHooks`. Actually, as they are internal only, I'd prefer renaming both `Hooks` structs to `FmtHooks` and `SerdeHooks`

packages/libs/error-stack/src/hook/mod.rs
```diff
@@ -181,4 +183,50 @@ impl Report<()> {
 
         closure(&hook)
     }
+
+    #[cfg(all(any(feature = "std", feature = "hooks"), feature = "serde"))]
+    pub fn install_serde_hook<T: serde::Serialize + Send + Sync + 'static>() {
+        crate::serde::install_builtin_hooks();
```
TimDiekmann, line 189: Same as above, could we also fully qualify the debug hooks or rename the functions?

packages/libs/error-stack/src/serde/hook.rs
```diff
@@ -0,0 +1,360 @@
...
+) -> Option<Box<dyn erased_serde::Serialize + 'a>> {
+    #[cfg(nightly)]
+    let value: &T = frame.request_ref()?;
+    #[cfg(not(nightly))]
+    let value: &T = frame.downcast_ref()?;
+    Some(Box::new(value))
+}
+
+// Thanks to https://users.rust-lang.org/t/hrtb-on-multiple-generics/34255/2 for the solution
+pub trait DynamicFn<'a, I>: Send + Sync + 'static
```
TimDiekmann, line 23: `Dynamic` does not really say anything about the implied meaning. Maybe we should rename it to something like `SerializeFn`?

packages/libs/error-stack/src/serde/hook.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    fn call(&self, value: &'a I, context: &mut HookContext<I>) -> Self::Output;
+}
+
+impl<'a, U: 'a, T, F> DynamicFn<'a, T> for F
+where
+    F: Fn(&'a T, &mut HookContext<T>) -> U,
+    F: Send + Sync + 'static,
+    U: serde::Serialize,
+    T: Send + Sync + 'static,
```
TimDiekmann, line 37: Is this the only implementation for `DynamicFn`? If yes do you see a way to directly use `Fn` instead without introducing the trait?

packages/libs/error-stack/src/hook/mod.rs
```diff
@@ -181,4 +183,50 @@ impl Report<()> {
...
+
+        // The spin RwLock cannot panic
+        #[cfg(all(not(feature = "std"), feature = "hooks"))]
+        let mut lock = SERDE_HOOK.write();
+
+        lock.insert_static::<T>();
+    }
+
+    #[cfg(all(any(feature = "std", feature = "hooks"), feature = "serde"))]
+    pub fn install_custom_serde_hook<F, T>(closure: F)
```
TimDiekmann, line 202: Is it possible to use `impl Trait` syntax for `F`, so we can call `install_custom_serde_hook::<MyType>(...)`?

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
```diff
@@ -0,0 +1,43 @@
...
+[
+  {
+    "context": "root error",
+    "attachments": [
+      {
+        "file": "tests/common/mod.rs",
+        "line": 4,
+        "column": 5
+      },
+      "backtrace (1)",
```
TimDiekmann, line 14: Can we/do we want to serialize backtraces as well?

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
```diff
@@ -0,0 +1,43 @@
...
+---
+[
+  {
+    "context": "root error",
+    "attachments": [
+      {
+        "file": "tests/common/mod.rs",
+        "line": 4,
+        "column": 5
+      },
```
TimDiekmann, line 13: By looking at the structure it's clearly a location, but do we somehow want to tag that? Or is that too opinionated? cc @Alfred-Mountfield

TimDiekmann (changes requested):
Generally, this looks very good! Thank you!
I have a few questions and suggestions, mainly about the public API.

packages/libs/error-stack/src/serde/hook.rs
indietyp: Yes, `DynamicFn` (I believe I have a comment referring to the post) is needed to specify lifetimes for the return type, which depends on the first argument (using HRTB) for the return type. Otherwise, the return type must be `'static`, which means that custom hooks need to return owned values, unable to borrow from the downcasted/requested value. Without the trait, we're unable to do that ._.

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
indietyp: We could serialize them as a string. Returning the backtrace structurally is a bit more complicated, because we'd likely need to regex match or *somehow* convert/use the backtrace crate.

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
Alfred-Mountfield: If we're expecting quite a few similar attachment things, I think it'd be nice to have a `kind` field. If we're not expecting that, then I think this is fine

indietyp: We could do:

```json
{
	"kind": "...",
	"value": {...}
}
```

The main problem is the `"kind"` value. We could use `TypeId` (but that value isn't stable across recompilations and is only valid for `'static` types), and another one would be https://doc.rust-lang.org/std/any/fn.type_name.html.

Alfred-Mountfield: I thought we're not deriving `Serialize`? So we could just choose the value of `"kind"` and call it `Location` or something, no?

Alfred-Mountfield: Again I'm not sure this is worth it unless it's a pattern we're establishing for things

indietyp: Yes, but ideally - if this were a pattern - we would automatically assign a `kind` to every attachment

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
TimDiekmann: I think, as long as `Serialize` is not implemented on `Backtrace` upstream, we should serialize it as string. If an upstream implementation is provided at some point we can switch to that.

packages/libs/error-stack/src/serde/hook.rs
indietyp: fixed in [0941b84](https://github.com/hashintel/hash/pull/1558/commits/0941b84cbc9939518d3f5ff1e6fabf6e89a0ea83)

packages/libs/error-stack/src/hook/mod.rs
indietyp: fixed in [83edbf2](https://github.com/hashintel/hash/pull/1558/commits/83edbf2a5191c8c8b38615601044655aea37277f)

packages/libs/error-stack/src/hook/mod.rs
indietyp: fixed in [83edbf2](https://github.com/hashintel/hash/pull/1558/commits/83edbf2a5191c8c8b38615601044655aea37277f)

packages/libs/error-stack/tests/snapshots/test_serialize__full__attachment.snap
indietyp: fixed in [efc3db4](https://github.com/hashintel/hash/pull/1558/commits/efc3db4212624c1564e5de5d024f6fca916a7ede)

indietyp:
Converting into draft, because everything is broken right now and the large change in formatting for 0.3 need to be applied to this PR

vilkinsons:
Closing this PR following internal discussion.

vilkinsons:
`deer` will be moved out to our [`labs` repository](https://github.com/hashintel/labs) until it is in a usable state, and this PR should be recreated there.

PR #1561, opened 2022-11-30
fix resolving comments in org workspaces

luisbettencourt:

Trying to resolve a comment in a org workspace throwed an error. This PR fixes that.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203466351235289/f) _(internal)_


packages/hash/frontend/src/blocks/page/Comments/CommentThread.tsx
```diff
@@ -73,6 +73,11 @@ export const CommentThread: FunctionComponent<CommentThreadProps> = ({
     [comment.author.properties],
   );
 
+  const authorId = useMemo(
+    () => comment.author.metadata.editionId.baseId.split("%")[1],
```
luisbettencourt, line 77: Is there a helper function/better way to get this?

Alfred-Mountfield: packages/hash/subgraph/src/types/identifier.ts:27

`extractOwnedByIdFromEntityId`

Alfred-Mountfield (commented):
I think it looks good to me, give me a re-request when you use the helper function :)

packages/hash/frontend/src/blocks/page/Comments/CommentThread.tsx
luisbettencourt: Thank you! Fixed in fffd735ae33f2407da48ad82fe5a791edd9f7285.

Alfred-Mountfield (approved):
Nice! 

Test it locally by writing a comment as Alice, resolve button appeared. Swapped workspace to the org and things continued to work as expected. Then I logged out and signed in as Bob. Comments were still there but no resolve button (as expected). Opened another thread and could resolve that.

More importantly, I also made a comment _as Alice_ _on Bob's_ block, signed back in as Bob, and was able to resolve Alice's thread. Nice!

PR #1564, opened 2022-12-01
Use `next/link` to redirect to the entity's type in the entity editor

benwerner01:

This PR ensures that the Next JS link component is used when redirecting from an entity to it's entity type in the entity editor. This prevents a white screen from flashing between the pages, as the page isn't fully refreshed.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203468350364504/f) _(internal)_

## 🔍 What does this change?

See commits.

Not that I'm aware of.

## ⚠️ Known issues

- sometimes when clicking on the entity's type, the user is redirected to the create new entity type page

## 🛡 What tests cover this?

None, tested manually.

## ❓ How to test this?

1. Log in to the frontend
2. Go to the entity editor
3. Click on it's type
4. Ensure the page is not refreshed during the page transition

## 📹 Demo

Before:

https://user-images.githubusercontent.com/42802102/205073739-75d044a0-fa8a-4dd7-ba28-070d4a5975f8.mov

After:

https://user-images.githubusercontent.com/42802102/205073245-48d5c99d-942b-408c-a02d-a3cc37fa2806.mov

### DIFF
--- a/packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx
+++ b/packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx
@@ -6,6 +6,28 @@ import {
   CardContent,
   CardContentProps,
 } from "@mui/material";
+import { FunctionComponent, ReactNode } from "react";
+import { Link } from "../../../shared/ui/link";
+
+const WhiteCardActionArea: FunctionComponent<
+  { children: ReactNode } & CardActionAreaProps
+> = ({ children, ...props }) => (
+  <CardActionArea
+    {...props}
+    disableRipple
+    disableTouchRipple
+    sx={[
+      {
+        [`&:hover .${cardActionAreaClasses.focusHighlight}`]: {
+          opacity: 0,
+        },
+      },
+      ...(Array.isArray(props.sx) ? props.sx : [props.sx]),
+    ]}
+  >
+    {children}
+  </CardActionArea>
+);
 
 export const WhiteCard = ({
   onClick,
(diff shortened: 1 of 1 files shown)


nonparibus:
@benwerner01 This is better, but the whole loading UX could still be improved further via introduction of a loading skeleton on the entity type frame (happy for this to follow in a separate PR). I thought we already had this, but I guess not app-wide: https://github.com/hashintel/hash/search?q=skeleton

packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx
```diff
@@ -43,19 +65,14 @@ export const WhiteCard = ({
...
-        >
+      {onClick ? (
+        <WhiteCardActionArea onClick={onClick}>
           {cardContent}
-        </CardActionArea>
+        </WhiteCardActionArea>
+      ) : href ? (
+        <Link href={href} sx={{ textDecoration: "none" }}>
+          <WhiteCardActionArea>{cardContent}</WhiteCardActionArea>
+        </Link>
```
Comment (nathggns) [inline packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx:75]: Have you considering using the `component` prop to `CardActionArea` to change the component from `"a"` to a `Link` when passing href? Seems a simpler solution. 

`CardActionArea` is based on `ButtonBase`, and this is how we make Buttons use `Link` elsewhere.

Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/shared/white-card.tsx]: Great suggestion, I didn't think of that. Have addressed in https://github.com/hashintel/hash/pull/1564/commits/bfd0ca1118fc5d97364de6e0c42c4db9d483a3e8

nathggns (approved):

PR #1513, opened 2022-11-28
Array property type creation

luisbettencourt:
## 🌟 What is the purpose of this PR?

This PR focuses on implementing the logic for creating array property types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203301853897838/1203366629132085/f) _(internal)_

## 🔍 What does this change?

- Added menus for creating a custom property type (only arrays in this PR);
- Moved some of the form logic on `property-type-form` to `property-utils`;
- Moved the dropdown on `property-type-selector-dropdown` to a generic component so it can be reused.

## 🚫 Blocked by

- [x]  #1527 
- [x]  #1521

## 🐾 Next steps

- Fix Modal positioning;
- Finish styling;
- Other subtasks [here](https://app.asana.com/0/1203301853897838/1203343922163681/f).

## ❓ How to test this?

1.  Start creating an entity type;
2. Add a custom property type;
3. Add an array (and nested properties) to the custom property type;
4. Save the property type;
5. Check if it was correctly saved in the database.

PR #1572, opened 2022-12-02
Property type form submission fix, autocomplete improvements

CiaranMn:

Fixes an issue whereby submitting the property type form (e.g. to create a new property type) would also submit the surrounding entity type's form if it had any draft changes.

While not nested in the HTML (as the property type form is in a modal), they are in the same React tree.

This also changes two things about the 'select a type' dropdown:
1. Makes it so that its container isn't taking up more width than it does (the click target previously included the empty space to the right)
2. Makes it close on select (since the majority of properties will be single type) – will need to confirm this doesn't mess with #1513 once that is merged

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203448542966678) _(internal)_

## ❓ How to test this?

1.  Have an entity type with unsaved changes (E.g. toggle 'required')
2. Open the 'create a property type form'
3. Create a new property type, check that it hasn't automatically saved the entity type (incremented the draft version)


yusufkinatas (approved):
Tested on my local, works as expected.

luisbettencourt:
Im resolving merge conflicts caused by my #1513

yusufkinatas (approved):

PR #1573, opened 2022-12-02
Reset the active workspace on login/logout and when local storage has unexpected value

benwerner01:

This PR fixes a bug where the active workspace is not set correctly when there is an unexpected/outdated value in local storage for `workspaceAccountId`. This PR addresses this by:
- explicitly setting the `activeWorkspaceAccountId` on login to the logged in user's `accountId`
- resetting `activeWorkspaceAccountId` to the logged in user's `accountId` whenever it has an unexpected value
- resetting the local storage values on logout

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203476705034112/f) _(internal)_

## 🔍 What does this change?

See commits/prior description.

Not that I'm aware of.

## 🛡 What tests cover this?

None, manually tested.

## ❓ How to test this?

1. Login and ensure the active workspace is correct in the dropdown
2. Update the value for `activeWorkspaceAccountId` in local storage to something invalid (like `"not-an-account-id"`)
3. Refresh the page, and verify the dropdown is not broken and the value for `activeWorkspaceAccountId` is reset to the logged in user's account ID

## 📹 Demo

https://user-images.githubusercontent.com/42802102/205269079-46e29d24-8582-4b5e-8d69-01f745f53f04.mov


yusufkinatas (approved):
Tested on my local, works as expected. Good UX improvement!

PR #1575, opened 2022-12-02
Frontend fixes: Formatting buttons. User mention caching. Entity labels

CiaranMn:

A few fixes to things I noticed while playing with the app, in the order they appear in the diff:

### 1. Cache the 'fetch all users' in the user mention hook, to stop loading state every time a mention is displayed

**Before**
Kapture 2022-12-02 at 15 11 47

**After**
Kapture 2022-12-02 at 15 12 19

---

### 2. Remove a background class from the formatting toolbar that was overwriting the class that is meant to apply when the mark is active, making the mark text invisible

**Before**
Screenshot 2022-12-02 at 15 05 23

**After**
Screenshot 2022-12-02 at 15 05 38

---

### 3. Fix the labels for entities in the entities tab on the entity type page (i.e. for a type, list entities belonging to it) – this was previously using whatever the root of the subgraph was, rather than the entity in question

**I'm not sure why the root of the subgraph was a `Block` entity?!**

**Before**
Screenshot 2022-12-02 at 15 04 32

**After**
Screenshot 2022-12-02 at 15 04 40

### DIFF
--- a/packages/hash/frontend/src/blocks/page/MentionView/MentionDisplay.tsx
+++ b/packages/hash/frontend/src/blocks/page/MentionView/MentionDisplay.tsx
@@ -21,7 +21,7 @@ export const MentionDisplay: FunctionComponent<MentionDisplayProps> = ({
   mentionType,
   accountId,
 }) => {
-  const { users, loading: usersLoading } = useUsers();
+  const { users, loading: usersLoading } = useUsers(true);
   const { data: pages, loading: pagesLoading } = useAccountPages(accountId);
 
   const { title, href, icon } = useMemo(() => {
--- a/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx
+++ b/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx
@@ -73,7 +73,7 @@ export const useEntitiesTable = (
 
     const rows: TypeEntitiesRow[] =
       entities?.map((entity) => {
-        const entityLabel = generateEntityLabel(subgraph);
+        const entityLabel = generateEntityLabel(subgraph, entity);
         const entityNamespace = getEntityByEditionId(
           subgraph,
           entity.metadata.editionId,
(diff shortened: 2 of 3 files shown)


packages/hash/frontend/src/blocks/page/MentionView/MentionDisplay.tsx
```diff
@@ -21,7 +21,7 @@ export const MentionDisplay: FunctionComponent<MentionDisplayProps> = ({
   mentionType,
   accountId,
 }) => {
-  const { users, loading: usersLoading } = useUsers();
+  const { users, loading: usersLoading } = useUsers(true);
```
Alfred-Mountfield, line 24: I think this is worth doing for now, but I do wonder about the implications of doing this generally. Especially if mentions are expanded to include more than users and pages, if I just made an entity in my page as a result of an action, I'd want it to appear in the mentions (or whatever analogue there is) selector.

I think the longer term solution is making these queries fast enough that it doesn't matter, or to invalidate the cache relatively frequently maybe.

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx
```diff
@@ -73,7 +73,7 @@ export const useEntitiesTable = (
 
     const rows: TypeEntitiesRow[] =
       entities?.map((entity) => {
-        const entityLabel = generateEntityLabel(subgraph);
+        const entityLabel = generateEntityLabel(subgraph, entity);
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx:76]: I believe this was appearing as a block because the `Subgraph` comes from [this call](https://github.com/hashintel/hash/blob/e2b8fff388244a1d106e9825e373a66bf21f5ac7/packages/hash/frontend/src/pages/%5Baccount-slug%5D/types/entity-type/use-entity-type-entities.tsx#L29-L45)

That call doesn't constrain based on the entity-type (maybe we should do that?), so I assume the first root was a Block by coincidence of the query

Alfred-Mountfield (approved):

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx
Comment (Alfred-Mountfield) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/use-entities-table.tsx:76]: Ah I remember why we can't do that presently. I [raised it in Slack](https://hashintel.slack.com/archives/C022217GAHF/p1669723758095149), although so far no responses.

packages/hash/frontend/src/blocks/page/MentionView/MentionDisplay.tsx
CiaranMn: Agreed – ideally this would all go through collab

PR #1577, opened 2022-12-02
Avoid more unneeded queries

TimDiekmann:

It's possible to avoid more queries to the database, this will increase the performance by up to 95%!!!1!11!
The effect is most probably most noticeable when having smaller depths.


Alfred-Mountfield (approved):
🚀 🔥 

It's unfortunate that we missed such an intense regression 😅 . Really emphasises how important it will be to get benchmarks running for CI (as planned)

This is a fantastic change though, I've just checked out the branch and a lot of the actions feel _**noticeably**_ faster.

PR #1579, opened 2022-12-02
Creating Array Data Types UX improvements 

luisbettencourt:

This PR improves the UX when creating array data types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203481015428204/f) _(internal)_

## 🔍 What does this change?

- Prevent popper from open to the side;
- Trigger a window resize event whenever a new data type is added/removed so that the popper component recalculates its position;
- Add expand/collapse animations to data types when adding/removing them from the schema;
- Add expand/collapse width animation to the prefix ("CONTAINING", "OR", etc...) container.

## 📹 Demo

https://www.loom.com/share/efa8cb19fc414afb9a0e36efed612c66

### DIFF
--- a/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx
+++ b/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx
@@ -10,7 +10,7 @@ import {
   tooltipClasses,
   Typography,
 } from "@mui/material";
-import { useState } from "react";
+import { useState, useRef, useEffect } from "react";
 import { dataTypeData } from "./property-type-utils";
 
 export interface DataTypeBadgeProps {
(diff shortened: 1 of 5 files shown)


packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx
```diff
@@ -47,9 +55,12 @@ export const DataTypeBadge = ({
             padding: 1.25,
             alignItems: "center",
             ...(isArray ? { borderBottomLeftRadius: 0 } : {}),
+            width: prefixContainerWidth,
+            transition: ({ transitions }) => transitions.create("width"),
```
Comment (benwerner01) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx:59]: I think this transition means it's possible to get into an odd state where the chip is jumping from line to line on hover. I think some right padding that is minimised when the width is expanded could address this.

https://user-images.githubusercontent.com/42802102/205361161-97d382e4-78c6-4082-a1e4-8ee89ae2440b.mov

benwerner01 (commented):
The changes look good to me, spotted one minor visual bug that can occur in an edge-case.

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx:59]: This could also be addressed in follow-up.

Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/data-type-badge.tsx:59]: Will do this in a follow-up

benwerner01 (approved):

PR #1580, opened 2022-12-02
Entity namespaces / page titles / UX tweaks

CiaranMn:

A few miscellaneous improvements:
1. Correct `namespace` on the entities tab on the type editor (introduces a new hook, `useGetOwnerForEntity`)
2. Close the edit property modal after saving
3. Remove misleading and non-functioning 'open in new icon' from dropdown entries
4. Add HTML titles to entity type and entity editing pages. Make page title consistent with them

### DIFF
--- a/packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
+++ b/packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
@@ -1,0 +1,44 @@
+import { Entity, extractOwnedByIdFromEntityId } from "@hashintel/hash-subgraph";
+import { useCallback } from "react";
+import { nilUuid } from "@hashintel/hash-shared/types";
+import { useUsers } from "./useUsers";
+import { useOrgs } from "./useOrgs";
+
+export const useGetOwnerForEntity = () => {
+  /*
+   * This is a simple way of getting all users and orgs to find an entity's owner's name
+   * @todo rethink caching here – users and orgs added since session start won't appear
+   * @todo probably replace this with something like fetching owners individually instead
+   */
+  const { users = [] } = useUsers(true);
+  const { orgs = [] } = useOrgs(true);
+
+  return useCallback(
+    (entity: Entity) => {
+      const ownerUuid = extractOwnedByIdFromEntityId(
+        entity.metadata.editionId.baseId,
+      );
+
+      const owner =
+        // @todo remove this hack when User and Orgs are owned by a real Org
+        ownerUuid === nilUuid
+          ? { orgAccountId: nilUuid, shortname: "system" }
+          : users.find((user) => ownerUuid === user.userAccountId) ??
+            orgs.find((org) => ownerUuid === org.orgAccountId);
+
+      if (!owner) {
+        throw new Error(
+          `Owner with accountId ${ownerUuid} not found – possibly a caching issue if it has been created mid-session`,
+        );
+      }
+
+      const isUser = "userAccountId" in owner;
+
+      return {
+        accountId: isUser ? owner.userAccountId : owner.orgAccountId,
+        shortname: owner.shortname ?? "",
+      };
+    },
+    [orgs, users],
+  );
+};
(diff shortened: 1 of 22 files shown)


packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page.tsx
```diff
@@ -93,48 +94,53 @@ const Page: NextPageWithLayout = () => {
   const entityLabel = generateEntityLabel(entitySubgraph);
 
   return (
-    <EntityPageWrapper label={entityLabel}>
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page.tsx:96]: Hide whitespace – only the `<Head>` has been added

packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
```diff
@@ -0,0 +1,44 @@
...
+  return useCallback(
+    (entity: Entity) => {
+      const ownerUuid = extractOwnedByIdFromEntityId(
+        entity.metadata.editionId.baseId,
+      );
+
+      const owner =
+        // @todo remove this hack when User and Orgs are owned by a real Org
+        ownerUuid === nilUuid
+          ? { orgAccountId: nilUuid, shortname: "system" }
```
Alfred-Mountfield, line 25: Has it been changed from "example" to system? I worry about having it inconsistent and then passing it into a route and having it fail. In the code I sent you on slack we special case the system shortname

packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
```diff
@@ -0,0 +1,44 @@
...
+        throw new Error(
+          `Owner with accountId ${ownerUuid} not found – possibly a caching issue if it has been created mid-session`,
+        );
+      }
+
+      const isUser = "userAccountId" in owner;
+
+      return {
+        accountId: isUser ? owner.userAccountId : owner.orgAccountId,
+        shortname: owner.shortname ?? "",
```
Alfred-Mountfield, line 39: Do we want to return an empty string here? This should only happen if the user sign-up hasn't completed (to my knowledge), would we want an empty string or an error if you're trying to do something with an incomplete user?

packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
CiaranMn: Has what? There's an org called `example-org` but it doesn't own these things (it has a real id)

Alfred-Mountfield: I'm AFK atm but if you search for "example" (with quotes) you'll find it I think. It was there an hour ago

Alfred-Mountfield: (for the root account special cased shortname, which is different from the system org account, and it does not have an associated entity)

CiaranMn: Both the root account shortname and the default org account use the same environment variable - so this will be checking against `example-org`. I'll update to use this same logic (this from `useGetAccountIdForShortname`).

```
    if (
      shortname === SYSTEM_ACCOUNT_SHORTNAME ||
      shortname === "example" // This is the root account
    ) {
      return nilUuid;
    }
```

CiaranMn: Done in https://github.com/hashintel/hash/pull/1580/commits/6cb468549715f9fa10ad9bcfe8c3769125815251

packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
CiaranMn: Made it more obvious in https://github.com/hashintel/hash/pull/1580/commits/6cb468549715f9fa10ad9bcfe8c3769125815251

benwerner01 (approved):

packages/hash/frontend/src/components/hooks/useGetOwnerForEntity.ts
CiaranMn: To flag that the fallback is redundant since `SYSTEM_ACCOUNT_SHORTNAME` itself is always defined, and falls back to `example-org` if the environment variables aren't present

PR #1581, opened 2022-12-02
Property type array expected value badges in entity type editor

luisbettencourt:

This PR adds the "Array" chip to expected values to properties in the entity type editor (the chips will be expanded on in the future, this is just a placeholder https://hashintel.slack.com/archives/C022217GAHF/p1670006397135869)
It also fixes this [bug](https://github.com/hashintel/hash/pull/1579#discussion_r1038417680) by putting each chip on its own line.

## 📹 Demo

image

image


benwerner01 (approved):
Works locally for me, thanks Luis 👍

PR #1582, opened 2022-12-03
Avoid unnecessary lookups and traversals in Subgraph generation

TimDiekmann:

We currently look up each type (knowledge + ontology) twice, this reduces the amount of lookups by using the entry API. Also, don't traverse the ontology types (on the Rust side, not on PSQL) if not needed.
Additionally, this avoids iterating through the link mappings twice.

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
@@ -51,7 +51,7 @@ impl<C: AsClient> PostgresStore<C> {
             let dependency_status = dependency_context
                 .knowledge_dependency_map
                 .insert(&entity_edition_id, current_resolve_depth);
-            let entity = match dependency_status {
+            let entity: Option<&KnowledgeGraphVertex> = match dependency_status {
                 DependencyStatus::Unresolved => {
                     match subgraph.vertices.knowledge_graph.entry(entity_edition_id) {
                         Entry::Occupied(entry) => Some(entry.into_mut()),
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
@@ -1,6 +1,6 @@
 mod read;
 
-use std::{future::Future, pin::Pin};
+use std::{collections::hash_map::Entry, future::Future, pin::Pin};
 
 use async_trait::async_trait;
 use error_stack::{bail, IntoReport, Report, Result, ResultExt};
@@ -53,28 +53,19 @@ impl<C: AsClient> PostgresStore<C> {
             let dependency_status = dependency_context
                 .knowledge_dependency_map
                 .insert(&entity_edition_id, current_resolve_depth);
-            let entity: Option<&KnowledgeGraphVertex> = match dependency_status {
-                DependencyStatus::Unknown => {
-                    if let Some(entity) = subgraph.vertices.knowledge_graph.get(&entity_edition_id)
-                    {
-                        Some(entity)
-                    } else {
-                        let entity = Read::<Entity>::read_one(
-                            self,
-                            &Filter::for_entity_by_edition_id(entity_edition_id),
-                        )
-                        .await?;
-                        Some(
-                            subgraph
-                                .vertices
-                                .knowledge_graph
-                                .entry(entity_edition_id)
-                                .or_insert(KnowledgeGraphVertex::Entity(entity)),
-                        )
+            let entity = match dependency_status {
+                DependencyStatus::Unresolved => {
+                    match subgraph.vertices.knowledge_graph.entry(entity_edition_id) {
+                        Entry::Occupied(entry) => Some(entry.into_mut()),
+                        Entry::Vacant(entry) => {
+                            let entity = Read::<Entity>::read_one(
+                                self,
+                                &Filter::<Entity>::for_entity_by_edition_id(entity_edition_id),
+                            )
+                            .await?;
+                            Some(entry.insert(KnowledgeGraphVertex::Entity(entity)))
+                        }
                     }
-                }
-                DependencyStatus::DependenciesUnresolved => {
-                    subgraph.vertices.knowledge_graph.get(&entity_edition_id)
                 }
                 DependencyStatus::Resolved => None,
             };
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
@@ -51,6 +51,9 @@ impl<C: AsClient> PostgresStore<C> {
             let dependency_status = dependency_context
                 .knowledge_dependency_map
                 .insert(&entity_edition_id, current_resolve_depth);
+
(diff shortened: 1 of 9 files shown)


packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -53,29 +53,20 @@ impl<C: AsClient> PostgresStore<C> {
...
+                DependencyStatus::Unresolved => {
+                    match subgraph.vertices.knowledge_graph.entry(entity_edition_id) {
+                        Entry::Occupied(entry) => Some(entry.into_mut()),
+                        Entry::Vacant(entry) => {
+                            let entity = Read::<Entity>::read_one(
+                                self,
+                                &Filter::<Entity>::for_entity_by_edition_id(entity_edition_id),
+                            )
+                            .await?;
+                            Some(entry.insert(KnowledgeGraphVertex::Entity(entity)))
```
Alfred-Mountfield, line 64: I had a quick dig through the API but didn't find anything nice to do this, but do we actually want mutable references here? I think it could be misleading

```suggestion
                        Entry::Occupied(entry) => Some(&*entry.into_mut()),
                        Entry::Vacant(entry) => {
                            let entity = Read::<Entity>::read_one(
                                self,
                                &Filter::<Entity>::for_entity_by_edition_id(entity_edition_id),
                            )
                            .await?;
                            Some(&*entry.insert(KnowledgeGraphVertex::Entity(entity)))
```

Ideally there's a way for us to avoid the mutable reference in the first place

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/data_type.rs
```diff
@@ -32,28 +34,32 @@ impl<C: AsClient> PostgresStore<C> {
...
-                    )
+        let data_type = match dependency_status {
+            DependencyStatus::Unresolved => {
+                match subgraph
+                    .vertices
+                    .ontology
+                    .raw_entry_mut()
+                    .from_key(data_type_id)
+                {
+                    RawEntryMut::Occupied(entry) => Some(entry.into_mut()),
```
Alfred-Mountfield, line 45: Same comment here about mutable refs

Alfred-Mountfield: Wish we could do it with `from_key().or_insert_with` but the `?` makes that impossible it seems

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -47,34 +47,37 @@ impl<C: AsClient> PostgresStore<C> {
...
-                DependencyStatus::DependenciesUnresolved => {
-                    subgraph.vertices.ontology.get(entity_type_id).cloned()
-                }
                 DependencyStatus::Resolved => None,
             };
 
-            if let Some(OntologyVertex::EntityType(entity_type)) = entity_type {
-                for property_type_ref in entity_type.inner().property_type_references() {
-                    if current_resolve_depth.constrains_properties_on.outgoing > 0 {
+            if let Some(OntologyVertex::EntityType(entity_type)) = entity_type.cloned() {
```
Alfred-Mountfield, line 78: Not asking for this PR, but is there a way we can get around these clones by making these records immutable somehow? It seems that this is caused by a mutable borrow to `subgraph` so I'm wondering if these records can be guaranteed to not change so their references aren't tied to the mutable borrow on `subgraph`.

Guessing it would not be straightforward and would involve some lifetime nonsense or a secondary collection to hold onto the records while the subgraph just holds references to them or something.

Alfred-Mountfield (commented):
Generally looks good, thank you. Few clarifications please

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/data_type.rs
TimDiekmann: `or_insert_with` is mainly not working because of the `async` nature of `Read`

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: Do we maybe want a small comment saying something like:
```rust
// Explicit typing means we don't take mutable references when these records shouldn't be mutable
```

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: That's not easily possible. The subgraph is mutably borrowed in all `traverse_*` functions to be extended by the new vertices. One possible solution is to store all references (`.property_type_references()`, `.data_type_references()` etc.) before, these are only called conditionally. I will do an attempt to remove the `cloned()`. (Note, we had the clone before, but further up)

Alfred-Mountfield: I noticed the move of the clone 😄 

I don't think it's worth sinking much effort into trying to get rid of this clone, and getting the references where they are now is probably faster, I'm not too concerned, just thought it was worth highlighting.

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Good point, I added a comment in a5d8d95d344ea9cb0c9648cbd8f9880db8dc071e

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: I avoided the cloning by conditionally collecting the references before traversing further in cbb81d730b6baa601e9ffd349b7268f55bcf6720. This also allowed me to remove the `Clone` implementation on all records: d8de9653ae6bbcaa1e6c54f34b2622919019d518

Alfred-Mountfield: Nice! Can you add a comment saying why we're collecting them there rather than inline?

Something like "Collecting these here rather than where they're used avoids multiple mutable reference errors" or something

Alfred-Mountfield: > I avoided the cloning by conditionally collecting the references before traversing further in https://github.com/hashintel/hash/commit/cbb81d730b6baa601e9ffd349b7268f55bcf6720. This also allowed me to remove the Clone implementation on all records: https://github.com/hashintel/hash/commit/d8de9653ae6bbcaa1e6c54f34b2622919019d518

Have you done a quick sanity check that we weren't using it in the benchmarks code or anything?

TimDiekmann: I didn't expect that, but the integration tests used an (unneeded) clone: 25a87b200cc95d0b36b43ecaf5a77e464c9cfae8.

655b726b51b6158fc47b33a0d42fca489ec6cbff

Alfred-Mountfield (approved):
Just tried it again locally, looks good 🚀

PR #1583, opened 2022-12-03
Bump axum to 0.6.1

TimDiekmann:

Axum 0.6 requires the JSON parameter to be the last in the function, so this PR adjusts this.


packages/graph/hash_graph/lib/graph/Cargo.toml
```diff
@@ -7,7 +7,7 @@ description = "HASH Graph API"
 
 [dependencies]
 async-trait = "0.1.59"
-axum = "0.5.16"
+axum = "0.6.1"
```
thehabbos007, line 10: (comment not for line)
Want to point out that there is a different URL routing behavior in 0.6 (PR1118 in the Axum repo, not linking to not cross-reference their PR). I don't think this affects us, but we have to keep in mind that this may be the cause of any routing issues if we do run into them

packages/graph/hash_graph/lib/graph/src/api/rest/data_type.rs
```diff
@@ -272,8 +272,8 @@ struct UpdateDataTypeRequest {
     request_body = UpdateDataTypeRequest,
 )]
 async fn update_data_type<P: StorePool + Send>(
-    body: Json<UpdateDataTypeRequest>,
     pool: Extension<Arc<P>>,
+    body: Json<UpdateDataTypeRequest>,
```
thehabbos007, line 276: Question: Was this a recommended change, or a compile error with 0.6?  :)

thehabbos007 (approved):
Thank you, Tim! I tested this locally going through the UI and it seems that the routes aren't messed up or anything. I added a couple comments with may be worth keeping in mind :)

packages/graph/hash_graph/lib/graph/src/api/rest/data_type.rs
TimDiekmann: It's a compile error since 0.6 to not pass the body as the last parameter to avoid having multiple bodies in the signature.

PR #1591, opened 2022-12-05
Rename link order fields

Alfred-Mountfield:

Links have been designed with the hope of supporting bi-directional links in the future. As such, a link might need an order (list index) for both directions. We had called this `left_order` and `right_order` to avoid accidental implication of order with `source` and `target`. The choice of the order names was deemed somewhat confusing, and `left_to_right_order` and `right_to_left_order` seems to be more explicit, albeit slightly wordy.

This PR renames all occurrences (I believe). It involves changing the DB so the volume will need to be wiped in between runs.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203401244553571/f) _(internal)_

## 🔍 What does this change?

See commits

- At the moment these fields aren't captured in standalone docs, however doc comments and such should have been successfully updated as part of this.

## 🛡 What tests cover this?

Any tests (E2E, integration, unit) that interact with link entities.

## ❓ How to test this?

1.  Test out the app as usual, noting features that depend on links.

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
@@ -553,7 +553,7 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
                     .attach_printable("cannot update link order of an entity that is not a link")
             ),
             (Some(link_data), left, right) => {
-                let new_left_order = match (link_data.left_order(), left) {
+                let new_left_order = match (link_data.left_to_right_order(), left) {
                     (None, None) => None,
                     (Some(_), None) => bail!(Report::new(UpdateError).attach_printable(
                         "left order was set on entity but new order was not provided"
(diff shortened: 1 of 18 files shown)


packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -553,7 +553,7 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
                     .attach_printable("cannot update link order of an entity that is not a link")
             ),
             (Some(link_data), left, right) => {
-                let new_left_order = match (link_data.left_order(), left) {
+                let new_left_order = match (link_data.left_to_right_order(), left) {
```
TimDiekmann, line 556: You missed this line 🙂

TimDiekmann (commented):
Looks good to me, only found one variable, which was not changed

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: Cheers, that made me think of a few more things to find and replace

benwerner01 (approved):
Thanks Alfie, looks good to me 👍

PR #1592, opened 2022-12-05
Use the unified type system package

Alfred-Mountfield:

Previously the Type System package was split into two variants (for node consumption and web consumption). This caused a number of issues when wanting to utilise the functionality inside packages like `hash-shared`.

The [Type System was unified upstream](https://github.com/blockprotocol/blockprotocol/pull/808), so this PR migrates to use the new version.

## 🔍 What does this change?

- See commits, aside from the first commit (pretty much a find and replace), this is a fairly small PR.

Not to my knowledge

## ⚠️ Known issues

- The `/slim` variant is currently causing TSC errors. This will be fixed upstream and then we should be able to use the slimmer variant across the application which should benefit from greater caching and smaller size. [Asana task](https://app.asana.com/0/0/1203492058388309/f) _(internal)_ 

## 🐾 Next steps

- Fix and use the `/slim` variant
- Migrate the HASH frontend to utilise `dynamic` on the root page to avoid needing to sprinkle `useInitTypeSystem` everywhere, while still benefiting from SSR.

## 🛡 What tests cover this?

- Integration tests

## ❓ How to test this?

1.  Run and try the app as usual, paying special attention to things like the type editor which uses the functionalities more heavily


packages/hash/api/src/index.ts
```diff
@@ -50,6 +51,9 @@ import { ensureSystemEntitiesExists } from "./graph/system-entities";
 const shutdown = new GracefulShutdown(logger, "SIGINT", "SIGTERM");
 
 const main = async () => {
+  await TypeSystemInitializer.initialize();
+  logger.info("Type System initialized");
```
kachkaev, line 55: Do we want to merge this line?

kachkaev (approved):
Calling `await TypeSystemInitializer.initialize();` everywhere looks a bit quirky but I hope that we'll eventually get rid of this hack via ES module imports.

packages/hash/api/src/index.ts
Alfred-Mountfield: How would you want to do that?

kachkaev: I was under the impression that this was just a `console.log` for local debugging. Happy with this line on `main` if it’s there on purpose.

Alfred-Mountfield: Gotcha, it's there on purpose :) There are other logger statements below, it's helpful to know how far the API got before it crashed during development

PR #1593, opened 2022-12-05
Generate property table rows using entity type schema

yusufkinatas:

This PR changes property table's row generation logic. Now it generates row using entity type's schema instead of entity's properties object. Because properties object of an entity could be an empty object. Which causes property table to render no rows at all.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203433085114587/f) _(internal)_

## 🔍 What does this change?

- Update the logic for generating `X empty, Y total` string. Now it uses the table rows instead of looking at property objects.

## ❓ How to test this?

1.  Open entity editor `/new` page to create a new entity
2. New entity should show all of the properties instead of an empty property table


packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts
```diff
@@ -35,14 +35,21 @@ import { getExpectedTypesOfPropertyType } from "./get-expected-types-of-property
...
+export const generatePropertyRowRecursively = ({
+  propertyTypeBaseUri,
+  propertyKeyChain,
+  entity,
+  entitySubgraph,
+  requiredPropertyTypes,
   depth = 0,
-): PropertyRow => {
+}: {
+  propertyTypeBaseUri: string;
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts:46]: Can we try and use the `BaseUri` type (it's just an alias for string so it is the same, but it's more descriptive)

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts
```diff
@@ -35,14 +35,21 @@ import { getExpectedTypesOfPropertyType } from "./get-expected-types-of-property
...
+  propertyKeyChain: string[];
+  entity: Entity;
+  entitySubgraph: Subgraph<SubgraphRootTypes["entity"]>;
+  requiredPropertyTypes: string[];
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts:50]: Similar comment actually for all of these, can we try and use more descriptive types, I'm assuming one of `BaseUri` or `VersionedUri`

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts
```diff
@@ -158,15 +84,7 @@ export const useCreateNewEntityAndRedirect = () => {
...
-          /**
-           * @todo after implementing this ticket: https://app.asana.com/0/1203312852763953/1203433085114587/f (internal)
-           * we should just use `properties: {}` here, and delete `generateDefaultProperties` function,
-           * this is a temporary workaround for entity table to show the rows with empty values
-           */
-          properties: generateDefaultProperties(
-            entityType.properties,
-            subgraph,
-          ),
+          properties: {},
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts:87]: Why are we creating an entity with empty properties? This is concerning to me. What if the entity type has required properties, will this code break the second we add validation?

Alfred-Mountfield (commented):
A few comments, will leave an actual review to someone else :)

No need to block merge on me resolving any comments

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts
Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts:87]: I think we can talk about "default properties" instead of "required properties" here. When we have "default values for properties", we can initialize the `properties` object using those. But I guess that's not priority right now. 

Entities can be an invalid state right? (e.g. missing required properties, invalid properties). Assuming we'll create the entity in draft state.

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts
Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts]: Thanks, updated here https://github.com/hashintel/hash/pull/1593/commits/00f0ca00e7e6aed5be438497ddabbef730b0e97d

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts
Comment (Alfred-Mountfield) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts:87]: Right, but default values wouldn't mean you have a valid entity. If I have a Person entity with a required "name", "age", and "location" field, and I have a default value for "location", I still wouldn't have a valid Person entity.

> Entities can be an invalid state right?

No. I'm sorry if we've been somewhat unclear on that, I know we said explicitly a while ago that we could do that to unblock development, because the backend does not have validation yet, but that's purely for short-term development speed. We should be building with the assumption that things actually are being validated.

Comment (Alfred-Mountfield) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts:87]: There is no concept of "draft" state

nathggns:
Reviewing now

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts
```diff
@@ -63,26 +71,27 @@ export const generatePropertyRowRecursively = (
...
 
-  let value;
+  const value =
+    get(entity.properties, propertyKeyChain) ?? (isArray ? [] : undefined);
+
   const children: PropertyRow[] = [];
 
-  if (isUnknownObject(properties)) {
-    value = properties[propertyTypeBaseUri];
+  const firstOneOf = propertyType.oneOf[0];
```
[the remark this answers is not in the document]
Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/use-row-data/generate-property-rows-from-entity/generate-property-row-recursively.ts:79]: because we assume if there is a nested property object, it holds one `oneOf` and that's a ref to the nested property (so we generate the rows, in future we can extend it to be dynamic but that's how it's initially coded, and it's out-of-scope from this PR. This behavior is not introduced in this PR)

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/flatten.ts
```diff
@@ -37,3 +37,22 @@ export const flattenExpandedItemsOfTree = <
 
   return flattened;
 };
+
+/**
+ * Does the same as `flattenExpandedItemsOfTree`, but assumes all items are expanded
+ */
+export const flattenAllItemsOfTree = <
```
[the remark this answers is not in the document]
Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/flatten.ts]: Yeah, good improvement! Applied here https://github.com/hashintel/hash/pull/1593/commits/0845a4061b84fa8bf7356579414e29e509780d4a

nathggns (approved):

PR #1594, opened 2022-12-05
Remove shortname from new entity type / new entity routes and reintroduce entity type list

nathggns:

https://github.com/hashintel/hash/pull/1555 made it so the new entity type page no longer created it in the shortname present in the URL (account slug), but in the currently active workspace. It did not make a similar change to the new entity page (I think this was a PR sequencing thing), and it also didn't remove the shortname from the routes for either.

This PR updates both, while also updating the ontology chips across the board to more accurately represent the routes, and updates the links across the app to be correct. 

I've also restored the entity type list in the sidebar which is not part of the task, as I had to update the URL for the entity type list items not being rendered which required passing through the right data, which was most of the work for reintroducing the menu. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203494569321463/f) _(internal)_
- [Second Asana task](https://app.asana.com/0/1203179076056209/1203502275950661/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- `/@shortname/new/types/entity-type` has been changed to `/new/types/entity-type` and links updated
- `/@shortname/new/entity` has been changed to `/new/entity` and links updated
- `EntityPageWrapper` now takes a react node as a `header` prop instead of passing props to `EntityPageHeader` itself. This reduces the number of layers props need passing through.
- `EntityPageHeader` now takes a `chip` component (`OntologyChip`), instead of rendering it itself, as the `path` of the chip can't be calculated from the props going to the header
- The breadcrumbs in `EntityPageHeader` can now use the active workspace to generate the link for the entities button, instead of only the account slug (as there is no account slug on the new entity page now)
- `createNewEntityAndRedirect` now takes the active workspace in order to work out the destination of the new entity
- `createNewEntityAndRedirect` no longer takes the authenticated user, as it no longer needs to check that the user has the right to create a new entity in the desired location, as this is assumed by the fact its the active workspace
- The new entity page no longer needs to check that the user has the right to create a new entity in the desired location, as thats indicated by the active workspace not by URL, so that can be assumed
- `url` renamed to `pagePath` for `AccountPageList` as its more accurate, and makes it clearer when used when generating a URL for pages that its only the path (and needs an origin prepended)
- New `pagePath` prop for `PageMenu` component for generating an accurate link for copy link option for pages. This is necessary because previously it uses the route account slug as that was always present, but that's now not present in some cases where the sidebar is. 
- New `getOwnedById` util for identifying the owned by id of a workspace. This will hopefully go away once the relevant todo in the code is addressed
- `useFrozenValue` has been moved into a shared library (this is useful for not changing a view while its being animated out). There might be a better way of handling this (i.e, a component which freezes its children) in the long run.
- The account entity type list now fetches and filters the entity type list and renders it, with updated (and cleaned up) menu code, and fixed links

N/A

## ⚠️ Known issues

N/A

## 🐾 Next steps

- [Ontology chips on "new" pages should update as user types singular name](https://app.asana.com/0/1203179076056209/1203493644105451/f) (_internal_)

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Ensure you can still create new entity types and they end up in the expected place, and that you can still create new entities (both from the new entity button on the entity type page, and from the new entity page)

Additionally ensure ontology chips are as expected.

Finally ensure all create buttons across sidebar/header link as expected.


packages/hash/frontend/src/pages/new/entity.page.tsx
```diff
@@ -21,19 +19,21 @@ const Page: NextPageWithLayout = () => {
...
   if (shouldBeCreatingEntity && !creatingEntity) {
     setCreatingEntity(true);
-  } else if (entityIdInvalid && creatingEntity) {
+  } else if (entityTypeInvalid && creatingEntity) {
     setCreatingEntity(false);
   }
 
+  // @todo better way to force login when using active workspace?
+  useLoggedInUser();
+
```
yusufkinatas, line 36: I think this is fine.

yusufkinatas (changes requested):
Tested on my local and reviewed, good work! Just an eslint issue

packages/hash/frontend/src/pages/new/entity.page.tsx
nathggns: This is going to be replaced with server side redirects.

yusufkinatas (approved):
Good work!

PR #1596, opened 2022-12-06
Create system types in system org, remove nil UUID

benwerner01:

This PR ensures that the HASH App system types are created in the system org. This let's the frontend treat the system types more consistently to other types, and removes the hardcoding of `nilUuid`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203427737626951/f) _(internal)_

## 🔍 What does this change?

See commits.

Not that I'm aware of.

## ⚠️ Known issues

None.

## 🛡 What tests cover this?

The existing integration tests ensure the system types can be initialised correctly.

## ❓ How to test this?

1. Run the Hash App and login to the frontend
2. Try viewing the `User` system type in the entity type editor (at `http://localhost:3000/@example-org/types/entity-type/user`)


packages/hash/api/src/graph/system-org.ts
```diff
@@ -0,0 +1,114 @@
...
+        all: [
+          { equal: [{ path: ["version"] }, { parameter: "latest" }] },
+          {
+            equal: [
+              { path: ["type", "versionedUri"] },
+              { parameter: types.entityType.org.entityTypeId },
+            ],
+          },
+        ],
+      },
```
benwerner01, line 38: We can't use the model classes here, because they rely on the system types being initialised (which happens after this function is called).

packages/hash/api/src/index.ts
```diff
@@ -132,13 +137,20 @@ const main = async () => {
...
+  await ensureSystemOrgAccountIdExists({ graphApi, logger });
+
   await ensureSystemTypesExist({ graphApi, logger });
 
+  await ensureSystemOrgExists({
+    graphApi,
+    logger,
+  });
+
   await ensureSystemEntitiesExists({ graphApi, logger });
```
Alfred-Mountfield, line 149: Do we want to just group these into a single method at this point? (We can still have 4 exported methods if we think they'd be helpful in tests for example to have fine-grained control)

packages/hash/api/src/graph/system-org.ts
```diff
@@ -0,0 +1,123 @@
...
+  extractEntityUuidFromEntityId,
+  Subgraph,
+  SubgraphRootTypes,
+} from "@hashintel/hash-subgraph";
+import { extractBaseUri } from "@blockprotocol/type-system";
+import { OrgModel, OrgSize } from "../model";
+import { PageDefinition, seedPages } from "../seed-data/seed-pages";
+
+// eslint-disable-next-line import/no-mutable-exports
+export let systemOrgAccountId: string;
```
Alfred-Mountfield, line 19: I know we already have some in the form of the `types` variable, but these global mutable variables do concern me :/ I think this is definitely something we should look into soon, perhaps next week (almost certainly out of scope for this PR), but generally these are an anti-pattern.

packages/hash/api/src/graph/system-org.ts
```diff
@@ -0,0 +1,123 @@
...
+      { title: "First" },
+      { title: "Second" },
+      { title: "Third" },
+    ];
+
+    await seedPages(pageTitles, systemOrgModel.getEntityUuid(), params);
+
+    logger.info(
+      `System Org with shortname = "${systemOrgModel.getShortname()}" now has seeded pages.`,
+    );
```
thehabbos007, line 121: Perhaps we shouldn't be seeding pages anymore for the system org. This was done originally as the org created was considered a "seeded org", but I don't think we should add pages to this for seeding reasons (at least not without having it be a flag you can toggle)

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
```diff
@@ -72,11 +72,11 @@ export const pages: ResolverFn<
 > = async (_, { ownedById }, { dataSources: { graphApi }, userModel }) => {
   const accountModel = ownedById
     ? await UserModel.getUserById(graphApi, {
-        entityId: userAccountIdToEntityId(ownedById),
+        entityId: `${systemOrgAccountId}%${ownedById}`,
```
thehabbos007, line 75: We should be using this instead: https://github.com/hashintel/hash/blob/88e84beca68a835394f270350bb6d9e59e7aa62b/packages/hash/subgraph/src/types/identifier.ts#L14

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
```diff
@@ -72,11 +72,11 @@ export const pages: ResolverFn<
...
       }).catch((error: Error) => {
         if (error instanceof EntityTypeMismatchError) {
           return OrgModel.getOrgById(graphApi, {
-            entityId: userAccountIdToEntityId(ownedById),
+            entityId: `${systemOrgAccountId}%${ownedById}`,
```
thehabbos007, line 79: As above, we should be using this instead: https://github.com/hashintel/hash/blob/88e84beca68a835394f270350bb6d9e59e7aa62b/packages/hash/subgraph/src/types/identifier.ts#L14

packages/hash/api/src/graph/system-org.ts
Alfred-Mountfield: Possibly not worth doing next week though. @CiaranMn is starting to look into TS type generation from schemas, which I've thought for a while could be used for the system types through a codegen step. It wouldn't help solve the `AccountId` thing, but ideally could solve the other bits to do with what should basically be static schemas which we're currently programmatically generating on every import of that file of the shared package 🙃

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
thehabbos007: Or perhaps even define a custom function that uses the helper under the hood in the `system-org.ts` file to create an `entityId` under that account space without needing to pass `systemOrgAccountId`

packages/hash/api/src/index.ts
```diff
@@ -132,13 +137,20 @@ const main = async () => {
...
 
   // This will seed users, an org and pages.
   // Configurable through environment variables.
-  await seedOrgsAndUsers({ graphApi, logger });
+  await seedUsers({ graphApi, logger, orgModel: systemOrgModel });
```
thehabbos007, line 153: I understand why `systemOrgModel` is exported as a mutable export, but it does feel strange/coupled that the `ensureSystemOrgAccountIdExists` and `ensureSystemOrgExists` functions don't return the accountId/orgModel to be used here instead of importing

packages/hash/api/src/graph/system-org.ts
```diff
@@ -0,0 +1,123 @@
...
+      `System Org available with shortname = "${systemOrgModel.getShortname()}"`,
+    );
+
+    const pageTitles: PageDefinition[] = [
+      { title: "First" },
+      { title: "Second" },
+      { title: "Third" },
+    ];
+
+    await seedPages(pageTitles, systemOrgModel.getEntityUuid(), params);
```
Alfred-Mountfield, line 117: Should we be doing this here? I'd like to see a clear separation of seeding the required thing for the app to run (the org), and random test data (seeded pages)

packages/hash/api/src/graph/system-types.ts
```diff
@@ -425,49 +400,41 @@ const dummyEntityTypeInitializer = async (graphApi: GraphApi) => {
...
   ...types.linkEntityType.contains,
-  actorId: systemAccountId,
 });
 
-const parentLinkEntityTypeInitializer = entityTypeInitializer({
-  ...types.linkEntityType.parent,
-  actorId: systemAccountId,
-});
+const parentLinkEntityTypeInitializer = entityTypeInitializer(
+  types.linkEntityType.parent,
```
Alfred-Mountfield, line 436: Can you remove the spread operator on the other objects as well? For example
```ts
const authorLinkEntityTypeInitializer = entityTypeInitializer({
  ...types.linkEntityType.author,
});
```

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
Alfred-Mountfield, line 75: Let's use `entityIdFromOwnedByIdAndEntityUuid` for these please :)

packages/hash/api/src/index.ts
```diff
@@ -132,13 +137,20 @@ const main = async () => {
...
   await ensureSystemTypesExist({ graphApi, logger });
 
+  await ensureSystemOrgExists({
+    graphApi,
+    logger,
+  });
+
   await ensureSystemEntitiesExists({ graphApi, logger });
 
   // This will seed users, an org and pages.
```
Alfred-Mountfield, line 151: This wasn't updated

packages/hash/api/src/model/knowledge/org.model.ts
```diff
@@ -82,7 +83,7 @@ export default class extends EntityModel {
     const entityTypeModel = SYSTEM_TYPES.entityType.org;
 
     const entity = await EntityModel.create(graphApi, {
-      ownedById: systemAccountId,
+      ownedById: orgAccountId,
```
Alfred-Mountfield, line 86: This is a subtle but big change? 

I thought we'd agreed that we wanted all `User`s and `Org`s to live under the root account. I'm not against us changing this but I think it's worth us having a discussion on Slack

packages/hash/api/src/model/knowledge/org.model.ts
```diff
@@ -64,7 +64,8 @@ export default class extends EntityModel {
       );
     }
 
-    const { data: orgAccountId } = await graphApi.createAccountId();
+    const orgAccountId =
+      params.orgAccountId ?? (await graphApi.createAccountId()).data;
```
thehabbos007, line 68: Oh, interesting, do we just return the accountId as a raw string from `createAccountId`?

packages/hash/api/src/model/knowledge/orgMembership.model.ts
```diff
@@ -84,7 +84,7 @@ export default class extends LinkEntityModel {
     };
 
     const entity = await user.createOutgoingLink(graphApi, {
-      ownedById: systemAccountId,
+      ownedById: systemOrgAccountId,
```
Alfred-Mountfield, line 87: This feels like a bug in the previous code. Fairly certain we want this to be owned by the org's accountId, not the _system_ org. Might not be the same here

packages/hash/api/src/model/knowledge/user.model.ts
```diff
@@ -200,7 +200,7 @@ export default class extends EntityModel {
     const entityTypeModel = SYSTEM_TYPES.entityType.user;
 
     const entity = await EntityModel.create(graphApi, {
-      ownedById: systemAccountId,
+      ownedById: systemOrgAccountId,
```
Alfred-Mountfield, line 203: See comment above about org entity ownership

packages/hash/api/src/model/ontology/util.ts
```diff
@@ -13,14 +13,14 @@ export const getNamespaceOfAccountOwner = async (
   params: { ownerId: string },
 ) => {
   const namespace =
-    params.ownerId === systemAccountId
+    params.ownerId === systemOrgAccountId
       ? SYSTEM_ACCOUNT_SHORTNAME
       : (
           (await UserModel.getUserById(graphApi, {
-            entityId: `${systemAccountId}%${params.ownerId}`,
+            entityId: `${systemOrgAccountId}%${params.ownerId}`,
```
Alfred-Mountfield, line 20: Can we use `entityIdFromOwnedByIdAndEntityUuid` for these please

packages/hash/api/src/model/ontology/util.ts
```diff
@@ -13,14 +13,14 @@ export const getNamespaceOfAccountOwner = async (
   params: { ownerId: string },
 ) => {
   const namespace =
-    params.ownerId === systemAccountId
+    params.ownerId === systemOrgAccountId
       ? SYSTEM_ACCOUNT_SHORTNAME
```
Alfred-Mountfield, line 17: We don't need to be doing this now right? I mean it will be _slightly_ faster but I think it'd be more consistent to use the actual Org entity here.

packages/hash/api/src/model/ontology/util.ts
thehabbos007, line 20: We can also use this here https://github.com/hashintel/hash/blob/88e84beca68a835394f270350bb6d9e59e7aa62b/packages/hash/subgraph/src/types/identifier.ts#L14

packages/hash/api/src/seed-data/index.ts
```diff
@@ -1,82 +1,31 @@
...
-  const sharedOrgModel = await OrgModel.createOrg(graphApi, {
-    name: SYSTEM_ACCOUNT_NAME,
-    shortname: SYSTEM_ACCOUNT_SHORTNAME,
-    providedInfo: {
-      orgSize: OrgSize.ElevenToFifty,
-    },
-    actorId: systemAccountId,
+  const createdUsers = await ensureUsersAreSeeded({
+    ...params,
+    systemOrgAccountId: orgModel.getEntityUuid(),
```
Alfred-Mountfield, line 16: This is assuming you're passing in the system org. Please change this.

packages/hash/api/src/seed-data/seed-users.ts
```diff
@@ -38,9 +37,11 @@ const devUsers: readonly SeededUser[] = [
 export const ensureUsersAreSeeded = async ({
   graphApi,
   logger,
+  systemOrgAccountId,
 }: {
   graphApi: GraphApi;
   logger: Logger;
+  systemOrgAccountId: string;
```
Alfred-Mountfield, line 44: I would have liked to see `type AccountId = string` added as a drive-by in this PR considering the amount of code we're adding that's using it.

Probably not worth it now but in the future we should consider doing such things if they're backwards-compatible (which a plain string is)

packages/hash/integration/src/tests/model/knowledge/block.model.test.ts
```diff
@@ -43,8 +41,7 @@ describe("Block model class", () => {
 
   beforeAll(async () => {
     await TypeSystemInitializer.initialize();
-    await ensureSystemTypesExist({ graphApi, logger });
-    await ensureSystemEntitiesExists({ graphApi, logger });
+    await ensureHashAppIsInitialized({ graphApi, logger });
```
Alfred-Mountfield, line 44: Previously in the review I asked for us to unify the initialize calls, can we just reuse this? :)

Alfred-Mountfield (changes requested):
Partial review, will continue after I grab lunch. Cheers! Really excited for how much more consistent this makes things

packages/hash/frontend/src/components/hooks/useGetAccountIdForShortname.ts
```diff
@@ -12,18 +10,6 @@ export const useGetAccountIdForShortname = (
...
 
   const accountId = useMemo(() => {
-    /**
-     * @todo - This is incredibly flakey right now.
-     *    We should be creating system types and entities under the actual org, _not_ the nilUuid
-     *    https://app.asana.com/0/1202805690238892/1203427737626951/f
-     */
-    if (
-      shortname === SYSTEM_ACCOUNT_SHORTNAME ||
-      shortname === "example" // This is the root account
```
thehabbos007, line 22: Quick question, how do we refer to the built-in link-type that is created in the graph layer? I assume that since it's not a system type but a builtin that it would have the nil-uuid as the owner, which we don't have a shortname for. Do we not make it browsable/addressable on purpose?

thehabbos007 (commented):
Thanks, Ben! I have some comments. Looks great! :)

packages/hash/shared/src/types.ts
```diff
@@ -1,17 +1,10 @@
 import { BaseUri, VersionedUri } from "@blockprotocol/type-system";
 import slugify from "slugify";
-import { EntityId } from "@hashintel/hash-subgraph";
 import { SYSTEM_ACCOUNT_SHORTNAME } from "@hashintel/hash-shared/environment";
 import { frontendUrl } from "./environment";
 
 type SchemaKind = "data-type" | "property-type" | "entity-type";
 
-export const nilUuid = "00000000-0000-0000-0000-000000000000" as const;
```
Alfred-Mountfield, line 9: 🎉

Alfred-Mountfield (commented):
Alright, turns out previous comments were comprehensive, when we sort those out I'm probably good to approve after a re-review :)

packages/hash/frontend/src/components/hooks/useGetAccountIdForShortname.ts
Alfred-Mountfield: > Do we not make it browsable/addressable on purpose?

Correct, somewhat in the same way that there's no way to discover the primitive data types in the frontend (outside of selecting them for values, I mean you can't actually _view_ them). This will all have to change when we implement a data type editor, and inheritance but for now it holds true.

Also the primitive types will need special casing anyway as their URLs will be blockprotocol ones, which means we won't have normal shortname-based routing anyway I think (so the nil-uuid should end up irrelevant and not exposed)

packages/hash/api/src/index.ts
benwerner01: That makes sense, I've already introduced such a methods in the tests so have moved and made use of it in the backend in https://github.com/hashintel/hash/pull/1596/commits/2f497441d396eeb58da8e3e6740225ad1fd5fbf6

packages/hash/api/src/graph/index.ts
```diff
@@ -34,3 +40,16 @@ export const createGraphClient = (
 
   return new GraphApiClient(config, basePath, axiosInstance);
 };
+
+export const ensureHashAppIsInitialized = async (params: {
```
Alfred-Mountfield, line 44: Glad we have a shared method for this now.

I think we should work on the naming, perhaps something like `ensureSystemGraphIsInitialized` or `ensureSystemPrerequisitesAreInGraph` or something equally unpleasant but at least clear.

This does not initialize the HASH app, it just makes sure the requisite information is within the graph so that the app can continue with initializing

packages/hash/integration/src/tests/model/knowledge/block.model.test.ts
Alfred-Mountfield: You've done that now, thanks 🎉

packages/hash/api/src/model/ontology/util.ts
benwerner01: I agree in theory we don't need to do this, but when I tried removing it I discovered that `getNamespaceOfAccountOwner` is used before the system types are created so an error is thrown without this custom check for `systemOrgAccountId`.

By further restructuring things this can probably be avoided, but I'm not sure it's worth the hassle as this should be pretty robust. If you disagree I can spend more time looking into it.

Alfred-Mountfield: Ah I see, this seems to be another issue to do with the model classes. There are two options here I guess:
* Leave it as is and track it in a task as an issue to tackle next week alongside the model classes
* ~~Just remove the dependency on the model classes here similar to what you did for [this](https://github.com/hashintel/hash/pull/1596#discussion_r1040839366)~~

Actually it turns out the model classes are used in the type initializers, definitely not worth trying to restructure as a drive-by now.

Can you put a todo in here and associate it with the model classes, saying we need to do this because the initialization logic uses the model classes, which in turn uses this.

packages/hash/api/src/graph/system-org.ts
benwerner01: I started out the implementation by having the `systemOrgAccountId` be returned by the `ensureSystemOrgAccountIdExists` method, and pass it explicitly as a parameter. Although I would prefer this approach, it exploded the scope of this PR because `systemAccountId` is being used in some quite hard to reach places, and the function arguments would have to be updated in a lot of places.

I am in favour of us more explicitly passing things around, and not relying on these uninitialised exports. It'd help ensure that things happen in the correct order.

packages/hash/api/src/graph/system-org.ts
benwerner01: This was addressed as part of introducing the system user:
- https://github.com/hashintel/hash/pull/1596/commits/2101b462063712aa6e9ab795f2db22ec59236864
- https://github.com/hashintel/hash/pull/1596/commits/7634f6395ff669efabfaed4fc34c28a02aceea86

packages/hash/api/src/index.ts
benwerner01: I'm not entirely sure what you're referring to, the comment? In which case this was removed

packages/hash/api/src/model/knowledge/org.model.ts
benwerner01: Thanks for catching

packages/hash/api/src/model/knowledge/orgMembership.model.ts
benwerner01: Agreed, have changed it in https://github.com/hashintel/hash/pull/1596/commits/ba5b6ccf460bee98b8b821be7244b007726609bf

packages/hash/api/src/model/knowledge/user.model.ts
benwerner01: This is now consistent with what we're doing in `org.model.ts` after the change in https://github.com/hashintel/hash/commit/5a6fc01c844c537a3a4c3ba30c9f6d903aa0fbc0

packages/hash/api/src/model/ontology/util.ts
benwerner01: Yes, addressed in https://github.com/hashintel/hash/commit/6e6c358e80915bc8fb3fb132d66f3382fbfaf395

packages/hash/api/src/seed-data/index.ts
benwerner01: This no longer makes such an assumption after the changes in:
- https://github.com/hashintel/hash/pull/1596/commits/2101b462063712aa6e9ab795f2db22ec59236864
- https://github.com/hashintel/hash/pull/1596/commits/7634f6395ff669efabfaed4fc34c28a02aceea86

packages/hash/integration/src/tests/model/knowledge/block.model.test.ts
benwerner01: Yeah, that's what I've done in https://github.com/hashintel/hash/commit/2f497441d396eeb58da8e3e6740225ad1fd5fbf6 :)

packages/hash/api/src/graph/system-org.ts
benwerner01: Agreed, this was addressed in:
- https://github.com/hashintel/hash/pull/1596/commits/2101b462063712aa6e9ab795f2db22ec59236864
- https://github.com/hashintel/hash/pull/1596/commits/7634f6395ff669efabfaed4fc34c28a02aceea86

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
benwerner01: This was addressed in https://github.com/hashintel/hash/pull/1596/commits/6e6c358e80915bc8fb3fb132d66f3382fbfaf395

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
benwerner01: https://github.com/hashintel/hash/pull/1596/commits/6e6c358e80915bc8fb3fb132d66f3382fbfaf395

packages/hash/api/src/index.ts
benwerner01: Agreed, see my comment to Alfie on why I stuck with this approach in https://github.com/hashintel/hash/pull/1596#discussion_r1041001326

packages/hash/api/src/model/knowledge/org.model.ts
benwerner01: Yes

packages/hash/api/src/model/ontology/util.ts
benwerner01: Addressed in https://github.com/hashintel/hash/pull/1596/commits/6e6c358e80915bc8fb3fb132d66f3382fbfaf395

packages/hash/frontend/src/components/hooks/useGetAccountIdForShortname.ts
benwerner01: This is a good point which I hadn't considered. I don't think we are currently exposing this in the browser, but it's a good point that there is still technically an entity type which has the `nilUuid`. Not sure whether we care about this.

cc @Alfred-Mountfield

packages/hash/api/src/graph/index.ts
benwerner01: I also prefer `ensureSystemGraphIsInitialized`, have renamed this in https://github.com/hashintel/hash/pull/1596/commits/2ee0ca0cfc695f172ec8d7969068556617debc0c

packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts
```diff
@@ -47,7 +46,11 @@ export const useCreateNewEntityAndRedirect = () => {
       if (!entityType) {
         throw new Error("persisted entity type not found");
       }
-      const ownedById = getOwnedById(activeWorkspace);
+
+      const ownedById =
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[account-slug]/entities/[entity-uuid].page/shared/use-create-new-entity-and-redirect.ts:50]: Just to flag that this can be nicer, but I'll update it in my PR #1600

packages/hash/frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list.tsx
```diff
@@ -163,7 +162,10 @@ export const AccountEntityTypeList: FunctionComponent<
 
   const accountEntityTypes = useMemo(() => {
     if (allEntityTypes) {
-      const ownedById = getOwnedById(activeWorkspace);
+      const ownedById =
```
Alfred-Mountfield, line 165: Can you actually change this function to take `activeWorkspaceAccountId`? It's available and passed in to a sibling component in the parent

Alfred-Mountfield (changes requested):
One final change and happy to approve 🎉

thehabbos007:
I'm happy with the changes, I'll let Alfie do the final approval if mine is not required :)

packages/hash/frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list.tsx
benwerner01: Addressed in https://github.com/hashintel/hash/pull/1596/commits/32a664811f4059a8d9129a453e6161edb8cdc987

benwerner01: Addressed in https://github.com/hashintel/hash/pull/1596/commits/32a664811f4059a8d9129a453e6161edb8cdc987

Alfred-Mountfield (approved):

PR #1598, opened 2022-12-06
Clarify entity UUID

Alfred-Mountfield:

When we were making the definition of `EntityId` consistent across the app, we missed a small parameter in the entity model creator params. This fixes that.


benwerner01 (approved):
Thanks Alfie! Looks good :)

PR #1604, opened 2022-12-06
Support creation of entities

TimDiekmann:

Calls the `create_entity` method of the database from the Graph API.

These changes were made without touching the OpenAPI specs, this will be covered in [this Asana task](https://app.asana.com/0/0/1203444301722133/f) _(internal)_.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203444301722132/f) _(internal)_

## 🚫 Blocked by

- #1599 

## 🔍 What does this change?

- Remove old creation/updating/archiving code, replace the two latter cases with TODOs
- Add TODOs in appropriate locations to plan further work
- Call the `create_entity` SQL method from Graph API
- Implement `ToSql` and `FromSql` for some common types to avoid calling the same function over and over again

## 🐾 Next steps

- [Support updating of entities](https://app.asana.com/0/0/1203464975244303/f) _(internal)_

## 🛡 What tests cover this?

The Rest endpoint tests will create entities. However, everything after the creation will fail as it's not been implemented yet.


packages/graph/hash_graph/lib/graph/src/shared/identifier/knowledge.rs
```diff
@@ -83,22 +82,21 @@ impl ToSchema for EntityId {
...
 
-    #[must_use]
-    pub const fn inner(&self) -> Timestamp {
-        self.0
+impl Serialize for EntityVersion {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // TODO: Expose temporal versions to backend
```
Alfred-Mountfield, line 97: This is just to have a non-breaking change I assume?

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -15,6 +15,12 @@ pub mod ontology;
 
 pub type Timestamp = DateTime<Utc>;
 
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
```
Alfred-Mountfield, line 18: Why put `ToSchema` on this for now? Appreciate it may not hurt to do so but I think we should figure out the OpenAPI spec for this stuff in one go when consciously doing it. The `EntityVersion` one is overridden so I presume we don't need to do this?

packages/graph/hash_graph/lib/graph/src/shared/provenance.rs
```diff
@@ -9,8 +10,20 @@ use crate::identifier::account::AccountId;
...
+            Clone,
+            PartialEq,
+            Eq,
+            Hash,
+            PartialOrd,
+            Ord,
+            Serialize,
+            Deserialize,
+            FromSql,
+            ToSql,
```
Alfred-Mountfield, line 24: Similar question to why we've added this to `EntityUuid`

packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -327,11 +326,12 @@ pub trait EntityStore: crud::Read<Entity> {
     /// - if an [`EntityUuid`] was supplied and already exists in the store
     async fn create_entity(
         &mut self,
-        properties: EntityProperties,
+        entity_id: EntityId,
```
Alfred-Mountfield, line 329: Why have you changed this to be required? 🤔

packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -327,11 +326,12 @@ pub trait EntityStore: crud::Read<Entity> {
...
+        decision_time: Option<Timestamp>,
```
Alfred-Mountfield, line 330: Do we want to _consider_ making new types for DecisionTime and TransactionTime?

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -122,19 +121,23 @@ impl<C: AsClient> PostgresStore<C> {
                             )
                             .await?
                             .into_iter()
-                            .map(|entity| entity.metadata().edition_id())
+                            .map(|entity| {
```
Alfred-Mountfield, line 124: Can you update the variable name for all of these

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -372,45 +385,82 @@ impl<C: AsClient> PostgresStore<C> {
...
-                entity_type_id,
-                updated_by_id,
-                link_data,
+        let row = self
+            .as_client()
+            .query_one(
+                r#"
+                SELECT
+                    entity_edition_id,
+                    lower(decision_time),
```
Alfred-Mountfield, line 411: Are we storing them as strings in the database?

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -372,45 +385,82 @@ impl<C: AsClient> PostgresStore<C> {
...
+                    NULLIF(upper(decision_time), 'infinity'),
+                    lower(system_time),
+                    NULLIF(upper(system_time), 'infinity')
+                FROM
+                    create_entity($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13);
```
Alfred-Mountfield, line 416: An unfortunate side-effect of shoving this in a stored-proc is it makes it _**very**_ hard for us to sanity check the ordering of parameters. Can you think of anything we can do here to make it better? @thehabbos007 any thoughts?

Alfred-Mountfield (changes requested):
Thanks Tim, generally looks good. Would like some discussion on some points though

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
thehabbos007: Postgres supports named parameter calling notation for functions, but not sure if this is restricted to the built in ones only. It would look something like:
```sql
create_entity(_owned_by_id => $1)
```
https://www.postgresql.org/docs/current/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-NAMED

Alfred-Mountfield: If there aren't meaningful restrictions with doing so, I'd be more comfortable with that. Shoving 13 unnamed things in a function is not fun

packages/graph/hash_graph/lib/graph/src/shared/identifier/knowledge.rs
TimDiekmann: Yes, exactly

packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -215,29 +220,13 @@ async fn archive_entity<P: StorePool + Send>(
...
-                return StatusCode::NOT_FOUND;
-            }
-
-            tracing::error!(error=?report, "Could not archive entity");
-
-            // Insertion/update errors are considered internal server errors.
-            StatusCode::INTERNAL_SERVER_ERROR
-        })
+    tracing::warn!(
+        "Use of deprecated endpoint `/entities/archive`, use `/entities/update` instead"
```
thehabbos007, line 224: There should be a OpenAPI flag for deprecation. No idea how to set it in utoipa though - and not worth doing if this function is to be removed entirely within the dev-branch

packages/graph/hash_graph/lib/graph/src/knowledge/entity/mod.rs
```diff
@@ -15,8 +15,21 @@ use crate::{
...
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
+    ToSchema,
+    FromSql,
+    ToSql,
```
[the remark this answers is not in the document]
thehabbos007: I would imagine the `#[postgres(transparent)]` added below would serve the same purpose as both would just call the `Uuid` `To/FromSql`. I don't think we could differentiate between the SQL newtypes when taking SQL parameters, could we?

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
thehabbos007: if you're asking about the `lower` it's the way to get the lower bound, not the lowercase version of the string

packages/graph/hash_graph/lib/graph/src/knowledge/entity/mod.rs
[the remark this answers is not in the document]
TimDiekmann: As discussed offline, I removed the `FromSql` implementation: 335193c558819a6b09279c87a1834e6774588c5b

packages/graph/hash_graph/lib/graph/src/shared/provenance.rs
TimDiekmann: As discussed offline, I removed the `FromSql` implementation: 335193c558819a6b09279c87a1834e6774588c5b

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
TimDiekmann: Good point, I removed it for now until we adjust the backend: 335193c558819a6b09279c87a1834e6774588c5b

packages/graph/hash_graph/lib/graph/src/store/mod.rs
Alfred-Mountfield: I was thinking we _might_ also want a `DecisionTimeSpan` 😅

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Updated the names in 2507807f458d5ee53a31c04ab84d5918e93a5096 and 36c72e68e8c133d66e8e14ddb61f12bf72c38124. The names may be a little bit too long but I expect to remove these soon.

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: Hmm, at least when exposing these this might be useful. I will consider that, thanks!

packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: I don't know if this function is going to be replaced in the `dev` branch or shortly after. It's going to be replaced in the linked task, which I hopefully get to this week.

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: Long names are better than unclear or incorrect ones 😁 Thanks!

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: https://www.w3resource.com/PostgreSQL/lower-function.php

I was misled 😭

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Changed in 9064133f29e9b3f8349d1d9ab80a9d327827d2bc

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: This was to align with the Postgres implementation. I expect to only return newly generated parameters from this function in the future and as the entity id can be known beforehand it feels more intuitive to pass it directly. Also, this encapsulates better, that the `owned_by_id` and `entity_uuid` belongs together. Starting with temporal versioning, we also store them next to each other in the database.

Alfred-Mountfield: I don't like this change being a drive-by in this PR if I'm honest. Can we revert it, and open a separate discussion in Slack to discuss this. It's quite a foundational change to say that the calling applications need to be responsible for generating EntityId's. It's something we've discussed before, but not agreed on, or decided the ramifications of.

packages/graph/hash_graph/lib/graph/src/store/mod.rs
thehabbos007: I would prefer them being newtypes if it isn't problematic with these SQL types, but not a strong preference. Would suck to pass the wrong timespan because of the type

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: If this is a custom function, do we want to pick something that doesn't clash with the in-built? And if it is the in-built... then what the heck

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
Alfred-Mountfield: Damn that does add a lot of lines. I think it's a lot better though, thank you!

packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
Alfred-Mountfield: I think it will have to be replaced as part of the dev branch if we do not reimplement it.

Alfred-Mountfield (commented):
I'll approve once we've resolved the open threads

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: The built-in functions are called `lower` and `upper`.

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: Reverted in 7f7fa9e0992b21605cfc296da69c41af8c76c839

Alfred-Mountfield (approved):

PR #1606, opened 2022-12-06
Temporal versioning: Support updating of entities

TimDiekmann:

Similar to #1604, this PR uses the `update_entity` method in SQL to update an entity.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203464975244303/f) _(internal)_

## 🚫 Blocked by

- #1604 

## 🔍 What does this change?

- Use the SQL function `update_entity` to ... update an entity 👁️ 👁️ 
- Remove unused error structs
- Fallback to get-entity + update entity for archiving. 

## ⚠️ Known issues

The fallback in the `/entities/archive` endpoint will result in data races if the same entity is updated at the same time, but I don't consider this as an issue as I'm going to remove this function as part of [this Asana task](https://app.asana.com/0/0/1203444301722133/f) _(internal)_ and the chance, that this will happen is very very unlikely. If it would happen, there would still be an error.

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
+++ b/packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
@@ -12,7 +12,6 @@ use futures::TryFutureExt;
 use serde::{Deserialize, Serialize};
 use type_system::uri::VersionedUri;
 use utoipa::{OpenApi, ToSchema};
-use uuid::Uuid;
 
 use crate::{
     api::rest::{
@@ -33,7 +32,7 @@ use crate::{
     },
     provenance::{OwnedById, ProvenanceMetadata, UpdatedById},
     store::{
-        error::{EntityDoesNotExist, QueryError},
+        error::{EntityDoesNotExist, RaceConditionOnUpdate},
         query::Filter,
         EntityStore, StorePool,
     },
@@ -120,7 +119,11 @@ impl RoutedResource for EntityResource {
                         .get(get_latest_entities::<P>)
                         .put(update_entity::<P>),
                 )
-                .route("/archive", post(archive_entity::<P>))
+                .route(
+                    "/archive",
+                    #[expect(deprecated)]
+                    post(archive_entity::<P>),
+                )
                 .route("/query", post(get_entities_by_query::<P>))
                 .route("/:entity_uuid", get(get_entity::<P>)),
         )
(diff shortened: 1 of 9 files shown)


packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -340,5 +387,42 @@ async fn update_entity<P: StorePool + Send>(
...
+            order,
+        )
+        .await
+        .map_err(|report| {
+            tracing::error!(error=?report, "Could not update entity");
+
+            if report.contains::<EntityDoesNotExist>() {
+                StatusCode::NOT_FOUND
+            } else if report.contains::<RaceConditionOnUpdate>() {
+                StatusCode::LOCKED
```
thehabbos007, line 421: It's fair enough we're using this return code, but I believe it's for WebDAV calendars specifically :D

packages/graph/hash_graph/lib/graph/src/store/error.rs
```diff
@@ -74,6 +74,18 @@ impl fmt::Display for EntityDoesNotExist {
 
 impl Context for EntityDoesNotExist {}
 
+#[derive(Debug)]
+#[must_use]
+pub struct RaceConditionOnUpdate;
+
+impl fmt::Display for RaceConditionOnUpdate {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.write_str("A race condition happened when updating the entity")
```
thehabbos007, line 83: Could this text error be more domain specific? Something about the entity being updated unexpectedly being changed?

packages/graph/migrations/postgres/migration/1656417312397_initial.ts
```diff
@@ -643,6 +630,8 @@ export const up = (pgm: MigrationBuilder): void => {
     },
     `
     BEGIN
+      SET CONSTRAINTS entity_versions_overlapping DEFERRED;
```
thehabbos007, line 633: Makes sense we'd have to do this because the update happens after wards :)

thehabbos007 (approved):
Thanks, Tim! I have some comments, but this is as we've discussed in length about. You've already addressed the changes we discussed, so it's just a couple nits. I'm happy to merge as-is, though :)

packages/graph/hash_graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: Happy to change it to another code. Just from the name it sounds reasonable to use it 😄

thehabbos007: We wouldn't have any clients that would interpret this as WebDAV i believe, but you should add it to the utoipa return code list in the proc macro :)

TimDiekmann: Good point: 2c10386786917a2d635cd04829f0db2a196996a3

thehabbos007 (approved):

PR #1607, opened 2022-12-06
Edit Bar on Entity Type Editor shouldn't scroll entire page when animating

nathggns:

When the edit bar on the entity type editor animates in/out, it scrolls the page to keep the content underneath it static. When this functionality was introduced, there was no sidebar on the type editor. Since this was introduced, it's been discovered the method it uses to scroll the page results in a gap appearing underneath the sidebar. 

This PR fixes that by scrolling the right element + makes this functionality work in Firefox, which it never did. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203462405114281/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- New `EditBarScroller` component which wraps the page with a DOM node it then provides to its children using context, as well as providing access to the "scrolling element". This is the node which will be scrolled. 
- This component is now used by the `LayoutWithSidebar` component
- Updated `useFreezeScrollWhileTransitioning` to read the nodes from this context instead of using `document.documentElement`. 

N/A

## ⚠️ Known issues

N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

1.  Make a change to a property in the entity type editor, and be sure the page doesn't shift as the edit bar animates in 
2. Make sure no gap appears beneath the sidebar while this happens
3. Ensure this also works in firefox as expected

### DIFF
--- a/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx
+++ b/packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx
@@ -1,10 +1,11 @@
 import { faSmile } from "@fortawesome/free-regular-svg-icons";
 import { FontAwesomeIcon } from "@hashintel/hash-design-system";
 import { Box, Collapse, Container, Stack, Typography } from "@mui/material";
-import { ReactNode, useEffect, useRef, useState } from "react";
+import { ReactNode, useContext, useEffect, useRef, useState } from "react";
 import { useFormState } from "react-hook-form";
 import { PencilSimpleLine } from "../../../../shared/icons/svg";
 import { Button, ButtonProps } from "../../../../shared/ui/button";
+import { EditBarContext } from "./edit-bar-scroller";
 import { EntityTypeEditorForm } from "./form-types";
 
 const useFrozenValue = <T extends any>(value: T): T => {
@@ -96,12 +97,13 @@ const EditBarContents = ({
 const useFreezeScrollWhileTransitioning = () => {
   const observerRef = useRef<ResizeObserver | null>(null);
   const ref = useRef<HTMLDivElement>(null);
+  const page = useContext(EditBarContext);
 
   useEffect(() => {
-    const docNode = document.documentElement;
-    const node = ref.current;
+    const scroller = page?.parentNode as HTMLElement | null | undefined;
+    const editBar = ref.current;
 
-    if (!node) {
+    if (!editBar || !page || !scroller) {
       return;
     }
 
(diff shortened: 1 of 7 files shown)


packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx
```diff
@@ -1,10 +1,11 @@
 import { faSmile } from "@fortawesome/free-regular-svg-icons";
 import { FontAwesomeIcon } from "@hashintel/hash-design-system";
 import { Box, Collapse, Container, Stack, Typography } from "@mui/material";
-import { ReactNode, useEffect, useRef, useState } from "react";
+import { ReactNode, useContext, useEffect, useRef, useState } from "react";
 import { useFormState } from "react-hook-form";
 import { PencilSimpleLine } from "../../../../shared/icons/svg";
 import { Button, ButtonProps } from "../../../../shared/ui/button";
+import { EditBarContext } from "../../../../shared/layout/layout-with-sidebar/edit-bar-scroller";
```
Comment (kachkaev) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx:8]: N.B.: should be `import { EditBarContext } from "../../../../shared/×××`"` (no need to fix this now)

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx
```diff
@@ -96,32 +97,34 @@ const EditBarContents = ({
 const useFreezeScrollWhileTransitioning = () => {
   const observerRef = useRef<ResizeObserver | null>(null);
   const ref = useRef<HTMLDivElement>(null);
+  const editBarContext = useContext(EditBarContext);
```
Comment (kachkaev) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx:100]: `useContext(FooBarContext)` is a bit too raw. I prefer seeing `useFooBar` as it makes code more readable. We can add checks for `undefined` into this custom hook and throw if the hook is called outside the context provider. Alternatively, we can return some default context too. If we do this, hook's return signature becomes simpler and the component ends up easier to read.

This is probably too minor for blocking this PR, but I'm happy to re-review it if you want to make changes.

kachkaev (approved):

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx
Comment (nathggns) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/edit-bar.tsx:100]: Completely agree - this happened as the context had been in this file, and then I refactored. Will fix.

packages/hash/frontend/src/shared/edit-bar-scroller.tsx
```diff
@@ -0,0 +1,31 @@
+import { Box } from "@mui/material";
+import { createContext, ReactNode, useContext, useMemo, useState } from "react";
+
+const EditBarContext = createContext<{
+  page: HTMLElement;
+  scrollingNode: HTMLElement;
+} | null>(null);
+
+export const useEditBarContext = () => useContext(EditBarContext);
```
kachkaev, line 9: We can potentially add _if `null`, `throw`_ here. This will allow us to do `const { page, scroller } = useEditBar()`. We only use this hook in one place though so we can probably leave the signature as is and revisit it if we see that this hook has become more popular.

kachkaev (approved):

packages/hash/frontend/src/shared/edit-bar-scroller.tsx
nathggns: I thought about that but would mean you would need to check both variables are defined in the effect. Will leave this as is for now. 

Thank you!

PR #1612, opened 2022-12-07
Serve ontology types as json on request

CiaranMn:

Ontology types – i.e. the types that make up the type system: entity types, property types, data types - are identified by URIs. A key part of the intended system is that the (mostly) JSON Schema representing these types is available at the URIs they are identified by, so that they are available for easy access / parsing outside of the application they are hosted in.

This PR makes them available as JSON when their URI is requested with `accept: "application/json"` in the header (visiting the URI normally will take you to their editing UI)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203507375651261/f) _(internal)_

## 🔍 What does this change?

- adds middleware to intercept requests with an `accept: "application/json"` header on type-related routes, and sends the JSON instead
- Enables sending a response directly from middleware body (otherwise you can only rewrite/redirect to another page)
- removes "must be logged in" gating from the `getEntityType` etc routes – we can consider types public until we introduce a public/private type distinction

## ⚠️ Known issues

- The data types currently all live at blockprotocol.org, which isn't yet serving them

## 🐾 Next steps

- Write a utility that, given the URI of an entity type, fetches the entity type and its dependents, and constructs TS types from it

## ❓ How to test this?

1.  Send some get requests to type URIs with `accept: "application/json"` in the header, e.g.
- http://localhost:3000/@example-org/types/entity-type/org/v/1
- http://localhost:3000/@example-org/types/property-type/shortname/v/1

### DIFF
--- a/packages/hash/frontend/src/middleware/return-types-as-json.ts
+++ b/packages/hash/frontend/src/middleware/return-types-as-json.ts
@@ -63,7 +63,7 @@ export const returnTypeAsJson = async (request: NextRequest) => {
   if (validationResultType === "Err") {
     return generateErrorResponse(
       400,
-      `${validationResult.reason}: requests for JSON representations of types must be made to a versioned URI, e.g. https://hash.ai/@example-org/types/entity-types/user/v/1`,
+      `${validationResult.reason}: requests for JSON representations of types must be made to a versioned URI, e.g. https://hash.ai/@example-org/types/entity-type/user/v/1`,
     );
   }
 


packages/hash/frontend/src/middleware/return-types-as-json.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  if (validationResultType === "Err") {
+    return generateErrorResponse(
+      400,
+      `Requests for JSON representations of types must be made to a versioned URI, e.g. ${url.replace(
+        /\/$/,
+        "",
+      )}/v/1`,
+    );
+  }
```
thehabbos007, line 71: Just flagging that there are a lot of error variants for the versioned URI:
https://github.com/blockprotocol/blockprotocol/blob/main/crates/type-system/src/ontology/uri/error.rs#L21-L36
do we care about making the response more precise, or is close enough good enough?

packages/hash/frontend/src/middleware/return-types-as-json/generate-query-args.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (!string) {
+    throw new Error("Node did not contain 'loc'");
+  }
+
+  return string;
+};
+
+const zeroDepth = { outgoing: 0 };
+
+type OntologyType = "data" | "entity" | "property";
```
thehabbos007, line 27: I have a feeling we may have this sort of enum somewhere else, maybe in another form, but may be in the subgraph lib?

packages/hash/frontend/src/middleware/return-types-as-json.ts
```diff
@@ -0,0 +1,139 @@
...
+    | GetPropertyTypeQueryVariables
+  >(query, variables, cookie);
+
+  if (errors || !data) {
+    const { code, message } = errors?.[0] ?? {
+      code: "INTERNAL_SERVER_ERROR",
+      message: "Unknown error",
+    };
+    return generateErrorResponse(
+      code === "FORBIDDEN" ? 401 : code === "INTERNAL_SERVER_ERROR" ? 500 : 400,
```
thehabbos007, line 101: In what case would be getting a `FORBIDDEN` response?

thehabbos007 (commented):
Thanks Ciaran! I have some comments, but looks good :)

packages/hash/frontend/src/middleware/return-types-as-json.ts
nonparibus: If we don't implement these now we should capture them in a follow-up.

packages/hash/frontend/src/middleware/return-types-as-json/generate-query-args.ts
nonparibus: @TimDiekmann Any idea?

CiaranMn: I can add it if we find it, although string unions are a bit easier to deal with than enums so would prefer the former if we're writing things ourselves

packages/hash/frontend/src/middleware/return-types-as-json.ts
CiaranMn: `FORBIDDEN` is the code given if a `ForbiddenError` from ApolloServer is thrown, which we do in a few places in the API, for both lack of authentication or authorization.

When we add private types, if we keep throwing the same error this will pick it up. When we add more codes (e.g. to cover authorization), we'll pick this up in the refactor.

packages/hash/frontend/src/middleware/return-types-as-json/generate-query-args.ts
TimDiekmann: We at least have `OntologyVertex` in the generated `api.ts`, which has a kind of unstable name for `OntologyVertexOneOf.kind`, so I guess we cannot use that.
We also have `OntologyVertex` defined in the subgraph library:
```ts
export type OntologyVertex =
  | DataTypeVertex
  | PropertyTypeVertex
  | EntityTypeVertex;

export type DataTypeVertex = { 
  kind: "dataType"; 
  inner: DataTypeWithMetadata 
};

export type PropertyTypeVertex = {
  kind: "propertyType";
  inner: PropertyTypeWithMetadata;
};

export type EntityTypeVertex = {
  kind: "entityType";
  inner: EntityTypeWithMetadata;
};
```

I'm not experienced enough with type definitions in TS, but maybe we can use the latter?

At least I'd prefer to be consistent with the strings, so change them to contain `type` as well (if possible)
```suggestion
type OntologyType = "dataType" | "entityType" | "propertyType";
```

CiaranMn: We cannot use `dataType` etc as these are extracted from the URL, e.g. `https://example.com/types/[something]-type`. `SchemaKind` in that linked file would be closest but it is not exported. 

I'm going to merge this for now but if there's some other suitable type please let me know and I'll update it. This will only be an issue if we ever add or change the name of an existing type though.

CiaranMn:
> Looks good to me and works locally (tested with and without cookies). Thank you! Up to you if you want another set of eyes on this or not

Thank you – I'm going to merge as it helps me in work I plan to do today. Happy to make further changes in follow up if people make suggestions post-merge.

packages/hash/frontend/src/middleware/return-types-as-json/generate-query-args.ts
Alfred-Mountfield: I don't think we have a `kebab-case` enum anywhere, so this seems fine to me. I agree with Tim though, I really dislike us having `"data" | "entity" | "property"` in code when talking about types, it's reallyyy confusing.

> We cannot use `dataType` etc as these are extracted from the URL, e.g. `https://example.com/types/[something]-type`. 

Can we not extract it like `https://example.com/types/[something]/` where `something` is `entity-type` for example?

CiaranMn: Sure, I can do this as a driveby. I think the potential for confusion is pretty limited in this case as it's localized to one file and called `OntologyType` but happy to clean it up later.

Alfred-Mountfield: Drive-by at some point would be good I think. I appreciate the name of the type is clear, but the issue is that type can be used in bigger objects, and then you inevitably end up seeing other objects like

```
updateStateArgs = {
  type: "entity",
  ...
}
```

Because `type` was `OntologyType`

packages/hash/frontend/src/middleware/return-types-as-json.ts
CiaranMn: I'd expect the error to be `InvalidBaseUri` in that case – you've explained that http://localhost:3000/@example-org/types/entity-type/user is invalid because it's missing a trailing slash

I don't think we need to sink time into this now though, "it's wrong" is fine

PR #1615, opened 2022-12-07
Array property type creation UX improvements

luisbettencourt:

This PR improved the UX around array property type creation.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203490593648614/f) _(internal)_

## 🔍 What does this change?

- Users should now be able to add multiple arrays to the same property (previously if there was already an array it would be removed);
- Added the `Array of Arrays` expected value chip for arrays with only arrays inside it.
- closing the deletion confirmation modal when the `Confirm Deletion` button is clicked (it wouldn't close when more than one array was present in the property type).

## 📹 Demo

image

image


packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/property-type-form/data-type-selector/custom-data-type-menu/array-data-type-menu.tsx
```diff
@@ -278,9 +278,14 @@ export const ArrayDataTypeMenu: FunctionComponent<ArrayDataTypeMenuProps> = ({
...
             if (typeId) {
+              const allowMultiple =
+                expectedValuesOptions[typeId]?.allowMultiple;
               const defaultData = getDefaultData(typeId);
 
-              if (reason === "selectOption") {
+              if (
+                reason === "selectOption" ||
+                (reason === "removeOption" && allowMultiple)
+              ) {
```
Comment (yusufkinatas) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/property-type-form/data-type-selector/custom-data-type-menu/array-data-type-menu.tsx:288]: I'm not sure if I understand what's happening here. If an option is selected, or if an option is removed, but it's "allowMultiple", then we add an item to the list?

yusufkinatas (commented):
Tested the functionality, works as expected. Asked about a logic I didn't understand.

yusufkinatas:
Here we can see an unexpected behavior when we try to delete the first array type. Do we have a ticket related with this on Asana?
https://user-images.githubusercontent.com/39553853/207005152-1274a642-3679-4f1d-b299-afbfe3fbe913.mov

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/property-type-form/data-type-selector/custom-data-type-menu/array-data-type-menu.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/property-type-form/data-type-selector/custom-data-type-menu/array-data-type-menu.tsx:288]: Autocomplete toggles items by default, which is something we want for the basic data types (text, number, boolean). So if we select text and text was already selected, it is removed. But for objects and arrays we want to be able to add as many as we want, because they can be different from each other.

luisbettencourt:
Thank you for catching this. Created this [task](https://app.asana.com/0/0/1203532455452915/f)

yusufkinatas (approved):
Looks good to me!

PR #1623, opened 2022-12-08
Rename `Grid`, `PropertyTable`, `LinkTable` related variables/ files

yusufkinatas:

To prevent confusion in future while maintaining grid related code, this PR renames some variabes / files around `Grid` component.

## 🔍 What does this change?

Renames following variables in the codebase around `Grid` component, `PropertyTable`, `LinkTable`
- `rowData` -> `rows`
- `sortRowData` -> `sortRows`
- `tableSort` -> `columnSort` (since component is called grid, not table)
- `[col, row]` -> `[colIndex, rowIndex]`
- `propertyRow` | `linkRow` -> `row` (in places where it’s obvious we’re working with which table, e.g. )
- `property: PropertyRow` -> `propertyRow: PropertyRow` (at data of some custom cell props)


packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/entities-tab.tsx
```diff
@@ -62,10 +62,10 @@ export const EntitiesTab: FunctionComponent = () => {
   }, [entities, activeWorkspaceAccountId]);
 
   const createGetCellContent = useCallback(
-    (rowData: TypeEntitiesRow[]) =>
+    (_rows: TypeEntitiesRow[]) =>
```
Comment (luisbettencourt) [inline packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/entities-tab.tsx:65]: not a big fan of `_rows`  as we will lose the unused variable warning if we ever change this piece of code, but I understand `rows` is declared above.

luisbettencourt (approved):

packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/entities-tab.tsx
Comment (yusufkinatas) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/entities-tab.tsx:65]: Discussion moved to a [slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1670525761916299) _(internal)_

Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[account-slug]/types/entity-type/[entity-type-id].page/entities-tab.tsx:65]: Discussion here: https://hashintel.slack.com/archives/C02TWBTT3ED/p1670525761916299

PR #1624, opened 2022-12-08
Temporal versioning: Use newtypes for timestamps and timespans

TimDiekmann:

As requested in https://github.com/hashintel/hash/pull/1604#discussion_r1042221384, this adds newtype structs for `TransactionTimestamp`, `DecisionTimestamp`, `TransactionTimespan`, and `DecisionTimespan`. Also, it adds a proper `FromSql` implementation for the timespans, which simplifies the logic on creating, updating, and (next PR) selecting.

Adding these newtypes already fixed a bug on the predefined filters, which were using the decision time, even though that is not exposed yet. (In fact, that does not make a difference, but it's wrong)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203505325130333/f) _(internal)_

## 🚫 Blocked by

- #1618 

## ⚠️ Known issues

`postgres_types` does not expose ranges, but the `postgres_protocol` crate, which is the underlying crate of `postgres_types` exposes these.


packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
```diff
@@ -472,11 +493,20 @@ pub struct ParameterConversionError {
...
+            Parameter::Boolean(boolean) => boolean.to_string(),
+            Parameter::Number(number) => number.to_string(),
+            Parameter::Text(text) => text.to_string(),
+            Parameter::Uuid(uuid) => uuid.to_string(),
+            Parameter::SignedInteger(integer) => integer.to_string(),
+            Parameter::Timestamp(timestamp) => match timestamp {
+                Bound::Included(timestamp) => format!("including bound of {timestamp}"),
+                Bound::Excluded(timestamp) => format!("excluding bound of {timestamp}"),
+                Bound::Unbounded => "unbounded timestamp".to_owned(),
+            },
```
TimDiekmann, line 506: Just to flag: This will be changed back in #1626

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -13,14 +23,177 @@ pub mod account;
...
+                    Bound::Excluded(timestamp)
+                }
+                RangeBound::Unbounded => Bound::Unbounded,
+            })
+        }
+
+        fn is_negative_infinity(bound: &RangeBound<Option<&[u8]>>) -> io::Result<bool> {
+            Ok(match bound {
+                RangeBound::Inclusive(Some(mut bytes)) | RangeBound::Exclusive(Some(mut bytes)) => {
+                    bytes.read_i64::<BigEndian>()? == i64::MIN
```
thehabbos007, line 98: Are we sure about the endianness, is it defined in the Postgres spec/code anywhere?

packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
```diff
@@ -504,14 +534,14 @@ impl Parameter<'_> {
             }
             (Parameter::Text(text), ParameterType::Timestamp) => {
                 if text != "latest" {
-                    *self = Parameter::Timestamp(
-                        Timestamp::from_str(&*text)
+                    *self = Parameter::Timestamp(Bound::Included(
+                        TransactionTimestamp::from_str(&*text)
```
TimDiekmann, line 538: Just to flag: This will be changed back in #1626

packages/graph/hash_graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -304,7 +304,12 @@ impl<'c, 'p: 'c, T: PostgresQueryRecord + 'static> SelectCompiler<'c, 'p, T> {
                     Parameter::Boolean(bool) => self.artifacts.parameters.push(bool),
                     Parameter::Uuid(uuid) => self.artifacts.parameters.push(uuid),
                     Parameter::SignedInteger(integer) => self.artifacts.parameters.push(integer),
-                    Parameter::Timestamp(timestamp) => self.artifacts.parameters.push(timestamp),
+                    Parameter::Timestamp(timestamp) => match timestamp {
+                        Bound::Unbounded => self.artifacts.parameters.push(&"infinity"),
+                        Bound::Included(timestamp) | Bound::Excluded(timestamp) => {
+                            self.artifacts.parameters.push(timestamp);
+                        }
+                    },
```
TimDiekmann, line 312: Just to flag: This will be changed back in #1626

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -13,14 +23,177 @@ pub mod account;
...
+    #[must_use]
+    pub(crate) fn new(timespan: Timespan) -> Self {
+        Self {
+            start_bound: timespan.start.map(DecisionTimestamp),
+            end_bound: timespan.end.map(DecisionTimestamp),
+        }
+    }
+}
+
+impl RangeBounds<DecisionTimestamp> for DecisionTimespan {
```
thehabbos007, line 151: How convenient that the `Bound` would be the implementation detail used for BTree ranges which happen to be the kind of index we're using for the spans (not that it would make any difference) :D

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -13,14 +23,177 @@ pub mod account;
...
+        self.start_bound.as_ref()
+    }
+
+    fn end_bound(&self) -> Bound<&TransactionTimestamp> {
+        self.end_bound.as_ref()
+    }
+}
+
+impl TransactionTimespan {}
+
```
thehabbos007, line 196: I suppose this could be removed
```suggestion
```

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -13,14 +23,177 @@ pub mod account;
...
+                RangeBound::Inclusive(None) | RangeBound::Exclusive(None) => {
+                    unimplemented!("null ranges are not supported")
+                }
+                RangeBound::Inclusive(Some(bytes)) => {
+                    let timestamp = DateTime::from_sql(&Type::TIMESTAMPTZ, bytes)?;
+                    Bound::Included(timestamp)
+                }
+                RangeBound::Exclusive(Some(bytes)) => {
+                    let timestamp = DateTime::from_sql(&Type::TIMESTAMPTZ, bytes)?;
+                    Bound::Excluded(timestamp)
```
thehabbos007, line 89: Quick question (maybe more related to the SQL `tstzrange`). I can see that the bounds are `'[)'` by default, but it's not clear from our usage that we're operating under that assumption. Should we be making that explicit in the SQL functions using `tstzrange`? And additionally, how could we encode that invariant in Rust (should we do that?)

I.e. `tstzrange(now(), 'infinity')` turns into `tstzrange(now(), 'infinity', '[)')` for explicitness

thehabbos007 (commented):
Looks good, Tim! I have some questions I think we should talk about before merging

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
TimDiekmann: the endianness is used in the `postgres-protocol` crate. However, I was able to find a function, which does that logic for us, so I was able to remove the `byteorder` crate.

PR #1632, opened 2022-12-09
Temporal versioning: Fix `owned_by_id` not included in GiST

TimDiekmann:

The `owned_by_id` is required for identifying an entity, so it *must* be included in the GiST.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203505325130332/f) _(internal)_


thehabbos007 (approved):
Thanks, Tim. As far as I understand this, we would now have quicker queries that filter on `owned_by_id` or `entity_uuid` as well for the table, which is neat! (Depending on what the query planner does, of course)

PR #1635, opened 2022-12-09
Upgrade kratos to v0.11.0

thehabbos007:

This PR bumps kratos to version [v0.11.0](https://github.com/ory/kratos/releases/tag/v0.11.0) in anticipation for simplifying email verification flows. 
The change `Add webhooks that can block the request, useful to do some additional validation.` will allow us to rely on rollback in the kratos service rather than undoing requests ourselves.

This also turned out to clean up the frontend well in places, using their new HTTP client names and getting rid of the `alpha` package we were using for the frontend self-service :)

For email sending, there are new flags to properly do retries and make sure it's robust.
Other things to consider:
- It's stated that email magic links for recovery may be problematic as some antivirus software may open links to check for malicious content, which means "recovery codes" may make sure the recovery can happen for users 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203509372806154/f) _(internal)_

## 🔍 What does this change?

- Bumps kratos to v0.11.0

None

## ⚠️ Known issues

None, but please do test to see if all flows work as expected

## 🐾 Next steps

- [Enable email flows](https://app.asana.com/0/1203363157432081/1203509372806154/f) and implement the flows in UI

## ❓ How to test this?

Spin up the app as you would do normally by checking out, `yarn external-services` down/up and `yarn dev`


packages/hash/api/src/auth/index.ts
```diff
@@ -99,7 +105,7 @@ const setupAuth = (params: {
   // eslint-disable-next-line @typescript-eslint/no-misused-promises -- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/50871
   app.use(async (req, _res, next) => {
     const kratosSession = await publicKratosSdk
```
indietyp, line 107: How about using also including the `xSessionToken` property? it would enable API clients (using the ory API self-service login) to use HASH, notably the graphql endpoint.

> Suggestion taken from https://www.ory.sh/docs/identities/sign-in/check-session-token-cookie-api

```suggestion
    const authHeader = req.headers.authorization;
    const hasAuthHeader = authHeader.startsWith("Bearer ");
    const sessionToken = hasAuthHeader ? authHeader.slice(7, authHeader.length) : null;

    const kratosSession = await publicKratosSdk
```

packages/hash/api/src/auth/index.ts
```diff
@@ -99,7 +105,7 @@ const setupAuth = (params: {
...
-      .toSession(undefined, req.header("cookie"))
+      .toSession({ cookie: req.header("cookie") })
```
indietyp, line 108: ... continuation of the previous suggestion

```suggestion
      .toSession({
          cookie: req.cookies.join("; "),
          xSessionToken: sessionToken,
      })
```

packages/hash/api/src/auth/index.ts
```diff
@@ -71,14 +71,20 @@ const kratosAfterRegistrationHookHandler =
...
         res.status(200).end();
       } catch (error) {
-        /**
-         * @todo: instead of manually cleaning up after the registration flow,
-         * use a "pre-persist" after registration kratos hook when the following
-         * PR is merged: https://github.com/ory/kratos/pull/2343
-         */
-        await adminKratosSdk.adminDeleteIdentity(kratosIdentityId);
-
-        next(error);
```
benwerner01, line 81: 🎉

packages/hash/external-services/kratos/kratos.dev.yml
```diff
@@ -1,6 +1,6 @@
 # Based on the quick-start example https://github.com/ory/kratos/blob/master/contrib/quickstart/kratos/email-password/kratos.yml
```
benwerner01, line 1: Don't we need to set `can_interrupt: true` for the registration-flow web-hook, as specified in [this example](https://github.com/ory/docs/pull/1142/files)? Otherwise to my understanding the kratos identity is created whether the user entity was successfully created in the graph or not.

packages/hash/external-services/kratos/kratos.prod.yml
```diff
@@ -1,6 +1,6 @@
 # Based on the quick-start example https://github.com/ory/kratos/blob/master/contrib/quickstart/kratos/email-password/kratos.yml
```
benwerner01, line 1: This suggestion also applies to this configuration file.

packages/hash/frontend/src/pages/shared/ory-kratos.ts
```diff
@@ -1,37 +1,24 @@
...
-     * Directly connecting to kratos (using "http://127.0.0.1:4433") would prevent the
-     * CRSF token from being set as an HTTP-Cookie, because the browser cannot send or
-     * receive cookies via the browser `fetch` method unless:
-     *  1. `credentials: "include"` is set in the HTTP request header
-     *  2. the correct CORS origin is configured in the kratos server
-     *
-     * Therefore requests to the ory kratos public endpoint are made on the server in a
-     * Next.js API handler.
-     */
-    basePath: "/api/ory",
```
benwerner01, line 30: This is going to conflict with changes made in [this PR](https://github.com/hashintel/hash/pull/1637) to make the `oryKratosClient` work on the server-side. Do you know whether this will work on the server-side?

benwerner01 (changes requested):
These kratos changes look great :)

I think I may have spotted a potential issue with the updated usage, and left a question, but apart from this looks good to me 👍

packages/hash/external-services/kratos/kratos.dev.yml
thehabbos007: Whoopsie, must have been removed when i rebased. You're right!

thehabbos007: 4a16d43d0d05558ce3f9e1aae2b78d98ef60a3da

packages/hash/frontend/src/pages/shared/ory-kratos.ts
thehabbos007: It would not, no. The change here isn't required, so I'll undo it. It just defined the same config we had already. I'll rename the endpoint again to `/api/ory` from `/api/.ory`

thehabbos007: Reverted in bbde5d84ba5f6797e3e4c5e64d53f4c68f1ec306

packages/hash/api/src/auth/index.ts
thehabbos007: Thank you so much! I've applied this in ac3955992d57c3cb7e702def5e1b965c6cad2854 <3

packages/hash/api/src/auth/index.ts
thehabbos007: Applied with the above change :)
I used the `request` function instead for case insensitivity

indietyp (approved):

packages/hash/api/src/auth/index.ts
```diff
@@ -71,14 +67,20 @@ const kratosAfterRegistrationHookHandler =
...
+        // The kratos hook can interrup creation on 4xx and 5xx responses.
```
Alfred-Mountfield, line 70:
```suggestion
        // The kratos hook can interrupt creation on 4xx and 5xx responses.
```

packages/hash/api/src/auth/index.ts
```diff
@@ -98,8 +100,19 @@ const setupAuth = (params: {
...
+    const authHeader = req.header("authorization");
+    const hasAuthHeader = authHeader?.startsWith("Bearer ") ?? false;
+    const sessionToken =
+      hasAuthHeader && typeof authHeader === "string"
+        ? authHeader.slice(7, authHeader.length)
+        : undefined;
+
     const kratosSession = await publicKratosSdk
-      .toSession(undefined, req.header("cookie"))
+      // .toSession({ cookie: req.header("cookie") }
```
Alfred-Mountfield, line 111: Was this intended to be left commented out?

thehabbos007: It was not, thanks :)
f64533fb36bf90408a05902af7870ef375c7fe64

packages/hash/api/src/auth/ory-kratos.ts
```diff
@@ -1,19 +1,20 @@
...
 
-export const publicKratosSdk = new OpenSourceV0alpha2Api(
+export const publicKratosSdk = new FrontendApi(
   new Configuration({ basePath: publicUrl }),
 );
 
 const adminUrl = getRequiredEnv("ORY_KRATOS_ADMIN_URL");
 
-export const adminKratosSdk = new OpenSourceV0alpha2Api(
+export const adminKratosSdk = new IdentityApi(
```
benwerner01, line 17: Maybe we can rename this accordingly to something like `kratosIdentityApi` to be more consistent with the changes made to kratos?

packages/hash/api/src/auth/ory-kratos.ts
```diff
@@ -1,19 +1,20 @@
...
+  CreateIdentityBody,
+  FrontendApi,
+  IdentityApi,
 } from "@ory/kratos-client";
 import { getRequiredEnv } from "../util";
 
 const publicUrl = getRequiredEnv("ORY_KRATOS_PUBLIC_URL");
 
-export const publicKratosSdk = new OpenSourceV0alpha2Api(
+export const publicKratosSdk = new FrontendApi(
```
benwerner01, line 11: Similarly we could rename this to something like `kratosFrontendApi`

packages/hash/frontend/src/pages/signup.page.tsx
```diff
@@ -104,11 +104,14 @@ const KratosRegistrationFlowForm: FunctionComponent = () => {
       .push(`/signup?flow=${flow.id}`, undefined, { shallow: true })
       .then(() =>
         oryKratosClient
-          .submitSelfServiceRegistrationFlow(String(flow?.id), {
-            csrf_token,
-            traits,
-            password,
-            method: "password",
+          .updateRegistrationFlow({
+            flow: String(flow?.id),
```
benwerner01, line 108: Given line 104 I don't think this would be nullable:
```suggestion
            flow: flow.id,
```

benwerner01 (approved):
Thank you Ahmad, the changes are working locally for me as expected. Have left a few more thoughts which I had.

packages/hash/frontend/src/pages/signup.page.tsx
thehabbos007: Yes, thanks!

thehabbos007: Applied in 1ca939b057484c10cf54f7660758dc1ce4e0cb19

packages/hash/api/src/auth/ory-kratos.ts
thehabbos007: Thanks, applied in 1ca939b057484c10cf54f7660758dc1ce4e0cb19

benwerner01 (approved):

PR #1625, opened 2022-12-08
Remove unused dependencies

kachkaev:
## 🌟 What is the purpose of this PR?

This PR removes unused dependencies thus improving repo maintainability. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1203452982053236/f) _(internal)_
- https://github.com/blockprotocol/blockprotocol/pull/816
- https://github.com/blockprotocol/blockprotocol/pull/824

## ⚙️ Methodology

1.  List workspaces

    ```sh
    yarn workspaces info
    ```

1.  Run these commands (one per workspace location)

  ```sh
  cd apps/hashdotai; npx npm-check@latest; cd -
  cd apps/hashdotdev; npx npm-check@latest; cd -
  cd blocks/calculation; npx npm-check@latest; cd -
  cd blocks/callout; npx npm-check@latest; cd -
  cd blocks/chart; npx npm-check@latest; cd -
  cd blocks/code; npx npm-check@latest; cd -
  cd blocks/countdown; npx npm-check@latest; cd -
  cd blocks/divider; npx npm-check@latest; cd -
  cd blocks/drawing; npx npm-check@latest; cd -
  cd blocks/embed; npx npm-check@latest; cd -
  cd blocks/github-pr-overview; npx npm-check@latest; cd -
  cd blocks/header; npx npm-check@latest; cd -
  cd blocks/html-para; npx npm-check@latest; cd -
  cd blocks/image; npx npm-check@latest; cd -
  cd blocks/paragraph; npx npm-check@latest; cd -
  cd blocks/person; npx npm-check@latest; cd -
  cd blocks/shuffle; npx npm-check@latest; cd -
  cd blocks/stopwatch; npx npm-check@latest; cd -
  cd blocks/table; npx npm-check@latest; cd -
  cd blocks/timer; npx npm-check@latest; cd -
  cd blocks/toggle-item; npx npm-check@latest; cd -
  cd blocks/video; npx npm-check@latest; cd -
  cd libs/javascript/@local/eslint-config; npx npm-check@latest; cd -
  cd libs/javascript/@local/tsconfig; npx npm-check@latest; cd -
  cd packages/engine/apache-arrow-js-bundle; npx npm-check@latest; cd -
  cd packages/graph/clients; npx npm-check@latest; cd -
  cd packages/graph/clients/typescript; npx npm-check@latest; cd -
  cd packages/graph/migrations; npx npm-check@latest; cd -
[...]

PR #1639, opened 2022-12-09
Enable `unicron/filename-case` and `canonical/filename-no-index` in most workspaces

kachkaev:

This PR supplements https://github.com/hashintel/hash/pull/1625. Initial motivation:

> Move `plugins: ["@typescript-eslint", "canonical", "unicorn"],` from `packages/hash/design-system/.eslintrc.cjs` to remove a warning about missing `@typescript-eslint/eslint-plugin` in this workspace. This warning will become an error after upgrading to Yarn Berry. `eslint-plugin-canonical` needs to be moved to `@local/eslint-config` too. 

After moving `eslint-plugin-canonical` to `@local/eslint-config`, I realised that it would not be too hard to enable these two rules in the whole repo:
- `unicorn/filename-case`
- `canonical/filename-no-index`

They should help us with switching to `kebab-case` and removing `index.js` files. To avoid refactoring now, I have temporarily disabled these rules where applicable.

## 🔗 Related links

- https://github.com/hashintel/hash/pull/1625

## 🐾 Next steps

Enable both rules in all workspaces, remove `eslint-disable` directives once we’ve renamed the files.

## 🛡 What tests cover this?

- CI


packages/graph/migrations/.eslintrc.cjs
```diff
@@ -4,4 +4,12 @@ module.exports = {
   env: {
     node: true,
   },
+  overrides: [
+    {
+      files: ["**/migration/*.ts"],
```
kachkaev, line 9: These files are autogenerated so I guess we can permanently allow `_` in their names.

CiaranMn (approved):

PR #1642, opened 2022-12-09
Remove the frontend unit tests

benwerner01:

This PR removes the frontend unit tests, as their functionality will be replaced by the playwright test-suite.

Not that I'm aware of.

## ❓ How to test this?

Check CI that the existing unit tests continue passing, but the frontend tests aren't run (because they've been removed).


kachkaev (dismissed):
🧹

kachkaev (approved):

kachkaev (approved):

PR #1516, opened 2022-11-28
`deer` implement `Deserialize` for built-in types (Part 1)

indietyp:
## 🌟 What is the purpose of this PR?

This implements `Deserialize` for a select built-in types.

PR #1645, opened 2022-12-10
`deer` housekeeping

indietyp:

This is a collection of required changes to implement other features (notably #1516).

This changes the following:

* `*Access::next()` now returns `Option<Result<T, *AccessError>>`
* `*Access::finish()` is now `*Access::end()`
* `*Access::set_bounded()`
* `Reflection` is no longer a supertrait
* remove `Deserialize::deserialize_none`

Open Questions:
* `ObjectAccess::value` rename?

### `*Access::next` return type

Before, it was `Result<Option<T>, *AccessError>` which was mirrored from `serde`, but it doesn't make sense in our case. While this is just a simple transpose, an error is often not an immediate fail, while `None` always stops iteration immediately.

### `*Access::finish` rename

The name is more fitting and is more in line with existing projects (serde)

### `*Access::set_bounded`

Before the calling of `Visitor::visit_none` through `*Access` was not well defined and would only happen for `ObjectAccess::value`, this is flawed, as one might want to fill a statically sized array (or use it for structs), as an compromise between forever looping and adding new items (which would have been the case with `Visitor::visit_none` on `Vec` elements), `*Access` callers can now opt into bounded "mode", this will hard limit the amount of times `*Access::next` can be called before a hard cut-off happens, and only `None` will be returned. This enables `VIsitor::visit_none` calls on arrays and other statically sized containers but will never result in a forever loop on dynamically sized containers like `Vec`.

`*Access::end` must still be called to ensure that not too many items have been supplied.

### `Reflection` no longer a supertrait

`Reflection` is no longer a super trait and instead an associated type, this allows us to implement `Deserialize` on types that are not `'static`. `Document` (and therefore `Reflection`) currently rely on `TypeId`s to collect all `Schema`s and reference them. This means that `Reflection` needs to be `'static`, forcing `Deserialize` to be `'static`. This means that `Deserialize` could not be implemented for `&str`. Now, `&str` can specify `str` (as `Reflection` no longer requires `Sized`) as the reflection type, and the corresponding reflection will be used instead.

To ease the use, a new trait function, `Deserialize::reflection`, has been added.

> Note: in theory, the reflection for non-static types should be the same as their `'static` counterpart, but there is no way in rust to remove references like (https://en.cppreference.com/w/cpp/types/remove_reference). Refer to https://stackoverflow.com/questions/51026536/implement-something-similar-to-cs-stdremove-reference-in-rust for more information.

### `Deserializer::deserialize_none` removed

This function has been removed, as `Visitor::visit_none` should only be called on objects that do not exist in objects and arrays. I do not see a use-case or possible state where one would like to call `deserialize_none` explicitly. Planned types like `Maybe<T>` would instead rely on `Visitor`. This change might be reverted if anyone sees a possible use case.

### `ObjectAccess::value` rename?

`ObjectAccess::value` is used to directly access the value of a specific key, used for types that explicitly require string names like structs. The name should probably be less ambiguous and must not be provided, so a better name should probably be chosen.


packages/libs/deer/src/lib.rs
```diff
@@ -44,23 +44,61 @@ mod schema;
...
+    /// [`Self::value`] also counts toward the length, behaviour of multiple calls to
+    /// [`Self::value`] will always decrement the counter.
+    ///
+    /// This is best suited for types where the length/amount of keys is already predetermined, like
+    /// structs or enum variants.
+    ///
+    /// # Errors
+    ///
+    /// This should error if a call to [`Self::next`] or [`Self::value`] has been made before
+    /// calling this function or this function has been called repeatably.
```
TimDiekmann, line 62: I don't know if there are any conventions on documentation on traits, but I'd prefer avoiding "should". This is a hint to the implementor, but the majority of users call that function instead of implementing it.

packages/libs/deer/src/lib.rs
```diff
@@ -44,23 +44,61 @@ mod schema;
 extern crate alloc;
 
 pub trait ObjectAccess<'de> {
+    /// This enables bound-checking for [`ObjectAccess`]. After calling this [`ObjectAccess`] will
```
TimDiekmann, line 47: rustdoc uses the first paragraph as summary on the parent page. This needs reformatting.
```suggestion
    /// This enables bound-checking for [`ObjectAccess`].
    ///
    /// After calling this [`ObjectAccess`] will
```

packages/libs/deer/src/lib.rs
```diff
@@ -44,23 +44,61 @@ mod schema;
...
+
+    fn size_hint(&self) -> Option<usize>;
 
-    fn finish(self) -> Result<(), ObjectAccessError>;
+    fn end(self) -> Result<(), ObjectAccessError>;
 }
 
 pub trait ArrayAccess<'de> {
-    fn next<T>(&mut self) -> Result<Option<T>, ArrayAccessError>
+    /// This enables bound-checking for [`ArrayAccess`], after calling this [`ArrayAccess`] will
```
TimDiekmann, line 80:
```suggestion
    /// Enables bound-checking for [`ArrayAccess`].
    ///
    /// After calling this [`ArrayAccess`] will
```

packages/libs/deer/src/schema.rs
```diff
@@ -6,7 +6,7 @@ use core::any::{type_name, TypeId};
 
 use serde::{ser::SerializeMap, Serialize, Serializer};
 
-pub trait Reflection: Sized + 'static {
+pub trait Reflection: 'static {
```
TimDiekmann, line 9: 🎉

TimDiekmann (changes requested):
A few nits, happy to approve as soon as the comments are resolved and conflices has been resolved.

packages/libs/deer/src/lib.rs
indietyp: fixed in [df9fb46](https://github.com/hashintel/hash/pull/1645/commits/df9fb46e67cea4dfeb63eb8ef61879f2d54340dd)

packages/libs/deer/json/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 mod error;
 
 extern crate alloc;
+extern crate core;
```
TimDiekmann, line 22:
```suggestion
```

packages/libs/deer/json/src/lib.rs
```diff
@@ -387,47 +366,91 @@ impl<'a, 'de> deer::Deserializer<'de> for Deserializer<'a> {
...
+            return Err(
+                Report::new(SetBoundedError::CalledMultipleTimes.into_error())
+                    .change_context(ArrayAccessError),
+            );
+        }
+
+        self.remaining = Some(length);
+
+        Ok(())
+    }
```
TimDiekmann, line 417: I wonder why we want to fail on `set_bounded` in JSON at all. There are formats, which require the bound to be set, but JSON does not.

packages/libs/deer/json/src/lib.rs
```diff
@@ -445,43 +468,110 @@ impl<'a, 'de> deer::ArrayAccess<'de> for ArrayAccess<'a> {
...
+
     inner: Map<String, Value>,
     context: &'a Context,
+
+    remaining: Option<usize>,
 }
 
 impl<'a> ObjectAccess<'a> {
-    const fn new(map: Map<String, Value>, context: &'a Context) -> Self {
+    fn new(map: Map<String, Value>, context: &'a Context) -> Self {
```
TimDiekmann, line 481: Instead of a `set_bounded` and having two errors, which basically ensures, that the parameter is passed first, why don't we just pass it here?

packages/libs/deer/json/src/lib.rs
indietyp: `set_bounded` is agnostic to the `*Access` and would be set in the visitor. It enables the following pattern:

```
Expected: [Maybe<u32>; 5]
Given: [1, 2, 3]
Returned: [Maybe::Some(1), Maybe::Some(2), Maybe::Some(3), Maybe::None, Maybe::None]
```

Basically: `T::deserialize` is guaranteed to be called 5 times, no more and no less, and enables `visit_none` for items that do not exit. This enables default values, filler values, etc. I hope that makes sense, but to guarantee that it is precisely `n` items, we need to ensure that `set_bounded` is called at the start and only once.

TimDiekmann (approved):

PR #1649, opened 2022-12-12
Initial shared type definitions

thehabbos007:

This PR introduced some type aliases for us to use for type branding. Having these aliases up front will allow the backend to start using references to types that will soon be branded.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203500651027069/f) _(internal)_ (note that it's not the complete task!)

## 🔍 What does this change?

- Renames `shared/types.ts` to `shared/ontology-types.ts`
- Introduces a `shared/types.ts` file that will contain shared, branded types

- We may want to extensively document these types, which we should do along with the branding work. For now we should get the types in so that @benwerner01 can make use of them for backend changes.
## 🐾 Next steps

- Implement type branding on these aliases
- Change usages of `string` in the backend
- Change usages of `string in the frontend

This should change nothing for ow.


packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,14 @@
-import { BaseUri, VersionedUri } from "@blockprotocol/type-system";
-import slugify from "slugify";
-import { systemUserShortname } from "@hashintel/hash-shared/environment";
-import { frontendUrl } from "./environment";
+// The types exported here are expected to be branded.
+// Perhaps some of these should be put into the subgraph stdlib instead?
```
Alfred-Mountfield, line 2: I think some of them already are, (in terms of the upstream subgraph lib we're not publishing/consuming yet).

About branding them there, ... there is a question as to if we should provide branded things in a public API though.  Maybe we just provide branded re-exports here or something.

Alfred-Mountfield: I'd probably delete this line from the code though, no need to commit it.

packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,14 @@
...
 
-type SchemaKind = "data-type" | "property-type" | "entity-type";
+export type ownedById = string;
+export type AccountId = string;
 
-/** Slugify the title of a type */
-export const slugifyTypeTitle = (title: string): string =>
-  slugify(title, { lower: true });
+export type EntityId = string;
```
Alfred-Mountfield, line 7: Not sure I love this, we've already defined it elsewhere, I'd like to see an explicit task to migrate `EntityId` into this file, and _not_ include it here in the interim, or to just do that now and fix the imports.

packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,14 @@
...
+export type EntityUuid = string;
+export type EntityEditionId = string;
```
Alfred-Mountfield, line 9: This is an object, shouldn't need branding. We do have an `EntityEditionIdString` type though 🤮  (still hate it and don't have a good idea of how to handle it)

packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,14 @@
-import { BaseUri, VersionedUri } from "@blockprotocol/type-system";
-import slugify from "slugify";
-import { systemUserShortname } from "@hashintel/hash-shared/environment";
-import { frontendUrl } from "./environment";
+// The types exported here are expected to be branded.
+// Perhaps some of these should be put into the subgraph stdlib instead?
 
-type SchemaKind = "data-type" | "property-type" | "entity-type";
+export type ownedById = string;
```
Alfred-Mountfield, line 4: Do we want to add
```ts
export type uuid = string
```

And change these to use that?

packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,14 @@
...
-
-      return { ...prev, [key]: definition };
-    },
-    {} as Record<SystemLinkEntityTypeKey, LinkEntityTypeDefinition>,
-  ),
-};
+// Since we already use a template string here, we will keep it. When we
+// brand these types, this may not be needed anymore.
+export type VersionedUri = `${string}/v/${number}`;
+export type BaseUri = string;
```
Alfred-Mountfield, line 14: Similar feeling as expressed above, I'm not sure we should be redefining these types when they're defined in an upstream package.

**_If_** we decide branding in a public library API is a bad idea, maybe we can look at re-exporting them here and shadowing with branded versions, but until then I'd leave them out.

Alfred-Mountfield: (even though `BaseUri` is one of the painful ones without branding)

Alfred-Mountfield (commented):
This makes me _**very**_ happy. Some comments before merging though

packages/hash/shared/src/types.ts
thehabbos007: The only other place I've seen this is in the `packages/hash/subgraph/src/types/identifier.ts` file. I can remove it from here for now

packages/hash/shared/src/types.ts
thehabbos007: Should I omit it from here for now, then?

packages/hash/shared/src/types.ts
thehabbos007: Sure we can do that for now

packages/hash/shared/src/types.ts
Alfred-Mountfield: image
Yeah I'm just saying `EntityId` is used a lot, let's not have competing definitions

packages/hash/shared/src/types.ts
Alfred-Mountfield: Let's omit it for now

packages/hash/shared/src/types.ts
```diff
@@ -1,393 +1,8 @@
...
-import slugify from "slugify";
-import { systemUserShortname } from "@hashintel/hash-shared/environment";
-import { frontendUrl } from "./environment";
+// The types exported here are expected to be branded.
 
-type SchemaKind = "data-type" | "property-type" | "entity-type";
+export type Uuid = string;
+export type AccountId = Uuid;
+export type ownedById = Uuid;
+export type updatedById = Uuid;
```
Alfred-Mountfield, line 6: These should be capitalised, right?

thehabbos007: Yes, you're right 98f9b73f8af33f98478c3f2ba80e81979780ab23

Alfred-Mountfield (approved):

PR #1650, opened 2022-12-12
Move page route to `/@shortname/<pageEntityUuid>`

benwerner01:

This PR moves pages from `/<accountId>/<pageEntityUuid>` to `/@shortname/<pageEntityUuid>`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203491213980897/f) _(internal)_

## 🔍 What does this change?

- adds a `getServerSideProps` function for pages to get the `pageEntityId` and the `pageWorkspace` on first render
- renames the `RouteAccountInfo` react context to `RouteWorkspaceInfo`

Not that I'm aware of.

## ⚠️ Known issues

- The `pageEntityId` is re-calculated in the sidebar even though it has already been calculated in `getServerSideProps`

## 🛡 What tests cover this?

None. This PR was manually tested.

## ❓ How to test this?

1. Run the HASH App locally, and log in
3. Navigate to a page, and observe the shortname in the URL. Ensure the page is behaving as expected.


packages/hash/frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
```diff
@@ -21,7 +21,9 @@ export const PageSidebar: FunctionComponent = () => {
   const router = useRouter();
   const { sidebarOpen, closeSidebar } = useSidebarContext();
   const { activeWorkspaceAccountId } = useContext(WorkspaceContext);
-  const { pageEntityId } = useRoutePageInfo({ allowUndefined: true }) ?? {};
+  /** @todo: re-use the `pageEntityId` provided by `getServerSideProps` */
+  const { routePageEntityId } =
+    useRoutePageInfo({ allowUndefined: true }) ?? {};
```
benwerner01, line 26: I wasn't sure what the best way of avoiding this would be, as the sidebar is not rendered as a child of the `[page-slug].page.tsx` page.

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
```diff
@@ -0,0 +1,75 @@
...
+    [systemUserWorkspace],
+  );
+
+  useEffect(() => {
+    if (systemUserOwnedById) {
+      void (async () => {
+        setLoading(true);
+
+        const entityId = entityIdFromOwnedByIdAndEntityUuid(
+          systemUserOwnedById,
```
thehabbos007, line 41: Quick question regarding the logic in this hook, we are doing this because we need to construct the `EntityId` of the user to use in the BP hook. That hook does not allow specifying `EntityUuid` to fetch the user's workspace, right?

packages/hash/frontend/src/components/hooks/useAccountPages.ts
```diff
@@ -28,8 +31,20 @@ export const useAccountPages = (ownedById: string): AccountPagesInfo => {
...
+  const { orgs } = useOrgs();
+
+  const ownerShortname = useMemo(
+    () =>
+      (
+        orgs?.find(({ orgAccountId }) => orgAccountId === ownedById) ??
+        users?.find(({ userAccountId }) => userAccountId === ownedById)
+      )?.shortname,
+    [users, orgs, ownedById],
+  );
```
thehabbos007, line 44: This seems like something we'd do in more places to get the shortname from an `OwnedById` (`AccountId`). Should this be a hook so we can swap out the implementation down the line?

packages/hash/frontend/src/components/hooks/use-get-workspace-by-shortname.ts
```diff
@@ -0,0 +1,31 @@
+import { useMemo } from "react";
+import { MinimalOrg } from "../../lib/org";
+import { User } from "../../lib/user";
+import { useOrgs } from "./useOrgs";
+import { useUsers } from "./useUsers";
+
+export const useGetWorkspaceByShortname = (
```
kachkaev, line 7: Nit:


```suggestion
export const useWorkspaceByShortname = (
```

We return data, so omitting `Get` from the hook name makes its purpose a bit more intuitive. Otherwise, one may expect something like this:

```js
const getFoo = useGetFoo();
getFoo(someParam);
```

We don't have to address this now, but maybe as part of https://app.asana.com/0/1201095311341924/1202863271046362/f.

We can also move this hook somewhere else. It's currently in `components/hooks`, which may become a dump for all hooks at one point. Instead, we can see which 'feature' this hook belongs to, put it to `../(../)../shared/some-feature/*` and re-export from `../(../)../shared/some-feature.js`.

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
```diff
@@ -0,0 +1,75 @@
+import { extractBaseUri } from "@blockprotocol/type-system";
+import { systemUserShortname } from "@hashintel/hash-shared/environment";
+import { types } from "@hashintel/hash-shared/ontology-types";
+import { entityIdFromOwnedByIdAndEntityUuid } from "@hashintel/hash-subgraph";
+import { getRootsAsEntities } from "@hashintel/hash-subgraph/src/stdlib/element/entity";
+import { useEffect, useMemo, useState } from "react";
+import { useBlockProtocolGetEntity } from "./blockProtocolFunctions/knowledge/useBlockProtocolGetEntity";
+import { useGetWorkspaceByShortname } from "./use-get-workspace-by-shortname";
+
+export const useGetWorkspaceShortnameByEntityUuid = (params: {
```
kachkaev, line 10: Same here re `get` in hook name.

kachkaev, line 10: I’d slightly prefer `payload` instead of `params` here. `params` is plural, which suggests that it’d be an array.

```js
const [foo, bar] = params; // ? 🤔
const { foo, bar } = params; // ? 🤔

const { foo, bar } = payload; // 💡
```

I know TS helps us see the inners of the var, but this does not mean that we can't use additional cognitive hints to help our brains a little.

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
```diff
@@ -0,0 +1,75 @@
...
+            extractBaseUri(types.propertyType.shortName.propertyTypeId)
+          ];
+
+        if (!shortname) {
+          throw Error(
+            `Could not get shortname of workspace with entityUuid "${entityUuid}"`,
+          );
+        }
+
+        setWorkspaceShortname(shortname as string);
```
kachkaev, line 68: Potential race condition here: `setWorkspaceShortname` from `useEffect` \#1 may take longer than for `useEffect` \#2 and we'll end up with old data in the state. We can fix this by adding `const cancelled = false` to the first line of `useEffect` and then set it to true in `return () => {...}`.

kachkaev (changes requested):
The app works well, the diff looks good! I have one comment re `useEffect`, others are optional.

Screenshot 2022-12-14 at 15 14 54

👍

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
```diff
@@ -39,50 +50,143 @@ import { PageIconButton } from "../../components/PageIconButton";
...
+  pageEntityId: EntityId;
   blocks: HashBlock[];
 };
 
+type PageParsedUrlQuery = {
+  "account-slug": string;
+  "page-slug": string;
+};
+
+export const tbdIsPageParsedUrlQuery = (
```
Comment (thehabbos007) [inline packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx:81]: what does `tbd` mean here? Is this naming intentional?

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
```diff
@@ -39,50 +50,143 @@ import { PageIconButton } from "../../components/PageIconButton";
...
+  if (!params || !tbdIsPageParsedUrlQuery(params)) {
+    throw new Error(
+      "Invalid page URL query params passed to `getServerSideProps`.",
+    );
+  }
+
+  const { workspaceShortname, pageEntityUuid } =
+    parsePageUrlQueryParams(params);
+
+  const cookieString = req?.headers.cookie;
```
Comment (thehabbos007) [inline packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx:118]: Should we call this `cookie` and let it be shorthanded below?

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
```diff
@@ -39,50 +50,143 @@ import { PageIconButton } from "../../components/PageIconButton";
...
+  if (!pageWorkspace) {
+    throw new Error(
+      `Could not find page workspace with shortname "${workspaceShortname}".`,
+    );
+  }
+
+  const pageOwnedById =
+    pageWorkspace.kind === "user"
+      ? pageWorkspace.userAccountId
+      : pageWorkspace.orgAccountId;
```
Comment (thehabbos007) [inline packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx:172]: Changing the property name to `accountId` would simplify this as well

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
```diff
@@ -362,7 +464,11 @@ const Page: NextPageWithLayout<PageProps> = ({ blocks }) => {
 
         <CollabPositionProvider value={[]}>
           <PageBlock
-            accountId={routeAccountId}
+            accountId={
+              pageWorkspace.kind === "user"
+                ? pageWorkspace.userAccountId
+                : pageWorkspace.orgAccountId
```
Comment (thehabbos007) [inline packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx:470]: as above, this could be simplified with a rename

thehabbos007 (commented):
Thanks, Ben. I have some comments I think we should address :)

packages/hash/frontend/src/components/hooks/use-get-workspace-by-shortname.ts
```diff
@@ -0,0 +1,31 @@
...
+  shortname?: string,
+): { workspace?: User | MinimalOrg; loading: boolean } => {
```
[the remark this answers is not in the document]
benwerner01: Good spot, addressed in https://github.com/hashintel/hash/pull/1650/commits/081d1b436d76254807e4c1040e30d2ac9e4c8a09

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
```diff
@@ -0,0 +1,75 @@
...
+    workspace: systemUserWorkspace,
+    loading: loadingSystemUserWorkspace,
+  } = useGetWorkspaceByShortname(systemUserShortname);
+
+  const systemUserOwnedById = useMemo(
+    () =>
+      systemUserWorkspace
+        ? systemUserWorkspace.kind === "user"
+          ? systemUserWorkspace.userAccountId
+          : systemUserWorkspace.orgAccountId
```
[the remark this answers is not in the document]
benwerner01: I think the reason I avoided this originally was because I didn't want to introduce usages of `accountId` in the codebase at a time when it was used in way too many places. This has been resolved to a large extent thanks to the introduction of `ownedByid` in the FE, so am happy to make this change now: https://github.com/hashintel/hash/pull/1650/commits/936a4734986f4d4cdfdc74feb321c997728a2e8c

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx]: Addressed in https://github.com/hashintel/hash/pull/1650/commits/936a4734986f4d4cdfdc74feb321c997728a2e8c

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx]: Addressed in https://github.com/hashintel/hash/pull/1650/commits/936a4734986f4d4cdfdc74feb321c997728a2e8c

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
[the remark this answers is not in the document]
thehabbos007: Makes sense. Especially with type branding (which I'll make sure to update for this PR after it's merged), it will be very clear what the types are :)

packages/hash/frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
benwerner01: Discussed this internally, and came to the reaslisation that the `pageEntityUuid` could be used instead. Have addressed in https://github.com/hashintel/hash/pull/1650/commits/48aab54f8e6b3b5e04845cdb1e70dbd731ac0bf7

packages/hash/frontend/src/components/hooks/use-get-workspace-by-shortname.ts
benwerner01: I've renamed the hook as part of https://github.com/hashintel/hash/pull/1650/commits/fa2ee8b97d56b079aca29075b602fd6243d6819b

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
benwerner01: Addressed in https://github.com/hashintel/hash/pull/1650/commits/fa2ee8b97d56b079aca29075b602fd6243d6819b

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
benwerner01: Thanks for spotting! Have addressed in https://github.com/hashintel/hash/pull/1650/commits/e26406722d1dfe390e608a9f679181da90ce2d29

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
benwerner01: If you feel strongly about this I think we should have an internal discussion, as we are using `params` in a lot of the TS code right now so I'd rather be consistent with this approach for now thank introducing a new concept such as `payload`.

packages/hash/api/src/graphql/resolvers/index.ts
```diff
@@ -92,7 +92,7 @@ export const resolvers = {
     pageComments: loggedInAndSignedUp(pageComments),
     blocks: loggedInAndSignedUp(blocks),
     getEntity: loggedInAndSignedUp(getEntity),
-    getAllLatestEntities: loggedInAndSignedUp(getAllLatestEntities),
```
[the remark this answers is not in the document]
benwerner01: It was intentional, yes. Pages are currently publicly accessible on `main`, so in order for that functionality to not be broken by this PR I need to make `getAllLatestEntites` publicly available here, as we need `getAllLatestEntities` to generate the `pageEntityId`. I mentioned this in standup the other day, and there were no objections/alternatives raised.

cc @CiaranMn @nonparibus @hashintel/backend

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx]: This is a bad habit of mine for writing typeguards, it stands for "to be determined". Have removed it in https://github.com/hashintel/hash/pull/1650/commits/136afbbdd12c5d6af013ab471d1aa06794d80228

packages/hash/api/src/graphql/resolvers/index.ts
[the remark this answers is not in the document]
nonparibus: @benwerner01 I added [this ticket](https://app.asana.com/0/1200211978612931/1203468941620598/f) (_internal_) as a priority for BE refinement. I guess this work will have crossover implications, so just wanted you to be aware of it.

kachkaev (commented):
Thanks for addressing my comments @benwerner01! This part of the diff looks great, so happy for someone in the backend team to resolve remaining discussions and approve this PR. We’ll also need to check why the Playwright tests are failing. [The logs](https://github.com/hashintel/hash/actions/runs/3703809685/jobs/6275690752#step:9:22) suggest that it’s something related to routing.

packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx
Comment (benwerner01) [reply packages/hash/frontend/src/pages/[account-slug]/[page-slug].page.tsx]: The reason I went with `cookieString` was because `cookie` makes it sound like we're passing a singular cookie, whereas it's a string of all the concatenated cookies. I realise this will probably not make sense to other readers though, so have addressed in https://github.com/hashintel/hash/pull/1650/commits/d4eeef5aca3b86902d2cf4474b9cc01eaa8613e6

packages/hash/frontend/src/components/hooks/use-get-workspace-shortname-by-entity-uuid.ts
benwerner01: We are doing this because we need to construct the the `entityId` of the workspace to get it using `getEntity`, in cases where we only have the workspaces `entityUuid`. Therefore we need the `ownedById`, which in the case of workspaces is the "system user"'s `accountId` (because users and orgs are owned by the system user).

packages/hash/frontend/src/components/hooks/useAccountPages.ts
benwerner01: Have added a `useWorkspaceShortnameByAccountId` in https://github.com/hashintel/hash/pull/1650/commits/aa68a29649c1a34d38f273c0aa806ca1de230106

packages/hash/frontend/src/blocks/page/MentionView/MentionDisplay.tsx
```diff
@@ -63,36 +63,32 @@ export const MentionDisplay: FunctionComponent<MentionDisplayProps> = ({
...
-          pageTitle =
-            pages.find((page) => page.entityId === entityId)?.title ??
-            "Unknown Page";
+          pageTitle = page?.title ?? "Unknown Page";
         }
 
         return {
           title: pageTitle,
-          href: `/${accountId}/${extractEntityUuidFromEntityId(entityId)}`,
+          href: `/@${page?.ownerShortname}/${extractEntityUuidFromEntityId(
```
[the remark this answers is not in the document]
benwerner01: I agree we should introduce something along the lines of what you're describing, so have added [a task](https://app.asana.com/0/1203179076056209/1203535666547772/f).

In the meantime, I've encapsulated the URL generation for the page in a `constructPageRelativeUrl` function in https://github.com/hashintel/hash/pull/1650/commits/03f1b93835f5ed6dee3f72701309c355fa56c58c

thehabbos007 (dismissed):
Great stuff!

PR #1652, opened 2022-12-13
Consolidate codegen scalars into shared file

thehabbos007:
## 🌟 What is the purpose of this PR?

This PR consolidates codegen scalars into a shared file through converting the codegen configs to typescript and importing the config.

This was done out of annoyances adding type branding (it was a 5 minute task)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203393349856541/f) _(internal)_

## 🔍 What does this change?

- Convert `codegen.yml` into `codegen.config.ts` which is a copy-paste through converting yaml to json
- Making a single scalar mapping definition in the shared package and importing it from the configs

## 📜 Does this require a change to the docs?

No changes

## 🐾 Next steps

- Type branding ([asana ticket](https://app.asana.com/0/0/1203500651027069/f))
## ❓ How to test this?

1.  Checkout the branch
1. run `yarn` or `yarn codegen`

## 📹 Demo

image

PR #1655, opened 2022-12-13
Initial type branding

thehabbos007:

This PR introduces type branding and adds it for the surface of the backend API.
This may already have found one issue at https://github.com/hashintel/hash/pull/1655/files#diff-f181b3a2e93b2e81dbe1e2f161e0892583d88835f89cc5bbe98332e96549bbbc

The reason why I did not use a library such as https://www.npmjs.com/package/ts-branding to do the branding is that it would add a transient dependency on everything that needs to make use of the branded type within the application. Having our own (fairly simple) type branding constructors did not seem like a devleopment burden, as they can easily be reasoned about and understood without needing to abstract away the implementation in a library. If we want to have "flavoring" and construtors for the type brands, we may want to look into using the linked library, but for now this suffices.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203500651027069/f) _(internal)_

## 🚫 Blocked by

- [x] #1650 
- [x] #1652 

## 🔍 What does this change?

- Adds a type branding framework 
  - Done through adding  two simple types that is able to add `_kind` and `_base` to emulate 'nominal types'
- Wrappers for the subgraph lib to cast to expected types
  - in anticipation for the subgraph not residing fully in the HASH repo, these wrappers are intentionally redundant to ensure we can smoothly migrate to the BP based subgraph stdlib.

## ⚠️ Known issues

- The frontend does _not_ pass lints right now. We may want to stack frontend changes onto this PR.
- Because we are wrapping only parts of the subgraph library, we aren't getting type branding in all the places we would like. Ideally the subgraph structure is well-typed with type branding, but that is something to be done in the BP.
- An open question is how we want to type `ownedById` throughout the codebase. Currently it is typed as the most permissive `AccountId` and not `OwnedById`. This means that an `OwnedById` or `UpdatedById` could be passed in for those methods. This does slightly make the two mentioned subtypes obsolete, as we don't use them anywhere. For these to be useful, we would want to type the subgraph properly and completely.

## 🛡 What tests cover this?

- `tsc` for the backend and integration tests

## ❓ How to test this?

1.  Checkout the branch 
1.  Run tsc lints for the backend/integration packages


packages/hash/shared/src/types.ts
```diff
@@ -1,8 +1,46 @@
-// The types exported here are expected to be branded.
+import {
+  EntityId,
+  extractEntityUuidFromEntityId as extractEntityUuidFromEntityIdSubGraph,
```
Alfred-Mountfield, line 3: Can we not capitalise graph here, `subgraph` is a single word :)

packages/hash/shared/src/types.ts
```diff
@@ -1,8 +1,46 @@
...
+ * The type-branding type to support nominal (name based) types.
+ */
+export type Brand<Base, Kind extends string> = Base & Branded<Base, Kind>;
+
+/** An account ID of an actor that is the owner of something */
+export type OwnedById = Brand<string, "OwnedById">;
+/** An account ID of an actor that has updated something */
+export type UpdatedById = Brand<string, "UpdatedById">;
+/** An ID to uniquely identify an account (e.g. a User or an Org) */
+export type AccountId = Brand<string, "AccountId"> | OwnedById | UpdatedById;
```
Alfred-Mountfield, line 25: I think we probably want an `ActorId` alongside these (inheriting from `AccountId`)

Alfred-Mountfield, line 25: Are you not able to do `export type OwnedById = Brand<AccountId, "OwnedById">`?

Is there a way we can rework the type to allow that possibly?, I'm guessing the issue is it would override the readonly annotations rather than wrapping them 🙃

packages/hash/shared/src/types.ts
```diff
@@ -1,8 +1,46 @@
...
+) => [OwnedById, EntityUuid];
+
+export const extractOwnedByIdFromEntityId =
+  extractOwnedByIdFromEntityIdSubGraph as (entityId: EntityId) => OwnedById;
+
+export const extractEntityUuidFromEntityId =
+  extractEntityUuidFromEntityIdSubGraph as (entityId: EntityId) => EntityUuid;
+
+/** If the underlying entityUuid is an accountId, use this cast to convert the type */
+export const extractAccountIdAsEntityUuid =
```
Alfred-Mountfield, line 45: I was thinking about this in the discussion we were just having about the hierarchy of branding and if we actually want/need an `OwnedById`.

First of all, I think the naming of this is backwards, it should be `extractEntityUuidAsAccountId` right?

Secondly, I'm not sure this helper method is a good idea, it'd be too easy to call this on the wrong kind of `EntityId` with a "trust me" approach rather than trying to embed further type safety.

What if we had an `EntityId` subtype for `AccountEntities` or something? (maybe a slightly less confusing name would be good), and then we have an `extractAccountId` function that takes in that branded `EntityId`. Then you'd need to explicitly say "the `entityId` I have here is for one of our unique account entities"

Alfred-Mountfield (commented):
Haven't done a full review by any means, just some questions and thoughts I'd like to know the answers to prior to further review :)

But I can't express how excited I am that we're following up on my suggestion for this, it looks like it'll save us such headaches and is a great pattern for us to start using a lot.

packages/hash/shared/src/types.ts
thehabbos007: I spent a fair bit of time trying to allow this, and unfortunately the type always becomes `never` because of the way we compose the types right now. I can give it another shot as it would make the hierarchy a lot simpler to define

packages/hash/shared/src/types.ts
thehabbos007: Yes, I see now that the name is backwards. And I do like that approach instead, I'll look into that :+1:

packages/hash/shared/src/types.ts
thehabbos007: aef0ef5ffe42b83cd9364e155474280232ee0714

packages/hash/shared/src/types.ts
thehabbos007: Hmm, I see why we may want that, but I have a feeling it may be equivalent to `UpdatedById` in terms of how useful it is. We wouldn't be returning `ActorId` from anything, only taking it as a param. If it's inheriting from `AccountId` it would require an explicit upcast to be used as an argument

Alfred-Mountfield: (We collaborated offline and came up with [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQgTgQwHYBMIoDwwQZ2hAD2AlRygFckBrJAewHckoAfWRVdDSmhpAGig5gcAJZIA5gD4BAaTEpJUALxQA3gCgoUAPQAqXVAAqAC2ihIUWgDMowU1ABuCADbloOMAgDGZ4wmBQImQARuxoKFC62ppQcBAIKLRIziBQAPrBuBAAXLBZANwxegYm0JQiAI5uUEgIALbQVrRwtvahyOG24BAAdJHRWnEJSSnpVPK5AIJwiCBYWVCExKRsHZzcdIwCQqISigD8UADaPafYeEcARGnjqJcAurLy97lyqJKF6gC+n8WaJfZzBAALTtVBicRdCzAWiCchgMDNAJ0OpiFxQAAUtQaUEyeBQAEooRAcD1-tFCIi4AEgatUPM8E9UIsiCQUGQdhDFCpztAAGR08IMiBMhSfdTabTEqCTLxeWiUYAASQiPLCGE5EgEl1l8sVKsuknUEqltIAYrRYSpNZCBRpJVp0nicoJhBC+CbHTcJsdLhbaJcBKcejajvd7p6vsbzZblIKNW6tVA-ZbDb9Td08i1rYnbWpPVoMllcjaPQ7C7cULkrv7A8nsHA68HQ+HI+paQ24-AOhh-dqG2njZ6OwgAF5xm1QO0Fp3F127cRlqUVn011P9hCNjejpunFsRh1RkfjtU9hvbwfqeVIIRQJq0XL+uOXS5QXBQf2Fa+3zJwXKdlQXzfMgGy-JIfzHf8x2fV932wUdPm-GkECoCAnxUDECEfS0iSURRVB+K9wOQ1CAMxLCs1w-DCKQ2wUIgeC40wqDRyotRCPbej-Qxe8CUKYAuMtDFfz4zjUO4zJWM+ATSM3HicP4+iG2EzdRJkhi5Mk0SxI00d5NoNSlLHFS4EM2S9K0wogA))

Alfred-Mountfield: Revamped [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQgTgQwHYBMCiAPYEkGcCWA9kgDwwK4QA0UA0vqlBFjirlAN4C+AfFALycAUFCgB6AFQSoAFQAW0UJCiEAZlGAKoANwQAbAK7RcYBAGNFchMCj52AI0SoIKKBLEiocCAhTE9IFAA+vYUEABcsGEA3J6S0vLQBkj4AI5GUEgIALbQqoRwGlqOyCguGuAQAHRuHqLevv6BQQDWDCiR9KixQlw98SIJWkoQALQlqAwA5hXKwIRQuAZgYAU2SITZDPpQABRZuVChlCgAlLMQuFWDHsyrcDYjsE5lKGRhNF2uzNio7LjAODTPiCciUJgsP7PUqYX4EYgkBiqCCFAByl2wKDB1FsSGRaIxLi+fE8AH4ouCAGTQ1CwnDw0jogEubE0JmYr5QakcADaQVxdHaAF1IskWhsAO5IXhQTyRbFcmnoSEM96UGi8-kMQWoEVQMWS6U8fpSQZQACCSBUYGARCyeigqmSZltxA0Cwmrk0ikquOZvhU6juenwZnwwACUDMFFtSBm4ZuQjuayjxABRxeAigJE5P1Y-0B0xorKVMl9eah8FKLjVOOJPF2GHlHygbVQpM67XO-D4HE83mABjgVowUAo+qQ4sIUrH7CrqDLkF6QiETwAYoQFqCXjWOFAytlCAB5ewAKwgLsiAKBcagXBoACIN4QH8aV09yIVt9W3s-H5-Xx6D8EAALyzedXneOB-1AwCVzMNMbHyQhImfLM9wPY8zwvYBIgfXBNggAA1fQjAfO9ZygZ9YjEMRUzwJDN1Qzcs09XYMIgQ8T3PS8oAfCVfAlciuFOHoEIYo4EDgZsvwzUpdmQ85x0-Gi6PE9NQmkqJZLYzTRPgxDJJA5swMEXSpKUudQNU+iNNAkzWJeXZQhA-ShHUx4EBaCA0MERtmMIbtezvWIPI0LyIE-LN-O0oLOBC9zDOACLyFMvYmyiVyBGCvp3wi58FM3fTku8gq9NiEqfM3ZzQLc2ioAAAWAXBRjuHDWrgOAClXFKpMKwKKt6uAargYqhpq1z+jopqWral0Oq6uAeu81L+v0+qZtajBIHmlFFuWyLQJGsaVqOlzRKAA)

thehabbos007: Added in e8af67a4df3371b721fd7a7c9ed00e40f2545246, adjusted the inner brand property names to not appear in intellisense :+1:

packages/hash/shared/src/types.ts
thehabbos007: 0e0e8ab0b3cdb887fc1f18aeb015bd5630f79675 should be addressed here. Using the `brand` fn directly

packages/hash/api/src/graph/system-user.ts
```diff
@@ -62,14 +64,15 @@ export const ensureSystemUserAccountIdExists = async (params: {
   );
 
   if (existingSystemUserEntity) {
-    systemUserAccountId = extractEntityUuidFromEntityId(
-      existingSystemUserEntity.metadata.editionId.baseId,
+    systemUserAccountId = extractAccountId(
+      brand(existingSystemUserEntity.metadata.editionId.baseId),
```
Alfred-Mountfield, line 68: Places like this make me wonder if we want to actually use the casting form rather than the function. Genuinely unsure which, the `brand` is slightly more ergonomic it seems because it infers from the type, but also here it's very unclear to me why you're branding the thing (although I'm pretty sure it's because of my suggestion about having a variant of `EntityId`)

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
```diff
@@ -72,7 +72,9 @@ export const pages: ResolverFn<
   QueryPagesArgs
 > = async (_, { ownedById }, { dataSources: { graphApi }, userModel }) => {
   const accountEntityId = ownedById
-    ? entityIdFromOwnedByIdAndEntityUuid(systemUserAccountId, ownedById)
+    ? // When type branding, we would need to cast `ownedById` to an `EntityUuid`.
+      // We may want to special-case this casting and have an explicit function to do it.
+      entityIdFromOwnedByIdAndEntityUuid(systemUserAccountId, ownedById)
```
Alfred-Mountfield, line 77: Are we not type branding? What's going on here?

packages/hash/api/src/graphql/resolvers/middlewares/canAccessAccount.ts
```diff
@@ -11,7 +11,7 @@ import { ResolverMiddleware } from "./middlewareTypes";
 export const canAccessAccount: ResolverMiddleware<
   GraphQLContext,
   {
-    ownedById: Scalars["ID"];
+    ownedById: Scalars["AccountId"];
```
Alfred-Mountfield, line 14: We probably want `OwnedById` here too right?

packages/hash/api/src/graphql/resolvers/ontology/property-type.ts
```diff
@@ -27,9 +28,9 @@ export const createPropertyTypeResolver: ResolverFn<
   const createdPropertyType = await createPropertyType(
     { graphApi },
     {
-      ownedById: ownedById ?? userModel.getEntityUuid(),
+      ownedById: brand(ownedById ?? userModel.getEntityUuid()),
```
Alfred-Mountfield, line 31: If we change `ownedById` to `OwnedById` then I guess we could have
```suggestion
      ownedById: brand(ownedById) ?? userModel.getEntityUuid(),
```

and ideally `userModel.getEntityUuid()` will change to a nicer standalone function anyway like `extractAccountIdFromUser` or something

packages/hash/frontend/src/components/hooks/useGetAccountIdForShortname.ts
```diff
@@ -14,7 +19,7 @@ export const useGetAccountIdForShortname = (
     const userBaseId = users?.find((user) => user.shortname === shortname)
       ?.entityEditionId.baseId;
     const userAccountId = userBaseId
-      ? extractEntityUuidFromEntityId(userBaseId)
+      ? extractAccountId(brand(userBaseId))
```
Alfred-Mountfield, line 22: Yep, the more I see this the more I think we'd want to see it as `extractAccountId(userBaseId as AccountEntityId)` 

But I don't know how we'd enforce that if both way works :(

Alfred-Mountfield (changes requested):
I think usage of `OwnedById` got lost in transit somewhere, can you try and update?

packages/hash/api/src/graph/ontology/primitive/data-type.ts
```diff
@@ -29,7 +29,7 @@ import { NotFoundError } from "../../../lib/error";
  */
 export const createDataType: ImpureGraphFunction<
   {
-    ownedById: OwnedById;
+    ownedById: AccountId;
```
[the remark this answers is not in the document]
thehabbos007: I made this pass in the last commit since I would expect us to allow any `AccountId` to be the owner of the ontology types. Otherwise we'd have to `brand` to upcast at the GQL resolvers to turn the `AccountId` into an `OwnedById`.
The types that inherit from `AccountId` such as `OwnedById` would make most sense to be returning from the subgraph, but for creation I suppose we want to be as permissive as possible as an `UpdatedById` is still a valid `ownedById` to pass into this function

packages/hash/api/src/graph/system-user.ts
thehabbos007: Hmm, good point. It is easy to cast something incorrectly, whereas the `as` cast would check the base type. And yes the reason for this is because of the `EntityId` variant

packages/hash/api/src/graph/ontology/primitive/data-type.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Right but it's the whole point to require branding for that 😅, especially for this:

> but for creation I suppose we want to be as permissive as possible as an UpdatedById is still a valid ownedById to pass into this function

That's the one where we wanted to absolutely require it to be explicit, because we don't want us mixing up IDs so we require you to go "I know this is an `UpdatedById`, I'm going to turn it into an `OwnedById`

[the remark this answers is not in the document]
thehabbos007: 2e79e039d40883c9728dd67f49e50266b1d43ebf Hopefully that should be all. I need to use a mix of `brand` and `as` because of `null` choices. I wonder if we want to/can make that more ergonomic?

packages/hash/api/src/graphql/resolvers/knowledge/page/page.ts
thehabbos007: 2e79e039d40883c9728dd67f49e50266b1d43ebf sorry, I missed my own 'todo' here :)

packages/hash/api/src/graphql/resolvers/middlewares/canAccessAccount.ts
thehabbos007: Yes, thank you! 3c9018ae5cc91804923a9ed854b390674f6bfba7

packages/hash/api/src/graphql/resolvers/ontology/property-type.ts
thehabbos007: Good idea! da18df6a18a02639b03d7c83bc9867551d087b09

packages/hash/api/src/graph/system-user.ts
thehabbos007: Changed in 79a98adc0227c976094cc723a4b84e1362ffa60d

packages/hash/frontend/src/components/hooks/useGetAccountIdForShortname.ts
thehabbos007: Changed in 79a98adc0227c976094cc723a4b84e1362ffa60d

packages/hash/shared/src/types.ts
```diff
@@ -1,8 +1,80 @@
...
+
+/** An ID to uniquely identify an account (e.g. a User or an Org) */
+export type AccountId = Brand<Uuid, "AccountId">;
+
+/** An account ID of an actor that is the owner of something */
+export type OwnedById = Brand<AccountId, "OwnedById">;
+/** An account ID of an actor that has updated something */
+export type UpdatedById = Brand<AccountId, "UpdatedById">;
+
+/** A Uuuid that points to an Entity without any edition */
```
Alfred-Mountfield, line 52:
```suggestion
/** A `Uuid` that points to an Entity without any edition */
```

packages/hash/shared/src/types.ts
```diff
@@ -1,8 +1,80 @@
...
+export type EntityUuid = Brand<Uuid, "EntityUuid">;
+
+/** A Uuuid that points to an Account Entity without any edition */
```
Alfred-Mountfield, line 55:
```suggestion
/** An `EntityId` which is the base of an Account Entity */
```

Alfred-Mountfield (changes requested):
Two tiny nits. I think the more I see the use of `brand` the less I like it, but also at this point I think we should stop iterating and just see how the ergonomics are in usage. We probably want to stick with one or the other (using the function or just doing casts)

Alfred-Mountfield (approved):

PR #1657, opened 2022-12-13
Support property arrays in entity editor

yusufkinatas:

This PR adds an important missing functionality to entity editor. Which is supporting properties with `allow multiple values` checked. Entity editor was supporting arrays before, but that was only for `references to array custom property array definitions`, not `allow property values`

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432084/1203549170064068/f) _(internal)_

## ❓ How to test this?

1.  Open entity editor
2. Create a type with some properties that ticks `allow multiple values` field
3. Create an entity from that type
4. Confirm that entity editor works as expected with `allow multiple values`

## 📹 Demo

Check the list icons on the first column of the entity editor.

Screenshot 2022-12-13 at 22 29 32

Screenshot 2022-12-13 at 22 29 37


nathggns:
Reviewing now

nathggns (approved):
This does allow entering multiple values when the checkbox is checked. It does not do any min/max values constraints. We'll need to do that in a follow up

PR #1664, opened 2022-12-14
Deprecate old endpoints

TimDiekmann:

In order to notify the TS backend, that the old endpoints are going to be removed, we annotate them with a `deprecated` attribute

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202979057056095/f) _(internal)_


benwerner01 (approved):
Thank you @TimDiekmann, the annotations are working for me locally as expected, and these deprecated annotations will come in handy in my model class refactor 👍

PR #1667, opened 2022-12-14
Cleanup traits and lifetime/generic names

TimDiekmann:

We have quite a lot traits a few functions were defined on multiple types, which could be moved to those traits. Also, we are not consequent with our generics names and lifetime names.

## 🔗 Related links

- [Move common methods on types (ontology and/or entities) to the traits](https://app.asana.com/0/1203363157432084/1203449432963809/f) _(internal)_
- [Unify traits, which are implemented on the same type](https://app.asana.com/0/1203363157432084/1203449432963806/f) _(internal)_

## 🔍 What does this change?

Please see the list of commits below, every commit is very scoped


packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -590,10 +590,7 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
             .collect())
     }
 
-    async fn get_entity<'f: 'q, 'q>(
-        &self,
-        query: &'f StructuralQuery<'q, Entity>,
-    ) -> Result<Subgraph, QueryError> {
+    async fn get_entity(&self, query: &StructuralQuery<Entity>) -> Result<Subgraph, QueryError> {
```
TimDiekmann, line 593: Not explicit in the commit list, but as a follow-up of #1665 this can greatly be simplified as well (same for ontology types).

packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -33,6 +33,9 @@ use crate::{
 /// [`store`]: crate::store
 pub trait Record {
     type Path<'q>: QueryPath;
+    type Metadata;
+
+    fn metadata(&self) -> &Self::Metadata;
```
Alfred-Mountfield, line 38: I'm wondering if it's worth a doc-comment explaining why we've done this. I think I'd be pretty confused if I went and looked at the `Entity` definition and saw methods for accessing everything, but then didn't find `metadata`.

Maybe we just have this shadow and call `self.metadata()`?

packages/graph/hash_graph/lib/graph/src/ontology/mod.rs
```diff
@@ -265,6 +246,10 @@ impl OntologyTypeWithMetadata for DataTypeWithMetadata {
             metadata,
         }
     }
+
+    fn inner(&self) -> &Self::OntologyType {
```
Alfred-Mountfield, line 248: Same question as before about just shadowing the previous method

Alfred-Mountfield (approved):
What a surprising pleasure that was to review, some of the best separation into commits I've seen, thank you!

A few questions/suggestions but I don't feel strongly about them

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#same_name_method) to avoid shadowing of methods, where the method is defined on the trait and on the struct. I don't know if we want to disable the lint, probably it's worth it.

Alfred-Mountfield: Hm, if we have the lint _enabled_ then it's maybe not as bad. Because at least if I went and tried to implement `metadata` (because I incorrectly though it was missing) then clippy would warn me and I'd discover the other one

Alfred-Mountfield: As much as I don't like it, maybe that's actually better, as now we only have one place to refactor if we rename the accessor for example

PR #1670, opened 2022-12-14
Use structural query over plain SQL in ontology update

TimDiekmann:

With #1665 and #1667 it's possible to use structural queries instead of using plain SQL in a generic context


packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/data_type.rs
```diff
@@ -156,7 +156,9 @@ impl<C: AsClient> DataTypeStore for PostgresStore<C> {
                 .change_context(UpdateError)?,
         );
 
-        let (_, metadata) = transaction.update(data_type, updated_by_id).await?;
+        let (_, metadata) = transaction
+            .update::<DataType>(data_type, updated_by_id)
+            .await?;
```
TimDiekmann, line 161: I have no idea, why the explicit generic is required here.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -345,17 +345,14 @@ where
                 .change_context(UpdateError));
         }
 
-        let base_uri = uri.base_uri();
-
         // TODO - address potential race condition
         //  https://app.asana.com/0/1202805690238892/1203201674100967/f
+        let previous_ontology_type =
+            <Self as Read<T::WithMetadata>>::read_one(self, &Filter::for_base_uri(uri.base_uri()))
```
TimDiekmann, line 351: Somehow, Rust fails to detect, that we don't want to read an `Entity` but `T::WithMetadata`, so this specifies the trait explicitly

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -332,6 +330,8 @@ where
     ) -> Result<(VersionId, OntologyElementMetadata), UpdateError>
     where
         T: OntologyDatabaseType<Representation: Send> + Send + Sync,
+        T::WithMetadata: PostgresRecord + Send,
+        for<'p> <T::WithMetadata as Record>::QueryPath<'p>: OntologyQueryPath + Send + Sync,
```
TimDiekmann, line 334: These bounds are required for `Read<T::WithMetadata>`

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -1,7 +1,6 @@
 mod knowledge;
 mod ontology;
 
-mod context;
```
TimDiekmann, line 4: Finally, I am able to remove this. 🎉

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
Alfred-Mountfield: I had a quick attempt locally and couldn't find an easy way to do it either, I do wonder if we have too many traits here, it does seem overly complicated

Alfred-Mountfield (approved):
I'm happy to approve this just because we've been able to delete so much unused code. I don't think we're quite where we want to be though, our traits seem to complicated, and they're causing problems for the compiler's inference, not to mention I'm really concerned about the maintenance burden, I think adding new methods would be an absolute pain at the moment because it really would be unclear what traits you need to be requiring.

I'd like us to continue to look at the traits and figure out how to simplify it even further

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: This is mainly because we have a generic implementation for `Read<T::WithMetadata>` instead `impl Read<DataTypeWithMetadata>` etc.. I expect this to be simplified in a not-too-distant future.

PR #1672, opened 2022-12-14
Pause hydration until type system has loaded

nathggns:

This attempts to use `dynamic` to pause hydration until the type system has initialized, avoiding the need to check within individual components

### Todo

- [x] Remove useless `useInitTypeSystem` and its calls

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203492058388310/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Wraps the whole app in a component wrapped in `dynamic`, which waits for type system to load


Alfred-Mountfield (commented):
Thanks Nate, I think it looks good, it _seems_ as simple as I'd hoped! 

Do you want to remove the old hooks before we merge this or were you intending that to be a follow-up? Also I'm guessing we don't need to leave the test.page once we're happy with the PR's functionality and can remove prior to merge?

nathggns:
Will remove the hook and test page before merge – thanks @Alfred-Mountfield

Alfred-Mountfield (changes requested):
I've tried out the branch and it would seem we did hit a problem :(

Doesn't seem like fonts are loading (not sure if there are other things going wrong).

Also linting is failing now

nathggns:
@Alfred-Mountfield I can't repro the fonts issue. Where did you witness that? 

image

(left main, right this branch)

Alfred-Mountfield (approved):
Fonts seem to be working for me again, no idea what could have caused that (At the time, I recreated it when pulling your branch locally, and also checking the deployment on the PR)

PR #1674, opened 2022-12-14
Zack/making macro

zdannar:

Adds a simple macro to the error-stack-macros crate.  The point is avoid boilerplate code and make error_stack easier to work with.
## 🔗 Related links

## 🚫 Blocked by

NA

- [ ] ...

## 🔍 What does this change?
This change just reduces boilerplate using error_stack. 

- ...

- I added a small doc/doctest to the lib.rs file. 
## ⚠️ Known issues

NA

## 🛡 What tests cover this?

-

## ❓ How to test this?

cargo test?

### DIFF
--- a/packages/libs/error-stack/macros/Cargo.toml
+++ b/packages/libs/error-stack/macros/Cargo.toml
@@ -13,9 +13,9 @@ keywords = ["errorstack", "error-handling", "macro", "derive", "no_std"]
 categories = ["rust-patterns", "no-std"]
 
 [lib]
-proc-macro = true
+proc-macro = false
 
 [dependencies]
+error-stack = { version = "0.2.4", default-features = false }
 
 [dev-dependencies]
-error-stack = { version = "0.2.4", default-features = false }
--- a/packages/libs/error-stack/macros/Cargo.toml
+++ b/packages/libs/error-stack/macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "error-stack-macros"
-version = "0.0.0-reserved"
+version = "0.0.1"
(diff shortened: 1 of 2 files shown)


zdannar:
Honestly I am not sure how you envisioned the error-stack-macros crate, but this change is simple.

Alfred-Mountfield:
Hi Zack,

Thanks for your contribution, appreciate all effort to try and improve the usage of the library! 

I can definitely see what you're going for with this change, but I think it's perhaps better placed somewhere else. The `error-stack-macros` crate is just a reservation for now, as we expect in the future we'll likely need a place to put proc-macros (that's why we had `proc-macro = true` in the `Cargo.toml`). The intentions for the crate that I just described also means we don't really want to depend on `error-stack` within this crate, as it will instead be a direct dependency _of_ `error-stack`.

If you're looking for an ergonomic way to make very simple `Error` types with a display implementation, we recommend in the docs that users look at using `thiserror`, it plays nicely with `error-stack` and it's what we utilise in our own code. Otherwise you can of course just keep this macro in your own code, but I think we'd prefer to leave the options more open to people, and let them pick the abstraction they find most helpful, rather than adding something like this which will inevitably lead people in a certain direction.

As such, I'm going to close this PR, but we might look at adding a README.md to this crate just to help convey the intentions a bit better! Thanks

nonparibus:
@zdannar Just to follow-up on what Alfie has said here... we're grateful for your contribution/thoughts! If you're interested in advancing error-stack and would like to kick about further ideas, please feel free to join us on our Discord (https://hash.ai/discord)

PR #1675, opened 2022-12-14
Optimize querying and add benchmarks for complete graphs

Alfred-Mountfield:

We were resolving an entity, and then passing its identifier down to the `traverse` function, which would promptly resolve it again as we hadn't inserted it into the dependency map yet. This fixes it.

This improves performance (in my tests) by around 20%.

This PR also adds some benchmarks for testing a "complete" graph, alongside the existing edgeless graph. (See module documentation)


packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -303,8 +318,15 @@ impl<C: AsClient> PostgresStore<C> {
...
+                        let left_entity_edition_id = left_entity.metadata().edition_id().clone();
+
+                        subgraph.vertices.knowledge_graph.insert(
+                            left_entity_edition_id,
+                            KnowledgeGraphVertex::Entity(left_entity),
+                        );
+
                         self.traverse_entity(
-                            left_entity.metadata().edition_id(),
+                            left_entity_edition_id,
```
Alfred-Mountfield, line 329: The code modification could have been even smaller if I just changed `traverse_entity` to take the actual record rather than the identifier, (it also would be very mildly faster as it'd save a lookup in the map), however I think we want to keep this general and based on identifiers. As we may be able to combine the various `traverse` methods (e.g. `traverse_entity_type`) and in those scenarios we may _not_ have the actual record, but just the URIs from inside the types.

packages/graph/hash_graph/bench/benches/read_scaling/knowledge/linkless/mod.rs
```diff
@@ -0,0 +1,7 @@
+/// An "edgeless" graph is a graph without any edges.
+///
+/// These benchmarks are for a datastore where the knowledge-graph component is an edgeless graph,
+/// meaning there are no link entities, hence the name "edgeless".
```
Alfred-Mountfield, line 4:
```suggestion
/// meaning there are no link entities, hence the name "linkless".
```

😅

packages/graph/hash_graph/bench/benches/read_scaling/knowledge/linkless/mod.rs
```diff
@@ -0,0 +1,7 @@
+/// An "edgeless" graph is a graph without any edges.
+///
+/// These benchmarks are for a datastore where the knowledge-graph component is an edgeless graph,
+/// meaning there are no link entities, hence the name "linkless".
```
Alfred-Mountfield, line 4: Just to note, I don't think we can do an "edgeless" graph easily, because every entity should have an entity type.. and in turn those probably should have property types and data types to actually have any meaningful size of data, so.. I think `linkless` is testing the case we care about

packages/graph/hash_graph/bench/benches/read_scaling/knowledge/complete/mod.rs
```diff
@@ -0,0 +1,11 @@
+/// An "complete" graph is a graph where every vertex pair is connected by a unique edge.
```
Alfred-Mountfield, line 1:
```suggestion
/// A "complete" graph is a graph where every vertex pair is connected by a unique edge.
```

packages/graph/hash_graph/bench/benches/read_scaling/knowledge/complete/entity.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    group.sample_size(10);
+    group.sampling_mode(SamplingMode::Flat);
+
+    // We use a hard-coded UUID to keep it consistent across tests so that we can use it as a
+    // parameter argument to criterion and get comparison analysis
+    let account_id =
+        AccountId::new(Uuid::from_str("bf5a9ef5-dc3b-43cf-a291-6210c0321eba").unwrap());
+
+    for size in [1, 5, 10, 25, 50] {
```
TimDiekmann, line 234: I think we shouldn't use `50` here as the benchmark will run quite a while (2500 links).
```suggestion
    for size in [1, 5, 10, 25 /* , 50 */] {
```

TimDiekmann (approved):
Amazing find, and thanks for improving the benchmarks as well!

One suggestion, but if you disagree feel free to ignore it.

packages/graph/hash_graph/bench/benches/read_scaling/knowledge/complete/entity.rs
Alfred-Mountfield: It takes about 10 minutes for that specific benchmark, which is bad and I'm already slightly concerned about how long our benchmarks take to run. However, for scenarios where we don't particularly care I think we can just disable the benchmark, but in most scenarios I think this case is quite important as it's hard to see how poor the performance change is as the numbers increase (as in 1 -> 5 -> 10 doesn't paint a good picture, so we kind of need 25 -> 50)

```
Finished seeding database entity_scale with 5 entities and 25 link entities after 271.603ms
Benchmarking scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #2: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 8.3s.
scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #2
                        time:   [725.90 ms 791.88 ms 853.66 ms]
                        change: [-23.084% -13.450% -3.2853%] (p = 0.04 < 0.05)
                        Performance has improved.
Seeding database: entity_scale
Finished seeding database entity_scale with 10 entities and 100 link entities after 335.602ms
Benchmarking scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #3: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 33.4s.
scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #3
                        time:   [2.5267 s 2.5890 s 2.6570 s]
                        change: [-20.533% -15.123% -10.038%] (p = 0.00 < 0.05)
                        Performance has improved.
Seeding database: entity_scale
Finished seeding database entity_scale with 25 entities and 625 link entities after 569.674ms
Benchmarking scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #4: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 146.6s.
scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #4
                        time:   [14.913 s 15.191 s 15.518 s]
                        change: [-17.158% -14.440% -11.811%] (p = 0.00 < 0.05)
                        Performance has improved.
Seeding database: entity_scale
Finished seeding database entity_scale with 50 entities and 2500 link entities after 626.593ms
Benchmarking scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #5: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 571.1s.
scaling_read_entity_complete_one_depth/get_entity_by_id/Account ID: `bf5a9ef5-dc3b-43cf-a291-6210c03... #5
                        time:   [57.350 s 57.491 s 57.650 s]
                        change: [-21.605% -19.122% -16.836%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Notice 15s to 57s, I think this is one of the most important metrics we should be targeting to bring down as we optimise further.

TimDiekmann: Good call, let's leave it in. 🙂

TimDiekmann (approved):

PR #1677, opened 2022-12-15
Avoid enumerations in Rust representation of the subgraph

TimDiekmann:

In order to dynamically allow insertion into the subgraph without implementing it for each ontology type separately, this changes the Rust representation of the subgraph to hold four maps instead of two general maps. This also makes the usage of the subgraph at Rust much easier as seen in the integration tests.

While this seems to be less efficient, I was not able to observe any performance change.

The OpenAPI specs did not change!

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1203363157432084/1203444301722127/f) _(internal)_

## 🔍 What does this change?

- Change the subgraph representation
- Move OpenAPI related structs to the `api` module
- Adjust integration tests
 
## 🐾 Next steps

- Generalize insertion of elements into the subgraph


packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/vertices/vertex.rs
```diff
@@ -9,12 +9,31 @@ use crate::{
 #[derive(Debug, PartialEq, Eq, Serialize)]
 #[serde(tag = "kind", content = "inner")]
 #[serde(rename_all = "camelCase")]
+#[expect(clippy::enum_variant_names)]
```
Alfred-Mountfield, line 12: I don't really understand how we keep removing and re-adding this? Is something confusing clippy

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/vertices/vertex.rs
```diff
@@ -84,6 +103,7 @@ impl ToSchema for KnowledgeGraphVertex {
 #[derive(Debug, PartialEq, Eq, Serialize)]
 #[serde(rename_all = "camelCase")]
 #[serde(untagged)]
+#[expect(dead_code, reason = "This is used in the generated OpenAPI spec")]
```
Alfred-Mountfield, line 106: How was this used before? Do we still need the `derive`s?

Alfred-Mountfield: Ah I see it was used in `remove` on `Vertices`

packages/graph/hash_graph/tests/integration/postgres/mod.rs
```diff
@@ -414,11 +392,11 @@ impl DatabaseApi<'_> {
...
             .into_iter()
-            .filter_map(|edition_id| subgraph.vertices.remove(&edition_id))
-            .map(|vertex| {
-                let Vertex::KnowledgeGraph(vertex) = vertex else { unreachable!() };
-                let KnowledgeGraphVertex::Entity(persisted_entity) = *vertex;
-                persisted_entity
+            .filter_map(|edition_id| match edition_id {
+                GraphElementEditionId::Ontology(_) => None,
+                GraphElementEditionId::KnowledgeGraph(edition_id) => {
+                    subgraph.vertices.entities.remove(&edition_id)
```
thehabbos007, line 398: Just a question about the removal here. I assume it mutates the subgraph, is this intentional? I see we did it before too

Alfred-Mountfield (commented):
Few questions, and I'm fine with merging this for the benefits we're seeing right now.

Generally I'm not convinced this is a great idea, I appreciate at the moment we don't consume/traverse/use a `Subgraph` in Rust, but I think that will change a lot in the future. But without being confident about what that future looks like, I think the gains we're getting now are perhaps worth it, I just also see us potentially undoing this. I think being able to get an element consistently by just passing in an ID is nicer than needing to match on the type of identifier, and then branch and look in the correct map (which could be 3 different maps for `BaseUri`) to find the thing.

thehabbos007 (approved):
Looks good to me, Tim. I have a single question, but not about something you changes

packages/graph/hash_graph/tests/integration/postgres/mod.rs
Alfred-Mountfield: Just to highlight, this is in the integration tests, and the return type of this function is `Entity`, so the `Subgraph` is dropped

thehabbos007: Oh of course, my bad!

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/vertices/vertex.rs
TimDiekmann: Cippy will only complain about these kinds of things if the struct is not exposed. It, however, cannot know, that we are exposing that via the OpenAPI specs.

PR #1678, opened 2022-12-15
Set up OpenTelemetry traces in Graph

thehabbos007:

This PR starts the work of introducing distributed tracing across the whole application.
This adds correlated tracing capabilities to the graph layer and allows exporting traces locally to Jaeger (added to external-services).

Unfortunately OTEL requires `protoc` to be available to build the project, which means for now we need `protoc` to compile the graph project. **This has already made redundant** on the main branch of OTEL, but we cannot add it as a git dependency because of compilation tricks required to bundle the proto typegen files.

I've added the Jaeger OTLP collecter/exporter/UI for us to use locally, but we can put in a Datadog collector in the future that will use the exact same protocol to ingest traces.

Note that we currently have "high cardinality" names for the `otel.name` but this is expected to change with #1684 and #1685 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203500651027063/f) _(internal)_

## 🔍 What does this change?

- OTLP tracing subscriber that published to Jaeger
- Adds new instrumentation attributes around the code to have spans for different calls
- `protoc` in dockerfile building the graph

- We may need to change docs to make it clear `protoc` is a dependency to build the project

## ⚠️ Known issues

- As pointed out, we need `protoc` to use the gRPC interface of the OTLP collectors. We could go with simple HTTP, but that would be less efficient than gRPC.

## ❓ How to test this?

1.  Checkout the branch
1.  Start the app through `yarn external-services up --build` and
1.  Open the Jaeger local UI at http://localhost:16686
1.  Explore traces created by initializing the system initially, run HTTP manual tests to generate more traces

## 📹 Demo

image

image

image


packages/graph/hash_graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -181,11 +182,12 @@ async fn create_entity_type<P: StorePool + Send>(
     tag = "EntityType",
     responses(
         (status = 200, content_type = "application/json", body = Subgraph, description = "A subgraph rooted at entity types that satisfy the given query, each resolved to the requested depth."),
-
+
```
TimDiekmann, line 185:
```suggestion

```

packages/graph/hash_graph/bin/hash_graph/src/main.rs
```diff
@@ -245,7 +246,7 @@ async fn main() -> Result<(), GraphError> {
 
     tracing::info!("Listening on {api_address}");
     axum::Server::bind(&addr)
-        .serve(rest_router.into_make_service())
+        .serve(rest_router.into_make_service_with_connect_info::<SocketAddr>())
```
thehabbos007, line 249: This is done for us to be able to supply `client_ip` to the top-level traces

packages/graph/hash_graph/lib/graph/src/api/rest/data_type.rs
```diff
@@ -96,7 +96,7 @@ impl RoutedResource for DataTypeResource {
     }
 }
 
-#[derive(Serialize, Deserialize, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
```
thehabbos007, line 99: I've tried to only have the requests pass their (potentially long) arguments. This means that we need debug on the Rest body structs for them to be put into the instrumented traces

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -74,3 +83,120 @@ where
...
+
+fn parse_x_forwarded_for(headers: &http::HeaderMap) -> Option<Cow<'_, str>> {
+    let value = headers.get("x-forwarded-for")?;
+    let value = value.to_str().ok()?;
+    let mut ips = value.split(',');
+    Some(ips.next()?.trim().into())
+}
+
+// Based on https://github.com/tokio-rs/axum/pull/769
+pub fn span_maker(request: &hyper::Request<hyper::Body>) -> tracing::Span {
```
thehabbos007, line 132: We can also specify `on_failure` and `on_response` to populate error codes and server faults

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -427,6 +428,7 @@ impl<C: AsClient> PostgresStore<C> {
 
 #[async_trait]
 impl<C: AsClient> EntityStore for PostgresStore<C> {
+    #[tracing::instrument(level = "info", skip(self, properties))]
```
thehabbos007, line 431: Please do tell if you see an `instrument` attribute that doesn't skip potentially large/expensive payloads. These should be put into the request trace

packages/hash/external-services/docker-compose.yml
```diff
@@ -75,13 +75,39 @@ services:
     environment:
       HASH_GRAPH_PG_MIGRATION_URL: "postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${HASH_GRAPH_PG_DEV_DATABASE}"
 
+  telemetry-collector:
```
thehabbos007, line 78: we can add this to the `hash_graph` docker compose too if we want it in isolation. I found it would be most useful for the external services one since errors would be more friendly to find for other folks

packages/graph/hash_graph/lib/graph/src/logging/init.rs
```diff
@@ -46,6 +58,41 @@ where
...
+    // Allow correlating trace IDs
+    global::set_text_map_propagator(TraceContextPropagator::new());
+    // If we need to set any tokens in the header for the tracing collector, this would be the place
+    // we do so.
+    let map = MetadataMap::new();
+
+    let pipeline = opentelemetry_otlp::new_exporter()
+        .tonic()
+        .with_endpoint(otlp_endpoint)
+        .with_timeout(std::time::Duration::from_secs(3))
```
TimDiekmann, line 73: Do we want to make this a constant?

packages/graph/hash_graph/lib/graph/src/logging/init.rs
```diff
@@ -46,6 +58,41 @@ where
...
+        .with_metadata(map);
+
+    // This can be changed to ratio-based sampling
+    let trace_config = opentelemetry::sdk::trace::config()
+        .with_sampler(Sampler::AlwaysOn)
+        .with_id_generator(RandomIdGenerator::default())
+        .with_max_events_per_span(64)
+        .with_max_attributes_per_span(16)
+        .with_max_events_per_span(16)
```
TimDiekmann, line 82: Same as above, does it makes sense to make these constants?

packages/hash/external-services/docker-compose.yml
```diff
@@ -75,13 +75,39 @@ services:
...
+    healthcheck:
+      # Port 14269 is the Jaeger admin endpoint
+      test:
+        [
+          "CMD-SHELL",
+          "wget --no-verbose --tries=1 --spider http://localhost:14269 || exit 1",
+        ]
+      interval: 2s
+      timeout: 2s
+      retries: 10
```
TimDiekmann, line 92: Are these the default values?

TimDiekmann (approved):
This is fantastic, thanks @thehabbos007! 🎉 

I have a few suggestions but don't feel strongly about them.

packages/graph/hash_graph/lib/graph/Cargo.toml
```diff
@@ -25,14 +25,18 @@ tower-http = { version = "0.3.5", features = ["trace"] }
 tracing = "0.1.37"
 tracing-appender = "0.2.2"
 tracing-error = "0.2.0"
+tracing-opentelemetry = "0.18.0"
+tonic = "0.8.3"
+opentelemetry = { version = "0.18.0", features = ["rt-tokio"]}
```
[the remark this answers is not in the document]
TimDiekmann: Answered offline, this wouldn't solve the issue as the artifacts have not been built. We have to wait for a new release.

packages/graph/hash_graph/lib/graph/src/logging/init.rs
thehabbos007: I don't see a reason to make this configurable right now at least. It would be the timeout for the gRPC request I believe. Maybe bumping it to 5 seconds is fine, but I don't think we need to add an arg to configure this

packages/graph/hash_graph/lib/graph/src/logging/init.rs
thehabbos007: These we may want to configure, yes. I think we should do that in the future when we know what kind of limits we want to set. I think it's fine to leave for later (we can always add the clap args + env variables later)

packages/hash/external-services/docker-compose.yml
thehabbos007: They are not, I chose them to be similar to what we have for graph. It's a Go service, so it should boot quickly. Did you run into timeouts?

packages/graph/hash_graph/lib/graph/src/logging/init.rs
TimDiekmann: I just mean a constant at the top of the file
```
const OPENTELEMETRY_TIMEOUT_DURATION: Duration = Duration::from_secs(3);
```

thehabbos007: Ah gotchu, sure why not

packages/hash/external-services/docker-compose.yml
TimDiekmann: No, I was just wondering 🙂

packages/graph/hash_graph/lib/graph/src/logging/init.rs
TimDiekmann: Having this configurable somehow would be good, but having a constant for now is probably good enough for now 🙂

packages/graph/hash_graph/lib/graph/src/logging/init.rs
thehabbos007: 621932df662fcaf5a1e676b200724eb1588ba86a changed to 5 to give a little more room for any bigger trace batches

packages/graph/hash_graph/lib/graph/src/logging/init.rs
TimDiekmann: The configuration seems good enough for us for now. I'd say we remove the configuration here and rely on their environment variables.

thehabbos007: image
Did a one-off test. Seems to be fine :+1:

TimDiekmann (approved):

TimDiekmann (approved):

PR #1683, opened 2022-12-15
Deduplicate dependency resolution - part 2: Generalizes the subgraph insertions

TimDiekmann:

The next step towards deduplicating of dependency resolution. It's required, that the insertion of records is aware of the `Record` type being inserted, so the deduplicated resolution can just call those function.

## 🔗 Related links

## 🚫 Blocked by

- #1677 
- #1679 
- #1680 
- #1682 

## 🔍 What does this change?

- Add method to `Record` to allow inserting of themself to the provided subgraph
- Added a similar method, but add it as a root as well
- Use the above functions to simplify dependency resolution logic
- Drive-by: Simplify some trait bounds

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/mod.rs
@@ -21,7 +21,9 @@ pub use self::{
     postgres::{AsClient, PostgresStore, PostgresStorePool},
 };
 use crate::{
-    identifier::{account::AccountId, knowledge::EntityId, DecisionTimestamp},
+    identifier::{
+        account::AccountId, knowledge::EntityId, DecisionTimestamp, GraphElementEditionId,
+    },
     knowledge::{Entity, EntityLinkOrder, EntityMetadata, EntityProperties, EntityUuid, LinkData},
     ontology::{
         DataTypeWithMetadata, EntityTypeWithMetadata, OntologyElementMetadata,
@@ -35,8 +37,8 @@ use crate::{
 /// A record stored in the [`store`].
 ///
 /// [`store`]: crate::store
-pub trait Record: Sized {
-    type EditionId: Clone + PartialEq + Eq + Hash + Send + Sync;
+pub trait Record: Sized + Send {
+    type EditionId: Clone + PartialEq + Eq + Hash + Send + Sync + Into<GraphElementEditionId>;
     type QueryPath<'p>: QueryPath + Send + Sync;
 
     fn edition_id(&self) -> &Self::EditionId;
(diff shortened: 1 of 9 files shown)


packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -47,6 +49,21 @@ pub trait Record: Sized {
         subgraph: &'s mut Subgraph,
         edition_id: &Self::EditionId,
     ) -> RawEntryMut<'s, Self::EditionId, Self, RandomState>;
+
+    fn insert_into_subgraph(self, subgraph: &mut Subgraph) -> &Self {
```
Alfred-Mountfield, line 53: Do you feel strongly that having it defined in the `Record` (where I know it benefits from knowing the type of `Self` in the impl) is that much cleaner?

It's really very different to the normal ways of doing things where we define `insert` on the collection.

Alfred-Mountfield: And honestly I just don't like it much 🙃

TimDiekmann: I'd prefer to have it on the collection, but this is not possible without introducing a trait like `SubgraphIndex<R: Record>`, where you define the access for `R`. This would work and is probably cleaner, but this also implies, that we need a bound on `SubgraphIndex` each time. I plan on looking into that at some later point again, defining it on `Record` is the easiest solution for now.

Really hoped to give you a more satisfying answer 🙁

Alfred-Mountfield: I think this aligns with one of my comments on the previous pull-request though, having everything in separate `HashMap`s and needing to branch to pick the correct place to insert them is generally a lot less ergonomic than just having the set be `HashMap<Identifier, Vertex>`.

I know you tend to avoid enumerations where you can, but it seems like we're ending up with even stranger 'clever' codepaths which aren't idiomatic, or straightforward to maintain.

We've already sunk significant investment into this so I'll approve because I don't know how much more we should put into refactoring this, but just flagging for the future.

Alfred-Mountfield (approved):

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: The issue with the enumeration would be almost the same as the insertion depends on `R`. I tried that approach before I refactored the subgraph, but the main motivation to change that is that we need to get `R` back out of the function to further use the record. With that, the implementation would be more complicated and we would need to implement the function on each record instead as we wouldn't get back `R`, but an enum variant with `R` in it.

Alfred-Mountfield: Gotcha, probably worth iterating on it some more at some point in the future

Alfred-Mountfield (approved):

PR #1689, opened 2022-12-16
Split store module into submodules

TimDiekmann:

Currently, the `store` `mod.rs` file is a very long file, which grew quite large over the last months. It's easier to read and find the traits defining a `Store` by looking at the sub modules instead.

The main motivation for this PR is that I expect to add subgraph related methods to the `Store` trait, which apply to all of stores.

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1203363157432084/1203444301722127/f) _(internal)_

## 🔍 What does this change?

**Without changing any implementation**, this breaks the `store` module into submodules.


packages/graph/hash_graph/lib/graph/src/lib.rs
```diff
@@ -65,6 +65,8 @@
     reason = "This encourages importing `as` which breaks IDEs"
 )]
 
+extern crate core;
+
```
TimDiekmann, line 69: CLion strikes again

```suggestion
```

Alfred-Mountfield (approved):

PR #1691, opened 2022-12-16
Define methods on `Subgraph` instead of `Record`

TimDiekmann:

As mentioned in a review earlier I was about to revisit where to define functions to modify the subgraph. I was able to come up with a solution, where the functions can actually be implemented on the `Subgraph` directly.

## 🔗 Related links

- https://github.com/hashintel/hash/pull/1683#discussion_r1050812094


packages/graph/hash_graph/lib/graph/src/shared/identifier/knowledge.rs
```diff
@@ -160,6 +164,15 @@ pub struct EntityEditionId {
     version: EntityVersion,
 }
 
+impl SubgraphIndex<Entity> for EntityEditionId {
+    fn subgraph_entry<'r>(
```
Alfred-Mountfield, line 168: What's the intended meaning behind `'r` by the way?

packages/graph/hash_graph/lib/graph/src/shared/subgraph/mod.rs
```diff
@@ -29,4 +38,56 @@ impl Subgraph {
...
+                            .read_one(&R::create_filter_for_edition_id(edition_id))
+                            .await?,
+                    )
+                    .1
+            }
+        })
+    }
+}
+
+pub trait SubgraphIndex<R: Record>: Clone + Eq + Hash + Into<GraphElementEditionId> {
```
Alfred-Mountfield, line 88: Can we have some documentation for this please

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -55,7 +55,8 @@ impl<C: AsClient> PostgresStore<C> {
 
             let entity = match dependency_status {
                 DependencyStatus::Unresolved => {
-                    <Self as Read<Entity>>::read_into_subgraph(self, subgraph, &entity_edition_id)
+                    subgraph
+                        .get_or_read::<Entity>(self, &entity_edition_id)
```
Alfred-Mountfield, line 59: Are there actually situations where this branch _should_ be hit? I'm wondering if we just don't want to allow the `or_read`

Alfred-Mountfield (commented):
Thanks for looking further into this, I knew we could do much better 🎉 

Few clarifications otherwise I think I'm happy to approve this

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: For ontology types, we may not have inserted an element before traversing it. I expect this to be changed at some point in the near future. For entities you are right, and I already adjusted this in a follow-up PR: https://github.com/hashintel/hash/commit/275f8a87c90d3494afecf868eb33df2588e4a1e3

Alfred-Mountfield: Great, cheers

packages/graph/hash_graph/lib/graph/src/shared/identifier/knowledge.rs
TimDiekmann: I have used `'r` as it belongs to the `Record`

Alfred-Mountfield: Isn't `'r` constrained by the lifetime of the reference to the `s`ubgraph though?

TimDiekmann: Yes, it's bound by the `Subgraph`, but the returned entry is a `Record`. I don't feel strongly either way, to avoid confusion we could just go with `'a`.

Alfred-Mountfield: I think let's go with `'a` here and similar places then.

TimDiekmann: Changed the lifetime parameter to `'a` in 15d717be9033590d7c7fe5936f0cf19075b4887d

packages/graph/hash_graph/lib/graph/src/shared/subgraph/mod.rs
TimDiekmann: Added a docstring to the trait and the function in 153f95a3bec949cda0ba9971c6179c4e8e08300c. I think the most important thing in the documentation is to state, that the operation will be performed on the vertices depending on `R`.

packages/graph/hash_graph/lib/graph/src/shared/subgraph/mod.rs
```diff
@@ -29,4 +38,60 @@ impl Subgraph {
...
+                    )
+                    .1
+            }
+        })
+    }
+}
+
+/// Used for index operations on a mutable [`Subgraph`].
+///
+/// Depending on `R`, the index operation will be performed on the vertices of the subgraph.
```
Alfred-Mountfield, line 90:
```suggestion
/// Depending on `R`, the index operation will be performed on the respective vertices collection of the subgraph.
```

Maybe something like that? I think the main thing we're trying to convey here is that this will go to `vertices.property_type` or `vertices.entity`, etc.

Also can this be used for Edges, or not? If it can't then maybe we should call this `SubgraphVertexIndex`

TimDiekmann: We probably can add a new function to access the edges as well but currently, we don't require that. This would require a few adjustments to the `Record` trait (adding an associated type on the edge kind) and the `Subgraph` structure (similar to `Vertices` to directly allow accessing the different ontology types), but I don't see a reason to not add the function as soon as we need generic access.
Maybe this?
```rust
/// Used for index operations on a mutable [`Subgraph`].
///
/// Depending on `R`, the index operation will be performed on the respective collection of the
/// subgraph.
pub trait SubgraphIndex<R: Record>: Clone + Eq + Hash + Into<GraphElementEditionId> {
    /// Returns a mutable reference to the [`Record`] vertex in the subgraph.
    fn subgraph_vertex_entry<'a>(
        &self,
        subgraph: &'a mut Subgraph,
    ) -> RawEntryMut<'a, R::EditionId, R, RandomState>;
}
```

Alfred-Mountfield: Looks good to me

TimDiekmann: Alright, changed it in 95d36c80bd790225412aaca6705703faf34f4c53

Alfred-Mountfield (approved):

PR #1692, opened 2022-12-17
Read the earliest edge timestamp on serialization

TimDiekmann:

Instead of querying the database for the earliest version, this can be read when serializing the subgraph. This roughly increases the performance by ~55 % when resolving links.

## 🔍 What does this change?

- Generate the subgraph without querying the earliest link timestamp
- When serializing the subgraph read the earliest link timestamp from the vertex set

## ⚠️ Known issues

I'm not sure if we actually *need* the earliest timestamp of a knowledge edge because, as demonstrated, this is not required to be present in the subgraph.

## 📹 Demo

Reading 50 entities with a depth of 1, every entity linked to each other entity. Red is old, blue is new.
image


packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/mod.rs
```diff
@@ -1 +1,41 @@
...
+
+// WARNING: This MUST be kept up to date with the struct names and serde attributes
+//   Necessary because Timestamp doesn't implement ToSchema
+impl ToSchema for EntityIdAndTimestamp {
+    fn schema() -> Schema {
+        ObjectBuilder::new()
+            .property(
+                "baseId",
+                // Apparently OpenAPI doesn't support const values, the best you can do is
+                // an enum with one option
```
Alfred-Mountfield, line 26: Appreciate this might have been copied but I don't think this makes sense here?

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -35,8 +56,8 @@ pub struct Edges {
     pub knowledge_graph: KnowledgeGraphRootedEdges,
 }
 
-impl From<crate::subgraph::edges::Edges> for Edges {
-    fn from(edges: crate::subgraph::edges::Edges) -> Self {
+impl Edges {
+    pub fn from_store_subgraph(edges: crate::subgraph::edges::Edges, vertices: &Vertices) -> Self {
```
Alfred-Mountfield, line 60: Naming of this function seems a bit weird if you don't actually pass in a `Subgraph`

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -56,7 +77,42 @@ impl From<crate::subgraph::edges::Edges> for Edges {
...
+                            crate::subgraph::edges::KnowledgeGraphOutwardEdges::ToOntology(edge) => {
+                                KnowledgeGraphOutwardEdges::ToOntology(edge)
+                            }
+                            crate::subgraph::edges::KnowledgeGraphOutwardEdges::ToKnowledgeGraph(
+                                edge,
+                            ) => {
+                                let earliest_timestamp = if edge.reversed {
+                                    // We want to log the time the link entity was *first* added
+                                    // from this entity. We therefore need to find the timestamp of
+                                    // the first link entity
```
Alfred-Mountfield, line 91: This is no longer true. I think we need to give some reasoning about deduplicating information of edges, where if we have N editions of an entity and M editions of a link with consistent endpoints, we don't want to store N*M edges when 1 is enough.

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -56,7 +77,42 @@ impl From<crate::subgraph::edges::Edges> for Edges {
...
+                                    vertices.find_earliest_entity(&edge.right_endpoint)
```
Alfred-Mountfield, line 92: If we stored the vertices in a btree-map we wouldn't need to do a search. In bad cases this is also pretty inefficient, it's two loops (although I appreciate the inner loop is only over the editions of that entity)

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -56,7 +77,42 @@ impl From<crate::subgraph::edges::Edges> for Edges {
...
-                    let edges = edges.into_iter().collect();
+                    let edges = edges.into_iter().map(|edge| {
+                        match edge {
+                            crate::subgraph::edges::KnowledgeGraphOutwardEdges::ToOntology(edge) => {
+                                KnowledgeGraphOutwardEdges::ToOntology(edge)
+                            }
+                            crate::subgraph::edges::KnowledgeGraphOutwardEdges::ToKnowledgeGraph(
+                                edge,
+                            ) => {
+                                let earliest_timestamp = if edge.reversed {
```
Alfred-Mountfield, line 88: I was struggling to wrap my head around it as I wasn't sure this branch logic is correct. I see now (at least I think) it is. Can you add a comment explaining something like this

We always need the earliest version of the _link_ entity, and there are four permutations
* hasLeftEntity (link to left)
* hasLeftEntity reversed  (left to link)
* hasRightEntity (link to right)
* hasRightEntity reversed (right to link)

If it's reversed the link is on the right endpoint, otherwise it's on the left

packages/graph/hash_graph/lib/graph/src/shared/subgraph/edges/edge.rs
```diff
@@ -63,18 +63,6 @@ impl ToSchema for OntologyOutwardEdges {
 #[derive(Debug, Hash, PartialEq, Eq, Serialize)]
 #[serde(untagged)]
 pub enum KnowledgeGraphOutwardEdges {
-    ToKnowledgeGraph(OutwardEdge<KnowledgeGraphEdgeKind, EntityIdAndTimestamp>),
+    ToKnowledgeGraph(OutwardEdge<KnowledgeGraphEdgeKind, EntityId>),
     ToOntology(OutwardEdge<SharedEdgeKind, OntologyTypeEditionId>),
 }
```
Alfred-Mountfield, line 68: I appreciate the optimisation we're getting from this change but the separation between the representations is increasing more and more and I really don't like what we're doing for maintainability of an already very complicated object.

(I'm not suggesting we shouldn't do this by the way, just I want us to be aware of it)

Alfred-Mountfield (commented):
Thanks Tim, this is a great change, we knew we needed to get rid of those lookups and this is a pretty clean way to do it.

That said, I'm getting quite concerned about maintainability, perhaps this is something we can try and mitigate in follow-ups but I'm not super comfortable with how the objects are deviating from one another. One of the things I dislike most is that if you look at the "datastore's" subgraph, you could accidentally think you have all the information required about the edges, but then when you look at a specific serialization of it.. you don't.

If we wanted to implement another serialization route, I could see us missing this post-processing step (for example), because it's pretty isolated in one step at the end of the process. Furthermore, if we actually do start interacting with the subgraph in Rust (which I am pretty confident we will), then we do not have the appropriate information right now to actually use it. (This is also why I disliked us getting rid of `Deserialize` because it's easy for us to make the mapping only work one-way easily).

I think if we move to `BTreeMap`s then we can have the earliest thing be done as part of querying the object, rather than a post-processing step, as it should be quick enough. We could then maybe use that same logic in an `Into` implementation, which could be consistent across representations (and used in the JSON representation), rather than the way we're doing it now which would require us to reimplement it if we serialized to flatbuffers for example.

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/mod.rs
TimDiekmann: I worked around this: fadffa6a6e351a054ecea93d4e989e74dd785a73

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: Yes, the name is not optimal, I renamed it b89f701f30178cf0dada4ff47c8b497489bf0c81
I'm still not super satisfied with the name, but as mentioned in the other comments, this will change anyway.

packages/graph/hash_graph/lib/graph/src/shared/subgraph/edges/edge.rs
TimDiekmann: I think it makes sense to use the `crate::subgraph::Subgraph` struct as the structure to build the subgraph, not as the subgraph itself. Maybe we define that in the `DependenyContext` (or `TraversalContext` in the future) and allow creating the `Subgraph` as defined in the `uitoipa_typedefs` module as the "real" subgraph. Then we would have the same subgraph available everywhere. 
If you are happy with this approach I will do that in a follow-up.

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: I changed the comment in 14938b548c5dea70a4e51bcd8f3a94cc3e00b5c8

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: That's a good point, I changed the representation to use a `BTreeMap` instead: 14938b548c5dea70a4e51bcd8f3a94cc3e00b5c8

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: Yes, it's a little bit too hard to get the branch logic. I added a better description to this location: 14938b548c5dea70a4e51bcd8f3a94cc3e00b5c8

TimDiekmann:
> That said, I'm getting quite concerned about maintainability, perhaps this is something we can try and mitigate in follow-ups but I'm not super comfortable with how the objects are deviating from one another. One of the things I dislike most is that if you look at the "datastore's" subgraph, you could accidentally think you have all the information required about the edges, but then when you look at a specific serialization of it.. you don't.
> 
> If we wanted to implement another serialization route, I could see us missing this post-processing step (for example), because it's pretty isolated in one step at the end of the process. Furthermore, if we actually do start interacting with the subgraph in Rust (which I am pretty confident we will), then we do not have the appropriate information right now to actually use it. (This is also why I disliked us getting rid of `Deserialize` because it's easy for us to make the mapping only work one-way easily).

I agree, In order to solve this we shouldn't expose the "Rust representation" at all and only use it as an implementation detail to build the subgraph. I proposed a solution in https://github.com/hashintel/hash/pull/1692#discussion_r1051598569


> I think if we move to `BTreeMap`s then we can have the earliest thing be done as part of querying the native object (e.g. `getEdgesForElement` or something), rather than a post-processing step, as it should be quick enough. We could then maybe use that same logic in an `Into` implementation, which could be consistent across representations (and used in the JSON representation), rather than the way we're doing it now which would require us to reimplement it if we serialized to flatbuffers for example.

I changed the implementation to use a `BTreeMap` in 14938b548c5dea70a4e51bcd8f3a94cc3e00b5c8.

TimDiekmann:
Thank you very much for reviewing this @Alfred-Mountfield, especially given that you are on vacation 🎉 
I addressed your review.

packages/graph/hash_graph/lib/graph/src/shared/subgraph/edges/edge.rs
Alfred-Mountfield: I think that would make sense except for what I spoke about before:
> Furthermore, if we actually do start interacting with the subgraph in Rust (which I am pretty confident we will), then we do not have the appropriate information right now to actually use it. (This is also why I disliked us getting rid of Deserialize because it's easy for us to make the mapping only work one-way easily).

I don't think we'd want to interface with the utoipa-polluted representation, I think we'd want to use the cleaner more native-friendly representation.

That's why I think we should go for this:
> I think if we move to BTreeMaps then we can have the earliest thing be done as part of querying the native object (e.g. getEdgesForElement or something), rather than a post-processing step, as it should be quick enough. We could then maybe use that same logic in an `Into` implementation, which could be consistent across representations (and used in the JSON representation), rather than the way we're doing it now which would require us to reimplement it if we serialized to flatbuffers for example.

So, instead of calling `earliest_entity_by_id` in the `Serialize` implementation, I think we should do it as part of an `IntoIterator` for the native `Edges` or something and `FromIterator`. Then we'd benefit from it on both sides.

packages/graph/hash_graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -56,13 +81,68 @@ impl From<crate::subgraph::edges::Edges> for Edges {
...
+                                // edge:
+                                //   1. `HAS_LEFT_ENTITY`, `reversed = false`
+                                //   2. `HAS_RIGHT_ENTITY`, `reversed = false`
+                                //   3. `HAS_LEFT_ENTITY`, `reversed = true`
+                                //   4. `HAS_RIGHT_ENTITY`, `reversed = true`
+                                //
+                                // For 1. and 2., the entity is a link and we want to store the
+                                // earliest version of this link as the earliest timestamp.
+                                // For 3. and 4., the endpoint of the edge is a link and we want to
+                                // store the earliest of that endpoint as the earliest timestamp.
```
thehabbos007, line 111: Is it correct that we are using the `right_endpoint` for both the `LEFT` and `RIGHT` permutation of `edge.reversed = true`? Or should we use a different endpoint for them?

TimDiekmann: Yes, that is correct. The "right_endpoint" refers to the edge itself, not the edge type.

PR #1693, opened 2022-12-18
`error-stack`: move `fmt::HookContext` into `hook::HookContext`

indietyp:

This is in preparation for #1558 and generalizes the `HookContext` (notably storage) into its own (isolated) type `HookContext<T, U>`, which is generic over two parameters: `T`, which are the extra values that might be needed nd `U`, the current type accessed and used for storage partitioning.

Why is this needed? `serde` hooks will also need a way to use and store context-sensitive information. I found a lot of code duplication when trying to mirror the storage API, so I chose the consolidation into a single struct. This is purely an internal change and does not change anything for library consumers.

## 👂 Open Questions

~~Because cargo-doc considers `hook::context::HookContext` to be private, we currently do not generate any docs for the new inner type. We would need a way to expose those.~~

Using `cfg(doc)` we now expose the struct, which is a bit hacky but enables documentation of all functions.

No, purely internal change.

## 🛡 What tests cover this?

Existing tests.


packages/libs/error-stack/src/hook/context.rs
```diff
@@ -0,0 +1,366 @@
...
+    ///
+    /// <pre>
+    #[doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/tests/snapshots/doc/fmt__hookcontext_cast.snap"))]
+    /// </pre>
+    #[must_use]
+    pub fn cast<U>(&mut self) -> &mut HookContext<Extra, U> {
+        // SAFETY: `HookContext` is marked as repr(transparent) and the changed generic is only used
+        // inside of the `PhantomData`
+        unsafe { &mut *(self as *mut Self).cast::<HookContext<Extra, U>>() }
+    }
```
TimDiekmann, line 174: I take the opportunity to ask this question as we touched this code: Do you see a way to remove the `unsafe` here? I know, why we have, I'm just wondering if there is a way to avoid it (this question is unrelated to this PR).

TimDiekmann (commented):
✅ Looks good, happy to approve when CI is green 🎉

packages/libs/error-stack/src/hook/context.rs
indietyp: I don't think so, if we would instead transmute `mut self` instead we could use RFC 2528 may be of help.

TimDiekmann (approved):

PR #1697, opened 2022-12-19
Enable Kratos SMTP with AWS SES

thehabbos007:

This PR adds an email transport for Kratos when we deploy to production on AWS using AWS SES.
The changes that were required was adding some variables to the terraform and stitching together an SMTP user to send emails through SES' verified identities.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203509372806153/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432081/1203475805898260/f) _(internal)_

## ❓ How to test this?

Try signing up with your email at the hosted instance


packages/hash/external-services/kratos/kratos.prod.yml
```diff
@@ -23,18 +23,17 @@ selfservice:
...
-    #   config:
-    #     # and make sure to enable the code method.
-    #     enabled: true
+    link:
+      config:
+        # The URL for verification emails are set through the link method
+        # but we're using the code method, so we disable this method for usage.
+        enabled: false
+        # Set through SELFSERVICE_METHODS_LINK_CONFIG_BASE_URL
+        base_url: http://localhost:3000/api/ory
```
benwerner01, line 32: Can we remove this if we're disabling this method?
```suggestion
```

I'm also a bit unsure why we would have `http://localhost:3000` in the production file in any case.

packages/hash/external-services/kratos/kratos.prod.yml
```diff
@@ -73,12 +72,12 @@ selfservice:
...
-    #   lifespan: 48h
-    #   # Set though SELFSERVICE_FLOWS_VERIFICATION_UI_URL
-    #   ui_url: http://localhost:3000/verification
-    #   enabled: true
+    verification:
+      use: code
+      lifespan: 48h
+      # Set though SELFSERVICE_FLOWS_VERIFICATION_UI_URL
+      ui_url: http://localhost:3000/verification
+      enabled: true
```
benwerner01, line 80: How can this be `http://localhost:3000` in the production kratos config? Wouldn't it have to be the public frontend URL (so `https://alpha.hash.ai`)?

packages/hash/external-services/kratos/kratos.prod.yml
thehabbos007: Because of some strageness in the way the email templates are configured, the base url of the link method is used for the code flow emails

packages/hash/external-services/kratos/kratos.prod.yml
thehabbos007: We're overriding all these through env vars at this diff: https://github.com/hashintel/hash/pull/1697/files#diff-a20ffb69d9d206b0d1749faf0e84f23ae7a85e38d016c8e4692946ffca6f36de
I understand it's very unclear, but I did not like the idea of configuration for production not being centralized in one place

benwerner01 (approved):
Thanks Ahmad, these changes look good to me 👍

PR #1705, opened 2022-12-20
Rename `/[account-slug]/` directory to `/[shortname]` in FE

benwerner01:

This PR renames the `/[account-slug]/` directory to `/[shortname]` in the frontend for the naming to be more consistent for recent changes to the frontend routes.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432087/1203488186089105/f) _(internal)_

## 🔍 What does this change?

See commits.

No that I'm aware of.

## 🛡 What tests cover this?

Manual tests

## ❓ How to test this?

1. Run the HASH App locally and check that routing works as expected


kachkaev (approved):
As a next step, we can see if we can read the value into some context provider and use it via `useSomeHook` rather than via `router.query.shortname`. This can make our codebase a bit more robust.

PR #1710, opened 2022-12-21
Globally specify common components for OpenAPI specs

TimDiekmann:

A lot of components used in the REST API are used commonly in multiple places. Currently, we specify them all on each file, which is unneeded and really annoying to deal with as IDEs don't properly provide auto-imports for those as utoipa has a custom proc macro syntax. This small PR moves the components to a common place, so the resulting specs do not change.


Alfred-Mountfield (approved):
Wanted to do this for a while but never found time, thank you!

PR #1715, opened 2022-12-21
Replace time-related structs with more generic ones

TimDiekmann:

In total, we need (at least) three different representations for time-related structures to be exposed to the backend, optimally, we need four:
- Simple timestamps: A single point in time
- Timespans for versions: spans in versions are guaranteed to be left-closed right-opened intervals or left-closed right-unbounded intervals, resulting in this representation
    ```ts
    type VersionTimespan = {
        start: string,
        end?: string,
    }
    ```
- Timespans with arbitrary bounds: To specify a time projection (see [Adjust structural queries for temporal versioning](https://app.asana.com/0/1203363157432087/1203491211535116/f) _(internal)_), we have to specify a timespan for the projected time axis. This cannot be left-closed right-opened for a few reasons:
    - `[start, end)` bound cannot specify a single timestamp as `end` is not included (`[x, x)` is an empty interval)
    - we want to support unbounded timespans, e.g. `[start, ∞)`, `[∞, end)`, or even `(∞, ∞)`
    - we want to support specifying the "now" time by passing `null`, so we cannot interpret `[start, null)` as `[start, ∞)`
- A Timespan similar to the above for the resolved time projection, i.e. replacing `null` with `now()`. The latter is not strictly needed but simplifies handling on the TS backend

To expose the correct `EntityVersion` has to be implemented, the first two bullets have to be implemented so this PR provides a more generic approach for `Timestamp` and `VersionTimespan`

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1201095311341924/1203331904553375/f) _(internal)_

## 🚫 Blocked by

- #1703 
- #1717

## 🔍 What does this change?

- Provide tags for `DecisionTime` and `TransactionTime` (they are enums as they will be used to specify the projections of the third and fourth bullet above)
- Re-implement `Timestamp`, same implementation as before but generic over the time axis
- Replace old timespans with `VersionTimespan` to correctly represent `EntityVersion`
- Expose `VersionTimespan` to the backend (not used yet)


packages/graph/hash_graph/lib/graph/src/shared/identifier/time/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+        TransactionTimeVersionTimespan, TransactionTimestamp,
+    },
+    timestamp::Timestamp,
+};
+
+fn parse_bound(
+    bound: &postgres_protocol::types::RangeBound<Option<&[u8]>>,
+) -> Result<Bound<Timestamp<()>>, Box<dyn Error + Send + Sync>> {
+    match bound {
+        postgres_protocol::types::RangeBound::Inclusive(Some(bytes)) => Ok(Bound::Included(
```
Alfred-Mountfield, line 20: Do we want to match the naming by the way? As in `Bound::Inclusive`

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/timestamp.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use chrono::{DateTime, Utc};
+use derivative::Derivative;
+use postgres_types::{private::BytesMut, FromSql, ToSql, Type};
+use serde::{Deserialize, Serialize};
+use utoipa::{openapi, ToSchema};
+
+/// Opaque structure to represent a single point in time.
+///
+/// The type parameter `A` is the time axis to distinguish between different time axis at compile
```
Alfred-Mountfield, line 12:
```suggestion
/// The type parameter `A` is the time axis to distinguish between different time axes at compile
```

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/timestamp.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct Timestamp<A> {
+    #[serde(skip)]
+    axis: PhantomData<A>,
+    time: DateTime<Utc>,
+}
+
+impl<A> fmt::Debug for Timestamp<A> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.debug_struct("timestamp")
+            .field("axis", &type_name::<A>())
```
Alfred-Mountfield, line 37: Does this stop us using `derivative(Debug..`?

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/axis.rs
```diff
@@ -0,0 +1,30 @@
...
+/// This is used as the generic argument to time-related structs and can be used as tag value.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum DecisionTime {
+    #[default]
+    Decision,
+}
+
+pub type DecisionTimestamp = Timestamp<DecisionTime>;
+pub type DecisionTimeVersionTimespan = VersionTimespan<DecisionTime>;
```
Alfred-Mountfield, line 17: Is `Version` superfluous in this maybe? I _do_ like us having quite explicit self-explanatory code but `DecisionTimeVersionTimespan` is realllyyy wordy

Alfred-Mountfield (commented):
Few questions/comments, generally looks good :)

packages/graph/hash_graph/lib/graph/src/api/rest/mod.rs
```diff
@@ -385,3 +386,15 @@ impl Modify for FilterSchemaAddon {
         }
     }
 }
+
+struct TimeSchemaAddon;
```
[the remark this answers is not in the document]
TimDiekmann: In an earlier version I needed this, mainly to abstract over `DecisionTimeVersionTimespan` and `TransactionTimeVersionTimespan` and just use `VersionTimespan`. The same applies to `Timestamp` and (later PR) `Timespan`. The current state does not require it, thus, I removed this as part of 6b10d3f0a472a31d58351b1ff23751901acf8f28

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/mod.rs
TimDiekmann: We neither define any of these variants. `Inclusive` is from `postgres_types`, while `Included` is from `std::ops`.

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/axis.rs
TimDiekmann: In https://github.com/hashintel/hash/pull/1738 I need a distinct `Timespan` struct, which contains different bounds. The version is guaranteed to always be `(Included, Excluded)` but for time projection defined in structural queries this does not apply. The different bounds are required as otherwise some timespans cannot be represented. With `(Included, Excluded)` a specific timestamp cannot be represented, without `(Included, Excluded)`, we cannot pass back `VersionTimespan` to the structural queries. Also `Unbounded` is required to select for a range until a certain timestamp or a range starting at a certain timestamp. Also, for getting all versions of an entity, `(Unbounded, Unbounded)` is required.
I elaborated a little bit more on the reasoning somewhere but I cannot find it currently 🙁

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/timestamp.rs
TimDiekmann: Yes. I mainly implemented it this way to allow `Debug(bound = "")` on structs, which are using `Timestamp`.

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/axis.rs
Alfred-Mountfield: I don't quite understand how that explains why we need the full length of the type name here, but we can revisit this as we tackle the stack of PRs.

Alfred-Mountfield (approved):

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/axis.rs
TimDiekmann: With `Timespan` added as well it's likely that we will also have a type alias for `DecisionTimespan` or something similar. However, It's also likely that we don't need any of these aliases later on. Yes, I will revisit this later.

PR #1717, opened 2022-12-21
Use unbounded time ranges in Postgres

TimDiekmann:

Postgres has the special value `'infinity'` which makes the handling in the Graph more complicated than needed. Instead, we can use unbounded ranges.


packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -884,10 +884,10 @@ impl PostgresStore<Transaction<'_>> {
                     entity_record_id,
                     tstzrange(
                         CASE WHEN decision_time IS NULL THEN now() ELSE decision_time END,
-                        'infinity',
+                        NULL,
                         '[)'
                     ),
-                    tstzrange(now(), 'infinity', '[)')
+                    tstzrange(now(), NULL, '[)')
```
thehabbos007, line 890: I assume you've looked into the semantics of `NULL` vs `'infinity'` for ranges.
Do you think we'd run into any problems with how the overlap is handled within the `gist` index we use? Does `NULL` behave the same as `'infinity'` for range queries?

thehabbos007 (approved):
Looks good to me, just a comment that asks for the range query semantics :)

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: The overlaps are handled almost the same. The only difference is, that `'infinity'` as an upper bound is slightly earlier than `NULL` and for a lower bound, `'infinity'` is slightly later than `NULL`, so only when comparing those two values, the semantic is different but as we don't use `'infinity'` anywhere anymore, this has no implications.

thehabbos007: Great. I can't imagine we'd be mixing up nulls and infinity, so that seems fine :)

PR #1718, opened 2022-12-21
Expose version timespans of entities to the backend

TimDiekmann:

In the Graph we switched to temporal versioning a few weeks ago but this was not exposed to the backend yet. With this PR the version timespans of entities are exposed and the backend was adjusted to use the timestamps to look-up the subgraph

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203331904553375/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432087/1203444301722133/f) _(internal)_

## 🚫  Blocked by

- #1703 
- #1717 
- #1715 

## 🔍 What does this change?

- Use derived `Serialize` and `ToSchema` for `EntityVersion`
- Adjust subgraph library to use the `transactionTime.from` to look up entities. This will slightly change in the future when we can specify time axis in structural queries, but for now this is the correct behavior. Actually, it wouldn't change the behavior if the `decisionTime` was used as it defaults to the `transactionTime` if not set and setting it is not possible yet.
- Use `decisionTime` to identify entities in other location in the code base


TimDiekmann:
As discussed offline I merged #1735 into this branch.

packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
```diff
@@ -552,149 +547,137 @@ mod tests {
     }
 
     #[test]
-    fn for_entity_by_edition_id() {
-        let entity_edition_id = EntityEditionId::new(
+    #[ignore = "TODO: Reevaluate if we need this after https://app.asana.com/0/0/1203491211535116/f"]
```
Alfred-Mountfield, line 550: Have you re-evaluated this yet in the next PRs?

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/create-entity-page.tsx
```diff
@@ -60,7 +60,7 @@ export const CreateEntityPage = ({ entityTypeId }: CreateEntityPageProps) => {
           throw new Error("subgraph not found");
         }
 
-        const draftEntityEditionId: EntityEditionId = {
+        const draftEntityEditionId: EntityVertexId = {
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/create-entity-page.tsx:63]: Can we rename this

packages/hash/subgraph/src/stdlib/roots.ts
```diff
@@ -28,17 +25,17 @@ import { getPropertyTypeByEditionId } from "./element/property-type";
...
+  subgraph.roots.map((rootVertexId) => {
     const root = mustBeDefined(
-      subgraph.vertices[rootEditionId.baseId]?.[
+      subgraph.vertices[rootVertexId.baseId]?.[
         // We could use type-guards here to convince TS that it's safe, but that would be slower, it's currently not
         // smart enough to realise this can produce a value of type `Vertex` as it struggles with discriminating
         // `EntityId` and `BaseUri`
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-        (rootEditionId as any).version
+        rootVertexId.version as any
```
Alfred-Mountfield, line 35: Swapping this around is interesting to me.. do you still need the `as any`?

packages/hash/subgraph/src/types/identifier.ts
```diff
@@ -35,27 +35,41 @@ export const extractEntityUuidFromEntityId = (entityId: EntityId): string => {
...
+
+export type EntityVersion = {
+  decisionTime: VersionTimespan;
+  transactionTime: VersionTimespan;
+};
+
+export type EntityRecordId = number;
 
 /**
  * An identifier of a specific edition of an `Entity` at a given `EntityVersion`
```
Alfred-Mountfield, line 51: Docs need updating

packages/hash/subgraph/src/types/identifier.ts
```diff
@@ -35,27 +35,41 @@ export const extractEntityUuidFromEntityId = (entityId: EntityId): string => {
...
-// ISO-formatted datetime string
-export type EntityVersion = Timestamp;
+export type VersionTimespan = {
+  start: Timestamp;
+  end?: Timestamp;
+};
+
+export type EntityVersion = {
+  decisionTime: VersionTimespan;
+  transactionTime: VersionTimespan;
```
Alfred-Mountfield, line 45: The more I see these the less I'm convinced by our naming, but I haven't managed to wrap my head around what we _should_ call it yet. 

The decision time (in my mind) is just the start _timestamp_, it's the _time_ at which you _decided_ it. 

There's then an associated time_span_ from that decision to the next decision.

packages/hash/subgraph/src/types/identifier.ts
```diff
@@ -35,27 +35,41 @@ export const extractEntityUuidFromEntityId = (entityId: EntityId): string => {
...
+};
+
+export type EntityRecordId = number;
```
Alfred-Mountfield, line 48: Should we type brand this?
cc @thehabbos007

packages/hash/subgraph/src/types/identifier.ts
```diff
@@ -113,16 +127,16 @@ export const isOntologyTypeEditionId = (
   );
 };
 
-export const isEntityAndTimestamp = (
-  editionId: object,
-): editionId is EntityIdAndTimestamp => {
+export const isEntityIdAndTimestamp = (
+  entityIdAndTimestamp: object,
```
Alfred-Mountfield, line 131: Good catch, thanks!

packages/hash/subgraph/src/types/vertex.ts
```diff
@@ -61,6 +61,6 @@ export type Vertices = {
   };
 } & {
   [_: EntityId]: {
-    [_: EntityVersion]: KnowledgeGraphVertex;
+    [_: Timestamp]: KnowledgeGraphVertex;
```
Alfred-Mountfield, line 64: This is a bit of a regression in expressiveness in my mind. 

We could have type-branded `EntityVersion` and then you'd know if you should be able to do a direct lookup for the thing or not. (Timestamps require a search, EntityVersions _should_ correspond to an exact vertex under a certain pinned axis)

I appreciate you've changed `EntityVersion` to now be this complex object thing (which I'm not entirely comfortable with but I don't have a better suggestion right now), maybe there's something else we can call these specific timestamps?

Alfred-Mountfield (commented):
Thanks Tim, monster amount of work going into this.

My comments sort of ended up being a train of thought, so I'd recommend you read them all before you start addressing them, as a lot of them can probably be answered in a single thread to avoid you having to write something like "see above" all over the place :)

packages/graph/clients/typescript/api.ts
```diff
@@ -700,63 +719,57 @@ export type Filter =
...
 
 /**
- * @type GraphElementEditionId
+ * @type GraphElementId
  * @export
  */
-export type GraphElementEditionId =
-  | GraphElementEditionIdOneOf
-  | GraphElementEditionIdOneOf1;
+export type GraphElementId = string;
```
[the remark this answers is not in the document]
TimDiekmann: It's either a base URI for ontology types or the entity id. The definition did not change, git was just not smart enough to notice, that this was not replaced but reordered. The order in the file changed as `GraphElementEditionid` is now called `GraphElementVertexid`, so `GraphElementId` was ordered to appear before `GraphElementVertexid`.

packages/graph/clients/typescript/api.ts
```diff
@@ -700,63 +719,57 @@ export type Filter =
...
+
+/**
+ * @type GraphElementVertexId
+ * @export
+ */
+export type GraphElementVertexId =
+  | GraphElementVertexIdOneOf
```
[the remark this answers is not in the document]
TimDiekmann: For this specific case, 98bd5de17c7288e696954a4431083f93f93db057 will improve the definition, but it's not really more readable. Generally, it's really hard to generate decent definitions for generic structs.

packages/graph/hash_graph/lib/graph/src/knowledge/entity/mod.rs
```diff
@@ -266,13 +279,21 @@ impl Entity {
...
     fn edition_id(&self) -> &Self::EditionId {
         &self.metadata.edition_id
     }
 
-    fn create_filter_for_edition_id(edition_id: &Self::EditionId) -> Filter<Self> {
-        Filter::for_entity_by_edition_id(*edition_id)
+    fn vertex_id(&self) -> Self::VertexId {
+        EntityVertexId::new(
+            self.edition_id().base_id(),
+            self.metadata().version().transaction_time().start,
```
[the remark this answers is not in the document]
TimDiekmann: Yes, this is going to be changed, but as there is no interface yet for the projection (#1738), this will be changed in #1743 to 
```rust
fn vertex_id(&self, time_axis: TimeAxis) -> Self::VertexId {
    EntityVertexId::new(self.edition_id().base_id(), match time_axis {
        TimeAxis::DecisionTime => self.metadata().version().decision_time().start.cast(),
        TimeAxis::TransactionTime => self.metadata().version().transaction_time().start.cast(),
    })
}
```

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -34,33 +33,57 @@ impl ToSchema for GraphElementId {
     }
 }
 
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Serialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct EntityVertexId {
+    base_id: EntityId,
+    version: TransactionTimestamp,
```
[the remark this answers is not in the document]
TimDiekmann: I think renaming `EntityVersion` is the better option here, but I don't really have a good name, `TemporalSlice` sounds good enough for me. Happy to change that in this PR or in a follow-up if desired.

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -34,33 +33,57 @@ impl ToSchema for GraphElementId {
...
 }
 
 // WARNING: This MUST be kept up to date with the enum variants.
 //   We have to do this because utoipa doesn't understand serde untagged:
 //   https://github.com/juhaku/utoipa/issues/320
-impl ToSchema for GraphElementEditionId {
+impl ToSchema for GraphElementVertexId {
     fn schema() -> openapi::Schema {
         openapi::OneOfBuilder::new()
             .item(OntologyTypeEditionId::schema())
```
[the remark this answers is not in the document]
TimDiekmann: I agree but I hesitate to do that in this PR. This PR is already pretty big and any bigger change here will have a huge impact on follow-ups as my stack of PRs has become pretty big. This is a general issue we have with utoipa and should be a standalone PR.

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/create-entity-page.tsx
Comment (TimDiekmann) [reply packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/create-entity-page.tsx]: ba619f9ca2f427265e684138d4d0b9ff0f677a1e

packages/hash/subgraph/src/stdlib/roots.ts
TimDiekmann: Yes, otherwise it results in 
> Element implicitly has an 'any' type because index expression is not of type 'number'

packages/hash/subgraph/src/types/identifier.ts
TimDiekmann: 63177e3e86d562d79e9759c51419882e80799d26

packages/hash/subgraph/src/types/identifier.ts
TimDiekmann: We could call it (applies to all locations) `decisionTimeRange`, but I don't really like that either.

Alfred-Mountfield: `decisionTimeValidRange`?

Alfred-Mountfield: `decisionTimeActiveRange`?

packages/graph/hash_graph/lib/graph/src/shared/identifier/mod.rs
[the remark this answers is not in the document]
Alfred-Mountfield: Let's visit this in a follow-up

Alfred-Mountfield (approved):

PR #1733, opened 2022-12-22
Replace requiredLinks with minItems 1 in entity type schemas

CiaranMn:

The original design for the entity type schemas had a `requiredLinks` field, the intention of which was to list the links from an entity of that type that are _required_.

This field is redundant, because each entry in the `links` array has a `minItems` option which has the same effect if set to a number greater than `0`.

I'm doing this now because it's been a bit annoying to have to handle the deprecated field in https://github.com/blockprotocol/blockprotocol/pull/861 and would ideally not have to. 

As a drive-by I also removed an inappropriate `maxItems: 1` from a user's `OrgMembership` links.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432087/1203602044074386/f) _(internal)_

## 🐾 Next steps

- The field needs to be removed from the Graph API and the auto-generated OpenAPI client types.
- The types at alpha.hash.ai won't actually change until the backend is redeployed/reseeded(?) – I'm not clear when/how they get generated.


TimDiekmann (approved):
Looks good to me, thank you @CiaranMn!

> - The field needs to be removed from the Graph API and the auto-generated OpenAPI client types.

Do we have an Asana ticket for this? I wasn't able to find any, would be good to capture this directly.

> - The types at alpha.hash.ai won't actually change until the backend is redeployed/reseeded(?) – I'm not clear when/how they get generated.

cc @thehabbos007

CiaranMn:
> Do we have an Asana ticket for this?

Thanks @TimDiekmann – here's the [task](https://app.asana.com/0/1202805690238892/1203522138014423/f) for the Graph API part of this work

PR #1741, opened 2022-12-23
Apply `simple-import-sort/*` and `unicorn/filename-case` in two packages

kachkaev:
## 🌟 What is the purpose of this PR?

This PR follows #1707, #1714, #1732 and #1731. Its goal is to reduce diff in #1699. The PR covers these packages:
- packages/hash/shared
- packages/hash/subgraph

I’m not enforcing new rules in the above workspaces because https://github.com/hashintel/hash/pull/1699 should land shortly. If there are any regressions until then, they’ll be caught by CI. If you see imports reshuffled by your IDE in the meantime, please let me know (this should not be happening).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202650508779539/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432087/1203388867587776/f) _(internal)_

## ❓ How to test this?

CI

PR #1699, opened 2022-12-20
Apply `simple-import-sort/*` and `unicorn/filename-case` in `packages/hash/frontend`

kachkaev:
## 🌟 What is the purpose of this PR?

This PR continues https://github.com/hashintel/hash/pull/1714, https://github.com/hashintel/hash/pull/1707, https://github.com/hashintel/hash/pull/1732, https://github.com/hashintel/hash/pull/1731, https://github.com/hashintel/hash/pull/1741 and https://github.com/hashintel/hash/pull/1742. It concludes the transition to `kebab-case` in file names and finishes applying `simple-import-sort/*` to imports and exports.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202650508779539/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432087/1203388867587776/f) _(internal)_

## 🚨 Blocked by

- ~https://github.com/hashintel/hash/pull/1721~

## 🔍 What does this change?

The entire repo uses `eslint-plugin-simple-import-sort` and `unicorn/filename-case`. These rules are now enfornced. `simple-import-sort/import`  and `simple-import-sort/export` autofixable.

## ⚠️ Known issues

- `unicorn/filename-case` does not look into folder names, i.e. `pathTo/my-file.ts` is not marked by this rule. I searched for `from "\..*[A-Z]` (regexp, case-sensitive) to make sure that all folders in `apps` are `kebab-case` too.

## 🛡 What tests cover this?

- CI

PR #1742, opened 2022-12-23
Apply `simple-import-sort/*` and `unicorn/filename-case` in `packages/hash/api`

kachkaev:

This PR follows #1707, #1714, https://github.com/hashintel/hash/pull/1732 and https://github.com/hashintel/hash/pull/1741. Its goal is to reduce diff in https://github.com/hashintel/hash/pull/1699. The PR covers `packages/hash/api`.

I’m not enforcing new rules in the above workspaces because https://github.com/hashintel/hash/pull/1699 should land shortly. If there are any regressions until then, they’ll be caught by CI. If you see imports reshuffled by your IDE in the meantime, please let me know (this should not be happening).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202650508779539/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432087/1203388867587776/f) _(internal)_

## 🔍 What does this change?

The entire repo uses `eslint-plugin-simple-import-sort` and `unicorn/filename-case`.

## ⚠️ Known issues

- `unicorn/filename-case` does not look into folder names, i.e. `pathTo/my-file.ts` is not marked by this rule. I searched for `from "\..*[A-Z]` (regexp, case-sensitive) to make sure that all folders in `apps` are `kebab-case` too.

## 🛡 What tests cover this?

- CI


packages/hash/api/src/graphql/resolvers/file/file-url-resolver.ts
```diff
@@ -3,7 +3,6 @@
 // @ts-nocheck
 import { FileProperties, ResolverFn } from "../../api-types.gen";
 import { GraphQLContext } from "../../context";
-import { File } from "../../../model";
 
 export const fileUrlResolver: ResolverFn<
   string,
```
yusufkinatas, line 8: Looks like this is removed instead of replacing?

packages/hash/api/src/graphql/resolvers/linked-aggregation/get-linked-aggregation.ts
```diff
@@ -5,7 +5,6 @@ import { ApolloError } from "apollo-server-express";
 
 import { QueryGetLinkedAggregationArgs, ResolverFn } from "../../api-types.gen";
 import { GraphQLContext } from "../../context";
-import { UnresolvedGQLLinkedAggregation, Aggregation } from "../../../model";
 
 export const getLinkedAggregation: ResolverFn<
```
yusufkinatas, line 9: Same on this file, this is removed, but I don't see a replacement

yusufkinatas (commented):
A bit confused about how linting is not failed due to my comments above.

kachkaev:
Thanks a lot @yusufkinatas, well spotted! Those files have `// @ts-nocheck` which I haven't noticed. Looks like we don't use them, but it's still nice to have them ready just in case!

yusufkinatas (approved):
Looks good to me.

@kachkaev Happy to help! 😊

PR #1745, opened 2022-12-23
Implement three simple Storybook stories for Design System components

kachkaev:

This PR follows https://github.com/hashintel/hash/pull/1660. In adds stories for `Button`, `Chip` and `LoadingSpinner` to begin with.

Screenshot 2022-12-23 at 19 29 03
Screenshot 2022-12-23 at 19 28 51

Default Storybook content is removed.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203513484217336/f) _(internal)_
- [Asana subtask](https://app.asana.com/0/0/1203513484217336/f) _(internal)_

## ⚠️ Known issues

- Inter font is not included in the design system, so Times New Roman is rendered instead.

  Screenshot 2022-12-23 at 19 29 10

  I could have moved font files into `packages/hash/design-system` but this would have blurred the scope of this PR. I suggest that we solve this problem separately.

## 🐾 Next steps

- Discuss the right approach to font loading
- Agree on when and how we write new stories
- Consider configuring Playwright for the Storybook.

## ❓ How to test this?

Check preview deployment


packages/hash/design-system/src/chip.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Chip } from "./chip";
+
+const meta = {
+  title: "Design System/Chip",
+  component: Chip,
+  tags: ["docsPage"],
+  args: {
+    label: "Example Chip",
+  },
+} satisfies Meta<typeof Chip>;
```
nathggns, line 12: What's the value of satisfies here vs just declaring meta as this type?

nathggns (approved):

packages/hash/design-system/src/chip.stories.tsx
kachkaev: This style comes from the official boilerplate, so there may be reasons beyond this simple example. I'd just stick with what's suggested upstream.

Issue #1746, opened 2022-12-26
Make Report compatible with std::error::Error based types


mayabyte:
### Related Problem

Let's say I have a library crate that uses error-stack for all its errors, and a pre-existing user-facing program that wants to use this library but doesn't want to do a big-bang migration of all its own error handling to use error-stack and may not ever care to - let's say it just uses `Box<dyn Error>` for all its errors. 

Integrating the library's error types into the application is currently very clumsy and weird, requiring us to either pull the `Context`s out of the `Report`s manually, stringify the errors, or some other tedious and undesirable pattern.

### Proposed Solution

Other error handling libraries provide some conversion options for integrating with std Error that would solve this issue for error-stack. By way of example, [eyre has the following](https://docs.rs/eyre/0.6.8/eyre/struct.Report.html#): 
- `impl AsRef<dyn Error + 'static> for Report`
- `impl Deref<Target = dyn Error + Send + Sync + 'static> for Report`
- `impl From<Report> for Box<dyn Error + 'static>`

### Alternatives

One possible alternative is implementing Error directly on Report. I can't tell whether this is deliberately being avoided, but neither eyre or anyhow do this for their error types so I assume there's a good reason not to choose this option.

indietyp:
Thanks for creating the issue! I just wanted to note that it may take a bit longer to resolve/respond to the issue.

Most of the people at HASH (and other contributors) are currently on Christmas/New Years vacation.🎄

Once everyone is back from enjoying some time with their close ones we'll be sure to take a closer look and respond properly!

Sorry for the inconvenience.

TimDiekmann:
Hi @mayabyte and thanks for filing this issue!

We’ve been looking into ways to improve the ergonomics of using `Report` in an environment that heavily relies on the usage of `Error`. From the start, we’ve considered implementing `Error` on `Report`. But we erred on the side of not doing it as it would allow you to construct a `Result<T, Report<Report<C>>` by using `into_report()` on `Result<T, Report<C>>`. In our own preliminary testing that ended up being quite frustrating, and easy to accidentally do.

We’ve experimented with a few other approaches and wonder if providing ways to turn the `Report` into an Error would be enough. Something like this:
```rust
fn into_error(self) -> impl Error + Send + Sync + 'static;
fn as_error(&self) -> &impl Error + Send + Sync + 'static;
```
Do you think this would help in situations such as yours?

mayabyte:
[...]

PR #1749, opened 2023-01-03
Allow conversion from `Report` to `Error`

TimDiekmann:

It's currently hard to deal with `Report` when an `Error` is required. This PR adds conversions from `Report` to `Error`.

## 🔗 Related links

- resolves https://github.com/hashintel/hash/issues/1746

## 🔍 What does this change?

- Add `Report::into_error` to convert `impl Error`
- Add `Report::as_error` to return `&impl Error`
- Add `From` implementation of `Report` for `Box<dyn Error>` (+ `Send` and/or `Sync`)

The changelog was changed to reflect these changes

## 🛡 What tests cover this?

Three new tests were added:
- A UI test to make sure, that `into_report` cannot be called on `Result<T, Report<C>>`
- Two tests to check the new methods/trait implementation by converting a `Report` to `Error` and ensure, that the provided values are still accessible

### DIFF
--- a/packages/libs/error-stack/src/error.rs
+++ b/packages/libs/error-stack/src/error.rs
@@ -1,0 +1,43 @@
+use core::fmt;
+#[cfg(nightly)]
+use core::{
+    any::{Demand, Provider},
+    error::Error,
+};
+#[cfg(not(nightly))]
+use std::error::Error;
+
+use crate::Report;
+
+#[repr(transparent)]
+pub(crate) struct ReportError<C>(Report<C>);
+
+impl<C> ReportError<C> {
+    pub(crate) const fn new(report: Report<C>) -> Self {
+        Self(report)
+    }
+
+    pub(crate) const fn from_ref(report: &Report<C>) -> &Self {
+        // SAFETY: `ReportError` is a `repr(transparent)` wrapper around `Report`.
+        unsafe { &*(report as *const Report<C>).cast() }
+    }
+}
+
+impl<C> fmt::Debug for ReportError<C> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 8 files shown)


packages/libs/error-stack/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+impl<C> fmt::Display for ReportError<C> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&self.0, fmt)
+    }
+}
+
+impl<C> Error for ReportError<C> {
+    #[cfg(nightly)]
+    fn provide<'a>(&'a self, demand: &mut Demand<'a>) {
+        self.0.frames().for_each(|frame| frame.provide(demand));
```
indietyp, line 41: Out of curiosity, any reason for choosing `for_each` over `for frame in self.frames()`?

packages/libs/error-stack/src/lib.rs
```diff
@@ -464,7 +464,6 @@
 )]
 
 extern crate alloc;
-extern crate core;
```
indietyp, line 467: I love how CLion always automatically adds this line...

indietyp (approved):
Looks good to me, just some minor non-blocking comments. I already noted this over private DMs, but with this change

```rust
#[test]
fn as_error_nest() {
    let report = create_report();
    let base = report.into_error();
    let report = Report::new(base);

    println!("{report:?}");
}
```

is possible. I think that is ok, as it does not break formatting and has quite the high barrier for creating it.

packages/libs/error-stack/src/error.rs
TimDiekmann: No specific reason. I used it here as it's only one line vs three lines.

PR #1751, opened 2023-01-03
Fix scroll lock issue on entity editor

yusufkinatas:

This PR enables scroll-lock while users are editing value cells on properties table for better UX.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432087/1203568198115136/f) _(internal)_

## ❓ How to test this?

1.  Open entity editor for any entity
2. Edit any value cell
3. Confirm that page scroll is locked while cells are being edited

## 📹 Demo

https://user-images.githubusercontent.com/39553853/210417714-fc1fadb1-0774-4382-a496-2a902ebe98c0.mov


packages/hash/design-system/src/use-scroll-lock.ts
```diff
@@ -3,54 +3,59 @@ import { useEffect, useLayoutEffect, useRef, useState } from "react";
...
+        ? element.offsetWidth - element.clientWidth
+        : getScrollbarSizeOfDocument();
 
       if (scrollbarSize > 0) {
         setLastScrollbarSize(scrollbarSize);
       }
     });
 
-    observerRef.current.observe(document.body);
+    observerRef.current.observe(element ?? document.body);
```
nathggns, line 26: You've put `document.body` as the default in multiple places. You can keep `element` as required in `useLastScrollbarSize` and just pass it from `useScrollLock`

packages/hash/design-system/src/use-scroll-lock.ts
```diff
@@ -3,54 +3,59 @@ import { useEffect, useLayoutEffect, useRef, useState } from "react";
...
+ * @param active is locked
+ * @param elementToLock an HTML element to lock it's scroll. Locks `document.body` if it's left empty
  */
-export const useScrollLock = (active: boolean) => {
-  const scrollbarSize = useLastScrollbarSize();
+export const useScrollLock = (active: boolean, elementToLock?: HTMLElement) => {
+  const scrollbarSize = useLastScrollbarSize(elementToLock);
 
   useLayoutEffect(() => {
+    const element = elementToLock ?? document.body;
```
nathggns, line 51: You can just default the `elementToLock` argument to `document.body`, right?

packages/hash/design-system/src/use-scroll-lock.ts
```diff
@@ -3,54 +3,59 @@ import { useEffect, useLayoutEffect, useRef, useState } from "react";
...
 
 /**
  * This function does the same thing as MUI's scroll-lock mechanism, but in a hook.
  * So we can use the same scroll-lock at custom components
+ * @param active is locked
+ * @param elementToLock an HTML element to lock it's scroll. Locks `document.body` if it's left empty
  */
-export const useScrollLock = (active: boolean) => {
-  const scrollbarSize = useLastScrollbarSize();
+export const useScrollLock = (active: boolean, elementToLock?: HTMLElement) => {
```
nathggns, line 47: You should probably make this take a ref, not the element itself, because reading the dom (as the consumer of this does with `getElementsByTagName`) can be a side effect so you probably ought not to be doing it at render time.

packages/hash/design-system/src/use-scroll-lock.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1751/commits/c0ac496c437d3eef0b0d97f21492359d945f0744

packages/hash/design-system/src/use-scroll-lock.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1751/commits/c0ac496c437d3eef0b0d97f21492359d945f0744

packages/hash/design-system/src/use-scroll-lock.ts
yusufkinatas: Makes sense, and updated it here https://github.com/hashintel/hash/pull/1751/commits/e0ebf7eb18b95df6cb0efa2550d3a92475a642b0 (reused `EditBarContext` even if it's not related with edit bar, because you already captured the ref of `main` element there)

PR #1753, opened 2023-01-03
Implement `transaction()` on `Store` trait

TimDiekmann:

From a generic `Store` context (such as the REST API) it's not to start a transaction. By adding a `transaction` function to the `Store` trait enables this feature.

## 🔍 What does this change?

- Make the `Store` trait alias an actual trait
- Add a `Transaction` trait to allow `commit()` and `rollback()`
- Add `Store::transaction()` to return a `Transaction
- Replace all usages of the raw API with the new transaction API


packages/graph/hash_graph/bench/benches/read_scaling/knowledge/complete/entity.rs
```diff
@@ -6,7 +6,7 @@ use graph::{
     identifier::account::AccountId,
     knowledge::{EntityMetadata, EntityProperties, LinkData},
     provenance::{OwnedById, UpdatedById},
-    store::{query::Filter, AccountStore, AsClient, EntityStore, PostgresStore},
+    store::{query::Filter, AccountStore, EntityStore, Store as _, Transaction},
```
TimDiekmann, line 9: We have a structure called "Store" in the benchmarks, thus importing the trait anonymously.

packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -30,4 +33,19 @@ pub use self::{
 ///
 /// In addition to the errors described in the methods of this trait, further errors might also be
 /// raised depending on the implementation, e.g. connection issues.
-pub trait Store = AccountStore + DataTypeStore + PropertyTypeStore + EntityTypeStore + EntityStore;
+#[async_trait]
+pub trait Store:
+    AccountStore + DataTypeStore + PropertyTypeStore + EntityTypeStore + EntityStore
+{
+    type Transaction<'t>: Transaction + Send
```
TimDiekmann, line 40: `Send` is required for utoipa and we probably don't want to add `Send` at every location

packages/graph/hash_graph/lib/graph/src/store/mod.rs
```diff
@@ -30,4 +33,19 @@ pub use self::{
...
+    where
+        Self: 't;
```
TimDiekmann, line 42: The syntax is ugly but sadly it's required currently. Even though GATs are stable, this may be lifted in the future.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -111,6 +111,43 @@ pub struct PostgresStore<C> {
     client: C,
 }
 
+#[async_trait]
+impl<C: AsClient> Store for PostgresStore<C> {
+    type Transaction<'t>
+    where
+        C: 't,
+    = PostgresStore<tokio_postgres::Transaction<'t>>;
```
TimDiekmann, line 119: Same as above, ugly syntax 😞

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -626,15 +663,9 @@ where
             .attach_printable_lazy(|| uri.clone())?
             .get(0))
     }
-
-    /// TODO - DOC
-    #[expect(clippy::missing_const_for_fn, reason = "Compile error")]
-    pub fn into_client(self) -> C {
-        self.client
-    }
```
TimDiekmann, line 634: We don't need this anymore as we call `commit` on the transaction directly

packages/graph/hash_graph/lib/graph/src/store/mod.rs
Alfred-Mountfield: Can you add a comment explaining this

packages/graph/hash_graph/lib/graph/src/store/mod.rs
Alfred-Mountfield: Worth a comment, or not really?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -111,6 +111,43 @@ pub struct PostgresStore<C> {
...
+                .await
+                .into_report()
+                .change_context(StoreError)?,
+        ))
+    }
+}
+
+#[async_trait]
+impl Transaction for PostgresStore<tokio_postgres::Transaction<'_>> {
+    async fn commit(self) -> Result<(), StoreError> {
```
Alfred-Mountfield, line 134: Is there a way for us to expose the client so we don't have to do this? As we could end up just redefining the whole `Transaction` interface.

(For example) I'm wondering how we'd call `Transaction::prepare`, or `Transaction::execute`.

Alfred-Mountfield: We can't just impl `AsClient` or something?

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -489,13 +479,7 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
         // The transaction is required to check if the update happened. If there were no returned
```
Alfred-Mountfield, line 479:
```suggestion
        // The transaction is required to check if the update happened. If there is no returned
```

packages/graph/hash_graph/lib/graph/src/store/mod.rs
TimDiekmann: Cannot say why it failed before but it's not needed 🤷🏻‍♂️ f058e1b938ed2e0d41709baef0c9e7dfc5c06af3

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: That's not possible as `Store` does not know anything about `Client`. It's intended, that you cannot call `prepare` or `execute` on a transaction directly as some stores may not have those methods defined. Instead, you directly call the functions defined on the `Store`, e.g. `create_entity`.

`AsClient` is an implementation detail for `Postgres`, which abstracts over `GenericClient`.

Alfred-Mountfield: Hm, I have the feeling our store abstraction isn't quite as satisfactory as we'd like, and that there are Postgres semantics bleeding over. But perhaps we'll find out with future development. Your explanation makes sense to me here though, thank you.

Alfred-Mountfield (approved):

PR #1757, opened 2023-01-04
Define range operations for timespans

TimDiekmann:

In order to allow structural queries to properly use the time projections defined in #1738, some common range operations has to be defined on `Timespan`

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1203363157432087/1203491211535116/f) _(internal)_

## 🚫 Blocked by

- #1715 
- #1718 
- #1735
- #1738 
- #1743

## 🔍 What does this change?

- Simplify the creation of `TimeProjection` as it's used in a lot of places
- Abstract away the `Kernel` and `Image` structs
- Define comparison operators for the timespan bounds. For this, I added two newtype structs: `LowerBound` and `UpperBound`, both private to the module. This is required as comparing the lower bounds and upper bounds behaves slightly different.
- With the operators defined, I added a few range operations, which I expect to use for the traversal of a subgraph:
    -  `intersect`
    - `union`
    - `overlaps`
    - `contains_range`
    - `contains` via `std::ops::RangeBounds` trait

An in-code documentation has been added to the parts, which likely lead to confusion


TimDiekmann:
Converting it to a draft as the logic to compare bounds is way more complicated than anticipated. Currently addressing the issues.

TimDiekmann:
Superseded by https://github.com/hashintel/hash/pull/1785

PR #1758, opened 2023-01-04
Open entity editor page when entity cell is clicked

yusufkinatas:

This PR lets users open entity editor from entity type editor's "entities" tab.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203507065426465/f) _(internal)_

## ❓ How to test this?

1.  Open an entity type page with existing entities
1.  Select "entities" tab from above
2. Click on one of the entity cells (check the demo below) and confirm that it opens entity editor

## 📹 Demo

https://user-images.githubusercontent.com/39553853/210568231-6279af9e-1665-4106-b68f-51612c1dbebe.mov


luisbettencourt (changes requested):
I found that an error is thrown when when clicking on entities outside of myworkspace. 

This happens in `http://localhost:3000/@system-user/types/entity-type/user?tab=entities` for example. When clicking on Alice I get redirected to `http://localhost:3000/@bob01/entities/3de72d40-622a-4e19-b7fe-ef11a992af74` when I should be redirected to `http://localhost:3000/@system-user/entities/3de72d40-622a-4e19-b7fe-ef11a992af74`.

We need to extract the workspaceId in the same way we are extracting the entityId I think.

yusufkinatas:
That's right, I forgot about testing entities on other accounts. Thanks!

luisbettencourt (approved):
Thanks for fixing the issue so quickly @yusufkinatas. LGTM

PR #1759, opened 2023-01-04
Add tracing to the TypeScript backend

thehabbos007:

This PR adds OpenTelemetry tracing to the TypeScript API. With the existing OTEL setup of the Graph, the systems can properly correlate spans in Jaeger.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203560791357489/f) _(internal)_

## 🔍 What does this change?

- Adds new environment variable for the OTEL gRPC endpoint in the TS API
- Adds tracing for express and GraphQL in the TS API
- Enriches the graph traces to indicate errors

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Checkout the branch,
1.  Run `yarn install`
1.  Run `yarn external-services up --build
1.  Run `yarn dev`
1.  Go to http://localhost:16686/ and explore the distributed trace logging

## 📹 Demo

image

image

image


.env
```diff
@@ -37,6 +37,8 @@ HASH_REALTIME_PORT=3333
 HASH_REDIS_HOST=localhost
 HASH_REDIS_PORT=6379
 
+HASH_OTLP_ENDPOINT=http://localhost:4317
```
Alfred-Mountfield, line 40: Do we need to do anything to disable this for the graph? (As in does the `.env` override the docker compose stuff, so if we drive-by comment out the env-var in the docker compose is this just going to re-enable it again?)

thehabbos007: The environment variables passed to docker compose are all explicit, and won't be passed to the containers unless specified in the `env` key. The variable names are also different, the one for the graph is `HASH_GRAPH_OTLP_ENDPOINT` and not `HASH_OTLP_ENDPOINT`, so we should be fine in that regard :)

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -84,6 +88,21 @@ where
...
 
+pub fn span_trace_layer() -> TraceLayer<
+    SharedClassifier<ServerErrorsAsFailures>,
+    impl Fn(&hyper::Request<hyper::Body>) -> tracing::Span + Clone,
+    DefaultOnRequest,
+    impl Fn(&hyper::Response<axum::body::BoxBody>, std::time::Duration, &tracing::Span) + Clone,
+    DefaultOnBodyChunk,
+    DefaultOnEos,
+    impl Fn(ServerErrorsFailureClass, Duration, &tracing::Span) + Clone,
+> {
```
Alfred-Mountfield, line 99: These are unnecessary
```suggestion
pub fn span_trace_layer() -> TraceLayer<
    SharedClassifier<ServerErrorsAsFailures>,
    impl Fn(&Request<Body>) -> tracing::Span + Clone,
    DefaultOnRequest,
    impl Fn(&hyper::Response<axum::body::BoxBody>, Duration, &tracing::Span) + Clone,
    DefaultOnBodyChunk,
    DefaultOnEos,
    impl Fn(ServerErrorsFailureClass, Duration, &tracing::Span) + Clone,
> {
```

Alfred-Mountfield: Also `hyper::header::HeaderName` below on line 117 doesn't need `hyper::` but it's too far away from the diff for me to comment.

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -129,7 +148,7 @@ fn parse_x_forwarded_for(headers: &http::HeaderMap) -> Option<Cow<'_, str>> {
 }
 
 // Based on https://github.com/tokio-rs/axum/pull/769
-pub fn span_maker(request: &hyper::Request<hyper::Body>) -> tracing::Span {
+fn span_maker(request: &hyper::Request<hyper::Body>) -> tracing::Span {
```
Alfred-Mountfield, line 151: Also unnecessary
```suggestion
fn span_maker(request: &Request<Body>) -> tracing::Span {
```

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -129,7 +148,7 @@ fn parse_x_forwarded_for(headers: &http::HeaderMap) -> Option<Cow<'_, str>> {
...
     let target = request.uri();
     let scheme: Cow<'static, str> = target.scheme().map_or_else(
         || "HTTP".into(),
```
Alfred-Mountfield, line 154: Also slightly too far from the diff, why do we need the `"http".into()` and `"https".into()`? Can't we just use `to_string` on all of them? Or is that to save us a few allocations with the `Cow` or something?

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -196,7 +215,36 @@ pub fn span_maker(request: &hyper::Request<hyper::Body>) -> tracing::Span {
...
         http.host = %host,
+        http.status_code = Empty,
+        otel.status_code = Empty,
     );
     tracing_opentelemetry::OpenTelemetrySpanExt::set_parent(&span, remote_context);
     span
 }
+
+fn span_on_response(
+    response: &hyper::Response<axum::body::BoxBody>,
```
Alfred-Mountfield, line 226: I believe this is redundant too
```suggestion
    response: &Response<axum::body::BoxBody>,
```

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
```diff
@@ -196,7 +215,36 @@ pub fn span_maker(request: &hyper::Request<hyper::Body>) -> tracing::Span {
...
+    _latency: Duration,
+    span: &tracing::Span,
+) {
+    let status = response.status().as_u16();
+    span.record("http.status_code", &tracing::field::display(status));
+    span.record("otel.status_code", "OK");
```
Alfred-Mountfield, line 232: Is this always `"OK"` or should it change depending on `status`?

Alfred-Mountfield: Is `on_response` always equivalent to `on_success`? In my head failures would also be responses

Alfred-Mountfield: Have you checked if these both trigger, and if they do does the order matter? Will `on_response` override the value from `on_failure` or vice versa?

packages/hash/api/src/graphql/opentelemetry.ts
```diff
@@ -0,0 +1,65 @@
...
+const unregisterInstrumentations = registerInstrumentations({
+  instrumentations: [
+    new HttpInstrumentation(),
+    new ExpressInstrumentation({
+      ignoreLayersType: [ExpressLayerType.MIDDLEWARE],
+    }),
+    new GraphQLInstrumentation({
+      allowValues: true,
+      depth: 5,
+    }),
```
Alfred-Mountfield, line 26: Are these settings based on anything? Do we want to add comments to explain them (e.g. is a depth of `5` a best practice?)

packages/hash/api/src/graphql/opentelemetry.ts
```diff
@@ -0,0 +1,65 @@
...
+  ],
+});
+
+export const registerOpenTelemetryTracing = (
+  otlpGrpcEndpoint: string | null,
+): (() => void) => {
+  if (!otlpGrpcEndpoint) {
+    logger.info(
+      "No OpenTelemetry Protocol endpoint given. Not sending tracespans anywhere.",
+    );
```
Alfred-Mountfield, line 36: Do we perhaps want to set this as a `warn`?

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
thehabbos007: Since we're returning a cow, we wouldn't allocate "HTTP" or "HTTPS" as a String using `into` here

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
thehabbos007: Thanks, done in 39700b961cfd014f3048ac96564a5a8c0837c584

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
thehabbos007: Thanks, done in 39700b961cfd014f3048ac96564a5a8c0837c584 :)

packages/graph/hash_graph/lib/graph/src/api/rest/middleware.rs
thehabbos007: From the spec, this should only be `"OK"` or `"ERROR"` https://opentelemetry.io/docs/reference/specification/common/mapping-to-non-otlp/#span-status

It does seem like we shouldn't actually set the `"OK"` state manually as it would override any error codes set by other parts of the application code. I think we can remove this `otel.status_code` record, but keep the one for the failure case

thehabbos007: Removed setting it in 50cad89ec9ed07786b7b71850a7fcf33ce231e4d

packages/hash/api/src/graphql/opentelemetry.ts
thehabbos007: The config isn't, the depth was just a best-guess for our deepest queries (our API is quite shallow because of of scalars, so it could probably be less.)

The `allowValues` makes it so values in queries are put into the tracing. We may want to set `sensitive` attributes in the GQL spec to prevent spilling anything that shouldn't be in logs

packages/hash/api/src/graphql/opentelemetry.ts
thehabbos007: I did consider that, I think I will do that. The app will function fine without tracing, though, but of course it could be useful to have it be a `warn` when it's misconfigured.

packages/hash/api/src/graphql/opentelemetry.ts
Alfred-Mountfield: Do you have an idea as to just how much we are logging and instrumenting? I'm wary to add all of this quite intrusive logging without us having any benchmarking in place. Obviously we don't have that infrastructure right now so we should make our best guess, just wondering if you have a sense if we're overdoing it already. (Needing a depth of 5 already makes me quite concerned about the quantity of things we're logging)

thehabbos007: I have tried to limit amount of logging by the `ignoreLayersType: [ExpressLayerType.MIDDLEWARE],` filter to prevent logging at every single pre/post-request step, so it definitely is a lot less that before, and all of our inner methods are not traced in TypeScript.

As for the depth, that depth of `5` is not needed, and honestly we can probably set it to 1 given our shallow API. I do want to point out that the default is `-1` which is infinite depth, so if anything we're setting an upper limit in this case :)

Alfred-Mountfield: That's a helpful thing to point out, and we can always configure it later. Cheers :)

Alfred-Mountfield (approved):
🎉

PR #1761, opened 2023-01-04
Move `rustfmt.toml` to the repository root

TimDiekmann:

To ensure we use the same rustfmt config on each crate, this should be moved to the repository root.

## 🔗 Related links

- [Slack discussion](https://hashintel.slack.com/archives/C03F7V6DU9M/p1672845445915869) _(internal)_

## 🔍 What does this change?

- Moves `rustfmt.toml` to the root
- Removes all other `rustfmt.toml`s


packages/engine/lib/stateful/src/field/field_type.rs
```diff
@@ -78,15 +78,15 @@ impl From<FieldTypeVariant> for DataType {
                 Field::new("item", DataType::from(field_type.variant), true),
             )),
             FieldTypeVariant::Struct(field_specs) => DataType::Struct(
+                // TODO: Enforce nullability of fields at initialisation.
```
Alfred-Mountfield, line 81:
```suggestion
                // TODO: Enforce nullability of fields at initialization.
```

Alfred-Mountfield (approved):

PR #1762, opened 2023-01-04
Adjust production dockerfile

thehabbos007:

This PR adjusts the production dockerfile to remove the hardcoded `kratos.dev.yml` references. Thanks @Alfred-Mountfield for finding this issue :green_heart: 

image


.env
```diff
@@ -22,6 +22,8 @@ HASH_KRATOS_PG_PASSWORD=kratos
 HASH_KRATOS_PG_DEV_DATABASE=dev_kratos
 HASH_KRATOS_PG_TEST_DATABASE=test_kratos
 
+HASH_GRAPH_LOG_LEVEL=info
```
Alfred-Mountfield, line 25: I know we discussed this yesterday, but does this end up used for the development docker compose too? Perhaps we just want to change the defaults in the prod docker compose file, and leave this undefined here

thehabbos007: It would take precedence over the dockerfile defaults, yes

Alfred-Mountfield: I think we should probably still have more in-depth logs in trace mode, just not in the prod docker compose

thehabbos007: In that case I'll revert the commit adding this .env value

Alfred-Mountfield: Can you also update 
```rml
      RUST_LOG: "${HASH_GRAPH_LOG_LEVEL:-graph=trace,hash_graph=trace,tokio_postgres=debug}"
```
in `docker-compose.prod.yml`

Alfred-Mountfield: Maybe to 
```yml
      RUST_LOG: "${HASH_GRAPH_LOG_LEVEL:-info}"
```

Alfred-Mountfield: If that's the right syntax 😅  (not sure)

thehabbos007: 169d82b97c245a57a4cd663902b8ff2e0ccdd37e and 4d014730279915ac1601e7aa51293320330c0ef9

Alfred-Mountfield (approved):

PR #1765, opened 2023-01-05
Remove `Dummy` entity type

Alfred-Mountfield:

The Type System previously didn't allow for link destinations to be "anything". As such, we introduced a "Dummy" entity type as a stand in for that during HASH development, and this PR removes it now that the Type System allows you to leave the destinations unconstrained (expressed as an empty array).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203449969023907/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Removes usages of the dummy entity type

No

## ⚠️ Known issues

None

## 🐾 Next steps

Update the type editors to allow for "any" link destinations.

## 🛡 What tests cover this?

Unsure about which specific ones, expect CI to pass.

## ❓ How to test this?

CI should be sufficient

## 📹 Demo

N/A


packages/hash/api/src/graph/system-types.ts
```diff
@@ -396,17 +383,6 @@ const textEntityTypeInitializer = async (graphApi: GraphApi) => {
   })(graphApi);
 };
 
-/**
- * @todo: remove this dummy entity type once we are able to define the block data link type without it
- * @see https://app.asana.com/0/1202805690238892/1203015527055368/f
- */
-const dummyEntityTypeInitializer = async (graphApi: GraphApi) => {
-  return entityTypeInitializer({
-    ...types.entityType.dummy,
```
benwerner01, line 405: I think it should also be removed from [the `ontology-types.ts` file](https://github.com/hashintel/hash/blob/93a13184191c24979f7fc941deee754da7f796cb/packages/hash/shared/src/ontology-types.ts#L113-L116)

benwerner01 (changes requested):
Looks good, think I spotted one remaining reference to the entity type

packages/hash/api/src/graph/system-types.ts
Alfred-Mountfield: Thanks, kind of annoyed I missed these 😅

thehabbos007 (approved):
Thanks, Alfie! Looks good to me :)

benwerner01 (dismissed):

packages/hash/api/src/graph/util.ts
```diff
@@ -218,11 +223,10 @@ export type EntityTypeCreatorParams = {
   }[];
   outgoingLinks?: {
     linkEntityType: EntityTypeWithMetadata;
-    destinationEntityTypes: (
-      | EntityTypeWithMetadata
-      // Some models may reference themselves. This marker is used to stop infinite loops during initialization by telling the initializer to use a self reference
-      | "SELF_REFERENCE"
-    )[];
+    destinationEntityTypes?: [
```
Alfred-Mountfield, line 226: Made this optional, the Type System doesn't allow an empty `oneOf`, it allows an _optional_ `oneOf`

thehabbos007 (dismissed):

thehabbos007 (approved):

PR #1769, opened 2023-01-05
Support editing property types with property objects or arrays

nathggns:

This removes a previously intentional crash when attempting to edit a property type which contained a property object or array, and implements the functionality to prefill the form with the current state of the object or array. 

It additionally corrects logic around saving whether a property array is infinite or not. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432084/1203554268413277/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Default values in a type form are now memoized, so will only get called again if the identity of the `getDefaultValues` function changes. This is an optimisation necessary by the more complicated nature of converting property arrays/objects into our internal draft state compared to pre-existing conversions for other type editing. This allows the wrapper of a type form (i.e, the property type form) to memoize their own `getDefaultValues` function to only change whenever a value it depends on changes, meaning `getDefaultValues` only changes when the property type itself changes.
- Removed lots of duplication in property type schema generation, and split up the files more sensibly, fixed some folder structure issue
- Fixed an issue where infinity on property type arrays wasn't being saved properly during property type schema generation
- `flattenedCustomExpectedValueList` in the form state is now the *entire* list, not just the list of whatever expected value is being edited. This means that data has been moved from the expected value type. This is because there is already a key which indicates which value is being edited, so this was duplicative and needlessly complicated.
- New `propertyTypeToFormDataExpectedValues` function which generates the expected values and flat list for form data for an existing property type
- `getDefaultValues` when editing a property type is now memoized so it only runs once, when opening the modal
- The form state is now split up for the expected value editor, so that it isn't saved to the property type editor form until you click save, which allows cancel to work as expected

## ❓ How to test this?

1. Create a property type with a complex set of expected values including nested property objects and arrays (with infinite and non infinite sizes)
2. Save it
3. Refresh
4. Go to edit it, and ensure all values are pre-filled as expected


packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/get-expected-value-descriptor.ts
```diff
@@ -0,0 +1,68 @@
+import { dataTypeOptions } from "./data-type-options";
```
Comment (nathggns) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/shared/get-expected-value-descriptor.ts:1]: This code is all abstracted from `custom-expected-value-builder`

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/get-property-type-schema.ts
```diff
@@ -0,0 +1,130 @@
+import {
```
Comment (nathggns) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/get-property-type-schema.ts:1]: Most of this comes from the `property-type-schema` file which has been deleted.

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/property-type-to-form-data-expected-values.ts
```diff
@@ -0,0 +1,108 @@
+import {
```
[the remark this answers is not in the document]
Comment (nathggns) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/property-type-to-form-data-expected-values.ts:1]: I think I do disagree. The variables it mutates are redefined on each call to `propertyTypeToFormDataExpectedValues`, and `processProperties` is only available within the scope of `propertyTypeToFormDataExpectedValues`. Cloning and merging arrays at every depth levels seems a) more complex and b) slower and c) harder to read. Going to leave as it is for now. 

I'm also wary of mutation generally, but mutation of locally owned variables when building new values is performant and sensible.

luisbettencourt (changes requested):
Thank you for fixing the issue @nathggns, and nice thinking splitting up the context in two!
Should be good to go after fixing the linting issue

luisbettencourt (approved):

PR #1773, opened 2023-01-05
Improve property type form validation

luisbettencourt:

This PR improves the validation on the property type form.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203490593648612/f) _(internal)_

## 🔍 What does this change?

- Changed validation to only happen `onSubmit`;
- Added error and success indicators, as well as a "Please select at least one expected value" to the expected values field;
- Removed `useTriggerValidation` which was triggering the validation of the expected values field when the form was initialized and was causing it show an error immediately for being empty.

## 📹 Demo

form


nathggns (approved):
> Removed useTriggerValidation which was triggering the validation of the expected values field when the form was initialized and was causing it show an error immediately for being empty.

This was there to ensure when the create property type form was opened with a title which already existed, an error would show. This is incidentally no longer necessary because you've changed the validation moment to onSubmit, so we don't need to validate when the form first opens anymore. 

This PR will conflict with https://github.com/hashintel/hash/pull/1769

PR #1775, opened 2023-01-06
Add helper function to get property types referenced in an entity type

Alfred-Mountfield:

This expands the `stdlib` to provide a helper to get all `PropertyType`s referenced by a given `EntityType` by utilising the `Edges` of a `Subgraph`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203662893654909/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

See commits, tiny diff

Not to my knowledge

## ⚠️ Known issues

We have a slightly unclear strategy about when to return errors for missing things

## 🐾 Next steps

Propagate this to the WIP new subgraph package in the Block Protocol

## 🛡 What tests cover this?

None

## ❓ How to test this?

We don't currently have tests for the subgraph library, if you want to see it in action you'll need to modify the existing code and use it somewhere

### DIFF
--- a/packages/hash/subgraph/src/stdlib/edge/entity-type.ts
+++ b/packages/hash/subgraph/src/stdlib/edge/entity-type.ts
@@ -1,0 +1,45 @@
+import {
+  BaseUri,
+  extractBaseUri,
+  extractVersion,
+} from "@blockprotocol/type-system";
+
+import { isConstrainsPropertiesOnEdge } from "../../types/edge/outward-edge-alias";
+import { OntologyTypeEditionId, VersionedUri } from "../../types/identifier";
+import { Subgraph } from "../../types/subgraph";
+
+/**
+ * Gets identifiers for all `PropertyType`s referenced within a given `EntityType` schema by searching for
+ * "ConstrainsPropertiesOn" `Edge`s from the respective `Vertex` within a `Subgraph`.
+ *
+ * @param subgraph {Subgraph} - The `Subgraph` containing the type tree of the `EntityType`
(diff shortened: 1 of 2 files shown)


packages/hash/subgraph/src/stdlib/edge/entity-type.ts
```diff
@@ -0,0 +1,45 @@
...
+ * "ConstrainsPropertiesOn" `Edge`s from the respective `Vertex` within a `Subgraph`.
+ *
+ * @param subgraph {Subgraph} - The `Subgraph` containing the type tree of the `EntityType`
+ * @param entityTypeId {OntologyTypeEditionId | VersionedUri} - The identifier of the `EntityType` to search for
+ * @returns {OntologyTypeEditionId[]} - The identifiers of the `PropertyType`s referenced from the `EntityType`
+ */
+export const getPropertyTypesReferencedByEntityType = (
+  subgraph: Subgraph,
+  entityTypeId: OntologyTypeEditionId | VersionedUri,
+): OntologyTypeEditionId[] => {
```
Alfred-Mountfield, line 22: I'm unsure if we want to return `PropertyType[]` or just the identifiers. If we do return identifiers I'm also unsure if we want to return `VersionedUri[]` or `OntologyTypeEditionId[]` as here. 

`OntologyTypeEditionId[]` is more efficient for future lookups as you can just do `subgraph.vertices[editionId.baseId][editionId.version]` but `VersionedUri` requires you calling `extract` again.

Returning `PropertyType[]` prematurely also doesn't feel ideal because we are assuming they need the whole object, and that involves more copying of data and more lookups

CiaranMn (approved):

PR #1783, opened 2023-01-09
Use version object on draft entity subgraph to prevent crash 

yusufkinatas:

This PR fixes a bug caused app crashing when users try to update any value cell on new entity page (draft / not-published).

## ❓ How to test this?

1.  Create an entity draft
1.  Change a value cell
2. Confirm that app is not crashing


TimDiekmann (approved):
Changes seems to work. However, I like to flag, that the subgraph should be [treated as immutable](https://app.asana.com/0/1203179076056209/1203636955054320/f) _(internal)_.

PR #1785, opened 2023-01-09
Implement `Interval` operations

TimDiekmann:

Interval operations like union, intersection, and difference are pretty complex. It's quickly complicated as the lower and upper bound have to be treaded differently (see the newly added `bounds.rs` file, which adds the operations).

We have quite a few different interval types in the Graph API, so instead of adding a single `Interval` type I decided to make this a trait and define the operations on that trait instead. To implement interval behavior, it's now as easy as implementing very few methods and the behavior is as expected (see Demo).

I played around a little bit with discrete ranges as well, thus I added this as a `CanonicalInterval`, which is either unbound or has a left-closed-right-open interval.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203701389454297/f) _(internal)_

## 🔍 What does this change?

- Adds the implementation of `Interval` and companion API
- Implement `Interval` for `VersionTimespan` and `Timespan`

## 🛡 What tests cover this?

A bunch of tests has been added to test the intervals

## ❓ How to test this?

Download the source code, go to the test module, and play around with the operations

## 📹 Demo

image
image


packages/graph/hash_graph/lib/interval-ops/src/lib.rs
```diff
@@ -0,0 +1,183 @@
+#![no_std]
+// Not required, reason: code quality
+#![feature(lint_reasons)]
+// Used to canonicalize intervals
+#![cfg_attr(feature = "canonicalize", feature(step_trait))]
```
thehabbos007, line 5: What is/was the reasoning behind the `canonical` name? :)

packages/graph/hash_graph/lib/interval-ops/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+            ContinuousInterval::from_range((Bound::Excluded(i32::MAX), Bound::Excluded(i32::MAX)))
+                .canonicalize()
+                .is_empty()
+        );
+    }
+
+    #[test]
+    #[rustfmt::skip]
+    fn test_merge() {
+        assert_eq!(ContinuousInterval::from_range(1..3) + ContinuousInterval::from_range(0..5), ContinuousInterval::from_range(0..5));
```
thehabbos007, line 115: Since we don't have a dedicated `union` operator, I want to point out that `+` could be interpreted as Minkowski addition (pairwise addition before union), https://math.stackexchange.com/a/659134. Could be slightly problematic, but likely not to cause confusion as-is

packages/graph/hash_graph/lib/interval-ops/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+        assert_eq!(ContinuousInterval::from_range(1..3) + ContinuousInterval::from_range(1..5), ContinuousInterval::from_range(1..5));
+        assert_eq!(ContinuousInterval::from_range(1..3) + ContinuousInterval::from_range(2..5), ContinuousInterval::from_range(1..5));
+        assert_eq!(ContinuousInterval::from_range(1..3) + ContinuousInterval::from_range(3..5), ContinuousInterval::from_range(1..5));
+        // ContinuousIntervals are disjoint
+        assert_eq!(ContinuousInterval::from_range(1..3).union(ContinuousInterval::from_range(4..5)), None);
```
thehabbos007, line 120: I'd lean on this syntax instead for consistency

packages/graph/hash_graph/lib/interval-ops/src/lib.rs
TimDiekmann: Originally, I had an implementation, which is only used for discrete values, such as integers. The main difference was to have unified bounds (left-closed-right-opened). It's probably worth just removing the struct again.

packages/graph/hash_graph/lib/interval-ops/src/lib.rs
TimDiekmann: The Minkowski Sum is defined on sets, but not on intervals. The main difference is that we don't store any values between the lower and upper bounds but only the bounds themselves. But generally, I agree that we should remove the operator overloading as using the functions are both, more expressive and more powerful.

packages/graph/hash_graph/lib/interval-ops/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[rustfmt::skip]
+    fn test_intersection() {
+        assert_eq!(ContinuousInterval::from_range(1..3) * ContinuousInterval::from_range(2..4), ContinuousInterval::from_range(2..3));
+        assert_eq!(ContinuousInterval::from_range(1..3) * ContinuousInterval::from_range(3..4), ContinuousInterval::empty());
+        assert_eq!(ContinuousInterval::from_range(1..2) * ContinuousInterval::from_range(3..4), ContinuousInterval::empty());
+        assert_eq!(ContinuousInterval::from_range(1..) * ContinuousInterval::from_range(3..4), ContinuousInterval::from_range(3..4));
+    }
+
+    #[test]
+    fn adjacency() {
```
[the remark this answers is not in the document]
TimDiekmann: Good point, will ade some more tests

packages/graph/hash_graph/lib/interval-ops/src/bounds.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+}
+
+impl<B, T> UpperBoundComparison<T> for B
+where
+    T: PartialOrd,
+    B: UpperBound<T>,
+{
+}
```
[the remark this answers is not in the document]
TimDiekmann: Yes. The trait is used to make the implementation details private while keeping the nicer way to call the methods instead of having free-standing functions.

TimDiekmann:
Thanks a lot @thehabbos007, I responded to the comments and will address them tomorrow! 🙂

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -193,84 +221,1191 @@ pub trait Interval<T>: Sized {
...
+            [included_excluded(0, 15)],
+            [included_excluded(0, 5)],
+        );
+        test(
+            included_excluded(5, 15),
+            included_included(0, 10),
+            [included_included(5, 10)],
+            [included_excluded(0, 15)],
+            [included_excluded(0, 15)],
+            [excluded_excluded(10, 15)],
```
thehabbos007, line 485: How does this turn into a `excluded_excluded` when using `difference`? I assume it isn't wrong, but it's strange to see :)

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -193,84 +221,1191 @@ pub trait Interval<T>: Sized {
...
+        // union:        (-------] | (-------]
+        // merge:        (-------] | (-------]
+        // difference:   (-)       |       (-]
+        test(
+            excluded_included(0, 10),
+            included_included(5, 15),
+            [included_included(5, 10)],
+            [excluded_included(0, 15)],
+            [excluded_included(0, 15)],
+            [excluded_excluded(0, 5)],
```
thehabbos007, line 523: I guess the same result here as above. I suppose it does make sense as we'd exclude the lower bound of range B, which would make the upper bound of the resulting interval open ended while the beginning of range A is open ended

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -193,84 +221,1191 @@ pub trait Interval<T>: Sized {
...
+            unbounded_unbounded(),
+            [included_included(5, 10)],
+            [unbounded_unbounded()],
+            [unbounded_unbounded()],
+            [],
+        );
+    }
+
+    #[test]
+    fn test_equal() {
```
thehabbos007, line 1396: Beautiful :D

packages/graph/hash_graph/lib/interval-ops/src/interval_bounds.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl<T: PartialOrd> Sub for IntervalBounds<T> {
+    type Output = Self;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        let difference = self.difference(rhs);
+        assert_eq!(
+            difference.len(),
+            1,
+            "interval union result in disjoint spans"
```
thehabbos007, line 104:
```suggestion
            "interval difference resulted in disjoint spans"
```

packages/graph/hash_graph/lib/interval-ops/src/interval_bounds.rs
```diff
@@ -0,0 +1,116 @@
...
+        (self.lower, self.upper)
+    }
+}
+
+impl<T: PartialOrd> Add for IntervalBounds<T> {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        let union = self.union(rhs);
+        assert_eq!(union.len(), 1, "interval union result in disjoint spans");
```
thehabbos007, line 91:
```suggestion
        assert_eq!(union.len(), 1, "interval union resulted in disjoint spans");
```

thehabbos007 (commented):
Thank you, Tim! I looked at the tests and tried to spot any issues, which I didn't (although I may have missed some as they are all sort of similar). This is a lot more assuring in the logic being correct!

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/timespan.rs
```diff
@@ -142,19 +139,15 @@ pub struct Timespan<A> {
 
 impl<A> Timespan<A> {
     #[must_use]
-    pub fn into_continuous_interval(self) -> ContinuousInterval<Timestamp<A>> {
-        ContinuousInterval::from_range((Bound::from(self.start), Bound::from(self.end)))
+    pub fn into_continuous_interval(self) -> IntervalBounds<Timestamp<A>> {
```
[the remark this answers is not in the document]
TimDiekmann: Good flag: a85b833c0c872de1ca24995e886f7d80aa06b0ec

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -1,14 +1,72 @@
...
+    fn upper_bound(&self) -> Bound<&T>;
+
+    fn into_lower_bound(self) -> Bound<T>;
+    fn into_upper_bound(self) -> Bound<T>;
+}
+
+enum Return<T> {
+    None,
+    One(Once<T>),
+    Two(Chain<Once<T>, Once<T>>),
```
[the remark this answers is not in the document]
TimDiekmann: The `Once` implicitly has an `Option`, which is `take`n when iterating. As we only return `IntervalIter<T>` as `impl ExactSizeIterator<T>` this is the easiest option. (btw, `ExactSizeIterator` does define `fn len()` and `fn is_empty()`)

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -65,124 +119,98 @@ pub trait Interval<T>: Sized {
...
     }
 
-    /// Returns a new interval that contains all points in both intervals.
-    ///
-    /// In comparison to [`Self::union`], this method does not require the intervals to be adjacent
-    /// or overlapping.
+    /// Returns the complement of this interval.
     #[must_use]
-    fn merge(self, other: impl Interval<T>) -> Self
+    fn complement(self) -> IntervalIter<Self>
```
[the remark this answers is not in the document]
TimDiekmann: Added a doc-string and some code comments: 4f9b71214e7b02bb1e3c53b9129796b20e5f7700

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: The upper bound of `lhs` is excluded and the upper bound of `rhs` is included. The upper bound of `lhs` is included in the difference, but the upper bound of `rhs` is not (as it's subtracted).

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: Yes, it's the same as above, but for the lower bounds

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: You can check the result by running
```sql
SELECT numrange(5, 15, '[)') - numrange(0, 10, '[]');
```

packages/graph/hash_graph/lib/interval-ops/src/interval_bounds.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn into_bound(self) -> (Self::LowerBound, Self::UpperBound) {
+        (self.lower, self.upper)
+    }
+}
+
+impl<T: PartialOrd> Add for IntervalBounds<T> {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        let union = self.union(rhs);
```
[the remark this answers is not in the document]
TimDiekmann: I didn't think of the `union` keyword but no, it does not conflict. 🙂 Sometimes, Rust is smart 🙃

TimDiekmann:
Thanks, @thehabbos007. I responded to your review and addressed the requested changes. On the last two comments this was an oversight as I wanted to remove the operator overloading entirely.

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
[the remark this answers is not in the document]
thehabbos007: Ah i see, that does make sense. Thanks :)

thehabbos007 (approved):
Great, thank you! Looks good to me :)

PR #1786, opened 2023-01-09
Introduce globally applied clippy lints for Rust crates

TimDiekmann:

As noticed previously and again noticed in https://github.com/hashintel/hash/pull/1785, specifying all lints each time is both, annoying and error-prone. This moves many lints a global `.cargo/config.toml`.

## 🔗 Related links

- [Slack discussion](https://hashintel.slack.com/archives/C03F7V6DU9M/p1672845445915869) _(internal)_

## 🔍 What does this change?

- Go through the list of clippy lints and pick a wide range of decent lints
- Adjust benchmarking code


packages/graph/hash_graph/.config/config.toml
```diff
@@ -0,0 +1,102 @@
+[target.'cfg(all())']
```
thehabbos007, line 1: When we say `global` here do we mean global for the hash graph project? Do we intend on sharing these lints with engine?

packages/graph/hash_graph/.config/config.toml
```diff
@@ -0,0 +1,102 @@
...
+rustflags = [
+    "-Wfuture_incompatible",
+    "-Wnonstandard_style",
+    # "-Wrust_2018_idioms",
```
thehabbos007, line 5: Why is this commented out vs. not just removed? Is there a 2021 equivalent we'd rather use that isn't out yet?

packages/graph/hash_graph/lib/graph/src/lib.rs
```diff
@@ -14,56 +14,11 @@
 #![feature(bound_map)]
 #![cfg_attr(all(doc, nightly), feature(doc_auto_cfg))]
 #![cfg_attr(not(miri), doc(test(attr(deny(warnings, clippy::all)))))]
-#![warn(
```
thehabbos007, line 17: I very much enjoy this being in a separate config file now :smile:

thehabbos007 (approved):
Looks good to me. Thank you for doing this and fixing the resulting lints, Tim!

packages/graph/hash_graph/.config/config.toml
TimDiekmann: Sorry, yes, for now I mean global for the Graph API workspace. I also intend to share it across the whole repository at some point.

packages/graph/hash_graph/.config/config.toml
TimDiekmann: `rust_2018_idioms` are idiomatic practices, which were introduced with the 2018 edition, those are not outdated. There are a few lints I like to disable, but I didn't want to address those in this PR. The 2021 edition AFAIK does not have such a lint.

PR #1790, opened 2023-01-10
Test cell update function of `Grid` component

yusufkinatas:

This PR implements a basic test case on entity editor page, just to test  `Grid` components most basic "update cell value" functionality.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432084/1203533294568155/f) _(internal)_

## 🔍 What does this change?

- Fix a console warning related with `notched` prop of MUI TextField's `outlined` variant
- Fix a console warning related with `controlling an uncontrolled component`
- Fix a console warning related with `usage of :first-child`

## 🐾 Next steps

- When data mutations on tests are supported, implement proper tests for entity editor

## ❓ How to test this?

1.  Check CI or run test on your local

## 📹 Demo

Check CI


packages/hash/playwright/tests/entity-creation.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  canvasLocator: Locator,
+  colIndex: number,
+) => {
+  const canvasPos = await canvasLocator.boundingBox();
+
+  if (!canvasPos) {
+    throw new Error("canvasPos not found");
+  }
+
+  const cellY = canvasPos.y + 20 + ROW_HEIGHT * (colIndex + 1);
```
kachkaev, line 29:
```suggestion
  const cellY = canvasPos.y + 20 + ROW_HEIGHT * (rowIndex + 1);
```
?

packages/hash/playwright/tests/entity-creation.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { loginUsingTempForm } from "./shared/login-using-temp-form";
+import { resetDb } from "./shared/reset-db";
+import { expect, Locator, Page, test } from "./shared/runtime";
+
+const getCellText = async (
+  canvasLocator: Locator,
+  colIndex: number,
+  rowIndex: number,
```
kachkaev, line 8: There is no conventional approach to indexing rows and columns. The numbers can be zero-based, one-based or even mixed. Happy for us to use any approach, but it'd be great to add a small JSDoc as a hint.

Example (this may be incorrect):

```suggestion
  /** zero-based (column A → 0) */
  colIndex: number,
  /** zero-based (header row → 0) */
  rowIndex: number,
```

packages/hash/playwright/tests/entity-creation.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { loginUsingTempForm } from "./shared/login-using-temp-form";
+import { resetDb } from "./shared/reset-db";
+import { expect, Locator, Page, test } from "./shared/runtime";
+
+const getCellText = async (
+  canvasLocator: Locator,
```
kachkaev, line 6: Playwright docs don’t add `Locator` suffix to variable names. You refer to canvas as `canvas` in the test itself, we could do so in the helper functions too:

```suggestion
  canvas: Locator,
```

packages/hash/playwright/tests/entity-creation.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  await page.mouse.move(cellX, cellY);
+  await page.mouse.click(cellX, cellY);
+};
+
+test.beforeEach(async () => {
+  await resetDb();
+});
+
+const value1 = "John";
+const value2 = "Doe";
```
kachkaev, line 40: Unlike in app code, string repetition in tests is encouraged. The closer to humans we are, the more likely the tests will react to real changes in product (which is what we want). We can keep `value1` and `value2`, but if we switch to reading and writing strings directly, the tests will read a bit better:

```diff
-await page.keyboard.type(value1);
+await page.keyboard.type("John");

-expect(cell1Text).toBe(value1);
+expect(cell1Text).toBe("John");

```

kachkaev (changes requested):
Looking good! Let's replace `colIndex` with `rowIndex` in one place (unless I’ve misread your intent). Other comments are optional are are more like general thoughts we can use when writing future tests.

packages/hash/playwright/tests/entity-creation.spec.ts
yusufkinatas: Oops, misnamed it sorry 🤦‍♂️

packages/hash/playwright/tests/entity-creation.spec.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1790/commits/29548ceb30f92b07c05ef2c19cce6daf2ad59f94

packages/hash/playwright/tests/entity-creation.spec.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1790/commits/29548ceb30f92b07c05ef2c19cce6daf2ad59f94

packages/hash/playwright/tests/entity-creation.spec.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1790/commits/29548ceb30f92b07c05ef2c19cce6daf2ad59f94

packages/hash/playwright/tests/entity-creation.spec.ts
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/1790/commits/29548ceb30f92b07c05ef2c19cce6daf2ad59f94

packages/hash/playwright/tests/entity-creation.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { loginUsingTempForm } from "./shared/login-using-temp-form";
+import { resetDb } from "./shared/reset-db";
+import { expect, Locator, Page, test } from "./shared/runtime";
+
+const getCellText = async (
+  canvas: Locator,
+  /** zero-based (first column -> 0) */
+  colIndex: number,
+  /** zero-based (first row after header row -> 0) */
```
kachkaev, line 9: I guess it’s `-1`-based then 😅 But we can keep it as is.

kachkaev (approved):

PR #1792, opened 2023-01-10
Support `startsWith`, `endsWith`, and `containsSegment` in structural queries

TimDiekmann:

We want more operators in structural queries

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203763222255819) _(internal)_

## 🛡 What tests cover this?

- Three more REST API tests were added to covert the new functions


packages/graph/hash_graph/lib/graph/src/store/postgres/query/condition.rs
```diff
@@ -82,6 +85,23 @@ impl Transpile for Condition<'_> {
...
+            Condition::EndsWith(lhs, rhs) => {
+                lhs.transpile(fmt)?;
+                fmt.write_str(" LIKE '%' || ")?;
+                rhs.transpile(fmt)
+            }
+            Condition::Contains(lhs, rhs) => {
+                lhs.transpile(fmt)?;
+                fmt.write_str(" LIKE '%' || ")?;
+                rhs.transpile(fmt)?;
+                fmt.write_str(" || '%'")
```
thehabbos007, line 103: You did end up using `like` for everything. Perfection :laughing:

TimDiekmann:
This was supposed to be in draft as I want to add more test cases

TimDiekmann:
Rebased onto main, will add some more tests as a next step

apps/hash-graph/lib/graph/src/store/postgres/query/condition.rs
```diff
@@ -89,6 +92,23 @@ impl Transpile for Condition<'_> {
...
+                lhs.transpile(fmt)?;
+                fmt.write_str(" LIKE '%' || ")?;
+                rhs.transpile(fmt)
+            }
+            Condition::Contains(lhs, rhs) => {
+                lhs.transpile(fmt)?;
+                fmt.write_str(" LIKE '%' || ")?;
+                rhs.transpile(fmt)?;
+                fmt.write_str(" || '%'")
+            }
```
thehabbos007, line 111: We've discussed this offline:
Since we're dealing with `text` here, we wouldn't be able to query into entity properties with this query. We'd have `JSONB` in the field and `TEXT` in the parameter. This causes an exception
```
╰─▶ db error: ERROR: operator does not exist: jsonb ~~ text
    HINT: No operator matches the given name and argument types. You might need to add explicit type casts.
    ├╴at /home/ahmad/dev/hash/hash/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs:84:14
    ├╴backtrace with 215 frames (1)
    ╰╴span trace with 4 frames (1)
```

This is unfortuante as it would have been correct if we could somehow cast the search parameter to JSONB, but we don't have enough information about the type we're operating on to decide to do this or not, just the fact that we're in `properties` for the query path which isn't ideal to base this casting logic on.

Furhtermore we'd want to support lists in the querying and not just strings as mentioned in the above comment - probably not trivial to implement.

thehabbos007 (commented):
Thank you Tim. We've talked offline about this and decided that it may be best to put this in draft for now. Otherwise great to see the query builder being so easy to add to!

I've added comments about what parts would need to change/need more tweaking from us :)

TimDiekmann:
Put back into draft due to decisions being made to change the implementation.

apps/hash-graph/lib/graph/src/store/postgres/query/condition.rs
TimDiekmann: This wasn't an easy fix but I solved it by converting `jsonb` to `text` on demand.

apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -225,16 +225,68 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
...
                 lhs.as_ref()
-                    .map(|expression| self.compile_filter_expression(expression)),
+                    .map(|expression| self.compile_filter_expression(expression).0),
                 rhs.as_ref()
-                    .map(|expression| self.compile_filter_expression(expression)),
+                    .map(|expression| self.compile_filter_expression(expression).0),
             ),
+            Filter::StartsWith(lhs, rhs) => {
+                let (left_filter, left_parameter) = self.compile_filter_expression(lhs);
+                dbg!(&left_filter, left_parameter);
```
thehabbos007, line 240: This one slipped in :)
```suggestion
```

thehabbos007 (commented):
Looks good to me. Thank you! Nice that you found that way to stringify the json object for now. I have a single comment and I suppose you want to rebase the `<'p>` away :)

TimDiekmann:
Yes, I have local changes to do this already 🙂

thehabbos007 (approved):

PR #1794, opened 2023-01-10
`error-stack` unify hook output of location hook

indietyp:

The location hook is currently inconsistent with pretty-print without color and no pretty-print; previously, if the color was deactivated, but `pretty-print` was activated, the output would read `<location>`, while without the pretty-print feature, it would read `at <location>`. This PR changes the behaviour, so that without color enabled, `pretty-print` will output `at <location>.`

## 🔗 Related links

This was discovered after inconsistencies were found in the snapshots of #1558.


Alfred-Mountfield (dismissed):

packages/libs/error-stack/src/fmt/hook.rs
```diff
@@ -492,12 +494,43 @@ mod default {
         });
     }
 
+    #[cfg(feature = "pretty-print")]
+    enum LocationDisplay<'a> {
+        Color(&'a Location<'static>),
+        None(&'a Location<'static>),
+    }
```
TimDiekmann, line 501: I'd prefer this to be a struct containing `location: &'a Location<'static>` and an enum with the color variants instead. That avoids having the same inner variant on all enum variants.

TimDiekmann (changes requested):
Thanks @indietyp!
The logic looks good to me. I have a small request on the code style.

packages/libs/error-stack/src/fmt/hook.rs
indietyp: fixed in [e3cb384](https://github.com/hashintel/hash/pull/1794/commits/e3cb38419de0d10e7e93f7c80c0721701fe95606)

TimDiekmann (approved):

PR #1797, opened 2023-01-10
Refactor impure graph context

thehabbos007:

This PR refactors the usages of the impure graph context such that we are able to more easily pass data sources down to every usage through the GQL data source list.

By doing so, I've introduced the `uploadProvider` into the dataSource list of GQL along with the `ImpureGraphContext` to verify the functionality (and because I will need it in file upload work)

## 🔗 Related links

- Prompted by [Asana task](https://app.asana.com/0/1203363157432094/1202636344465854/f) _(internal)_

## 🔍 What does this change?

- A new `dataSourceToImpureGraphContext` helper that is called from every Graph-related resolver
- Added `uploadProvider` to be used in the `ImpureGraphContext`
- Adjusted tests for a mock uploader that will fail instantly

## 🐾 Next steps

- Continue working on file uploads [Asana task](https://app.asana.com/0/1203363157432094/1202636344465854/f) _(internal)_

## 🛡 What tests cover this?

- Existing integration tests

## ❓ How to test this?

Start the API/workspace as usual and observe everything should be the same


packages/hash/api/src/graph/util.ts
```diff
@@ -160,37 +160,31 @@ export const generateSystemPropertyTypeSchema = (
  */
 export const propertyTypeInitializer = (
   params: PropertyTypeCreatorParams,
-): ((graphApi: GraphApi) => Promise<PropertyTypeWithMetadata>) => {
+): ((context: ImpureGraphContext) => Promise<PropertyTypeWithMetadata>) => {
   let propertyType: PropertyTypeWithMetadata;
 
-  return async (graphApi?: GraphApi) => {
+  return async (ctx?: ImpureGraphContext) => {
```
Alfred-Mountfield, line 166: This is why I really don't like the shadowing / re-define lint 🙃  

We're presumably breaking our own conventions about not shortening variable names just so that the lint doesn't complain and these two things don't conflict. Which sort of worries me even more because it'd be too easy to say `context.` inside here and not know there's a problem.

packages/hash/api/src/graph/util.ts
```diff
@@ -160,37 +160,31 @@ export const generateSystemPropertyTypeSchema = (
...
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- @todo improve logic or types to remove this comment
     if (propertyType) {
       return propertyType;
-    } else if (!graphApi) {
+    } else if (!ctx) {
       throw new Error(
-        `property type ${params.title} was uninitialized, and function was called without passing a graphApi object`,
+        `property type ${params.title} was uninitialized, and function was called without passing a impure graph object`,
```
Alfred-Mountfield, line 172: Can we change log lines like this in this diff? It's not an impure graph object right? It's a context object that encapsulates the accessors for the API, it's not _a graph_

packages/hash/api/src/graph/util.ts
Alfred-Mountfield: Actually maybe we're not even doing that here? The `context` above should just be in the scope of the return type presumably, and not cause problems.

packages/hash/api/src/graphql/resolvers/knowledge/block/block.ts
```diff
@@ -1,16 +1,19 @@
...
 import { UnresolvedBlockGQL } from "../graphql-mapping";
 
 export const blocksResolver: ResolverFn<
   Promise<UnresolvedBlockGQL[]>,
   {},
   GraphQLContext,
   QueryBlocksArgs
-> = async (_, params, { dataSources: { graphApi } }) => {
+> = async (_, params, { dataSources }) => {
+  const ctx = dataSourceToImpureGraphContext(dataSources);
```
Alfred-Mountfield, line 13: Nit: can we try and use `context` instead of `ctx`? It's more consistent with the rest of our code.

packages/hash/api/src/graphql/resolvers/util.ts
```diff
@@ -1,20 +1,12 @@
...
-//       entityId: "c1"
-//     }
-//   }
-// },
-export type LinkedDataDefinition = {
-  aggregate?: AggregateOperationInput;
-  entityTypeId?: string;
-  entityId?: string;
-  entityVersionId?: string;
+export const dataSourceToImpureGraphContext = ({
```
Alfred-Mountfield, line 4:
```suggestion
export const dataSourcesToImpureGraphContext = ({
```

packages/hash/api/src/graph/index.ts
```diff
@@ -17,11 +18,12 @@ import {
 
 export type ImpureGraphContext = {
   graphApi: GraphApi;
+  uploadProvider: UploadableStorageProvider;
   /** @todo: add logger? */
 };
```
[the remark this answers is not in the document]
thehabbos007: The biggest alternative I see is that we use dependency injection, but I don't think we should do that. We could make the context items optional if we don't require them everywhere, but that makes it annoying to deal with for operations that require the dependencies

packages/hash/api/src/graph/system-types.ts
```diff
@@ -8,6 +7,7 @@ import {
 } from "@hashintel/hash-subgraph";
 
 import { logger } from "../logger";
+import { ImpureGraphContext } from ".";
```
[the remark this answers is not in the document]
thehabbos007: I think this was technically correct as it was importing from the current relative directory which contained the `ImpureGraphContext` in `index.ts`?

packages/hash/api/src/graph/util.ts
thehabbos007: It's really annoying that the names clash. I think we don't use the outermost context, so perhaps we can just call it `_` outside and `context` inside?

thehabbos007: It seems we aren't using those outer contexts at all. Should we get rid of them?

thehabbos007: 197773ac120481acea936543ecee2034e9ee1257

packages/hash/api/src/graph/util.ts
thehabbos007: Will call this a `'context object'`

thehabbos007: 9816be5129076f9d6ad32f8369b5e31c103e9f44

packages/hash/api/src/graphql/resolvers/knowledge/block/block.ts
thehabbos007: Sorry, i actively wanted this to be `context`, but I seem to have missed these. Thanks!

thehabbos007: eccdaa1a01bc30a69de4100d385ad13ee319892c

packages/hash/api/src/graphql/resolvers/util.ts
thehabbos007: c768e5a8990d4c9759cbe2b7c9bf6a6e81129958 Thanks!

packages/hash/api/src/graph/index.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Yeah, perhaps it is a non-issue :)

Lets see how it ends up feeling in usage

packages/hash/api/src/graph/system-types.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Ah it's from an `index.ts` 🙃 Gross. Would we want to say `import { ImpureGraphContext } from "./index"` considering we want to move away from index files eventually anyway?

packages/hash/api/src/graph/util.ts
Alfred-Mountfield: <insert egg emoji here>

packages/hash/api/src/graph/system-types.ts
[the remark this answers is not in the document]
thehabbos007: We can do that, yes

Alfred-Mountfield (approved):

packages/hash/api/src/graph/system-types.ts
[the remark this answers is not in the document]
thehabbos007: Updated in 28594b72e318da5ab0fc71c798010669233aff89 also driveby of any other imports like it

Alfred-Mountfield (dismissed):

Alfred-Mountfield (approved):
🚀

PR #1798, opened 2023-01-10
Use wasm binary for type system instead of base64 version

nathggns:

This updates the type system loader in HASH to use the wasm binary alongside the slim package instead of the one compiled to base64. 

## 🔗 Related links

- Use the `/slim` entry-point of the Type System package [Asana task](https://app.asana.com/0/1201095311341924/1203492058388309/f) _(internal)_

## 🔍 What does this change?

- Uses a webpack/tsconfig alias to enforce use of the slim package
- Adds code to fetch/package the wasm binary (from @Alfred-Mountfield) and to pass it in when initialising the type system

N/A

## ⚠️ Known issues

There's an issue where webstorm doesn't understand the type system's "exports" field so we can't import the wasm that way, but importing it from its proper location also fails because webpack expects you to use the values in exports. So we use an alias to overwrite that.


packages/hash/frontend/next.config.js
```diff
@@ -52,33 +52,70 @@ process.env.NEXT_PUBLIC_SENTRY_REPLAY_SESSION_SAMPLE_RATE =
...
-          test: /\.svg$/,
-          use: [require.resolve("@svgr/webpack")],
-        });
+          // WebStorm doesn't understand exports so we have to use the full import path,
+          // but we can't because it's not in the exports so webpack complains
+          // This fixes that
+          // eslint-disable-next-line no-param-reassign
+          webpackConfig.resolve.alias[
+            "@blockprotocol/type-system/dist/wasm/type-system.wasm"
+          ] = require.resolve("@blockprotocol/type-system/type-system.wasm");
```
Alfred-Mountfield, line 113: This also feels uncomfortable to me. Are you sure it's necessary? 

I've just opened your branch in CLion (which I'd expect to be largely the same as WebStorm) and I'm not seeing an error when I delete this.

packages/hash/frontend/tsconfig.json
```diff
@@ -6,7 +6,12 @@
     "jsx": "preserve",
     "lib": ["dom", "dom.iterable", "esnext"],
     "target": "es5",
-    "noEmit": true
+    "noEmit": true,
+    "paths": {
+      "@blockprotocol/type-system": [
+        "../../../../node_modules/@blockprotocol/type-system/dist/cjs-slim/index-slim.d.ts"
+      ]
```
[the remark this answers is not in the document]
Alfred-Mountfield: I also didn't need to do it in [the example template](https://github.com/Alfred-Mountfield/wasm-sandbox/blob/main/nextjs-app/tsconfig.json) so I'm not sure why we need this here.

Deleting it doesn't seem to complain in my IDE (perhaps you're using a feature of the IDE I'm not?), maybe we should jump on a call and have a look?

packages/hash/frontend/next.config.js
Alfred-Mountfield: Actually I can recreate the error after rewriting the import. I'm not sure why this is happening, when it's not happening when I open the demo. I'm wondering if it's to do with the [custom.d.ts](https://github.com/Alfred-Mountfield/wasm-sandbox/blob/main/nextjs-app/custom.d.ts) or something else

Alfred-Mountfield: We need to do two things to fix the errors I believe. 
1. We need to add an exclusion for the type system here
https://github.com/hashintel/hash/blob/9723f5f012b0b75f8797a4a6966b1ad2f04095dd/libs/javascript/%40local/eslint-config/legacy-base-eslintrc-to-refactor.cjs#L65-L74

2. We need to add the custom declaration (@kachkaev recommends we do it in https://github.com/hashintel/hash/blob/main/packages/hash/frontend/assets.d.ts) like I have in the `custom.d.ts` linked

packages/hash/frontend/tsconfig.json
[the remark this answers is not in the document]
nathggns: Discussed on a call that this is so typescript knows to use the slim package's typings instead of the non-slim typings (which happen to be the same here, but this is for completeness/future proofing)

packages/hash/frontend/next.config.js
Alfred-Mountfield: Jumped on a call to pair on this, and pushed a commit

Alfred-Mountfield (approved):
Looks good! I like the idea of the alias so we don't have to remember `/slim`. If it ends up being problematic or too magical we should move to a lint rule instead like you suggested as an alternative :)

Issue #1782, opened 2023-01-08
Can't see crash output when using solarized-dark color scheme


dpc:
### Describe the bug

![Screenshot from 2023-01-07 23-48-27](https://user-images.githubusercontent.com/9209/211185860-22b210dd-f924-4447-988f-e05c9a4c109c.png)

I use solarized-dark color scheme in my terminals and noticed after upgrading that I can not see some parts of the output (until I highlight it).

### Expected behavior

Would be better if I was able to see colors. I think I've seen that color output is behind an optional cargo feature, so I'm going to try disabling them, and that's good enough workaround for me, I think.

BTW. Recently `clap` has made a decision to abandon colors and rely only on underline, bold font, etc. to make the output readable. 

The reasoning was that not only it's hard to find color scheme that works for everyone, but also - businesses and teams have sometimes their own branding etc. and error colors might get in a way.

indietyp:
Hey! Just as a side-note we also support the NO_COLOR environment variable. Setting NO_COLOR=1 will disable color output for error-stack.

TimDiekmann:
With #1800 it's now possible to choose the `ColorMode` and it will be available with error-stack 0.3, which is planned to be released today (maybe tomorrow, but we're targeting today).

PR #1800, opened 2023-01-11
`error-stack`: introduction of color modes

indietyp:

This implements the notion of color modes, determining if colors and which type of coloring/styling are requested.

~~This enables users to override the preferred mode while preserving the convenience of environment variables like `NO_COLOR` and `FORCE_COLOR`.~~ This has been moved into an example.

The available color modes are `Color`, `None`, and `Emphasis`. This mode is available for hooks so that they can easily adapt. A perfect example of this is the location hook, which now will use the `mode` to either print in bright black or italic.

~~If no color mode is set, the mode is automatically determined using `owo-colors` using `supports-colors` and the specified override.~~ If no color mode is set, the mode is set to `Color::Emphasis` if the `color` feature (opt-in) is enabled.

> **Note:** as a drive-by, `supports-colors` is now only enabled on `std`, `supports-color` implies `std` (not as a feature, found this by accident), which meant that no-std colors would fail, even tho they shouldn't.

## 🔗 Related links

* #1782

## 🔍 What does this change?

* `Report::format_color_mode_preference`
* `HookContext::mode()` is now available

Yes, the new `ColorMode` needs to be explained. This is already done.

We might want to add additional explanation/info in `lib.rs`

## 🏃 Next Steps

1. Ideally we'd temporarily set the mode from owo-colors, if possible, this needs https://github.com/jam1garner/owo-colors/pull/84

## 🛡 What tests cover this?

* Existing tests
* New snapshot tests

## 📹 Demo

image

image


indietyp:
discussed online with @TimDiekmann, rework in progress which will remove `pretty-print`, introduce a `Charset` and `ColorMode`

packages/libs/error-stack/tests/snapshots/doc/fmt__charset_ascii.snap
```diff
@@ -0,0 +1,9 @@
+<b>invalid input parameter</b>
+<b>|- </b><i>src/fmt/charset.rs:21:5</i>
```
TimDiekmann, line 2: I'm not sure but do we want to add the `"at "` here as well? If so, we could also remove the whitespace between `-` and `at` again I guess.

packages/libs/error-stack/tests/snapshots/doc/fmt__preference_color.snap
```diff
@@ -0,0 +1,9 @@
+<b>invalid input parameter</b>
+<span style='color:#a00'>├</span><span style='color:#a00'>╴</span><span style='color:#a00'> </span><span style='color:#555'>src/fmt/color.rs:24:5</span>
```
TimDiekmann, line 2: I have not noticed that before as the output is the same, but do we emit unneeded spans here, or is this due to the conversion to HTML-codes?

packages/libs/error-stack/src/fmt/charset.rs
```diff
@@ -0,0 +1,128 @@
...
+            .map_or(Self::Utf8, |charset| charset)
+    }
+}
+
+/// Value layout:
+/// `0x00`: `Charset::Ascii`
+/// `0x01`: `Charset::Utf8`
+///
+/// all others: unset/none
+struct AtomicOverride(AtomicU8);
```
TimDiekmann, line 35: `AtomicOverride` is very similar to `ColorPreference`, Is it possible to reuse the implementation?

TimDiekmann (commented):
Looks good to me, thanks @indietyp! 🎉 

I have a few comments which should be addressed before we merge it, but nothing, which is really blocking the merge.

packages/libs/error-stack/tests/snapshots/doc/fmt__preference_color.snap
indietyp: Great catch. After looking at the code, we actually emit the ANSI codes multiple times, while in reality, there's no real change. It is considered poor etiquette. Fixed in [8f5918b](https://github.com/hashintel/hash/pull/1800/commits/8f5918bfa5cec7b666cf544ed74ac8663a6ce85d)

packages/libs/error-stack/src/fmt/charset.rs
indietyp: fixed in [de12b05](https://github.com/hashintel/hash/pull/1800/commits/de12b05ff5dae1663f87c3009307f929b38355bf)

packages/libs/error-stack/tests/snapshots/doc/fmt__charset_ascii.snap
indietyp: 🤔 I am unsure. We could if that is the case, would it be "at <color>LOCATION</color>" where color is either color or emphasis or another way?

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Value layout:
+/// `0x00`: `ColorMode::None`
+/// `0x01`: `ColorMode::Color`
+/// `0x02`: `ColorMode::Emphasis`
+///
+/// all others: unset/none
+#[cfg(feature = "color")]
+impl AtomicPreference for ColorMode {
+    fn load(value: u8) -> Option<Self> {
```
TimDiekmann, line 53: The behavior using `Option` feels weird to me. We either have `None` or `Some(None)`. I'd prefer to always return a value, if none was set before we should just return our default value.

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub(super) fn load() -> Self {
+        // The default is `ColorMode::Emphasis`, as colors are hard. ANSI colors are not
+        // standardized, and some colors may not show at all.
+        COLOR_OVERRIDE.load().map_or(Self::Emphasis, |mode| mode)
+    }
+
+    #[cfg(not(feature = "color"))]
+    pub(super) const fn load() -> Self {
+        ColorMode::None
+    }
```
TimDiekmann, line 42: Does this mean we cannot store `ColorMode` when the `"color"` feature is not enabled?

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/// Value layout:
+/// `0x00`: `ColorMode::None`
+/// `0x01`: `ColorMode::Color`
+/// `0x02`: `ColorMode::Emphasis`
+///
+/// all others: unset/none
+#[cfg(feature = "color")]
```
TimDiekmann, line 51: Why this requires a feature?

packages/libs/error-stack/src/fmt/color.rs
TimDiekmann, line 53: Why do we have two implementations of `load` with different semantics?

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// </pre>
+    ///
+    /// <pre>
+    #[doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/tests/snapshots/doc/fmt__preference_emphasis.snap"))]
+    /// </pre>
+    ///
+    /// <pre>
+    #[doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/tests/snapshots/doc/fmt__preference_color.snap"))]
+    /// </pre>
+    #[cfg(feature = "color")]
```
TimDiekmann, line 149: Why this is gated behind a feature flag? I expect that downstream binaries want to set `ColorMode::None` or `ColorMode::Emphasis` and downstream libraries can pick up those values, this has nothing to do on our dependency on `owo-colors`.

packages/libs/error-stack/src/lib.rs
```diff
@@ -464,6 +470,7 @@
 )]
 
 extern crate alloc;
+extern crate core;
```
TimDiekmann, line 473:
```suggestion
```

TimDiekmann (changes requested):
I found an issue regarding the feature flags, the behavior for color support should be similar to `charset` and the `color` feature enables the usage of colored output e.g. for the location.

packages/libs/error-stack/src/fmt/color.rs
indietyp: Yes, I tied the `color` feature to the ability to enable color, but after thinking about it, that's a mistake. We should not gate that behavior.

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [f83228f](https://github.com/hashintel/hash/pull/1800/commits/f83228f7d0c91828e09c642bd916119794d17402)

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [f83228f](https://github.com/hashintel/hash/pull/1800/commits/f83228f7d0c91828e09c642bd916119794d17402)

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [f83228f](https://github.com/hashintel/hash/pull/1800/commits/f83228f7d0c91828e09c642bd916119794d17402)

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [f83228f](https://github.com/hashintel/hash/pull/1800/commits/f83228f7d0c91828e09c642bd916119794d17402)

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [f83228f](https://github.com/hashintel/hash/pull/1800/commits/f83228f7d0c91828e09c642bd916119794d17402)

packages/libs/error-stack/src/fmt/charset.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub(super) fn load() -> Self {
+        CHARSET_OVERRIDE.load()
+    }
+}
+
+impl Default for Charset {
+    fn default() -> Self {
+        Self::DEFAULT
+    }
+}
```
TimDiekmann: Can we `derive` it instead?

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn default() -> Self {
+        Self::Emphasis
+    }
+
+    #[cfg(not(feature = "color"))]
+    fn default() -> Self {
+        Self::None
+    }
+}
```
TimDiekmann: This feels wrong, especially when in mind, that we are going to remove the `color` feature in a follow-up. Can we just derive `Default` and set it to `Emphasis`?

packages/libs/error-stack/src/fmt/charset.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::{
+    fmt::r#override::{AtomicOverride, AtomicPreference},
+    Report,
+};
+
+/// The available supported charsets
+///
+/// Can be accessed through [`crate::fmt::HookContext::charset`], and set via
+/// [`Report::set_charset`].
+#[derive(Debug, Copy, Clone)]
```
TimDiekmann, line 10: We should derive `PartialEq`, `Eq`, and `Hash` as well

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -0,0 +1,161 @@
+use crate::{
+    fmt::r#override::{AtomicOverride, AtomicPreference},
+    Report,
+};
+
+/// The available modes of color support
+///
+/// Can be accessed through [`crate::fmt::HookContext::color_mode`], and set via
+/// [`Report::set_color_mode`]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
```
TimDiekmann, line 10: We should derive `Hash` as well

indietyp: fixed in [a727237](https://github.com/hashintel/hash/pull/1800/commits/a727237a55d6a1fdcad4abc805e546800833dd14)

packages/libs/error-stack/src/fmt/charset.rs
indietyp: fixed in [a727237](https://github.com/hashintel/hash/pull/1800/commits/a727237a55d6a1fdcad4abc805e546800833dd14)

packages/libs/error-stack/src/fmt/color.rs
indietyp: fixed in [a727237](https://github.com/hashintel/hash/pull/1800/commits/a727237a55d6a1fdcad4abc805e546800833dd14)

packages/libs/error-stack/src/fmt/charset.rs
indietyp: fixed in [a727237](https://github.com/hashintel/hash/pull/1800/commits/a727237a55d6a1fdcad4abc805e546800833dd14)

packages/libs/error-stack/CHANGELOG.md
```diff
@@ -11,13 +11,16 @@ All notable changes to `error-stack` will be documented in this file.
...
 
 ### Features
 
-- Add serializing support using [`serde`](https://serde.rs) ([#1290](https://github.com/hashintel/hash/pull/1290))
+- Add initial serializing support using [`serde`](https://serde.rs) ([#1290](https://github.com/hashintel/hash/pull/1290))
 - Support `Debug` hooks on `no-std` platforms via the `hooks` feature ([#1556](https://github.com/hashintel/hash/pull/1556))
 - Support converting `Report` into [`Error`](https://doc.rust-lang.org/core/error/trait.Error.html) via `Report::as_error` and `Report::into_error` ([#1749](https://github.com/hashintel/hash/pull/1749))
 - Support converting `Report` into `Box<dyn Error>` via the `From` trait ([#1749](https://github.com/hashintel/hash/pull/1749))
+- Programmatic selection of color mode and charset used for `Debug` output ([#1800](https://github.com/hashintel/hash/pull/1800))
+- New feature `color`, supersedes `pretty-print` feature ([#1800](https://github.com/hashintel/hash/pull/1800))
```
TimDiekmann, line 23: Just as a reminder, that we have to tweak this entry when removing the feature again 🙂

TimDiekmann (dismissed):
Thanks @indietyp, a nice, concise interface! 🙂 

Too bad, that `rustdoc` does not render `#[default]`.

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #1801, opened 2023-01-11
Use time projection kernel in compiled queries

TimDiekmann:

The time projection is passed to the query compiler since #1743. Up until now we defaulted to decision time as the pinned axis and the transaction time as the variable axis. With this PR, this can be changed. The kernel will be correctly picked up in compiled queries and the output is adjusted.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203701389454309/f) _(internal)_

## 🚫 Blocked by

- #1743 

## 🔍 What does this change?

- Rename `LowerTransactionTime` to `ProjectedTime`
- Adjust the compiler to store information about temporal tables, which have been pinned to avoid duplicated statements or parameters
- Adjust test outputs to check for the correct output

## 🐾 Next steps

- [Use time projection image in structural queries](https://app.asana.com/0/1203543021352032/1203701389454310/f) _(internal)_

## 🛡 What tests cover this?

All tests, which have been adjusted


packages/graph/hash_graph/lib/graph/src/store/postgres/query/condition.rs
```diff
@@ -77,10 +77,11 @@ impl Transpile for Condition<'_> {
                 fmt.write_str(" != ")?;
                 rhs.transpile(fmt)
             }
-            Condition::RangeContains(lhs, rhs) => {
+            Condition::TimerangeContainsTimestamp(lhs, rhs) => {
                 lhs.transpile(fmt)?;
                 fmt.write_str(" @> ")?;
-                rhs.transpile(fmt)
+                rhs.transpile(fmt)?;
+                fmt.write_str("::TIMESTAMPTZ")
```
TimDiekmann, line 84: Somehow, `tokio_postgres` tries to serialize a `TSTZRANGE` instead of `TIMESTAMPTZ`

thehabbos007 (commented):
Hey Tim, I have some comments here. It may be good for Alfie to take a look as well

packages/graph/hash_graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -87,6 +89,37 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
...
+        });
+
+        if !temporal_table_info.tables.contains(&table) {
+            self.statement
+                .where_expression
+                .add_condition(Condition::TimerangeContainsTimestamp(
+                    Expression::Column(
+                        Column::Entities(match self.time_projection.time_axis() {
+                            TimeAxis::DecisionTime => Entities::TransactionTime,
+                            TimeAxis::TransactionTime => Entities::DecisionTime,
```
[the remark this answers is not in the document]
TimDiekmann: Yes, this is correct. I added a small comment to clarify it: 9678ef6245bfa05f653e682c3a1d9739af6abe55

TimDiekmann:
Thanks, @thehabbos007! I responded to your comments and requested another review from @Alfred-Mountfield.

packages/graph/hash_graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -87,6 +89,39 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
...
+                parameter_index: self.artifacts.parameters.len(),
+            }
+        });
+
+        if !temporal_table_info.tables.contains(&table) {
+            // Adds the kernel timestamp condition, so for the projected decision time, we use the
+            // transaction time and vice versa.
+            self.statement
+                .where_expression
+                .add_condition(Condition::TimerangeContainsTimestamp(
```
Alfred-Mountfield, line 111: This might have been in another PR but can we change this to `TimeRange` not `Timerange` please? (Can be done in follow-up).

"timespan" is one word, but "timerange" is not.

packages/graph/hash_graph/lib/graph/src/store/postgres/query/table.rs
```diff
@@ -131,9 +131,9 @@ pub enum Entities<'p> {
     RecordId,
     DecisionTime,
     TransactionTime,
-    // TODO: Remove when adjusting structural queries
-    //   see https://app.asana.com/0/0/1203491211535116/f
-    LowerTransactionTime,
+    // TODO: Remove when correctly resolving time intervals in subgraphs.
+    //   see https://app.asana.com/0/0/1203701389454316/f
+    ProjectedTime,
```
Alfred-Mountfield, line 136: Does the documentation for "projected time" exist anywhere else? I don't think the name is very intuitive, I get the choice of "projection" from a mathematical point of view but I feel like we could improve accessibility of the code by going for things like "pinned time axis" and such.

At the very least I'd want these terms explained somewhere (and I can't find that at the moment)

TimDiekmann: There is a short documentation in `shared/identifier/time/axis.rs`:
```rust
/// Time axis for the projected image time.
///
/// This is used as the generic argument to time-related structs.
#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Hash)]
pub struct ProjectedTime;
```

Alfred-Mountfield: Gotcha, can we try and explain it a *lot* more. Explaining "projected time" by saying it's "the axis for the projected image time" isn't reallyyy an explanation

packages/graph/hash_graph/lib/graph/src/store/postgres/query/compile.rs
TimDiekmann: Addressed in 070bd2846de71d66d2cbefe04e7cf999e6478d3d

packages/graph/hash_graph/lib/graph/src/store/postgres/query/table.rs
TimDiekmann: I added a more in-depth documentation in 3ddd19f27775101a499c922b45080ece2489a687 describing the `TimeProjection` and it's axes. I linked the `ProjectedTime` axis documentation to that documentation.

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/axis.rs
```diff
@@ -27,8 +27,12 @@ pub enum TimeAxis {
     TransactionTime,
 }
 
-/// Time axis for the projected image time.
+/// Time axis for the [`Image`] time axis used in [`TimeProjection`]s.
```
thehabbos007, line 30: This comments read slightly strange to me, perhaps just this?
```suggestion
/// Time axis for the [`Image`] used in [`TimeProjection`]s.
```

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/projection.rs
```diff
@@ -218,6 +218,9 @@ impl ToSchema for UnresolvedTimeProjection {
     }
 }
 
+/// The pinned axis of a [`TimeProjection`].
```
thehabbos007, line 221: If we have the term `pinned axis` for `kernel`, why do we not call it `PinnedAxis` instead? Not sure if this was brought up before :)

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/projection.rs
```diff
@@ -253,6 +256,9 @@ impl ToSchema for TransactionTimeKernel {
     }
 }
 
+/// The variable time of a [`TimeProjection`].
```
thehabbos007, line 259: Same as above but for `VariableTime`. These terms seem to be a lot more natural within our context

Alfred-Mountfield (approved):

PR #1803, opened 2023-01-11
Move `hash-realtime` and `hash-search-loader` workspaces to `apps`

kachkaev:

This PR continues #1795 and moves two workspaces with apps into `apps`. It also changes their workspace names to `@apps/hash-*`. More workspaces will be moved in follow-up PRs.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203608234995336/f) _(internal)_

No


thehabbos007 (approved):
Looks good to me and realtime runs locally.
The docker images build/run for me (at least realtime runs, search-loader is outdated because of other reasons)

We aren't deploying realtime in prod, so we don't have to worry about that part. Thank you Alex! :sparkles:

nonparibus (changes requested):
There should be AGPL license files in each of these repositories.

nonparibus (approved):
Directories now correctly include the appropriate license files.

PR #1787, opened 2023-01-09
Introduce HASH Status Package and Model

Alfred-Mountfield:
## 🌟 What is the purpose of this PR?

This PR introduces the foundations of an Error and Status Model to be used across our services. Most of the motivation is captured within the README.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432087/1203639876018733/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432094/1203639884730777/f) _(internal)_
- [Notion Doc](https://www.notion.so/hashintel/The-HASH-Status-and-Error-Model-4d726d7f03624fd1b5a96590452cabf9)_(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

See README and commits please, it's all new code.

## 📜 Does this require a change to the docs?

- We should probably update the top-level README or CONTRIBUTING docs to point to the existence of this model.
- Usage instructions will need to be supplied, this is done in a follow up (#1810)

## ⚠️ Known issues
- [ ] CI needs updating to check the new package, especially for the Rust portion
- [x] Clippy lints need disabling for generated files (unsure of the best way to do this at the moment)
 - This is addressed in the follow-up as we generate the files into the `target` directory and they aren't checked via clippy.
- The codegen script isn't currently as re-usable or portable as we'd like. @kachkaev has some thoughts on how we can hoist it into its own `@local` package, and establish a pattern for all packages (such as unifying all codegen steps inside a `codegen.ts` file in each package that needs code generation) 
- The generator struggles with complex types, or files that aren't just plain type definitions. This is why definitions are in their own folder outside of `src`, for the most part they should be treated like static schema files that just happen to share the syntax of TypeScript

## 🐾 Next steps

- Demonstrate usage of the library in the HASH Graph API (#1810) 
- Add usage instructions (#1810)
- Migrate the whole of the HASH Graph API to be compliant with the model
- Migrate workspace backend code to parse and consume the model payloads
- Establish a pattern for the error model over the GraphQL API

## 🛡 What tests cover this?

- None at present beside lints and similar checks

## ❓ How to test this?

1. Checkout the branch
2. Check that `yarn build` successfully generates code in the crate
3. Check that the crate can build
4. Look at #1810 for more confidence

## 📹 Demo

N/A

PR #1810, opened 2023-01-12
Use the HASH Status lib in the Graph

Alfred-Mountfield:

This PR follows on from #1787 to provide a demonstration of how to use the model across various services. It does so by 
- changing an endpoint on the HASH Graph to return the `Status` type, including the work involved in 
  - setting up the service to import from the status package
  - defining the payloads definition for the service
  - setting up codegen to create types for the payloads
  - configuring Turborepo to define the dependency graph
  - extending the core definitions of payloads with a new one
- parsing the `Status` objects on the boundary of the HASH Graph and the HASH API service.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203639884730779/f) _(internal)_

## 🚫 Blocked by

- [x] #1787 

## 🔍 What does this change?

- See commits, effort has been made to keep them self explanatory

- We may need to improve docs around building of packages, and add some follow-up tasks to track the necessary following work

## ⚠️ Known issues

- Some quirks around `IntoResponse`, see the comment opened on the diff
- ~~Not expressing the bodies in the OpenAPI spec, see comment opened on the diff~~

## 🐾 Next steps

- Migrate the whole of the HASH Graph API to be compliant with the model
- Establish a pattern for _using_ and propagating the `Status` objects in the API service code
- Establish a pattern for the error model over the GraphQL API

## 🛡 What tests cover this?

- The normal suite of tests (see CI), I have not extended the HTTP tests as part of this work

## ❓ How to test this?

Check as usual, check compilation, CI, etc. Also using postman or similar you should call the modified endpoint to check the response body now contains the new object.

## 📹 Demo

image


packages/graph/hash_graph/lib/graph/src/api/error/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum StatusPayloads {
+    ErrorInfo(ErrorInfo),
+    RequestInfo(RequestInfo),
+    ResourceInfo(ResourceInfo),
+}
+
+impl ErrorInfo {
+    pub fn new(metadata: HashMap<String, Option<serde_json::Value>>, reason: String) -> Self {
```
TimDiekmann, line 16: Will the reason ever be a non-static string? From the code below it seems it could be changed to `&'static str`

Alfred-Mountfield: The `ErrorInfo` type is generated with `String`, we don't control that

libs/status/README.md
```diff
@@ -17,7 +17,63 @@ A consistent structure and pattern for producing API-agnostic definitions of sta
...
+- The propagation of errors purely _within_ a service do not have to conform to the shape of the model defined here. This is a compromise to attempt to maintain idiomatic behaviour within various environments. Therefore `error-stack` can continue to be used consistently in Rust, and TS can throw errors that will be caught in a tight scope when required. However, once a path allows for that error to be propagated through an API, type systems should be used to ensure that all errors in there are defined in a publicly consumable fashion.
+
+## Usage
+
+### Importing the dependency
+
+This package is exposed as a Rust crate and as an NPM package. Services in those languages should add the respective
```
nonparibus, line 26:
```suggestion
This package is exposed as a Rust crate and as an npm package. Services in those languages should add the respective
```

(`npm` not `NPM`: https://docs.npmjs.com/about-npm)

apps/hash-api/package.json
```diff
@@ -14,6 +14,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@apps/hash-graph": "0.0.0-private",
```
TimDiekmann, line 17: I guess this is required to have a correct dependency graph?

apps/hash-graph/docker/Dockerfile
```diff
@@ -1,20 +1,42 @@
...
+COPY yarn.lock .
+
+COPY ./libs/@local/status/. /usr/local/libs/@local/status
+
+COPY libs/@local/eslint-config libs/@local/eslint-config
+COPY libs/@local/tsconfig libs/@local/tsconfig
+
+RUN yarn workspace @local/status install --frozen-lockfile --ignore-scripts --prefer-offline
+
+FROM node_base AS builder
```
TimDiekmann, line 31: Having two blank lines makes it easier to find new layers
```suggestion

FROM node_base AS builder
```

apps/hash-graph/lib/graph/src/api/gen.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(dead_code)]
+
+pub mod status_payloads {
+    include!(concat!(env!("OUT_DIR"), "/status_payloads.rs"));
+}
```
TimDiekmann, line 5: Is there a reason you don't add the `status.rs` module as well?

apps/hash-graph/lib/graph/src/api/error.rs
```diff
@@ -0,0 +1,25 @@
...
+use serde::{Deserialize, Serialize};
+
+pub(super) use crate::api::gen::status_payloads::{ErrorInfo, RequestInfo, ResourceInfo};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum StatusPayloads {
+    ErrorInfo(ErrorInfo),
+    RequestInfo(RequestInfo),
+    ResourceInfo(ResourceInfo),
+}
```
TimDiekmann, line 13: Shouldn't these be generated?

apps/hash-graph/lib/graph/src/api/error.rs
TimDiekmann, line 1: This file generally seems as if this should be generated entirely

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -93,29 +102,75 @@ async fn create_entity_type<P: StorePool + Send>(
...
+                HashMap::from([(
+                    "validationError".to_owned(),
+                    serde_json::to_value(err)
+                        .expect("Could not serialize entity type validation error"),
+                )]),
+                // TODO: We should encapsulate these Reasons within the type system, perhaps
+                //  requiring top level contexts to implement a trait `ErrorReason::to_reason`
+                //  or perhaps as a big enum, or as an attachment
+                "INVALID_SCHEMA".to_owned(),
+            ))],
```
TimDiekmann, line 129: I guess these should be provided by `error-stack` at some point?

apps/hash-api/package.json
Alfred-Mountfield: Yes indeed

apps/hash-graph/docker/Dockerfile
```diff
@@ -1,20 +1,42 @@
...
 
-COPY rust-toolchain.toml .
+COPY apps/hash-graph/rust-toolchain.toml .
 RUN apk add --no-cache gcc musl-dev bash && \
     wget -q -O- https://sh.rustup.rs | sh -s -- -y --default-toolchain none --profile minimal && \
     rustup show
 
 SHELL ["bash", "-c"]
 
+FROM rust as node_base
```
Alfred-Mountfield, line 16:
```suggestion

FROM rust as node_base
```

apps/hash-graph/docker/Dockerfile
Alfred-Mountfield:
```suggestion

FROM node_base AS builder
```

apps/hash-graph/lib/graph/Cargo.toml
```diff
@@ -11,12 +11,15 @@ type-fetcher = { path = "../type-fetcher", optional = true }
 async-trait = "0.1.66"
 axum = "0.6.11"
 bb8-postgres = "0.8.1"
+bytes = "1.4.0"
 clap = { version = "4.1.9", features = ["derive", "env"], optional = true }
 time = { version = "0.3.20", features = ['serde', 'serde-well-known'] }
 derivative = "2.2.0"
 error-stack = { version = "0.3.1", features = ["spantrace"] }
 futures = "0.3.27"
+hash-status = { path = "../../../../libs/@local/status/crate" }
```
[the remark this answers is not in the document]
Alfred-Mountfield: We're going to try and rely on Turbo's dependency graph to decide what CI checks to run for Node things. I think we should commit to that as a repo-level solution and rely on it for Rust as well so we can remove our bespoke logic

apps/hash-graph/lib/graph/build.rs
```diff
@@ -0,0 +1,54 @@
...
+            // Uncomment this to generate JSON Schema definitions in the `OUT_DIR` as well
+            // "--json-schema-out-dir",
+            // &out_dir,
+        ])
+        .current_dir(virtual_manifest_dir)
+        .status()
+        .expect("Failed to run codegen")
+        .success()
+    {
+        panic!("Failed to run codegen");
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yes

apps/hash-graph/lib/graph/src/api/error.rs
```diff
@@ -0,0 +1,25 @@
+use std::collections::HashMap;
+
+use hash_status::Status as HashStatus;
+use serde::{Deserialize, Serialize};
+
+pub(super) use crate::api::gen::status_payloads::{ErrorInfo, RequestInfo, ResourceInfo};
```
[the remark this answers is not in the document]
Alfred-Mountfield: I believe it is required, yes
```suggestion
pub(in crate::api) use crate::api::gen::status_payloads::{ErrorInfo, RequestInfo, ResourceInfo};
```

Would you prefer this?

apps/hash-graph/lib/graph/src/api/error.rs
Alfred-Mountfield: I think the enum should be generated but it doesn't preserve the actual previous definitions in my experience. It will instead create new duplicative definitions for each variant like we've experienced with Utoipa before.

There are potentially ways around this but it might involve upgrading the `quicktype` version (which we can't do right now because it requires an upgraded `node`) and it might involve swapping to JSON schema definitions instead of TS ones

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
Alfred-Mountfield: I think that'd be good, yes

apps/hash-graph/lib/graph/Cargo.toml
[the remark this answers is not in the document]
TimDiekmann: Yes, this is possible, but this will greatly slow down CI times. Given this example, the graph depends on `status`, but also, let's say, 5 other packages depend on `status`, all 6 packages need to be tested. While Turbo is able to test all of those packages, every test would run in the same GitHub Runner. Unless it's possible to let Turbo generate a `matrix` object (like the current `setup,py`), all of our test CI jobs would run in a single, gigantic job and it's impossible to re-run only a single, failed job. Especially for the graph (which has currently one of the longest-running CIs for testing), this implies compiling it for different configurations with different targets multiple times (e.g. debug, release, test, binaries, integration, docker), which are currently in various jobs, which run parallel on different runners. 

If we can use turbo to create a list of packages, which needs to be tested, we can use that as our entry point for the CI and I'm happy to do some discovery here, so this could almost entirely replace the current `setup.py` job.

apps/hash-graph/lib/graph/src/api/gen.rs
TimDiekmann: So I assume if the Graph would _consume_ errors, it would be useful?

Alfred-Mountfield: No, I think it would never be useful for Rust. We could add an `exclude` pattern for the Rust generation but that is more work than is worth it at this stage considering the files are harmless.

The local `status` package has a crate with a natively implemented nicer generic `Status` type (which is what we import and use), and I think that would always be better than the generated one. We'd just need to define an alias of it which fills in the generic with the appropriate imported payload enum of whatever service is being _consumed_.

TimDiekmann (commented):
As this PR adds `yarn` to the `Rust` CI, additional parameters have to be set in that workflow. https://github.com/hashintel/hash/pull/2253 does these changes, so this PR is currently blocked on that PR.

Update: PR is approved and queued ✅

turbo.json
```diff
@@ -18,6 +17,11 @@
     },
     "lint:tsc": {
       "outputs": []
+    },
+    "@apps/hash-graph#build": {
+      "cache": false,
+      "dependsOn": ["^build"],
+      "outputs": ["target/**"]
```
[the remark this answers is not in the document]
Alfred-Mountfield: Because it tells Turbo what the side-effects of running build is. It's not telling other packages about `target`.

turbo.json
```diff
@@ -18,6 +17,11 @@
     },
     "lint:tsc": {
       "outputs": []
+    },
+    "@apps/hash-graph#build": {
+      "cache": false,
+      "dependsOn": ["^build"],
```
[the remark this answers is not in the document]
Alfred-Mountfield: The simplest answer is that currently `^build` does not need to be specified, as the `status` package does not have a build step. If we added a bundling step, it would need to be there (and we'd need to think about how the `build.rs` works at that point as we wouldn't want to directly call `codegen.ts`)

The longer answer is that we're using Turborepo, and we should be appropriately modelling the dependency graph. The `hash-graph` application has local dependencies (`status`), and the build step of the `hash-graph` depends on those dependencies (`status`).

Not appropriately modelling these things is what causes unwanted effects and is why we are currently finding ourselves writing `TURBO_FORCE` so often.

This should not affect graph development for the most part (or at all right now), so I wouldn't worry about it.

It may also be helpful when it comes to CI later on.

apps/hash-graph/README.md
```diff
@@ -59,6 +59,39 @@ Every command line argument passed will also be forwarded to the subcommand, e.g
...
+The spec is mostly generated from the code using [`utoipa`](https://github.com/juhaku/utoipa/), although some of the more complex types are specified manually within [`lib/graph/src/api/rest/json_schemas`](lib/graph/src/api/rest/json_schemas).
+
+As such, when modifying return values related to these types, it's important to update the accompanying schemas.
+
+#### Status
+
+Responses containing non-OK statuses are returned according to the `Status` format defined in the [`@local/status`](/libs/@local/status/README.md) package.
+
+The [`status.json`](lib/graph/src/api/rest/json_schemas/status.json) schema is responsible for the definition of this type, and should be updated whenever new payloads are added within [`./type-defs`](./type-defs).
+JSON schema definitions can be generated within the build directory by uncommenting the line in the lib's [`build.rs`](./lib/graph/build.rs).
```
thehabbos007, line 87: Are we being vague about what line to uncomment here so we don't need to keep the docs in sync?

TimDiekmann (commented):
PR looks good to me, @thehabbos007 will give a final approval ✅

turbo.json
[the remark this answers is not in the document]
TimDiekmann: Thanks for the link, makes sense to me!

thehabbos007 (commented):
Changes look good to me. Just have one comment here

apps/hash-graph/README.md
Alfred-Mountfield: Pretty much, yes 😅

packages/graph/hash_graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -92,7 +99,9 @@ async fn create_entity_type<P: StorePool + Send>(
     pool: Extension<Arc<P>>,
     domain_validator: Extension<DomainValidator>,
     body: Json<CreateEntityTypeRequest>,
-) -> Result<Json<OntologyElementMetadata>, StatusCode> {
+    // TODO: We want to be able to return `Status` here we should try and create a general way to
+    //  call `status_to_response` for our routes that return Status
+) -> Result<Json<OntologyElementMetadata>, impl IntoResponse> {
```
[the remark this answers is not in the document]
thehabbos007: I haven't dealt with many of Axum's response type expectations, so I'm not sure what alternatives there are other than what you've outlined. What would we add to the `post(create_entity_type::<P>)` path to do this mapping? A handler that somehow calls the mapping function?

thehabbos007 (approved):

thehabbos007 (approved):

TimDiekmann (approved):
Happy if CI is happy, otherwise it would fail anyway 🙃

packages/graph/hash_graph/lib/graph/src/api/rest/entity_type.rs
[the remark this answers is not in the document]
Alfred-Mountfield: > A handler that somehow calls the mapping function?

That's my thinking, if we make every method return `Result<OkVal, Status>` and then wrap those methods with `map_err(status_into_response)`. But not entirely sure how to wrap them cleanly, it feels like it'd be a mess of generics and might screw up the proc macros. 

We could potentially do it with a request/response interceptor (or middleware, can't remember if they're the same thing) at the axum level

TimDiekmann (changes requested):
The CI error is reproducible as it occurred in all previous CI runs. This needs to be fixed first.

TimDiekmann:
I reverted my commits and moved to https://github.com/hashintel/hash/pull/2272

TimDiekmann (approved):

PR #1811, opened 2023-01-12
Rename `Span` to `Interval`

TimDiekmann:

To be consistent in the terms, we decided to rename `Span` to `Interval` and `VersionTimespan` to `VersionInterval`

## 🚫 Blocked by

- #1801 
- #1802


Alfred-Mountfield (commented):
I'm happy with this but will approve once the base is in to avoid polluting its diff

Alfred-Mountfield (approved):

PR #1815, opened 2023-01-12
Display object property types in the entity type editor

luisbettencourt:

This ticked implements displaying object property types in the properties table of the entity type editor.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203198755412051/f) _(internal)_

## 🔍 What does this change?

- Added an expected value chip to object property types with the nested number of properties and a tooltip displaying the names of the properties;
- Created a generic `PropertyRow` component that is used by `PropertyTypeRow` (for property types with depth = 0), and by itself (for property types with depth != 0);
- Width of table columns is now set on the `TableBody` level, this is because nested rows require a table at their level, and this makes it so the tables have the same column width.

## 📹 Demo

demo


luisbettencourt:
@nonparibus can you please check if the tooltips' text is okay?

nonparibus (changes requested):
Rather than say "Can't edit this on the X level", can we say "Can only be edited from Y"?

This provides users with a bit more of a directional hint about how to get themselves unstuck. 

Otherwise this is great. Thank you!

luisbettencourt:
Have changed the tooltip:

image

nonparibus:
I think this is still confusing.

1. we should refrain from referring to properties as "props" in end user-facing copy.
1. when I read this now I think "this is the nested property?! where else am I meant to go?" -- we need to refer to the intended edit path by another name. cc @CiaranMn @nathggns for suggestions.

CiaranMn:
> we should refrain from referring to properties as "props" in end user-facing copy.
> when I read this now I think "this is the nested property?! where else am I meant to go?" -- we need to refer to the intended edit path by another name. cc @CiaranMn @nathggns for suggestions.

@nonparibus  "nested props" isn't copy but the name of the property. So for other properties it might say things like "This can only be edited from address instead"

I would say something like "Edit the 'address' property to change this". Optionally you could add 'by hovering over it and clicking the '...' button' or something else to explain how to do that

luisbettencourt:
Updated the tooltip as suggested by @CiaranMn. Did not add further instruction ('by hovering over it and clicking the '...' button'), as I'm unsure how that should read. Most property types won't be editable on the entity type page, because they were created outside of it.

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
```diff
@@ -40,15 +57,282 @@ import {
...
+            mb: `-${lineHeight}px`,
+          }}
+          appear
+        >
+          <Box
+            sx={{
+              position: "absolute",
+              height: lineHeight,
+              width: "1px",
+              left: `${13.4 + 20 * depth}px`,
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:99]: Is there anything we can do to make this less magical? What is it related to?

CiaranMn (approved):
I haven't been able to test this as creating object property types isn't working at the moment, but let's get it in and can tweak in follow up as needed. Demo looks good

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Got rid of it in f5073edca4b96d5972f46218d39365e4c9afe26d

Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: This is still magic to me – it appears in two components with identical values, but they aren't connected. 

Is it a problem if they become out of sync? If so, we should have some function to generate it rather than duplicate it.

Also, can `10 * 20` be derived from some other value we have access to? Or is it just to match the design?

This isn't a big deal but if there's an easy way to ensure we don't update something and this breaks, let's do it – whether 'something' is one of these components being updated without the other one, or some other value somewhere that affects this

Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Thanks for the explanation, that makes sense. I think I fixed it in 69a24d5ccd38aa6d925d38e0cd5a44158d2504ea.

CiaranMn (approved):

PR #1816, opened 2023-01-12
Adjust backend to use time projections over `"latest"` for entities

TimDiekmann:

With temporal versioning in place in the Graph, we can remove the `version` parameter for entity queries and use the `TimeProjection` instead.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203720600730930/f) _(internal)_

## 🚫 Blocked by

- #1801 
- #1802 
- #1811 
- ~~[Correctly resolve time intervals when traversing the Graph](https://app.asana.com/0/1203543021352032/1203701389454316/f) _(internal)_~~ I expected this to be a blocker but it's not

## 🔍 What does this change?

- Make the `timeProjection` parameter required in structural queries. This also requires a `timeProjection` parameter in ontology queries. I always set those to `[now(), now()]` as this will have no impact on querying the ontology types, however, if we chose to support reversed edges, this will be the most conservative approach. If more data is needed later on, this can still be changed at that point
- Remove the `version` parameter in entity query paths. This now has to be specified in the `timeProjection`
- Adjust the backend to take these changes into account
- Remove functions in the Graph, which were rely on the present of the `version` parameter

The documentation on `TimeProjection` were added as part of #1801

## 🐾 Next steps

We currently still have the `Default` implementation on `UnresolvedTimeProjection`, which should be removed. Currently, we use those in tests, so I kept the implementation for now

## 🛡 What tests cover this?

Basically, all tests are covering this. I also adjusted the Graph Integration tests to more precisely query for versioned entities.

## ❓ How to test this?

Run the App and check, that everything is working. I did this locally as well


packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
```diff
@@ -129,161 +128,97 @@ impl<'p> Filter<'p, Entity> {
         ])
     }
 
-    /// Creates a `Filter` to search for a specific entity edition, identified by its
-    /// [`EntityVertexId`].
+    /// Creates a `Filter` to search for linked entities where the specified [`EntityId`] identifies
```
thehabbos007, line 131: Should we make it clear that it's the `outgoing linked entities`?
```suggestion
    /// Creates a `Filter` to search for outgoing linked entities where the specified [`EntityId`] identifies
```
(breaks rustfmt)

packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
```diff
@@ -129,161 +128,97 @@ impl<'p> Filter<'p, Entity> {
...
-                Some(FilterExpression::Parameter(Parameter::Timestamp(
-                    vertex_id.version(),
+                    entity_id.entity_uuid().as_uuid(),
                 ))),
             ),
         ])
     }
 
-    /// TODO
+    /// Creates a `Filter` to search for linked entities where the specified [`EntityId`] identifies
```
thehabbos007, line 155: Like above, probably more clear to make it `incoming linked entities`

packages/graph/clients/typescript/api.ts
```diff
@@ -635,7 +634,7 @@ export interface EntityStructuralQuery {
    * @type {UnresolvedTimeProjection}
    * @memberof EntityStructuralQuery
    */
-  timeProjection?: UnresolvedTimeProjection;
+  timeProjection: UnresolvedTimeProjection;
```
thehabbos007, line 637: We talked offline about changing the `UnresolvedTimeProjection` type to make the undefined axis timestamp value a `null` value instead. Just writing it here as well.
We could do this by making the `Timestamp` object nullable, but that really isn't ideal. Perhaps the other solution is to somehow make use of an enum for either the `Timestamp` object or null

thehabbos007 (approved):
Looks good to me and works locally. The changes are quite minimal and mostly touches the tests :+1:

packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
TimDiekmann: 75bc0acc3b6894c1bf282105e1b19bade510c748

packages/graph/hash_graph/lib/graph/src/store/query/filter.rs
TimDiekmann: 75bc0acc3b6894c1bf282105e1b19bade510c748

thehabbos007 (approved):
Thank you, as said it has been tested with login, signup, pages, type editor and entity editor routes and it does seem like there has been zero changes

TimDiekmann:
>  it does seem like there has been zero changes

I consider this as a good thing 😆

PR #1817, opened 2023-01-12
Move `engine` to `apps`

kachkaev:

This PR continues migrating the contents of `packages` into `apps` and `libs`. It moves `packages/engine` to `apps/engine` and cleans things up a little.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203324923411371/f) _(internal)_

## 🔍 What does this change?

- Files are moved and paths are updated accordingly

- `@hashengine/apache-arrow-bundle` is no longer a workspace in a top-level yarn project. This matches the way we treat `hash_stdlib` in `apps/engine/stdlib/package.json`. By removing this workspace, we avoid nested workspaces and also simplify `check-dependency-version-consistency` setup. We can reconsider this decision when we focus on developing `engine` app futher.

- `apps/engine` is converted into a dummy yarn workspace (`@app/engine`). This workspace does not have any scripts or dependencies, but it helps us with repo maintenance. For example, we can check presence of `LICENSE.md` in `apps/engine` just as we do for all other workspaces.

- Paths in `md` files are updated accordingly

## ⚠️ Known issues

- All workspaces should be in `/{apps,libs,tests}/workspace-name` or `/{apps,libs,tests}/@namespace/workspace-name`, with package name mapping to folder name. We end up with `@hashengine/apache-arrow-js-bundle` in `/apps/engine/apache-arrow-js-bundle`, which creates an exception. At the same time, `/apps/engine` does not have `package.json` so is not a workspace. I’ll attempt to fix this in a follow-up PR, as discussed [in Slack](https://hashintel.slack.com/archives/C022217GAHF/p1673552553548749) (internal).

## 🛡 What tests cover this?

CI


.prettierignore
```diff
@@ -23,8 +23,7 @@
 LICENSE.md
 
 # Skip autogenerated files tracked by git
-/packages/engine/lib/execution/src/runner/javascript/bundle_arrow.js
```
kachkaev, line 26: Non-existing file path, we have this one instead:

```
hash/engine/lib/execution/src/runner/javascript/apache-arrow-bundle.js
↓
apps/engine/lib/execution/src/runner/javascript/apache-arrow-bundle.js
```

apps/engine/lib/execution/src/runner/javascript/apache-arrow-bundle.js
```diff
@@ -3,7 +3,7 @@ export var arrow;
   // webpackBootstrap
   /******/ "use strict";
   /******/ var __webpack_modules__ = {
-    /***/ 717: /***/ (
+    /***/ 35: /***/ (
```
kachkaev, line 6: Result of re-running instructions in `apps/engine/apache-arrow-js-bundle/README.md`

TimDiekmann (commented):
Changes to the CI and the labels LGTM. Probably worth that @Alfred-Mountfield will have a look at the `npm` related things.

apps/engine/apache-arrow-js-bundle/README.md
```diff
@@ -4,10 +4,12 @@
...
 ### Building
 
-Run
+Move to the package folder and run:
 
-- `yarn install`
-- `yarn workspace @hashengine/apache-arrow-js-bundle bundle`
+```sh
+npm install
+npm run bundle
```
Alfred-Mountfield, line 11: What's the reasoning behind using `npm` here? That feels very strangely inconsistent with the rest of the repository.

I appreciate we don't want it to be a workspace within the root one, but we can avoid that by just not including it in the paths of all the workspaces surely? I don't see why we have to revert to npm here

kachkaev: A saw `npm` and `package-lock.json` in `apps/engine/stdlib`, so decided to do the same in this sibling folder. Both packages are ‘islands’, i.e. the are not visible from the root yarn workspace.

Alfred-Mountfield (approved):
I'd prefer if we moved to `yarn` and tried making the folders consistent with our guidelines considering we're already touching these files (e.g. making `stdlib` a `lib`), but it's also not worth the time debating it, if CI passes I'm fine with this.

apps/engine/apache-arrow-js-bundle/README.md
kachkaev: I was choosing between 'upgrading' `stdlib` to `yarn` and 'downgrading' `apache-arrow-js-bundle` to npm. The latter option won as I found it less risky. I fully agree with you about consistency, I’m just trying to make changes to `engine` as safe as possible.

TimDiekmann (changes requested):
Before merging I like to check, that the CI properly passed and did not skip any tests. Will approve afterwards.

nonparibus (approved):
Legal +1 follows Alfie's prior tech +1

Waiting for tests to pass and then good to merge.

nonparibus:
@TimDiekmann Happy to approve?

TimDiekmann (approved):
I wonder why this warning was not picked up by the lint job or the dev-build, but that's unrelated. LGTM 🎉 
image

PR #1821, opened 2023-01-13
Update Code block for Block Protocol 0.3

CiaranMn:

Updates the Code block for the forthcoming `0.3` version of the Block Protocol specification. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203673154406411/f) _(internal)_

## 🔍 What does this change?

- Installs canary version of block packages which were published from https://github.com/blockprotocol/blockprotocol/pull/879
- Ignores these packages in the yarn dependency check to avoid having to touch other packages in this PR
- Reference an externally-hosted `schema` in `package.json` which replaces the previous local definition of the block's entity type (previously `BlockEntityProperties`) – this schema is read and an equivalent type – now `RootEntity` - produced via the new `yarn codegen` script
- Update the block's props for the new shape of `graph`, and reads `blockEntity` from the new `blockEntitySubgraph` (which replaces `blockEntity, `linkGroups` and `linkedEntities`, although the latter two are irrelevant to this block)
- Adds some extra code to deal with the fact that property keys are now URIs and are a bit more of a pain to deal with (they are invalid variable names and therefore cannot be deconstructed from objects)
- Updates the props sent to `MockBlockDock`

## 🐾 Next steps

- #1823 makes this block loadable in HASH.
- Update more blocks to `0.3`

## ❓ How to test this?

1.  `yarn install && yarn dev`
2. Update some properties in the code block, check it's reflected in the mock datastore (bottom tab)
3. Edit some properties in the mock datastore, check it's reflected in the code block


blocks/code/src/dev.tsx
```diff
@@ -7,22 +7,34 @@ import { createRoot } from "react-dom/client";
...
  */
-const initialData = {
-  content: 'var foo = "bar";',
-  language: "javascript",
+const initialData: RootEntity = {
+  metadata: {
+    entityTypeId: "https://alpha.hash.ai/@ciaran/types/entity-type/code/v/1",
+    editionId: {
+      baseId: "entity-code",
+      versionId: "1",
```
Alfred-Mountfield, line 22: Does this not cause problems?

Alfred-Mountfield: (I would have expected a version to be a timestamp, or this to be `recordId`)

package.json
```diff
@@ -31,14 +31,14 @@
     "external-services": "docker compose --file packages/hash/external-services/docker-compose.yml --env-file .env",
     "fix": "npm-run-all --continue-on-error \"fix:*\"",
     "@todo.1": "Remove the ignore argument below for mock-block-dock once repo is upgraded fully to 0.2 spec",
-    "fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-dep=mock-block-dock",
+    "fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-dep=mock-block-dock --ignore-dep=@blockprotocol/graph --ignore-dep=block-scripts",
```
Alfred-Mountfield, line 34: Any thoughts on how we can track this? 

Really wish we could put TODO comments in package.json 😭

CiaranMn: We can (sort of), there's one two lines above this 

`yarn` will warn you if one of these checks is _unused_, i.e. we will be immediately told if we've made everything consistent and this remains... but if inconsistencies remain then we just need to somehow externally remember that they need fixing (maybe a clean up task for 0.3, but might not be a priority to go through all the other packages and give them the new versions)

blocks/code/src/dev.tsx
CiaranMn: No, because it's only used for getting a root from the subgraph, and both the vertices and the roots in this mock data are using the same `versionId`. Nothing is actually using version history. We'll be prompted to update this when we make the entity edition ids consistent between BP and HASH

Alfred-Mountfield: If you run `yarn dev`, what happens when you actually load it into HASH though? Does the demo data get ignored?

CiaranMn: Yes, because HASH does not use this file. This is the entry point for the development server. HASH uses the bundled source of `app.tsx`

Alfred-Mountfield: Forgot about how it loaded them from a URL (wasn't sure if it was the bundled source or if it just sorta embedded the live view 😅)

Alfred-Mountfield (approved):

PR #1823, opened 2023-01-13
Make 0.3 blocks load in HASH

CiaranMn:

Updates HASH to load blocks compliant with the `0.3` specification.

This means that we can reintroduce blocks that are currently not working in HASH by updating them to be `0.3` compliant. Prior to this PR, HASH had been partially updated to `0.3` in a way which meant no blocks were working (except text blocks, which do not rely on any message passing under the graph service).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203714761594036/f) _(internal)_

## 🔍 What does this change?

- Installs new canary versions of packages in `hash/frontend`
- Removes redundant code related to:
  - an earlier data mapping prototype which is not worth retaining, as it depends heavily on a previous way of linking entities. this includes code to fetch a block's schema, which the data mapping prototype used
  - rewriting entities to be in the shape defined in the BP. In `0.3`, HASH and BP entity shape will be consistent
- updates `block-loader.tsx` to fetch the subgraph rooted at the block entity. There are a few temporary measures involved which are covered in 'known issues'
- replaces the `updateEntity` shim with the new version imported from `@blockprotocol/graph@canary`
- **driveby**: fix filename for consistently, `loadremote-block.tsx` -> `load-remote-block.tsx`
- **driveby**: update the branded type to use `{ [_ in Kind]: true }>` instead of `{ [_ in Kind]: void }>` – makes a bit more sense semantically and didn't seem to be any objection to doing this

## ⚠️ Known issues

- There is a mismatch between the type of `Subgraph` in `@blockprotocol/graph` and in `hash/subgraph` – particularly in the identifiers for entities (`version` versus `versionId`). This leads to some temporary assertions in the code. 
- Because blocks use the `graph` package and expect types in its shape, there is some _adding_ fields they expect (`versionId`) so that they can correctly look up items in the subgraph map
- Each block fetching its own `subgraph` is a temporary measure that should be replaced by fetching the subgraph rooted at the `Page` to a sufficient depth to also fetch the block's linked entities. Currently, the special-cased `getPage` query terminates the query at the block entity. This should be fixed by removing the special casing of page retrieval in favour of using the general `getEntity` to fetch the appropriate subgraph.
- Paragraph blocks are automatically loaded on every page, and require temporarily adding properties in the pre-`0.3` shape to stop them crashing.

## 🐾 Next steps

- Sort out mismatch between `@blockprotocol/graph` and types in `hash`.
- Update more blocks to `0.3`

## ❓ How to test this?

1. `yarn install`
2.  in `blocks/code` run `yarn dev` so that the block's dev server is running
3. run the HASH app by `yarn dev`
4. Navigate to a page, click the context menu next to a block, 'load block', and paste in the code block's URL (should be `http://localhost:9090`


packages/hash/frontend/src/blocks/page/component-view.tsx
```diff
@@ -195,21 +191,16 @@ export class ComponentView implements NodeView {
               }}
             >
               <BlockLoader
-                key={entityId} // reset the component state when the entity changes, e.g. to reset the data map state
-                blockEntityId={entityId}
+                key={entityId} // reset the component state when the entity changes
+                blockEntityId={
+                  childEntity?.metadata.editionId.baseId as EntityId | undefined
+                } // @todo make this always defined
+                blockEntityTypeId={this.block.meta.schema as VersionedUri} // @todo-0.3 remove when @blockprotocol/core types updated
```
Alfred-Mountfield, line 198: Is this tracked in an asana task?

packages/hash/frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -1,44 +1,41 @@
...
 import { RemoteBlock } from "../remote-block/remote-block";
-import { DataMapEditor } from "./data-map-editor";
 import { fetchEmbedCode } from "./fetch-embed-code";
-import { SchemaMap } from "./shared";
 
-// @todo consolidate these properties, e.g. take all entityX, linkX into a single childEntity prop
-// @see https://app.asana.com/0/1200211978612931/1202807842439190/f
 type BlockLoaderProps = {
-  blockEntityId: string;
+  blockEntityId?: EntityId; // @todo make this always defined
```
Alfred-Mountfield, line 33: Out of curiosity, why did this have to become `undefined`?

If it takes more than a minute or two to explain, don't worry about it.

packages/hash/frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -1,44 +1,41 @@
...
+  blockEntityTypeId: VersionedUri;
   blockMetadata: HashBlockMeta;
-  blockSchema: JsonSchema;
   editableRef: (node: HTMLElement | null) => void;
-  entityId: EntityId;
-  entityTypeId: string;
-  entityProperties: {};
   onBlockLoaded: () => void;
+  wrappingEntityId: string;
```
Alfred-Mountfield, line 38: Is this not an actual `EntityId`?

packages/hash/frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -50,53 +47,105 @@ type BlockLoaderProps = {
...
+    if (!blockEntityId) {
+      // when inserting a block in the frontend we don't yet have an entity associated with it
+      // there's a delay while the request to the API to insert it is processed
+      // @todo some better way of handling this – probably affected by revamped collab.
+      //    or could simply not load a new block until the entity is created?
+      const now: string = new Date().toISOString();
+      const placeholderEntity = {
+        metadata: {
+          editionId: {
+            baseId: "placeholder-account%entity-id-not-set",
```
Alfred-Mountfield, line 88: I see the todo above and I really think we should consider not loading it until an entity is created. (Not asking for us to change it in this PR as I don't have a good sense at how much effort would be involved). Just this code feels exactly like the kind of code that will cause really annoying bugs, we're setting random placeholder strings (which are not UUIDs and may cause validation failures, although we wouldn't want them to masquerade as being valid to be fair), inside a pretty constrained template-string type, which will give us false confidence.

packages/hash/frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -50,53 +47,105 @@ type BlockLoaderProps = {
...
+          throw new Error(
+            `Could not get entity ${blockEntityId} ${
+              errors ? JSON.stringify(errors, null, 2) : ""
+            }`,
+          );
+        }
+
+        setGraphProperties({
+          blockEntitySubgraph: {
+            ...(data as unknown as Subgraph<SubgraphRootTypes["entity"]>), // @todo-0.3 do something about this,
```
Alfred-Mountfield, line 131: I appreciate where these casts have come from, I've got something similar in my branch but I'm not sure we should commit it to `main`, even temporarily. Just feels like there are far far too many gotchas and we may end up with some non-insignificant lost dev time to bugs we accidentally introduce here.

What do you think about blocking this PR on us either updating the BP package, or introducing a refactoring PR in HASH to address the mismatches?

Alfred-Mountfield (commented):
Thanks Ciaran this is great progress to us having things working again, I feel like the end is in sight for the big move to the type system. 

I'm not entirely sure how to move forwards with this PR though. There are going to definitely be some migration pains, things will almost certainly _**have**_ to be out of sync at some point during the process, and that will bring unpleasantries with it I'm sure. However, I hope we can (still efficiently) do slightly better than the number of casts and such we have to do in this PR. I've opened a thread to discuss it more, as I feel very unsure.

packages/hash/frontend/src/blocks/page/component-view.tsx
CiaranMn: No – I'm just going to come back and fix all these `todo-0.3` when we have consistency between the BP types and the HASH types (or least a strategy to reconcile them without assertions)

packages/hash/frontend/src/components/block-loader/block-loader.tsx
CiaranMn: In practice this has always been potentially null. We previously had this prop as required and asserted that it was always defined when calling the component, but then handled it being missing inside this component anyway (current code in `main` [here](https://github.com/hashintel/hash/blob/ae4949d654012828d319816ef8739457c79e97ef/packages/hash/frontend/src/components/block-loader/block-loader.tsx#L83))

This just makes the type reflect reality

packages/hash/frontend/src/blocks/page/component-view.tsx
Alfred-Mountfield: That could work (although a bit of a bus factor there 😅)

Just another possible approach, in the backend we've started putting asana task links in almost all the TODOs. If we had a task to handle that reconciliation, we could just put the asana task in all of these and `ctrl+f` for that, it'd make it easier for someone in the future to also be able to guess that something was missed, rather than needing to know what `todo-0.3` is.

Not a big thing though

packages/hash/frontend/src/components/block-loader/block-loader.tsx
Alfred-Mountfield: Gotcha, not ideal

packages/hash/frontend/src/blocks/page/component-view.tsx
CiaranMn: Created a [task](https://app.asana.com/0/1203543021352041/1203756383156843/f) (internal) in which I'll do this. Since I'll be fixing them this week I don't think we need to link the task in all these todos, although I agree with the principle as a good thing to do to make sure context isn't lost

If I don't fix some of the todos I'll create tasks and link them

packages/hash/frontend/src/components/block-loader/block-loader.tsx
CiaranMn: I agree we need to fix this. Something we've spoken about occasionally is the FE being able to generate its own ids, so it can start to refer to entities without having a response from the db. We've planned to consider this in the context of new collab – [task](https://app.asana.com/0/1201095311341924/1203257689029607/f) (internal). I think this would be better than a loading state, although we could implement the latter if the former proved impossible

packages/hash/frontend/src/components/block-loader/block-loader.tsx
CiaranMn: It should be, although as part of the wider 'block might be new and entities not created' problem it's possibly an empty string... to type it properly I'd have to either (a) assert it when loading this component, or (b) implement 'don't load this until entities are back from db', which I've no idea how much work it is

I can do (a) if you think it preferable 

Here's the relevant field being defined in the calling file
Screenshot 2023-01-16 at 13 45 29

packages/hash/frontend/src/components/block-loader/block-loader.tsx
CiaranMn: We spoke about this and agreed to get this PR in on the basis that I would fix the type mismatches soon this week (following a discussion on approach tomorrow), and warn anyone who was planning to work in this area about possibly waiting until they are fixed

packages/hash/frontend/src/components/block-loader/block-loader.tsx
Alfred-Mountfield: I don't have a good sense of what's worth the investment at the moment. Perhaps it's fine to leave as is. Just quite uncomfortable to see an `entityId` variable which is not an `EntityId`

Alfred-Mountfield (dismissed):

Alfred-Mountfield (approved):

PR #1825, opened 2023-01-13
Add soft CI check for LICENSE.md files

kachkaev:

This PR should help us maintain repo integrity. It introduces a script that should fail CI when `LICENSE.md` file is missing in a Yarn workspace (including root workspace).

The initial implementation is called _soft CI check_ because exit code is always zero. This will be changed in a follow-up PR after #1831 and similar PRs are merged.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203669287079639/f) _(internal)_

## 📹 Demo

https://github.com/hashintel/hash/actions/runs/3931754090/jobs/6723419910#step:7:23

Screenshot 2023-01-16 at 15 43 51


nonparibus:
I wonder if it would be worth us checking for LICENSES alongside Cargo.toml and pyproject.toml, as well.

kachkaev:
@nonparibus that's a good idea, we can do this in a follow-up. We might need to come up with some special rule for crates because there exist `Cargo.toml` in nested folders:

Screenshot 2023-01-13 at 17 34 49


Going forward, we can supplement this initial script with more checks. For example, we can check if `each `{apps,libs,tests}/*` and `libs/@*/*` folder has `package.json` in it.

nonparibus:
@kachkaev Captured the follow-up for future triage [here](https://app.asana.com/0/1203193968491715/1203754204131510/f) (_internal ticket_)

nonparibus:
@kachkaev Please can we add a check that I (or any ext. contributor, etc.) hasn’t accidentally named a file `LICENCE` again? Exceptionally dumb, but I just did it in https://github.com/hashintel/hash/pull/1847 (see https://github.com/hashintel/hash/pull/1847/commits/ee4af7284b89d493ba7fd37ff82139735b764ae9), and I would like to catch this automatically if it reoccurs. Should be simple enough to add, and I don’t think unnecessarily bloats the script. Thank you.

kachkaev:
If we add `LICENCE.md`, there won’t be `LICENSE.md` so this check will fail. Will this suffice?

libs/@local/repo-chores/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "fix:eslint": "eslint --fix .",
+    "lint:eslint": "eslint --report-unused-disable-directives .",
+    "lint:tsc": "tsc --noEmit"
+  },
+  "dependencies": {
+    "chalk": "4.1.2",
+    "execa": "5.1.1",
+    "fs-extra": "10.1.0",
+    "globby": "11.0.1",
+    "typescript": "4.9.4"
```
CiaranMn, line 16: `devDependencies`?

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
```diff
@@ -0,0 +1,139 @@
...
+    ...Object.entries(workspaceInfoLookup).map(([, workspaceInfo]) =>
+      path.resolve(monorepoRootDirPath, workspaceInfo.location),
+    ),
+  ];
+
+  const licenseFilePaths = await globby("**/license*", {
+    absolute: true,
+    caseSensitiveMatch: false,
+    cwd: monorepoRootDirPath,
+    ignore: ["**/node_modules/**", "**/apps/engine/*/**/*"],
```
CiaranMn, line 39: Why is `engine` ignored?

nonparibus:
@kachkaev I thought about this, but believe it would still be beneficial - e.g. first, in our specific instance `LICENCE-CC.md` was the specific example which occurred, which would not have been flagged - as well as second in the event a malicious (or unwitting) user inserts, and we somehow end up with, co-located `LICENCE` and `LICENSE` files which go unchecked during review.

libs/@local/repo-chores/package.json
kachkaev: We have
```
    "ts-node": "10.9.1",
    "typescript": "4.9.4"
```
in other workspaces in which we have `yarn exe`. I added `ts-node` in https://github.com/hashintel/hash/pull/1825/commits/507d472eca160d5ecf5154f7b48516bf9ce19e1d. If that’s inaccurate, we can move these dependencies in all workspaces one day, but at least it will be consistent for now.

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
kachkaev: I clarified the reason in https://github.com/hashintel/hash/pull/1825/commits/f3fd69842a32dc974e7a4282674686781931f199. If we don't have this glob, we get this in output:

```
[EXTRA]    apps/engine/lib/memory/src/arrow/util/LICENSE
[EXTRA]    apps/engine/lib/memory/src/arrow/ipc/serialize/LICENSE
```

After some more thinking, I removed this glob. Lines marked as `[EXTRA]` don’t fail CI. https://github.com/hashintel/hash/pull/1825/commits/94a0ec1c3ef04650f6c4643dafc3683d783e14a7

CiaranMn (approved):

kachkaev:
>  I thought about this, but believe it would still be beneficial - e.g. first, in our specific instance `LICENCE-CC.md` was the specific example which occurred, which would not have been flagged - as well as second in the event a malicious (or unwitting) user inserts, and we somehow end up with, co-located `LICENCE` and `LICENSE` files which go unchecked during review.

https://github.com/hashintel/hash/pull/1852

PR #1834, opened 2023-01-16
Change return value of `Interval::merge` to always return a value

TimDiekmann:

The `Interval::merge` operation currently returns an iterator, which always has a single value. Instead, we can return the value directly.

## 🔍 What does this change?

- Change the return value of `Interval::merge` from `impl ExactSizeIterator<Item = Self>` to `Self`

## 🛡 What tests cover this?

- The Rust compiler
- The unit tests of `Interval` were adjusted to use only a single value instead of an iterator

### DIFF
--- a/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
+++ b/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
@@ -143,7 +143,7 @@ pub trait Interval<T>: Sized {
     ///
     /// In comparison to [`Self::union`], this method does also return the points between the
     /// intervals if they do not overlap.
-    fn merge(self, other: Self) -> IntervalIter<Self>
+    fn merge(self, other: Self) -> Self
     where
         T: PartialOrd,
     {
@@ -153,7 +153,7 @@ pub trait Interval<T>: Sized {
         // Range A:   [-----] | [-----]   | [-----]         |         [-----] | [---------]
         // Range B: [-----]   |   [-----] |         [-----] | [-----]         |   [-----]
         // Result:  [-------] | [-------] | [-------------] | [-------------] | [---------]
-        Return::one(Self::from_bounds(
+        Self::from_bounds(
             match lhs_lower.cmp_lower(&rhs_lower) {
                 Ordering::Less | Ordering::Equal => lhs_lower,
                 Ordering::Greater => {
@@ -166,7 +166,7 @@ pub trait Interval<T>: Sized {
                     <Self::UpperBound as UpperBound<T>>::from_bound(rhs_upper.into_bound())
                 }
             },
-        ))
+        )
     }
 
     /// Returns a new interval that contains all points in both intervals.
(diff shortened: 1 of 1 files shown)


packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -183,7 +183,7 @@ pub trait Interval<T>: Sized {
         // Range A:   [-----] | [-----]   | [-----]         |         [-----] | [---------]
         // Range B: [-----]   |   [-----] |         [-----] | [-----]         |   [-----]
         // Result:  [-------] | [-------] | [-------------] | [-------------] | [---------]
```
Alfred-Mountfield, line 185: This confused me for quite a while, can you add a heading above each column to explain they're examples or something (e.g. `Example 1 | Example 2`, or maybe just a top level heading saying "Each column constitutes an example"

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -554,15 +553,15 @@ mod tests {
...
             [excluded_excluded(0, 5)],
         );
         test(
             included_included(5, 15),
             excluded_included(0, 10),
             [included_included(5, 10)],
             [excluded_included(0, 15)],
-            [excluded_included(0, 15)],
+            excluded_included(0, 15),
             [excluded_included(10, 15)],
```
Alfred-Mountfield, line 565: Do we want to consider making a struct for these? I appreciate these are tests but it'd be too easy to accidentally swap one of these around.

Alfred-Mountfield (approved):
Some thoughts, but fine to merge without

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: This is just a shorthand for
```rust
IntervalBounds::<u32>::from_bounds(Bound::Excluded(0), Bound::Included(15))
```
to make it more readable

Alfred-Mountfield: Sorry I mean encapsulating the parameters to the actual `test` function

TimDiekmann: I cannot really follow. Could you give an example?

Alfred-Mountfield:
```Rust
test(IntervalTestParameters {
    lhs: included_included(5, 15),
    rhs: excluded_included(0, 10),
    intersection: [included_included(5, 10)],
    union: [excluded_included(0, 15)],
    merge: excluded_included(0, 15),
    difference:[excluded_included(10, 15)]
}

TimDiekmann: Oh, yes, that makes sense. I will adjust this as part of #1838

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
Alfred-Mountfield: Thanks, I'll respond there

PR #1835, opened 2023-01-16
Add `Interval::contains_interval` function

TimDiekmann:

Adds a function to the `Interval` trait to check if another interval is contained in this interval

## 🚫 Blocked by

- #1834 

## 🛡 What tests cover this?

An additional test was added to the test suite based on `lhs.merge(rhs) == lhs`, which is only true, if `lhs` contains `rhs`

### DIFF
--- a/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
+++ b/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
@@ -123,6 +123,25 @@ pub trait Interval<T>: Sized {
             || other.upper_bound().is_adjacent_to(self.lower_bound())
     }
 
+    /// Checks if this interval completely contains the other interval.
+    ///
+    /// Returns `true` if this interval's lower bound is less than or equal to the other interval's
+    /// lower bound and this interval's upper bound is greater than or equal to the other
+    /// interval's.
+    #[must_use]
+    fn contains_interval(&self, other: &impl Interval<T>) -> bool
+    where
+        T: PartialOrd,
+    {
+        matches!(
+            self.lower_bound().cmp_lower(other.lower_bound()),
+            Ordering::Less | Ordering::Equal
+        ) && matches!(
+            self.upper_bound().cmp_upper(other.upper_bound()),
+            Ordering::Greater | Ordering::Equal
+        )
+    }
+
     /// Returns the complement of this interval.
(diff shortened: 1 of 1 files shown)


packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -354,6 +373,18 @@ mod tests {
             "difference calculated by complement",
         );
 
+        if lhs.merge(rhs) == lhs {
+            assert!(
+                lhs.contains_interval(&rhs),
+                "{lhs:?} contains {rhs:?}, but does not report so"
```
Alfred-Mountfield, line 379: I don't quite get this, what do you mean by "but does not report so"

TimDiekmann: `contains_interval` should return true, so that `lhs` contains `rhs`, but it reports `false`

Alfred-Mountfield: Then perhaps we should say 
```suggestion
                "{lhs:?} contains {rhs:?}, but `contains_interval` reported otherwise"
```

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -384,6 +403,18 @@ mod tests {
...
 
+        if lhs.merge(rhs) == lhs {
+            assert!(
+                lhs.contains_interval(&rhs),
+                "{lhs:?} contains {rhs:?}, but `contains_interval` reported otherwise"
+            );
+        } else {
+            assert!(
+                !lhs.contains_interval(&rhs),
+                "{lhs:?} does not contain {rhs:?}, but reports so"
```
TimDiekmann, line 414:
```suggestion
                "{lhs:?} does not contain {rhs:?}, but `contains_interval` reports so"
```

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: I also changed this for the other messages in 8b83780096846b94935db84c9f9172266b46d685

Alfred-Mountfield (approved):

PR #1827, opened 2023-01-13
Revamp Graph migrations

thehabbos007:
## 🌟 What is the purpose of this PR?

This PR introduces a new migrations system using [refinery](https://github.com/rust-db/refinery) instead of our semi-bespoke typescript based setup using [node-pg-migrate](https://github.com/salsita/node-pg-migrate).

The CLI for the Graph has been changed to include migrations in-binary with a subcommand `migrate` to run the migrations. To keep the CLI simple, the webserver subcommand is the default.

The docker composes have all been updated to use the new migration system through the graph binary.

We may want to consider auto-formatting the SQL files (but most SQL formatters produce rather ugly output :( )

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1201970098275906/f) _(internal)_

## 🔍 What does this change?

- Remove old TS migrations for graph
- Add new migration with refinery in `hash_graph/lib` with a `hash_graph/postgres_migrations` directory containing the migration files
- Add new trait for migrations with support for getting current applied migration status (will be useful for health checks in the future)
- Update clap arg parser to allow giving a `migrate` subcommand

## 📜 Does this require a change to the docs?

- I've updated the readme of the graph to include how to run/create migrations

## 🐾 Next steps

- Automated deployments [Asana task](https://app.asana.com/0/1203363157432094/1203410952443002/f)
- Figure out if it makes sense to include migration in health checks for rolling out new systems

## 🛡 What tests cover this?

- CI and dev environment. Should be seamless and no changes to normal workflow. Please do test it locally

## ❓ How to test this?

Start the app as you always would (whether it be from the graph directory with `cargo make deployment-up && cargo run` or through external-services )

PR #1837, opened 2023-01-16
Support searching for `rust-toolchain.toml` files in parent directories in CI

TimDiekmann:

rustup is able to search for the `rust-toolchain.toml` in the crate's or parent directory. Our CI currently requires the file to be present in the crate's directory (next to `Cargo.toml`). This is restricting, if we want to run `cargo` from another directory than the crate root.

Currently, #1827 fails because of this. Also, the toolchain files of `error-stack` and `deer` are currently duplicated because of this issue.

## 🔗 Related links

- #1827

## 🔍 What does this change?

- Add a function to the `setup.py` script to find the correct toolchain file (or error otherwise)
- Remove the obsolete toolchain files in `error-stack//macro` and `deer/macros`

### DIFF
--- a/packages/libs/deer/macros/rust-toolchain.toml
+++ b/packages/libs/deer/macros/rust-toolchain.toml
@@ -1,2 +1,0 @@
-[toolchain]
-channel = "nightly-2022-09-27"
--- a/.github/scripts/rust/setup.py
+++ b/.github/scripts/rust/setup.py
@@ -66,6 +66,29 @@ def find_local_crates():
     return checked_crates
 
 
+def find_toolchain(crate):
+    """
+    Returns the toolchain for the specified crate.
+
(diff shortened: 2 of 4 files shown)


packages/libs/deer/macros/rust-toolchain.toml
```diff
@@ -1,2 +0,0 @@
-[toolchain]
-channel = "nightly-2022-09-27"
```
thehabbos007, line 2: Probably a good call to remove this as it's desynced with the parent create (which is at 
`nightly-2022-11-14`)

thehabbos007: cc: @indietyp beware that we are now using the parent toolchain for the macro folder; likely not a problem though

thehabbos007 (approved):
Looks good to me and CI passes. Thanks! :+1:

PR #1838, opened 2023-01-16
Improve comments in bound comparison logic

TimDiekmann:

A minor change to slightly improve the in-code comments on how bounds are compared

### DIFF
--- a/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
+++ b/packages/graph/hash_graph/lib/interval-ops/src/interval.rs
@@ -98,10 +98,10 @@ pub trait Interval<T>: Sized {
         let rhs_lower = other.lower_bound();
         let rhs_upper = other.upper_bound();
 
-        // Example |      A     |     B
-        // ========|============|============
-        // Range A |    [-----] | [-----]
-        // Range B | [-----]    |    [-----]
+        // Examples |      1     |     2
+        // =========|============|============
+        // Range A  |    [-----] | [-----]
+        // Range B  | [-----]    |    [-----]
         matches!(
             lhs_lower.cmp_lower(rhs_lower),
             Ordering::Greater | Ordering::Equal
@@ -164,7 +164,7 @@ pub trait Interval<T>: Sized {
     where
         T: PartialOrd,
     {
-        // Examples   |       1     |    2    |    3    |    4    |    5
+        // Examples   |      1      |    2    |    3    |    4    |    5
         // =========================|=========|=========|=========|=========
         // Range      |   [-----]   | ---)    | ---]    |    (--- |    [---
         // -------------------------|---------|---------|---------|---------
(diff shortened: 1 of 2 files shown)


packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -98,8 +98,10 @@ pub trait Interval<T>: Sized {
         let rhs_lower = other.lower_bound();
         let rhs_upper = other.upper_bound();
 
-        // Range A:    [-----] | [-----]
-        // Range B: [-----]    |    [-----]
+        // Example |      A     |     B
```
Alfred-Mountfield, line 101: I like a lot of the different format strategies in this commit, but I think we should try and be consistent with them. 

For instance, we differ between a heading of `Examples` and `Example`, we also differ between using numeric indices, or alphabetic ones. I'd probably encourage us to use numeric as you seem to be calling things "Range A" and "Range B" and such.

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -98,8 +98,10 @@ pub trait Interval<T>: Sized {
...
+        // ========|============|============
+        // Range A |    [-----] | [-----]
+        // Range B | [-----]    |    [-----]
```
Alfred-Mountfield, line 104: Does this comment need an output/result row?

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -162,8 +164,11 @@ pub trait Interval<T>: Sized {
     where
         T: PartialOrd,
     {
-        // Range:        [-----]   | ---)    | ---]    |    (--- |    [---
-        // Complement: --)     (-- |    [--- |    (--- | ---]    | ---)
+        // Examples   |       1     |    2    |    3    |    4    |    5
```
Alfred-Mountfield, line 167: Not important, and maintaining it might prove a bit annoying, but the `1` is off center
```suggestion
        // Examples   |      1      |    2    |    3    |    4    |    5
```

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -231,10 +239,12 @@ pub trait Interval<T>: Sized {
...
-            // The ranges overlaps
-            // Range A:   [-----] | [-----]
-            // Range B: [-----]   |   [-----]
-            // Result:    [---]   |   [---]
+            // Examples  |      1    |     2
+            // ==========|===========|===========
+            // Range A   |   [-----] | [-----]
+            // Range B   | [-----]   |   [-----]
+            // ----------|-----------|-----------
+            // Intersect |   [---]   |   [---]
```
Alfred-Mountfield, line 247: I don't feel strongly about this, I know the method is called `intersect`, just reading the table felt a bit strange.

```suggestion
            // Examples     |     1     |     2
            // =============|===========|===========
            // Range A      |   [-----] | [-----]
            // Range B      | [-----]   |   [-----]
            // -------------|-----------|-----------
            // Intersection |   [---]   |   [---]
```

Alternatively, could just fix the heading for `1`
```suggestion
            // Examples  |     1     |     2
            // ==========|===========|===========
            // Range A   |   [-----] | [-----]
            // Range B   | [-----]   |   [-----]
            // ----------|-----------|-----------
            // Intersect |   [---]   |   [---]
```

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
```diff
@@ -231,10 +239,12 @@ pub trait Interval<T>: Sized {
             let (lhs_lower, lhs_upper) = self.into_bound();
```
Alfred-Mountfield, line 239: I can't comment above this line, but the docs for this method say `intersection` while the method is called `intersect`, do we perhaps want to pick one or the other?

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: IMO it doesn't because it just shows how ranges could overlap

packages/graph/hash_graph/lib/interval-ops/src/interval.rs
TimDiekmann: Actually, the comment was outdated. I implemented the operator for the operations as well but they got removed. I removed the comments as well.

Alfred-Mountfield (approved):

PR #1843, opened 2023-01-16
Make test cases for interval operations more readable

TimDiekmann:

As raised in https://github.com/hashintel/hash/pull/1834#discussion_r1071097117, this will improve the readability of the interval operation tests


Alfred-Mountfield (approved):
Shame you can't do `impl IntoIterator` in the struct and avoid nasty generics :(

PR #1844, opened 2023-01-16
Reimplement `queryEntities` with transpilation to structural queries

Alfred-Mountfield:

As part of experimenting with the Type System in HASH, we removed the `aggregateEntities` resolvers and implemented a stand-in `getAllLatestEntities`. This reimplements the (now called `queryEntities`) endpoint to match the updates made in the `0.3` release of the Block Protocol specification.

There are some operations which are currently unsupported but will be in the near(ish) future.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203749694686234/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Implements the resolver for `queryEntities` which takes the BP aggregation syntax and translates it into a structural query across the HASH graph

Not presently

## ⚠️ Known issues

Quite a few deficiencies were discovered, please see the added TODOs in the code 

## 🐾 Next steps

- Expand structural querying to support the deficiencies identified
- ~~Update the aggregate syntax in the BP to take a list of fields instead of a single one~~

## 🛡 What tests cover this?

Unsure if there are any besides code analysis

## ❓ How to test this?

1.  Go to the GraphQL endpoint
1.  Confirm that the `aggregateEntities` query works as expected, for example:
```graphql
query {
  queryEntities(
    operation: {
      multiFilter:{
        filters: [
          {
            field: ["metadata", "entityTypeId"]
            operator:"EQUALS"
            value: "http://localhost:3000/@system-user/types/entity-type/page/v/1"
          }
        ]
        operator: "AND"
      }
    }
    constrainsLinksOn: {outgoing:0}
		constrainsValuesOn:{outgoing:0}
    constrainsPropertiesOn:{outgoing:0}
    constrainsLinkDestinationsOn:{outgoing:0}
    isOfType:{outgoing:0}
    hasLeftEntity:{outgoing:0,incoming:0}
    hasRightEntity:{outgoing:0,incoming:0}
  ) {
    roots
    vertices
    edges
  }
}
```

or

```graphql
query {
  queryEntities(
    operation: {
      multiFilter:{
        filters: [
          {
            field: ["properties", "http://localhost:3000/@system-user/types/property-type/title/"]
            operator: "EQUALS"
            value: "Middle"
          }
          {
            field: ["properties", "http://localhost:3000/@system-user/types/property-type/title/"]
            operator: "EQUALS"
            value: "First"
          }
        ]
        operator: "OR"
      }
    }
    constrainsLinksOn: {outgoing:0}
		constrainsValuesOn:{outgoing:0}
    constrainsPropertiesOn:{outgoing:0}
    constrainsLinkDestinationsOn:{outgoing:0}
    isOfType:{outgoing:0}
    hasLeftEntity:{outgoing:0,incoming:0}
    hasRightEntity:{outgoing:0,incoming:0}
  ) {
    roots
    vertices
    edges
  }
}
```

## 📹 Demo
image


packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
   return entitySubgraph as Subgraph;
 };
 
+export const aggregateEntitiesResolver: Extract<
```
Alfred-Mountfield, line 192: Unfortunately we need this `Extract`. The generated type is a `Resolver<…>` which is:
```ts
export type Resolver<TResult, TParent = {}, TContext = {}, TArgs = {}> =
  | ResolverFn<TResult, TParent, TContext, TArgs>
  | ResolverWithResolve<TResult, TParent, TContext, TArgs>;
```

Where `ResolverFn` is a function (duh) and `ResolverWithResolve` is 
```ts
export type ResolverWithResolve<TResult, TParent, TContext, TArgs> = {
  resolve: ResolverFn<TResult, TParent, TContext, TArgs>;
};
```

So if I use the generated type, this then causes problems in our middleware-types
```ts
export type ResolverMiddleware<
  TStartContext,
  TArgs,
  TEndContext = TStartContext,
> = (
  next: ResolverFn<any, any, TEndContext, any>,
) => ResolverFn<any, any, TStartContext, TArgs>;
```

Changing that to take `Resolver` instead would be quite a pain to deal with as we’d have to type-guard to check if it’s an object everywhere we use it

packages/hash/api/src/graphql/resolvers/index.ts
```diff
@@ -99,6 +100,7 @@ export const resolvers = {
     blocks: loggedInAndSignedUpMiddleware(blocksResolver),
     getEntity: loggedInAndSignedUpMiddleware(getEntityResolver),
     getAllLatestEntities: getAllLatestEntitiesResolver,
+    aggregateEntities: loggedInAndSignedUpMiddleware(aggregateEntitiesResolver),
```
Alfred-Mountfield, line 103: The resolver above (which this will eventually replace) does not have loggedInAndSignedUpMiddleware. I think someone told me this was to do with viewing pages without an account and such, do we have any information about this decision, perhaps a task to track the security of this?

cc: @CiaranMn @thehabbos007 @benwerner01

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+    isOfType,
+    hasLeftEntity,
+    hasRightEntity,
+  },
+  { dataSources },
+  __,
+) => {
+  const { graphApi } = dataSources;
+
+  /** @todo - should we throw an error, or should we just ignore the constraint and hope for the best */
```
Alfred-Mountfield, line 212: @CiaranMn any idea how we want to handle this at the moment? (regarding not having support for pagination)

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+    });
+  }
+
+  if (operation.multiFilter !== undefined && operation.multiFilter !== null) {
+    const subFilters: Filter[] = [];
+
+    /**
+     * @todo - support more than just "properties" as a field so we can filter on other things in the entity
+     *   https://app.asana.com/0/0/1203749694686236/f
+     */
```
Alfred-Mountfield, line 255: Right now there is no way to filter based on metadata (such as link destinations, order, timestamps, etc.)

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+        case "IS_EMPTY":
+          subFilters.push({
+            equal: [
+              { path: ["properties", operationFilter.field] },
+              /**
+               * @todo - This is incorrect and will return TRUE for properties with value `null`, not just missing ones
+               *   https://app.asana.com/0/0/1203749694686236/f
+               */
+              // @ts-expect-error -- Our typings don't allow us to pass `null` here, when they should.
+              null,
```
Alfred-Mountfield, line 273: This is one of @TimDiekmann's next-ups (Asana Task (internal))

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+            throw new ApolloError(
+              "Equality checks with `null` are currently unsupported",
+            );
+          }
+          subFilters.push({
+            equal: [
+              /**
+               * @todo - This assumes this is just one field
+               *   https://app.asana.com/0/0/1203749694686236/f
+               */
```
Alfred-Mountfield, line 307: We need to update the BP aggregation type to take an array of fields (currently untracked), and update structural queries to support it (This is one of @TimDiekmann's next-ups (Asana Task (internal)))

cc @CiaranMn

packages/hash/frontend/src/graphql/queries/knowledge/entity.queries.ts
```diff
@@ -75,6 +75,33 @@ export const getAllLatestEntitiesQuery = gql`
   ${subgraphFieldsFragment}
 `;
 
+export const aggregateEntitiesQuery = gql`
```
Alfred-Mountfield, line 78: I opted to not remove getAllLatestEntities in this PR. I thought it would make sense to do it partially, and wait until we've fully implemented this query before migrating frontend code that doesn't use the hook)

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx
```diff
@@ -38,14 +38,17 @@ export const EntitySelector = ({
...
           data: {
-            rootEntityTypeIds: expectedEntityTypes.map(
-              ({ schema }) => schema.$id,
-            ),
+            operation: {
+              /** @todo - we want more than one entity type here */
+              entityTypeId: expectedEntityTypes.map(
+                ({ schema }) => schema.$id,
+              )[0],
+            },
```
Comment (Alfred-Mountfield) [inline packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx:46]: The Aggregation syntax doesn't take multiple right now, I'm not sure what to do in this PR. We could replace this with an unconstrained aggregateEntities call, but I'm not sure if the logic using this data relies on the roots being filtered already. We could split it into multiple queries, but that's temporary hacky investment I'm trying to avoid.

packages/hash/api/src/graphql/resolvers/index.ts
CiaranMn: I'm not sure why `getAllLatestEntities` does not have the middleware on it - it's the `page` query that needs to be ungated to allow for anonymous viewing of a page.

I would not read any security into these middlewares – we know that the app is insecure (made explicit in the second line in the README) and it will not be secure until we introduce a proper authorization system.

Regardless of these middlewares, in the absence of proper authorization gating on the supply of individual entities (not queries), anyone can sign up for an account, construct a graph in which they own an entity connected to someone else's entities of interest, and retrieve anything.

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
CiaranMn: Where does support for this need to be added? the BP, SQ, or both?

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx:46]: can we make it an array?

package.json
```diff
@@ -30,14 +30,14 @@
     "external-services": "docker compose --file packages/hash/external-services/docker-compose.yml --env-file .env",
     "fix": "npm-run-all --continue-on-error \"fix:*\"",
     "@todo.1": "Remove the ignore argument below for mock-block-dock once repo is upgraded fully to 0.2 spec",
-    "fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-dep=mock-block-dock --ignore-dep=@blockprotocol/graph --ignore-dep=@blockprotocol/hook --ignore-dep=block-scripts",
+    "fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-dep=mock-block-dock --ignore-dep=@blockprotocol/hook --ignore-dep=@blockprotocol/graph --ignore-dep=@blockprotocol/hook --ignore-dep=block-scripts",
```
thehabbos007, line 33: as Ciaran mentioned in the same file, `hook` also appears twice here

packages/hash/api/src/graphql/resolvers/index.ts
thehabbos007: image
`getAllLatestEntities` is called from the login page without a logged in user. I'm sure @benwerner01 knows the reason :)

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+  if (
+    (operation.pageNumber !== undefined && operation.pageNumber !== null) ||
+    (operation.itemsPerPage !== undefined && operation.itemsPerPage !== null)
+  ) {
+    throw new ApolloError(
+      "Pagination on aggregateEntities results is not currently supported",
+    );
+  }
+
+  if (operation.multiSort !== undefined && operation.multiSort !== null) {
```
thehabbos007, line 222: doesn't `!= null` cover both `null` and `undefined` or does our linting not allow this? (applies for the other similar checks in this file)

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx
Comment (thehabbos007) [reply packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx:46]: How come `entityTypeId` is a standalone field and not just a `filter`? We'd get multiple constrained entity types (Using `OR`) if we allow it as just a filter and remove this explicit field

thehabbos007 (commented):
Thanks Alfie, I've left some comments

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx:46]: Yes – I think ultimately we just want it to be part of the filters, along with timestamps etc. I think there was some issue with doing this though that we needed to address (those fields being split across multiple tables?)

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: > Could we not just paginate the roots (skip and limit based on the pagination arguments) when returning them? At least then FE clients can be constructed to make calls for pagination and get progressive pages of the roots.

I guess we could.. I'm not sure if it would be a good thing. I'll try and think on this a little more.

> I know it will mean there is additional stuff in the subgraph but I don't think that matters – I may have forgotten an explanation of why it does in fact matter though.

This is a harder conversation, and one I was meaning to bring up in another forum today. @TimDiekmann and I were thinking about it in the [context of resolving the graph for various time intervals](https://github.com/hashintel/hash/pull/1848#discussion_r1072504403). I'll initiate the conversation somewhere with some context and tag you in it.

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: SQ supports almost all of those (except nested fields in `properties`), the BP could support it if we just took an array of fields and said in the spec that it represents a path within the `Entity` rather than `entity.properties`. 

That would mean that _any_ field we add inside the `Entity` in BP, we'd have to add native support for in SQ, and we'd have to check if there were any that we hadn't done that for yet (there are _maybe_ some metadata fields that we didn't bother with yet)

packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx
Comment (Alfred-Mountfield) [reply packages/hash/frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/links-section/link-table/cells/linked-with-cell/entity-selector.tsx:46]: > I think there was some issue with doing this though that we needed to address (those fields being split across multiple tables?)

I think the main thing we need to address is just taking an array of `field`s instead of a single one in the aggregation syntax. Once we do that we can conditionally match on the contents of the fields to identify if something is in `properties` or not.

package.json
```diff
@@ -30,14 +30,14 @@
...
     "fix:eslint": "turbo run --continue fix:eslint",
     "fix:markdownlint": "markdownlint --dot --fix .",
     "fix:prettier": "prettier --write  --ignore-unknown .",
     "fix:yarn-deduplicate": "yarn install && yarn-deduplicate --strategy=fewer && yarn install",
     "lint": "npm-run-all --continue-on-error \"lint:*\"",
     "@todo.2": "Remove the ignore argument below for mock-block-dock once repo is upgraded fully to 0.2 spec",
-    "lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-dep=mock-block-dock --ignore-dep=@blockprotocol/graph --ignore-dep=@blockprotocol/hook --ignore-dep=block-scripts",
+    "lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-dep=mock-block-dock --ignore-dep=@blockprotocol/hook --ignore-dep=@blockprotocol/graph --ignore-dep=@blockprotocol/hook --ignore-dep=block-scripts",
```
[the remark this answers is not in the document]
Alfred-Mountfield: Oops, thank you. I thought I had got an error when I added two duplicate ignores previously but apparently not.

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -183,6 +189,198 @@ export const getAllLatestEntitiesResolver: ResolverFn<
...
+        equal: [{ path: ["archived"] }, { parameter: false }],
+      },
+    ],
+  };
+
+  if (operation.entityTypeId) {
+    filter.all.push({
+      equal: [
+        { path: ["type", "versionedUri"] },
+        { parameter: operation.entityTypeId },
```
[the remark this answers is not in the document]
Alfred-Mountfield: I believe @TimDiekmann said it will _just_ work with `STARTS_WITH`, yes. Alternatively we could add some hard-coded logic to actively spot a type-ID related query joined with a "STARTS_WITH" as structural queries actually support a "base URI" parameter

packages/hash/api/src/graphql/type-defs/knowledge/entity.typedef.ts
```diff
@@ -57,6 +58,20 @@ export const entityTypedef = gql`
...
+    """
+    aggregateEntities(
+      operation: AggregateOperationInput!
+      constrainsValuesOn: OutgoingEdgeResolveDepthInput!
+      constrainsPropertiesOn: OutgoingEdgeResolveDepthInput!
+      constrainsLinksOn: OutgoingEdgeResolveDepthInput!
+      constrainsLinkDestinationsOn: OutgoingEdgeResolveDepthInput!
+      isOfType: OutgoingEdgeResolveDepthInput!
+      hasLeftEntity: EdgeResolveDepthsInput!
+      hasRightEntity: EdgeResolveDepthsInput!
```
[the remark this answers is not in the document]
Alfred-Mountfield: I think it's just so we don't have to add all of the combinations. Until we support incoming depths on things like "isOfType" we won't have all of the depths on all of the queries. `getAllLatestPropertyTypes` will only take `constrainsValuesOn` and `constrainsPropertiesOn` for example

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: It does cover that, but I know other devs on the team don't like that check, and I think they would have recommended I do `!operation.multiSort`, however I think it's better here to be explicit. We're not doing a whole falsey check, we're just checking it's not null or undefined.

CiaranMn: It does, and our linting allows it. I would tend to use `x != null` if I was just writing code for myself, but it relies on people knowing this not exactly obvious behavior. 

We currently leave it up to personal preference – if I had to choose one to enforce I'd probably go for the  longer formulation shown in this diff, as it is clearer and less distracting (I would prioritise clarity over verbosity).

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
TimDiekmann: I added an [Asana task](https://app.asana.com/0/1202805690238892/1203774687353264/f) _(internal)_ to track if we consider it as a bug if the subgraph contains more vertices/edges than requested.

packages/hash/api/src/graphql/type-defs/knowledge/entity.typedef.ts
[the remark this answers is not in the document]
CiaranMn: I mean, why don't we nest all of these arguments under a `graphResolveDepths` field, as we do with `getEntity`?

It's very noisy having them at the top level, and would be easier to parse the request's arguments if you had the context of "this is all `graphResolveDepths` stuff" namespacing them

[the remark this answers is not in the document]
Alfred-Mountfield: We don't do that within `getEntity`?

```graphql
    """
    Get a subgraph rooted at an entity resolved by its id.
    """
    getEntity(
      """
      The id of the entity.
      """
      entityId: EntityId!
      """
      The version of the entity. Defaults to the latest version.
      """
      entityVersion: String
      constrainsValuesOn: OutgoingEdgeResolveDepthInput!
      constrainsPropertiesOn: OutgoingEdgeResolveDepthInput!
      constrainsLinksOn: OutgoingEdgeResolveDepthInput!
      constrainsLinkDestinationsOn: OutgoingEdgeResolveDepthInput!
      isOfType: OutgoingEdgeResolveDepthInput!
      hasLeftEntity: EdgeResolveDepthsInput!
      hasRightEntity: EdgeResolveDepthsInput!
    ): Subgraph!
```

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
CiaranMn: > the BP could support it if we just took an array of fields and said in the spec that it represents a path within the Entity rather than entity.properties.

I think we should do this for `0.3` – mainly thinking of the current special treatment of `entityTypeId` as a bit weird and undesirable.

> That would mean that any field we add inside the Entity in BP, we'd have to add native support for in SQ, and we'd have to check if there were any that we hadn't done that for yet (there are maybe some metadata fields that we didn't bother with yet)

I think we'll want this anyway (though full support in HASH can lag behind a bit)

packages/hash/api/src/graphql/type-defs/knowledge/entity.typedef.ts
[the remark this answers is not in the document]
Alfred-Mountfield: This would introduce _in_consistencies with the other typedefs though. Because as I mentioned, for non-entity queries the selection of depths are different.

So in GraphQL we would have to create multiple objects such as `DataTypeQueriesResolveDepths` and `PropertyTypeQueriesResolveDepths`, etc.

At the time these queries were being written, this seemed to be a lesser evil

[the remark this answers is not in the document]
CiaranMn: >  we would have to create multiple objects such as DataTypeQueriesResolveDepths and PropertyTypeQueriesResolveDepths, etc.

This seems better to me. it seems more natural / ergonomic to nest arguments if you have 7 of them related to one thing rather than spread them out. this is what we've done elsewhere with metadata and everything else.

I guess we can translate between the BP and HASH (since we have to do that for `getEntity` anyway) if it's better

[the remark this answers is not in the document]
Alfred-Mountfield: > this is what we've done elsewhere with metadata and everything else.

Yes but the metadata and such is consistent. We would be creating N permutations of the same object with just varying levels of properties here and GraphQL doesn't allow us to benefit from type composition at all.

> I guess we can translate between the BP and HASH

We are doing that already, no? I think you're just commenting on the GraphQL signature here, which is completely distinct to the signature of the hook, which _does_ take in the depths grouped in an object (`graphResolveDepths`)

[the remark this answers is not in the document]
Alfred-Mountfield: It's sunk cost if we do allow incoming edges on types for example. (Because then you could explore _from_ ontology types _to_ entities and thus you'd need all the edges and we'd have one consistent unified type.

It's also just polluting the space with a bunch of confusingly named types that all look somewhat similar to each other.

If you feel strongly that we should do this I'll put in the 15 minutes it takes to do so, just feels a bit redundant if we're hoping to get rid of most of the GraphQL API with the reintroduction of a collab service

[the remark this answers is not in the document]
CiaranMn: > If you feel strongly that we should do this I'll put in the 15 minutes it takes to do so, just feels a bit redundant if we're hoping to get rid of most of the GraphQL API with the reintroduction of a collab service

On the basis that these aren't sticking around, I agree it would be redundant. I suspect we'll have them for a good while – it's not clear to me that the only public API is going to be a collab service. Let's leave it and see if someone else suggests consolidating them at some point.

> It's also just polluting the space with a bunch of confusingly named types that all look somewhat similar to each other.

They would be the least of it, we're in the process of introducing more 😁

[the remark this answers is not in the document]
Alfred-Mountfield: > we're in the process of introducing more 😁

Let's just slap more on top then ey 😂

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: This still applies in the rebased PR

packages/hash/api/src/graphql/resolvers/index.ts
Alfred-Mountfield: This still applies in the rebased PR, however we now _only_ have `queryEntities` (we do not have competing query endpoints)

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: This is now redundant as pagination was removed

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: This was rectified

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: Unfortunately we diverted priorities and this is still present

packages/hash/api/src/graphql/resolvers/knowledge/entity/entity.ts
Alfred-Mountfield: This has been rectified

packages/hash/frontend/src/graphql/queries/knowledge/entity.queries.ts
Alfred-Mountfield: This is now redundant, I removed it at another point

package.json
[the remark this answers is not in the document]
Alfred-Mountfield: This is redundant

package.json
Alfred-Mountfield: Now redundant

thehabbos007 (commented):
Looks good to me and works locally! (althought should the second test case in the PR description be "OR" instead of "AND"? :) )
App also works as expected with the new query usages, just have a some comments

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
```diff
@@ -0,0 +1,276 @@
...
+      const [pathRoot, ...rest] = field;
+
+      // We check for any specially-handled paths, equivalent to the `EntityQueryToken`s
+
+      // case `EntityQueryToken.Properties`
+      if (pathRoot === "properties") {
+        return ["properties", ...rest];
+      }
+
+      if (pathRoot === "metadata") {
```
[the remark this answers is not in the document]
Alfred-Mountfield: > but we don't provide the constant keys in JSON Schema

What do you mean sorry?

https://blockprotocol.org/types/modules/graph/0.3/schema/entity

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
```diff
@@ -0,0 +1,276 @@
...
+        if (metadataRoot === "archived") {
+          throw new InvalidEntityQueryError(
+            "Filtering by `metadata.archived` is currently not exposed, all queries will return unarchived entities",
+          );
+          // if (metadataRest.length > 0) {
+          //   throw new InvalidEntityQueryError(
+          //     `Invalid filter field path, unable to index inside \`metadata.archived\``,
+          //   );
+          // }
+          // return ["archived"];
```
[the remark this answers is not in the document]
Alfred-Mountfield: A valid question, I'm happy to remove it if you'd prefer (generally I dislike seeing commented out code), but I do think it'd make it easy for us to re-enable it should we decide we don't want to default to removing `archived` entities.

And I'd be _slightly_ concerned we'd forget to do the `if (metadataRest.length > 0)` check, but I'm not suggesting that's a good argument to leave it in.

If anything, the large comment serves as a big indicator that we're introducing a (perhaps unexpected) invariant in this part of the code.

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
```diff
@@ -0,0 +1,276 @@
...
+  );
+};
+
+export const bpMultiFilterToGraphFilter = (bpFilter: MultiFilter): Filter => {
+  const { filters: bpFilters, operator } = bpFilter;
+
+  const filter: Filter = {
+    all: [
+      {
+        equal: [{ path: ["archived"] }, { parameter: false }],
```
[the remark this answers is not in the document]
Alfred-Mountfield: Correct. As `archived` is not part of the BP spec, we don't really have a way for blocks to know they should be filtering them out.. and I believe we intend to use (or are already using) `archive` as the implementation of `deleteEntity` from the spec. Whether that is the long term decision is still up for debate, but while blocks call `deleteEntity`, I don't think we want to be returning such entities back to them..

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
```diff
@@ -0,0 +1,276 @@
...
+type MultiFilter = NonNullable<QueryOperationInput["multiFilter"]>;
+
+export class InvalidEntityQueryError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "InvalidEntityQueryError";
+  }
+}
+
+const bpMultiFilterFieldPathToPathExpression = (
```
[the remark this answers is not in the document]
Alfred-Mountfield: I mean, we can do, but I'd prefer people just look at the `Entity` type, and I'd hope anyone touching this code doesn't need to be pointed to it considering how essential and primitive it is to everything 😅

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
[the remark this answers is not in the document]
thehabbos007: I mean in the query operation https://blockprotocol.org/types/modules/graph/0.3/schema/query-operation
I guess it doesn't have much benefit, just wondering if it's useful to add as a definition in the schema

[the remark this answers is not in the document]
Alfred-Mountfield: The query operation is technically generic across entities and types, so I would not. At least not until that changes :)

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I'll add a bit of text above it so it looks less like commented out code 😅

thehabbos007 (dismissed):

apps/hash-api/src/graph/knowledge/primitive/entity/query.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Provide context to lack of archived filtering](https://github.com/hashintel/hash/pull/1844/commits/e9df8d00bce8dc4da6a6eb4de06e6a6434a2461a)

thehabbos007 (dismissed):

thehabbos007 (approved):

thehabbos007 (approved):

thehabbos007 (approved):

PR #1848, opened 2023-01-16
Correctly resolve time intervals when traversing the Graph

TimDiekmann:

We correctly resolve the roots of the subgraph based on the `timeProjection` passed to the structural query. It's however not required to resolve all edges for the graph.

This changes the logic to only resolve edges where the edge exists for the vertex in both, the vertex time interval (i.e. the entity version) and the time interval provided by the time projection. Effectively, this is an intersection of these two intervals.

Applying this logic reduces the number of queries issued to the Graph when running the REST API tests by roughly 15% (with only very few versioned data).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203701389454316/f) _(internal)_

## 🔍 What does this change?

- Some operations were required to convert between the different temporal structs, c679ae5df55cbf822cf54c94030226bc15669b65 adds those
- The `DependencMap` was changed to identify already resolved intervals and report which intervals have to be resolved further. The logic is implemented in d0a0810110ee6c4ba6342e1d43c6d24b43862404
- faad89fd19b2d2f5c0e41a215ca958574e909dd1 uses the new interface defined in the two commits above. As mentioned above, the time projection's image will be intersected with the version of the entity, which then will be used in the dependency map. When the given resolve depths and intervals are not sufficiently resolved, information is returned on how to traverse the graph further.

Some extensive documentation has been added to the `DependencyMap` update logic. A follow-up PR requires some renaming changes. With this in place, the documentation will be slightly enhanced as well.

## ⚠️ Known issues

In some rare edge cases, it's possible, that the subgraph is expected to resolve an entity, which was already resolved, but the resolved interval and the newly requested interval are disjoint. In this case, we couldn't come up with a proper representation (Postgres' multirange would probably be an option), so this PR handles this case by merging the two intervals into one big interval. This may result in a few more edges, which will be resolved. I don't consider this as a big issue as this is only inside of _a single entity_ based on its vertex id. I added a warning for this specific case so if we encounter this warning frequently, we may want to further take a look.

## 🐾 Next steps

- Rename `TimeProjection` to `TimeAxes`
- Rename `TimeProjection::kernel` to `TimeAxes::pinned`
- Rename `TimeProjection::image` to `TimeAxes::variable`

## 🛡 What tests cover this?

All tests, which are using the subgraph, including
- Graph Integration tests
- Rest API tests
- backend integration tests
- playwright tests
- manual testing 🙃

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
@@ -13,7 +13,7 @@ use crate::{
     identifier::{
         knowledge::{EntityEditionId, EntityId, EntityRecordId, EntityVersion},
         ontology::OntologyTypeEditionId,
-        time::{DecisionTime, Timestamp, VersionInterval},
+        time::{DecisionTime, TimeProjection, Timestamp, VersionInterval},
         EntityVertexId,
     },
     knowledge::{Entity, EntityLinkOrder, EntityMetadata, EntityProperties, EntityUuid, LinkData},
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
         entity_vertex_id: EntityVertexId,
         dependency_context: &'a mut DependencyContext,
         subgraph: &'a mut Subgraph,
-        current_resolve_depth: GraphResolveDepths,
+        mut resolve_depths: GraphResolveDepths,
+        time_projection: TimeProjection,
     ) -> Pin<Box<dyn Future<Output = Result<(), QueryError>> + Send + 'a>> {
         async move {
-            let dependency_status = dependency_context
-                .knowledge_dependency_map
-                .insert(&entity_vertex_id, current_resolve_depth);
-
             let time_axis = subgraph.resolved_time_projection.image_time_axis();
 
-            let entity: &Entity = match dependency_status {
-                DependencyStatus::Unresolved => {
-                    match entity_vertex_id.subgraph_vertex_entry(subgraph) {
-                        RawEntryMut::Occupied(entry) => entry.into_mut(),
-                        RawEntryMut::Vacant(_) => {
-                            // Entities are always inserted into the subgraph before they are
-                            // resolved, so this should never happen. If it does, it is a bug.
-                            unreachable!("entity should already be in the subgraph")
-                        }
-                    }
+            let entity: &Entity = match entity_vertex_id.subgraph_vertex_entry(subgraph) {
+                RawEntryMut::Occupied(entry) => entry.into_mut(),
+                RawEntryMut::Vacant(_) => {
+                    // Entities are always inserted into the subgraph before they are resolved, so
+                    // this should never happen. If it does, it is a bug.
+                    unreachable!("entity should already be in the subgraph")
                 }
+            };
 
+            let version_interval = entity
+                .metadata()
+                .version()
+                .projected_time(time_axis)
+                .into_time_interval();
+
+            let Some(mut time_projection) = time_projection.intersect_image(version_interval) else {
+                // The entity is not in the time projection, so we don't need to resolve it.
+                // This should never happen as the time projection is based on this entity's
+                // version interval, but we emit a warning just in case.
+                tracing::warn!(
+                    entity_id = ?entity_vertex_id.base_id(),
+                    "entity is not in the time projection"
+                );
+                return Ok(());
+            };
+
+            let dependency_status = dependency_context.knowledge_dependency_map.update(
+                &entity_vertex_id,
+                resolve_depths,
+                time_projection.image(),
+            );
+
+            match dependency_status {
+                DependencyStatus::Unresolved(depths, interval) => {
+                    // The dependency may have to be resolved more than anticipated, so we update
+                    // the resolve depth and time projection.
+                    resolve_depths = depths;
+                    time_projection.set_image(interval);
+                }
                 DependencyStatus::Resolved => return Ok(()),
             };
 
-            if current_resolve_depth.is_of_type.outgoing > 0 {
+            if resolve_depths.is_of_type.outgoing > 0 {
                 let entity_type_id =
                     OntologyTypeEditionId::from(entity.metadata().entity_type_id());
                 subgraph.edges.insert(Edge::KnowledgeGraph {
@@ -88,20 +112,21 @@ impl<C: AsClient> PostgresStore<C> {
                     subgraph,
                     GraphResolveDepths {
                         is_of_type: OutgoingEdgeResolveDepth {
-                            outgoing: current_resolve_depth.is_of_type.outgoing - 1,
-                            ..current_resolve_depth.is_of_type
+                            outgoing: resolve_depths.is_of_type.outgoing - 1,
+                            ..resolve_depths.is_of_type
                         },
-                        ..current_resolve_depth
+                        ..resolve_depths
                     },
+                    time_projection.clone(),
                 )
                 .await?;
             }
 
-            if current_resolve_depth.has_left_entity.incoming > 0 {
+            if resolve_depths.has_left_entity.incoming > 0 {
                 for outgoing_link_entity in <Self as Read<Entity>>::read(
                     self,
                     &Filter::for_outgoing_link_by_source_entity_vertex_id(entity_vertex_id),
-                    &subgraph.resolved_time_projection,
+                    &time_projection,
                 )
                 .await?
                 {
@@ -125,21 +150,22 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph,
                         GraphResolveDepths {
                             has_left_entity: EdgeResolveDepths {
-                                incoming: current_resolve_depth.has_left_entity.incoming - 1,
-                                ..current_resolve_depth.has_left_entity
+                                incoming: resolve_depths.has_left_entity.incoming - 1,
+                                ..resolve_depths.has_left_entity
                             },
-                            ..current_resolve_depth
+                            ..resolve_depths
                         },
+                        time_projection.clone(),
                     )
                     .await?;
                 }
             }
 
-            if current_resolve_depth.has_right_entity.incoming > 0 {
+            if resolve_depths.has_right_entity.incoming > 0 {
                 for incoming_link_entity in <Self as Read<Entity>>::read(
                     self,
                     &Filter::for_incoming_link_by_source_entity_vertex_id(entity_vertex_id),
-                    &subgraph.resolved_time_projection,
+                    &time_projection,
                 )
                 .await?
                 {
@@ -163,21 +189,22 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph,
                         GraphResolveDepths {
                             has_right_entity: EdgeResolveDepths {
-                                incoming: current_resolve_depth.has_right_entity.incoming - 1,
-                                ..current_resolve_depth.has_right_entity
+                                incoming: resolve_depths.has_right_entity.incoming - 1,
+                                ..resolve_depths.has_right_entity
                             },
-                            ..current_resolve_depth
+                            ..resolve_depths
                         },
+                        time_projection.clone(),
                     )
                     .await?;
                 }
             }
 
-            if current_resolve_depth.has_left_entity.outgoing > 0 {
+            if resolve_depths.has_left_entity.outgoing > 0 {
                 for left_entity in <Self as Read<Entity>>::read(
                     self,
                     &Filter::for_left_entity_by_entity_vertex_id(entity_vertex_id),
-                    &subgraph.resolved_time_projection,
+                    &time_projection,
                 )
                 .await?
                 {
@@ -201,21 +228,22 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph,
                         GraphResolveDepths {
                             has_left_entity: EdgeResolveDepths {
-                                outgoing: current_resolve_depth.has_left_entity.outgoing - 1,
-                                ..current_resolve_depth.has_left_entity
+                                outgoing: resolve_depths.has_left_entity.outgoing - 1,
+                                ..resolve_depths.has_left_entity
                             },
-                            ..current_resolve_depth
+                            ..resolve_depths
                         },
+                        time_projection.clone(),
                     )
                     .await?;
                 }
             }
 
-            if current_resolve_depth.has_right_entity.outgoing > 0 {
+            if resolve_depths.has_right_entity.outgoing > 0 {
                 for right_entity in <Self as Read<Entity>>::read(
                     self,
                     &Filter::for_right_entity_by_entity_vertex_id(entity_vertex_id),
-                    &subgraph.resolved_time_projection,
+                    &time_projection,
                 )
                 .await?
                 {
@@ -239,11 +267,12 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph,
                         GraphResolveDepths {
                             has_right_entity: EdgeResolveDepths {
-                                outgoing: current_resolve_depth.has_right_entity.outgoing - 1,
-                                ..current_resolve_depth.has_right_entity
+                                outgoing: resolve_depths.has_right_entity.outgoing - 1,
+                                ..resolve_depths.has_right_entity
                             },
-                            ..current_resolve_depth
+                            ..resolve_depths
                         },
+                        time_projection.clone(),
                     )
                     .await?;
                 }
@@ -442,19 +471,20 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
         let StructuralQuery {
             ref filter,
             graph_resolve_depths,
-            ref time_projection,
+            time_projection: ref unresolved_time_projection,
         } = *query;
+
+        let time_projection = unresolved_time_projection.clone().resolve();
+        let time_axis = time_projection.image_time_axis();
 
         let mut subgraph = Subgraph::new(
             graph_resolve_depths,
+            unresolved_time_projection.clone(),
             time_projection.clone(),
-            time_projection.clone().resolve(),
         );
         let mut dependency_context = DependencyContext::default();
-        let time_axis = subgraph.resolved_time_projection.image_time_axis();
 
-        for entity in Read::<Entity>::read(self, filter, &subgraph.resolved_time_projection).await?
-        {
+        for entity in Read::<Entity>::read(self, filter, &time_projection).await? {
             let vertex_id = entity.vertex_id(time_axis);
             // Insert the vertex into the subgraph to avoid another lookup when traversing it
             subgraph.insert(&vertex_id, entity);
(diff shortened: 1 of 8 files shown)


packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+                    // resolved for the full interval anyway.
+                    DependencyStatus::Unresolved(*current_depths, current_interval.clone())
+                } else if old_interval.contains_interval(&new_interval) {
+                    tracing::info!("{old_interval:?} contains {new_interval:?}");
+                    // The dependency is already resolved for the required interval
+                    // old: [-----)
+                    // new:  [---)
                     DependencyStatus::Resolved
+                } else if new_interval.contains_interval(&old_interval)
+                    | new_interval.is_adjacent_to(&old_interval)
```
thehabbos007, line 125: Are we intentionally using `bitwise or` here?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+                    DependencyStatus::Unresolved(
+                        *current_depths,
+                        difference
+                            .into_iter()
+                            .next()
+                            .expect("difference must be a single interval"),
+                    )
+                } else {
+                    // The time intervals are disjoint. Ideally, we only would resolve the new
+                    // interval, but we did not came up with a good way to store the different
```
thehabbos007, line 152:
```suggestion
                    // interval, but we did not come up with a good way to store the different
```

thehabbos007 (commented):
Thank you Tim. I have some comments, and lets await Alfie's review as well. There was nothing that stood out as problematic to me though :)

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
...
+                resolve_depths,
+                time_projection.image(),
+            );
+
+            match dependency_status {
+                DependencyStatus::Unresolved(depths, interval) => {
+                    // The dependency may have to be resolved more than anticipated, so we update
+                    // the resolve depth and time projection.
+                    resolve_depths = depths;
+                    time_projection.set_image(interval);
```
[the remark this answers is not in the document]
TimDiekmann: Generally, we should rely on the return value of `DependencyMap::update` to decide which parameters to use, but, yes, `depths` is always greater than or equal to `resolve_depths`. This may change with [this Asana task](https://app.asana.com/0/1201095311341924/1203763222255827/f) _(internal)_.

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/data_type.rs
```diff
@@ -25,15 +25,22 @@ impl<C: AsClient> PostgresStore<C> {
...
+            data_type_id,
+            resolve_depths,
+            time_projection.image(),
+        );
 
-        let _data_type = match dependency_status {
-            DependencyStatus::Unresolved => {
-                let time_projection = subgraph.resolved_time_projection.clone();
+        #[expect(unused_assignments, unused_variables)]
+        let data_type = match dependency_status {
```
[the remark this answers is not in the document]
TimDiekmann: I wanted to make sure, that `resolve_depths` is updated, so we can just add the new code below. This requires the `unused_assignments` lint so I decided to just ignore the `unused_variable` lint as well. Happy to change this to
```suggestion
        #[expect(unused_assignments)]
        let _data_type = match dependency_status {
```

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
...
+                }
```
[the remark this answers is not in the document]
TimDiekmann: Yes, as the ontology types do not have a time interval associated, so we have nothing to `.intersect` with.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: This was by accident, locally I had a `match` statement here. The result is the same, but `||` is safer: c14634ee8ceebfac2e0a75ddf9228679741fbe35

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+                );
+                DependencyStatus::Unresolved(new_resolve_depth, new_interval)
             }
             RawEntryMut::Occupied(entry) => {
-                if entry.into_mut().update(resolved_depth) {
-                    DependencyStatus::Unresolved
-                } else {
+                let (current_depths, current_interval) = entry.into_mut();
+                let old_interval = current_interval.clone();
+                *current_interval = current_interval.clone().merge(new_interval.clone());
```
Alfred-Mountfield, line 112: Is the `merge` safe to use here (rather than `union`)? 

My guess is that we aren't confident right now that the intervals will be contiguous, so this feels like it would return subtly incorrect information.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+                    // old: [-----)
+                    // new:  [---)
                     DependencyStatus::Resolved
+                } else if new_interval.contains_interval(&old_interval)
+                    || new_interval.is_adjacent_to(&old_interval)
+                {
+                    // The dependency is already resolved, but not for the required interval or the
+                    // intervals are adjacent so we only have to resolve the new interval
+                    // old:  [---)   |  [---)
+                    // new: [-----)  |      [---)
```
Alfred-Mountfield, line 130: This is another location where I think we should highlight that this isn't _necessarily_ optimum, and in the first case we'd ideally just resolve the symmetric difference.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+
+                    DependencyStatus::Unresolved(
+                        *current_depths,
+                        difference
+                            .into_iter()
+                            .next()
+                            .expect("difference must be a single interval"),
+                    )
+                } else {
+                    // The time intervals are disjoint. Ideally, we only would resolve the new
```
Alfred-Mountfield, line 151: The intervals can be disjoint as well above, right? If we're going to bother to start the comment with this line I think we should try and be comprehensive. Is the correct statement that they're disjoint and non-adjacent?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+
+                    // This case is expected to happen very rarely, so falling back to this logic
+                    // is fine for now. If it turns out to be a performance bottleneck, we can
+                    // find and implement a better solution. To track this, we log a warning.
+                    tracing::warn!(
+                        vertex_id=?identifier,
+                        ?old_interval,
+                        ?new_interval,
+                        "dependency is resolved for disjoint intervals"
+                    );
```
Alfred-Mountfield, line 163: I appreciate you may want to have this here for us to track how frequently it happens but I don't know if I'm comfortable having a tracing statement, with such a high-priority log level (this will appear even if we disable `info` logging) that isn't gated behind a feature or something, and one that is in a potentially majorly hot loop. This _could_ happen tens of thousands of times _per request_.

The log line is also really not self-descriptive, I'd have to actively search for it in the codebase to have any clue what it's about.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
+
+                    // We only require this logic when traversing edges of the graph, so the roots
+                    // of the graph are always precisely resolved correctly. By this implementation,
+                    // we may get a few more dependencies resolved than necessary, which will appear
+                    // as vertices/edges in the subgraph. As we do not guarantee that the subgraph's
+                    // vertices/edges are minimal, this is not a problem.
```
Alfred-Mountfield, line 169: I disagree with this, and I thought we explicitly talked about this in a call. I _do_ view this as a problem. If we want to accept this as a "feature not a bug" then I think we need to talk about the ramifications for this, I can easily see a lot of block developers creating UIs to display the "number of friends of a person" for example, which would be based on the elements of the edge set. If you cannot rely on that, then I do not know what you can correctly rely on. This seems like a foundational essential part of having a temporal datastore.

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
         entity_vertex_id: EntityVertexId,
         dependency_context: &'a mut DependencyContext,
         subgraph: &'a mut Subgraph,
-        current_resolve_depth: GraphResolveDepths,
+        mut resolve_depths: GraphResolveDepths,
```
Alfred-Mountfield, line 49: I somewhat liked the previous definition of `current_resolve_depth`(s) because it had some implication about it being temporary (as in "current"), and considering we mutate the heck out of these depths as we pass them in, this feels a bit like an unnecessary change.

I don't feel strongly, not asking for it to be changed back, just thought I'd highlight it.

Alfred-Mountfield (changes requested):

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
...
+                .into_time_interval();
+
+            let Some(mut time_projection) = time_projection.intersect_image(version_interval) else {
+                // The entity is not in the time projection, so we don't need to resolve it.
+                // This should never happen as the time projection is based on this entity's
+                // version interval, but we emit a warning just in case.
+                tracing::warn!(
+                    entity_id = ?entity_vertex_id.base_id(),
+                    "entity is not in the time projection"
+                );
```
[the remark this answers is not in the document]
TimDiekmann: `traverse_entity` is called with the returned entities from `read` with `time_projection`. This implies, that the version interval of `entity` is overlapping with `time_projection`. `intersect_image` returns `None` if there are no overlapping points, so this should never happen. I'm happy to make this an error.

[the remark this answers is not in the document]
TimDiekmann:
```suggestion
                // `traverse_entity` is called with the returned entities from `read` with
                // `time_projection`. This implies, that the version interval of `entity` is
                // overlapping with `time_projection`. `intersect_image` returns `None` if there are
                // no overlapping points, so this should never happen.
                unreachable!("the version interval of the entity does not overlap with the time projection");
```

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
...
 
+            let dependency_status = dependency_context.knowledge_dependency_map.update(
+                &entity_vertex_id,
+                resolve_depths,
+                time_projection.image(),
+            );
+
+            match dependency_status {
+                DependencyStatus::Unresolved(depths, interval) => {
+                    // The dependency may have to be resolved more than anticipated, so we update
```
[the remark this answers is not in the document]
TimDiekmann: We expected, that `entity` has to be resolved for depths of `resolve_depths` and the intersection of the time projection and the entity version interval. The dependency map may return a higher resolve depths or a larger time projection. This is captured in the documentation of `DependencyStatus` and `DependencyMap::update`.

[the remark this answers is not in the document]
TimDiekmann: I added a few more lines to this: 050b4795d1bff9300d8c0958af0056dcf72aeb6f

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -46,31 +46,55 @@ impl<C: AsClient> PostgresStore<C> {
...
                 }
+            };
+
+            let version_interval = entity
+                .metadata()
+                .version()
+                .projected_time(time_axis)
+                .into_time_interval();
+
+            let Some(mut time_projection) = time_projection.intersect_image(version_interval) else {
```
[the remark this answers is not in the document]
TimDiekmann: I did both, a better name and a comment describing the logic: 669cf15657c85f93428651c007d60e2cf65a8575

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: I hear you and the reasoning makes sense. Happy to revert it

TimDiekmann: Reverted (and added an `s`) in ec9a2c45b28142e03fabb6bfa62a5ea8ecdad598

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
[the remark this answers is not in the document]
Alfred-Mountfield: If we're confident this would be a bug, then the suggestion seems fine. Ideally we'd have _some_ confidence that automated tests would pick it up if that were the case

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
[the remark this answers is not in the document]
Alfred-Mountfield: I'm not sure that's more than _anticipated_ though. That would imply we wouldn't expecting to have to resolve it more, but we don't really have an expectation of one or the other

packages/graph/hash_graph/lib/graph/src/store/postgres/knowledge/entity/mod.rs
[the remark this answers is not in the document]
Alfred-Mountfield: That's probably fine, definitely better. I'd still prefer something that helps the reader understand it's the overlap of the entity's version with the timespan that was given as an input.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: `merge` cannot fail. If the intervals do not overlap or are not adjacent, the points in between will also be included. 
pseudo code:
```
merged = (min(lhs.lower, rhs.lower), max(lhs.upper, rhs.upper))
```

Alfred-Mountfield: I know which is why I said about them being contiguous and returning incorrect information. And also is why I spoke about `union` :)

Alfred-Mountfield: This seems related to the discussion below about returning too much information being a bug or not

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -70,34 +80,115 @@ impl<K> Default for DependencyMap<K> {
...
-                if entry.into_mut().update(resolved_depth) {
-                    DependencyStatus::Unresolved
-                } else {
+                let (current_depths, current_interval) = entry.into_mut();
+                let old_interval = current_interval.clone();
+                *current_interval = current_interval.clone().merge(new_interval.clone());
+
+                if current_depths.update(new_resolve_depth) {
+                    // It does not matter if the interval is updated or not, as it has to be
+                    // resolved for the full interval anyway.
```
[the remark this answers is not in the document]
TimDiekmann: I improved the comment: aa2451486c77da1801ea14c4d1aa739537d2a6ee

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: Improved the comment: 14e2eb59923d783be881a75cc221d065e038cf6f

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: Actually, the comment above was wrong. I fixed the comment above in d2e49fd417e2928f50fb1a454ad09996db9f44a4 and added, that the intervals are not adjacent either in b02de604b0598abcc5cadc1a0042149d12d51956.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: Do you think it's fine to leave it as a `tracing::debug!` log and create a task to capture this?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
Alfred-Mountfield: This one's great!

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
Alfred-Mountfield: Thank you :)

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
Alfred-Mountfield: I think I'd err on the side of removing the log for now and creating a task to figure this out properly. (One we can talk about in refinement and perhaps make a task to establish a metrics collection strategy) 
cc: @thehabbos007

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
Alfred-Mountfield: We've spoken about this more on a call. We're returning too much information in `main` at the moment anyway, so we will probably address this in a follow-up if we feel it is worth addressing (as this PR shouldn't be a regression).

TimDiekmann:
Thanks a lot @Alfred-Mountfield! I tweaked the comments further.

Alfred-Mountfield (approved):

Alfred-Mountfield (approved):

PR #1850, opened 2023-01-16
Update version of property types

luisbettencourt:

This PR implements updating the version of out of date property types and disables editing for out of date property types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203198755412050/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203543021352041/1203673362090856/f) _(internal)_

## 🚫 Blocked by

- [x] #1815 

## 🔍 What does this change?

- The entity type editor is now getting the property types relevant to the entity type in order to display them in the properties table;
- Added a `usePropertyTypeVersions` to get the current and latest version of a property type;
- Added an update version button when the property type is out of date;
- Disabled the edit button for out of date property types.

## 📹 Demo

image


packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
```diff
@@ -19,10 +25,47 @@ import {
   useFetchEntityTypes,
 } from "../../../../../shared/entity-types-context/hooks";
 
+interface EntityTypeAndPropertyTypes {
+  entityType: EntityType;
+  propertyTypes: Record<string, PropertyType>;
+}
+
+const getPropertTypes = (
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:33]: Typo
```suggestion
const getPropertyTypes = (
```

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
```diff
@@ -19,10 +25,47 @@ import {
...
+  properties: any,
+  subgraph: Subgraph,
+  propertyTypes?: Map<string, PropertyType>,
+) => {
+  let propertyTypesMap = propertyTypes ?? new Map<string, PropertyType>();
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:38]: `getPropertyTypesReferencedByEntityType` in the stdlib might help simplify this – it will return an array of property type ids referenced by an entity type, given a subgraph and an entity type id. 

This will include property types referenced by property types, which you might not need for this purpose, but probably don't hurt to include.

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
```diff
@@ -164,6 +171,33 @@ const DisabledCheckboxCell = ({
...
+    const baseUri = extractBaseUri(propertyTypeId);
+
+    const propertyTypeIds = propertyTypes
+      ? Object.values(propertyTypes).map(({ $id }) => $id)
+      : [];
+
+    const versions = propertyTypeIds
+      .filter((versionedUri) => {
+        return versionedUri.startsWith(`${baseUri}v/`);
+      })
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:188]: If you had the `subgraph` here this could be replaced by `getPropertyTypesByBaseUri(subgraph, baseUri)`

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
```diff
@@ -366,7 +421,13 @@ export const PropertyTypeRow = ({
...
 
   const propertyTypes = usePropertyTypes();
-  const property = propertyTypes?.[$id];
+  const { propertyTypes: entityTypePropertyTypes } = useEntityType();
+  const property = entityTypePropertyTypes[$id] ?? propertyTypes?.[$id];
+
+  const [currentVersion, latestVersion] = usePropertyTypeVersions(
+    $id,
+    propertyTypes,
+  );
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:430]: These would always be the same if it isn't already on the entity, right?

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
```diff
@@ -19,10 +25,47 @@ import {
...
+      if (propertyType) {
+        propertyTypesMap.set(propertyUri, propertyType);
+      }
+    }
+  }
+
+  return propertyTypesMap;
+};
+
 export const useEntityTypeValue = (
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:65]: I think we're approaching (or at) the point where it'd be better to just have a subgraph we can pass around and extract things from as needed, since we already have helper utilities to do so – but I appreciate that might involve a lot of changes, and be incompatible with how we currently store draft state.

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:38]: It returns only the `baseId` and `version` of the property type in the entity type, and also does not return property types for `depth > 0`.

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx]: Fixed typo in ac19ecbff4056f0befa28dc47a0762441553bb0b

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Using  `getPropertyTypesByBaseUri(subgraph, baseUri)` in 60c5ba7e3bc80cd4a33cae6ec47d2ec8103afb9e

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: If the property type is being added you mean?

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:65]: In the end we just need the property types relativant to the entity type, which we get here. Is there something else I'm missing?

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Yes – if this property type only exists in `propertyTypes` then presumably we already have the latest (unless `propertyTypes` can end up with multiple versions of the same one). Not a big deal though

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
```diff
@@ -39,6 +46,7 @@ import { useBlockProtocolCreatePropertyType } from "../../../../../../components
 import { useBlockProtocolUpdatePropertyType } from "../../../../../../components/hooks/block-protocol-functions/ontology/use-block-protocol-update-property-type";
 import { StyledPlusCircleIcon } from "../../../../shared/styled-plus-circle-icon";
 import { useRouteNamespace } from "../../../../shared/use-route-namespace";
+import { useEntityType } from "../shared/entity-type-context";
 import { EntityTypeEditorForm } from "../shared/form-types";
 import {
   usePropertyTypes,
```
Comment (CiaranMn) [inline packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:52]: Can we rename this to `useLatestPropertyTypes`, if that's accurate? I think it would be easier to tell what's going on when we have both `entityTypesAndPropertyTypes` and `propertyTypes`

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: There can be multiple versions of the same one inside `propertyTypes`, by updating a property type (the previous and new version get stored there). In the end I don't think it's worth doing a lot optimizations here since we won't be fetching all the latest property types in the future anyway, right?

Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Sure, I don't think it's worth it

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
Comment (luisbettencourt) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx]: Renamed in 3d5787116febb8c5d3726720beb99487ac471e1d

packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
Comment (CiaranMn) [reply packages/hash/frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:65]: I was just thinking that we do some extraction of property types from a subgraph here, and then do more manipulation / extraction of them elsewhere, and maybe we could reduce some of it. But maybe we can't – it's fine as it is

CiaranMn (approved):

CiaranMn (approved):

PR #1851, opened 2023-01-16
Enable exit code 1 in `yarn lint:license-in-workspaces`

kachkaev:

Continues https://github.com/hashintel/hash/pull/1825, makes sure that CI fails when `LICENSE.md` is missing in a workspace.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203669287079639/f) _(internal)_

## 🛡 What tests cover this?

- CI

## 📹 Demo

https://github.com/hashintel/hash/actions/runs/3961597906/jobs/6787220355#step:7:119

Screenshot 2023-01-19 at 19 17 15


libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
```diff
@@ -49,7 +49,7 @@ const script = async () => {
 
   for (const licenseFilePath of licenseFilePaths) {
     const licenseFileName = path.basename(licenseFilePath);
-    if (!licenseFileName.match(/^LICENSE(-[A-Z0-9]+)?.md$/)) {
+    if (!licenseFileName.match(/^LICENSE(-[A-Z0-9]+)?(\.(md|txt))?$/)) {
```
kachkaev, line 52: Converts

```
🔴 [NAMING]  apps/engine/lib/memory/src/arrow/util/LICENSE
🔴 [NAMING]  apps/engine/lib/memory/src/arrow/ipc/serialize/LICENSE
```

```
🟡 [EXTRA]   apps/engine/lib/memory/src/arrow/util/LICENSE
🟡 [EXTRA]   apps/engine/lib/memory/src/arrow/ipc/serialize/LICENSE
```

These two files do not use markdown are external, which seems reasonable to tolerate. LICENSE files within worksppaces still need to have `md` extension.

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
```diff
@@ -134,8 +134,11 @@ const script = async () => {
     }
   }
 
-  if (checkFailed) {
+  if (checkFailed || extraLicenseFilesArePresent) {
```
kachkaev, line 137: Adds empty line when check has passed, but with warnings (slightly improves output)

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
```diff
@@ -151,10 +154,9 @@ const script = async () => {
     );
   }
 
-  // @todo Uncomment when all workspaces have a license file
-  // if (checkFailed) {
-  //   process.exit(1);
-  // }
+  if (checkFailed) {
+    process.exit(1);
+  }
```
kachkaev, line 159: The main point of this PR

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
nonparibus: I feel like this should conditionally read 'EXTRA' only when the files exist outside of an expected workspace location, and if they occur within one they should be red 'NAMING' errors.

These are fine to mark as extra, though, as they are the first case of the two.

nonparibus (approved):
Legal LGTM ✅  Have taken a look at the logic and shared thoughts. Somebody else should take a peek, too.

CiaranMn (approved):

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
kachkaev: Possibly, I just think that any further improvements have diminishing returns so are not worth extra time, at least in the near future. If you disagree, happy to pick a new ticket!

nonparibus: If not simple enough to ship as part of this (which has now already merged, so nm) we can ticketize for a future cleanup week. See [ticket](https://app.asana.com/0/1203193968491715/1203767287750344/f) (_internal link_).

PR #1853, opened 2023-01-16
Use `null` instead of `undefined` for time related structs

TimDiekmann:

In https://github.com/hashintel/hash/pull/1816#discussion_r1071443624 was raised, that using `null` instead of `undefined` is better suited. This PR makes this change.

## 🔗 Related links

- https://github.com/hashintel/hash/pull/1816#discussion_r1071443624

## 🔍 What does this change?

- e3d54e3e5cbd4690bb4d085f53efa060c63b4e6f: Update `utoipa` as the `nullable` field was added to `OneOf` after the `2.4.2` release
- 72081c1f1748a167cf5be1392e18b7b56cb3fd61: Adjust time related structs in the Graph to generate `| null` instead
- 83303bdfdea5f1a34d7e4555a53003bb3f92cf1c: Adjust workspace to use `null` instead of `undefined`


packages/graph/clients/typescript/api.ts
```diff
@@ -2078,36 +2078,23 @@ export interface UnresolvedDecisionTimeImage {
   axis: DecisionTime;
   /**
    *
-   * @type {TimeIntervalBound}
+   * @type {UnresolvedDecisionTimeImageEnd}
    * @memberof UnresolvedDecisionTimeImage
    */
-  end?: TimeIntervalBound;
```
thehabbos007, line 2084: Was this wrongly typed before?

packages/graph/clients/typescript/api.ts
```diff
@@ -2078,36 +2078,23 @@ export interface UnresolvedDecisionTimeImage {
...
+  end: UnresolvedDecisionTimeImageEnd | null;
   /**
    *
-   * @type {TimeIntervalBound}
+   * @type {UnresolvedDecisionTimeImageEnd}
    * @memberof UnresolvedDecisionTimeImage
    */
-  start?: TimeIntervalBound;
+  start: UnresolvedDecisionTimeImageEnd | null;
```
thehabbos007, line 2090: The property here is `start` but the type is `UnresolvedDecisionTimeImageEnd`. Is the struct itself misnamed?

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/projection.rs
```diff
@@ -71,35 +89,63 @@ pub type UnresolvedDecisionTimeImage = UnresolvedImage<DecisionTime>;
...
 impl ToSchema for UnresolvedTransactionTimeImage {
     fn schema() -> openapi::RefOr<openapi::Schema> {
-        openapi::Schema::AllOf(
-            openapi::AllOfBuilder::new()
-                .item(
-                    openapi::ObjectBuilder::new()
-                        .property("axis", openapi::Ref::from_schema_name("TransactionTime"))
-                        .required("axis"),
-                )
-                .item(UnresolvedTimeInterval::<DecisionTime>::schema())
```
thehabbos007: Curious about the change made here. Instead of `allOf`-ing the `UnresolvedTimeInterval` we now explicitly define `start` and `end` with nullable `TimeIntervalBound`. Aren't these equivalent or did the schema of `UnresolvedTimeInterval` turn the `Option`s into `undefined`?

thehabbos007 (commented):
Hey, got a couple comments :)

packages/graph/clients/typescript/api.ts
TimDiekmann: `utoipa` wrongly generates the type parameter, but it's aliased to `TimeIntervalBound`, which is correct.

packages/graph/hash_graph/lib/graph/src/shared/identifier/time/projection.rs
TimDiekmann: The `Option` turns it into `undefined`. Also, the `allOf` generates those weird duplicated definitions. I will test if I can tweak this a little bit and report back.

TimDiekmann:
Thanks @thehabbos007! I was able to solve the issue of weird intermediate names. Could you have another look?

packages/graph/clients/typescript/api.ts
```diff
@@ -2078,23 +2078,36 @@ export interface UnresolvedDecisionTimeImage {
...
 }
 /**
- * @type UnresolvedDecisionTimeImageEnd
+ *
  * @export
+ * @interface UnresolvedDecisionTimeImageAllOf
  */
-export type UnresolvedDecisionTimeImageEnd = TimeIntervalBound;
-
+export interface UnresolvedDecisionTimeImageAllOf {
```
thehabbos007, line 2097: This seems unused?

packages/graph/hash_graph/lib/graph/src/api/rest/mod.rs
```diff
@@ -462,6 +462,14 @@ impl Modify for TimeSchemaAddon {
             components
                 .schemas
                 .insert("Timestamp".to_owned(), Timestamp::<()>::schema());
+            components.schemas.insert(
+                "NullableTimestamp".to_owned(),
+                ObjectBuilder::new()
+                    .schema_type(SchemaType::String)
+                    .format(Some(SchemaFormat::KnownFormat(KnownFormat::DateTime)))
```
thehabbos007, line 469: I suppose it's intentional that we are using a `string` with a `DateTime` format, I would have just thought the client generator would create better types from this :laughing: 

Just to clarity `Timestamp` is also just `string` in the output, right?

packages/graph/clients/typescript/api.ts
TimDiekmann: I don't know how you can comment on that line, possible via an old commit in this PR, but this line is not touched by the PR anymore. However, you are right. That's the same issue as https://github.com/hashintel/hash/pull/1738#discussion_r1063429672

packages/graph/hash_graph/lib/graph/src/api/rest/mod.rs
TimDiekmann: Yeah, I hoped that as well 😢 

Yes, `TimeStamp` is defined the same way, but without `.nullable`. In fact, I copied over their implementation. Which makes me wonder if I can reuse it somehow.

TimDiekmann: Sadly, this would generate `type NullableTimeinterval = string`, so it's probably best to just keep the current approach

thehabbos007 (approved):
Looks good to me. I hope there are no API problems with unexpected types as a result of this PR, i think we can rely on the backend integration tests for the most part for these changes

PR #1855, opened 2023-01-17
Editing object property types doesn't display required field correctly

luisbettencourt:

When editing an object property type with one of its properties with required set to true, it won't display the correct state of the required checkbox. This PR fixes that.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203720352875052/f) _(internal)_

## 🔍 What does this change?

- The required fields array consists of a list of `baseUri`s and we were comparing it against `versionedUri`s.


kachkaev (approved):

nathggns:
good catch!

PR #1859, opened 2023-01-17
Automate Rust nightly toolchain version upgrades

indietyp:

This is an attempt to bump toolchain versions using renovate automatically. Right now, this is done manually and sporadically.

There have been attempts (https://github.com/renovatebot/renovate/issues/11488) to support this, but currently, renovate does not natively support `rust-toolchain.toml` files, so how does this work?

This is a combination of two things and makes multiple assumptions:

1) using a regex matcher (as proposed in the PR), we're able to select the appropriate version we need
2) to update, we have two possibilities. In theory, it'd be great if we could send an HTTP request, but that is sadly not possible
	1) [repology](https://repology.org/)
	2) GitHub releases

### repology

To implement this, repology was my first choice, as it is a giant aggregate of package vendors. None of the [repositories](https://repology.org/project/rust/versions) has an up-to-date version. FreeBSD has a manually updated version that isn't updated that often (https://www.freshports.org/lang/rust-nightly). 

So repology isn't a viable candidate.

### GitHub releases

GitHub releases is a great candidate, except the `rust-lang/rust` repo does not have tags for individual day releases, which is a bummer. Still, another adjacent repo does.`rust-analyzer`! `rust-analyzer` is officially supported and releases a version every week. Coincidentally that version has the same format to rust nightly versions.

This has one downside: we cannot check if that release e.g., supports `miri` on `aarch64`. It is primarily a guess in the dark, but the PR would fail if `miri` (or other components) were unavailable.

### datasource

renovate currently does not support custom datasources, if we would instead self-host renovate we could do this better by creating our own datasource which takes into account the features required, versioning, etc. Still, for now, this isn't feasible.

## 🔍 What does this change?

Use renovate to update the rust toolchain automatically.

## 📷 Demo

image

dry-run via:
```
LOG_LEVEL=debug RENOVATE_BASE_BRANCHES="bm/meta/renovate" renovate --token <TOKEN> hashintel/hash --dry-run --use-base-branch-config 'merge'
```

> Another small note: this is my 50th PR 🎉 thank you all so much for the continued support and for giving me the ability to contribute/create stuff alongside you all ❤️


TimDiekmann (commented):
This looks very promising, thank you very much @indietyp! Also. for being a regular contributor to the repository!

The changes look good, I like the "hack" using rust-analyzer 🙃 
I'd like to have another look from @kachkaev, otherwise happy to approve!

kachkaev (approved):
Interesting – I did not know about the [regex](https://docs.renovatebot.com/modules/manager/regex/) manager. Let’s see how it works, thanks for the PR!

nonparibus:
@indietyp **Congratulations** on PR number 50! 🪄✨ Collaborating with you is a delight.  🎉

PR #1862, opened 2023-01-17
Ignore `dist`, `runner_venv`, `target` and `venv` when checking license files

kachkaev:

This PR reduces spam in `yarn lint` output:

image

## 🔗 Related links

- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1673963900236779) _(internal)_


libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
```diff
@@ -36,7 +36,7 @@ const script = async () => {
     absolute: true,
     caseSensitiveMatch: false,
     cwd: monorepoRootDirPath,
-    ignore: ["**/node_modules/**"],
+    ignore: ["**/node_modules/**", "**/venv/**"],
```
Alfred-Mountfield, line 39: That ignore seems fine considering we don’t really use Python much, but the actual folder name is not prescribed, you could call your venv folder anything you want so if there was a more intelligent way of excluding it’d be good

Alfred-Mountfield (approved):

libs/@local/repo-chores/scripts/check-license-in-workspaces.ts
kachkaev: Which command can return `venv`?

Alfred-Mountfield: Actually I am realising that we also have a `runner_venv` folder.

_Generally_ the command is `python -m venv <FOLDER_NAME>` but that depends on how you are creating virtual environments, your version of Python, etc.

Alfred-Mountfield: apps/engine/lib/execution/src/runner/python/setup.sh:23

kachkaev: If you want, we can merge this as a `yarn lint` hotfix and then think again.

Alfred-Mountfield: > Is there any way we can limit the variety of those folders to be able to exclude them statically?

Not practically. We could establish guidelines but again a question of capacity and investment.

Let's add `runner_venv` to the ignore list and a comment above it explaining this isn't a fully-fleshed out solution, then can merge as a hotfix, yes.

kachkaev: > Let's add `runner_venv`
https://github.com/hashintel/hash/pull/1862/commits/e84de85d0c994d9b0fc1fdc461339277bb8dd675

Alfred-Mountfield (approved):

PR #1864, opened 2023-01-17
Re-enable backend build/push workflow

thehabbos007:

This PR re-enables the continuous building of the HASH backend docker images and pushes them to our private ECR. The workflow is triggered on merge to main and can be manually triggered through workflow dispatch

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203410952443001/f) _(internal)_

## 🔍 What does this change?

- Fixes workflow for building/pushing images
- Adds arg to composite action to allow build-time args
- Anticipates needing to migrate/redeploy backend

## 🐾 Next steps

- Auto-deploying the backend [asana task](https://app.asana.com/0/1203363157432094/1203410952443002/f) (_internal_)

## ❓ How to test this?

A little complicated, but manually re-enabling the workflow and then triggering the workflow dispatch in this ui: https://github.com/hashintel/hash/actions/workflows/hash-backend-cd.yml

## 📹 Demo

image

https://github.com/hashintel/hash/actions/runs/3951355941


.github/workflows/hash-backend-cd.yml
```diff
@@ -13,85 +14,105 @@ env:
...
 
   HASH_ECS_CLUSTER_NAME: ${{ secrets.HASH_ECS_CLUSTER_NAME }}
+  HASH_APP_SERVICE_NAME: ${{ secrets.HASH_APP_SERVICE_NAME }}
+
+  # To be uncommented once we migrate in the job below
+  # HASH_GRAPH_PG_USER: ${{ secrets.HASH_GRAPH_PG_USER }}
+  # HASH_GRAPH_PG_PASSWORD: ${{ secrets.HASH_GRAPH_PG_PASSWORD }}
+  # HASH_GRAPH_PG_HOST: ${{ secrets.HASH_GRAPH_PG_HOST }}
+  # HASH_GRAPH_PG_PORT: ${{ secrets.HASH_GRAPH_PG_PORT }}
+  # HASH_GRAPH_PG_DATABASE: ${{ secrets.HASH_GRAPH_PG_DATABASE }}
```
kachkaev, line 31: Reeeeally minor: it’s helpful to distinguish comments from commented-out lines of code by using `##` and `#`. This makes it possible to select a bunch of lines in an idea, use ‘toggle comment‘ shortcut and end up with a valid shell / yaml / etc. Not suggesting to make a change now, just sharing a side thought.

```suggestion
  ## To be uncommented once we migrate in the job below
  # HASH_GRAPH_PG_USER: ${{ secrets.HASH_GRAPH_PG_USER }}
  # HASH_GRAPH_PG_PASSWORD: ${{ secrets.HASH_GRAPH_PG_PASSWORD }}
  # HASH_GRAPH_PG_HOST: ${{ secrets.HASH_GRAPH_PG_HOST }}
  # HASH_GRAPH_PG_PORT: ${{ secrets.HASH_GRAPH_PG_PORT }}
  # HASH_GRAPH_PG_DATABASE: ${{ secrets.HASH_GRAPH_PG_DATABASE }}
```

kachkaev (approved):
Approving based on diff which makes sense. If there are any types, we can fix them in a follow-up.

.github/workflows/hash-backend-cd.yml
thehabbos007: I definitely think this makes sense, but given I'll be removing that line in #1877, I think it's fine to skip for now :)

PR #1873, opened 2023-01-18
Add prettier plugin for `*.sql` files

TimDiekmann:

The new migration scripts require, that the migration files did not change since the previous migration. Without consistent formatting, this will be very hard and, honestly, the generated format from Postgres is ugly.

## 🔍 What does this change?

- Adds the `prettier-plugin-sql` 
- Specifies `postgresql` as the language

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -18,6 +18,7 @@
 !*.ts
 !*.tsx
 !*.yml
+!*.sql
 
 # Keep license files as is
 LICENSE.md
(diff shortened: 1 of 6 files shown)


.prettierignore
```diff
@@ -18,6 +18,7 @@
 !*.ts
 !*.tsx
 !*.yml
+!*.sql
```
kachkaev, line 21: It’s best to keep these extensions alphabetically sorted, but I can do this in https://github.com/hashintel/hash/pull/1874. Not worth re-triggering CI in this PR.

kachkaev (dismissed):

thehabbos007 (approved):
Looks good to me :+1:

kachkaev (approved):

PR #1882, opened 2023-01-19
Move `OwnedById` and `UpdatedById` to `type_ids` table

TimDiekmann:

For future work and to avoid duplication we can move these types to the shared `type_ids` table

## 🔗 Related links

- This is required for [this Asana task](https://app.asana.com/0/0/1203774687353268/f) _(internal)_

## 🔍 What does this change?

- Change the database layout
- Adjust inserting logic
- Adjust querying logic
- Drive-by: Query base URI and the version from the `type_ids` table as well

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
@@ -229,16 +229,24 @@ where
         &self,
         uri: &VersionedUri,
         version_id: VersionId,
+        owned_by_id: OwnedById,
+        updated_by_id: UpdatedById,
     ) -> Result<(), InsertionError> {
         let version = i64::from(uri.version());
         self.as_client()
             .query_one(
                 r#"
-                    INSERT INTO type_ids (base_uri, version, version_id)
-                    VALUES ($1, $2, $3)
+                    INSERT INTO type_ids (base_uri, version, version_id, owned_by_id, updated_by_id)
+                    VALUES ($1, $2, $3, $4, $5)
                     RETURNING version_id;
                 "#,
-                &[&uri.base_uri().as_str(), &version, &version_id],
+                &[
+                    &uri.base_uri().as_str(),
+                    &version,
+                    &version_id,
+                    &owned_by_id,
+                    &updated_by_id,
+                ],
             )
             .await
             .into_report()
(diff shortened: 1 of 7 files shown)


packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/read.rs
```diff
@@ -56,23 +63,26 @@ where
...
                     .into_report()
                     .change_context(QueryError)?;
-                let owned_by_id = OwnedById::new(row.get(owned_by_id_index));
-                let updated_by_id = UpdatedById::new(row.get(updated_by_id_path_index));
 
                 Ok(T::new(
                     record,
                     OntologyElementMetadata::new(
-                        OntologyTypeEditionId::from(&versioned_uri),
+                        OntologyTypeEditionId::new(base_uri, version),
```
thehabbos007, line 85: The change made here to change the queried`versioned_uri` seems unrelated to the rest of the changes. Was this done to prevent looking into `schema->>$id` to get the versioned URI?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -345,11 +353,11 @@ where
 
         let version_id = VersionId::new(Uuid::new_v4());
         self.insert_version_id(version_id).await?;
-        self.insert_uri(&uri, version_id).await?;
-
-        self.insert_with_id(version_id, database_type, owned_by_id, updated_by_id)
+        self.insert_uri(&uri, version_id, owned_by_id, updated_by_id)
             .await?;
 
+        self.insert_with_id(version_id, database_type).await?;
```
thehabbos007, line 359: I suppose the SQL `function`s are yet to be made, are they going to be in a follow up PR?

thehabbos007 (commented):
Looks good to me! I have a couple questions, but other than that I'm happy to merge

packages/graph/hash_graph/lib/graph/src/store/postgres/ontology/read.rs
TimDiekmann: Yes, I wanted to avoid the schema lookup here. We have to look in the `type_ids` table anyway, so there is no reason to not directly look them up there.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: Yes, I'm currently cleaning up my branch and will open a PR with the functions in a couple minutes 🚀

thehabbos007 (approved):

PR #1884, opened 2023-01-19
Automatically update rust version (4th attempt)

indietyp:

I am sorry that this took four tries. The logs from `renovate` haven't been helpful at all. They are useless (attached screenshot), which required a change in tactics. I found that creating a fork and using `--include-forks` in the CLI works to test the integration properly.

* error-stack shields now use the query parameter version
	* using the path (previous version) led to a significant problem -> replacement, even with our template, would fail. I wonder if this is intentional. Replacing `-` with `--` did not work. We do not need to do that if we use the query parameter version of shields.
* regex those shields instead and use pretty much the same regex we use in the `rust-toolchain.toml`

image

Preview: https://github.com/indietyp/hash-renovate/issues/1
Preview: https://github.com/indietyp/hash-renovate/pull/3

> Disclaimer: The CLI returns `211:220: syntax error: A identifier can’t go after this “"”. (-2740)`, but everything still works, and there are no apparent error messages in the giant log. I also couldn't find anything related to that on GitHub. This could be just a CLI-related error and has nothing to do with renovate or the configuration. My suspicion is enforced by looking at the [renovate code](https://github.com/search?q=repo%3Arenovatebot%2Frenovate+syntax+error&type=code), there's no mention of a similar error. 🤔 

> <sup> This has been a lot more painful than I anticipated^^ </sup>


TimDiekmann (dismissed):
Mainly reviewed offline and in the fork, the results are looking perfect, thank you very much! 🎉

kachkaev (approved):

PR #1886, opened 2023-01-19
Add transaction time to ontology types

TimDiekmann:

Storing the transaction time for ontology types gives us the ability to use it at some later point. This also enables validation, if we are actually updating to the version after the latest version.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203596982776833/f) _(internal)_

## 🔍 What does this change?

- Add `transaction_time` to `type_ids` table and adjust functions
- raise an exception, if the version does not match

### DIFF
--- a/packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
+++ b/packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
@@ -26,13 +26,15 @@ BEGIN
     "version",
     "version_id",
     "owned_by_id",
-    "updated_by_id"
+    "updated_by_id",
+    "transaction_time"
   ) VALUES (
     create_ontology_id.base_uri,
     create_ontology_id.version,
     _version_id,
     create_ontology_id.owned_by_id,
-    create_ontology_id.updated_by_id
+    create_ontology_id.updated_by_id,
+    tstzrange(now(), NULL, '[)')
   ) RETURNING type_ids.version_id;
 END $create_ontology_id$ VOLATILE LANGUAGE plpgsql;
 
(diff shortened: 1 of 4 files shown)


packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
```diff
@@ -53,36 +55,46 @@ BEGIN
...
   SET CONSTRAINTS type_id_unique IMMEDIATE;
+  SET CONSTRAINTS type_ids_overlapping IMMEDIATE;
 END $update_ontology_id$ VOLATILE LANGUAGE plpgsql;
 
 CREATE
 OR REPLACE FUNCTION "update_type_ids_trigger" () RETURNS TRIGGER AS $pga$
     BEGIN
+      IF (OLD.version != NEW.version - 1) THEN
+        RAISE EXCEPTION 'Not updating the latest id: % -> %', OLD.version, NEW.version
+        USING ERRCODE = 'invalid_parameter_value';
```
thehabbos007, line 81: Oh cool, does the SQL type crate provide translations for these built-in enums? (or at least I think they are enums)

This is incredibly useful for the setup we have with the SQL functions!

thehabbos007 (approved):
Very sweet that the diff for this was so small. Great stuff! Thank you :)

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
TimDiekmann: Yes, all error codes are available via `SqlState` from `tokio_postgres`. Really handy!

PR #1893, opened 2023-01-20
Simplify ontology table definitions

TimDiekmann:

With #1886, it's possible to greatly simplify the definition of the ontology tables. This is possible as inserting a new ontology type with a base URI will fail as there will be overlapping transaction times.

## 🔗 Related links

- #1886

## 🚫 Blocked by

- #1886
## 🔍 What does this change?

- Make `version_id` the primary key of `type_ids` (we still require `(base_uri, version)` to be unique)
- Remove `base_uris` table
- Remove `version_ids` table
- Change the insertion order on the update trigger to avoid unneeded deferring of constraints
- Adjust the Graph to also catch the overlapping-constraint on creating

### DIFF
--- a/packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
+++ b/packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
@@ -190,7 +190,7 @@ where
             .into_report()
             .map(|row| row.get(0))
             .map_err(|report| match report.current_context().code() {
-                Some(&SqlState::UNIQUE_VIOLATION) => report
+                Some(&SqlState::EXCLUSION_VIOLATION | &SqlState::UNIQUE_VIOLATION) => report
                     .change_context(BaseUriAlreadyExists)
                     .attach_printable(uri.base_uri().clone())
                     .change_context(InsertionError),
--- a/packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
+++ b/packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
@@ -5,33 +5,19 @@ OR REPLACE FUNCTION create_ontology_id (
   "owned_by_id" UUID,
   "updated_by_id" UUID
 ) RETURNS TABLE (version_id UUID) AS $create_ontology_id$
-DECLARE
-  _version_id UUID;
 BEGIN
-  INSERT INTO "base_uris" (
-    base_uri
-  ) VALUES (
-    create_ontology_id.base_uri
-  );
-
-  INSERT INTO version_ids (
-    version_id
-  ) VALUES (
-    gen_random_uuid()
-  ) RETURNING version_ids.version_id INTO _version_id;
-
   RETURN QUERY
   INSERT INTO type_ids (
+    "version_id",
     "base_uri",
     "version",
-    "version_id",
     "owned_by_id",
     "updated_by_id",
     "transaction_time"
   ) VALUES (
+    gen_random_uuid(),
     create_ontology_id.base_uri,
     create_ontology_id.version,
-    _version_id,
     create_ontology_id.owned_by_id,
     create_ontology_id.updated_by_id,
     tstzrange(now(), NULL, '[)')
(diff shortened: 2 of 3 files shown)


packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
```diff
@@ -45,60 +31,54 @@ OR REPLACE FUNCTION update_ontology_id (
...
-        OLD.version_id,
-        OLD.owned_by_id,
-        OLD.updated_by_id,
-        tstzrange(lower(OLD.transaction_time), lower(NEW.transaction_time), '[)')
-      );
+  OLD.transaction_time = tstzrange(lower(OLD.transaction_time), lower(NEW.transaction_time), '[)');
 
-      RETURN NEW;
-    END$pga$ VOLATILE LANGUAGE plpgsql;
+  RETURN OLD;
```
thehabbos007, line 80: What is the implication of returning `OLD` here? I see that `update_entity_version_trigger` for knowledge returns `NEW`. Do we need it somewhere in code?

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
```diff
@@ -45,60 +31,54 @@ OR REPLACE FUNCTION update_ontology_id (
...
 CREATE
-OR REPLACE FUNCTION "update_type_ids_trigger" () RETURNS TRIGGER AS $pga$
-    BEGIN
-      IF (OLD.version != NEW.version - 1) THEN
-        RAISE EXCEPTION 'Not updating the latest id: % -> %', OLD.version, NEW.version
-        USING ERRCODE = 'invalid_parameter_value';
-      END IF;
+OR REPLACE FUNCTION "update_type_ids_trigger" () RETURNS TRIGGER AS $update_type_ids_trigger$
+BEGIN
+  IF (OLD.version != NEW.version - 1) THEN
```
thehabbos007, line 57: (just to preface, I haven't though through this too much)
Would we ever be in a situation where let's say:
```
OLD = 2
NEW = 3

LATEST = 4
```
would get to this point and insert a new row thinking it's a valid insertion? 

I suppose we would run into the unique check on `UNIQUE ("base_uri", "version")`, but was that the intention of the unique constraint to do specifically? I ask because I can imagine us slightly changing things leading to the above situation passing the insertion checks

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
```diff
@@ -190,7 +190,7 @@ where
             .into_report()
             .map(|row| row.get(0))
             .map_err(|report| match report.current_context().code() {
-                Some(&SqlState::UNIQUE_VIOLATION) => report
+                Some(&SqlState::EXCLUSION_VIOLATION | &SqlState::UNIQUE_VIOLATION) => report
```
thehabbos007, line 193: When does `EXCLUSION_VIOLATION` happen for us? I see in the docs it can happen due to serialization errors, so is it to do with the function call handing off to the trigger?

thehabbos007 (commented):
Thanks, Tim! I have some comments :)

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
TimDiekmann: The returned record of a `BEFORE UPDATE` trigger is the value the row will be updated to. I reverted the logic here to avoid the usage of `DEFERRED` constraints.

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
TimDiekmann: For your example this is true, this would violate the `UNIQUE` constraint. However, if `LATEST` is `4` and you try to update from `5 -> 6`, it wouldn't fail.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: If you have a type with `id=1` with the time of `[10, ∞)` and then try to insert a new type with `id=1` the time intervals will overlap, which is not allowed. So this check ensures, that the type does not exist already in the database.

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
thehabbos007: Do we see this as a potential problem as we'd be skipping 5?

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
thehabbos007: So it happens when the GIST index is violated?

packages/graph/hash_graph/postgres_migrations/V4__ontology_functions.sql
TimDiekmann: Currently, we enforce that behavior in the Rest API (by passing `typeToUpdate` and adding `+1` to the version). we can remove that check here though.

As a side note: This PR only removed the leading whitespace at this location.

packages/graph/hash_graph/lib/graph/src/store/postgres/mod.rs
TimDiekmann: Yup 🙂

thehabbos007 (approved):
Looks good to me!

PR #1897, opened 2023-01-20
Set integration tests step count to `10`

TimDiekmann:
## 🔗 Related links

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Try X
1.  Confirm that Y

### DIFF
--- a/apps/engine/tests/examples/air_defense_system/experiments.json
+++ b/apps/engine/tests/examples/air_defense_system/experiments.json
@@ -1,6 +1,6 @@
 {
   "radar_zone_arange": {
-    "steps": 100,
+    "steps": 10,
     "type": "arange",
     "field": "radarZone",
     "start": 5,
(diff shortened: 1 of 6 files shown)


nonparibus:
We'll probably want to revert these to more impactful step counts when we find ourselves focusing on simulation infra again, but while the primary Rust work takes place on the graph this change makes sense to me, speeding up CI in those instances where repo-wide Rust changes have consequence for hEngine and we can't just skip these tests entirely (as we already do).

Alfred-Mountfield (approved):
Agreed with David's comment when we go back to changing implementation logic within the engine (as opposed to global repo stuff like updating Rust version, etc.

Alfred-Mountfield (approved):

PR #1912, opened 2023-01-23
Disable Apollo cache on SSR requests

thehabbos007:

This PR disables apollo cache for any request by default on the NextJS server-side.
This reason why we do this is to prevent requests that take no params to be shared responses to users (i.e. the `me` query)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203788788550913/f) _(internal)_

## 🔍 What does this change?

- The Apollo client shared by client/server now conditionally disable caching

nope

## ❓ How to test this?

1.  Start local development setup as usual
1. Follow steps by @yusufkinatas to trigger response sharing:
```
1. Open two different browsers on incognito mode, side by side
2. Register `charlie`  on left
3. Register `dean`  on right
4. Refresh `charlie`’s page
    a. `charlie`  becomes `dean`
5. After step 4 , if we log out from everything, then try logging in to another account, that one gets replaced on refresh as well 
```
1. Observe if step `4.a.` does not happen anymore

### DIFF
--- a/apps/hash-frontend/src/lib/apollo-client.ts
+++ b/apps/hash-frontend/src/lib/apollo-client.ts
@@ -1,3 +1,5 @@
 import { createApolloClient } from "@local/hash-isomorphic-utils/graphql/create-apollo-client";
 
-export const apolloClient = createApolloClient();
+import { isBrowser } from "./config";
+
+export const apolloClient = createApolloClient({ isBrowser });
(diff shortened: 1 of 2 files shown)


yusufkinatas (approved):
Great work! 💪  Tested, works as expected on my local.

PR #1915, opened 2023-01-23
Support any JSON values as parameter to structural queries

TimDiekmann:

Currently, we only support `string | number | boolean` as the parameter values. With this Pull Request we also enable `object`, `Array`, and `null` as parameters. This implies, that we can compare a full `properties` object or arrays. Also, we can check, if a value is `null`.

## 🚫 Blocked by

- #1914

## 🔍 What does this change?

- Allow `properties` as terminating token in `path`
- Allow deserializing `Any` values
- Update OpenAPI specs


apps/hash-graph/lib/graph/src/store/query/filter.rs
```diff
@@ -275,7 +275,6 @@ pub enum Parameter<'p> {
     Boolean(bool),
     Number(f64),
     Text(Cow<'p, str>),
-    #[serde(skip)]
     Any(Value),
```
thehabbos007, line 278: Just a quick question about the ordering when deserializing. Would serde know to not make values of the `Any` variant when it acts like a match-all? Or does it try from top down? The order here makes sense in relation to top-down, I'm just curious :)

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
```diff
@@ -355,12 +355,7 @@ impl<'de> Visitor<'de> for EntityQueryPathVisitor {
             EntityQueryToken::Type => EntityQueryPath::Type(
                 EntityTypeQueryPathVisitor::new(self.position).visit_seq(seq)?,
             ),
-            EntityQueryToken::Properties => {
-                let property = seq
-                    .next_element()?
-                    .ok_or_else(|| de::Error::invalid_length(self.position, &self))?;
```
thehabbos007, line 361: Do we not want to throw this error or is it just not needed anymore?

thehabbos007 (approved):
Looks good to me, Looked at the consuming code locally and it seemed good with `any` regardless of it being very permissive. I have a couple questions, but nothing blocking merge :)

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
TimDiekmann: We don't want to as we can compare the whole JSON object

apps/hash-graph/lib/graph/src/store/query/filter.rs
TimDiekmann: It’s top-down. For the deserialization we only need `Any` as the parameter will be converted anyway, but for predefined queries (defined on `Filter`) it`s easier to have all types available, so we can skip conversion.

thehabbos007: Oh interesting. Thanks for explaining

PR #1919, opened 2023-01-23
Sort properties and links alphabetically in entity type editor tables

nathggns:

Sorts the properties and links tables in the entity editor alphabetically. Compensates for new rows no longer being inserted at the end by 'flashing' then in blue and, if necessary, scrolling them into view. 

**There may still be an issue with not scrolling into view when the edit bar is visible and covering the new role. I'll investigate, but will likely need to fix in a follow up**

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203661700052726/f) _(internal)_

## 🔍 What does this change?

- Moved EditBar as needed to make use of `EDIT_BAR_HEIGHT` in the scroll logic for flashing table rows
- Moved the logic for resolving `link` and `property` up from the table rows to the tables themselves, as its necessary to have access to the titles of each in order to sort the rows.
	- New `sortRows` function to help sort the rows of a react hook form field array
- New `useFlashRows` hook to manage the timing of flashing and unflashing new rows in property/links tables
-  New `flash` prop on `EntityTypeTableRow`, which will flash the row and scroll it into view (if its out of view), ensuring it's not under the edit bar.

## ⚠️ Known issues

N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Add a new property or link
2. See its in the right place and flashes as expected

## 📹 Demo

(Recorded before merging with main, so the type selector went off screen when it shouldn't)

https://user-images.githubusercontent.com/719814/214285367-cef6c62a-3afc-4763-9279-09afb2f89d20.mov


nathggns:
@nonparibus Requested your review for visual approval. Video should be enough to review.

nonparibus:
> @nonparibus Requested your review for visual approval. Video should be enough to review.

I'm happy with this UX, thanks.

I wonder if we should have some "jump back down to add another" type function when users are whizzed up to the property's location in the table, if the "Add new" button is now off-screen/no longer visible. It's NBD in short lists like the one shown in the demo, but in long/expanded ones it will be a bit meh.

I think this implementation is fine for now, though, and we could always add this later. Let's definitely not block this PR on it.

nonparibus (approved):
Functional approval ✅

**Still requires code review.**

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx
```diff
@@ -386,21 +391,25 @@ const PropertyRow = ({
...
+  property: PropertyType | undefined;
+  propertyId: VersionedUri;
   onRemove: () => void;
   onUpdateVersion: (nextId: VersionedUri) => void;
+  flash: boolean;
 }) => {
   const { control } = useFormContext<EntityTypeEditorForm>();
 
-  const [$id, array] = useWatch({
+  // @todo move down
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:409]: What does this `@todo` mean? Can it happen now, or be better described for later?

Comment (nathggns) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:409]: Apologies, I thought I'd updated all the comments. This variable is only used in a lower component, and subscribing to it at the lowest possible level reduces the size of the re-render tree when its changed. It's a principle of react-hook-form that you should isolate your re-renders. I'll fix it.

Comment (nathggns) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card.tsx:409]: Sorry, I just remembered why I didn't make this change. Because of the property object recursiveness, the logic for working out if is array is not the same for every instance of `PropertyRow`, so we can't move this down. I meant to remove this comment. Doing that now.

CiaranMn (approved):

PR #1920, opened 2023-01-23
Use 0.3 staging block hub

CiaranMn:

HASH is updated to load Block Protocol v0.3 blocks as of #1823, but the production [**Þ _Hub_**](https://blockprotocol.org/hub) will be serving 0.2 blocks until we've got everything ready for 0.3 to release.

In the meantime, to be able to test out 0.3 blocks in HASH, we can use a version of blockprotocol.org built off the `0.3` branch in `blockprotocol`.

This PR does that.

## 🔗 Related links

- [Asana task for restoring the production hub](https://app.asana.com/0/1203358502199087/1203801045329699/f) _(internal)_

## 🔍 What does this change?

- Centralise all the references to the BP URL for loading blocks to a shared `blockProtocolHubOrigin` variable
- Set the value of that variable to `https://blockprotocol-git-03.stage.hash.ai`, which is the Vercel alias for 'deployment built from the tip of `0.3`'
- Use `paragraphBlockComponentId` variable in a couple of places where we were previously repeating the literal string for the paragraph's URL in blockprotocol.org

## 🐾 Next steps

- Revert this when the `0.3` branch of `blockprotocol` is merged into `main`

## ❓ How to test this?
**You may need to clear your `localStorage` value for "user workspace blocks" if you have any issue with the below**

1.  Load a page
1.  Load the code block, which is 0.3-compliant, edit the content, refresh to check that the update call succeeded
1.  To confirm that the BP API works, set `BLOCK_PROTOCOL_API_KEY` to a staging API key (generate one in the linked deployment)


libs/@local/hash-isomorphic-utils/src/blocks.ts
```diff
@@ -203,13 +203,19 @@ export const fetchBlock = async (
...
 };
 
-export const paragraphBlockComponentId =
-  "https://blockprotocol.org/blocks/@hash/paragraph";
+/**
+ * @todo-0.3 replace this temporary domain with blockprotocol.org
+ * https://app.asana.com/0/1203358502199087/1203788113163116/f
+ */
+export const blockProtocolHubOrigin =
+  "https://blockprotocol-git-03.stage.hash.ai";
```
kachkaev, line 211: Can this be made an env var with a default value? It might make it easier to test things in the future

kachkaev (approved):

libs/@local/hash-isomorphic-utils/src/blocks.ts
CiaranMn: Yes, good idea – I've done this in https://github.com/hashintel/hash/pull/1920/commits/a6b33347e0400a5aa6bb2ca2db94466b16c20a3f

I've kept the default as the temporary domain, as I think it will be less work for me to come back and replace this than to ensure that everyone has the temporary domain in an environment variable on their machine (with those who miss the prompt encountering bugs / starting Slack threads)

kachkaev (approved):

kachkaev (dismissed):

kachkaev (approved):

PR #1340, opened 2022-11-03
Allow logged-out users to view pages

CiaranMn:
## 🌟 What is the purpose of this PR?

This PR makes it so that if you visit a page's URL while you are logged out, you will see a readonly version of the page, stripped of any UI.

This involves removing the 'is logged in' middleware on the API `page` query, but as per the README the app is not yet secure and this middleware was only ever superficial (it is possible to retrieve entities you do not own via other routes, and there is self sign-up, making the app as a whole insecure). A proper 'view page while logged out' solution would involve users either (a) choosing to make particular pages public or private, or (b) generating 'anyone with the link can view' links which encode a token of some kind.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1200211978612931/1203216920361670/f) _(internal)_

## 🔍 What does this change?

- Allow anonymous retrieval of pages
- Update `useReadonlyMode` to also put the app into readonly if there is no authenticated user
- Also hide the navbar and comment UI if in readonly mode
- don't load ProseMirror plugins if in readonly (they're all related to editing)

## ⚠️ Known issues

- This is a temporary solution that makes it even easier for anyone to see any page.
- The page content doesn't seem quite centered to me, but this also happens in editing view. We should fix that at some point

## 🐾 Next steps

- A better solution when we introduce proper auth

## ❓ How to test this?

1.  Get a page URL by logging in and visiting a page
2. Open the page URL in incognito
3. See the page in readonly mode

## 📹 Demo

Screenshot 2022-11-03 at 17 56 45

PR #1921, opened 2023-01-23
Remove get entity logged in gate

CiaranMn:

#1340 introduced a temporary way of having 'anonymous readonly' pages which allowed any anonymous user to see any page.

#1823 broke this by introducing a new dependency on `getEntity` for resolving a page's data (in order to fetch the subgraph rooted at a block).

This PR fixes that by removing the logged in gate on `getEntity`. As mentioned in #1340 and elsewhere, these gates are superficial and do not provide proper security for any entity – which will be solved by introducing proper authorization – and this therefore does not represent removal of an effective security control.

## ❓ How to test this?

1.  Log in, find a page URL
2. View the page in incognito, check it loads without crashing

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/index.ts
+++ b/apps/hash-api/src/graphql/resolvers/index.ts
@@ -97,7 +97,7 @@ export const resolvers = {
     pages: loggedInAndSignedUpMiddleware(pagesResolver),
     pageComments: loggedInAndSignedUpMiddleware(pageCommentsResolver),
     blocks: loggedInAndSignedUpMiddleware(blocksResolver),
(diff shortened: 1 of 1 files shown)


kachkaev (commented):
Playwright tests [failed](https://github.com/hashintel/hash/actions/runs/3989921938/jobs/6842931093#step:9:52) and I restarted them.

The change is somewhat related to security so I’ll leave the approval to someone else just in case.

kachkaev (approved):

thehabbos007:
Oh I didn't see this was a 1 line change, I thought it was a bigger PR :smile: 
LGTM too though

PR #1927, opened 2023-01-24
Support `null`, `emptyList` & `object` data type in entity editor

yusufkinatas:

This PR adds support for `null`, `emptyList` and `object` types in entity editor. So instead of seeing the fallback `TextEditor`, users will see proper UI for these three types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203715515231598/f) _(internal)_

## 🔍 What does this change?

- Add `prismjs` to highlight JSON in `json-editor`
- Create a const for glide-grid util classname `click-outside-ignore`
- Rename `entityFromDB` -> `entityFromDb`
- Update single-value and array editor logics to support new types
- Move `editorSpecs` to it's own file since it's getting bigger with more editors
- Refactor render code of `sortable-row`
- add `arrayValueToString` util function to render a preview of arrays on grid cells
- improve `guessEditorTypeFromExpectedType`, now it searches for the given type in `editorSpecs`

## ❓ How to test this?

1.  Create an entity type with `object` in it (if you want to test `emptyList` and `null` types, you need to use the API to create properties from those types manually via code.)
2. Confirm that new object editor & property table UI/UX works as expected

## 📹 Demo

https://user-images.githubusercontent.com/39553853/214262235-29e37f49-5e20-4f2b-b3f3-91e761726fbd.mov


apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/inputs/json-input/json-editor.tsx
```diff
@@ -0,0 +1,294 @@
...
+  useRef,
+} from "react";
+
+interface JsonEditorProps {
+  value: string;
+  onChange: (value: string) => void;
+  height: string | number;
+}
+
+export const JsonEditor = ({ onChange, value, height }: JsonEditorProps) => {
```
Comment (yusufkinatas) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/inputs/json-input/json-editor.tsx:22]: This is almost an exact copy of  the `json-editor` at block-protocol repository. https://github.com/blockprotocol/blockprotocol/blob/3e6e44a95c0d9b9bd9f0f8f5efc165e5fafb151e/apps/site/src/components/pages/hub/json-editor.tsx

CiaranMn (changes requested):
Thanks @yusufkinatas – I haven't looked at the code yet but I had a test and have a few comments.

I added the different data types to the type editor to make this easier to test. If you're happy to I think we might as well add these in here [[diff](https://github.com/hashintel/hash/compare/yk/support-more-types-on-properties-table...cm/support-extra-types-in-type-editor)]

Could you update the icons in the entity editor to use the icons I've used there? We might want to look at sharing the icons from `expectedValuesOptions` across both editors so they are consistent (could be follow up if this is a pain / blows up the diff)

I have some suggestions for UX improvements – please let me know if any of these are lots of work, because we don't need to over-invest in this.

**1. Selection of fixed-value types, e.g. Null and EmptyList**
On UX – I think we can improve how we set fixed-value types, i.e. `Null` and `EmptyList`. At the moment you have to click to select the type, _and then_ click the pill. Once I've selected the type, and I see a `Null` pill, I assume that it's already set and blur the input, but that doesn't set it.

I think the _ideal_ is that I choose the type and then the value is automatically set – no need to click the pill again, it's pointless.

A slight improvement would be that the click target for the pill is the entire width of the input, not having to hunt the pill out.

https://user-images.githubusercontent.com/37743469/214379562-64292939-1afc-4f8f-a5ad-15bffa589b6f.mp4

**2. Display of fixed-value types**

Ideally we would show these as pills rather than strings, so that they are not confused with the strings Null and EmptyList

**3. Clearing of fixed value**

It is not possible to remove e.g. `Null` once it is added – for non-required properties, we should have some way of clearing the value. 

**4. Resizing expected value column**

This has a resize handle but it does not resize.

**5. Validate JSON is an object**

We should check that the input starts with `{` because it is supposed to be a JSON object not any JSON (e.g. not a number or string, nor an array)

~**6. Sorting properties in the table**~

~The properties are sorted by value, not by property name, which means the rows jump about when the value changes. We should fix this.~

I didn't realise there were options for this and I had clicked the 'Value' header at some point which sorted it on value 😅 

~**7. Rename Object -> JSON Object**~

~Let's rename it so we can better distinguish it from the property object~ We decided to park this for now and see if anyone is confused in practice.

Thoughts welcome on the investment required for these – thanks!

yusufkinatas:
Hey Ciaran, thanks for the suggestions. Some of them are pretty small, some might be better in follow-up PRs. Going to update the PR tomorrow.

CiaranMn (approved):

CiaranMn (approved):

PR #1931, opened 2023-01-24
Remove docker-compose from `hash-graph`

TimDiekmann:
## 🔗 Related links

- [Parent task](https://app.asana.com/0/1203543026679373/1203767287750365/f) (_internal_)
- As a drive-by, this also solves [this task](https://app.asana.com/0/1201095311341924/1203071961523005/f) _(internal)_

## 🔍 What does this change?

- Remove the `docker-compose` file from the `hash-graph` and use external services instead
- Adjust tests and makefiles to take this into account

PR #1932, opened 2023-01-24
Flatten directory structure of `hash-graph/hash_graph`

TimDiekmann:

We previously had the Rust project in `packages/graph/hash_graph`, but as the `apps/hash-graph` directory is more cleaned up now and will be more cleaner in the future, the nesting is not needed (and we don't need to come up with a better name)

## 🚫 Blocked by

- #1931 

## 🔍 What does this change?

- Merge `Makefile.toml` files
- Use `cargo run -- migrate` instead of using the external service
- Move everything from `./hash_graph` to `.`.


apps/hash-graph/Makefile.toml
```diff
@@ -17,12 +31,16 @@ args = ["external-services", "up", "postgres", "--wait"]
...
 description = "Run DB migrations using the Graph binary"
-extend = "yarn"
-args = ["external-services", "up", "graph-migrate", "--build"]
-cwd = "${REPO_ROOT}"
+args = ["run", "--", "migrate"]
+
+[tasks.migrate-db.env]
+HASH_GRAPH_PG_USER = "${POSTGRES_USER}"
+HASH_GRAPH_PG_PASSWORD = "${POSTGRES_PASSWORD}"
+HASH_GRAPH_PG_DATABASE = "${HASH_GRAPH_PG_DEV_DATABASE}"
```
thehabbos007, line 43: This is cool!

apps/hash-graph/Makefile.toml
```diff
@@ -34,7 +52,7 @@ cwd = "${REPO_ROOT}"
```
thehabbos007, line 52: (for deployment-down task)
do we perhaps want to pass `"${@}"` at the end of the args here to be able to do something like `cargo make deployment-down -v` to wipe volumne? We could add it in `*-up` too.

We already do this in `tasks.build-docker`

apps/hash-graph/Makefile.toml
```diff
@@ -60,4 +78,74 @@ run_task = { name = ["yarn-install-openapi", "graph-up", "test-rest-api", "graph
...
 # This is a temporary solution until we have e2e tests in place
 extend = "yarn"
-args = ["httpyac", "send", "--all", "${REPO_ROOT}/apps/hash-graph/hash_graph/tests/rest-test.http", "${@}"]
+args = ["httpyac", "send", "--all", "${REPO_ROOT}/apps/hash-graph/tests/rest-test.http", "${@}"]
+
+
+[tasks.yarn-install-openapi]
+description = "Installs the required yarn dependencies for generating the API client"
+extend = "yarn"
+cwd = "${REPO_ROOT}/apps/hash-graph/clients"
```
thehabbos007, line 87: tiny nit that's probably not worth addressing:
Instead of the `cwd` path here we could pass the workspace name of graph client generator (`@hashintel/hash-graph-client-generator`). Would make it so we don't need to specify the path here and potentially get it wrong in 2 places. But since we're extending yarn I'm not sure what's best

apps/hash-graph/README.md
```diff
@@ -31,7 +31,7 @@ cargo run -- server
 ### Logging configuration
 
 Some of the libraries used are very talkative in `trace` logging configurations, especially `mio`, `hyper`, and `tokio_util`.
-If you're interested in just increasing the logs for the Graph, we recommend specifically targeting the crates with `RUST_LOG=graph=trace,hash_graph=trace`.
+If you're interested in just increasing the logs for the Graph, we recommend specifically targeting the crates with `RUST_LOG=graph=trace,hash-graph=trace`.
```
thehabbos007, line 34: We have `RUST_LOG` values using `hash_graph` in:
- [apps/hash-external-services/docker-compose.yml#L93](https://github.com/hashintel/hash/blob/d5b9071524c3094059d48466fd32a59716b08a0d/apps/hash-external-services/docker-compose.yml#L93) (and line 145)
- [infra/terraform/hash/prod-usea1.tfvars#L33](https://github.com/hashintel/hash/blob/d5b9071524c3094059d48466fd32a59716b08a0d/infra/terraform/hash/prod-usea1.tfvars#L33)

thehabbos007 (commented):
Thank you! Awesome follow-up to the previous move PR. I have some comments :)

apps/hash-graph/Makefile.toml
thehabbos007: Discussed offline, we tried this before and it conflicts with CI params

apps/hash-graph/Makefile.toml
thehabbos007: Discussed offline, this will be addressed in a follow-up moving the clients into `libs`

thehabbos007 (approved):
Ty!

PR #1935, opened 2023-01-24
Fix path to hash-graph `Dockerfile`

TimDiekmann:

As flagged in https://github.com/hashintel/hash/pull/1931#issuecomment-1402284288 I missed a few locations to update in https://github.com/hashintel/hash/pull/1931


kachkaev (approved):
👍

PR #1936, opened 2023-01-24
`graph` switch for timestamp from chrono to time

indietyp:

This switches from `chrono` and `RFC3339` parsing to `time` and `ISO8601`. This now also allows any timezone instead of just UTC.

While `RFC3339` and `ISO8601` are similar, they are not identical. Notably `ISO8601` is a lot more encompassing.

Regarding date-time, the following is the TL;DR difference:

* `RFC3339` only: `2023-01-24_19:19:28Z`
* `RFC3339` only: `2023-01-24t20:19:48+01:00`
* `RFC3339` only: `2023-01-24 20:19:48Z`
* `ISO8601` only: `2023` (**not** accepted by `time::OffsetDateTime`)
* `ISO8601` only: `T20:20:53` (**not** accepted by `time::OffsetDateTime`)

TL;DR: `ISO8601` requires `T` as a separator, and `RFC3339` permits any character. `ISO8601` separators are case-sensitive, and `T` and `Z` of `RFC3339` are case-insensitive. `ISO8601` requires `T` to be present if date and time are used.

> **Fun Fact:** I've actually read through the whole `ISO8601` specification

> **Fun Fact:** `ISO8601` also supports durations and one can specify duration with a start date (and repeats), e.g., if I want a duration of 3 months from 2022-12-12, then I'd write: `2022-12-12/P3M`, a range between two dates is described as `2022-12-12/2023-03-12`, and one can have a duration into the other direction `P3M/2023-03-12`. (`ISO8601-1:2019 5.5.3`)

## 🔗 Related links

- [`ISO8601` vs `RFC3339` differences](https://ijmacd.github.io/rfc3339-iso8601/)


indietyp:
I had lockfile merge collisions, so I executed `cargo generate-lockfile`. It seems like that `cargo generate-lockfile` acts like `cargo-update` :/

apps/hash-graph/Cargo.lock
```diff
@@ -2076,24 +2012,29 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.1.44"
+version = "0.3.17"
```
TimDiekmann, line 2015: This version bump makes me very happy!

apps/hash-graph/tests/integration/postgres/mod.rs
```diff
@@ -4,9 +4,8 @@ mod entity_type;
 mod links;
 mod property_type;
 
-use std::{borrow::Cow, ops::Bound, str::FromStr, time::SystemTime};
+use std::{borrow::Cow, str::FromStr};
```
TimDiekmann, line 7: I think I removed `ops::Bound` several times now, but somehow it keeps appearing. Maye it's now gone forever 😆

TimDiekmann (approved):
LGTM, thanks @indietyp!

PR #1937, opened 2023-01-24
Move `hash-graph-client` into `@local/hash-graph-client`

TimDiekmann:

Currently, we have a nested workspace in `apps/hash-graph`. This PR cleans this up and moves the generation into `libs/@local`

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679373/1203767287750365/f) _(internal)_

## 🔍 What does this change?

- Moves `hash-graph-client` to `apps/@local/hash-graph-client`
- Moves `hash-graph-client-generator` to `apps/@local/hash-graph-client-generator
- Adjust and cleans up Makefile to generate the files


apps/hash-graph/Makefile.toml
```diff
@@ -123,29 +124,25 @@ args = ["llvm-cov", "--no-report", "nextest", "--workspace", "--all-features", "
 extend = "coverage-task"
 args = ["llvm-cov", "report", "${@}"]
 
+
+## OpenAPI client
+####################
+
 # If this step fails, the OpenAPI spec is invalid.
 # When running in CI, a non-empty git diff will fail the pipeline.
 [tasks.generate-openapi-client]
```
thehabbos007, line 133: Just for my understanding, is the intention that we run this task only when the Graph API is running? The script to start/stop the API is now gone.

I see that it's part of the `tasks.test-docker` step, so I suppose this is the intention

apps/hash-graph/README.md
```diff
@@ -99,7 +99,7 @@ cargo run -- migrate
 
 ## Generate OpenAPI client
 
-The HASH Graph produces an OpenAPI Spec while running, which can be used to generate the `@hashintel/hash-graph-client` typescript client. In the `/apps/hash-graph` directory run:
+The HASH Graph produces an OpenAPI Spec while running, which can be used to generate the `@local/hash-graph-client` typescript client. In the `/apps/hash-graph` directory run:
 
 ```shell
 cargo make generate-openapi-client
```
thehabbos007, line 105: Above comment is a leading question for this, we'd need to adjust docs to specify the app must be running before this command is useful.

libs/@local/hash-graph-client-generator/openapitools.json
```diff
@@ -7,11 +7,11 @@
...
         "generateAliasAsModel": true,
         "generatorName": "typescript-axios",
-        "output": "#{cwd}/typescript",
+        "output": "#{cwd}/../hash-graph-client",
         "inputSpec": "http://localhost:4000/api-doc/openapi.json",
         "additionalProperties": {
-          "npmName": "@hashintel/hash-graph-client",
-          "npmVersion": "0.0.0",
+          "npmName": "@local/hash-graph-client",
+          "npmVersion": "0.0.0-private",
```
thehabbos007, line 14: I think we `openapi-generator-ignore` the generated `package.json`, so the values here don't technically make a difference. Thanks for keeping them in-sync though!

libs/@local/hash-subgraph/jest.config.cjs
```diff
@@ -13,7 +13,7 @@ module.exports = {
   // The chosen default give a lot of room to the integration test.
   testTimeout: 60000,
   moduleNameMapper: {
-    "@hashintel/hash-graph-client": "<rootDir>/../../graph/clients/typescript",
+    "@local/hash-graph-client": "<rootDir>/../../graph/clients/typescript",
```
thehabbos007, line 16: I believe we should adjust this
```suggestion
    "@local/hash-graph-client": "<rootDir>/../hash-graph-client",
```
Although I don't think it affects anything as we're just importing types which are erased during the tests

thehabbos007 (commented):
App works locally and lints pass. I have some small observations

apps/hash-graph/Makefile.toml
TimDiekmann: Yes, this is the intention so we can easily create it either from a "native" graph or from the graph started through docker

apps/hash-graph/README.md
thehabbos007: (Discussed offline), it does say in the paragraph above that the `OpenAPI` spec is there when the graph is running, so my comment would only be helpful to provide copy-paste lines to run to quickly generate the spec

thehabbos007 (dismissed):

nonparibus (dismissed):

PR #1938, opened 2023-01-24
Rename `module/mod.rs` to `module.rs` in error-stack

TimDiekmann:

As we currently restructure our repository, it's a good opportunity to move over to the recommended module file names. This renames all `dirname/mod.rs` to `dirname.rs`.

## 🔗 Related links

- [Module Source Filenames (doc.rust-lang.org/)](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames)

## 🔍 What does this change?

```sh
find . -name 'mod.rs' -exec bash -c 'git mv {} $(dirname {}).rs' \; 
```


thehabbos007 (approved):
Looks good to me. No changes to actual code, just tests, and all tests run work locally. Thank you!

packages/libs/error-stack/tests/ui/macro_invalid_args.stderr
```diff
@@ -95,7 +95,7 @@ error[E0308]: mismatched types
    |             arguments to this function are incorrect
    |
 note: associated function defined here
-  --> $RUST/core/src/convert/mod.rs
```
thehabbos007, line 98: I did notice this when testing locally through search. What is this `convert` module? It isn't renamed in the PR which is why I didn't bring it up. Is this change valid?

thehabbos007 (approved):
Thanks, looking good again :+1:

PR #1939, opened 2023-01-24
Rename `module/mod.rs` to `module.rs` in deer

TimDiekmann:
## 🌟 What is the purpose of this PR?

As we currently restructure our repository, it's a good opportunity to move over to the recommended module file names. This renames all `dirname/mod.rs` to `dirname.rs`.

## 🔗 Related links

- [Module Source Filenames (doc.rust-lang.org/)](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames)

## 🔍 What does this change?

```sh
find . -name 'mod.rs' -exec bash -c 'git mv {} $(dirname {}).rs' \; 
```

PR #1941, opened 2023-01-24
Rename `module/mod.rs` to `module.rs` in hash-graph

TimDiekmann:

As we currently restructure our repository, it's a good opportunity to move over to the recommended module file names. This renames all `dirname/mod.rs` to `dirname.rs`.

## 🔗 Related links

- [Module Source Filenames (doc.rust-lang.org/)](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames)

## 🔍 What does this change?

```sh
find . -name 'mod.rs' -exec bash -c 'git mv {} $(dirname {}).rs' \; 
```


indietyp:
I find this change quite controversial. While encouraged the documentation does not note that this is recommended. I also find that this change has some significant drawbacks:

1. cognitive load: many tools like CLion or GitHub separate files and directories by default. Associating the `module.rs` with the `module` folder makes this a lot harder (everything is also contained in a single place, making it easier to work with imp)
2. precedence and ecosystem
	1. other languages: other languages, like python (`__init__.py`) and typescript (`index.ts`) use a similar approach to a `mod.rs` file
	2. the rust ecosystem at large currently uses `mod.rs` files. There are very few projects (only `rust-analyzer` comes to mind) that use the outlined approach
3. https://github.com/rust-lang/reference/issues/1215, as outlined in this issue, this note might even be removed.
4. I won't deny this, as this, of course, plays into my opinion: personal taste, but I still think that the points mentioned above are valid

TimDiekmann:
The decision was made based on two factors:
- A while ago I noticed that advice in the linked documentation. I also have seen the issue you linked to; even if this decision is controversial, there has not been any recent action to change this. Initially, this advice was even [stronger phrased](https://github.com/rust-lang/book/commit/597d00a91b7f4302a168a5cede379421f7fa8d5f). I'm not sure if I like this approach more than the other and I hear your concern on (1), however, personally, I encountered several issues when the editor was only showing `mod.rs` and I had to hover the file to get the actual location.
- With our current restructuring of the repository we set some principles to maintain consistency, not only inside of a library but across the whole repository. This makes it easier to develop in different areas of the repository. For JavaScript(-like) languages, we are in the process of removing the `index.js` files, which more-or-less corresponds to the `mod.rs` file: an entry point to the module.

indietyp:
Thank you very much for the insight and the reasoning as to why. I still don't think that this is the right move. <sub>(and I won't block on this, not that I have the power to do so^^, I just wanted to voice my opinion)</sub>.

> With our current restructuring of the repository we set some principles to maintain consistency, not only inside of a library but across the whole repository. 

But isn't this just making everything differently consistent? Currently, all rust libraries are consistent by using the `mod.rs` method and now move the `module.rs` through #1938, #1939, #1940, and #1941.

> encountered several issues when the editor was only showing mod.rs and I had to hover the file to get the actual location.

I, too, had this problem from time to time, so I get your point, but this never caused issues for me personally.

> For JavaScript(-like) languages, we are in the process of removing the index.js files, which more-or-less corresponds to the mod.rs file: an entry point to the module.

Yes, but doesn't this make the interop harder? The `index.[tj]s` file is optional and can either be there or not be there, while a `mod.rs` or `module.rs` file is required. At least for me, the switch from a self-contained module in javascript, typescript, or python to a `module + module.rs` could be confusing, as it is a paradigm shift and could lead a less experienced user to assume that, like in javascript, an index file is optional.

<sub>Looking at /r/rust it seems like we're not the only ones discussing this^^</sub>

indietyp:
Thank you for the response! I understand.

I just wanted to note something else too, the module.rs method has been available since edition 2018, so it seems a bit odd that even crates that are never than that (the top 100 is comprised of many different crates, only some older than 2018) still chose to use the mod.rs pattern.

thehabbos007 (commented):
Smallest nit: 
https://github.com/hashintel/hash/blob/d7afa8a7d19de0746ef822d449b2aa6ae49eb0c6/apps/hash-graph/bench/benches/util.rs#L18
refers to `mod.rs`

TimDiekmann:
Thanks @thehabbos007, I fixed it in da898a01df410268cc4c00c7620f213620a0e02d

thehabbos007 (approved):
Thanks!

PR #1943, opened 2023-01-25
Update refs to `/packages/hash` with `/apps/hash` 

nonparibus:

Updates `packages/hash` to `apps/hash` in various places throughout the repo.

### DIFF
--- a/.github/workflows/hash-backend-cd.yml
+++ b/.github/workflows/hash-backend-cd.yml
@@ -11,7 +11,7 @@ on:
       - "libs/@local/hash-isomorphic-utils/**"
       # TODO: remove/rewrite these as we remove the package directory.
       - "apps/hash-graph/**"
-      - "packages/hash/**"
+      - "apps/hash/**"
 
 env:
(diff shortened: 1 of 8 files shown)


TimDiekmann (dismissed):
LGTM, happy to reapprove when CI is green

CiaranMn (approved):

.github/workflows/hash-backend-cd.yml
```diff
@@ -11,7 +11,7 @@ on:
       - "libs/@local/hash-isomorphic-utils/**"
       # TODO: remove/rewrite these as we remove the package directory.
       - "apps/hash-graph/**"
-      - "packages/hash/**"
+      - "apps/hash/**"
```
kachkaev, line 14: Something is off here. I guess we mean 

```suggestion
      - "apps/hash-*/**"
```

because `apps/hash` only contains a single `README.md`. Overall, I wonder if we an place that file somewhere else, e.g. turning it into a section of `apps/README.md`. This way we won't have to have `apps/hash`, which is currently a folder amongst other Yarn workspaces that is not a workspace itself.

PR #1946, opened 2023-01-25
Add new `Table` block (MVP supporting local data only)

yusufkinatas:

This is the first PR for the new Table block. It will be follow up PRs implementing new features like query building, linking data, filtering, sorting etc. This PR just introduces the most basic functionality and the usage of `glide-data-grid`.

Here is the list of the supported functionalities for now

- local columns
- local rows
- adding colums / rows
- editing cells
- updating table title
- readonly mode

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1200211978612931/1203179653350866/f) _(internal)_

## 🐾 Next steps

Implement rest of the features for Table block

## ❓ How to test this?

1.  Run `table-v2` block on your local
2. Confirm that it works as expected

## 📹 Demo

⚠️ Showing copy-paste using keyboard shortcuts at the beginning of the demo

https://user-images.githubusercontent.com/39553853/223970783-8c182290-6d1a-4027-9e56-214b01e6bb13.mov

---
  - https://app.asana.com/0/0/1203922830586217


nonparibus:
@yusufkinatas A couple of observations:

1. we highlight the active column when you click in a cell, but not the active row. Could we do that, too?
1. can we have the 'New row' row flush with the row above it? It's a bit odd to me that there's a partial gap between.

yusufkinatas:
@nonparibus Addressed all of the comments!

nonparibus (dismissed):
Legal approval ✅

Code review pending

Functional review pending

nathggns:
@nonparibus @CiaranMn what's the plan for this one?

nonparibus:
@nathggns OOS for your updates. Some YX fixes required. We're going to revisit later in the week, or early the following.

blocks/table-v2/src/components/settings/settings.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+            <div className={styles.option}>
+              <input
+                type="checkbox"
+                id="isStriped"
+                checked={isStriped}
+                onChange={() => updateEntity({ [isStripedKey]: !isStriped })}
+              />
+              <label htmlFor="isStriped">
+                Enable stripes for alternating rows
```
nonparibus, line 66: We should expand the pop-up so this text doesn't bleed onto two lines.

blocks/table-v2/src/app.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  return (
+    <div className={styles.block} ref={blockRootRef}>
+      <div className={styles.titleWrapper}>
+        <TableTitle onChange={setTitle} title={title} readonly={readonly} />
+        {!readonly && (
+          <Settings blockEntity={blockEntity} updateEntity={updateEntity} />
+        )}
+      </div>
+      {!!selectedRowCount && !readonly && (
```
nonparibus, line 190: Once a single checkbox is selected, we should then display checkboxes next to all other rows (in unchecked state, unless/until flipped).

We also want to allow for multiple checkboxes to then be selected, without having to shift+click or cmd+click (if not already supported).

blocks/table-v2/src/components/grid/grid.tsx
```diff
@@ -0,0 +1,22 @@
+import { DataEditor, DataEditorProps } from "@glideapps/glide-data-grid";
+import "@glideapps/glide-data-grid/dist/index.css";
+import { useRenderGridPortal } from "./use-render-grid-portal";
+
+type GridProps = DataEditorProps;
+
+export const Grid = (props: GridProps) => {
+  useRenderGridPortal();
+
```
nonparibus, line 9: Wasn't sure where to drop this comment... below a certain number of rows, I believe it is still possible to get left with a weird partial row between the final actual row of cells, and the "new row" row. We should make sure this is flush always.

Capture 2023-03-09 at 17 51 50@2x

blocks/table-v2/LICENSE-APACHE.md
```diff
@@ -0,0 +1,191 @@
+<!-- markdownlint-disable -->
+
```
nonparibus, line 2: As part of this PR, please remove the old `blocks/table` folder and replace it with this (renaming this from `table-v2` to just `table` in the process).

We don't want the old table code (targeting an unsupported version of the spec) hanging around any more.

blocks/table-v2/src/app.tsx
yusufkinatas: Re: displaying all checkboxes

https://github.com/glideapps/glide-data-grid/commit/91ad40b8bc39a22d98994530530b037083509ab9 `glide-data-grid` will add a row-marker option called `checkbox-visible`  on their next patch (`5.2.2`). I suggest waiting until they release this (they updated their repo 9days ago, I guess they’ll release new patch soon)

blocks/table-v2/LICENSE-APACHE.md
yusufkinatas: https://github.com/hashintel/hash/pull/1946/commits/2048c0c2395c40a93e49db706cb444763e379794 done here

blocks/table-v2/src/components/grid/grid.tsx
yusufkinatas: Addressed this issue here https://github.com/hashintel/hash/pull/1946/commits/13b8a7de657b465ab8f27366cc7887841936dfb6

blocks/table-v2/src/app.tsx
yusufkinatas: Addressed the other issue here https://github.com/hashintel/hash/pull/1946/commits/13b8a7de657b465ab8f27366cc7887841936dfb6

blocks/table-v2/src/components/settings/settings.tsx
yusufkinatas: https://github.com/hashintel/hash/pull/1946/commits/13b8a7de657b465ab8f27366cc7887841936dfb6 expanded it here

blocks/table/.gitignore
```diff
@@ -0,0 +1,3 @@
+dist
+node_modules
+.blockprotocolrc
```
CiaranMn, line 3: these are covered by the root `.gitignore`, you can delete this

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
 {
   "name": "@blocks/table",
-  "version": "0.2.1",
-  "private": true,
+  "version": "0.0.0",
```
CiaranMn, line 3: suggest making this `0.0.1`

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
...
     "directory": "blocks/table"
   },
   "license": "(MIT OR Apache-2.0)",
   "author": "HASH",
   "scripts": {
     "build": "block-scripts build",
     "dev": "block-scripts dev",
-    "fix:eslint": "eslint --fix .",
-    "lint:eslint": "eslint --report-unused-disable-directives .",
+    "format": "prettier --write --ignore-unknown .",
```
CiaranMn, line 19: Covered by root formatter
```suggestion
```

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
...
+    "url": "https://github.com/hashintel/hash.git#main",
     "directory": "blocks/table"
   },
   "license": "(MIT OR Apache-2.0)",
   "author": "HASH",
   "scripts": {
     "build": "block-scripts build",
     "dev": "block-scripts dev",
-    "fix:eslint": "eslint --fix .",
-    "lint:eslint": "eslint --report-unused-disable-directives .",
```
CiaranMn, line 22: These should be here and you should keep the `eslintrc.cjs` file

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
...
-    "@local/eslint-config": "0.0.0-private",
-    "@local/tsconfig": "0.0.0-private",
-    "@types/react-table": "7.7.1",
-    "block-scripts": "0.0.14",
-    "eslint": "8.33.0",
-    "mock-block-dock": "0.0.38",
+    "@types/react-dom": "18.0.9",
+    "block-scripts": "0.1.0",
+    "mock-block-dock": "0.1.0",
+    "prettier": "2.8.2",
```
CiaranMn, line 40:
```suggestion
```

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
...
+    "@blockprotocol/graph": "0.1.0",
+    "@glideapps/glide-data-grid": "5.2.1",
+    "clsx": "1.2.1",
+    "immer": "9.0.6",
+    "react-laag": "2.0.5",
+    "rooks": "7.4.3"
   },
   "devDependencies": {
-    "@local/eslint-config": "0.0.0-private",
-    "@local/tsconfig": "0.0.0-private",
```
CiaranMn, line 36: These should remain

blocks/table/package.json
```diff
@@ -1,93 +1,63 @@
...
-    "@types/react-table": "7.7.1",
-    "block-scripts": "0.0.14",
-    "eslint": "8.33.0",
```
CiaranMn, line 39: This should remain

blocks/table/src/app.tsx
```diff
@@ -0,0 +1,305 @@
...
+import {
+  CompactSelection,
+  DataEditorProps,
+  GridCellKind,
+  GridColumn,
+  GridSelection,
+  Rectangle,
+} from "@glideapps/glide-data-grid";
+import produce from "immer";
+import styles from "./base.module.scss";
```
CiaranMn, line 18: when you restore linting these should be autosorted

blocks/table/src/app.tsx
```diff
@@ -0,0 +1,305 @@
...
+  // this should definitely be a sync function
+  const handleCellsEdited: DataEditorProps["onCellsEdited"] = (newValues) => {
+    const newRows = produce(rows, (draftRows) => {
+      for (const { value, location } of newValues) {
+        const [colIndex, rowIndex] = location;
+
+        const columnId = columns[colIndex]?.id;
+
+        if (columnId) {
+          // @ts-ignore
```
CiaranMn, line 114: if we `expect-error` we will be warned if this line becomes unnecessary – also suggest mentioning what error we are expecting in case more appear when someone revisits this
```suggestion
          // @ts-expect-error -- type instantiation is deep and possibly infinite
```

blocks/table/src/app.tsx
```diff
@@ -0,0 +1,305 @@
...
+    updateEntity({ [localRowsKey]: newRows });
+
+    return true;
+  };
+
+  const setTitle = async (val: string) => {
+    await updateEntity({ [titleKey]: val });
+  };
+
+  const updateEntity = async (newProperties: RootEntity["properties"]) => {
```
CiaranMn, line 129: `updateEntity` is defined after it is used, I'd move this up (linting will probably flag this)

blocks/table/src/app.tsx
```diff
@@ -0,0 +1,305 @@
...
+            onClose={() => setHeaderMenu(undefined)}
+            updateTitle={(title) => {
+              updateEntity({
+                [localColumnsKey]: localColumns.map((col, index) =>
+                  index === headerMenu.col
+                    ? { ...col, [columnTitleKey]: title }
+                    : col,
+                ),
+              });
+              // update column title here
```
CiaranMn, line 299: what's this for? is it a `@todo`?
```suggestion
              // update column title here
```

CiaranMn (commented):
Thanks Yusuf – tested this out and works well.

I've a few comments, mainly around making sure this has the same linting and commands as other blocks in the repo – once you reactive linting there might be a few other things to fix, hopefully mostly autofixable

blocks/table/package.json
yusufkinatas: Since previous version of the table block was `0.2.0`, could making it `0.2.1` makes more sense?

CiaranMn: Eh, I lean towards no, because this is really just a new block that happens to be taking over the name of the old one. It has no relation to it other than the name. I don't feel strongly about it though – maybe @nonparibus has a preference

yusufkinatas: I thought there might be a technical reason to always increase the version while using the same `name` for the package (everything is the same, directory, package name etc.)

CiaranMn: If anyone was using / relying on the old one I think there would be, but they aren't at the moment. I think restarting the version gives a more accurate description of the development history of this specific block – i.e. it's just started

nonparibus: I would personally increment the version number still. I think it is extremely confusing not to do so. A whole rewrite might be a v1 -> v2 jump in ordinary product-land, but with this being pre-v1 I would just go 0.2 -> 0.3

I absolutely don't think the version number should _decrease_ while using the same reserved block name. Recipe for confusion.

nonparibus: Please make this `0.3`

yusufkinatas:
@CiaranMn Addressed all of the comments and fixed some eslint errors here https://github.com/hashintel/hash/pull/1946/commits/4668257c96693f4e8c19bed6db273ecac15bc3f4

yusufkinatas:
I guess next step before merging this is, if it gets approved we should create the block entity type at BP production env, then merge this?

CiaranMn (approved):

PR #1947, opened 2023-01-25
implement value deserializers

indietyp:

These deserializers only act upon a single value and are inspired by `serde`. They enable us to easily implement types like `Option<T>`.

## 🔗 Related links

- [Inspiration](https://docs.rs/serde/latest/serde/de/value/index.html)

### DIFF
--- a/libs/deer/src/macros.rs
+++ b/libs/deer/src/macros.rs
@@ -1,0 +1,103 @@
+// Adapted from serde
+
+#[macro_export(local_inner_macros)]
+macro_rules! forward_to_deserialize_any {
+    (<$visitor:ident: Visitor<$lifetime:tt>> $($func:ident)*) => {
+        $(forward_to_deserialize_any_helper!{$func<$lifetime, $visitor>})*
+    };
+    // This case must be after the previous one.
+    ($($func:ident)*) => {
+        $(forward_to_deserialize_any_helper!{$func<'de, V>})*
+    };
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! forward_to_deserialize_any_method {
+    ($func:ident < $l:tt, $v:ident > ()) => {
+        #[inline]
+        fn $func<$v>(self, visitor: $v) -> error_stack::Result<$v::Value, $crate::DeserializerError>
+        where
+            $v: $crate::Visitor<$l>,
+        {
+            self.deserialize_any(visitor)
+        }
+    };
+}
+
+#[doc(hidden)]
+#[macro_export(local_inner_macros)]
+macro_rules! forward_to_deserialize_any_helper {
+    (bool < $l:tt, $v:ident >) => {
+        forward_to_deserialize_any_method! {deserialize_bool<$l, $v>()}
+    };
+    (i8 < $l:tt, $v:ident >) => {
+        forward_to_deserialize_any_method! {deserialize_i8<$l, $v>()}
+    };
+    (i16 < $l:tt, $v:ident >) => {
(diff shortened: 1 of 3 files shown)


libs/deer/src/value/bytes.rs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        BytesDeserializer::new(self, context)
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub struct BorrowedBytesDeserializer<'a, 'de> {
+    context: &'a Context,
+    value: &'de [u8],
+}
```
thehabbos007, line 66: What are the differences for `BorrowedBytesDeserializer` and others (`StrDeserializer`, `NoneDeserializer`, etc.) that can't use `impl_owned`? Is is the fact that they require additional lifetimes to be specified?

libs/deer/src/macros.rs
```diff
@@ -0,0 +1,103 @@
...
+    ($func:ident < $l:tt, $v:ident > ()) => {
+        #[inline]
+        fn $func<$v>(self, visitor: $v) -> error_stack::Result<$v::Value, $crate::DeserializerError>
+        where
+            $v: $crate::Visitor<$l>,
+        {
+            self.deserialize_any(visitor)
+        }
+    };
+}
```
thehabbos007, line 26: Would it not be sufficient to provide default impls. for all these in `Deserializer` instead of these macros or are we reyling on being able to specify generics in a complicated manner?

thehabbos007 (commented):
Thanks, Bilal! I am wondering if there is a simpler way using default impls for the trait (haven't played around with it on this PR), but it feels like the repetition in the code is only partially addressed through the macro setup

libs/deer/src/value/bytes.rs
indietyp: That is correct, the additional lifetimes would complicate the macro too much, so for now, I did not include them.

libs/deer/src/macros.rs
indietyp: I know the pain; this is possible, but it would lead implementors to accidentally miss a function when implementing. This way, they need to implement every function without accidentally missing a deserializer function, and if they want to, they can use the macro to short-circuit. This is also what serde does.

So while technically possible, it would make potential new implementations more error-prone, IMO.

What do you think?

thehabbos007: I expected that to be the reason. A subtle bug due to missing a case seems way more problematic than having some macro indirection.

I do think that an extensive test suite should pick up any problems with that and make the default impls a better option. What about we keep this for now, but try to use default implementations at some point in the future when we're confident (i.e. we have extensive tests)?

indietyp: I am not 100% confident with either (default implementation would be ok for us, but foreign crates implementing `Deserializer` would suffer if they do not have extensive tests). As you outlined above, the current version also has problems associated with it.

I'd move this to the backlog for now, and maybe through discussion, we are able to come to a better solution! :D I am very much open to ideas.

But I do not expect that there's a large number of implementations for `Deserializer` (only data formats will implement them), see serde or any other serialization/deserialization crate, there are a dozen `Deserializer` implementations but thousands of `Deserialize` implementations. So the tradeoff macro vs. default impl is hard to quantify 🤔 Maybe once we touch documentation an implementation guide could help with that.

thehabbos007: Sure, sounds good. I am okay with merging this right now, then :)

thehabbos007 (approved):

PR #1954, opened 2023-01-26
Improve `Debug` output for time-related structs

TimDiekmann:

Currently, the time interval output is very unreadable in log files, this is especially annoying when trying to debug the Graph and it's cluttered with struct names as the `Debug` implementation is used to show the prepared parameters, which are passed to Postgres. This small PR uses the common convention to display time intervals.


thehabbos007 (approved):
Looks good to me :+1:

PR #1955, opened 2023-01-26
Refetch block subgraph when block entity is updated

yusufkinatas:

To keep the block subgraphs provided to blocks up-to-date when they get updated, we need to refetch the block subgraph. Ideally, this should be handled by Collab service in future.

For more context, check this [Slack thread](https://hashintel.slack.com/archives/C022217GAHF/p1674725747666169) _(internal)_

### DIFF
--- a/apps/hash-frontend/src/components/block-loader/block-loader.tsx
+++ b/apps/hash-frontend/src/components/block-loader/block-loader.tsx
@@ -16,9 +16,11 @@ import {
 import { useBlockLoadedContext } from "../../blocks/on-block-loaded";
 import { useBlockContext } from "../../blocks/page/block-context";
 import { useFetchBlockSubgraph } from "../../blocks/use-fetch-block-subgraph";
+import { useBlockProtocolAggregateEntities } from "../hooks/block-protocol-functions/knowledge/use-block-protocol-aggregate-entities";
 import { useBlockProtocolFileUpload } from "../hooks/block-protocol-functions/knowledge/use-block-protocol-file-upload";
 import { useBlockProtocolUpdateEntity } from "../hooks/block-protocol-functions/knowledge/use-block-protocol-update-entity";
 import { RemoteBlock } from "../remote-block/remote-block";
+import { fetchEmbedCode } from "./fetch-embed-code";
 
 type BlockLoaderProps = {
   blockEntityId?: EntityId; // @todo make this always defined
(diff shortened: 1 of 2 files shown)


apps/hash-frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -141,7 +140,20 @@ export const BlockLoader: FunctionComponent<BlockLoaderProps> = ({
...
+          const res = await updateEntity(...args);
+
+          const newBlockSubgraph = await fetchBlockSubgraph(
+            blockEntityTypeId,
+            blockEntityId,
+          );
+          setBlockSubgraph(newBlockSubgraph);
+
+          return res;
+        },
```
CiaranMn, line 155: Why not just add this in the object it was already in? This is making a new object, do we need to?

CiaranMn (approved):

apps/hash-frontend/src/components/block-loader/block-loader.tsx
yusufkinatas: Did that to easily get the types of `...args`, thinking this is a temporary fix. But I can give the type there as well.

CiaranMn: I think these callback might become stale in the graph hadnler too, if `blockEntityId` changes – can you memoize this based on the value of its dependencies, and in `RemoteBlock` add htis please?

```ts
  useEffect(() => {
    if (graphService) {
      graphService.registerCallbacks(graphCallbacks);
    }
  }, [graphCallbacks, graphService]);
```
this will reregister the callbacks if they change

CiaranMn: It's an existing defiency with the `useGraphEmbedderService` hook that it does not update the callbacks if they change – it just takes whatever was first passed in. We have a [task](https://app.asana.com/0/1201095311341924/1203257689029614/f) to improve this so that callers don't need to manually re-register callbacks when they change

yusufkinatas: Addressed here https://github.com/hashintel/hash/pull/1955/commits/93feab3f9bee16354259ee6b8f9588005c33c0bf

CiaranMn (approved):

yusufkinatas:
trunk merge

PR #1956, opened 2023-01-26
Support nested fields and array index for ontology schemas

TimDiekmann:

We have the functionality to query for nested fields and arrays in entity properties but not for ontology schema. The functionality can be transferred, so this PR adds the functionality to the `"schema"` field.
This, however, does not expose querying for the schema, yet.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679373/1203821263193159/f) _(internal)_

## 🚫 Blocked by

- #1952 

## 🔍 What does this change?

- Renames `JsonField::JsonParameter` to `JsonField::JsonPathParameter` to make the intention more clear
- Adds `JsonPath` to the `Schema` variants for ontology types
- Adds `Schema` to ontology type tokens, but hides them via adding `#[serde(skip)]` for now

### DIFF
--- a/apps/hash-graph/lib/graph/src/ontology/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/ontology/data_type.rs
@@ -202,7 +202,7 @@ impl fmt::Display for DataTypeQueryPath<'_> {
             Self::VersionedUri => fmt.write_str("versionedUri"),
             Self::OwnedById => fmt.write_str("ownedById"),
             Self::UpdatedById => fmt.write_str("updatedById"),
-            Self::Schema(Some(path)) => write!(fmt, "schema{path}"),
+            Self::Schema(Some(path)) => write!(fmt, "schema.{path}"),
             Self::Schema(None) => fmt.write_str("schema"),
             Self::Title => fmt.write_str("title"),
             Self::Description => fmt.write_str("description"),
--- a/apps/hash-graph/lib/graph/src/ontology/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/ontology/data_type.rs
@@ -6,7 +6,7 @@ use serde::{
 };
 use utoipa::ToSchema;
 
-use crate::store::query::{OntologyQueryPath, ParameterType, QueryPath};
+use crate::store::query::{JsonPath, OntologyQueryPath, ParameterType, PathToken, QueryPath};
 
 /// A path to a [`DataType`] field.
 ///
@@ -18,7 +18,7 @@ use crate::store::query::{OntologyQueryPath, ParameterType, QueryPath};
 // TODO: Adjust enum and docs when adding non-primitive data types
 //   see https://app.asana.com/0/1200211978612931/1202464168422955/f
 #[derive(Debug, PartialEq, Eq)]
-pub enum DataTypeQueryPath {
+pub enum DataTypeQueryPath<'p> {
     /// The [`BaseUri`] of the [`DataType`].
     ///
     /// ```rust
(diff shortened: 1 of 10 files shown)


apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -315,7 +315,8 @@ impl fmt::Display for EntityTypeQueryPath {
             Self::VersionedUri => fmt.write_str("versionedUri"),
             Self::OwnedById => fmt.write_str("ownedById"),
             Self::UpdatedById => fmt.write_str("updatedById"),
-            Self::Schema => fmt.write_str("schema"),
+            Self::Schema(Some(path)) => write!(fmt, "schema{path}"),
```
TimDiekmann, line 318:
```suggestion
            Self::Schema(Some(path)) => write!(fmt, "schema.{path}"),
```

apps/hash-graph/lib/graph/src/ontology/property_type.rs
```diff
@@ -220,7 +220,8 @@ impl fmt::Display for PropertyTypeQueryPath {
             Self::VersionedUri => fmt.write_str("versionedUri"),
             Self::OwnedById => fmt.write_str("ownedById"),
             Self::UpdatedById => fmt.write_str("updatedById"),
-            Self::Schema => fmt.write_str("schema"),
+            Self::Schema(Some(path)) => write!(fmt, "schema{path}"),
```
thehabbos007, line 223:
```suggestion
            Self::Schema(Some(path)) => write!(fmt, "schema.{path}"),
```

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -259,10 +259,10 @@ pub enum EntityTypeQueryPath {
     /// Only used internally and not available for deserialization.
     VersionId,
     /// Only used internally and not available for deserialization.
-    Schema,
+    Schema(Option<JsonPath<'p>>),
```
thehabbos007, line 262: What would be an easy way to test this locally now that this is not deserializable?

thehabbos007 (commented):
Looks good to me, have some comments :)

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
TimDiekmann: You probably just have to remove `#[sersde(skip)]` on the `EntityTypeQueryToken::Schema`

apps/hash-graph/lib/graph/src/ontology/data_type.rs
```diff
@@ -202,7 +202,8 @@ impl fmt::Display for DataTypeQueryPath {
             Self::VersionedUri => fmt.write_str("versionedUri"),
             Self::OwnedById => fmt.write_str("ownedById"),
             Self::UpdatedById => fmt.write_str("updatedById"),
-            Self::Schema => fmt.write_str("schema"),
+            Self::Schema(Some(path)) => write!(fmt, "schema{path}"),
```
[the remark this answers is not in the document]
TimDiekmann: Thanks! 😅

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
TimDiekmann: Thanks for spotting this! I fixed that in https://github.com/hashintel/hash/pull/1956/commits/3e981db0627da564b5eead724e816b3e6f2d198b. The approach is not necessarily pretty, but it's out of scope to solve this problem in this PR.

thehabbos007: Awesome, thanks! Works now :)

thehabbos007 (approved):
Thank you! I tested locally by uncommenting the `serde(skip)`. Works just as expected :)

PR #1960, opened 2023-01-27
Update `utoipa` to v3.0.0

TimDiekmann:

Currently, we rely on a dev-version of that crate, this PR migrates to the latest release

## 🔍 What does this change?

- Bump utoipa to `3.0.0`
- Adjust changes to `ToSchema` trait
- Use a few more `derive(ToSchema)` instead of manually implementing it. I expected to be able to do that on way more structs, but the changes to the `ToSchema` trait seem to be not used in the derive macro.


apps/hash-graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -26,14 +26,16 @@ pub enum KnowledgeGraphOutwardEdges {
...
 // WARNING: This MUST be kept up to date with the enum variants.
-//   We have to do this because utoipa doesn't understand serde untagged:
-//   https://github.com/juhaku/utoipa/issues/320
-impl ToSchema for KnowledgeGraphOutwardEdges {
-    fn schema() -> RefOr<Schema> {
-        OneOfBuilder::new()
-            .item(<OutwardEdge<KnowledgeGraphEdgeKind, EntityIdAndTimestamp>>::schema())
-            .item(<OutwardEdge<SharedEdgeKind, OntologyTypeEditionId>>::schema())
-            .into()
+//   We have to do this because utoipa does use `OutwardEdge` as the schema reference
```
Alfred-Mountfield, line 29: I don't understand what this means

apps/hash-graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -134,7 +136,7 @@ impl Edges {
                         Entry::Occupied(entry) => {
                             entry.into_mut().insert(
                                 id.version(),
-                                edges
+                                edges,
```
Alfred-Mountfield, line 139: Not important, but any idea why this changed? Wouldn't a rustfmt check have failed a previous PR if this was a chance from it?

apps/hash-graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
```diff
@@ -153,16 +155,21 @@ impl Edges {
...
-            .into()
+impl ToSchema<'_> for Edges {
+    fn schema() -> (&'static str, RefOr<Schema>) {
+        (
+            "Edges",
+            ObjectBuilder::new()
+                .additional_properties(Some(Schema::from(
+                    ObjectBuilder::new().additional_properties(Some(Array::new(
+                        OneOfBuilder::new()
+                            .item(Ref::from_schema_name(OntologyOutwardEdges::schema().0))
```
Alfred-Mountfield, line 166: Nice to remove the magic strings!

apps/hash-graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -24,14 +24,17 @@ pub enum GraphElementId {
 // WARNING: This MUST be kept up to date with the enum variants.
 //   We have to do this because utoipa doesn't understand serde untagged:
 //   https://github.com/juhaku/utoipa/issues/320
```
Alfred-Mountfield, line 26: Does this need updating (and other similar ones)

apps/hash-graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -81,11 +84,14 @@ impl From<EntityVertexId> for GraphElementVertexId {
...
-        openapi::OneOfBuilder::new()
-            .item(openapi::Ref::from_schema_name("OntologyTypeEditionId"))
-            .item(openapi::Ref::from_schema_name("EntityVertexId"))
-            .into()
+impl ToSchema<'_> for GraphElementVertexId {
+    fn schema() -> (&'static str, openapi::RefOr<openapi::Schema>) {
+        (
+            "GraphElementVertexId",
+            openapi::OneOfBuilder::new()
+                .item(openapi::Ref::from_schema_name("OntologyTypeEditionId"))
```
Alfred-Mountfield, line 92: Can you use the trick with calling `schema().0` to avoid the magic strings here?

apps/hash-graph/lib/graph/src/shared/identifier/time/version.rs
```diff
@@ -14,17 +14,20 @@ pub struct VersionInterval<A> {
...
-                .build(),
+impl<A> ToSchema<'_> for VersionInterval<A> {
+    fn schema() -> (&'static str, openapi::RefOr<openapi::Schema>) {
+        (
+            "VersionInterval",
+            openapi::Schema::Object(
+                openapi::ObjectBuilder::new()
+                    .property("start", Timestamp::<A>::schema().1)
+                    .required("start")
+                    .property("end", openapi::Ref::from_schema_name("NullableTimestamp"))
```
Alfred-Mountfield, line 25: Similarly, are you able to rely on `schema().0` here, or not?

libs/@local/hash-graph-client/api.ts
```diff
@@ -1918,7 +1836,8 @@ export interface Subgraph {
  */
 export type TimeIntervalBound =
   | TimeIntervalBoundOneOf
-  | TimeIntervalBoundOneOf1;
+  | TimeIntervalBoundOneOf1
+  | TimeIntervalBoundOneOf2;
```
Alfred-Mountfield, line 1840: This is still equivalent right?

apps/hash-graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: It's using `Ref::from_schema_name("OutwardEdge")` (not even `OutwardEdge::<_, _>::schema().0)`) to generate the schema, which results in a wrong schema.

TimDiekmann: Sure: 3d9fa77ebce40e91105d16e9171c6b38d32f96cd

apps/hash-graph/lib/graph/src/api/rest/utoipa_typedef/subgraph/edges.rs
TimDiekmann: I don't know, it seems that `rustfmt` does bail on this function. For consistency, I also added another `,` below in 3d9fa77

apps/hash-graph/lib/graph/src/shared/identifier/mod.rs
TimDiekmann: Even better: I was able to derive it: 1ccc2695f81b510a40af5c0b5453237bc78b7669

apps/hash-graph/lib/graph/src/shared/identifier/time/version.rs
TimDiekmann: That's not possible here as we don't have a `NullableTimestamp` struct.

apps/hash-graph/lib/graph/src/shared/identifier/mod.rs
TimDiekmann: I was able to derive it by adding a schema for `BaseUri`: 1dfe10810e8ffa893be9fe0e2a997c744609d5c0

libs/@local/hash-graph-client/api.ts
```diff
@@ -1096,31 +1096,6 @@ export const KnowledgeGraphVertexOneOfKindEnum = {
 export type KnowledgeGraphVertexOneOfKindEnum =
   (typeof KnowledgeGraphVertexOneOfKindEnum)[keyof typeof KnowledgeGraphVertexOneOfKindEnum];
 
-/**
- * A record of an [`Entity`] that has been persisted in the datastore, with its associated metadata.
- * @export
- * @interface KnowledgeGraphVertexOneOfInner
- */
-export interface KnowledgeGraphVertexOneOfInner {
```
thehabbos007, line 1104: Good thing this is just `Entity` now :D

thehabbos007 (approved):
This looks good to me! The tuple return type of `ToSchema` turned out to be quite helpful, but it.s annoying we have `.0` and `.1` accesses everywhere. I suppose we can't do much about that. At least the types are different :sweat_smile: 

I ran type checks for TS locally and nothing seems to be broken. Looks good to me :+1:

PR #1962, opened 2023-01-27
Bring `td/dev/db-changes` in line with `main`

TimDiekmann:

We are going to merge `td/dev/db-changes` into `main` soon™️, so we need to make sure it's in line with `main`.

## 🔍 What does this change?

This is the `td/dev/db-changes` branch with an almost clean merge.


TimDiekmann:
GitHub is not able to detect a simple merge.

PR #1965, opened 2023-01-27
Fix bug that tables may be joined multiple times

TimDiekmann:

@indietyp encountered an issue, where the same table was joined multiple times with the same name. The bug occurred, if from the same base table the same table was joined multiple times, where another table was joined on the joined tables, e.g. `entity_types` were joined on `entities` twice and on each `entity_types` an `type_ids` table was joined. The `entity_types` tables are intermediate and not required in the final query, so they are optimized out. However, then `type_ids` was joined with the same name. This PR fixes this issue.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203811012038440/f) _(internal)_
- [Slack discussion](https://hashintel.slack.com/archives/C03TWBSRT8B/p1674570505708979) _(internal)_

## 🔍 What does this change?

- Instead of modifying the last table on the stack, we pop it and treat it as the current table unless we need it, then we push it back to the stack
- Reverse the logic, when an entity table is pinned as the `current_table` is not necessarily an entity table

## 🛡 What tests cover this?

A test case was added to match the expected result. Without this fix, the test would fail.

## ❓ How to test this?

Run this query with and without a fix. Without the fix, there should be an error. Wiih the fix the query will succeed (but may not return any roots)
```json

{
  "filter": {
    "all": [
      {
        "equal": [
          { "path": ["leftEntity", "type", "baseUri"] },
          { "parameter": "http://localhost:3000/@example-org/types/entity-type/category/" }
        ]
      },
      {
        "equal": [
          { "path": ["rightEntity", "type", "baseUri"] },
          { "parameter": "http://localhost:3000/@example-org/types/entity-type/category/" }
        ]
      }
    ]
  }
}
```

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
@@ -477,10 +477,14 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
     fn add_join_statements(&mut self, path: &R::QueryPath<'_>) -> Alias {
         let mut current_table = self.statement.from;
 
+        if current_table.table == Table::Entities {
+            self.pin_entity_table(current_table.alias);
+        }
+
         for relation in path.relations() {
             let current_alias = current_table.alias;
             for (current_column, join_column) in relation.joins() {
-                let current_column = current_column.aliased(current_table.alias);
+                let mut current_column = current_column.aliased(current_table.alias);
                 let mut join_column = join_column.aliased(Alias {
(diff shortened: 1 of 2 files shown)


indietyp (approved):
Works on the query and dataset where discovered no longer throws the previously encountered error.

> **Edit:** I have also reviewed the code. I am still relatively new to the source code of the hash graph and query compiler, so proceed with caution, but LGTM 🎉

thehabbos007 (approved):
LGTM as well, CI passes and I tested with `test-rest-api` locally. I suppose we know that we changed this if any errors pop up :)

PR #1972, opened 2023-01-30
Rename `type_ids` table and `version_id` column

TimDiekmann:

We use the term `type_id` when referring to a versioned URI. Similarly, the term `version_id` is confusing. This PR renames the `type_ids` table to `ontology_ids` and the `version_id` column to `ontology_id`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203214689883089/f) _(internal)_

## 🔍 What does this change?

- Rename `type_ids` to `ontology_ids`
- Rename `ontology_ids::version_id` to `ontology_ids::ontology_id`

This is an internal change only

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/ontology/ontology_id.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/ontology/ontology_id.rs
@@ -5,15 +5,12 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 use uuid::Uuid;
 
-// TODO: rename this to TypeInternalVersionId or something to distinguish it from versioned
-//  URIs and from entity ids
-//  https://app.asana.com/0/1202805690238892/1203214689883089/f
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema, FromSql, ToSql)]
 #[repr(transparent)]
 #[postgres(transparent)]
-pub struct VersionId(Uuid);
+pub struct OntologyId(Uuid);
 
-impl fmt::Display for VersionId {
+impl fmt::Display for OntologyId {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(fmt, "{}", &self.0)
     }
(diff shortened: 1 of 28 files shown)


apps/hash-graph/lib/graph/src/store/postgres/ontology/ontology_id.rs
```diff
@@ -5,15 +5,12 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 use uuid::Uuid;
 
-// TODO: rename this to TypeInternalVersionId or something to distinguish it from versioned
-//  URIs and from entity ids
-//  https://app.asana.com/0/1202805690238892/1203214689883089/f
```
thehabbos007, line 10: I didn't even know we had this task when you talked about doing this. Great thing we already wanted to change this!

apps/hash-graph/postgres_migrations/V2__ontology_tables.sql
```diff
@@ -1,11 +1,11 @@
 CREATE TABLE IF NOT EXISTS
-  "type_ids" (
-    "version_id" UUID PRIMARY KEY,
+  ontology_ids (
+    ontology_id UUID PRIMARY KEY,
```
thehabbos007, line 3: small nit, does our formatter not ensure quoting/non-quoting for table/column names? I do like to quote column names as it doesn't get strange highlighting overlaps with builtin keywords.

thehabbos007 (commented):
Looks good to me, Tim! I just have one comment in the migration files

apps/hash-graph/lib/graph/src/store/postgres/ontology/ontology_id.rs
TimDiekmann: Me neither, I found the TODO and was happy to remove it 😆

apps/hash-graph/postgres_migrations/V2__ontology_tables.sql
TimDiekmann: The formatter does not really work as expected. I fixed this in https://github.com/hashintel/hash/pull/1975, but I'm happy to do this in this PR as well.

thehabbos007: No worries, I'm fine with the change being in that PR instead :+1:

thehabbos007 (approved):

PR #1975, opened 2023-01-30
Rename `Entity` related ID columns and Graph representation

TimDiekmann:

We currently use `edition_id` in the database and `record_id` in the Graph. This changes both names to `revision_id`.

## 🚫 Blocked by

- #1972 

## 🔍 What does this change?

- Swap naming for `EntityEditionId` and `EntityRecordId`
- Rename `EntityVersion` to `EntityReivision`
- Adjust surrounding code to match variables and methods the new naming schema
- Drive-by: Remove unused code snippets from the query compiler

Generally, this applies this schema:
```ts
type EntityRecordId = {
  entityId: EntityId;
  editionId: <Opaque>; // currently u64, will be changed to string
};

type EntityVertexId = {
  baseId: EntityId;
  revisionId: RevisionId; // Timestamp
};

type EntityRevision {
    transactionTime: TimeInterval,
    decisionTime: TimeInterval,
}
```

## 🐾 Next steps

This only changes the Graph. We want to do the same renaming in the rest of the repository as well, but not yet.


apps/hash-graph/lib/graph/src/knowledge/entity/mod.rs
```diff
@@ -173,8 +173,10 @@ impl LinkData {
 // TODO: deny_unknown_fields on other structs
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct EntityMetadata {
-    edition_id: EntityEditionId,
-    version: EntityVersion,
+    #[serde(rename = "editionId")]
+    record_id: EntityRecordId,
+    #[serde(rename = "version")]
+    revision: EntityRevision,
```
TimDiekmann, line 179: Renaming is done to avoid adjusting the REST API

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
```diff
@@ -347,7 +347,7 @@ impl<'de> Visitor<'de> for EntityQueryPathVisitor {
 
         Ok(match token {
             EntityQueryToken::Uuid => EntityQueryPath::Uuid,
-            EntityQueryToken::RecordId => EntityQueryPath::RecordId,
+            EntityQueryToken::RecordId => EntityQueryPath::EditionId,
```
TimDiekmann, line 350: `EntityQueryToken::RecordId` was not renamed as I avoided touching the REST API. `utoipa` does not understand `serde(alias)`, which would have been the preferred way.

apps/hash-graph/lib/graph/src/ontology/mod.rs
```diff
@@ -197,16 +197,11 @@ pub struct DataTypeWithMetadata {
 }
 
 impl Record for DataTypeWithMetadata {
-    type EditionId = OntologyTypeEditionId;
     type QueryPath<'p> = DataTypeQueryPath<'p>;
-    type VertexId = Self::EditionId;
-
-    fn edition_id(&self) -> &Self::EditionId {
-        self.metadata().edition_id()
-    }
```
TimDiekmann, line 206: This is not used anywhere

apps/hash-graph/lib/graph/src/shared/identifier/knowledge.rs
```diff
@@ -163,9 +163,11 @@ impl EntityRecordId {
...
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Serialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
-pub struct EntityEditionId {
-    base_id: EntityId,
-    record_id: EntityRecordId,
+pub struct EntityRecordId {
+    #[serde(rename = "baseId")]
+    entity_id: EntityId,
+    #[serde(rename = "recordId")]
+    edition_id: EntityEditionId,
```
TimDiekmann, line 170: Same as above: I avoided changing the REST API

apps/hash-graph/lib/graph/src/shared/identifier/mod.rs
```diff
@@ -39,13 +39,17 @@ impl ToSchema for GraphElementId {
 #[serde(rename_all = "camelCase")]
 pub struct EntityVertexId {
     base_id: EntityId,
-    version: Timestamp<ProjectedTime>,
+    #[serde(rename = "version")]
+    revision_id: Timestamp<ProjectedTime>,
```
TimDiekmann, line 43: Same as above: I avoided changing the REST API

apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -283,31 +283,6 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
         }
     }
 
-    fn compile_latest_entity_version_filter(
```
TimDiekmann, line 286: We use temporal versioning to determine "latest", it was forgotten to be removed.

thehabbos007 (commented):
It looks good to me, have tested locally and the app seems to work. I think it would be good if @Alfred-Mountfield takes a look as well!

apps/hash-graph/lib/graph/src/knowledge/entity/mod.rs
```diff
@@ -173,8 +173,10 @@ impl LinkData {
...
 pub struct EntityMetadata {
-    edition_id: EntityEditionId,
-    version: EntityVersion,
+    #[serde(rename = "editionId")]
+    record_id: EntityRecordId,
+    #[serde(rename = "version")]
+    revision: EntityRevisionVersion,
     #[schema(value_type = String)]
     entity_type_id: VersionedUri,
     #[serde(rename = "provenance")]
```
Alfred-Mountfield, line 182: The PR that will be opened soon on the BP will have the following shape and names for this:

```ts
export type EntityMetadata<Temporal extends boolean> = {
  recordId: EntityRecordId;
  entityTypeId: VersionedUri;
  temporalVersioning: Temporal extends true
    ? EntityTemporalVersioningMetadata
    : never;
};
```

(You can ignore the conditional bit as in HASH it will always be versioned)

I'd probably suggest we revert the `revision` metadata field rename in this PR, and address it in a follow-up once we've decided on the terminology in the BP. I feel pretty strongly that we _don't_ call the field `revision` here, because a revision of the entity _is a specific instance_ of it, it's not the identifier of that instance. 

I'm suggesting we revert rather than modifying, because I really don't like the term `EntityTemporalVersioningMetadata` and I'd like to see us find something better, but that discussion should really happen in the upcoming PR and I appreciate we don't want to block this PR on that as there are more things stacked on top of it

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
```diff
@@ -43,37 +43,37 @@ pub enum EntityQueryPath<'p> {
     /// [`EntityId`]: crate::identifier::knowledge::EntityId
     /// [`Entity`]: crate::knowledge::Entity
     OwnedById,
-    /// The [`EntityRecordId`] of the [`EntityRecordId`] belonging to the [`Entity`].
```
Alfred-Mountfield, line 46: Oops 😅

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
Alfred-Mountfield: When are we going to touch the REST API then? 

I think I'd be more comfortable if these PRs actually renamed things across the stack in one go each time, rather than doing it in pieces where we have an inconsistent state of the branch again. I appreciate that we're trying to get DB changes stacked and ready, and I also appreciate that we're blocked on updating the whole branch until the BP stuff is in, so making TS changes now would just lead to fighting conflicts... I defer to your judgement, just wanted to flag my thoughts.

libs/@local/hash-graph-client/api.ts
```diff
@@ -492,25 +492,6 @@ export interface Entity {
    */
   properties: object;
 }
-/**
- *
- * @export
- * @interface EntityEditionId
- */
-export interface EntityEditionId {
```
Alfred-Mountfield, line 500: Seems like the REST API ended up changed anyway. I appreciate the shape of the objects should probably still be the same, but were any of these types imported?

apps/hash-graph/bench/benches/read_scaling/knowledge/complete/entity.rs
```diff
@@ -106,8 +106,8 @@ async fn seed_db(
                             None,
```
Alfred-Mountfield, line 106: Just commenting to thread replies.

Do we know why the CI lint job is failing?

TimDiekmann:
Superseded by #1999

apps/hash-graph/lib/graph/src/knowledge/entity/mod.rs
TimDiekmann: This was also a thing I was unsure about and the main reason why I opened #1999 to only change the required things to move forward.

apps/hash-graph/bench/benches/read_scaling/knowledge/complete/entity.rs
TimDiekmann: There is a corresponding [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1675079929051119) (_internal_) and [a PR](https://github.com/hashintel/hash/pull/1973) to at least make the rest of the CI run.

PR #1976, opened 2023-01-30
Reverse logic to detect feature powerset on `cargo make`

TimDiekmann:

We have a logic in place to only run a subset in PRs and only run the full set of tests before merge. However, in order to test locally, this logic should be reversed and instead of testing for a `push` event, this should test for `pull_request` events.

## 🔗 Related links

- #1800

### DIFF
--- a/.github/scripts/rust/Makefile.toml
+++ b/.github/scripts/rust/Makefile.toml
@@ -15,8 +15,8 @@ CARGO_RUN_FLAGS = ""
 CARGO_FORMAT_FLAGS = ""
 CARGO_FORMAT_HACK_FLAGS = "--workspace"
 
-CARGO_CLIPPY_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--no-deps --all-targets --all-features", mapping = {"push" = "--no-deps --all-targets" } }
-CARGO_CLIPPY_HACK_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--workspace", mapping = {"push" = "--workspace --optional-deps --feature-powerset" } }
+CARGO_CLIPPY_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--no-deps --all-targets", mapping = {"pull_request" = "--no-deps --all-targets --all-features" } }
+CARGO_CLIPPY_HACK_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--workspace --optional-deps --feature-powerset", mapping = {"pull_request" = "--workspace" } }
 
 CARGO_DOC_FLAGS = "--no-deps --all-features -Zunstable-options -Zrustdoc-scrape-examples"
 CARGO_DOC_HACK_FLAGS = "--workspace"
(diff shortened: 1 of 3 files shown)


indietyp (approved):
LGTM

.github/scripts/rust/Makefile.toml
```diff
@@ -15,17 +15,17 @@ CARGO_RUN_FLAGS = ""
 CARGO_FORMAT_FLAGS = ""
 CARGO_FORMAT_HACK_FLAGS = "--workspace"
 
-CARGO_CLIPPY_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--no-deps --all-targets --all-features", mapping = {"push" = "--no-deps --all-targets" } }
-CARGO_CLIPPY_HACK_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--workspace", mapping = {"push" = "--workspace --optional-deps --feature-powerset" } }
+CARGO_CLIPPY_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--no-deps --all-targets", mapping = {"pull_request" = "--no-deps --all-targets --all-features" } }
+CARGO_CLIPPY_HACK_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--workspace --optional-deps --feature-powerset", mapping = {"pull_request" = "--workspace" } }
```
thehabbos007, line 19: Okay this mapping feature is so cool, very useful. Hadn't noticed that before.

.github/scripts/rust/Makefile.toml
```diff
@@ -15,17 +15,17 @@ CARGO_RUN_FLAGS = ""
...
 
 CARGO_DOC_FLAGS = "--no-deps --all-features -Zunstable-options -Zrustdoc-scrape-examples"
 CARGO_DOC_HACK_FLAGS = "--workspace"
 
 CARGO_RUSTDOC_FLAGS = "--all-features -Zunstable-options -Zrustdoc-scrape-examples -- -Zunstable-options"
 CARGO_RUSTDOC_HACK_FLAGS = "--workspace"
 
-CARGO_TEST_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--all-features", mapping = {"push" = "" } }
-CARGO_TEST_HACK_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "--workspace", mapping = {"push" = "--workspace --optional-deps --feature-powerset" } }
+CARGO_TEST_FLAGS = { source = "${GITHUB_EVENT_NAME}", default_value = "", mapping = {"pull_request" = "--all-features" } }
```
thehabbos007, line 27: I notice that this doesn't have the `all-features` flag as as default anymore while the clippy flags does. Is this intentional?

thehabbos007 (approved):
Looks good to me too, thanks. I have a small question about the `CARGO_TEST_FLAGS` :)

.github/scripts/rust/Makefile.toml
TimDiekmann: The `--all-features` flag is implied by `--feature-powerset` in `CARG_TEST_HACK_FLAGS` below

PR #1977, opened 2023-01-30
Attempt to fix linting job in CI

TimDiekmann:

In various PR the Rust CI linting job is failing as `cargo make` is supposed to be installed, but it's not. This PR tries to solve this by not caching the tools (which is probably a good idea anyway). If this PR is able to be merged, this issue should be solved, otherwise, it's very likely hitting the same bug.

## 🔗  Related links

This error happened in
- #1800
- #1973 
- #1976


indietyp (approved):
LGTM. I think it's worth a try.

PR #1979, opened 2023-01-30
Change `EntityEditionId` to be a `UUID`

TimDiekmann:

To be more future proof, we want the (opaque!) edition id of an entity to be a UUID as this is easier when sharding.

## 🚫 Blocked by

- #1975 

## 🔍 What does this change?

Change the inner type of `EntityEditionId` to be a `UUID`


TimDiekmann:
Superseded by https://github.com/hashintel/hash/pull/1999

PR #1980, opened 2023-01-30
`error-stack`: remove `owo-colors` as dependency

indietyp:

Remove `owo-colors` as a dependency. This means we can push the enable `colors` support without an extra feature! `owo-colors` is being replaced with a vendored solution that does the same thing, only more correct.

This also implements the suggestion from https://github.com/hashintel/hash/pull/1800#discussion_r1089753480

### Why remove `owo-colors`?

Don't get me wrong, `owo-colors` is a great crate, but for our purposes is a bit too heavy. 

Initially, we chose `owo-colors` because of the convenience, and we thought that it might provide Windows support (my bad 😅) and automatic detection of color support.

In 0.3, we have removed automatic detection support and discovered that `owo-colors` does not - in fact - provide Windows support. In that case, it is pretty easy to create an alternative with a similar level of convenience, and after going on an adventure 🥾  with @TimDiekmann through the lands of Windows 11 terminals and discovered that pretty much every built-in terminal either supports ANSI escape sequences natively out-of-the-box, or it can be enabled (on older installations).

### Correctness of `owo-colors`

What about correctness? After looking through snapshots I have discovered that for style attributes like `bold` or `italic` `owo-colors` actually does the wrong thing when resetting the styles. I will be posting a follow-up PR for `owo-colors`, but due to it being passively maintained I think there's a low chance it will be in time for 0.3 anyway.

image

The problem: when using style attributes like `bold` (`\e[1m`) or `italic` (`\e[3m`) when you want to end the portion of text there are two options:
1) nuclear: `\e[0m`, this removes all formatting (`owo-colors` uses this)
2) 🎩 gentleman: `\e[22m` (not `bold`), `\e[23m` (not `italic`), instead of removing and resetting everything it just disables bold or italic respectively, allowing for nested styles and even colored text in italic text (not possible with the nuclear option)

We also compress as much as possible, meaning that `bold italic` is not `\e[1m\e[3m` but instead `\e[1;3m`. Small optimization, but a nice one c:

## 👂 Open Questions

I went ahead and just implemented all ANSI escape sequences, they are currently lacking tests (coming soon), but most are currently unused. Should we remove the ones that are not needed or future-proof the solution?

Do we want to expose the coloring methods through `error-stack`?

> IMO, no, if we want to go ahead with something like that, it should go in a separate crate

### DIFF
--- a/packages/libs/error-stack/Cargo.toml
+++ b/packages/libs/error-stack/Cargo.toml
@@ -21,7 +21,6 @@ categories = ["rust-patterns", "no-std"]
 tracing-error = { version = "0.2", optional = true, default_features = false }
 anyhow = { version = "1.0.65", default-features = false, optional = true }
 eyre = { version = "0.6", default-features = false, optional = true }
-owo-colors = { version = "3", default-features = false, optional = true }
 serde = { version = "1", default-features = false, optional = true }
 spin = { version = "0.9", default-features = false, optional = true, features = ['rwlock', 'once'] }
 
@@ -39,6 +38,7 @@ ansi-to-html = "0.1.2"
 once_cell = "1.17.0"
 supports-color = "2.0.0"
 supports-unicode = "1.0.2"
+owo-colors = "3"
 
 [build-dependencies]
 rustc_version = "0.4"
@@ -46,7 +46,6 @@ rustc_version = "0.4"
 [features]
 default = ["std"]
 
-color = ["dep:owo-colors"]
 spantrace = ["dep:tracing-error", "std"]
 std = ["anyhow?/std"]
 eyre = ["dep:eyre", "std"]
--- a/packages/libs/error-stack/Cargo.toml
+++ b/packages/libs/error-stack/Cargo.toml
@@ -38,7 +38,7 @@ ansi-to-html = "0.1.2"
 once_cell = "1.17.0"
 supports-color = "2.0.0"
 supports-unicode = "1.0.2"
-owo-colors = "3"
+owo-colors = "3.5.0"
 
 [build-dependencies]
 rustc_version = "0.4"
--- a/packages/libs/error-stack/Cargo.toml
+++ b/packages/libs/error-stack/Cargo.toml
@@ -71,7 +71,7 @@ required-features = ["std"]
 
 [[example]]
 name = "detect"
-required-features = ['std', 'color']
+required-features = ['std']
 
 
 [[test]]
--- a/packages/libs/error-stack/src/fmt/color.rs
+++ b/packages/libs/error-stack/src/fmt/color.rs
@@ -141,12 +141,6 @@ impl Report<()> {
 pub(crate) enum Color {
     Black,
     Red,
-    Green,
-    Yellow,
-    Blue,
-    Magenta,
-    Cyan,
-    White,
 }
 
 impl Color {
@@ -154,12 +148,6 @@ impl Color {
         match self {
             Self::Black => b'0',
             Self::Red => b'1',
-            Self::Green => b'2',
-            Self::Yellow => b'3',
-            Self::Blue => b'4',
-            Self::Magenta => b'5',
-            Self::Cyan => b'6',
-            Self::White => b'7',
         }
     }
 }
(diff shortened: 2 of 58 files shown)


packages/libs/error-stack/Cargo.toml
```diff
@@ -39,14 +38,14 @@ ansi-to-html = "0.1.2"
 once_cell = "1.17.0"
 supports-color = "2.0.0"
 supports-unicode = "1.0.2"
+owo-colors = "3"
```
TimDiekmann, line 41: We can pin a specific version here

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -68,12 +66,7 @@ static COLOR_OVERRIDE: AtomicOverride<ColorMode> = AtomicOverride::new();
 impl Report<()> {
     /// Set the color mode preference
     ///
-    /// If the value is [`None`], a previously set preference will be unset, while with [`Some`] a
-    /// specific color mode will be set, this mode will be used, otherwise if `detect` is enabled
-    /// the capabilities of the terminal are queried through [`owo_colors`].
-    ///
-    /// The value defaults to [`ColorMode::Emphasis`] (if the terminal supports it), otherwise
-    /// [`ColorMode::None`].
+    /// The value defaults to [`ColorMode::Emphasis`], otherwise [`ColorMode::None`].
```
TimDiekmann, line 69: What do you mean by "otherwise `ColorMode::None`"?

TimDiekmann (commented):
> I went ahead and just implemented all ANSI escape sequences, they are currently lacking tests (coming soon), but most are currently unused. Should we remove the ones that are not needed or future-proof the solution?

I'd prefer removing them as they provide no purpose. I'd say we add new codes as we need them.

> Do we want to expose the coloring methods through `error-stack`?

I love the simplicity of the colors, but the purpose of `error-stack` is different, so I'd say no. Maybe, we should keep the idea to publish this as a mini library at some point if there is none with this minimalistic implementation, but I think the interface of `error-stack` should remain unchanged (besides the feature flag change).

Please don't get me wrong, I love the simple support and also much appreciate the explanation on what this approach does better than other approaches.

Generally, the code looks good to me 🙂

packages/libs/error-stack/src/fmt/color.rs
indietyp: whoops, that must've been left there when we had the different color modes depending on the feature in #1800, removed in [1a19702](https://github.com/hashintel/hash/pull/1980/commits/1a1970264f0551a255344db50ec11613e5f2aed0)

packages/libs/error-stack/Cargo.toml
indietyp: fixed in [0d7afa6](https://github.com/hashintel/hash/pull/1980/commits/0d7afa6416abf3777b740df69a9b4e50cf358676)

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -1,3 +1,8 @@
+use core::{
+    fmt,
+    fmt::{Display, Formatter},
+};
```
TimDiekmann, line 4:
```suggestion
use core::fmt::{self, Display, Formatter};
```

packages/libs/error-stack/src/fmt/color.rs
```diff
@@ -68,12 +66,7 @@ static COLOR_OVERRIDE: AtomicOverride<ColorMode> = AtomicOverride::new();
 impl Report<()> {
     /// Set the color mode preference
     ///
-    /// If the value is [`None`], a previously set preference will be unset, while with [`Some`] a
-    /// specific color mode will be set, this mode will be used, otherwise if `detect` is enabled
-    /// the capabilities of the terminal are queried through [`owo_colors`].
-    ///
-    /// The value defaults to [`ColorMode::Emphasis`] (if the terminal supports it), otherwise
-    /// [`ColorMode::None`].
+    /// The value defaults to [`ColorMode::Emphasis`].
```
TimDiekmann, line 69:
```suggestion
    /// If no [`ColorMode`] is set, it defaults to [`Emphasis`].
```

packages/libs/error-stack/tests/snapshots/doc/fmt__charset_ascii.snap
```diff
@@ -1,7 +1,7 @@
 <b>invalid input parameter</b>
-<b>|- </b><i>src/fmt/charset.rs:21:5</i>
-<b>|- </b>backtrace (1)
-<b>|- </b>suggestion: oh no, try again
+|- at <i>src/fmt/charset.rs:21:5</i>
```
TimDiekmann, line 2: Do we want to remove the whitespace after `-` again as a drive-by?

TimDiekmann (commented):
Looks good to me, only a few minor things.

Also I wonder, as we default to `ColorMode::Emphasis`, if we want to use that mode in the docs as well, I don't really have an opinion on this, though.

packages/libs/error-stack/tests/snapshots/doc/fmt__charset_ascii.snap
indietyp: removed i n[20ef0a0](https://github.com/hashintel/hash/pull/1980/commits/20ef0a082978f68e92a4062a72598daa7dab4a77)

indietyp:
> Also I wonder, as we default to ColorMode::Emphasis, if we want to use that mode in the docs as well, I don't really have an opinion on this, though.

This might be a good idea, otherwise, people might have different expectations, e.g.: why are docs in color, but when I use error-stack there is none. We should then maybe add further emphasis on the different modes in the `lib.rs` intro text to ensure people are aware that colors are supported (maybe the current info is already enough?)

TimDiekmann (approved):
Great, thank you!
I also took a look at the generated docs and an issue I noticed with `owo-colors` is now gone 🎉

PR #1985, opened 2023-01-31
Move `error-stack` to `/libs`

TimDiekmann:

This moves the `error-stack` library to the `/libs` folder, which is the last task in order to restructure the `packages/` folder into `/apps` and `/libs`

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203767287750368/f) _(internal)_

## 🔍 What does this change?

- Moves all files to `/libs/error-stack`
- Update links to point to the correct location
- Remove `packages/` folder


README.md
```diff
@@ -84,10 +84,6 @@ This repository's contents is divided across four primary sections:
 
 - [`/infra/terraform`](infra/terraform): contains Terraform modules for deploying HASH on AWS
 
-### Libraries
```
indietyp, line 87: Removing this will leave a deadlink for the following lines:

https://github.com/hashintel/hash/blob/8539c0ae7760e19ee39a201b0c9a33fa9434ca4e/README.md?plain=1#L62

libs/error-stack/README.md
```diff
@@ -101,7 +101,7 @@ This will most likely result in an error and print
```
thehabbos007, line 101: (comment for line 98)
This path is still valid after the move because we are using a sha to refer to an exact commit, but just pointing out the path in `main` would be incorrect

README.md
TimDiekmann: Good spot, thanks! It should be fixed with e1f3a5fddcf4ebb945428b4405cd3ed65d1d3a27

thehabbos007 (commented):
It looks mostly good to me, I have a single comment, but I don't even thing we can/should address this right now (we can't predict the sha of the squash-commited PR)

libs/error-stack/README.md
nonparibus: Do we want to move this file into the `.github/assets` folder anyway? https://github.com/hashintel/hash/tree/main/.github/assets (If this is adding weight to `error-stack`, maybe? If it's not, then I guess it's more convenient to leave it inline? Haven't looked at this closely and am doing a brief run-through here)

nonparibus (approved):
Legal LGTM ✅

libs/error-stack/README.md
TimDiekmann: I'd leave it next to `error-stack` as it's very related and generated output form the library. However, I think we should update the link after this PR is merged.

kachkaev (dismissed):
Diff LGTM! I found a couple of remaining references to `packages/libs` but they are all justifiable.

Screenshot 2023-01-31 at 13 21 17

The first file points to git tags and the second one points to a commit. We cannot update this commit (yet) because of a chicken-and-egg problem here. Doing this some time later this week would be good.

TimDiekmann:
Thanks, @kachkaev, I probably will update that hash to point to the `0.3.0` tag when it's released 🙂

indietyp:
We were able to remove the `full.png` in #1989 as a drive-by. Therefore we no longer need to update the URL for the image 🎉

thehabbos007 (approved):
LGTM, test suite passes locally, no grep results for `packages/libs/error-stack`. Thanks for addressing the image ref in the other PR @indietyp!

nonparibus (approved):

PR #1987, opened 2023-01-31
`error-stack`: clarify documentation of `HookContext`

indietyp:

#1693 separated `fmt::HookContext` into `hook::HookContext` and `fmt::HookContext`, so that it can be used in #1558. While working, the problem is that documentation doesn't render correctly because `hook::HookContext` is private in public. Therefore unreachable rustdoc is unable to navigate to the type, and the documentation is incomplete.

This PR moves to a macro-based approach to enable better documentation. Instead of using an unreachable public type, we have a macro that implements the necessary common functions. This is by far not ideal, but I found this to be the best way without compromising #1558 or future hooks.


TimDiekmann (approved):
Thanks, @indietyp, this is much better (and also does unbreak SemVer for this struct) 🎉

indietyp:
I took the liberty of cleaning `fmt::hook` up and re-enabling `dead_code` and `unused_imports`.

TimDiekmann (approved):

indietyp:
Sorry, @TimDiekmann, I need another review D: I just realized that I can remove all cfg guards in the `hook.rs` because it is only conditionally enabled anyway.

TimDiekmann (approved):
I think the last commit was a good idea 😄

PR #1990, opened 2023-01-31
Fix Renovate not picking up toolchain for `hash-graph`

TimDiekmann:

When moving `packages/graph` to `apps/hash-graph` we forgot to update the renovate config


indietyp (approved):
LGTM

PR #1992, opened 2023-02-01
Temporarily disable `Rust / publish` job

TimDiekmann:

In order to move `error-stack` to `libs/` (#1985), which is touching the `version` line in the `Cargo.toml`, we need to disable the `publish` job, otherwise, it will try to publish the crate with an unchanged version. There is probably a better way to detect a version change, but this is currently out of scope, and moving a whole crate is unlikely to happen again in the future.

## 🔗 Related links

- #1985

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -350,7 +350,7 @@ jobs:
 
   merging-enabled:
     name: Merging enabled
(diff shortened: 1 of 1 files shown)


thehabbos007 (approved):
Let's get that move in :)

PR #1995, opened 2023-02-01
introduce new crate: `antsi`

indietyp:

`antsi` is a mini-crate that correctly supports coloring (and SGR (Select Graphic Rendition) as defined in ISO 6429) without any external crates, fast, const, no-std in an easy-to-use package with a concise and explicit.

Current crates do not support all of these feature sets, so I think that adding a create like this would be valuable to the rust ecosystem.

There are currently three actively/passively maintained libraries out there that support coloring:
- `termcolor`: std and support for older Windows versions, only supports color
- `nu-ansi-term`: std, a fork from `ansi-term`.
- `owo-colors`: no-std, convoluted API, pollutes the namespace

Why only ANSI and not dedicated windows support?

After looking through the different terminals on Windows 10 and 11, it is apparent that ANSI support is often enough for cross-platform nowadays. Older Windows versions do not natively support ANSI escape sequences, but in that case, `termcolor` is the better choice.

> and yes... I have read the relevant standards to ensure that we're compliant.

Please feel free to let me know if creating this crate in the HASH repo is ok!

The name: antsi

Names are hard, and this came immediately to mind, like most of my names it is a play on words, this time: 3 in one!
* `ant` as in smol, but still very useful, like an ant
* `ansi` as in ANSI escape sequences
* `antsy` as in fidgety ~> fast

New crate, new docs.


indietyp:
After looking into it, I completely forgot the setup and just went ahead with my first batch of changes, so this is now a two piece PR:

* setup (.github changes etc)
* initial documentation and outline of most outside-facing functions.

I can split this into 2 separate PRs if that helps.

libs/antsi/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+`antsi` is a no-std mini-crate that provides support for SGR[^1] codes (better known as ANSI escape sequences)
+
+The name is a play on words and encapsulates the three goals of the crate:
+
+- ant: small, productive and extremely useful (🐜)
+- ansi: implementation of ANSI escape sequences
+- antsy: restless as in fast, with near to no overhead (🏎️💨)
+
+The crate tries to be as correct as possible, with the goal of being not only a crate, but also a comprehensive reference about terminal support and related specifications.
```
nonparibus, line 13: While I think this is a great goal for the crate, I suggest we seek to "undersell and over-deliver" here, and avoid:

- claiming comprehensiveness
- committing ourselves to ongoing maintenance, making clear that correctness is claimed only at the point of publication (while we likely want to do this, we should gauge the initial reaction to `antsi` first)

```suggestion
The crate tries to be as correct as possible, acting both as a library and as an up-to-date reference guide regarding terminal support and related specifications (correct as of the time of publication).
```

libs/antsi/README.md
```diff
@@ -0,0 +1,48 @@
...
+        style = style.font_mut().set_strikethrough();
+    }
+
+    eprintln!("O no! {}", PANIC_STYLE.apply(format_args!("mainframe breach {} has been {}", style.apply("(from an unknown user)"), BOLD_STYLE.apply("detected"))))
+}
+```
+
+## Contributors
+
+`antsi` was created by [Bilal Mahmoud](https://github.com/indietyp). It is being developed in conjunction with [HASH](https://hash.dev/). As an open-source project, we gratefully accept external contributions and have published a [contributing guide](https://github.com/hashintel/hash/blob/main/CONTRIBUTING.md) that outlines the process. If you have questions, please reach out to us on our [Discord server](https://hash.ai/discord). You can also report bugs [directly on the GitHub repo](https://github.com/hashintel/hash/issues/new?assignees=Alfred-Mountfield%2CTimDiekmann%2Cindietyp&labels=A-antsi%2CC-bug&template=bug-report-antsi.yml).
```
nonparibus, line 42:
```suggestion
`antsi` was created by [Bilal Mahmoud](https://github.com/indietyp) for use in [`error-stack`](https://github.com/hashintel/hash/tree/main/libs/error-stack). It is being developed in conjunction with [HASH](https://hash.dev/) as an open-source project. We gratefully accept external contributions and have published a [contributing guide](https://github.com/hashintel/hash/blob/main/CONTRIBUTING.md) that outlines the process. If you have questions, please reach out to us on our [Discord server](https://hash.ai/discord?utm_medium=organic&utm_source=github_readme_hash-repo_libs-antsi-readme). You can also report bugs [directly on the GitHub repo](https://github.com/hashintel/hash/issues/new?assignees=Alfred-Mountfield%2CTimDiekmann%2Cindietyp&labels=A-antsi%2CC-bug&template=bug-report-antsi.yml).
```

nonparibus, line 42: I've created an @hashintel/antsi team. Not sure if we can use this as the assignee instead here.

indietyp: 🤔 after looking into it, it seems that the URL parameter version here does not work for people that are not at least contributors. We might want to move this into the template instead

nonparibus:
We're going to block a new PR updating the repo's READMEs (https://github.com/hashintel/hash/pull/2008) on `antsi` merging.

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
+// Good reference to begin with: https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+#![no_std]
+#![cfg_attr(
+    nightly,
+    feature(provide_any, error_in_core, error_generic_member_access)
+)]
+#![cfg_attr(all(doc, nightly), feature(doc_auto_cfg))]
+#![warn(
```
TimDiekmann, line 8: Could you copy over the `.config/config.toml` from `apps/hash-graph` instead?

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
+// Good reference to begin with: https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+#![no_std]
+#![cfg_attr(
+    nightly,
+    feature(provide_any, error_in_core, error_generic_member_access)
```
TimDiekmann, line 5: For what do we need these?

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+    clippy::print_stdout,
+    clippy::print_stderr,
+    clippy::alloc_instead_of_core,
+    clippy::std_instead_of_alloc,
+    clippy::std_instead_of_core,
+    clippy::if_then_some_else_none
+)]
+#![allow(clippy::redundant_pub_crate)] // This would otherwise clash with `unreachable_pub`
+#![allow(clippy::module_name_repetitions)]
+#![allow(missing_docs)]
```
TimDiekmann, line 24: I'd say we remove this and put it in appropriate places (or just add the docs directly) instead.

TimDiekmann (changes requested):

libs/antsi/src/lib.rs
indietyp: deer and error-stack both use this, to stay consistent I chose this approach instead. We might want to change that for all of them/

libs/antsi/src/lib.rs
indietyp: we don't, will remove.

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+///
+/// ## History
+///
+/// Nowadays every terminal emulator supports these colors. To programmatically check for
+/// support it is advised to use a crate similar to [`supports-color`](https://lib.rs/crates/supports-color)
+///
+/// ## Support
+///
+/// Support for bright colors was not part of [ISO 6429](https://www.iso.org/standard/12782.html)
+/// the specification, which introduced and standardized ANSI escape sequences.
```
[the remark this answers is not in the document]
indietyp: I don't think we need to, due to the complexities associated we would end up with *a lot* of feature flags.

* `bright` (aixterm)
* `rgb` (ISO 8613-6)
* `rgb-compat` (`xterm`)
* `underline` (kitty + VTE)
* `underline-color` (kitty + VTE)
* `rgba` (wezterm)

ANSI defines that unrecognized escape sequences are just ignored, so even if they are used in a terminal that does not support them the user will just have no change in appearance.

libs/antsi/src/lib.rs
TimDiekmann: Yes, the plan is to move it to the repository root. For now, it's only in `apps/hash-graph` as this is the most important crate we have currently. We could add it to `/libs` in another PR for now and block this PR on that

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            weight: None,
+            family: None,
+            style: 0x00,
+            underline: None,
+            blinking: None,
+        }
+    }
```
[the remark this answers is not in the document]
indietyp: Yes, it's the [builder-lite](https://matklad.github.io/2022/05/29/builder-lite.html)-mut-fusion-pattern (patent pending™). I will add getter methods.

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: If unknown sequences are ignored this seems good enough. @hashintel/antsi thoughts?

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+        self
+    }
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct Foreground(Color);
+
+impl Foreground {
+    #[must_use]
+    pub const fn new(color: Color) -> Self {
```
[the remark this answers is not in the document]
indietyp: That was my initial thought, the only problem is that const trait implementations are not `const`, so we would need two functions.

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+
+impl Style {
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            font: Font::new(),
+            foreground: None,
+            background: None,
+        }
+    }
```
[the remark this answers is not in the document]
indietyp: fixed in [4cf4e0a](https://github.com/hashintel/hash/pull/1995/commits/4cf4e0a89d1d72308e8cbd094731bbf04c0d46dc)

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
indietyp: fixed in [4cf4e0a](https://github.com/hashintel/hash/pull/1995/commits/4cf4e0a89d1d72308e8cbd094731bbf04c0d46dc)

libs/antsi/src/lib.rs
indietyp: fixed in [4cf4e0a](https://github.com/hashintel/hash/pull/1995/commits/4cf4e0a89d1d72308e8cbd094731bbf04c0d46dc)

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: [It seems to work just as expected](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=ce358256b62b3d6d85d29712ed151ac0)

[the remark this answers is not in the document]
indietyp: Yes, but what I meant is that it wouldn't work on stable.

libs/antsi/src/lib.rs
```diff
@@ -1,14 +1,500 @@
...
+}
+
+impl RgbColor {
+    #[must_use]
+    pub const fn new(r: u8, g: u8, b: u8) -> Self {
+        Self { r, g, b }
+    }
+}
+
+// TODO: Default
```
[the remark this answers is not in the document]
indietyp: Wrongly worded `TODO`. This was a (now removed) placeholder to consider also allowing to reset to the default background/foreground color, but I determined this is unnecessary, as (if we color) we reset to the default anyway once we're done.

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: Would this be a problem? I don't feel strongly, but adding a `const` is a non-breaking change. Changing the parameter to `impl Into`, however, is.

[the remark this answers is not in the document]
indietyp: 🤔 we could in theory have two functions:

* `new(impl ~const Into<Color>) -> Self` (`const` on nightly)
* `const new_color(Color) -> Self` (deprecated once const traits are stable)

libs/antsi/src/color.rs
```diff
@@ -0,0 +1,313 @@
...
+impl Color {
+    /// Return [`BasicColor`], if value of variant [`Color::Basic`]
+    #[must_use]
+    pub const fn as_basic(&self) -> Option<BasicColor> {
+        if let Self::Basic(color) = self {
+            Some(*color)
+        } else {
+            None
+        }
+    }
```
TimDiekmann, line 258: As `match` statements are allowed in `const` context, do we really want these methods?

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: We could also just implement `From` and deprecate `new` later

libs/antsi/src/font.rs
```diff
@@ -0,0 +1,254 @@
...
+/// The specification supports 9 alternative fonts.
+///
+/// [ISO 6429]: https://www.iso.org/standard/12782.html
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum FontFamily {
+    /// The primary (default) font family
+    Primary,
+    /// Fraktur
+    ///
+    /// What is Fraktur? [ISO 6429] was first introduced in 1976 as [ECMA-48], a time where in
```
TimDiekmann, line 71: Needs reformatting to adjust line length:
```suggestion
    /// ## What is Fraktur?
    ///
    /// [ISO 6429] was first introduced in 1976 as [ECMA-48], a time where in
```

libs/antsi/src/lib.rs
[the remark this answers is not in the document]
indietyp: Yes, but one of the goals of `antsi` is the easy creation of styles (and reusing them) in const contexts, therefore we need some sort of constructor that allows `const` in stable environments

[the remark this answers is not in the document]
TimDiekmann: Sorry, I think my comment was misleading.
We could implement `From` for the different color variants (or even `where T: Into<Color>`?) and provide `const fn new` until const traits are stabilized.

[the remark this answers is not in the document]
indietyp: fixed in [5c969e0](https://github.com/hashintel/hash/pull/1995/commits/5c969e06494eec25ba0fa44727f2bd08ff08cab2)

libs/antsi/src/lib.rs
indietyp: fixed in [33fbd97](https://github.com/hashintel/hash/pull/1995/commits/33fbd97888e967ed368cbdfc9487de030a6be70e)

libs/antsi/src/color.rs
indietyp: sure, tho I added them as convenience getters (to e.g. allow easy chaining), if you think that is not valuable I'll remove them :D

TimDiekmann: I think the signatures are misleading. From reading they suggest, that they may get converted into `BasicColor` and return `None` if there is no representation for that.

TimDiekmann (approved):

TimDiekmann (approved):

PR #2001, opened 2023-02-01
Add HASH plugin for `IntelliJ`-based IDEs

indietyp:

This PR introduces a new experimental IDE plugin, which aims to make some things specific to HASH easier. The first of those is: 

* `module.rs` fix

### `module.rs` fix

`module` and `module.rs` files are now correctly grouped. With git colors enabled and the native theme, it looks like this:

image

Suggestions are welcome!

### 👂 Open Questions

* Name of the plugin
* Should the plugin be available on the JetBrains marketplace?
* Is this wanted?
* Where should it be placed in the monorepo?


indietyp:
Minor status update: This is mostly working (I am using it actively). There's currently just a UI bug remaining that I'd like to solve. (Java mutability rules ftw)

image

nonparibus:
I imagine this would live at `apps/intellij-plugin` but am open to alternative suggestions.

We should add a`README` making clear that this is experimental, similar to that as we've done with `deer`.

We might want to consider instrumenting this with very basic privacy-preserving telemetry to give us some rough data on usage (to help inform whether this is something worth investing in or maintaining), if that's congruent with the JetBrains plugin developer terms of service. Ahmad is currently working on our telemetry pipeline with the Block Protocol WordPress plugin in mind, so I suggest we consider mirroring the approach we end up adopting for that (when settled).

We lack existing Java/Kotlin code and there is nothing in the repo set up to handle this. If we need to update our `.ignore` file patterns, CI, etc, it would be good to understand this upfront and consider the impact as part of this PR.

It is likely we will lack the capacity to effectively review this PR in the near-term, so we should be comfortable approving a version of it without subjecting the PR to our usual level of intense review. In this case, a security review should still be conducted, and a disclaimer should be added to the plugin README advising of its status.

indietyp:
> We should add a README making clear that this is experimental [...]

I will add a README, with an explanation and clear statement that this is to be considered experimental with an additional security disclaimer.

> We lack existing Java/Kotlin code and there is nothing in the repo set up to handle it specifically (yet). If we need to update our .ignore file patterns, CI, etc, it would be good to understand this upfront and consider the impact as part of this PR.

I'd say that for now, as this is highly experimental, configuring CI/CD is not necessary.

The project itself is completely self-contained. Therefore no update to `.ignore` should be needed. `.prettierignore` already ignores the build directory (https://github.com/hashintel/hash/blob/8ae3629aaddb16158a36d20152813eba3f572478/.prettierignore#L100-L103) as does markdown-lint (https://github.com/hashintel/hash/blob/8ae3629aaddb16158a36d20152813eba3f572478/.markdownlintignore#L87-L90).

> I imagine this would live at apps/intellij-plugin, or intellij-* if we come up with another name. Alternative suggestions are, however, welcome.

I will move the project to `apps/` and rename it. If we want to publish the plugin, a name (and icon) might be good. The prototype's name is `rust-extra`, which isn't great.

> [...] if that's congruent with the JetBrains plugin developer terms of service.

I haven't yet taken a look at the terms of service just yet (not that I have the ability to properly judge the legal text), but popular plugins like [material theme do integrate analytics](https://material-theme.com/docs/analytics/)

nonparibus (approved):
Legal approval ✅

apps/intellij-plugin/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
+import org.gradle.api.JavaVersion.VERSION_17
+
+plugins {
+    id("java")
+    id("org.jetbrains.kotlin.jvm") version "1.7.20"
+    id("org.jetbrains.intellij") version "1.12.0"
+}
+
+group = "ai.hash"
+version = "1.0-SNAPSHOT"
```
TimDiekmann, line 10: Where this version is used? Shouldn't it be something like `0.0.1`?

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        // To merge all `module` with their corresponding `module.rs` file we need to do two iterations
+        // ... first collect all directories and files via lookup tables
+        for (child in children) {
+            val value = child.value
+
+            if (value is PsiDirectory) {
+                // Statement above checks that this is the case
+                @Suppress("UNCHECKED_CAST")
+                directories[value.name] = child as AbstractTreeNode<PsiDirectory>
```
TimDiekmann, line 29: I don't know much about java, but can't the check from above be part of the cast?

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            // all rust files need to end with `.rs`, therefore it is safe to just look for `*.rs` files
+            // this way we do not need to require an implicit dependency on the `Rust` plugin to provide us
+            // with the correct `value.virtualFile.fileType` value
```
TimDiekmann, line 34: Would depending on Rust have other benefits?

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
```diff
@@ -0,0 +1,77 @@
...
+        children: MutableCollection<AbstractTreeNode<*>>,
+        settings: ViewSettings?,
+    ): MutableCollection<AbstractTreeNode<*>> {
+        val nodes = ArrayList<AbstractTreeNode<*>>()
+
+        val directories = HashMap<String, AbstractTreeNode<PsiDirectory>>()
+        val files = HashMap<String, AbstractTreeNode<PsiFile>>()
+
+        // To merge all `module` with their corresponding `module.rs` file we need to do two iterations
+        // ... first collect all directories and files via lookup tables
```
TimDiekmann, line 22:
```suggestion
        // To merge all `module` with their corresponding `module.rs` file we need to do two iterations.
        // First collect all directories and files via lookup tables...
```

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
```diff
@@ -0,0 +1,77 @@
...
+            // as a direct dependency, this is bad, because that'd mean we would need to update our plugin
+            // everytime the rust plugin is updated, with multiple versions etc.
+            if (value is PsiFile && value.virtualFile.extension == "rs") {
+                // Statement above checks that this is the case
+                @Suppress("UNCHECKED_CAST")
+                files[value.virtualFile.nameWithoutExtension] = child as AbstractTreeNode<PsiFile>
+            }
+        }
+
+        // ... then only retain directories that have corresponding files
```
TimDiekmann, line 45:
```suggestion
        // ... then only retain directories that have corresponding files.
```

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleNodeToolWindowListener.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.wm.ex.ToolWindowManagerListener
+
+class ModuleNodeToolWindowListener(project: Project) : ToolWindowManagerListener {
+    init {
+        ProjectView.getInstance(project).currentProjectViewPane.tree.addTreeExpansionListener(
+            ModuleNodeTreeExpansionListener()
+        );
+    }
+}
```
TimDiekmann, line 13:
```suggestion
}

```

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleNodeTreeExpansionListener.kt
```diff
@@ -0,0 +1,32 @@
...
+                        NavigationUtil.openFileWithPsiElement(navi.value, true, true)
+                    }
+                }
+            }
+        }
+    }
+
+    override fun treeCollapsed(event: TreeExpansionEvent?) {
+    }
+}
```
TimDiekmann, line 32:
```suggestion
}

```

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <!-- Extension points defined by the plugin.
+         Read more: https://plugins.jetbrains.com/docs/intellij/plugin-extension-points.html -->
+    <extensions defaultExtensionNs="com.intellij">
+        <treeStructureProvider implementation="ai.hash.rustextra.projectView.ModuleGroupTreeStructureProvider" order="last, after org.rust.lang"/>
+    </extensions>
+    <projectListeners>
+        <listener class="ai.hash.rustextra.projectView.ModuleNodeToolWindowListener" topic="com.intellij.openapi.wm.ex.ToolWindowManagerListener" />
+    </projectListeners>
+</idea-plugin>
```
TimDiekmann, line 33:
```suggestion
</idea-plugin>

```

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,33 @@
+<!-- Plugin Configuration File. Read more: https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html -->
+<idea-plugin>
+    <!-- Unique identifier of the plugin. It should be FQN. It cannot be changed between the plugin versions. -->
+    <id>ai.hash.rust-extra</id>
+
+    <!-- Public plugin name should be written in Title Case.
+         Guidelines: https://plugins.jetbrains.com/docs/marketplace/plugin-overview-page.html#plugin-name -->
+    <name>Rust-extra</name>
```
TimDiekmann, line 8: This is not Title Case 😉

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <!-- A displayed Vendor name or Organization ID displayed on the Plugins Page. -->
+    <vendor email="support@yourcompany.com" url="https://www.yourcompany.com">YourCompany</vendor>
```
TimDiekmann, line 11: @nonparibus what email we want to put here?

```suggestion
    <vendor email="support@hash.ai" url="https://hash.ai">HASH</vendor>
```

TimDiekmann (changes requested):

apps/intellij-plugin/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Disclaimer
+
+This plugin is to be considered **highly experimental**, HASH does not guarantee any long-term support. This project might vanish one day if it wasn't deemed feasible.
+
+**Security Disclaimer:** Due to the highly experimental status, the code is currently not as thoroughly tested and reviewed as other projects in this repository.
+
+## Running the plugin
+
+To install and run the plugin it is recommended to have a JetBrains IDE installed, navigate to this folder and execute `./gradlew buildPlugin` to build the plugin (can be found at `/build/distributions`) and `./gradlew runIde` to test run the plugin in a sandboxed CLion installation.
```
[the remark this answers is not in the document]
indietyp: Yes, this setting sets the target IDE to CLion. Therefore will always use CLion for `runIde`. To install the plugin on multiple platforms, we need to build the plugin multiple times (at least twice). I opted out of this process and set it to CLion.

https://github.com/hashintel/hash/blob/8b894afc52aeb92ee69e45f8b553a0b3e8c849a6/apps/intellij-plugin/build.gradle.kts#L20

apps/intellij-plugin/build.gradle.kts
indietyp: on the JetBrains Marketplace, I can switch it to `0.0.1`, `1.0-SNAPSHOT` was simply the default from the template I generated the plugin from.

TimDiekmann: Yes, let's switch it 🙂 
```suggestion
version = "0.0.1"
```

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
indietyp: This is a limitation of Kotlin. I tried to fuzz around with this, but sadly to no avail (this is also why I split `value` out of `child` here).

Because fearless concurrency is not a concept in Java, it does not create a smart-cast. (https://kotlinlang.org/docs/typecasts.html#smart-casts)

> var properties - never, because the variable can be modified at any time by other code.

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleGroupTreeStructureProvider.kt
indietyp: For now, no, I was initially dependent on Rust, but plugins (and Java, for that fact) are unlike Rust dependencies. You need to pin a specific version. This makes it super complicated because the Rust plugin actually publishes **2-3** distinct versions per release. (You can see it here: https://plugins.jetbrains.com/plugin/8182-rust/versions/stable). This would also mean that our plugin would need to update every time their plugin updates, which is a headache (otherwise our plugin would get disabled). We can indirectly influence the plugin using JetBrain plugin internals, but should the occasion arise for a specific feature that needs a direct dependency I am very open to it! :D

TimDiekmann: Thanks for the clarification!
Also, I wonder if we want to add the same behavior to TS files at some point. Depending on the Rust plugin would be suboptimal I think.

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
indietyp:
```suggestion
    <name>Rust-Extra</name>
```

TimDiekmann (approved):

nonparibus (approved):

PR #2004, opened 2023-02-02
Update database to allow external metadata

TimDiekmann:

In order to have external type support we need to store some additional metadata in the database

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203877122765672/f) _(internal)_

## 🔍 What does this change?

- Change database to allow external metadata
- Adjust error conditions to properly catch attempts to update wrong ontology types (updating an external ontology type with owned metadata and vice versa)
- #2009

## 🛡 What tests cover this?

The code is not automatically tested. @thehabbos007 and I spend quite some time testing various conditions to ensure correctness.


thehabbos007 (approved):
Looks good to me, tested it locally on the app and types seem to function as expected :+1: 

Let's merge this when we get a chance to wipe the migrations

thehabbos007 (approved):
Changes LGTM, tested the app locally with no observable changes

PR #2009, opened 2023-02-03
Raise a better error if entity update failed

TimDiekmann:

Similar to #2004, this also adds a better error message when an entity update failed

## 🔍 What does this change?

raises an exception, if the updated did not happen

## ⚠️ Known issues

When I initially implemented that logic, I had a pretty complicated (manual) test setup, which I haven't reproduced it. Mainly, I was testing two concurrent updates with different parameters. The logic here didi not change, only how we report the error, but we should investigate further when exactly an update can fail. Currently, we're checking from Rust, if the entity exist.


TimDiekmann:
Offline, we decided to merge this PR into #2004 to avoid touching the migration files from two different PRs, so I pushed the changes there directly

PR #2014, opened 2023-02-06
Add `Address` block

luisbettencourt:

This PR implements the Address block, powered by Mapbox.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203624215426857/f) _(internal)_

## ⚠️ Known issues

When the user closes the selected address the link entities to the map image and address entity are deleted. In Mock block dock this is not correctly working as the `linkedEntities` are still being returned. This leads to a slow down when selecting multiple addresses as the `linkedEntities` array will grow in size.

## 🐾 Next steps

- [x] Add mapbox watermark to map
- [x] Zoom levels

Follow-up PRs:
- Route Mapbox API calls via Þ middleware;
- Add interactive map option for paid subscribers.

## ❓ How to test this?

1.  Run `yarn dev` inside `blocks/address`;
2. Get a mapbox accessToken or ask for mine in slack;
3. Test the block.

## 📹 Demo
demo

image

demo2


blocks/address/src/app.tsx
```diff
@@ -0,0 +1,514 @@
...
+
+type BlockEntityProperties = {
+  label: string;
+  fullAddress: string;
+  mapUrl: string;
+  description: string;
+};
+
+const accessToken =
+  "pk.eyJ1IjoibHVpc2JldHRlbmNvdXJ0IiwiYSI6ImNsZGl2ZXdvODBuY2YzcW1lb3N5bng4NTQifQ.HW1cG865jlptDTbJBNwhQw";
```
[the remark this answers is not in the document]
luisbettencourt: Pushed it by mistake, already refreshed the token

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,514 @@
...
+const regionKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/addressregion/";
+const postalCodeKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/postalcode/";
+const streetKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/streetaddress/";
+
+const popperPositionDataAttribute = "data-popper-placement";
+
+const popperPlacementSelectors = {
```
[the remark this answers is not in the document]
luisbettencourt: Refreshed the token

nonparibus (changes requested):
Please can you add explicit `LICENSE`, `LICENSE-MIT`, and `LICENSE-APACHE` files as found in our other blocks?

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,605 @@
...
+const streetKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/streetaddress/";
+const countryKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/addresscountry/";
+const fullAddressKey =
+  "https://alpha.hash.ai/@luisbett/types/property-type/fulladdress/";
+
+// const accessToken = "";
+const accessToken =
+  "pk.eyJ1IjoibHVpc2JldHRlbmNvdXJ0IiwiYSI6ImNsZHphbG9pdzB3cWIzb29hNzFxbjNtMjAifQ.IZNEdPk4pFvPoY3npFbAtQ";
```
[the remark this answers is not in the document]
luisbettencourt: Access token refreshed and removed from code

nonparibus:
I followed up directly w/ Luís regarding a couple of changes (copy button on address requested by Joel, and icon/text that currently reads `Close x`).

nonparibus:
Do we want the `CHANGELOG` file in here? Assuming it's a stray remnant of the template.

blocks/address/src/editable-field.tsx
```diff
@@ -0,0 +1,103 @@
...
+          ...(Array.isArray(sx) ? sx : [sx]),
+        }}
+      />
+
+      {!readonly ? (
+        <Fade in={hovered && !editing}>
+          <Box sx={{ position: "relative" }}>
+            <IconButton
+              onClick={() => {
+                setEditing(!editing);
```
CiaranMn, line 84: Shouldn't this be `true`? This is to enable editing, not to toggle it.

if it's a toggle it should only focus the input if editing is being set to `true`

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,677 @@
...
+
+    const addressEntityId =
+      createAddressEntityResponse?.data?.metadata.recordId.entityId;
+
+    if (addressEntityId) {
+      if (!addressLinkEntity) {
+        await graphService?.createEntity({
+          data: {
+            entityTypeId:
+              "https://alpha.hash.ai/@luisbett/types/entity-type/address-link/v/1",
```
CiaranMn, line 331: This can be `addressLinkTypeId`

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,677 @@
...
+
+  return (
+    <ThemeProvider theme={theme}>
+      <Box
+        ref={blockRootRef}
+        sx={{ display: "inline-block" }}
+        onMouseEnter={() => setHovered(true)}
+        onMouseLeave={() => setHovered(false)}
+      >
+        <Fade
```
CiaranMn, line 390: This shouldn't appear in `readonly` mode

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,677 @@
...
+              >
+                <MapboxIcon sx={{ fontSize: 16, mr: 0.375 }} />
+                Mapbox Static Images
+              </Box>
+              to render a fixed map
+            </Typography>
+          </Box>
+        </Fade>
+
+        <Collapse
```
CiaranMn, line 472: This probably shouldn't appear in `readonly` mode

blocks/address/src/address-card.tsx
```diff
@@ -0,0 +1,340 @@
...
+            <Fade in={hovered}>
+              <Stack
+                sx={{
+                  display: "inline-flex",
+                  position: "absolute",
+                  top: 13,
+                  left: 13,
+                }}
+              >
+                <Button
```
CiaranMn, line 240: This should not appear in `readonly`

blocks/address/src/address-card.tsx
```diff
@@ -0,0 +1,340 @@
...
+              >
+                OpenStreetMap
+              </Link>
+            </Typography>
+          </Box>
+        ) : (
+          <CircularProgress sx={{ color: ({ palette }) => palette.gray[40] }} />
+        )}
+
+        <Fade in={hovered}>
```
CiaranMn, line 305: This should not appear in `readonly`

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,677 @@
...
+                    color: ({ palette }) => palette.black,
+                  }}
+                >
+                  <Box
+                    component="span"
+                    sx={{ color: ({ palette }) => palette.red[60] }}
+                  >
+                    Error connecting
+                  </Box>{" "}
+                  to the Block Protocol
```
CiaranMn, line 616: this may be implemented in other ways than connecting to blockprotocol.org
```suggestion
                  to the Mapbox API
```

blocks/address/src/app.tsx
```diff
@@ -0,0 +1,677 @@
...
+                  Check your network connection or{" "}
+                  <Box
+                    component="span"
+                    sx={{
+                      fontWeight: 700,
+                      color: ({ palette }) => palette.gray[70],
+                    }}
+                  >
+                    contact support
+                  </Box>{" "}
```
CiaranMn, line 636: Since this is not a link, I would not bold it – it looks like it should be clickable

CiaranMn (changes requested):
Thanks @luisbettencourt – this looks great!

I have a few functionality changes requested, mainly around readonly mode.

Also to note that the error text appears when Mapbox does not recognise the address (e.g. as you start typing) - I suggest we come back to better error handling when the external service module is available.

CiaranMn:
Forgot to mention – can we add mobile styling? It's almost there but overflows horizontally

blocks/address/README.md
```diff
@@ -0,0 +1,77 @@
+# Block template: React
```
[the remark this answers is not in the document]
luisbettencourt: Done in 4cb2a032698a105cf2e06ad533bb84ebec097cd6

blocks/address/src/editable-field.tsx
luisbettencourt: This button was a toggle in its first iteration, forgot to change it. Addressed in 39eeca67e607adb15706c7258e6fa85d5b6c6277

blocks/address/src/app.tsx
luisbettencourt: Thank you. 2af46146fb4f66ce7c7a30a794edd03721c029e5

blocks/address/src/app.tsx
luisbettencourt: Hid it in readonly mode in 38c4a19ea9d64daea84706f708acccffd3e7ebc9

blocks/address/src/address-card.tsx
luisbettencourt: Hid it in readonly mode in 38c4a19ea9d64daea84706f708acccffd3e7ebc9

blocks/address/src/address-card.tsx
luisbettencourt: Hid it in readonly mode in 38c4a19ea9d64daea84706f708acccffd3e7ebc9

blocks/address/src/app.tsx
luisbettencourt: We should probably have an empty state for when the block is in readonly mode and without an address selected?

blocks/address/src/app.tsx
luisbettencourt: I'm not sure if it's supposed to be a link. Here's the [design](https://www.figma.com/file/T10d2P2iNb1LdRC5a9J7vA/Blocks?node-id=1403%3A5185&t=qzUaNLUATCz9wAjo-4).

@nonparibus

CiaranMn: It was originally intended to but we don't have a way of EAs telling blocks what 'contact support' should do at the moment. We can make it a link later, but while it isn't I would not bold it as it implies interactivity.

luisbettencourt: Addressed in 1f10fc11ee423380e7e0900acbec6876523ddb76

blocks/address/src/app.tsx
CiaranMn: I think potentially we should have nothing displayed. 

At least for sites like WordPress, 'readonly' is the state you want to show when displaying a page for information only, e.g. in this case simply the map and address. If there isn't a map and address to show, arguably you are better off showing nothing than some error-looking message about a map not being selected that the viewer can do nothing about.

I suggest we think about this further in follow-up since we need to do a v2 of the block next week anyway.

blocks/address/src/address-card.tsx
CiaranMn: We can do that, although it would trigger more API requests than pageviews. I would default to no, but it could be added later as a configuration option for the block.

CiaranMn (approved):

CiaranMn (approved):

PR #2017, opened 2023-02-07
Replace Graph initialization stop-gap solution to use the type fetcher

TimDiekmann:

In order to properly implement the type fetcher, it's required that it's set up properly in external services.
This PR adds the `hash-type-fetcher` binary to the graph Docker image and runs it from the external services. To make sure the deployment is working, it's used to replace the current stop gap solution to insert primitive datatypes but instead it fetches it from blockprotocol.org.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203882684071078/f) _(internal)_

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2038

## 🔍 What does this change?

- ~~Add type fetcher binary to Dockerfile~~ https://github.com/hashintel/hash/pull/2038
- ~~Use rust-tls instead of OpenSSL to avoid the dependency on glibc in the Dockerfile (The error messages on alpine on link errors are not helpful at all ("file does not exist"), by luck I've seen the OpenSSL dependency and removed it)~~ https://github.com/hashintel/hash/pull/2038
- ~~Run the type fetcher from external services~~ https://github.com/hashintel/hash/pull/2038
- Implement `FetcherPool`, which is a wrapper over another `Pool`, but also holds information for the type fetcher
- Provide `FetcherStore`, but don't implement `Store` on it yet as this is better suited as a follow-up after some clean-up of the `Store`, which would otherwise result in a huge diff. `FetcherPool<P>` currently acquires `P::Store` instead of `FetcherStore<P::Store>` but added a commented-out section on how this will be done in a follow-up.
- To test this functionality, I replaced the stop-cap-solution we currently use by a function, which uses the type fetcher instead.

In order to run the graph we still use `cargo make deployment-up`, which now also covers the type fetcher.

## ⚠️ Known issues

Currently, `Store` requires _all_ different stores to be implemented on that type and `Pool` requires the associated type to implement `Store`. This is a very unfortunate situation and requires some refractoring to sort out. Also I have some issues with `#[async_trait]` as it requires `Send` and cannot capture the inner lifetimes implicitly. When removing `#[async_trait]` or the `Send` bound (`#[async_trait(?Send)]`), `axum` is not able anymore to use methods from those traits. There is a way to work around this, but this will uglify the `Store` trait even more. I will tackle this in a follow-up PR.

## 🐾 Next steps

Dynamically use the type fetcher when creating or updating an ontology type

## 🛡 What tests cover this?

The `type-fetcher` is used to fetch built-in data types from blockprotocol.org. Any test, which uses those built-in types, will also cover this implementation. 

## ❓ How to test this?

Spin up the external services and check if everything works as before

### DIFF
--- a/apps/hash-graph/Cargo.lock
+++ b/apps/hash-graph/Cargo.lock
@@ -750,9 +750,11 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
+ "tarpc",
  "time",
  "tokio",
  "tokio-postgres",
+ "tokio-serde",
  "tonic",
  "tower",
  "tower-http",
@@ -2376,9 +2378,9 @@ dependencies = [
 
 [[package]]
 name = "tokio-util"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bb2e075f03b3d66d8d8785356224ba688d2906a371015e225beeb65ca92c740"
+checksum = "bc6a3b08b64e6dfad376fa2432c7b1f01522e37a623c3050bc95db2d3ff21583"
 dependencies = [
  "bytes",
  "futures-core",
--- a/apps/hash-graph/bin/cli/src/subcommand/server.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/server.rs
@@ -19,7 +19,6 @@ use graph::{
 };
 use regex::Regex;
 use reqwest::Client;
-use serde_json::json;
 #[cfg(feature = "type-fetcher")]
 use tarpc::context;
 use tokio::time::timeout;


apps/hash-external-services/docker-compose.yml
```diff
@@ -117,6 +117,28 @@ services:
...
+      dockerfile: docker/Dockerfile
+      context: ../../apps/hash-graph
+      args:
+        PROFILE: dev
+    environment:
+      HASH_GRAPH_LOG_FORMAT: "${HASH_GRAPH_LOG_FORMAT:-pretty}"
+      RUST_LOG: "${HASH_GRAPH_LOG_LEVEL:-info,hash_type_fetcher=trace,reqwest=debug}"
+      RUST_BACKTRACE: 1
+    entrypoint: "/hash-type-fetcher"
+    command: ["--type-fetcher-host", "0.0.0.0", "--type-fetcher-port", "4444"]
```
thehabbos007, line 132: Any reason for these to be args rather than environment variables? I suppose we're overwriting the existing command, but that can also be empty I assume

apps/hash-external-services/docker-compose.yml
```diff
@@ -117,6 +117,28 @@ services:
...
+    healthcheck:
+      test:
+        [
+          "CMD-SHELL",
+          "netstat -an | grep 4444 > /dev/null; if [ 0 != $? ]; then exit 1; fi",
+        ]
```
thehabbos007, line 138: I think we have at 3 types of TCP health checks in the codebase :D

I don't think it's worth it for us to try to simplify it, I just find it funny we have so many slightly different setups that require very different health checks

apps/hash-graph/bin/cli/src/subcommands/server.rs
```diff
@@ -239,29 +275,97 @@ async fn stop_gap_setup(pool: &PostgresStorePool<NoTls>) -> Result<(), GraphErro
...
-            ProvenanceMetadata::new(UpdatedById::new(root_account_id)),
-            OwnedById::new(root_account_id),
-        ));
+    Ok(())
+}
 
-    if let Err(error) = connection
-        .create_entity_type(link_entity_type, &link_entity_type_metadata)
+#[cfg(feature = "type-fetcher")]
+async fn stop_gap_setup_type_fetcher<A: tokio::net::ToSocketAddrs + Send + Sync + Clone>(
```
thehabbos007, line 284: Very cool that we can get rid of most of the old stop-gap setup! We have to keep in mind the implications of this for when we do firewall in the hosted instance as we'd need arbitrary HTTPS egress, the IP of blockprotocol.org could be anything since we use Cloudflare

apps/hash-graph/bin/cli/Cargo.toml
```diff
@@ -14,6 +14,7 @@ axum = "0.6.3"
 clap = { version = "4.1.4", features = ["cargo", "derive", "env", "wrap_help"] }
 clap_complete = "4.1.1"
 error-stack = { version = "0.3.0", features = ["spantrace"] }
+futures = "0.3.25"
```
thehabbos007, line 17: Do we need this dependency? If this isn't optional and enabled by `type-fetcher`, I dont' think we'd need this

apps/hash-graph/docker/Dockerfile
```diff
@@ -20,14 +20,20 @@ RUN apk add --no-cache make openssl-dev protobuf-dev
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/usr/local/cargo/git \
     --mount=type=cache,target=/usr/local/src/hash-graph/target \
-    cargo install --path bin/cli --profile $PROFILE --locked
+    cargo install --path bin/cli --profile $PROFILE --features type-fetcher --locked
```
thehabbos007, line 23: Having this here would currently crash the production instance as it would spin up a server without a type-fetcher. Ideally we would have this be a built time arg in the dockerfile such that we can condtionally build this feature in until we fix the CD

apps/hash-graph/docker/Dockerfile
```diff
@@ -20,14 +20,20 @@ RUN apk add --no-cache make openssl-dev protobuf-dev
...
+    --mount=type=cache,target=/usr/local/src/hash-graph/target \
+    cargo install --path bin/type-fetcher --profile $PROFILE --locked
 
 # Replace with `FROM scratch` when a health check command was added to hash-graph
 FROM alpine:3.17 AS runtime
 
 RUN addgroup -S appgroup && adduser -S appuser -G appgroup
 
 COPY --from=builder --chown=appgroup:appuser /usr/local/cargo/bin/hash-graph /
+COPY --from=builder --chown=appgroup:appuser /usr/local/cargo/bin/hash-type-fetcher /
```
thehabbos007, line 36: Not sure what I feel about having this as a shared dockerfile, I think I like it? It's just slightly unconventional I guess. Did you have any comments on this multi-binary dockerfile setup?

apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -0,0 +1,133 @@
...
+    // type Store<'pool> = FetchingStore<P::Store<'pool>, A>;
+    //
+    // async fn acquire(&self) -> Result<Self::Store<'_>, Self::Error> {
+    //     Ok(FetchingStore {
+    //         store: self.pool.acquire().await?,
+    //         address: self.address.clone(),
+    //         config: self.config.clone(),
+    //         domain_validator: self.domain_validator.clone(),
+    //     })
+    // }
```
thehabbos007, line 78: This is the comment block you are referring to in the PR description, right?

thehabbos007 (commented):
Thank you! I have some comments, mostly concerned about the result of merging this (with the type-fetcher feature enabled by default)

TimDiekmann:
I rebased this branch on https://github.com/hashintel/hash/pull/2038, so this PR is now only replacing the stop-gap solution. As this is already implemented, we can just leave it in, for the follow-ups I mainly need the `FetcherStore`

apps/hash-external-services/docker-compose.yml
TimDiekmann: I replaced the healthchecks in https://github.com/hashintel/hash/pull/2038

apps/hash-graph/bin/cli/Cargo.toml
TimDiekmann: Made this optional in https://github.com/hashintel/hash/pull/2038

apps/hash-graph/docker/Dockerfile
TimDiekmann: We discussed offline that we will merge the binaries (done in #2037) and #2038 adds a conditional feature flag instead of hardcode it

apps/hash-graph/docker/Dockerfile
TimDiekmann: The Dockerfile now only contains one binary (`FROM scratch`!) since https://github.com/hashintel/hash/pull/2038

apps/hash-graph/lib/graph/src/store/fetcher.rs
TimDiekmann: Yes, exactly 🙂

thehabbos007 (approved):
Looks good to me. Thank you!
As we aren't touching the dockerfile, we shouldn't worry about breaking anything this time around

PR #2028, opened 2023-02-08
Make `Interval` a struct

TimDiekmann:

With upcoming tasks we require more interval with different bound types. Implementing `Interval` on all of them is annoying and error-prone. A good example is that we currently don't validate, if `lower <= upper` (a bug no one noticed yet 🙃). By making `Interval` a struct instead, this will enable us to use it everywhere.

## 🔗 Related links

This is part of [forbidding the "unbounded" bound in temporal axes](https://app.asana.com/0/1202805690238892/1203791431510341/f) and [renaming time related types](https://app.asana.com/0/1203543021352041/1203783495017458/f), but was moved to a separated PR to avoid huge diffs.

With the linked tasks I plan to introduce a few aliases for the different intervals to avoid writing the bounds each time, this was captured as TODO.

---
  - https://app.asana.com/0/0/1203936439791005


apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
```diff
@@ -0,0 +1,263 @@
...
+    ToSchema,
+};
+
+use crate::{
+    identifier::time::{axis::TemporalTagged, Timestamp},
+    interval::IntervalBound,
+};
+
+#[derive(Derivative, Serialize, Deserialize, ToSchema)]
+#[derivative(
```
Alfred-Mountfield, line 17: Why do we need to use `derivative` for these again?

Alfred-Mountfield (approved):

apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
TimDiekmann: The `#[derive(Debug)]` (and others as well) will generate `impl<A: Debug> ...` instead of `impl<A> ...`.

PR #2033, opened 2023-02-08
Pin `thread_local` crate to version `1.1.4`

TimDiekmann:

With `thread_local` v1.1.5 `miri` reports a memory leak. This pins the version of `thread_local` until the issue is fixed.

## 🔗 Related links

- https://github.com/Amanieu/thread_local-rs/issues/45


indietyp (approved):
Hopefully we won't have to pin the dependency for too long D:

PR #2035, opened 2023-02-08
Extract type editor into its own publishable package

CiaranMn:

Makes a separate package out of the type editor, `@hashintel/type-editor`, so that we can use it on blockprotocol.org also.

Cannot be merged due to type inconsistencies that are being fixed separately, but getting into review now because the only thing will change is the removal of some type assertions, and it would be good to keep this moving.

This involves:
1. Package scaffolding, with thanks to @kachkaev 
2. Shifting a bunch of components into `@hashintel/design-system`, which will also be published
3. Setting an interface for the type editor and moving code around to respect it. This is a starting point we can move from – I've kept it pretty close to what the definition tab was getting before (lists of property and entity types)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352041/1203443872976464/f) _(internal)_

## 🔍 What does this change?

- Introduces new `@hashintel/type-system` package, which offers the functionality of the `DefinitionTab` in the entity type editor, i.e. the properties and link tables
- Move various components into the `design-system`
- Moves 'is name allowed' validation out of the type editor (I've created a `useGenerateTypeUrisForUser` hook as we also need this in the create entity type form)
- Removes some context from the HASH frontend which is now inside the `type-editor`
  - entity types context remains in HASH but without the split between link types / normal types, which is only relevant to the type editor
  - the type editor has internal context, but only to avoid prop drilling
  - I made it receive and use options as a map of `uri:Type` (not `typeWithMetadata`) – we may want to change this to some other thing later, e.g. with metadata, or have it receive a subgraph it picks through. But using simply types is the easiest thing for now (previously we had property types, but entity types with metadata). I originally thought to use a subgraph for the entity but we'd have to construct it properly in HASH only to deconstruct it again, so it seemed pointless.
  - the type editor relies on the consumer to keep property/entity type options up to date – all refetching has been moved into HASH. At some point we'll do subscriptions to updates properly. The type editor doesn't need to know any of this, it should just get updated options as updated props.
- Upgraded `@mui/` versions – 5.11.x introduces a breaking change to `experimental_sx`, and `blockprotocol` has `5.11.x` installed. Because it is a peer dependency of the `type-editor` package, we rely on the version provided by the dependent application. 
- Copies `theme-override.d.ts` into `@hashintel/type-editor`. The type editor makes assumptions about non-standard values being present in the MUI theme. This previously was enabled by _importing_ `theme-override.d.ts` in the `design-system`, so that it was evaluated when `type-editor` imported things from the design system. But this breaks the design system once it's compiled. We should fix this at some point ([task](https://app.asana.com/0/1203179076056209/1203983715193924/f))

## ⚠️ Known issues

- The key issue blocking merging this is the fact that the subgraph types are mismatched between this repo and the `@blockprotocol/graph` package. I have to use the local `getRoots` function to make this work in HASH, but that function will not work in `blockprotocol`. We need to be able to use `getRoots` from `@blockprotocol/graph` before this can be used elsewhere.

## 🐾 Next steps

- Try and use this in blockprotocol.org once type mismatches are fixed
- In the meantime, I will be adding the necessary API endpoints to `blockprotocol` to support this

## 🛡 What tests cover this?

- Some existing

## ❓ How to test this?

1.  Checkout that the type editor still works and looks as it's supposed to

## 📹 Demo

https://user-images.githubusercontent.com/37743469/217792005-599c0ab2-e468-46c9-970c-67820a83a767.mp4

---
  - https://app.asana.com/0/0/1203975175898856
  - https://app.asana.com/0/0/1203904050345555


apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page.tsx
```diff
@@ -22,88 +26,19 @@ import {
...
 import { EntityTypeEntitiesContext } from "./[entity-type-id].page/shared/entity-type-entities-context";
-import { EntityTypeEditorForm } from "./[entity-type-id].page/shared/form-types";
 import { getEntityTypeBaseUri } from "./[entity-type-id].page/shared/get-entity-type-base-uri";
-import { LatestPropertyTypesContext } from "./[entity-type-id].page/shared/latest-property-types-context";
 import { useCurrentTab } from "./[entity-type-id].page/shared/tabs";
-import { useLatestPropertyTypesContextValue } from "./[entity-type-id].page/shared/use-latest-property-types-context-value";
 import { useEntityTypeEntitiesContextValue } from "./[entity-type-id].page/use-entity-type-entities-context-value";
 import { useEntityTypeValue } from "./[entity-type-id].page/use-entity-type-value";
 
-const getSchemaFromEditorForm = (
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page.tsx:42]: I moved this into the type editor and exported it from there

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab.tsx
```diff
@@ -1,34 +1,52 @@
...
-            {entityType.title}
-          </Box>
-        </Typography>
-        <LinkListCard />
-      </Box>
-    </Stack>
+  const propertyTypeOptions = useMemo(() => {
+    return {
+      ...possiblyIncompletePropertyTypeOptions,
+      ...entityTypeAndPropertyTypes.propertyTypes,
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab.tsx:31]: This ensures that the property types the entity type needs are available (this was previously handled in the old property type context)

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts
```diff
@@ -0,0 +1,86 @@
...
+  const refetchPropertyTypes = useFetchLatestPropertyTypes();
+
+  const wrappedCreateEntityType = useCallback<
+    OntologyFunctions["createEntityType"]
+  >(
+    (args) => {
+      return createEntityType(args).then(async (res) => {
+        await refetchEntityTypes();
+        return res;
+      }) as any; // @todo fix this when types consistent
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts:37]: This replicates what was previously happening for mutations from inside the type editgor – creating/updating, then awaiting the refetch, before moving on.

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/latest-property-types-context.tsx
```diff
@@ -0,0 +1,43 @@
+import { PropertyType } from "@blockprotocol/graph";
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/latest-property-types-context.tsx:1]: This is a move + expansion of `[entity-type-id].page/shared/latest-property-types-context.ts`

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/shared/latest-property-types-context.tsx
```diff
@@ -0,0 +1,43 @@
+import { PropertyType } from "@blockprotocol/graph";
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/shared/latest-property-types-context.tsx:1]: Above deleted file with more stuff added / return type tweaked

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/property-type-form.tsx
```diff
@@ -1,49 +0,0 @@
-import { BaseUri } from "@blockprotocol/type-system";
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/property-list-card/property-type-form.tsx:1]: This file has been removed and the 'check if name exists' code extracted, so it appears to be a deletion

apps/hash-frontend/src/blocks/use-fetch-block-subgraph.ts
```diff
@@ -106,7 +106,7 @@ export const useFetchBlockSubgraph = () => {
           ],
           vertices: {
             [placeholderEntity.metadata.recordId.entityId]: {
-              [now as EntityRevisionId]: {
+              [now]: {
```
Alfred-Mountfield, line 109: Was this ESLint that removed this or did you? I just tried it locally and I'm surprised to see that it's not required, I thought it would be (maybe because it's cast-able so it's happening implicitly..).

If it was manually removed, I think it's nicer as it was though, it's more explicit/correct.

apps/hash-frontend/src/middleware/return-types-as-json.ts
```diff
@@ -3,7 +3,7 @@ import {
   EntityType,
   PropertyType,
   VersionedUri,
-} from "@blockprotocol/type-system";
+} from "@blockprotocol/type-system/slim";
```
Alfred-Mountfield, line 6: We have a rewrite rule in the frontend, does that not cover the middleware?

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts
```diff
@@ -0,0 +1,133 @@
...
+  const refetchPropertyTypes = useFetchLatestPropertyTypes();
+
+  const wrappedCreateEntityType = useCallback<
+    OntologyFunctions["createEntityType"]
+  >(
+    (args) => {
+      return createEntityType(args as any).then(async (res) => {
+        await refetchEntityTypes();
+        return res;
+      }) as any; // @todo fix these when types consistent: additionalProperties/ownedById removed
```
Comment (Alfred-Mountfield) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts:42]: If this branch has been updated are these now redundant?

If it works, perhaps a `ts-expect-error` would be a good idea next time

Alfred-Mountfield (commented):
Got about a third of the way through, but as discussed offline, going to focus on other work and let others take the review here

nonparibus (approved):
Legal ✅

Still requires further code/functionality reviews, and response to prior review.

luisbettencourt (approved):
code changes LGTM. Tested Hash's entity type editor and seems to also be working as expected

blocks/address/theme-override.d.ts
```diff
@@ -0,0 +1,240 @@
+import { Components } from "@mui/material";
```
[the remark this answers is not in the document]
CiaranMn: There's a task to avoid duplicating this [here](https://app.asana.com/0/1203179076056209/1203983715193924/f) – there's a potential solution I didn't have the time to progress

libs/@hashintel/design-system/src/icon-arrow-right.tsx
```diff
@@ -1,7 +1,7 @@
 import { SvgIcon, SvgIconProps } from "@mui/material";
 import { FunctionComponent } from "react";
 
-export const ArrowRightIcon: FunctionComponent<SvgIconProps> = (props) => {
+export const IconArrowRight: FunctionComponent<SvgIconProps> = (props) => {
```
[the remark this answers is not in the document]
CiaranMn: Probably, although we're not going to have export subpaths in `design-system` so it will make no difference to consumers, will just be internal organization of the package – will leave it to some future dev to put them in an icon folder when there are too many to leave at the root

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/definition-tab/use-editor-ontology-functions.ts:42]: I am going to fix this in a follow up PR which will make changes to the type editor, including bringing it up to date with packages

luisbettencourt (dismissed):

nonparibus (approved):

nonparibus (approved):

luisbettencourt (approved):

PR #2037, opened 2023-02-09
Make `type-fetcher` a subcommand of the Graph CLI

TimDiekmann:

To only require a single binary, this makes the `type-fetcher` binary a subcommand of the CLI.

## 🔗 Related links

- This was discussed offline as part of the review of https://github.com/hashintel/hash/pull/2017

## 🔍 What does this change?

- Move the `type-fetcher` binary to `hash-graph`
- Drive-by: Simplify call to `connect` methods
- Drive-by: Avoid `use module::function` imports where the name of the module and the functions are equal


thehabbos007 (approved):
Thank you! Looks good to me, I tested with my testing client through `git cherry-pick -n adb5342cde5565032aa51b2ec865bd72bd35d49d` and it seems to be working just fine. Rest of the app works as expected as we didn't re-add the dockerfile flags yet :)

thehabbos007 (approved):

PR #2038, opened 2023-02-09
Adjust docker for type fetcher

TimDiekmann:

This splits up #2017 and only adds support for the type fetcher from docker.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203882684071078/f) _(internal)_

## 🔍 What does this change?

- Adds a build argument to enable the `type-fetcher` feature in the Graph
- Adds a very simple `--healthcheck` subcommand
- Adjust the docker-compose file to use the type fetcher
- Drive-by: Unify order of configuration of `hash-graph` configurations and apply `SemGrep` lints

## ❓ How to test this?

Test the external services with and without the type fetcher. Also test, if the `--healthcheck` flag correctly shows the healthy state


apps/hash-graph/docker/Dockerfile
```diff
@@ -5,22 +5,29 @@ ENV RUSTUP_HOME=/usr/local/rustup \
     PATH=:$PATH:/usr/local/cargo/bin
 
 COPY rust-toolchain.toml .
-RUN apk add --no-cache gcc musl-dev && \
+RUN apk add --no-cache gcc musl-dev bash && \
     wget -q -O- https://sh.rustup.rs | sh -s -- -y --default-toolchain none --profile minimal && \
     rustup show
 
+SHELL ["bash", "-c"]
```
TimDiekmann, line 12: This is required for the check, which is used in `builder`

apps/hash-graph/docker/Dockerfile
```diff
@@ -5,22 +5,29 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
 
 # To be removed once https://github.com/open-telemetry/opentelemetry-rust/issues/934 is sorted
-RUN apk add --no-cache make openssl-dev protobuf-dev
+RUN apk add --no-cache make protobuf-dev
 
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/usr/local/cargo/git \
     --mount=type=cache,target=/usr/local/src/hash-graph/target \
-    cargo install --path bin/cli --profile $PROFILE --locked
+    if [[ ${TYPE_FETCHER^^} == Y* || ${TYPE_FETCHER^^} == T* || $TYPE_FETCHER == 1 ]] ; then  \
```
TimDiekmann, line 26: I tried to capture most common boolean flags like `yes`, `YES`, `true`, `True`, and `1`

TimDiekmann: I haven't tested it, but even `YesPleaseFFS!` should work 😆

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -314,3 +328,20 @@ pub async fn server(args: ServerArgs) -> Result<(), GraphError> {
 
     Ok(())
 }
+
+pub async fn healthcheck(address: ApiAddress) -> Result<(), GraphError> {
+    let request_url = format!(
+        "http://{}:{}/api-doc/openapi.json",
+        address.api_host, address.api_port
+    );
```
TimDiekmann, line 336: I don't know if there is a better option, I'm not particularly happy about `http`. We could use the `router` instead, but this also requires connecting the pool.

apps/hash-graph/docker/Dockerfile
```diff
@@ -5,22 +5,29 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
+      cargo install --path bin/cli --profile $PROFILE --features type-fetcher --locked; \
+    else \
+      cargo install --path bin/cli --profile $PROFILE --locked;  \
+    fi
 
 # Replace with `FROM scratch` when a health check command was added to hash-graph
```
TimDiekmann, line 32: I was not able to run the healthchecks as `/bin/sh` is not available. Probably needs further investigations or it's simply not possible.

apps/hash-graph/docker/Dockerfile
```diff
@@ -5,36 +5,34 @@
...
-RUN install -d -m 0755 -o appuser -g appgroup /log
 
-USER appuser
+FROM scratch AS runtime
 
-HEALTHCHECK --interval=2s --timeout=1s --retries=5 \
-    CMD wget --no-verbose --tries=1 --spider http://localhost:4000/api-doc/openapi.json || exit 1
+COPY --from=builder /usr/local/cargo/bin/hash-graph /
 
 ENTRYPOINT ["/hash-graph"]
```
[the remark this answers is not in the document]
TimDiekmann: Thanks for flagging, [we added nobody](https://github.com/hashintel/hash/pull/2038/commits/33f258eb405bccb033230b40d5a647aedbe89c7f)

apps/hash-external-services/docker-compose.yml
```diff
@@ -72,11 +73,17 @@ services:
         condition: service_healthy
       telemetry-collector:
         condition: service_healthy
-    build:
+    build: &graph-build
       dockerfile: docker/Dockerfile
       context: ../../apps/hash-graph
       args:
         PROFILE: dev
+    image: hash-graph
```
thehabbos007, line 81: Do we need to specify image as well? I thought htis would conflict with `build`? This goes for the other services that have this

apps/hash-graph/bin/cli/src/subcommand/server.rs
thehabbos007: I think it's fine as we wouldn't be setting up TLS from the container but have a reverse proxy in front with TLS termination anyway. Although we may want to introduce a `mTLS` ceritficate or some sort of Origin cert at some point. This is fine for now :+1:

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -314,3 +328,20 @@ pub async fn server(args: ServerArgs) -> Result<(), GraphError> {
...
+
+    Client::new()
+        .head(&request_url)
```
thehabbos007, line 339: I suppose we do `head` because we don't really want to fetch the data but check for responsiveness

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
```diff
@@ -58,3 +80,15 @@ pub async fn type_fetcher(args: TypeFetcherArgs) -> Result<(), GraphError> {
 
     Ok(())
 }
+
+async fn healthcheck(address: TypeFetcherAddress) -> Result<(), GraphError> {
+    let transport = tarpc::serde_transport::tcp::connect(
```
thehabbos007, line 85: Does this `connect` call timeout or spin/wait indefintely until a connection is made? Their docs don't mention anything regarding this

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -27,37 +28,36 @@ use uuid::Uuid;
...
+
+    /// The address the REST client is listening at.
+    #[clap(flatten)]
+    pub api_address: ApiAddress,
+
+    /// The address for the type fetcher RPC server is listening at.
     #[cfg(feature = "type-fetcher")]
-    pub type_fetcher_port: u16,
+    #[clap(flatten)]
+    pub type_fetcher_address: TypeFetcherAddress,
```
thehabbos007, line 60: So many nice things came of consolidating the binary! Very nice

apps/hash-graph/docker/Dockerfile
thehabbos007: How come this afect the next stage of the dockerfile? Shouldn't this call be under line 15 for it to influence anything?

apps/hash-graph/docker/Dockerfile
```diff
@@ -5,36 +5,42 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
+      cargo install --path bin/cli --profile $PROFILE --features type-fetcher --locked; \
+    else \
+      cargo install --path bin/cli --profile $PROFILE --locked;  \
+    fi
 
-RUN addgroup -S appgroup && adduser -S appuser -G appgroup
+RUN mkdir -p /out/etc/ \
+ && echo 'nobody:x:65534:65534:nobody:/:' > /out/etc/passwd \
+ && echo 'nobody:x:65534:' > /out/etc/group \
+ && install -d -m 0755 -o nobody -g nobody /out/log \
```
thehabbos007, line 36: I suppose this is good to have in case a volume isn't mounted. Although `read_only: true` likely would make the binary fail

thehabbos007 (commented):
Looks good to me, just have some comments.

I don't know how we want to do this, though, as the docker image will make the deployed backend endlessly not pass health checks. Because of rolling updates the backend won't actually crash. DO you have any thoughts on what we should do regarding the order? We can just merge and fix the healthcheck after

apps/hash-graph/docker/Dockerfile
TimDiekmann: We reuse this image in the next stage, so this propagates.

apps/hash-external-services/docker-compose.yml
```diff
@@ -72,11 +73,17 @@ services:
         condition: service_healthy
       telemetry-collector:
         condition: service_healthy
-    build:
+    build: &graph-build
```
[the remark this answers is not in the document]
TimDiekmann: Yes, YAML is [full of neat features](https://quickref.me/yaml) 🚀 
`*` is called `alias` and `&` is called `anchor`

apps/hash-external-services/docker-compose.yml
TimDiekmann: This ensures that we use the exact same image for all three of the services, it's just the name of the image.

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: Yes, exactly 🙂

apps/hash-graph/docker/Dockerfile
TimDiekmann: `read_only: true` is only specified in `docker-compose`, where we also specify the volume to mount. Without a compose-environment, you can still write to the image.

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
thehabbos007: We'd have timeouts in the hosted instance as well to prevent this from passing, so we should be good. We can come back to this if we see problems with indefinite waiting anywhere and add a timeout

thehabbos007 (approved):

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
thehabbos007: I approved just now, but we decided we may as well add the timeout now

thehabbos007 (approved):

thehabbos007 (approved):

PR #2040, opened 2023-02-09
Hoist shared GraphQL code from `isomorphic-utils`

Alfred-Mountfield:

We had a dumping ground package for various utilities, and we intend to separate it into more meaningful ones. This migrates the GraphQL related definitions and types into its own package, which in turn reduces the surface area for cyclical dependencies, which unblocks incoming work to migrate HASH to the new canary versions of the Block Protocol packages.

## 🔍 What does this change?

- Introduces a new `@localhash-graphql-shared` package (taking suggestions for naming)
- Moves GraphQL related code to that package, and fixes imports

No.

## ⚠️ Known issues

- There are still some cyclical dependencies, and messy types that are redefined in various places, this is fixed in the next PR.

## 🐾 Next steps

- Migrate HASH to the new canary versions of BP 0.3

## 🛡 What tests cover this?

- All normal tests I would guess (linting, jest, etc.)

## ❓ How to test this?

Check the full repo as usual for E2E functionality.


libs/@local/hash-graphql-shared/LICENSE.md
```diff
@@ -0,0 +1,607 @@
+GNU Affero General Public License
```
Alfred-Mountfield, line 1: @nonparibus do we want AGPL here as well? I copied the same license as the one from the package I hoisted these from.

nonparibus: Yes, this is correct. Thank you.

nonparibus (approved):
Legal ✅

thehabbos007 (approved):
LGTM, tested locally and over tandem. Thanks!

PR #2042, opened 2023-02-09
Expose external types to TS

TimDiekmann:

We have the concept of external types in the graph, but not in the TS side yet. This exposes the external ontology metadata.


libs/@local/hash-subgraph/src/types/element.ts
```diff
@@ -32,7 +32,11 @@ export type EntityTypeWithMetadata = Omit<
   "schema"
 > & { schema: EntityType };
 
-export type { OntologyElementMetadata } from "@local/hash-graph-client";
+export type {
+  ExternalOntologyElementMetadata,
+  OntologyElementMetadata,
+  OwnedOntologyElementMetadata,
+} from "@local/hash-graph-client";
```
Alfred-Mountfield, line 39:
```suggestion
} from "@local/hash-graph-client";

export const isExternalOntologyElementMetadata = (metadata: ontologyElementMetadata): metadata is  ExternalOntologyElementMetadata => {
  return (metadata as ExternalOntologyElementMetadata).fetchedAt !== undefined;
}

export const isOwnedOntologyElementMetadata = (metadata: ontologyElementMetadata): metadata is  OwnedOntologyElementMetadata => {
  return (metadata as OwnedOntologyElementMetadata).ownedById !== undefined;
}
```

Let's add type-guards and then replace the cast above in `account-entity-type-list.tsx`

thehabbos007 (approved):

PR #2044, opened 2023-02-09
Correctly import external types from the type fetcher

TimDiekmann:

Currently, types from blockprotocol.org are treated as "owned" types. This changes this and imports them as "external" types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203936439790997/f) _(internal)_

## 🚫 Blocked by

The CI is likely to fail before these PRs are merged

- #2038 
- #2017 
- #2042

## 🔍 What does this change?

- Store the fetched-at time and pass it from the type-fetcher to the graph
- Fix wrong (de-)serialization
- Store external types as "external"


thehabbos007 (approved):
Awesome! Tested locally and seems to work just fine :)

I suppose this would just not insert data types if there is no network connection or if `blockprotocol.org` is down?

TimDiekmann:
Correct, the `server` subcommand will if the data types are not fetchable (for whatever reason).

PR #2048, opened 2023-02-10
Rename `EditionId` to `RecordId`

Alfred-Mountfield:
## 🌟 What is the purpose of this PR?

As part of bringing HASH back in sync with the changes made in the Block Protocol for the new specification, we need to rename various fields. This PR renames `EditionId` to `RecordId` within the TypeScript code, and introduces some temporary mapping functions which should allow us to merge at least a few PRs without needing a dev branch.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203756383156843/f) _(internal)_

## 🚫 Blocked by

- [x] #2040 

## 🔍 What does this change?

- Renames `EntityEditionId` to `EntityRecordId`
  - Renames the fields inside it to be consistent with the `0.3` branch of the BP
- Renames `OntologyTypeEditionId` to `OntologyTypeRecordId`
  - Renames the fields inside it to be consistent
- Introduces correct `VertexId` types
- Creates temporary mapping functions.   

## 📜 Does this require a change to the docs?

N/A

## ⚠️ Known issues

Many more inconsistencies to be addressed before we can remove the mapping functions

## 🐾 Next steps

- Rename `EditionId` in the Rust Graph Service
- Continue migrating types

## 🛡 What tests cover this?

All the normal ones

## ❓ How to test this?

Test as usual, ensure the app works and CI passes

## 📹 Demo

N/A

PR #2059, opened 2023-02-10
Refactor branded types, and advanced types

Alfred-Mountfield:

As part of bringing HASH back in sync with the changes made in the Block Protocol for the new specification, we need to refactor various things which will cause conflicts otherwise, and also will uncover some cyclical dependencies.

Please see commits for details, but generally this PR introduces a new package `@local/advanced-types`, and moves all of the branded types into the `@local/hash-subgraph` package. (Which should be renamed in a follow-up)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203756383156843/f) _(internal)_

## 🚫 Blocked by

- [x] #2048 

## 🔍 What does this change?

See commits 

N/A

## ⚠️ Known issues

Many more inconsistencies to be addressed before we can remove the mapping functions

## 🐾 Next steps

- Rename `EditionId` in the Rust Graph Service
- Continue migrating types

## 🛡 What tests cover this?

All the normal ones

## ❓ How to test this?

Test as usual, ensure the app works and CI passes

## 📹 Demo

N/A


nonparibus (approved):
Legal ✅

libs/@local/advanced-types/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0-private",
   "private": true,
   "description": "A collection of helper types",
-  "license": "AGPL-3.0",
+  "license": "MIT",
```
nonparibus, line 6: In `package.json` when dual-licensing do we always only declare MIT? @Alfred-Mountfield @TimDiekmann

nonparibus: IIRC in Crates we do both? Does npm not support this?

nonparibus (dismissed):

libs/@local/advanced-types/src/typed-entries.ts
```diff
@@ -0,0 +1,39 @@
...
+  : T extends ReadonlyArray<infer U>
+  ? [`${number}`, U]
+  : ObjectEntry<T>;
+
+/** `Object.entries` analogue which returns a well-typed array */
+export function typedEntries<T extends {}>(object: T): ReadonlyArray<Entry<T>> {
+  return Object.entries(object) as unknown as ReadonlyArray<Entry<T>>;
+}
+
+/** `Object.values` analogue which returns a well-typed array */
```
thehabbos007, line 31: This comment is a copy of `typedValues`
```suggestion
/** `Object.keys` analogue which returns a well-typed array */
```

apps/hash-frontend/next.config.js
```diff
@@ -1,10 +1,11 @@
 const { config } = require("dotenv-flow");
 const withTM = require("next-transpile-modules")([
   "@hashintel/design-system",
+  "@local/advanced-types",
   "@local/hash-graph-client",
   "@local/hash-graphql-shared",
   "@local/hash-isomorphic-utils",
-  "@local/hash-subgraph",
+  "@local/hash-subgraph/main",
```
thehabbos007, line 8: I don't think this is correct. Shouldn't it just be the package?
```suggestion
  "@local/hash-subgraph",
```

Otherwise I get the following error: `next-transpile-modules - an unexpected error happened when trying to resolve "@local/hash-subgraph/main"`

libs/@local/advanced-types/.babelrc
```diff
@@ -0,0 +1,15 @@
+{
```
thehabbos007, line 1: I don't think we need this file

thehabbos007 (changes requested):
Thanks, Alfie! I have some comments as the app isn't working

libs/@local/advanced-types/.babelrc
Alfred-Mountfield: I think you're right. It's still unclear to me which configs/scripts/tools we need to put into new packages and which we don't. It'd be really great if we had a guide (or even a command) on how to make a new package in our monorepo to be honest. cc @kachkaev

libs/@local/advanced-types/src/typed-entries.ts
Alfred-Mountfield: Oops good catch

libs/@local/advanced-types/.babelrc
thehabbos007: Yeah I agree, that would be nice to have. I think this babelrc is for the block? I am actually not sure at all where babel is used :sweat_smile:

thehabbos007 (approved):

libs/@local/advanced-types/.babelrc
Alfred-Mountfield: I think I copied it from `isomorphic-utils`, so I'm not sure it's necessary for a block?

thehabbos007 (approved):

PR #2060, opened 2023-02-10
Change vercel disabled branches to match `trunk-merge/*`

thehabbos007:
Seems that the pattern is `trunk-merge/2048/1qeq9ATg-UYT2` and not with a slash before the PR number. Maybe this would work?


Alfred-Mountfield (approved):
Previous one didn't break deployments so lets try it

PR #2067, opened 2023-02-14
Use `VersionedUri` instead of `String` in type fetcher

TimDiekmann:

Currently, the type fetcher takes in a string, but we have a more descriptive way for ontology types. This also makes the traversal operations public, so they can be used in the fetching client.

This is a required change to have a working type fetcher at HASH to unblock further work on the `FetchingStore`

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203958456055089/f) _(internal)_

## 🔍 What does this change?

- Use `VersionedUri` in the type fetcher
- Expose traversal functions

### DIFF
--- a/apps/hash-graph/lib/type-fetcher/src/fetcher.rs
+++ b/apps/hash-graph/lib/type-fetcher/src/fetcher.rs
@@ -3,7 +3,7 @@ use std::fmt;
 use error_stack::Context;
 use serde::{Deserialize, Serialize};
 use time::OffsetDateTime;
-use type_system::repr;
+use type_system::{repr, uri::VersionedUri};
 
 // We would really like to use error-stack for this. It's not possible because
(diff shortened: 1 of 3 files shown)


apps/hash-graph/lib/type-fetcher/src/fetcher.rs
```diff
@@ -56,6 +56,6 @@ impl TypeFetchResponse {
 pub trait Fetcher {
     /// Fetch an entity type by its URL and return all types that are reachable from it.
```
thehabbos007, line 57:
```suggestion
    /// Fetch an ontology type by its URL and return all types that are reachable from it.
```

thehabbos007 (commented):
Looks good to me, just as we've talked about :+1:

Just a small missed comment, happy to approve after

thehabbos007 (approved):

PR #2068, opened 2023-02-14
Implement `Store` for `FetchingStore`

TimDiekmann:

This mainly adds boilerplate code, which is currently required to implement `Store`. In a follow-up, the logic to invoke the type fetcher will be added to this location

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203958456055093/f) _(internal)_

## 🚫 Blocked by

- #2067 

## 🔍 What does this change?

- Implement `Store` logic on `FetchingStore` by passing down the parameters to `S`
- Remove the `Transaction` trait and directly implement it on `PostgresStore` instead

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/fetcher.rs
+++ b/apps/hash-graph/lib/graph/src/store/fetcher.rs
@@ -1,3 +1,5 @@
+use std::borrow::Borrow;
+
 use async_trait::async_trait;
 use error_stack::{IntoReport, Result, ResultExt};
 use tokio::net::ToSocketAddrs;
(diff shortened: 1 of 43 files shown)


apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -131,3 +109,213 @@ where
...
+    S: AccountStore + Send,
+    A: Send + Sync,
+{
+    async fn insert_account_id(&mut self, account_id: AccountId) -> Result<(), InsertionError> {
+        self.store.insert_account_id(account_id).await
+    }
+}
+
+#[async_trait]
+impl<S, A> DataTypeStore for FetchingStore<S, A>
```
thehabbos007, line 140: A little bit verbose/annoying we have to impl everything and can't rely on fallbacks. You've already mentioned before you pushed for a modular approach, unfortunate that we didn't end up with that :/

apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -131,3 +109,213 @@ where
...
+        &mut self,
+        property_types: impl IntoIterator<
+            Item = (
+                PropertyType,
+                impl Borrow<OntologyElementMetadata> + Send + Sync,
+            ),
+            IntoIter: Send,
+        > + Send,
+    ) -> Result<(), InsertionError> {
+        self.store.create_property_types(property_types).await
```
thehabbos007, line 187: By the way, what does this delegate to at the end? Is it passed on to the postgres store, or what happens? And if it is passed to postgres, how does that happen in the router (my questions are likely answered in your next PR in the stack :) )

I read some stuff about using semigroups to model the routing chain of a HTTP request and I suppose dependencies could work in a sort of similar way for us, but honestly I have no clue where we would even begin to leverage something like that. Just seems like a powerful, existing pattern that would be highly composable.

thehabbos007 (approved):
Looks good to me, mostly plumbing as you pointed out :+1:

apps/hash-graph/lib/graph/src/store/fetcher.rs
TimDiekmann: Yes, it will still be passed down to the Postgres store, only some logic will be added in the follow-up to also insert other ontology types before.

apps/hash-graph/lib/graph/src/store/fetcher.rs
TimDiekmann: Yeah, it's a little unfortunate, but we need to adjust both, the creation and the updating logic, so it's not that bad. Only the `AccoutStore` is completely untouched.

PR #2072, opened 2023-02-14
(Mostly) Migrate HASH to BP 0.3

Alfred-Mountfield:

For quite a while now we've been iterating upon the Block Protocol spec natively within HASH. We've been testing out a variety of changes to the Type System, the Graph Module, etc. With an incoming launch of a new version of the Block Protocol we want to put the two projects back in sync (mostly).

This PR upgrades the versions of the Block Protocol packages to the newest canary versions which contain [most of the changes being made on the Block Protocol 0.3 branch](https://github.com/blockprotocol/blockprotocol/tree/0.3). In doing so, we have removed a lot of our early implementations of `Subgraph` traversal and now opt for the official ones we ported to the Protocol repo. We also fix a large quantity of mismatches in shapes of data (and namings of fields) that have come about of upstream changes.

We do however, want to maintain more information than the Protocol specifies, including things like provenance metadata, greater information captured inside `EntityId`s, distinct metadata for owned Ontology types, etc. These differences in types caused this PR to be much, much, bigger than originally hoped for, and even getting this in required a lot of prior work to remove some pre-existing cyclical dependencies and package management problems. (See #2040, #2048, #2059).

As such, this PR is absolutely _**massive**_, it was near impossible to separate it into components that were working, you may see that some commits have _(Broken Interim State)_ in their messages, this is to indicate that the app was broken for that commit. Commits without that message should still work as expected. While it is possible to therefore split this PR into a few PRs at those various inflection points, they won't make much semantic sense as chunks of work, and the additional overhead costs of multiple PR reviews, conflicts, etc. mean that a giant PR makes the most sense in this instance.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203701650452068/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203358502199087/1203756383156843/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203543021352041/1203783495017458/f) _(internal)_
- [Asana task](https://app.asana.com/0/1202805690238892/1203791431510341/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Effort has been made to keep individual commits well encapsulated, and clear, and as such the review should be a fairly logical linear process (making bigger ones that involve a tonne of import renames fairly easy to speed through). As such, please look through the commits to understand, and feel free (obviously) to open review comments where it is not clear.

- Weirdly enough, I don't think this gigantic PR actually does. We'll probably want to follow-up by renaming `hash-subgraph` to something that better encapsulates what it is, and probably adding rules around its usages (to avoid directly calling `@blockprotocol/graph` in most instances for example)

## ⚠️ Known issues

(All are linked to their relevant threads opened in GitHub)

- [The Subgraph Types are not assignable](https://github.com/hashintel/hash/pull/2072/files#r1106090797)
- [The compatibility of the message callbacks aren't checked](https://github.com/hashintel/hash/pull/2072/files#r1106014332)
- [The message handlers we _do_ implement aren't actually necessarily implemented correctly, or aren't compatible](https://github.com/hashintel/hash/pull/2072/files#r1106017238)
- [There's an issue with the compatibility of the `MergeReturn` type](https://github.com/hashintel/hash/pull/2072/files#r1106020367)
- A portion of the `page-creation` Playwright test was disabled in a previous PR due to the code block being broken, this is still the case

## 🐾 Next steps

- Decide on any actions to take on the issues described above
- Migrate the callback implementations to be compatible
- Fix the blocks that are in the HASH repo, especially the code block so it can be re-enabled within the Playwright tests

## 🛡 What tests cover this?

- `tsc`
- `eslint`
- `test:backend-integration`
- `test:playwright`
- The HASH graph test suite

## ❓ How to test this?

Aside from checking CI passes, please run the app and ensure it all works as usual. It should basically be non-visible change to the end-user.

## 📹 Demo

N/A (would just be a plain video of the app working)


apps/hash-frontend/src/components/hooks/block-protocol-functions/knowledge/knowledge-shim.ts
```diff
@@ -6,38 +6,55 @@
...
+/** @todo-0.3 - we really want some type safety on these if we can find it, but this doesn't work */
+// export type GetEntityMessageCallback = Subtype<
+//   EmbedderGraphMessageCallbacks<true>["getEntity"],
+//   MessageCallback<
+//     GetEntityRequest,
+//     null,
+//     MessageReturn<Subgraph<EntityRootType>>,
+//     ReadOrModifyResourceError
+//   >
+// >;
```
Alfred-Mountfield, line 40: This feels _**very**_ unfortunate and somewhat important. We can't make sure that our custom callbacks are compatible with the ones from the BP because our local `Subgraph` isn't assignable to the `Subgraph` from the BP.

Maybe make a custom `Subtype` helper specifically for these which ignores `Subgraph` or something?

This is important if we continue to enhance the typings of the BP hooks rather than using the vanilla ones (which is somewhat needed if we keep using the hooks across the frontend rather than just at the EA/block boundary)

apps/hash-frontend/src/components/remote-block/remote-block.tsx
```diff
@@ -14,7 +14,8 @@ import { useRemoteBlock } from "./use-remote-block";
 
 type RemoteBlockProps = {
   graphCallbacks: Omit<
-    EmbedderGraphMessageCallbacks,
+    /** @todo-0.3 - Add these back */
+    EmbedderGraphMessageCallbacks<true>,
```
Alfred-Mountfield, line 18: This is perhaps the biggest "well not really migrated" part of this PR.

I wasn't able to just use the methods from the BP, even though we migrated a lot of their message definitions to match, this is partly [due to our messageCallbacks not being assignable](https://github.com/hashintel/hash/pull/2072/files#r1106014332), but also because some of our implementations don't actually match yet (for example `uploadFile`). I'm not sure how we go about fixing these

apps/hash-frontend/src/pages/type-editor/blockprotocol-ontology-functions-hook.ts
```diff
@@ -32,28 +35,28 @@ import { useBlockProtocolUpdatePropertyType } from "../../components/hooks/block
...
-  | "deleteLink"
+  EmbedderGraphMessageCallbacks<true>,
   | "getLinkedAggregation"
   | "deleteEntity"
   | "deleteEntityType"
   | "createLinkedAggregation"
   | "updateLinkedAggregation"
   | "deleteLinkedAggregation"
-  // Replaced by new ontology callbacks
+  /** @todo-0.3 fix these inconsistencies */
```
Alfred-Mountfield, line 45: Very similar to [this thread](https://github.com/hashintel/hash/pull/2072/files#r1106017238)

libs/@local/hash-subgraph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,272 @@
...
+  left: LeftInterval,
+  right: RightInterval,
+): MergeReturn<LeftInterval, RightInterval> =>
+  intervalMergeWithIntervalBp<LeftInterval, RightInterval>(
+    left,
+    right,
+  ) as unknown as MergeReturn<
+    LeftInterval,
+    RightInterval
+  >; /** @todo fix this unknown cast */
```
Alfred-Mountfield, line 199: I'm not actually sure why this one is needed, I _think_ the types should be compatible

libs/@local/hash-subgraph/src/types/subgraph.ts
```diff
@@ -1,42 +1,84 @@
...
+/** @todo - Figure out the incompatible vertices/edges is it the `&` instead of `|`? */
+// export type Subgraph<
+//   RootType extends SubgraphRootType = SubgraphRootType,
+//   > = Subtype<{
+//   roots: RootType["vertexId"][];
+//   vertices: Vertices;
+//   edges: Edges;
+//   depths: GraphResolveDepths;
+//   temporalAxes: SubgraphTemporalAxes
+// }, SubgraphBp<true, RootType>;
```
Alfred-Mountfield, line 76: I'm really not sure what to do about this. I think it's to do with the index types on the inner objects (`Vertices` and `Edges`). The changes we've made here give us what is, in many ways, a better type than the upstream Block Protocol one, as the index types to the inner objects (e.g. `EntityId` or `BaseUri`) are all branded, and do not overlap. Because of this TS is able to actually know what thing we're referring to, so if you do `Vertices[BaseUri]`, it knows the values inside can only be knowledge graph vertices, whereas in the BP those types completely overlap (they are just `string`s), which means that you have to tell it what the inner type is, even when you know you're dealing with an `entityId`.

Because of this, I think the types actually _aren't_ compatible as the BP one is more permissive than we intend it to be (and the spec/definition we associate with a `Subgraph` will probably be more constrained than the type), and thus it's actually somewhat incorrect and **won't** be assignable.

This causes annoyances elsewhere (it means anything using a `Subgraph` isn't assignable either, such as the Graph callback definitions), and means we have to do `as unknown as Subgraph..` casts which isn't particularllyy safe. (We could create `mapLocalSubgraphToBp` and `mapSubgraphBpToLocal` functions I guess, and how in-depth we decide to make those beyond `as` casts would need to have considerations made about performance).

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
```diff
@@ -192,9 +194,7 @@ export const useEntityTypeValue = (
 
       const newUrl = extractBaseUri(res.data.schema.$id);
 
-      if (newUrl) {
-        await router.replace(newUrl, newUrl, { shallow: true });
-      }
+      await router.replace(newUrl, newUrl, { shallow: true });
```
Comment (thehabbos007) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:197]: The null/undefined check was removed from here. Are we sure that this would be valid always?

apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
```diff
@@ -98,31 +89,31 @@ impl<A> TemporalTagged for LimitedTimeIntervalBound<A> {
...
     }
 }
 
 impl<A> IntervalBound<Timestamp<A>> for LimitedTimeIntervalBound<A> {
     fn as_bound(&self) -> Bound<&Timestamp<A>> {
         match self {
-            Self::Included(limit) => Bound::Included(limit),
-            Self::Excluded(limit) => Bound::Excluded(limit),
+            Self::Inclusive(limit) => Bound::Included(limit),
+            Self::Exclusive(limit) => Bound::Excluded(limit),
```
thehabbos007, line 102: The `Bound`s are still called `Included` and `Excluded`, is that something we want to change in the `interval-ops` crate at some point?

apps/hash-graph/lib/graph/src/shared/identifier/time/interval.rs
```diff
@@ -49,7 +51,7 @@ impl<A> ToSchema<'_> for UnresolvedTimeInterval<A> {
                     openapi::Schema::OneOf(
                         openapi::OneOfBuilder::new()
                             .item(openapi::Ref::from_schema_name(
-                                TimeIntervalBound::<A>::schema().0,
+                                LimitedTimeIntervalBound::<A>::schema().0,
```
thehabbos007, line 54: The `start` property is using `TimeIntervalBound`, is that correct if the end is `Limited*`?

thehabbos007 (commented):
(most of the way through, so submitting a partial review as GH has been messing up on me this week)

apps/hash-frontend/src/components/hooks/block-protocol-functions/knowledge/use-block-protocol-update-entity.ts
```diff
@@ -22,69 +21,67 @@ export const useBlockProtocolUpdateEntity = (
...
+        leftToRightOrder,
+        rightToLeftOrder,
+        properties,
+      } = data;
 
-        const {
-          entityId,
+      const { data: updateEntityResponseData } = await updateFn({
+        variables: {
+          entityId, // @todo-0.3 consider validating that this matches the id format,
```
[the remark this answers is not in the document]
Alfred-Mountfield: I believe it is still relevant. We've typed it within the function but there's nothing stopping a block from passing some garbage string, so we may want to consider some form of validation step here.

This is a wider issue whereby we are nicely typing the message callbacks for convenience, but are implicitly casting a bunch of things away from what the BP spec guarantees. We do this mostly because we also use the hooks all over the frontend _as_ the interface to our backend.

I think a potential solution would involve introducing another layer in the middle. So we have
BP Layer (plain imports) <-Transform A-> Nicely Typed Layer <-Transform B-> GraphQL

Whereby the frontend can call the Nicely Typed Layer but we can add secondary validations and such within Transform A for some proper type safety. (It would also mean we wouldn't have to be redefining all of the message callbacks..

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx
Comment (Alfred-Mountfield) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page/use-entity-type-value.tsx:197]: I believe so, because `extractBaseUri` will throw an error if it's invalid (which we do not catch or handle). So either way `newUrl` should be valid here

apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
Alfred-Mountfield: These are `use std::ops::Bound;`, I made the same mistake in a previous PR when asking @TimDiekmann 😁

apps/hash-graph/lib/graph/src/shared/identifier/time/interval.rs
Alfred-Mountfield: @TimDiekmann can you look at this Q

thehabbos007 (commented):
One more comment. Got through it now, thank you, Alfie this is a big piece of work!

I'll test locally in a bit and report any errors

apps/hash-graph/lib/graph/src/shared/identifier/time/interval.rs
TimDiekmann: Yes, this is the intention. We don't allow `"unbounded"` upper bounds in the time axes

apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
TimDiekmann: `std` uses `Included` and `Excluded`. This is the reason why I argued for `"included"` and `"excluded"` as well.

tests/hash-backend-integration/jest.config.cjs
```diff
@@ -16,8 +16,6 @@ module.exports = {
     "@local/hash-isomorphic-utils(.*)":
       "<rootDir>/../../libs/@local/hash-isomorphic-utils/src$1",
     "@local/hash-graph-client": "<rootDir>/../../libs/@local/hash-graph-client",
-    "@local/hash-subgraph(.*)":
```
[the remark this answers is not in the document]
Alfred-Mountfield: The import should be resolved without the mapping, yes. It [uses Jest's native support](https://www.github.com/facebook/jest/issues/9771) for the `exports` field of a package, as opposed to relying on our file system mapping to remove the `src` (which is now handled by `exports`, as it should be in my opinion)

apps/hash-graph/lib/graph/src/shared/identifier/time/bound.rs
Alfred-Mountfield: > This is the reason why I argued for "included" and "excluded" as well.

Just for posterity, we mean in terms of the BP spec and the wider system. Instead we've chosen to have "inclusive" and "exclusive" across the system but (mostly) keep "included" and "excluded" within the Graph

thehabbos007 (approved):
LGTM, tested locally and the bugs were addressed. Thank you! We can always address follow-up post-merge

TimDiekmann (approved):

PR #2073, opened 2023-02-15
Properly run `clippy` on `hash-graph`

TimDiekmann:

We had the `config.toml` in `.config` instead of `.cargo` so clippy did not run recently. This fixes this issue and also makes sure, that clippy will be invoked for integration tests.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1203958456055101/f) _(internal)_

## 🔍 What does this change?

- [Address clippy lints](https://github.com/hashintel/hash/commit/c7887f84e267546331af65b2f218aa1b51a97ef4)
- [Run linting on integration tests as well](https://github.com/hashintel/hash/commit/3393d302e4653b2ecfa1ec8ee85dd08cf83f448a)
- [Make integration tests part of the workspace](https://github.com/hashintel/hash/commit/2a48af621b89ea32cc91217c94059eb620bb7142)
- [Apply same structure to benches](https://github.com/hashintel/hash/commit/5d3389ee6646c2da6fd60bf58a72373a13e87aa4)

## 🐾 Next steps

The approach on adding a parameter in `Makefile.toml` worked pretty well. We should use this pattern more often instead of overwriting everything.


apps/hash-graph/lib/graph/src/shared/subgraph/query.rs
```diff
@@ -166,6 +166,10 @@ pub struct StructuralQuery<'p, R: Record> {
 }
 
 impl<'p, R: Record> StructuralQuery<'p, R> {
+    #[expect(
+        clippy::same_name_method,
+        reason = "This is supposed to replace the derived `schema` method"
+    )]
     fn schema() -> RefOr<Schema> {
```
Alfred-Mountfield, line 173: It's not implementing the trait though right? So do we maybe we want to just call it something else like `open_api_schema` or is it actually important it's called the same thing?

apps/hash-graph/lib/graph/src/store/postgres/ontology/read.rs
```diff
@@ -101,8 +101,7 @@ where
                 let base_uri = BaseUri::new(row.get(base_uri_index))
                     .into_report()
                     .change_context(QueryError)?;
-                let version: i64 = row.get(version_index);
-                let version = OntologyTypeVersion::new(version as u32);
+                let version: OntologyTypeVersion = row.get(version_index);
```
Alfred-Mountfield, line 104: What changed to make this safe?

Alfred-Mountfield: Or was it safe originally

apps/hash-graph/tests/integration/postgres/links.rs
```diff
@@ -2,12 +2,12 @@ use graph::knowledge::{EntityLinkOrder, EntityProperties};
...
 use type_system::uri::{BaseUri, VersionedUri};
 
-use crate::postgres::DatabaseTestWrapper;
+use crate::DatabaseTestWrapper;
 
 #[tokio::test]
 async fn insert() {
-    let person_a = serde_json::from_str(entity::PERSON_A_V1).expect("could not parse entity");
-    let person_b = serde_json::from_str(entity::PERSON_B_V1).expect("could not parse entity");
+    let alice = serde_json::from_str(entity::PERSON_A_V1).expect("could not parse entity");
```
Alfred-Mountfield, line 9: Should we rename `PERSON_A_V1` to `PERSON_ALICE_V1` then? I think `alice` and `bob` in code is less likely to get mixed up compared to `person_a` and `person_b` so this is probably a good change, I just don't want us randomly giving variables human names if the thing inside it doesn't relate.

So perhaps us renaming it to `PERSON_ALICE_V1` will also help us try and make sure the contents of the JSON stay as "alice"

Alfred-Mountfield (commented):
Looks generally good, thanks! Bit worrying we didn't realise for so long but we caught it now, and almost all of the lints are tiny

apps/hash-graph/lib/graph/src/store/postgres/ontology/read.rs
TimDiekmann: We are now implementing `FromSql` for `OntologyTypeVersion`, which does this logic (and fails in the case we have a negative value, which should never happen)

apps/hash-graph/lib/graph/src/shared/subgraph/query.rs
TimDiekmann: The trait is implemented on the four type aliases. I will figure out if we can just implement the trait manually.

apps/hash-graph/lib/graph/src/store/postgres/ontology/read.rs
Alfred-Mountfield: Cheers

apps/hash-graph/lib/graph/src/shared/subgraph/query.rs
TimDiekmann: I renamed the function

apps/hash-graph/tests/integration/postgres/links.rs
TimDiekmann: [Rename person entity file names](https://github.com/hashintel/hash/pull/2073/commits/65ddd8e02d8c00addfc6a0f1d095cd4f5a428ea9)

Alfred-Mountfield: Can you also rename the variable names in the `test_data` files? They're still `entity::PERSON_A_V1`

TimDiekmann: Oh, I missed that, thanks! e61b7c84ee13a679540022e084747eb172b7b672

Alfred-Mountfield (approved):

thehabbos007 (dismissed):
Deer LGTM. @indietyp this touches Deer but none of the code. Feel free to have a look as well

Alfred-Mountfield (dismissed):

indietyp (approved):

thehabbos007 (approved):

PR #2021, opened 2023-02-08
CI: switch from quickinstall to `install-action`

indietyp:
## 🌟 What is the purpose of this PR?

Currently, we use `quickinstall` to install all dependencies required, but (especially the lint job) jobs have been failing spuriously. This tries to fix the issue by installing the big brother of `cargo-quickinstall`: `cargo-binstall`, which is from the same organization, but more actively maintained and active.

`binstall` falls back to the `quickinstall` repo of precompiled binaries if available.

## 🔗 Related links

- [Example of failed CI](https://github.com/hashintel/hash/actions/runs/4106758312/jobs/7085394791)

PR #2075, opened 2023-02-15
`CI`: automatically generate and check rust lint across repo

indietyp:

`clippy` does not support the use configuration files, so one needs to either set lint levels via a `.cargo/config.toml` or `#![allow(...)]`, etc in code.

We want to keep all lints in sync but have a fundamental problem: Different projects require different subsets of lints.

This is where this PR comes in! A new cargo-make command has been added: `cargo make update-clippy {check/apply}`. This is inspired by `cargo-cranky` but supports:

* no need to have a separate command, just use `cargo clippy`, only modifies `.cargo/config.toml`
* inheritance
	* files higher in the tree are taken into account as "parents", e.g. `/.config/lints.toml` specifies `allow = ["clippy::x"]`, `/libs/deer/.config/lints.toml` specifies `deny = ["clippy::x"]`, then the config file will write `-Dclippy::x`.

There are in total 5 levels:
* `allow` (`-A`)
* `warn` (`-W`)
* `force-warn` (`--force-warn`)
* `deny` (`-D`)
* `forbid` (`-F`)
* `default` (unsets any previous lint level configuration to default - omits it in the final)

Using this command every workspace/project that opts into it will have their separate `.cargo/config.toml`, this is necessary as we would otherwise `rustflags`.

Each file must have a section with delimited by `## START CLIPPY LINTS ##` and `## END CLIPPY LINTS ##`, the cargo-make command will only ever touch the content between them!

There are two modes:
* `cargo make update-clippy check` (for CI, will fail if there's a change in any of the files, telling why)
* `cargo make update-clippy apply` (apply lint changes to `.cargo/config.toml`)

## 🚫 Blocked By

* #2073

## 📹 Demo

image


.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+fn main() {
+    let cwd = env::var("CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY").expect("environment variable should exist");
+    let cwd = PathBuf::from(cwd);
+
+    let args = env::var("CARGO_MAKE_TASK_ARGS").expect("environment variable should exist");
+    let mode = args.split(" ").next().expect("at least one position argument specifying mode required");
+
+    match mode {
+        "apply" => apply(&cwd),
+        "check" => check(&cwd),
```
TimDiekmann, line 252: I don't think we need a `check` command, the CI will catch this when ensuring empty diffs

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+    allow: HashSet<String>,
+    #[serde(default)]
+    warn: HashSet<String>,
+    #[serde(default)]
+    deny: HashSet<String>,
+    #[serde(default)]
+    forbid: HashSet<String>,
+    #[serde(default)]
+    default: HashSet<String>,
+}
```
TimDiekmann, line 80: `force-warn` is missing but I don't know if there is a flag for it

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    // brute-force, search for the line:
+    // ## START CLIPPY LINTS ##
+    // and then:
+    // ## END CLIPPY LINTS ##
+    // take all values in between
+
+    if !contents.contains("## START CLIPPY LINTS ##") || !contents.contains("## END CLIPPY LINTS ##") {
+        panic!("malformed .cargo/config.toml, please add the required markers")
+    }
```
TimDiekmann, line 139: Not necessarily happy about this error message as it does not show how the file is expected to look like

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    // strip out content between the markers
+    let prefix = contents.lines().take_while(|value| value.trim() != "## START CLIPPY LINTS ##");
+    let suffix = contents.lines().skip_while(|value| value.trim() != "## END CLIPPY LINTS ##");
+
+    let indent = "    ";
```
TimDiekmann, line 185: We could use the same indent as `## START CLIPPY LINTS ##`

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+    // read the `config.toml`
+    let contents = fs::read_to_string(&path).expect(&format!("should be able to read {}", path.display()));
+
+    if !contents.contains("## START CLIPPY LINTS ##") || !contents.contains("## END CLIPPY LINTS ##") {
+        panic!("malformed .cargo/config.toml, please add the required markers")
+    }
+
+    // strip out content between the markers
+    let prefix = contents.lines().take_while(|value| value.trim() != "## START CLIPPY LINTS ##");
+    let suffix = contents.lines().skip_while(|value| value.trim() != "## END CLIPPY LINTS ##");
```
TimDiekmann, line 183: Would be good if the strings are `const PREFIX: &str` instead

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+        .chain(lints.deny.into_iter().map(|lint| format!(r#"{indent}"-D{lint}","#)))
+        .chain(lints.forbid.into_iter().map(|lint| format!(r#"{indent}"-F{lint}","#)));
+
+    let contents: Vec<_> = prefix.map(|line| line.to_owned())
+        .chain(once(format!("{indent}## START CLIPPY LINTS ##")))
+        .chain(body)
+        .chain(suffix.map(|line| line.to_owned()))
+        .collect();
+
+    fs::write(&path, contents.join("\n")).expect("able to write `.cargo/config.toml`");
```
TimDiekmann, line 197: The way the `expect` function is called is wrong. This would generate an output like
> thread 'main' panicked at 'able to write `.cargo/config.toml`'

```suggestion
    fs::write(&path, contents.join("\n")).expect("unable to write `.cargo/config.toml`");
```

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,255 @@
...
+
+fn main() {
+    let cwd = env::var("CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY").expect("environment variable should exist");
+    let cwd = PathBuf::from(cwd);
+
+    let args = env::var("CARGO_MAKE_TASK_ARGS").expect("environment variable should exist");
+    let mode = args.split(" ").next().expect("at least one position argument specifying mode required");
+
+    match mode {
+        "apply" => apply(&cwd),
```
TimDiekmann, line 251: Can we run these as dependency on the clippy task?

libs/deer/.cargo/config.toml
```diff
@@ -0,0 +1,107 @@
...
+    "-Wclippy::string_add",
+    "-Wclippy::match_wildcard_for_single_variants",
+    "-Wclippy::mod_module_files",
+    "-Wclippy::empty_structs_with_brackets",
+    "-Wunreachable_pub",
+    "-Wclippy::from_iter_instead_of_collect",
+    "-Wclippy::default_union_representation",
+    "-Dunsafe_code",
+    ## END CLIPPY LINTS ##
+]
```
TimDiekmann, line 107:
```suggestion
]

```

libs/deer/.config/lints.toml
```diff
@@ -0,0 +1,14 @@
+# deer specific lints
```
TimDiekmann, line 1: Are you going to remove these lints from `lib.rs` then?

.github/scripts/rust/Makefile.lint.toml
```diff
@@ -0,0 +1,3 @@
+[tasks.update-clippy]
+script_runner = "@rust"
+script = { file = "${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}/lint.rs", absolute_path = true }
```
TimDiekmann, line 3: I guess this will always be executed from the toolchain which is used to invoke `cargo make`?

TimDiekmann (commented):
I like the idea!

.github/scripts/rust/lint.rs
indietyp: There is, `--force-warn`, will add.

.github/scripts/rust/lint.rs
indietyp: `check` does a bit more actually. You can re-order and comment if you want to, and `check` will not fail. I can remove that part of the code tho.

Also(!) I wanted to have a check command as it is a common courtesy to have an equivalent non-write operation (see prettier, clippy, etc.)

indietyp: also if we run as a dependency, then running would start with the new lints and "swallow" the change, with this we immediately terminate, ensuring the user knows why this happened, and also what has changed (as a sanity check).

indietyp: also currently we're using `HashSet` for easy comparison and duplication elimination, which means that `apply` isn't necessarily stable (this can be fixed tho). `check` will also pass if you e.g. reorder your lints in the `lints.toml`

.github/scripts/rust/lint.rs
indietyp: fixed in [6296e8e](https://github.com/hashintel/hash/pull/2075/commits/6296e8e0910adbf20de2627ac7966415a999b836)

.github/scripts/rust/lint.rs
indietyp: fixed in [f798dce](https://github.com/hashintel/hash/pull/2075/commits/f798dce512d0598b591c5034af17ba792096a1fd)

.github/scripts/rust/lint.rs
indietyp: fixed in [f798dce](https://github.com/hashintel/hash/pull/2075/commits/f798dce512d0598b591c5034af17ba792096a1fd)

.github/scripts/rust/lint.rs
indietyp: reworked in [287f64a](https://github.com/hashintel/hash/pull/2075/commits/287f64abe7ef298b32ea879b6a602d14f160a8f5)

libs/deer/.config/lints.toml
indietyp: removed in [e729f4d](https://github.com/hashintel/hash/pull/2075/commits/e729f4dbc50b9b88d9dad7091bad61eacb5e8783)

.github/scripts/rust/lint.rs
indietyp: added in [8febc14](https://github.com/hashintel/hash/pull/2075/commits/8febc14322b2212a2e0276c2ebb9140290eb63a4)

.github/scripts/rust/lint.rs
TimDiekmann: I think stable ordering is very important as you may overwrite lint groups (e.g. `-Wclippy::pedantic`) with some specific lints (e.g. `-Aclippy::whatever_is_in_padentic`). Is the order we currently have correct to account for this? the `-A` flags should be the last.

indietyp: yep [8febc14](https://github.com/hashintel/hash/pull/2075/commits/8febc14322b2212a2e0276c2ebb9140290eb63a4) changed this, groups are ordered, this items within the group are not ordered.

indietyp: The group order is:

* forbid
* deny
* force-warning
* warning
* alloy

TimDiekmann: If we can make sure that `check` returns the same order as `apply` I'm fine with keeping it. But IMO the order should be deterministic. We actively ask to not change the section, so changing it should result in an error.

libs/deer/src/lib.rs
```diff
@@ -8,23 +8,6 @@
...
-    clippy::if_then_some_else_none,
-    clippy::print_stdout,
-    clippy::print_stderr,
-    clippy::mod_module_files
-)]
-// TODO: once more stable introduce: warning missing_docs, clippy::missing_errors_doc
-#![allow(clippy::module_name_repetitions)]
-#![allow(clippy::redundant_pub_crate)]
-#![allow(clippy::missing_errors_doc)]
-#![deny(unsafe_code)]
```
TimDiekmann, line 27: 🎉

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,281 @@
...
+
+fn main() {
+    let cwd = env::var("CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY").expect("environment variable should exist");
+    let cwd = PathBuf::from(cwd);
+
+    let args = env::var("CARGO_MAKE_CLIPPY_LINT_MODE").or_else(|_| env::var("CARGO_MAKE_TASK_ARGS")).expect("environment variable should exist");
+    let mode = args.split(" ").next().expect("at least one position argument specifying mode required");
+
+    match mode {
+        "apply" => apply(&cwd),
```
TimDiekmann, line 277: How do you feel about "generate"? "apply" sounds like you are fixing the lints

indietyp: fixed in [6b76071](https://github.com/hashintel/hash/pull/2075/commits/6b76071f8ced88156189b74f604d7d4e34588c67)

.github/scripts/rust/Makefile.lint.toml
```diff
@@ -0,0 +1,8 @@
...
+
+[tasks.update-clippy-check]
+private = true
+env = { CARGO_MAKE_CLIPPY_LINT_MODE = "check" }
```
TimDiekmann, line 7:
```suggestion
env = { CARGO_MAKE_CLIPPY_LINT_MODE = "generate" }
```

libs/error-stack/.cargo/config.toml
```diff
@@ -0,0 +1,106 @@
...
+    "-Wclippy::verbose_file_reads",
+    "-Wclippy::zero_sized_map_values",
+    "-Wfuture_incompatible",
+    "-Wnonstandard_style",
+    "-Wunreachable_pub",
+    "-Aclippy::missing_errors_doc",
+    "-Aclippy::module_name_repetitions",
+    "-Aclippy::redundant_pub_crate",
+    ## END CLIPPY LINTS ##
+]
```
TimDiekmann, line 106:
```suggestion
]

```

TimDiekmann (changes requested):

.github/scripts/rust/Makefile.lint.toml
indietyp: Why?

TimDiekmann: We don't want clippy to run with an outdated configuration I suppose

TimDiekmann: It's pretty annoying if you want to run clippy but `cargo make` does not let you run `clippy` because the check failed.

indietyp: well the commend wouldn't run, because `check` returns an exit code `1`, but I get your second point

indietyp: otherwise, (I am not as proficient in GitHub workflow magic), if the git repo is dirty after a task, would the GitHub action fail?

TimDiekmann: Yes, it will run a `git diff`, prints the diff and error

indietyp:
in #2021 we might want to pre-install `rust-script` if available, once the `binstall` issue has been solved

.github/scripts/rust/Makefile.lint.toml
indietyp: 🤝

TimDiekmann: Actually, we only do this for the `test` job. But you could add this snippet to the end of the `lint` job as well:

```yaml
      - name: Ensure empty git diff
        run: git --no-pager diff --exit-code --color
```

TimDiekmann (approved):
Great!

libs/antsi/.cargo/config.toml
```diff
@@ -0,0 +1,107 @@
...
+    "-Wclippy::zero_sized_map_values",
+    "-Wfuture_incompatible",
+    "-Wnonstandard_style",
+    "-Wunreachable_pub",
+    "-Aclippy::missing_errors_doc",
+    "-Aclippy::module_name_repetitions",
+    "-Aclippy::redundant_pub_crate",
+    "-Amissing_docs",
+    ## END CLIPPY LINTS ##
+]
```
TimDiekmann, line 107:
```suggestion
]

```

TimDiekmann (approved):

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    // read the `config.toml`
+    let contents = fs::read_to_string(&path).expect(&format!("should be able to read {}", path.display()));
+
+    if !contents.contains(PREFIX) || !contents.contains(SUFFIX) {
+        panic!(concat!(
+            "malformed .cargo/config.toml, please add the required markers ",
+            r###""## START CLIPPY LINTS ##" and "## END CLIPPY LINTS ##" on a separate line to "###,
+            "delimit the region this script is allowed to modify."
+        ))
```
TimDiekmann, line 194:
```suggestion
        panic!(
            "malformed .cargo/config.toml, please add the required markers \"{}\" and \"{}\" on a \
            separate line to delimit the region this script is allowed to modify.",
            PREFIX,
            SUFFIX
        )
```

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,281 @@
...
+        .unwrap_or_else(|| "    ".to_owned());
+
+    let body = lints.forbid.into_iter().map(|lint| format!(r#"{indent}"-F{lint}","#))
+        .chain(lints.deny.into_iter().map(|lint| format!(r#"{indent}"-D{lint}","#)))
+        .chain(lints.force_warn.into_iter().map(|lint| format!(r#"{indent}"--force-warn{lint}","#)))
+        .chain(lints.warn.into_iter().map(|lint| format!(r#"{indent}"-W{lint}","#)))
+        .chain(lints.allow.into_iter().map(|lint| format!(r#"{indent}"-A{lint}","#)));
+
+    let contents: Vec<_> = prefix.map(|line| line.to_owned())
+        .chain(once(format!("{indent}## START CLIPPY LINTS ##")))
```
TimDiekmann, line 213:
```suggestion
        .chain(once(format!("{indent}{PREFIX}")))
```

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,281 @@
...
+        .chain(body)
+        .chain(suffix.map(|line| line.to_owned()))
+        .collect();
+
+    fs::write(&path, contents.join("\n")).expect("unable to write `.cargo/config.toml`");
```
TimDiekmann, line 218:
```suggestion
    fs::write(&path, format!("{}\n", contents.join("\n"))).expect("unable to write `.cargo/config.toml`");
```

indietyp: fixed differently in [6a91206](https://github.com/hashintel/hash/pull/2075/commits/6a91206d27365ba06e2103d2dc9a916703ebd562)

.github/scripts/rust/lint.rs
```diff
@@ -0,0 +1,284 @@
...
+        .map(|value| &value[1..(value.len() - 2)]).fold(LintFile::default(), |mut acc, line| {
+            let (arg, lint) = line.split_at(2);
+
+            match arg {
+                "-A" => acc.allow.insert(lint.to_owned()),
+                "-W" => acc.warn.insert(lint.to_owned()),
+                "--force-warn" => acc.force_warn.insert(lint.to_owned()),
+                "-D" => acc.deny.insert(lint.to_owned()),
+                "-F" => acc.forbid.insert(lint.to_owned()),
+                _ => panic!("unrecognized rust flag {arg}")
```
TimDiekmann, line 172:
```suggestion
                _ => panic!("unrecognized rust flag {}", arg)
```

libs/error-stack/src/fmt/color.rs
```diff
@@ -161,7 +161,7 @@ impl Foreground {
 
         let buffer = &[if bright { b'9' } else { b'3' }, color.digit()];
         // This should never fail because both are valid ASCII
-        let control = core::str::from_utf8(buffer).unwrap();
+        let control = core::str::from_utf8(buffer).expect("valid utf-8 buffer");
```
TimDiekmann, line 164: Same as above, the output of `expect` is different

indietyp: fixed in [c96800b](https://github.com/hashintel/hash/pull/2075/commits/c96800bc492e75dfa78f7c665d009bf691af73b4)

TimDiekmann (approved):

PR #2076, opened 2023-02-15
Disable caching for `CI` job

TimDiekmann:

We encountered a lot of issues recently due to caching not working. This disables the caching on `warm-up repo`.

## 🔗 Related links

- [Slack discussion](https://hashintel.slack.com/archives/C02TWBTT3ED/p1676466201924619) _(internal)_

## ⚠️ Known issues

Caching usually speeds up the CI. In our case, however, running `yarn install` does not take very long. See the slack discussion linked above for more information.

### DIFF
--- a/.github/actions/warm-up-repo/action.yml
+++ b/.github/actions/warm-up-repo/action.yml
@@ -14,7 +14,7 @@ runs:
     - uses: actions/setup-node@v3
       with:
         node-version: 16 ## aligned with Node version on Vercel
-        cache: yarn
+        # cache: yarn ## Currently disabled because of frequent timeouts
 
     - run: yarn install
       shell: bash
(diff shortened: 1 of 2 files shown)


kachkaev (commented):
Cause: https://github.com/actions/cache/issues/810

I’d suggest doing something like https://github.com/vercel/next.js/pull/45198/

If anyone else approves though, I won’t mind.

TimDiekmann:
We considered caching with a timeout of 2 minutes but `actions/setup-node` does not expose it and moving a way is a way bigger task than just removing the cache.

thehabbos007 (dismissed):
As Tim mentioned, we did want to try out the timeouts, but it is not yet exposed in `actions/setup-node@v3` unfortunately

thehabbos007 (approved):

PR #2079, opened 2023-02-16
Add type fetcher to production infra

thehabbos007:

This PR adds the type fetcher service to our production terraform config.

I cheated a little by already deploying this in production.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203893826309840/f) _(internal)_

## 🔍 What does this change?

- Adds type fetcher in terraform module for the app
- Builds the container in CD

## 📹 Demo

Look at hosted instance!

### DIFF
--- a/.github/workflows/hash-backend-cd.yml
+++ b/.github/workflows/hash-backend-cd.yml
@@ -61,8 +61,6 @@ jobs:
           AWS_REGION: ${{ env.AWS_REGION }}
           AWS_ECR_URL: ${{ env.AWS_ECR_URL }}
           IMAGE_NAME: ${{ env.IMAGE_NAME }}
-          BUILD_ARGS: |
-            TYPE_FETCHER=yes
 
   build-api:
     name: Build and push HASH api image
--- a/.github/workflows/hash-backend-cd.yml
+++ b/.github/workflows/hash-backend-cd.yml
@@ -62,7 +62,7 @@ jobs:
           AWS_ECR_URL: ${{ env.AWS_ECR_URL }}
           IMAGE_NAME: ${{ env.IMAGE_NAME }}
           BUILD_ARGS: |
-            TYPE_FETCHER=no
+            TYPE_FETCHER=yes
 
   build-api:
     name: Build and push HASH api image
--- a/infra/terraform/hash/main.tf
+++ b/infra/terraform/hash/main.tf
@@ -143,14 +143,23 @@ module "application" {
   param_prefix                 = local.param_prefix
   cpu                          = 1024
   memory                       = 2048
-  graph_image                  = module.graph_ecr
   ses_verified_domain_identity = var.ses_verified_domain_identity
+  graph_image                  = module.graph_ecr
   graph_env_vars = concat(var.hash_graph_env_vars, [
+    { name = "HASH_GRAPH_TYPE_FETCHER_HOST", secret = false, value = "127.0.0.1" },
+    { name = "HASH_GRAPH_TYPE_FETCHER_PORT", secret = false, value = "4444" },
     { name = "HASH_GRAPH_PG_USER", secret = false, value = "graph" },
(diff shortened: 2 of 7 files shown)


infra/terraform/modules/hash_application/variables.tf
```diff
@@ -95,6 +112,6 @@ variable "api_env_vars" {
 }
 
 variable "ses_verified_domain_identity" {
-  type = string
+  type        = string
```
thehabbos007, line 115: Will adjust this spacing..

infra/terraform/modules/hash_application/type_fetcher.tf
```diff
@@ -0,0 +1,51 @@
...
+      timeout  = 5
+
+    }
+    logConfiguration = {
+      logDriver = "awslogs"
+      options = {
+        "awslogs-create-group"  = "true"
+        "awslogs-group"         = local.log_group_name
+        "awslogs-stream-prefix" = local.type_fetcher_service_name
+        "awslogs-region"        = var.region
```
thehabbos007, line 40: Logs haven't been published in prod yet, not sure what's happening. We may want to keep an eye on that. Perhaps it's because of the new `scratch` images? Seems like the logs are just not submitted to aws

TimDiekmann: This could be true. Is it possible for `aws` to mount a volume and `tail -f` that? Otherwise, we may want to register a logger via `tracing`, which can be used from `aws`. For a temporary solution we could use `alpine` again.

TimDiekmann (approved):
Changes LGTM, thanks! :tada:

I have a few questions, but nothing is blocking the PR.

TimDiekmann (approved):
Changes LGTM, thanks! 🎉 

I have a few questions, but nothing is blocking the PR.

.github/workflows/hash-backend-cd.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           AWS_ECR_URL: ${{ env.AWS_ECR_URL }}
           IMAGE_NAME: ${{ env.IMAGE_NAME }}
           BUILD_ARGS: |
-            TYPE_FETCHER=no
```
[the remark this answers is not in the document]
thehabbos007: Sure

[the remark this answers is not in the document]
thehabbos007: e577976ad20d3488636990c6fcca428de8fff7b9

infra/terraform/hash/main.tf
```diff
@@ -143,15 +143,24 @@ module "application" {
   param_prefix                 = local.param_prefix
   cpu                          = 1024
   memory                       = 2048
-  graph_image                  = module.graph_ecr
   ses_verified_domain_identity = var.ses_verified_domain_identity
+  graph_image                  = module.graph_ecr
   graph_env_vars = concat(var.hash_graph_env_vars, [
+    { name = "HASH_GRAPH_TYPE_FETCHER_HOST", secret = false, value = "127.0.0.1" },
+    { name = "HASH_GRAPH_TYPE_FETCHER_PORT", secret = false, value = "4444" },
```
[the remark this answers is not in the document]
thehabbos007: I think `localhost` would still work, I chose this to be explicit about using the loopback address. We use `awsvpc` networking mode which does some funky network interface fusing. Didn't want to risk something strange happening. `0.0.0.0` would likely not work though

TimDiekmann (approved):

infra/terraform/modules/hash_application/type_fetcher.tf
thehabbos007: Yeah i think we can using Amazon EFS volumes, but we already use the Cloudwatch log collector which relies on stdout. We should look into this when we get a chance

PR #2081, opened 2023-02-16
Make fields on common structs `pub`

TimDiekmann:

For the same reason as https://github.com/blockprotocol/blockprotocol/pull/987 this PR makes structs, which are only a collection of information (like `OntologyVertexId`, or `EntitMetadata`) publicly accessible. With the linked PR it's also possible to properly implement `From` implementations for `VersionedUri`.

These changes makes a few accesses a little bit more verbose when otherwise chaining is possible, however, in other cases, this greatly improves readability, especially when type hints are available.

I planned to do these changes as a drive-by to my next PR but have split the changes off to make reviewing easier.

## 🚫 Blocked by

- https://github.com/blockprotocol/blockprotocol/pull/987

## 🔍 What does this change?

Please see the list of commits to see the changes

This PR also implements `FromSql` for more types, which previously we didn't because from just looking at the code (without typehints), it was not possible to see, if the parameters returned from `row.get(N)` were passed to the correct location. With these changes it's now required to type the parameter explicitly, so it's easy to check if a call was made correctly (the parameter passed to `get()` can't be checked at compile time). 

Example:

```diff

< let owned_by_id = OwnedById::new(row.get(owned_by_id_index));
< let entity_uuid = EntityUuid::new(row.get(entity_uuid_index));
< let updated_by_id = UpdatedById::new(row.get(updated_by_id_index));
< 
< Ok(Entity::new(
<     properties,
<     link_data,
<     EntityRecordId::new(
<         EntityId::new(owned_by_id, entity_uuid),
<         EntityEditionId::new(row.get(edition_id_index)),
<     ),
<     EntityVersion {
<         decision_time: row.get(decision_time_index),
<         transaction_time: row.get(transaction_time_index),
<     },
<     entity_type_uri,
<     ProvenanceMetadata::new(updated_by_id),
<     // TODO: only the historic table would have an `archived` field.
<     //   Consider what we should do about that.
<     row.get(archived_index),
< ))
---
> Ok(Entity {
>     properties,
>     link_data,
>     metadata: EntityMetadata {
>         record_id: EntityRecordId {
>             entity_id: EntityId {
>                 owned_by_id: row.get(owned_by_id_index),
>                 entity_uuid: row.get(entity_uuid_index),
>             },
>             edition_id: row.get(edition_id_index),
>         },
>         version: EntityVersion {
>             decision_time: row.get(decision_time_index),
>             transaction_time: row.get(transaction_time_index),
>         },
>         entity_type_id,
>         provenance: ProvenanceMetadata {
>             updated_by_id: row.get(updated_by_id_index),
>         },
>         archived: row.get(archived_index),
>     },
> })
```

### DIFF
--- a/apps/hash-graph/Cargo.lock
+++ b/apps/hash-graph/Cargo.lock
@@ -2701,7 +2701,7 @@ dependencies = [
 [[package]]
 name = "type-system"
 version = "0.0.0"
-source = "git+https://github.com/blockprotocol/blockprotocol?rev=646b7d1#646b7d1c6cc35197ac76872d6c34e8e4185f2f0c"
+source = "git+https://github.com/blockprotocol/blockprotocol?rev=ce8df71#ce8df719ca03657f30849140e83843f60ecefa59"
 dependencies = [
  "console_error_panic_hook",
  "regex",
--- a/apps/hash-graph/Cargo.lock
+++ b/apps/hash-graph/Cargo.lock
@@ -2701,7 +2701,7 @@ dependencies = [
 [[package]]
 name = "type-system"
 version = "0.0.0"
-source = "git+https://github.com/blockprotocol/blockprotocol?rev=ce8df71#ce8df719ca03657f30849140e83843f60ecefa59"
+source = "git+https://github.com/blockprotocol/blockprotocol?rev=adab205#adab205d4724c992b0f4b8efbf21f8942050479b"
 dependencies = [
  "console_error_panic_hook",
  "regex",
--- a/apps/hash-graph/bench/Cargo.toml
+++ b/apps/hash-graph/bench/Cargo.toml
@@ -21,7 +21,7 @@ serde = { version = "1.0.152", features = ["derive"] }
 serde_json = "1.0.91"
 tokio = { version = "1.25.0", features = ["rt-multi-thread", "macros"] }
 tokio-postgres = { version = "0.7.7", default-features = false }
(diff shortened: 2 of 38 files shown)


indietyp:
I like this! I wonder if https://github.com/rust-lang/rust/issues/105077 might be of use.

TimDiekmann:
> I like this! I wonder if [rust-lang/rust#105077](https://github.com/rust-lang/rust/issues/105077) might be of use.

I'm really looking forward to that RFC, thanks for linking this here!

Alfred-Mountfield (commented):
I'm not really a fan of making the fields on `EntityMetadata` pub, the metadata shouldn't be mutable (and while I appreciate we can handle that by just not having `const mut entity_metadata ...`, not having any way of setting inner fields after construction is just more idiomatic and consistent with the rest of our code).

I don't think the gains in two functions of having `field: x` on the constructor is worth it, that explicitness is provided by IDEs with parameter hints already, and all of the types are distinct so you can't put them in in the wrong order.

We also don't have a need (to my knowledge) of destructuring these

TimDiekmann:
Thanks, @Alfred-Mountfield, I appreciate the concern. I have reverted changes to `OntologyElementMetadata`, `EntityMetadata`, and `ProvenanceMetadata` and added a TODO to track the RFC, which was linked by @indietyp.

Alfred-Mountfield (approved):

Alfred-Mountfield (approved):

PR #2087, opened 2023-02-17
Fix numbers not being passed correctly to Postgres

TimDiekmann:

We currently only require integers for predefined fields (fields inside of JSON like `"properties"."..."` are treated differently):
- unsigned 32-bit value for ontology versions
- signed 32-bit value for ordering (this may change in the future)

- It's possible to truncate the value for ontology versions if the number was too big (resulting in the wrong error, annoying but not a major issue)
- It's *not* possible to specify a parameter for `"leftToRightOrder"` or `"rightToLeftOrder"` *at all* as this will fail with the following error:
```
graph::api::rest::entity: Could not read entities from the store, error: Could not query from store
│
╰─▶ error serializing parameter 2: cannot convert between the Rust type `f64` and the Postgres type `int4`
```

## 🔗 Related links

- [Asana Task](https://app.asana.com/0/0/1203977361907410) (_internal_)
- [Another Asana Task](https://app.asana.com/0/0/1202884883200973) (_internal_)

## 🔍 What does this change?

Change the parameter to take an `i32` instead. This does not cover all values of an unsigned 32-bit integer, which is used for the ontology version, but changing this to an `i64` would require adding a new variant specifically for the ordering, the effort required for this is not worth it currently.

## ❓ How to test this?

- Make sure you can query for an entity by its link ordering
- Make sure you can still compare a property by numbers

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/query/filter.rs
+++ b/apps/hash-graph/lib/graph/src/store/query/filter.rs
@@ -267,7 +267,7 @@ pub enum FilterExpression<'p, R: Record + ?Sized> {
     Parameter(Parameter<'p>),
 }
 
-#[derive(Debug, PartialEq, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Deserialize)]
 #[serde(untagged)]
 pub enum Parameter<'p> {
     Boolean(bool),
--- a/apps/hash-graph/lib/graph/src/store/query/filter.rs
+++ b/apps/hash-graph/lib/graph/src/store/query/filter.rs
@@ -271,7 +271,7 @@ pub enum FilterExpression<'p, R: Record + ?Sized> {
 #[serde(untagged)]
 pub enum Parameter<'p> {
     Boolean(bool),
-    Number(f64),
+    Number(i32),
     Text(Cow<'p, str>),
     Any(Value),
     #[serde(skip)]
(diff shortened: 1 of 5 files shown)


apps/hash-graph/lib/graph/src/store/query/filter.rs
```diff
@@ -345,33 +345,31 @@ impl Parameter<'_> {
...
-                *self = Parameter::Number(number.as_f64().ok_or_else(|| {
+                let number = number.as_i64().ok_or_else(|| {
                     Report::new(ParameterConversionError {
                         actual: self.to_owned(),
                         expected,
                     })
-                })?);
+                })?;
+                *self =
+                    Parameter::Number(i32::try_from(number).into_report().change_context_lazy(
```
thehabbos007, line 358: Not sure if I'm missing something, but are we not converting back and forth here?

thehabbos007: I.e. `i32 -> i64 -> i32`

TimDiekmann: No, we have a `serde_json::Number`, which has to be parsed as either `i64`, `u64` or `f64`, so we're parsing it as `i64`.

PR #2090, opened 2023-02-17
Support inserting the same external type with different versions

TimDiekmann:

The current database layout asserts, that we only call "create" once per Base URI.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643284/1203958456055099/f) _(internal)_

## 🔍 What does this change?

- Adjust database to allow multiple versions of external ontology data to be **created** (we can't _update_ them as they may be inserted in arbitrary order)
- Greatly improve the error handling code in Postgres

## 🛡 What tests cover this?

- A bunch of new integration tests was added to catch different possibilities of inserting various data. This also asserts the correct errors.

## ❓ How to test this?

- Modify the Rest API tests and try out different configurations
- keep in mind that you cannot insert external types from the Rest API, so may you also want to play around with the newly added tests

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres.rs
@@ -262,8 +262,8 @@ where
                 FROM create_owned_ontology_id(
                     base_uri := $1,
                     version := $2,
-                    record_created_by_id := $4,
-                    owned_by_id := $3
+                    record_created_by_id := $3,
+                    owned_by_id := $4
                 );"#,
                 &[
                     &metadata.record_id().base_uri.as_str(),
@@ -306,14 +306,14 @@ where
                 FROM create_external_ontology_id(
                     base_uri := $1,
                     version := $2,
-                    record_created_by_id := $4,
-                    fetched_at := $3
+                    record_created_by_id := $3,
+                    fetched_at := $4
                 );"#,
                 &[
                     &metadata.record_id().base_uri.as_str(),
                     &metadata.record_id().version,
-                    &metadata.fetched_at(),
                     &metadata.provenance_metadata().updated_by_id(),
+                    &metadata.fetched_at(),
                 ],
             )
             .await
--- a/apps/hash-graph/lib/graph/src/store/postgres.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres.rs
@@ -35,10 +35,9 @@ use crate::{
     },
     provenance::{OwnedById, ProvenanceMetadata, UpdatedById},
     store::{
-        error::{VersionedUriAlreadyExists, WrongOntologyVersion},
+        error::{OntologyTypeIsNotOwned, OntologyVersionDoesNotExist, VersionedUriAlreadyExists},
         postgres::ontology::{OntologyDatabaseType, OntologyId},
-        AccountStore, BaseUriAlreadyExists, BaseUriDoesNotExist, InsertionError, QueryError,
-        StoreError, UpdateError,
+        AccountStore, BaseUriAlreadyExists, InsertionError, QueryError, StoreError, UpdateError,
     },
     subgraph::edges::GraphResolveDepths,
 };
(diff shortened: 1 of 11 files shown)


apps/hash-graph/lib/graph/src/store/postgres.rs
```diff
@@ -263,21 +262,21 @@ where
...
-                    record_created_by_id := $4
+                    record_created_by_id := $4,
+                    owned_by_id := $3
                 );"#,
                 &[
                     &metadata.record_id().base_uri.as_str(),
                     &metadata.record_id().version,
-                    &metadata.owned_by_id(),
                     &metadata.provenance_metadata().updated_by_id(),
+                    &metadata.owned_by_id(),
```
thehabbos007, line 272: nit: 
Should we swap these around so they're in order with respect to the params in `create_owned_ontology_id`? It's strange to see the order being `1,2,4,3` above. I suppose this is the case for the calls function below in this file as well

apps/hash-graph/postgres_migrations/V5__ontology_metadata_functions.sql
```diff
@@ -39,6 +45,20 @@ DECLARE
...
+    ) VALUES (
+      create_external_ontology_id.base_uri
+    );
+  EXCEPTION WHEN unique_violation THEN
+    IF EXISTS (SELECT FROM ontology_ids NATURAL JOIN ontology_owned_metadata WHERE ontology_ids.base_uri = create_external_ontology_id.base_uri) THEN
+      RAISE EXCEPTION 'Owned ontology with base_uri `%` already exists',
+        create_external_ontology_id.base_uri
+      USING ERRCODE = 'invalid_parameter_value';
+    END IF;
+  END;
```
thehabbos007, line 60: Is this nested transaction required for us to catch the unique violation?

And for my understanding, is this explictitly ignoring the exception if the base URI already exists for a non-owned ontology type?

apps/hash-graph/postgres_migrations/V5__ontology_metadata_functions.sql
```diff
@@ -60,97 +80,50 @@ CREATE
...
   "version" BIGINT,
+  "version_to_update" BIGINT,
   "record_created_by_id" UUID
 ) RETURNS TABLE (ontology_id UUID, owned_by_id UUID) AS $update_owned_ontology_id$
 DECLARE
   "_ontology_id" UUID;
 BEGIN
-  SET CONSTRAINTS ontology_owned_metadata_pk DEFERRED;
-  SET CONSTRAINTS ontology_owned_metadata_fk DEFERRED;
-  SET CONSTRAINTS ontology_ids_overlapping DEFERRED;
```
thehabbos007, line 70: Very happy about these being gone!

apps/hash-graph/postgres_migrations/V5__ontology_metadata_functions.sql
```diff
@@ -60,97 +80,50 @@ CREATE
...
   WHERE ontology_ids.ontology_id = metadata.ontology_id
-    AND ontology_ids.base_uri = update_external_ontology_id.base_uri
-    AND ontology_ids.transaction_time @> now()
-  RETURNING metadata.ontology_id;
+    AND ontology_ids.base_uri = update_owned_ontology_id.base_uri
+    AND ontology_ids.version = update_owned_ontology_id.version_to_update
+  RETURNING _ontology_id, metadata.owned_by_id;
   IF NOT FOUND THEN
-    RAISE EXCEPTION 'No external ontology type with base_uri % does not exist', update_external_ontology_id.base_uri
+    RAISE EXCEPTION 'No owned ontology type with base_uri `%` and version `%` does exist',
```
thehabbos007, line 109:
```suggestion
    RAISE EXCEPTION 'No owned ontology type with base_uri `%` and version `%` exists',
```

thehabbos007 (commented):
Awesome! I have some comments, but it looks great. Especially the new integration tests.

This would require a DB wipe, though. We need to figure out when we would want to merge/wipe

apps/hash-graph/postgres_migrations/V5__ontology_metadata_functions.sql
TimDiekmann: I was not able to come up with an easier solution 🙁 

Yes, this was the only way I found to catch the exception. Happy to change this if you have a better idea 😄

thehabbos007: I see, I don't have a better solution, just wanted to know the context. This makes sense, though :)

thehabbos007 (approved):

PR #2097, opened 2023-02-18
`antsi`: extensions

indietyp:

This creates the types that remained from the ISO 6429 standard (51-69), `CMY`, `CMYK`, and chooses to **not** implement `26` and `50` (`CCITT T.61`), as the related standard was withdrawn.

This also includes extensions of popular modern terminal emulators:
* wezterm (`RGBA`)
* kitty + VTE (underline color, additional underline types)
* mintty (overstrike, subscript, superscript)

`kitty + VTE` is also supported by `wezterm` and `mintty`.
`mintty` *script is also supported by `wezterm`

This PR also adds and refines some more documentation.


libs/antsi/src/config.rs
```diff
@@ -0,0 +1,130 @@
...
+//! advanced colors, like [`crate::IndexedColor`], or [`crate::RgbColor`].
+//!
+//! The two differences across implementations are:
+//! * [`Delimiter`]
+//! * [`Compliance`]
+//!
+//! For more information please look at the respective documentation. `antsi` uses
+//! [`Delimiter::Colon`] and [`Compliance::Iso`] as default in all documentation.
+// TODO: recommendations for most used terminal emulators
+use core::sync::atomic::{AtomicU8, Ordering};
```
TimDiekmann, line 14: I'm not particularly happy to require those types. Why do we need to store the configuration globally? It implies, that we assume `antsi` is used in multithreaded applications, but IMO `antsi` should not be agnostic over those assumptions. Why don't we let the downstream crates handle this behavior?

TimDiekmann: > a user outputs their content into a single type of terminal

I think this assumption is wrong, e.g. when writing to a terminal _and_ to a log file you may want different formats.

indietyp: 🤔 even then, we need a way to globally configure this to stay consistent while allowing for compatibility with older terminals and I feel like consistency across all formatting should be a top priority, as to not confuse the library user or end user of the application. This is also what all current applications do, they use a specific combination (usually `;` + `Legacy`) and then stick w/ it, but I'd like to give the user the choice of what that combination is. 🤔 

We *could*, of course, also do something similar to: `let _guard = Delimiter::Color.set()`. If we wanted to output to both, the log file stream could use that instead when interpolating the `Display`.

The log file is a great example, as we have another problem: if we want to output to a log file, meaning non-interactive, one should be able to strip color. This is what all other apps usually do. This, of course, could also be done by a downstream crate.

indietyp: We have decided offline to remove the configuration part. To properly support nested color reset (e.g. `[blue]Value [red]Value[/red] Value[/blue]`), we will need a stateful printer anyway. Configuration of the delimiter will default to the legacy version and be configurable on the printer itself.

libs/antsi/src/font.rs
```diff
@@ -100,17 +105,29 @@ pub enum Blinking {
...
+
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Default)]
 pub struct Font {
     weight: Option<FontWeight>,
     family: Option<FontFamily>,
+    // mintty extension
+    script: Option<FontScript>,
 
-    // Value layout: `XXXX_IRHS`
+    // Value layout: `XXÖO_IRHS`
```
TimDiekmann, line 122: Should we use the bitflags crate here instead?

indietyp: I considered that but did not want to pull in an additional dependency, although it would be better.

libs/antsi/src/font.rs
```diff
@@ -91,6 +89,13 @@ pub enum FontFamily {
 pub enum Underline {
     Single,
     Double,
+
+    // kitty + vte extension
+    Curly,
+    // kitty + vte extension
+    Dotted,
+    // kitty + vte extension
+    Dashed,
```
TimDiekmann, line 98: It doesn't seem that these are in the standard but extensions? Not a blocker for this PR but we really should consider feature flags.

libs/antsi/src/font.rs
TimDiekmann: I think for now it's fine, but in the case where we hit more than 8 bits, we definitely should pull in a crate, otherwise, we would need to deal with endianness.

TimDiekmann (dismissed):

libs/antsi/src/font.rs
indietyp: Yes, but ANSI says that unrecognized escape sequences are to be ignored. Therefore even if not enabled, their formatting will still be ignored. I agree, tho that in the future (next PR), these will move into feature flags, while ignored they still provide the library user w/ more insight, by telling them: "this is an extension and not supported by every terminal emulator"

TimDiekmann (approved):

PR #1854, opened 2023-01-17
Rename time related structs to avoid the term "Projection"

TimDiekmann:
## 🌟 What is the purpose of this PR?

To make the temporal versioning easier to understand, this PR renames the time related structs around `Projection`.

🎉 **With this PR the temporal versioning implementation is considered done!** 🎉 

Any further changes are most likely optimization or fixes, but the functionality has landed.

## 🔗  Related links

- [Asana Task](https://app.asana.com/0/1203543021352041/1203783495017458/f) _(internal)_

## 🔍 What does this change?

- Rename `TimeProjection` to `TemporalAxes` (+ `Unresolved*`)
- Removed `Projection`
- Removed `DecisionTimeProjection`
- Removed `TransactionTimeProjection`
- Rename `ProjectedTime` to `VariableTimeAxis`
- Greatly simplified the definition of `TemporalAxes` to generate a better OpenAPI spec and remove a lot of intermediate structs.

## 📜 Does this require a change to the docs?

The docs were adjusted. I changed all documentation I found around "projection", "kernel", and "image"

PR #2106, opened 2023-02-20
Make `utoipa` definitions more readable

TimDiekmann:

We currently have still a lot of `OneOf` and more unreadable things in our `api.ts` file. This makes it very unreadable. By only tweaking a few small bits on the Graph API we are able to greatly improve readability. One example is removing `OntologyOutwardEdgesOneOfKindEnum` and just using (the already defined) `OntologyEdgeKind` enumeration.

This also fixes a naming issue on the `*Edges` enumeration and removes the trailing `s`.

## 🚫 Blocked by

- #1854


Alfred-Mountfield (approved):
Fantastic improvement, it's a shame this stuff is so fiddly, can appreciate how difficult/frustrating this might have been to accomplish

PR #2110, opened 2023-02-21
Generalize edge creation in the graph to abstract over `EntityidWithInterval`s

TimDiekmann:
## 🌟 What is the purpose of this PR?

Currently, our representation of edges in the Graph is very limited. We only store a generic `OutwardEdge` struct, which holds exactly one base id and one revision id, merging the intervals for overlapping edge lifespans is only very hard to achieve. This introduces a few small crates to identify the `VertexId` and the `EdgeEndpoint` to abstract over them. A generic `AdjecencyList` was added, which can handle the insertion logic depending on the required edge layout.

This PR does not introduce the union of intervals but provides an entry point for this, where the logic can simply be added in a follow-up PR.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1203998818778056/f) _(internal)_

## 🚫 Blocked by

- #1854 
- #2106 
- #2109 

## 🔍 What does this change?

I encapsulated every change in fine-grained commits. Please see the commit list.

## 🐾 Next steps

Merge overlapping edge endpoints ([Asana task](https://app.asana.com/0/1201095311341924/1203399924452451/f))

PR #2115, opened 2023-02-23
Implement `Interval` ordering

TimDiekmann:

Currently, we cannot sort intervals. Sorting is required to have a more or less efficient and easy way to merge overlapping intervals in a list.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203835712237857/f) _(internal)_

## 🚫 Blocked by

- #2110 

## 🔍 What does this change?

- Use the correct bounds for `Interval` operations: The bounds needs to be `Ord` and `Eq` over `PartialOrd` and `PartialEq`, otherwise weird panic messages will happen when for example comparing `[NaN, +∞)` and `[NaN, +∞)`. The intervals can be constructed (as `NaN` is not greater than `+∞`), but comparing `NaN` to `NaN` will panic)
- Implement `PartialOrd` and `Ord` for `Interval`

### DIFF
--- a/apps/hash-graph/lib/graph/src/shared/interval.rs
+++ b/apps/hash-graph/lib/graph/src/shared/interval.rs
@@ -115,7 +115,7 @@ impl<T, S: IntervalBound<T>, E: IntervalBound<T>> Interval<T, S, E> {
     /// Panics if the start bound is greater than the end bound.
     pub fn new(start: S, end: E) -> Self
     where
-        T: PartialOrd,
+        T: Ord,
     {
         assert_ne!(
             compare_bounds(
@@ -123,6 +123,7 @@ impl<T, S: IntervalBound<T>, E: IntervalBound<T>> Interval<T, S, E> {
                 end.as_bound(),
                 BoundType::Start,
                 BoundType::End,
+                Ord::cmp,
             ),
             Ordering::Greater,
             "Start bound must be less than or equal to end bound"
@@ -149,7 +150,7 @@ impl<T, S: IntervalBound<T>, E: IntervalBound<T>> Interval<T, S, E> {
         other: &Interval<T, impl IntervalBound<T>, impl IntervalBound<T>>,
     ) -> bool
     where
-        T: PartialOrd,
+        T: Ord,
     {
         // Examples |      1     |     2
         // =========|============|============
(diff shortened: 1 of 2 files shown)


apps/hash-graph/lib/graph/src/shared/interval.rs
```diff
@@ -448,6 +451,45 @@ where
...
+    S: IntervalBound<T>,
+    E: IntervalBound<T>,
+{
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let start_ordering = compare_bounds(
+            self.start_bound(),
+            other.start_bound(),
+            BoundType::Start,
+            BoundType::Start,
+            PartialOrd::partial_cmp,
```
thehabbos007, line 466: Is `partial_cmp` only compatible with the return type because the `Option<Ordering>` return type satisfies `From<Ordering>`?

What happens here in case of None from the `partial_cmp` comparison?

thehabbos007 (commented):
I do have a question below, but am wondering if we want to ensure/test for the rules given in https://doc.rust-lang.org/core/cmp/trait.PartialOrd.html? i.e. `transitivity` and `duality`. Or do you think that's not needed given our use case for the trait?

apps/hash-graph/lib/graph/src/shared/interval.rs
TimDiekmann: Yes, that's the reason why it's working. If `None` is returned from `partial_cmp`, `None` will be returned fro `compare_bounds` and this will be forwarded to the caller as well (`?` also passes `None` up)

TimDiekmann:
Thanks for the review!
> Or do you think that's not needed given our use case for the trait?

Generally, I try not to follow this principle. If we implement a trait, it should be implemented correctly. I'd love to add some test cases here, but my concern is that without fuzzing data, this will be very hard to test properly, as I'm unable to come up with a test case, where transitivity or duality is not fulfilled.
I'm pretty sure, that this is implemented correctly given the number of tests we have for intervals, which all rely on the `compare_bounds` function (which is the only location, where transitivity or duality constraints could be violated).

thehabbos007 (dismissed):

thehabbos007 (approved):

PR #2113, opened 2023-02-22
Update Þ packages

Alfred-Mountfield:
## 🌟 What is the purpose of this PR?

We have published new canary versions of the BP packages. This PR updates the versions and fixes the code.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204024091026379/f) _(internal)_

## 🚫 Blocked by

- [x] #1445 

## 🔍 What does this change?

- Changes all imports from the "@blockprotocol/graph" package to import from the "@blockprotocol/graph/temporal" entrypoint where relevant
- Changes all imports from the "@blockprotocol/graph/stdlib" package to import from the "@blockprotocol/graph/temporal/stdlib" entrypoint where relevant
- Updates imports that have had their names changed like `GraphEmbedderMessageCallbacks`
- Hotfixes a mistake in typings
- Updates messages to match those upstream (create and update now doesn't have `kind` in it)
- Updates the type editor to correctly set "additionalProperties: false"

## 📜 Does this require a change to the docs?

No

## ⚠️ Known issues

N/A

## 🐾 Next steps

- Type editor improvements 

## 🛡 What tests cover this?

All normal ones, `tsc`, `eslint`, playwright, etc.

## ❓ How to test this?

Try the app, ensure it works, ensure CI is green

PR #2117, opened 2023-02-23
Improve type editing (introduce read-only mode, show type versions)

CiaranMn:

UX improvements in the type editor package and the type page in HASH:

1. Type editor implements `readonly` mode
2. HASH can handle versioned URIs for entity types (previously 404'd) – just redirects to latest for now
3. Show version numbers on types in dropdown, and in ontology chips
4. Link types in links table are clickable to go to their URI

## 🔗 Related links

- [Asana task for readonly mode](https://app.asana.com/0/1203179076056209/1203623179643234/f) _(internal)_
- [Slack thread of potential UX improvements](https://hashintel.slack.com/archives/C02LG39FJAU/p1677068435104359) _(internal)_

## 🚫 Blocked by

- [x] Merging of #2113 

## 🔍 What does this change?

- Add a readonly context inside the type editor and use it to disable various hover effects and fields
- Update type editor route in HASH so that it captures versioned URIs
- Allow passing custom routing into type-editor (so that we can use Next routing on links and not require full page navigation)
- Update `OntologyChip` so that it doesn't strip out version information
- Remove no longer unnecessary type assertions

## 🐾 Next steps

**Type editor updates**

1. Enable updating link types
5. Fix updating property types (clicking the button does not load the new property type in)
6. Only show the latest version of an entity type in the dorpdown
7. Publish new version

**blockprotocol.org updates**
1. Add new version of type editor
2. Show that a link entity is a link entity
3. Allow switching an entity to a link entity and vice versa

## ❓ How to test this?

1.  Try a type that should be readonly (not yours), e.g. `/@system-user/types/entity-type/user/v/1`, check it's readonly
2. Try a type that should work as normally (one of yours), check it's editable
3. Try different version URLs, should go to latest

## 📹 Demo

https://user-images.githubusercontent.com/37743469/220891459-72a506fe-9d86-4dc2-a7ea-dec24e65cb91.mp4


nathggns:
Reviewing now

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx
```diff
@@ -41,12 +47,22 @@ export const DefinitionTab = ({
     );
   }, [entityTypesContext.entityTypes]);
 
+  const onNavigate = (url: string) => {
```
Comment (nathggns) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx:50]: Feels like you're replicating a link?

libs/@hashintel/type-editor/src/entity-type-editor/property-list-card.tsx
```diff
@@ -393,7 +407,12 @@ export const PropertyTypeRow = ({
...
-  const array = useWatch({
+  const isRequired = useWatch({
+    control,
+    name: `properties.${propertyIndex}.required`,
+  });
+
+  const isArray = useWatch({
     control,
     name: `properties.${propertyIndex}.array`,
   });
```
nathggns, line 418: You can do this with one `useWatch`, like so

```ts
const [isRequired, isArray] = useWatch({ control, name: [`properties.${propertyIndex}.required`, `properties.${propertyIndex}.array`] })
```

libs/@hashintel/type-editor/src/entity-type-editor/shared/link.tsx
```diff
@@ -0,0 +1,40 @@
+import { HTMLProps, KeyboardEvent, MouseEvent, PropsWithChildren } from "react";
+
+import { useCustomizationSettings } from "../../shared/customization-context";
+
+export const Link = ({
+  children,
+  ...props
+}: PropsWithChildren<HTMLProps<HTMLAnchorElement>>) => {
+  const { onNavigate } = useCustomizationSettings();
```
nathggns, line 9: I don't think I understand the point of this component?

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
```diff
@@ -48,9 +50,14 @@ const Page: NextPageWithLayout = () => {
   const isDraft = !!router.query.draft;
   const { loading: loadingNamespace, routeNamespace } = useRouteNamespace();
 
-  const entityTypeId = router.query["entity-type-id"] as string;
+  const [slug, _, requestedVersion] = router.query["slug-maybe-version"] as [
+    string,
+    "v" | undefined,
+    `${number}` | undefined,
+  ];
```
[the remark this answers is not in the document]
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx]: Yep - done in https://github.com/hashintel/hash/pull/2117/commits/654b242d1beff3aa7f6d15f7f29e66e0a62c68fd

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
```diff
@@ -102,6 +111,28 @@ const Page: NextPageWithLayout = () => {
     [entityType, remotePropertyTypes],
   );
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx]: Nope, thank you – moved in https://github.com/hashintel/hash/pull/2117/commits/32cf493ef485387fe504aae335a9971ad8bdb49a

libs/@hashintel/type-editor/src/entity-type-editor/property-list-card.tsx
CiaranMn: Thank you – done in https://github.com/hashintel/hash/pull/2117/commits/32cf493ef485387fe504aae335a9971ad8bdb49a

libs/@hashintel/type-editor/src/entity-type-editor/shared/link.tsx
CiaranMn: The type editor is a separate package now, which is also used in blockprotocol.org. It therefore can't use special components like `Link` which is inside `frontend`.

To avoid having simple `a` tags which would cause full page refresh, we need some way of users of the component to handle navigation rather than the package. 

I went for this callback rather than passing in a link component, as it seemed more flexible and less of a type issue. I've renamed it and documented it in https://github.com/hashintel/hash/pull/2117/commits/0827b1749bc6c3777530bd636c5c7ba83dce52bc – happy to take other suggestions as to how to deal with this

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx]: See https://github.com/hashintel/hash/pull/2117#discussion_r1115796381

libs/@hashintel/type-editor/src/entity-type-editor/shared/link.tsx
nathggns: I think the better way to handle this is to pass a link component to be honest – can we mark that as a follow up? Means we can just use our normal link components (i.e, `next`'s)

nathggns (approved):

libs/@hashintel/type-editor/src/entity-type-editor/shared/link.tsx
CiaranMn: Yes – I will do this in the next PR I do on the type editor, which will be soon

PR #2121, opened 2023-02-23
Rename `URI` to `URL`

TimDiekmann:

Originally when writing the Type System RFC, we picked "URI" to try and be more agnostic about the resource locator. After some more research it turns out that the distinction between the two terms has been eroded over time and URI just serves to be more confusing when you go and look at the respective standards.

As such we're opting to just settle on URL. This PR attempts to find all occurrences and replaces them, including patching the RFC.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204053055714030/f) _(internal)_
- https://github.com/blockprotocol/blockprotocol/pull/1019

## 🛡 What tests cover this?

- all


Alfred-Mountfield (dismissed):
Tested it after merging #2122 in locally and everything seems to work as expected. Tried quite a few variations of searching for "URI" elsewhere in the code but didn't find anything. 

Great work! Not a massive deal if some slip through, we can always pick them up as we see them

Alfred-Mountfield (approved):

thehabbos007 (approved):

PR #2122, opened 2023-02-23
Hotfix type editor

Alfred-Mountfield:

The type editor was broken with #2113. This fixes it.

## ❓ How to test this?

Run the app and check the type editor can correctly create an entity type


apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page.tsx
```diff
@@ -58,13 +58,21 @@ const Page: NextPageWithLayout = () => {
 
   const draftEntityType = useMemo(() => {
     if (router.query.draft) {
-      // @todo use validation when validateEntityType doesn't return undefined
```
Comment (Alfred-Mountfield) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page.tsx:61]: Not entirely sure why we opted to just not use it, probably an oversight

Comment (Alfred-Mountfield) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[entity-type-id].page.tsx:61]: This would have informed us that our draft entity type was broken
image

Alfred-Mountfield:
/ trunk merge

PR #2123, opened 2023-02-24
Type editor: allow link type upgrading, fix property type upgrading, better option filtering

CiaranMn:

Makes the following improvements to the entity type editor:
1. Allows upgrading a link type, when a newer version is available
2. Fix the same process for property types
3. Improves the filtering of options for property and link type dropdowns so that it:
  - excludes all versions of a property already added (previously wasn't happening for link types)
  - only shows the latest version of a type in the dropdown 

## 🔗 Related links

- [Slack thread suggestions](https://hashintel.slack.com/archives/C02LG39FJAU/p1677068435104359) _(internal)_

## 🔍 What does this change?

- Extract and share type version upgrade indicator from property type to link type row
- Introduce version upgrade to link types
- Fix version upgrade in property types (the change to rely on `useFieldArray`'s return in #1919 meant that the property became stale, because it wasn't being watched – I've moved it back to being selected from a watch)
- New filtering hook for better/consistent filtering of type options for dropdowns
- Move `DisabledCheckbox` into its own component (I started trying to split out the property/link list card components because they each are very long and have multiple components in them, but stopped after one as it was an unnecessary diversion)
- Linting rule to prevent import of `@local/` packages in published packages because I got caught out introducing one.

## 🐾 Next steps

- PR in the `blockprotocol` repo to use this version (which I've already published)

## ❓ How to test this?

1.  Try the type editors and do the things this PR suggests it enables


libs/@hashintel/type-editor/src/entity-type-editor/property-list-card/disabled-checkbox-cell.tsx
```diff
@@ -0,0 +1,55 @@
+import {
```
CiaranMn, line 1: This has been moved only

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card.tsx
```diff
@@ -288,7 +328,7 @@ export const LinkListCard = () => {
             Expected entity types{" "}
             <QuestionIcon tooltip="When specified, only entities whose types are listed in this column will be able to be associated with a link" />
           </TableCell>
-          <EntityTypeTableCenteredCell width={200}>
+          <EntityTypeTableCenteredCell width={210}>
```
CiaranMn, line 331: Needed for blockprotocol

libs/@hashintel/type-editor/src/entity-type-editor/shared/typed-values.ts
```diff
@@ -0,0 +1,31 @@
+type TupleEntry<
```
luisbettencourt, line 1: What's up with this file?

luisbettencourt (approved):
Just one file I'm not sure what its purpose is, but otherwise LGTM

libs/@hashintel/type-editor/src/entity-type-editor/shared/typed-values.ts
CiaranMn: Needed for typedEntries in the option filtering hook

luisbettencourt (changes requested):
Actually, after further testing I noticed that after refeshing the type editor the properties and links disappear

CiaranMn:
> Actually, after further testing I noticed that after refeshing the type editor the properties and links disappear.

@luisbettencourt can you share exact reproduction steps? I can't recreate this. Can you run `yarn` in the root and make sure the `react-hook-form` is being applied please?

luisbettencourt (approved):
It was probably the patch not being applied correclty. Thank you for fixing the entities tab aswell!

PR #2125, opened 2023-02-25
Update to new versions of the Block Protocol packages

Alfred-Mountfield:

We've published a few new versions of the Block Protocol packages, most importantly there were 2 breaking changes
- https://github.com/blockprotocol/blockprotocol/pull/1039
- https://github.com/blockprotocol/blockprotocol/pull/1041

This PR implements both of those

## 🔍 What does this change?

- Updates versions
- Renames `aggregate*` things to `query*`
- Removes `additionalProperties` logic from type creation
- Renames the `getAllLatest` queries (which have been confusing for a while) so that they align with their BP equivalent

I don't believe we have docs affected by this

## ⚠️ Known issues

- The sandbox component seems very out of date, I fixed linting in it but I suspect if we try putting it in the app it is going to be very broken
- We are using a hard-coded URL for the primitive link type on blockprotocol.org, until `0.3` is merged this requires us to keep it up to date with changes. All deployments of HASH (including local development) will be broken once https://github.com/blockprotocol/blockprotocol/pull/1042 goes in, until this PR is merged.

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

All normal ones

## ❓ How to test this?

Try the app as usual, ensure linting passes, etc. This shouldn't be an end-user visible change


TimDiekmann (commented):
✅ Rust changes are looking good

apps/hash-api/src/graphql/type-defs/knowledge/entity.typedef.ts
```diff
@@ -43,7 +43,7 @@ export const entityTypedef = gql`
     """
     Get a subgraph rooted at all entities at their latest version.
```
thehabbos007, line 44: Likely not a high priority right now, but we should probably update the docs for these and the ones for ontology types

libs/@local/hash-isomorphic-utils/src/blocks.ts
```diff
@@ -124,7 +125,10 @@ const transformBlockConfig = ({
     variants,
     icon: deriveAbsoluteUrl({ baseUrl, path: metadata.image }),
     image: deriveAbsoluteUrl({ baseUrl, path: metadata.icon }),
```
thehabbos007, line 127: Is this an unrelated mistake? :D
```suggestion
    icon: deriveAbsoluteUrl({ baseUrl, path: metadata.icon }),
    image: deriveAbsoluteUrl({ baseUrl, path: metadata.image }),
```

thehabbos007 (dismissed):
Thank you, Alfie! The commit-separation made this very pleasant to review. I have a couple comments, but mostly nits :purple_heart: 

I've tested locally and the type/entity editors seem to work on my end.

libs/@local/hash-isomorphic-utils/src/blocks.ts
Alfred-Mountfield: That does look like a bug to me haha
Fix presumed bug

apps/hash-api/src/graphql/type-defs/knowledge/entity.typedef.ts
Alfred-Mountfield: [Update docs](https://github.com/hashintel/hash/pull/2125/commits/d909a3429cc06d46a218918df4e93b84cb461cfc)

thehabbos007 (approved):

PR #2126, opened 2023-02-25
Allow running external services in "offline" mode

TimDiekmann:

When developing locally it's annoying that an internet connection is required to use the type fetcher. The type fetcher has to pull data from `blockprotocol.org` for primitive data types and entity types.

This PR adds the option to run external services without the type-fetcher, which currently falls back to the stopgap solution we still have in the Graph

## 🔗 Related links

- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1677348564924359) _(internal)_

## 🔍 What does this change?

- Move the `type-fetcher` to another compose file
- Use the above compose file as an override configuration for the default behavior and adjust `package.json` to pick that up automatically
- Add a configuration to not run the type fetcher at all and make this accessible with `yarn external-services-offline`

A small section to the readme was added

## ⚠️ Known issues

This is only a hotfix. To properly support an offline mode, this requires more programming and design work.

## ❓ How to test this?

Test that `yarn external-services up --wait --build` does work as before
Test that `yarn external-services-offline up --wait --build` does work without starting the `type-fetcher` service


indietyp (commented):
looks good so far, but be aware that the `deploy` parameter is completely optional, and some installations may not support it (https://docs.docker.com/compose/compose-file/deploy/). I would recommend basing this on an environment variable and creating a busy box if not set. This would also remove the need for an additional `docker-compose.yml` file.

> Compose Specification is extended to support an OPTIONAL deploy subsection on services [^1]

[^1]: https://docs.docker.com/compose/compose-file/deploy/#definitions

TimDiekmann:
Thanks, @indietyp!

> I would recommend basing this on an environment variable and creating a busy box if not set

How do you conditionally use another image and omit the `build` command in `type-fetcher`? Also, this still has the overhead of a busybox, but in comparison to the other containers this is most probably negligible.

indietyp:
I don't know how easy it is, but instead, modify the Dockerfile for the type-fetcher:

```dockerfile
ARG TYPE_FETCHER_RUN="yes";

// https://stackoverflow.com/a/55734437 about running nothing in a container
RUN if [ $(TYPE_FETCHER_RUN) = "yes" ]; then <run type-fetcher>; else exec /bin/bash -c "trap : TERM INT; sleep infinity & wait"; fi
```

In the docker-compose.yml file, one can then:

```yaml
args:
	TYPE_FETCHER: ${TYPE_FETCHER_RUN}

// ... type-fetcher
build:
	TYPE_FETCHER_RUN: ${TYPE_FETCHER_RUN:-yes}
```

The healthcheck would need to be substituted via an alias in the dockerfile, with the following contents:

```sh
if [$TYPE_FETCHER_RUN = "yes"]; then
	// execute type-fetcher
fi
```

TimDiekmann:
This requires another image as well, currently, we use a `scratch` image. Otherwise, we could simply run `true`, exit the container, and change the check to `exit succesfully`

indietyp:
AFAIK docker-compose supports all POSIX substitutions so something similar to down below should work: 

`${TYPE_FETCHER_OFFLINE:+service_healthy}${TYPE_FETCHER_OFFLINE:-service_completed_successfully}` could be used.

`:+` sets the value to `service_healthy` if `TYPE_FETCHER_OFFLINE` has any value
`:-` sets the value to `service_completed_successfully` if `TYPE_FETCHER_OFFLINE` is either empty or unset.

In that case the definition of `TYPE_FETCHER` on the hash-graph service should be changed to something similar.

TimDiekmann:
I appreciate your effort to avoid the additional configuration file, but we want a `service_healthy` condition set if and only if the type fetcher is enabled. Thanks for the explanation, but I know how variable substitution is working, it just does not work in the `depends_on` section as it either requires the user to specify `service_completed_successfully` as an argument (which is a pretty unexpected value for something like `HASH_GRAPH_TYPE_FETCHER_ENABLED`) or otherwise it will fail with
> condition must be one of the following: "service_started", "service_healthy", "service_completed_successfully"

Also, this does not solve the issue that the type fetcher should not run at all, and erroring as the subcommand is not found is not an option here. It's required that the type fetcher has a `build` option to ensure it's using the correct image, so specifying another container will not work (also, this implies, that the service will be started regardless, which I try to avoid).

One possible solution is to completely move out the `type-fetcher` of the compose file and only add it in the `docker-compose.default.yml`, this at least avoids the `replicas` part. But generally, `docker-compose.yml` files should only contain the services, which should run regardless, conditionally enabled services should be defined in other files.

indietyp (commented):
I like this a lot more!

(Edit because previous statement was flat out wrong)

apps/hash-external-services/docker-compose.type-fetcher.yml
```diff
@@ -0,0 +1,40 @@
...
+      timeout: 2s
+      retries: 10
+
+  graph:
+    depends_on:
+      type-fetcher:
+        condition: service_healthy
+    build:
+      args:
+        TYPE_FETCHER: yes
```
thehabbos007, line 40: Would this override the existing
```
     args:
        PROFILE: dev # <-- this
        TYPE_FETCHER: no
```
In case it would, we should add it here too

thehabbos007 (commented):
Looks good to me, works locally as well. Thank you, Tim! Just have one comment, perhaps we should just address that pre-approval

apps/hash-external-services/docker-compose.type-fetcher.yml
TimDiekmann: We talked about this offline and decided to not put it here. We will generally revisit the docker-compose files at some point and base the `.prod` file on a base file, so we would overwrite it there.

thehabbos007 (approved):

PR #2129, opened 2023-02-27
Update to latest BP Canary version and ensure `$schema` is used in ontology types

TimDiekmann:

We require `$schema` to be present in ontology types. This PR updates the HASH workspace to account for these changes

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204065621964625/f) _(internal)_


apps/hash-api/src/graph/ontology/primitive/data-type.ts
```diff
@@ -54,7 +51,13 @@ export const createDataType: ImpureGraphFunction<
     kind: "data-type",
     title: params.schema.title,
   });
-  const schema = { $id: dataTypeUrl, ...params.schema };
+  const schema = {
+    $schema:
+      "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type" as const,
```
Alfred-Mountfield, line 56: Can you replace these with imports of `DATA_TYPE_META_SCHEMA` and such from the exports of the `@blockprotocol/graph` or `@blockprotocol/type-system` package (preferably graph if it's already a dependency)

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
```diff
@@ -50,7 +50,13 @@ export const createEntityType: ImpureGraphFunction<
...
   });
 
-  const schema = { $id: entityTypeId, ...params.schema };
+  const schema = {
+    $schema:
+      "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type" as const,
+    kind: "entityType" as const,
+    $id: entityTypeId,
+    ...params.schema,
+  };
```
Alfred-Mountfield, line 59: For these I think you can just do this
```suggestion
  const schema = {
    $schema:
      "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
    kind: "entityType",
    $id: entityTypeId,
    ...params.schema,
  } as const;
```

If that does work It'd be nice if it was applied elsewhere, it's just a bit cleaner. (Not blocking the PR though)

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
```diff
@@ -128,7 +134,12 @@ export const updateEntityType: ImpureGraphFunction<
   const updateArguments: UpdateEntityTypeRequest = {
     actorId,
     typeToUpdate: entityTypeId,
-    schema,
+    schema: {
+      kind: "entityType",
+      $schema:
```
Alfred-Mountfield: Ridiculously nit-picky suggestion (feel free to ignore), would be nice to try and establish a consistent ordering for these fields. We've been going with `$schema`, `kind`, `$id`, `title`, ...

Only worth addressing if you're already going to touch these files, 100% not blocking

apps/hash-api/src/graph/ontology/primitive/data-type.ts
TimDiekmann: It only suggested imports from `type-system`: https://github.com/hashintel/hash/pull/2129/commits/0241c04ec2c6616e744b9b2382a31e18fd9a3e44

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
TimDiekmann: I tried but obviously missed a few. I will go through the changed files again and apply the order.

apps/hash-api/src/graph/ontology/primitive/data-type.ts
Alfred-Mountfield: I think it probably did that because there was already an import from the `type-system`. It'd be good to unify all of our imports at some point so we're not importing from both packages, no worries though

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
TimDiekmann: With https://github.com/hashintel/hash/pull/2129/commits/0241c04ec2c6616e744b9b2382a31e18fd9a3e44 it removes the `const` from `$schema`. Do we still want the `const` at the end? I've seen both approaches in the code and found this more clearer.

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
Alfred-Mountfield: No need! Really not important

apps/hash-api/src/graph/ontology/primitive/entity-type.ts
Alfred-Mountfield: I don't mind, let's just stick with what's already implemented

Alfred-Mountfield (approved):

PR #2145, opened 2023-03-01
Improve various block docs and metadata

Alfred-Mountfield:

This updates the `license` fields in `package.json`s to be consistent, also adds missing `repository` and/or `keywords` fields.

It also improves the `README.md`s of some of the blocks, as they were either outdated or a bit unclear.

### DIFF
--- a/apps/hashdotdev/package.json
+++ b/apps/hashdotdev/package.json
@@ -2,6 +2,12 @@
   "name": "@apps/hashdotdev",
   "version": "0.0.0-private",
   "private": true,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hashintel/hash.git#main",
+    "directory": "@apps/hashdotdev"
+  },
+  "license": "(MIT OR Apache-2.0)",
   "scripts": {
     "build": "next build",
     "dev": "next dev",
--- a/apps/hashdotdev/package.json
+++ b/apps/hashdotdev/package.json
@@ -5,7 +5,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/hashintel/hash.git#main",
-    "directory": "@apps/hashdotdev"
+    "directory": "apps/hashdotdev"
   },
   "license": "(MIT OR Apache-2.0)",
   "scripts": {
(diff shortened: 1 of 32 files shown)


blocks/divider/README.md
```diff
@@ -1,5 +1,12 @@
-The Divider block is an `<hr>` element which can be customized with [`https://blockprotocol.org/@blockprotocol/types/property-type/height-in-pixels/`](https://blockprotocol.org/@blockprotocol/types/property-type/height-in-pixels/) and [`https://blockprotocol.org/@blockprotocol/types/property-type/css-background-color/`](https://blockprotocol.org/@blockprotocol/types/property-type/css-background-color/) properties.
+The Divider block is a customizable wrapper around an `<hr>` element.
 
-Its width is set to 100% and its border to "none", allowing the [`https://blockprotocol.org/@blockprotocol/types/property-type/height-in-pixels/`](https://blockprotocol.org/@blockprotocol/types/property-type/height-in-pixels/) property to control the line's thickness. The [`https://blockprotocol.org/@blockprotocol/types/property-type/css-background-color/`](https://blockprotocol.org/@blockprotocol/types/property-type/css-background-color/) property sets the element's background color.
+## Programmatic Usage
```
Alfred-Mountfield, line 3: I'm unsure what the hub is going to do with these headings, if it does go wrong, it feels like something we should be able to deal with.

nonparibus: The new Figma designs for the Þ Hub specifically show how we want to handled these styled headers, so I suggest we leave them in for now (since the current experience isn't _dire_).

Alfred-Mountfield: What I meant is I'm not sure how the README rendering component is going to do when it sees high-level headings (like `h1` or `h2`).

Unless we patch them, or override their styling to be smaller, it might be quite strange

apps/hashdotdev/package.json
```diff
@@ -2,6 +2,12 @@
   "name": "@apps/hashdotdev",
   "version": "0.0.0-private",
   "private": true,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hashintel/hash.git#main",
+    "directory": "@apps/hashdotdev"
```
nonparibus, line 8:
```suggestion
    "directory": "apps/hashdotdev"
```

blocks/divider/README.md
Alfred-Mountfield: I'm sure it won't be too hard to fix, and should hopefully be relatively quick to hotfix if needs be, just saying the hub _might_ look kinda terrible until we patch it (if there's a problem), and we should check when we publish the blocks

PR #2150, opened 2023-03-02
Address Block fixes

luisbettencourt:

This PR makes some fixes to the Address Block

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204086167143604/f) _(internal)_

## 🔍 What does this change?

- Hide outline when input is focused;
- Autofocus input.

### DIFF
--- a/blocks/address/src/app.tsx
+++ b/blocks/address/src/app.tsx
@@ -520,6 +520,7 @@ export const App: BlockComponent<RootEntity> = ({
                       return (
                         <TextField
                           {...params}
+                          autoFocus
                           placeholder={
                             isMobile
                               ? "Enter an address"
@@ -611,6 +612,7 @@ export const App: BlockComponent<RootEntity> = ({
                       width: 1,
                       [`.${autocompleteClasses.input}`]: {
                         paddingLeft: "0 !important",
+                        outline: "none",
                       },
                       [`.${autocompleteClasses.inputRoot}`]: {
                         paddingX: ({ spacing }) =>


blocks/address/src/app.tsx
```diff
@@ -516,10 +544,21 @@ export const App: BlockComponent<RootEntity> = ({
...
+                     * popup.
+                     *
+                     * @see LinkEntityTypeSelector
+                     */
+                    ref={setAnchorEl}
                     renderInput={({ InputProps, ...params }) => {
                       return (
                         <TextField
                           {...params}
+                          autoFocus
```
benwerner01, line 561: I just want to double check, do we actually want this? Wouldn't this mean that if this block were in a page with other blocks, there'd be a race between which block gets focused first?

blocks/address/src/app.tsx
```diff
@@ -545,10 +584,32 @@ export const App: BlockComponent<RootEntity> = ({
...
+                                  border: `1px solid ${theme.palette.gray[30]} !important`,
+                                },
+                              }),
+                              ...(open && suggestions.length
+                                ? [
+                                    popperPlacementInputNoRadius,
+                                    popperPlacementInputNoBorder,
+                                    {
+                                      borderRadiusBottomLeft: "0 !important",
+                                      borderRadiusBottomRight: "0 !important",
```
benwerner01, line 608: Ideally we would reframe from using `!important` in favour of specificity, did you try using the relevant MUI classname to achieve the override using specificity? Or is there some other issue you ran into

libs/@hashintel/design-system/src/autocomplete-dropdown.tsx
```diff
@@ -37,50 +38,53 @@ export const AutocompleteDropdown = ({
...
+                backgroundColor: `${theme.palette.blue[20]} !important`,
+              },
+
+              "&.Mui-focused": {
+                backgroundColor: `${theme.palette.gray[10]} !important`,
+
+                [`&[aria-selected="true"]`]: {
+                  backgroundColor: `${theme.palette.gray[20]} !important`,
+                },
               },
```
benwerner01, line 83: Similar comment, why the need for using `!important`?

blocks/address/src/app.tsx
```diff
@@ -499,8 +520,15 @@ export const App: BlockComponent<RootEntity> = ({
                   }
                 }}
               >
-                <Box sx={{ display: "flex", gap: 1.5 }}>
+                <Box sx={{ display: "flex", gap: 1.5, maxWidth: 420 }}>
```
benwerner01, line 523: Nit: would be nice to capture the max width of the input in a named variable, and make use of it here

blocks/address/src/app.tsx
```diff
@@ -410,7 +431,7 @@ export const App: BlockComponent<RootEntity> = ({
                 <Box sx={{ display: "flex", columnGap: 3, flexWrap: "wrap" }}>
                   <Link
                     //  @todo: link this to the block's hub page
-                    href=""
+                    href="https://blockprotocol.org/@hash/blocks/address"
```
benwerner01, line 434: We can remove the `@todo` now, right?

libs/@hashintel/design-system/src/autocomplete-dropdown.tsx
```diff
@@ -37,50 +38,53 @@ export const AutocompleteDropdown = ({
...
-            [`&[aria-selected="true"]`]: {
-              backgroundColor: `${theme.palette.blue[20]} !important`,
+        sx={[
+          (theme) => ({
+            p: 1,
+            border: 1,
+            boxSizing: "border-box",
+            borderColor: theme.palette.gray[30],
+            boxShadow: "none",
+            [`${popperPlacementSelectors.top} &`]: {
```
benwerner01, line 48: For my understanding, what does the `&` do here?

benwerner01 (commented):
Nice work Luis, the changes are working for me locally - and the address block is looking really good! Had a few questions/thoughts, let me know what you think.

benwerner01:
@luisbettencourt this wasn't introduced in this PR, so could be done in follow-up, but just wanted to flag that whilst testing I noticed that the block is sending a `suggestAddress` request when `searchText === ""`, which throws a "searchText is required" error in the mapbox API. We should therefore add a check that `searchText !== ""` somewhere before sending the request.

image

To reproduce type some text in the input, and then delete it.

blocks/address/src/app.tsx
luisbettencourt: This will be moved to the design-system in the follow-up PR, so I suggest we leave it for now

libs/@hashintel/design-system/src/autocomplete-dropdown.tsx
luisbettencourt: Same as above

libs/@hashintel/design-system/src/autocomplete-dropdown.tsx
luisbettencourt: I wasn't sure myself, as this was taken from the AutocompleteDropdown in the design-system. But [here](https://sass-lang.com/documentation/style-rules/parent-selector)'s what I found

blocks/address/src/app.tsx
luisbettencourt: Thanks for catching 0f6baee8618547a9897638bde3c07ec425be07ba

blocks/address/src/app.tsx
luisbettencourt: Done in 59259c798be853ed9062b9068a9602ef82f39dab

blocks/address/src/app.tsx
luisbettencourt: Yeah, thanks for checking on this. It's the desired behaviour, I think ([slack thread](https://hashintel.slack.com/archives/C03FRBNFM7T/p1677688755039609?thread_ts=1677688731.958359&cid=C03FRBNFM7T)). The openAI blocks do it with a `useEffect` that focuses the input

benwerner01: Interesting, have we thought about the implications of having multiple blocks with autofocus inputs in the same page? If not perhaps worth flagging in that thread. Won't block the PR on this.

luisbettencourt:
Created a [task](https://app.asana.com/0/1203623179643287/1204105361082361/f) to track that issue, and will address it in a follow-up PR

benwerner01 (approved):

benwerner01 (approved):

PR #2153, opened 2023-03-02
Resolve only one type and remove the exhaustive variant in the type fetcher

TimDiekmann:

Currently, we pass in a single URI to the type fetcher and this is used to traverse the whole dependency map. To avoid fetching types, which already were added to the database and avoid querying HASH types, we should fetch a single ontology type instead.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1203958456055086/f) _(internal)_

## 🔍 What does this change?

- Allow the type fetcher to take a list of URLs to resolve
- Only resolve the URLs provided to the type fetcher
- Move logic to traverse ontology types to the graph


apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
```diff
@@ -74,7 +74,7 @@ pub async fn type_fetcher(args: TypeFetcherArgs) -> Result<(), GraphError> {
         .filter_map(|r| future::ready(r.ok()))
         .map(server::BaseChannel::with_defaults)
         .map(|channel| {
-            let server = FetchServer;
+            let server = FetchServer { buffer_size: 10 };
```
TimDiekmann, line 77: We could make this a CLI parameter

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
```diff
@@ -74,7 +74,7 @@ pub async fn type_fetcher(args: TypeFetcherArgs) -> Result<(), GraphError> {
...
             channel.execute(server.serve())
         })
         .buffer_unordered(255)
```
TimDiekmann, line 80: If we add a parameter for the above, we should also do this here

apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -1,18 +1,17 @@
...
+    collections::{HashMap, HashSet},
+    iter::once,
+    mem,
+};
 
 use async_trait::async_trait;
 use error_stack::{IntoReport, Result, ResultExt};
 use tarpc::context;
 use tokio::net::ToSocketAddrs;
-#[cfg(feature = "type-fetcher")]
```
TimDiekmann, line 7: The module is gated behind this flag as well, there is no need to repeat it.

apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -196,106 +203,113 @@ where
...
-                .fetch_ontology_type_exhaustive(context::current(), reference.url().clone())
+        ontology_type_references: Vec<VersionedUrl>,
+        actor_id: UpdatedById,
+    ) -> Result<FetchedOntologyTypes, StoreError> {
+        let provenance_metadata = ProvenanceMetadata::new(actor_id);
+
+        let mut queue = ontology_type_references;
+        let mut seen = queue.iter().cloned().collect::<HashSet<_>>();
+
+        let mut fetched_ontology_types = FetchedOntologyTypes::default();
```
TimDiekmann, line 214: Previously, we used `stream::unfold`, but I chose the manual approach here as this way I can directly add them to the three different sets for ontology types, `unfold` only supports a single collection. We could implement the additional traits for `FetchedOntologyTypes`, but as it's only used here it's not worth the effort.

apps/hash-graph/lib/graph/src/ontology.rs
```diff
@@ -157,9 +177,26 @@ impl OntologyType for EntityType {
...
+            )
+            .chain(self.link_mappings().into_iter().flat_map(
+                |(link_entity_type, destination_entity_type_constraint)| {
+                    {
+                        once(link_entity_type)
+                            .chain(destination_entity_type_constraint.unwrap_or_default())
+                    }
+                    .map(OntologyTypeReference::EntityTypeReference)
+                },
+            ))
```
TimDiekmann, line 198: This implementation avoids the extra allocation we did before. `<&[]>::default()` returns an empty slice

thehabbos007 (approved):
Thank you a lot, Tim! This is great :)

I tested this locally by creating an entity with an external type, works as expected/before. I don't think we need to add the buffer size CLI params yet. Great stuff!

apps/hash-graph/lib/graph/src/ontology.rs
TimDiekmann: `destination_entity_type_constraint` is of type `Option<&[]>`. Calling `.unwrap_or_default` will either return the slice or an empty slice.

TimDiekmann: Previously we used a temporary vector

PR #2154, opened 2023-03-02
Fix Timestamps on ontology types not being updated on `update`

TimDiekmann:

When calling update on an ontology type the timestamp is not updated.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204091652018645/f) _(internal)_

## 🚫 Blocked by

- This PR is not blocked on anything

## 🔍 What does this change?

Fix that the timestamp is being set on the update method

This is a feature, which is currently not exposed at all, thus, it does not need a change for documentation

## ⚠️ Known issues

We should investigate how we can fix these kind of bugs later when we cannot simply change old migration files

## 🐾 Next steps

After this PR has been merged we need to redeploy the Backend

## 🛡 What tests cover this?

This is a feature, which is currently not exposed, so there is no way to automatically test it

## ❓ How to test this?

1.  Run the REST API tests
1.  open the Database
1.  make sure, that the timestamps on the same type differ between their versions

## 📹 Demo

image

### DIFF
--- a/apps/hash-graph/postgres_migrations/V4__ontology_functions.sql
+++ b/apps/hash-graph/postgres_migrations/V4__ontology_functions.sql
@@ -37,7 +37,8 @@ BEGIN
   SET
     "ontology_id" = update_ontology_id.ontology_id,
     "version" = update_ontology_id.version,
-    "record_created_by_id" = update_ontology_id.record_created_by_id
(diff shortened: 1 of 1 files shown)


thehabbos007:
I suppose we want to merge this just before we wipe/redeploy next time?

thehabbos007 (approved):

PR #2156, opened 2023-03-03
`error-stack`: improve formatting docs

indietyp:

This adds additional documentation on formatting in `error-stack` through a new section: `Implementation Details`. This is part of a larger effort to reduce the accumulated tech debt in the `fmt` module by:

1) reduce allocations
    1.1) use of eager evaluation and use of `core::fmt` instead of requiring `Into<String>`
2) improve (internal) documentation
3) remove redundant code

This chain of PRs should have *no* user-facing changes, except documentation. Formatting will stay the same.

### DIFF
--- a/libs/error-stack/src/fmt.rs
+++ b/libs/error-stack/src/fmt.rs
@@ -204,10 +204,10 @@
 //! parent, we travel up the tree until we encounter our first [`Context`] node. Groups always
 //! contain lists, for the sake of clarity this explanation only shows the first element.
 //!
-//! Additional rules that apply:
+//! The rules stated above also derive some additional rules:
 //! * lists are never empty
 //! * lists are nested in groups
-//! * groups are always preceded by lists
+//! * groups are always preceded by a single list
 //! * groups are ordered left to right
 //!
 //! Using the aforementioned delimiters for lists and groups the end result would be:
--- a/libs/error-stack/src/fmt.rs
+++ b/libs/error-stack/src/fmt.rs
@@ -198,7 +198,7 @@
 //! more than `1` child, this means that `(2, 6)` is a group (they share `1` as an immediate context
 //! parent), while `(3, 4, 6)` is not. `(3, 4, 6)` share the same parent with more than 1 child
 //! (`1`), but `1` is not the immediate context parent of `3` and `4` (`2`) is. In the more detailed
-//! example `(Dᶜ, Hᶜ, Cᶜ)` is considered a group because they share the same *immediate* context
+//! example `(Dᶜ, Hᶜ, Iᶜ)` is considered a group because they share the same *immediate* context
 //! parent `Aᶜ`, important to note is that we only refer to immediate context parents, `Fᵃ` is the
 //! immediate parent of `Iᶜ`, but is not a [`Context`], therefore to find the immediate context
 //! parent, we travel up the tree until we encounter our first [`Context`] node. Groups always
--- a/libs/error-stack/src/fmt.rs
+++ b/libs/error-stack/src/fmt.rs
@@ -215,7 +215,9 @@
 //! Overview Tree: `[0, 1] ([2] ([3], [4, 5]), [6, 7, 8])`
 //! Detailed Tree: `[Aᶜ] ([Dᶜ], [Hᶜ], [Iᶜ])`
 //!
-//! Attachments are not ordered by insertion order but by depth in the tree. Each context uses the
+//! Attachments are not ordered by insertion order but by depth in the tree. The depth in the tree
+//! is the inverse of the insertion order, this means that the [`Debug`] output of all
+//! attachments is reversed from the calling order of [`Report::attach`]. Each context uses the
 //! attachments that are it's parents until the next context node. If attachments are shared between
 //! multiple contexts, they are duplicated and output twice.
 //!
@@ -269,11 +271,13 @@
 //!     ╰─▶ 8
 //! ```
 //!
-//! Attachments have been added to various places to simulate a real usecase with attachment and
-//! visualise their placement.
+//! Attachments have been added to various places to simulate a real use-case with attachments and
+//! to visualise their placement.
 //!
-//! The spacing and characters used are chosen carefully, as even when you have a context vs. group
-//! of contexts (as seen above) alignment is still ensured.
+//! The spacing and characters used are chosen carefully, to reduce indentation and increase visual
+//! legibility in large trees. The indentation of the group following the last entry in the
+//! preceding list is the same. To indicate that the last entry in the preceding list is the parent
+//! a new indentation of the connecting line is used.
 //!
 //! [`Display`]: core::fmt::Display
 //! [`Debug`]: core::fmt::Debug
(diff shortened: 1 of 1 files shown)


libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//! A list is a list of nodes where each node in the list if the parent of the following element and
+//! has only a single child, the last element of a list, can have `0..n` children. In the examples
+//! above, `[6, 7, 8]` is considered a list, while `[1, 6]` is not, because while `1` is a parent of
+//! `6`, `1` has more than 1 child.
+//!
+//! A group is a list of nodes where each node shares a common immediate context parent that has
+//! more than `1` child, this means that `(2, 6)` is a group (they share `1` as an immediate context
+//! parent), while `(3, 4, 6)` is not. `(3, 4, 6)` share the same parent with more than 1 child
+//! (`1`), but `1` is not the immediate context parent of `3` and `4` (`2`) is. In the more detailed
+//! example `(Dᶜ, Hᶜ, Cᶜ)` is considered a group because they share the same *immediate* context
```
TimDiekmann, line 201:
```suggestion
//! example `(Dᶜ, Hᶜ, Iᶜ)` is considered a group because they share the same *immediate* context
```

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//! parent `Aᶜ`, important to note is that we only refer to immediate context parents, `Fᵃ` is the
+//! immediate parent of `Iᶜ`, but is not a [`Context`], therefore to find the immediate context
+//! parent, we travel up the tree until we encounter our first [`Context`] node. Groups always
+//! contain lists, for the sake of clarity this explanation only shows the first element.
+//!
+//! Additional rules that apply:
```
TimDiekmann, line 207: Are these rules you have set?

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//! * lists are never empty
+//! * lists are nested in groups
+//! * groups are always preceded by lists
```
TimDiekmann, line 210: What does this mean? Every context is in exactly one list?

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//! * groups are ordered left to right
+//!
+//! Using the aforementioned delimiters for lists and groups the end result would be:
+//!
+//! Overview Tree: `[0, 1] ([2] ([3], [4, 5]), [6, 7, 8])`
+//! Detailed Tree: `[Aᶜ] ([Dᶜ], [Hᶜ], [Iᶜ])`
+//!
+//! Attachments are not ordered by insertion order but by depth in the tree. Each context uses the
```
TimDiekmann, line 218: Isn't the depth in the tree determined by the insertion order?

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//! heading of the whole list, while all other contexts are intended. The last entry in that
+//! indentation is (if present) the group that follows, taking the detailed example this means that
+//! the following output would be rendered:
+//!
+//! ```text
+//! Aᶜ
+//! │
+//! ╰┬▶ Dᶜ
+//!  │  ├╴Bᵃ
+//!  │  ╰╴Cᵃ
```
TimDiekmann, line 234: It would be good if you could describe why the attachments here are in reversed order. As far as I can see, we created an error with `D` as context. Then we attached `C` and `B` (in that order). After that, we changed the context to `A`

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,148 @@
...
+//!     ├─▶ 7
+//!     │
+//!     ╰─▶ 8
+//! ```
+//!
+//! Attachments have been added to various places to simulate a real usecase with attachment and
+//! visualise their placement.
+//!
+//! The spacing and characters used are chosen carefully, as even when you have a context vs. group
+//! of contexts (as seen above) alignment is still ensured.
```
TimDiekmann, line 276: I think this is not the case. `1` is the parent of the `(2, 3)` group but they are rendered in the same column.

TimDiekmann (commented):
Thanks for writing this down, this makes formatting way easier to understand!

I have a few commends and clarification questions.

libs/error-stack/src/fmt.rs
indietyp: Will clarify, yes the depth in the tree is the inverse of the insertion order, as we "build the tree up".

indietyp: clarified in [c27f8c6](https://github.com/hashintel/hash/pull/2156/commits/c27f8c6658293e871570c07e9948f9c70ed634f5)

libs/error-stack/src/fmt.rs
indietyp: clarified in preceding paragraph in [c27f8c6](https://github.com/hashintel/hash/pull/2156/commits/c27f8c6658293e871570c07e9948f9c70ed634f5)

libs/error-stack/src/fmt.rs
indietyp: These are rules that are derived from the rules stated above, clarified in [3a68ec5](https://github.com/hashintel/hash/pull/2156/commits/3a68ec5e73d3dd9fc8f3fa7a3777f62e2852bdde)

libs/error-stack/src/fmt.rs
indietyp: hopefully clarified in [3a68ec5](https://github.com/hashintel/hash/pull/2156/commits/3a68ec5e73d3dd9fc8f3fa7a3777f62e2852bdde), it just means that there's always at least a single context before a group, as otherwise, a group can never happen. The only place where this might not be the case is the top level, where we render them out as two separate trees.

indietyp: additional clarification in [288470f](https://github.com/hashintel/hash/pull/2156/commits/288470f80eba6aab6bb3b3ffdb3e6ffafa89eee7)

libs/error-stack/src/fmt.rs
```diff
@@ -133,6 +133,155 @@
...
+//! Detailed Tree: `[Aᶜ] ([Dᶜ], [Hᶜ], [Iᶜ])`
+//!
+//! Attachments are not ordered by insertion order but by depth in the tree. The depth in the tree
+//! is the inverse of the insertion order, this means that the [`Debug`] output of all
+//! attachments is reversed from the calling order of [`Report::attach`]. Each context uses the
+//! attachments that are it's parents until the next context node. If attachments are shared between
+//! multiple contexts, they are duplicated and output twice.
+//!
+//! Groups are always preceded by a single list, the only case where this is not true is at the top
+//! level, in that case we opt to output separate trees for each member in the group.
```
TimDiekmann, line 225: Thanks for adding this as well!

TimDiekmann (approved):
Sorry, forgot to approve 🙃

PR #2157, opened 2023-03-03
`deer`: introduce `FieldAccess`

indietyp:

This introduces a new API, `FieldAccess`, which is conjunction with `ObjectAccess` and can be used to conditionally get the value depending on the underlying type. This enables `Enum` style access, e.g. for `Result`.

### Differences between `serde` and `deer`

Because we decided to have a minimal API surface and unify the access through `Object` vs. `Array` we have a fundamental problem: conditional values based on the key, this is best shown with `Result`, `Result<T, E>` has two variants: `Ok` and `Err` and depending on the key. This is now possible, by implementing something similar to:

```rust
enum Key {
	Ok,
	Err
}

/* impl Deserialize for Key */

pub trait FieldAccess<'de> {
	type Key: Key;
    type Value: Result<T, E>;

    fn value(&self, key: &Key, deserializer: D) -> Result<Self::Value, FieldAccessError> {
       match key {
          Key::Ok => T::deserialize(deserializer).map(Ok).change_context(FieldAccessError),
          Key::Err => ...
       }
    }
}
```

in `serde` this is implemented through the `EnumAccess` in the `visit_enum` variant, which is a lot more complicated, otherwise one can use `next_key` and `next_value`. The problem with those is that they are prone to abuse, there is no type-system contract that `next_value` must always follow `next_key`, hand that they are only used in combination. With `FieldAccess` this is not possible, because `value()` required `&Key` and therefore can only be called after `Key`

### Downsides

Because `next()` now defaults to `field()` and the `field()` value depends on the key we cannot probe the value for an error if the key errored out. Previously we got the error from both.

### 👂 Open Questions

Does `FieldAccess::Value` and `FieldAccess::Key` need to be `Deserialize<'de>`? Because `FieldAccess::Value` is not used in a default implementation we could relax the bound, but I do not know if we want to keep the bound, just to ensure that a user does not misuse the API, for `FieldAccess::Key` the bound is needed for the default implementation, which I think is fine. @TimDiekmann , @thehabbos007 , what do you think?

### DIFF
--- a/libs/deer/desert/src/object.rs
+++ b/libs/deer/desert/src/object.rs
@@ -3,7 +3,7 @@ use deer::{
         BoundedContractViolationError, ExpectedLength, ObjectAccessError, ObjectLengthError,
         ReceivedLength, Variant,
     },
-    Deserialize, Deserializer as _,
+    Deserialize, Deserializer as _, FieldAccess,
 };
 use error_stack::{Report, Result, ResultExt};
 
@@ -180,10 +180,9 @@ impl<'de> deer::ObjectAccess<'de> for ObjectAccess<'_, '_, 'de> {
         .change_context(ObjectAccessError)
     }
 
-    fn next<K, V>(&mut self) -> Option<Result<(K, V), ObjectAccessError>>
+    fn field<F>(&mut self, access: F) -> Option<Result<(F::Key, F::Value), ObjectAccessError>>
     where
-        K: Deserialize<'de>,
-        V: Deserialize<'de>,
+        F: FieldAccess<'de>,
     {
         if self.remaining == Some(0) {
             return None;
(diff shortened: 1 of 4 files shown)


TimDiekmann (approved):
> Does FieldAccess::Value and FieldAccess::Key need to be Deserialize<'de>? 

I think it's fine either way. Generally, unless required by the trait, my first instinct is to say no, let's leave out the constraints. However, as we require the bound on `Key`, it makes sense to also have it on value, especially as we expect to require it for `Value` downstream anyway. 

I'm fine with both approaches, I'd probably keep it.

indietyp:
After thinking about it some more, I think I am inclined to remove the bound of type, the reason is because it enables us to do a pretty interesting pattern for structs.

```rust
struct Example {
	field_a: u8,
	field_b: u16,
	field_c: u8
}

// we can now create an enum for every field and a field identifier
// the naming here is non-PascalCase because we could have a clash in identifier (tho in reality we might just want to disallow that).
struct ExampleFieldIdent {
	#[deer(default)]
	field_a,
	field_b,
	field_c
}

impl Deserialize for ExampleFieldIdent {
	fn deserialize(deserializer: Deserialize) -> Result<Self, _> {
		struct Visitor {
			fn visit_str(value: &str) {
				match value {
					"field_a" => Self::field_a,
					/* ... */
				}
			}
		}
	}
}

struct ExampleField {
	field_a(u8),
	field_b(u16),
	field_c(u8)
}

struct ExampleFieldIdentAccess;

impl FieldAccess for ExampleFieldIdentAcces {
	type Key = ExampleFieldIdent;
	type Value = ExampleField;

	fn value(&self, key: &ExampleFieldIdent, deserializer: D) -> Result<Self::Value, _> {
		match key {
			ExampleFieldIdent::field_a => u8::deserialize(deserializer).map(ExampleField::field_a),
			/* ... */
		}
	}
}

impl Deserialize for Example {
	fn deserialize(&self, deserializer: D) -> Result<Self, _> {
		struct DVisitor;

		impl Visitor for DVisitor {
			fn visit_object(&self, visitor: V) {
				visitor.set_bounded(3);
				let mut error = None;

				let field_a = None;
				let field_b = None;
				let field_c = None;

				while let Some(v) = visitor.next() {
					let v = v.map(|(_, value)| value);
					
					match v {
						Ok(ExampleField::field_a(value)) => field_a = Some(value),
						/* ... */
					}
				}
				
				/* insert error accumulator if field_b and field_c are None */
				
			  if let Some(error) = error {
					return Err(error);
				}

				Self {
					field_a: field_a.unwrap_or_default(),
					field_b: field_b.unwrap(), // infallible to check before,
					field_c: field_c.unwrap()
				}
			}
		}
	}
}
```

This way, we do not need arbitrary access to the object itself. We can get the keys one by one! And we keep the same functionality.

The example is incomplete, I know, but hopefully brings across the point I am trying to make.

TimDiekmann (approved):

PR #2164, opened 2023-03-06
Fix and improve `Address` block

luisbettencourt:

This PR makes some improvements to the address block and fixes some issues with its implementation that were exposed while using it from within WordPress.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204086167143789/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203623179643287/1204105361082361/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203623179643287/1204086167143853/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203623179643290/1204147539752983/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203623179643290/1204145931939215/f) _(internal)_

## 🔍 What does this change?

- Created an `Autocomplete` component in the design-system so it's easier to reuse accross blocks and apps;
- Prevented the address block from updating and creating entities in readonly mode;
- Added styles to override WP global styles (in inputs and links);
- Made changes to the block's state so that it is correctly persisted and displayed;
- Changed the static map API call to hide the logo and attribution watermark (dependent on [this PR](https://github.com/hashintel/internal-api/pull/76) being merged);
- Fixed the map file name and extension that prevented it from being uploaded to WP;
- Changed media query for a container query to make the block responsive according to its container size instead of window size;
- Prevented searching for addresses when the query is empty;
- Refetch coordinates, between sessions, when a map with a different zoom level is requested (Mapbox doesn't allow for persisting coordinates);
- Allow shift+enter in editable fields;
- Added eslint;
- Replaced block icon;
- Retained zoom controls when the map doesn't load.


CiaranMn (changes requested):
I don't get an error displayed if a request fails (I ran out of credits), can you have a look please?

CiaranMn (approved):

PR #2166, opened 2023-03-06
Add the `merge_group` trigger to CI workflows

TimDiekmann:

In the BP repository we enabled GH Merge Queue and this went pretty well. This is going to allow the GH Merge Queue in this repository as well.

## 🔗 Related links

- https://github.com/blockprotocol/blockprotocol/pull/1100
- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1677788837126439) _(internal)_

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -11,6 +11,7 @@ on:
       - canary
       - dev/*
       - "trunk-merge/**"
+  merge_group:
 
   schedule:
     - cron: "30 0 1,15 * *" # scheduled for 00:30 UTC on both the 1st and 15th of the month
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
     branches:
       - main
       - "trunk-merge/**"
+  merge_group:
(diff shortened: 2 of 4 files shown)


.github/workflows/semgrep.yml
```diff
@@ -11,6 +11,7 @@ on:
       - canary
       - dev/*
       - "trunk-merge/**"
+  merge_group:
```
TimDiekmann, line 14: I don't know if we added `trunk-merge` here by accident, but this does not run on PRs. I added it to the merge queue as well to be consistent, but I'm happy to remove both lines.

nonparibus:
I think we can probably remove the `.trunk` folder entirely as part of this PR, right, and all references to Trunk?

We're not using **Trunk Check** (at least at the moment), and there is no custom config for that or **Trunk Analytics** (which simply needs repo access) anyway. The directory purely houses (now redundant) merge-queue related stuff.

nonparibus (approved):

TimDiekmann:
I chose to not remove trunk yet so we have a fallback if it's not working as expected.

PR #2171, opened 2023-03-07
Remove optimization path in traversal logic

TimDiekmann:

We currently have a bug in the traversal logic, which happens due to optimization. This PR removes the optimization and uses a plain recursion

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1203783495017463/f) _(internal)_
- [Asana task](https://app.asana.com/0/0/1204117845662860/f) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C03F7V6DU9M/p1678107081222519) _(internal)_

## 🐾 Next steps

- [Move current implementation to unfold loop instead of recursive calls](https://app.asana.com/0/1203623179643290/1204117845662864/f)

## 🛡 What tests cover this?

Basically every test, which is using the traversal logic

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres.rs
@@ -5,12 +5,6 @@ mod migration;
 mod pool;
 mod query;
 
-use std::{
-    collections::{hash_map::RawEntryMut, HashMap},
-    fmt::Debug,
-    hash::Hash,
-};
-
 use async_trait::async_trait;
 use error_stack::{IntoReport, Result, ResultExt};
 #[cfg(feature = "__internal_bench")]
@@ -23,12 +17,7 @@ use type_system::{
 
 pub use self::pool::{AsClient, PostgresStorePool};
 use crate::{
-    identifier::{
-        account::AccountId,
-        ontology::OntologyTypeRecordId,
-        time::{TemporalInterval, VariableAxis},
-        EntityVertexId, OntologyTypeVertexId,
-    },
+    identifier::{account::AccountId, ontology::OntologyTypeRecordId},
     ontology::{
         ExternalOntologyElementMetadata, OntologyElementMetadata, OwnedOntologyElementMetadata,
     },
@@ -38,7 +27,6 @@ use crate::{
         postgres::ontology::{OntologyDatabaseType, OntologyId},
         AccountStore, BaseUrlAlreadyExists, InsertionError, QueryError, StoreError, UpdateError,
     },
-    subgraph::edges::GraphResolveDepths,
 };
 #[cfg(feature = "__internal_bench")]
 use crate::{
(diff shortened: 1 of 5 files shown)


apps/hash-graph/lib/graph/src/store/postgres.rs
```diff
@@ -49,164 +37,8 @@ use crate::{
...
-        }
-    }
-}
-
 #[derive(Default)]
-pub struct TraversalContext {
-    pub ontology_traversal_map: TraversalMap<OntologyTypeVertexId>,
-    pub knowledge_traversal_map: TraversalMap<EntityVertexId>,
-}
+pub struct TraversalContext;
```
Alfred-Mountfield, line 41: For future knowledge, we're leaving this here to make sure we're accounting for lifetimes, mutability, etc. in the methods that use it (as we will be bringing it back in a new form after this set of PRs)

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -124,7 +101,7 @@ impl<C: AsClient> PostgresStore<C> {
                     },
                     intersected_temporal_axes.clone(),
                 )
-                    .await?;
```
Alfred-Mountfield, line 127: How did `rustfmt` not pick these up before 🤔

apps/hash-graph/lib/graph/src/store/postgres/ontology/data_type.rs
```diff
@@ -23,39 +22,15 @@ impl<C: AsClient> PostgresStore<C> {
...
-                // different to those provided, so we update the resolve depths and the temporal
-                // axes.
-                //
-                // `TraversalMap::update` may return a higher resolve depth than the one
-                // requested, so we update the `resolve_depths` to the returned value.
-                current_resolve_depths = depths;
-                temporal_axes.set_variable_interval(interval.convert());
-                subgraph
-                    .get_or_read::<DataTypeWithMetadata>(self, data_type_id, &temporal_axes)
-                    .await?
```
Alfred-Mountfield, line 54: Don't we still need to do this? 🤔

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
TimDiekmann: Rustfmt currently fails as soon as it sees `let Some(foo) = bar else {};`. Removing that line, applying `rustfmt`, adding it back solved this. AFAIK the rustfmt team is aware of this.

apps/hash-graph/lib/graph/src/store/postgres/ontology/data_type.rs
TimDiekmann: No, the return type was unused before and was only here for consistency. It's not added to the subgraph here but in other locations.

Alfred-Mountfield (approved):

PR #2174, opened 2023-03-08
Move traversal implementation to unfold loop instead of recursive calls

TimDiekmann:

The recursive lookup is easy to implement but very hard to optimize. Instead, we want to use an unfolding loop.

I highly recommend trimming the whitespace when reviewing this PR.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204117845662864/f) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C03F7V6DU9M/p1678119337542689?thread_ts=1678107081.222519&cid=C03F7V6DU9M) _(internal)_

## 🔍 What does this change?

- Use a loop instead of recursion in traversal functions
- As we don't use recursion anymore it's possible to remove the `Pin<Box>` construct
- Pass in a batch of ids to traverse and adjust the `get_` methods
- Rename `current_resolve_depths` to `graph_resolve_depths` and `intersected_temporal_axes` to `temporal_axes` to shadow previous variables to avoid using old variables by accident. I encountered this issue while doing these changes.

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
@@ -73,18 +73,19 @@ impl<C: AsClient> PostgresStore<C> {
                 // Intersects the version interval of the entity with the variable axis's time
                 // interval. We only want to resolve the entity further for the overlap of these two
                 // intervals.
-                let Some(temporal_axes) = temporal_axes
+                let temporal_axes = temporal_axes
                     .clone()
-                    .intersect_variable_interval(entity_interval) else {
-                    // `traverse_entity` is called with the returned entities from `read` with
-                    // `temporal_axes`. This implies, that the version interval of `entity` overlaps
-                    // with `temporal_axes`. `variable_interval` returns `None` if there are
-                    // no overlapping points, so this should never happen.
+                    .intersect_variable_interval(entity_interval)
+                    .unwrap_or_else(|| {
+                        // `traverse_entity` is called with the returned entities from `read` with
+                        // `temporal_axes`. This implies, that the version interval of `entity`
+                        // overlaps with `temporal_axes`. `variable_interval` returns `None` if
+                        // there are no overlapping points, so this should never happen.
                         unreachable!(
-                            "the version interval of the entity does not overlap with the variable \
-                            axis's time interval"
+                            "the version interval of the entity does not overlap with the \
+                             variable axis's time interval"
                         );
-                    };
+                    });
 
                 if graph_resolve_depths.is_of_type.outgoing > 0 {
                     let entity_type_id =
--- a/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
@@ -55,6 +55,8 @@ impl<C: AsClient> PostgresStore<C> {
             .collect::<Vec<_>>();
 
         while !queue.is_empty() {
+            // TODO: We could re-use the memory here but we expect to batch the processing of this
+            //       for-loop. See https://app.asana.com/0/0/1204117847656663/f
             for (entity_vertex_id, graph_resolve_depths, temporal_axes) in mem::take(&mut queue) {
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -40,251 +39,267 @@ impl<C: AsClient> PostgresStore<C> {
...
-                        .metadata
-                        .temporal_versioning()
-                        .variable_time_interval(time_axis);
+        let mut queue = entity_vertex_ids
+            .into_iter()
+            .map(|id| (id, graph_resolve_depths, temporal_axes.clone()))
+            .collect::<Vec<_>>();
+
+        while !queue.is_empty() {
+            for (entity_vertex_id, graph_resolve_depths, temporal_axes) in mem::take(&mut queue) {
```
Alfred-Mountfield, line 58: What's the reason for the loop within the `while` rather than just doing this as part of the `while`?

I guess it could impact bounds checks?

While what I'm about to say is potentially a micro-optimisation, I also want to make sure we don't have this `for` loop for a different micro-optimisation as it seems strange to me.

Wouldn't it likely be faster to not call `mem::take` here and reuse capacity of the current queue? Otherwise we're allocating a new queue every time and have to wait for its grow logic whenever we append a certain amount to the queue (whereas if we started off with a big queue we'd likely not have to wait for reallocations)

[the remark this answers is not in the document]
TimDiekmann: I actively decided to shadow the parameters here as while developing I accidentally used the initial parameters. The initial parameters are only for creating the initial queue and should never be used in the function again.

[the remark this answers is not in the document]
Alfred-Mountfield: That's the exact reason why I wouldn't want us to do this though. Adding `initial_` something makes it pretty clear if we're using the initial ones.

Whereas here I was confused while reading it which we were using, and while I appreciate it's not currently `mut`, a _potential_ change we could make would allow us to mutate it in place to decrement the depth, and then we could hit a really unpleasant bug to find.

I hear your reasoning, so I won't _ask_ you to change it, now that I've explained my concern (and also the confusion I had while reading) I'll leave it to your judgement which risk is higher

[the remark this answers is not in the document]
TimDiekmann: I hear your concern as well. I'd prefer to keep it that way for now but will keep that concern in mind. I guess it's possible to get the best of both worlds when not passing around a tuple of things in the queue (which is a source of confusion here as well), but abstracting it in a struct.

TimDiekmann: The memory cannot be reused here as we're pushing new items to the queue as we loop over the old queue.

Further down the road, I expect to remove the `for` loop here and pass a big part of the queue to the database layer directly. Mainly, I like to see the `queue` as "Items, I like to process at once" or at least as "Items, which _could_ be processed at once". Having everything in one big `for` loop does not help with this approach.

Alfred-Mountfield: > The memory cannot be reused here as we're pushing new items to the queue as we loop over the old queue.

I don't think that's true?
```rust
while !queue.is_empty() {
    let (entity_vertex_id, graph_resolve_depths, temporal_axes) = queue.pop().unwrap();
    ...

    queue.push((...));
}

TimDiekmann: You're right. I added a comment to describe the reasoning: https://github.com/hashintel/hash/pull/2174/commits/b5af96050f865106cb226cc562b700efa4db10d0

Alfred-Mountfield: Thanks, makes sense that we're leaving the pattern in to be ready to process that whole queue in one go as a batch (rather than iteratively in a `for` loop)

Alfred-Mountfield (approved):

PR #2183, opened 2023-03-09
Store adjacency list for each (planned) reference table

TimDiekmann:

We are going to have reference tables between entities and entities, and between entities and entity types. To simplify batched reading of the data from the database, we store a set of edges per reference table

## 🔗 Related links

## 🚫 Blocked by

- #2178 
- #2181 

## 🔍 What does this change?

- Use adjacency lists for each kind of edge (determined by the endpoints, not by the edge kind itself)
    - If we chose to have reference tables for each edge kind, this can easily be added later on
- Merge `VertexIndex` and `VertexId` traits. They are now implemented on exactly the same types in the same context
- Move `OntologyTypeVertexId` to `utoipa_typedef`. As mentioned previously, this could be removed entirely, but to avoid touching the `utoipa_typedef` too many times moving it is sufficient for now. The subgraph typedefs have a very close representation to the current representation of the internal subgraph already. I'm confident that we can remove the `utoipa_typedef` module entirely soon.
- Rename the methods on `VertexId` and `EdgeKind` to align

No

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest.rs
@@ -37,8 +37,8 @@ use crate::{
         middleware::log_request_and_response,
         utoipa_typedef::subgraph::{
             Edges, KnowledgeGraphOutwardEdge, KnowledgeGraphRootedEdges, KnowledgeGraphVertex,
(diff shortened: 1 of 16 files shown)


thehabbos007 (approved):
Thank you! I did some manual tests locally with no observed changes to functionaility

PR #2185, opened 2023-03-09
Make step ordering deterministic in How To block

CiaranMn:

The How To block intends the steps linked to it to be in a certain order, but was not specifying one – it happened to work in some environments where the order the application returned the links was deterministic in the right way, but the block cannot rely on this (applications might send the latest createst links first, for example).

This PR fixes that by specifying a position for the link when creating it.


luisbettencourt (approved):
Thanks Ciaran

PR #2191, opened 2023-03-10
Always run Rust jobs when `.config` changed

TimDiekmann:

In a recent PR we changed the allowed lints in `.config`, but this did not run all expected test, thus, currently `cargo make clippy` will fail when running in the `antsi` crate

## 🔍 What does this change?

- Change `setup.py` to also cover `.config`
- fix config file for `antsi`


TimDiekmann:
This PR is currently in conflict with #2192. I will rebase the PR as soon as the other PR has left the merge queue.

TimDiekmann:
#2192 is failing because of the issue this PR solves

PR #2192, opened 2023-03-10
Use sparse cargo profile

TimDiekmann:

With the sparse registries now being stabilized, we should enable this feature to speed up CI times

## 🔗 Related links

- [Rust 1.68 announcement post](https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#cargos-sparse-protocol)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[registries.crates-io]
+protocol = "sparse"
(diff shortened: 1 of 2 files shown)


thehabbos007 (dismissed):
Excited for this!

thehabbos007 (approved):

PR #2195, opened 2023-03-13
Remove potential optimization path in Query compiler

TimDiekmann:

We have a small deduplication logic in the query compiler, however, the current implementation does not generate the correct SQL Query in all circumstances. The optimization is able to remove an intermediate table, but it's currently not taking the `entities` table into account correctly.

## 🔗 Related links

- #1965
- [Slack discussion](https://hashintel.slack.com/archives/C03TWBSRT8B/p1678703368190799) _(internal)_

## 🔍 What does this change?

This PR removes this optimization to prefer correctness over speed (the speed gain wouldn't be very high).

## ⚠️ Known issues

In queries, where we only use an intermediate table, this table appears in the SQL query again.

## 🐾 Next steps

Re-implement this optimization: [Asana Task](https://app.asana.com/0/1202805690238892/1204165137291093/f) _(internal)_

## 🛡 What tests cover this?

Two unit tests are affected by this change, they have been adjusted

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
@@ -509,33 +509,35 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
         for relation in path.relations() {
             let current_alias = current_table.alias;
             for (current_column, join_column) in relation.joins() {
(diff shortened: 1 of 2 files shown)


indietyp (approved):
LGTM (take w/ a grain of salt), tested locally. The previous query no longer produces an error.

thehabbos007 (approved):

PR #2200, opened 2023-03-14
Fix `owned_by_id` not used when joining entity tables

TimDiekmann:

We currently do not support joining entities by `entity_uuid` *and* `owned_by_id`. This is likely to be one cause why our experiments with Citus results in slower performance than expected.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204165137291102/f) _(internal)_

## 🔍 What does this change?

- Enhance the `Relation::joins` function to allow returning multiple columns
- Adjust compiler and test outputs to take these changes into account
- Drive-by: Remove obsolete `ProjectedTime`

## ⚠️ Known issues

The `Relation` enumeration is quite large. Also, the `ForeignKeyReference` does not check, if the columns belong to the same table. I have a few ideas how to improve this, but this is out of scope in this PR.

## 🐾 Next steps

Initially, this was done as a drive-by when migrating to reference tables, but the issue was actually harder so solve than expected. Based on these changes it's easier to implement reference tables (correctly)

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/query/expression/join_clause.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/query/expression/join_clause.rs
@@ -1,37 +1,119 @@
-use std::fmt;
+use std::{fmt, fmt::Write};
 
-use crate::store::postgres::query::{AliasedColumn, Transpile};
+use crate::store::postgres::query::{
+    table::{Column, ForeignKeyReference},
+    Alias, AliasedTable, Transpile,
+};
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct JoinCondition<'p> {
+    pub join: Column<'p>,
+    pub on: Column<'p>,
+}
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum JoinType {
+    Inner,
+    LeftOuter,
+    RightOuter,
+    FullOuter,
+}
+
+impl Transpile for JoinType {
+    fn transpile(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Inner => fmt.write_str("INNER JOIN"),
+            Self::LeftOuter => fmt.write_str("LEFT OUTER JOIN"),
+            Self::RightOuter => fmt.write_str("RIGHT OUTER JOIN"),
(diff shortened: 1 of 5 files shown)


apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
```diff
@@ -479,97 +474,134 @@ pub enum Relation {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum ForeignKeyReference {
+    Single {
+        on: Column<'static>,
+        join: Column<'static>,
+    },
+    Double {
+        on: (Column<'static>, Column<'static>),
+        join: (Column<'static>, Column<'static>),
+    },
```
thehabbos007, line 486: Any reason for this being split up like this instead of just defining the column names in something like a const array or equivalent? Could probably simplify the `from_foreign_key` function if we change the representation

thehabbos007 (commented):
Thank you, Tim! The adjusted tests look as expected, I suppose we can lean on those. Happy to merge after we address the comments below :)

apps/hash-graph/lib/graph/src/store/postgres/query/expression/join_clause.rs
```diff
@@ -1,37 +1,119 @@
...
+            (false, true) => Self::RightOuter,
+            (true, true) => Self::FullOuter,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct JoinOn<'p> {
+    pub left: Column<'p>,
+    pub right: Column<'p>,
```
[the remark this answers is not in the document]
TimDiekmann: I renamed them to `join` and `on` as well. Thanks for the suggestions!

apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
TimDiekmann: I changed the definition of `Double` to be `[Colum<'static>; 2]` instead, which makes it more readable. I was not able to make the number of columns static as it's not known at compile time as this comes from user input enumerations. I used this approach over just a `Vec<_>` to have better control over the size of the join statement and avoid unneeded allocations, as otherwise, we'd need a vector for each join.

PR #2201, opened 2023-03-14
Simplify reference table definitions

TimDiekmann:

As noted in #2200, the `Relation::joins` function got very long and is expected to get even longer with more reference tables added, which is the next step.

## 🔗 Related links

- #2200

## 🚫 Blocked by

- #2200 

## 🔍 What does this change?

- Introduce `ReferenceTable` enumeration, which defines the layout of a reference table
- Use `ReferenceTable` inside of `Table::Reference(ReferenceTable)` to provide the same API
- Return `impl Iterator` instead of a dynamically sized list when requesting the required joins for a relation

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
@@ -2,6 +2,7 @@ use std::{
     borrow::Cow,
     fmt::{self, Debug},
     hash::Hash,
+    iter::{once, Chain, Once},
 };
 
 use crate::{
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
```diff
@@ -336,12 +413,18 @@ impl<'p> Column<'p> {
...
             Self::PropertyTypePropertyTypeReferences(_) => {
-                Table::PropertyTypePropertyTypeReferences
+                Table::ReferenceTable(ReferenceTable::PropertyTypePropertyTypeReferences)
+            }
+            Self::EntityTypePropertyTypeReferences(_) => {
+                Table::ReferenceTable(ReferenceTable::EntityTypePropertyTypeReferences)
+            }
+            Self::EntityTypeEntityTypeReferences(_) => {
+                Table::ReferenceTable(ReferenceTable::EntityTypeInheritance)
             }
```
thehabbos007, line 427: I may be a little out of the loop, but we expect to add an explicit reference table for `EntityTypeLinks` as well here, right?

thehabbos007 (approved):
Looks good to me, I suppose this shouldn't have any observable effects, so tests should just pass and nothing to check, right?

apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
TimDiekmann: Yes, this is done in #2202. Currently, the reference table is the same.

thehabbos007 (approved):

PR #2203, opened 2023-03-15
Improve `How-to` block

luisbettencourt:

This PR makes some improvements to the **How-to** block

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204117779689068/f) _(internal)_

## 🔍 What does this change?

- Removed the background and the drop shadow box around text in readonly mode;
- Shortened placeholder on smaller screen widths while editing to read "Additional instructions here";
- Made several changes to `<EditableField />` and moved it to the design system, so it can be used by other blocks (FAQ block will use it);
- Added eslint and fixed errors;
- Removed unused component.


blocks/how-to/src/app.tsx
```diff
@@ -271,217 +296,248 @@ export const App: BlockComponent<RootEntity> = ({
...
+                        }}
+                        appear
+                      >
+                        <Box
+                          sx={{
+                            mb: index === steps.length - 1 ? 0 : 3,
+                            transition: ({ transitions }) =>
+                              transitions.create("margin-bottom"),
+                          }}
+                        >
```
yusufkinatas, line 492: Curious if this can be achieved via using  `display: flex` & `gap: 3` on the parent container instead of using `margin-bottom`? Or is the push down animation here something we want to keep?

yusufkinatas (changes requested):
Great work! I guess this wasn't at the scope of this ticket, but I guess we can fix the issue shown in the video above in this PR as well. When block is at `readonly` mode, I can still edit the fields.

https://user-images.githubusercontent.com/39553853/226539629-083d3b28-106b-44a4-bfc9-6b840b8f3d5f.mov

yusufkinatas:
Also looks like linting fails.

blocks/how-to/src/app.tsx
luisbettencourt: It does work to keep the spacing between steps, but it makes the delete step animation jump.

yusufkinatas (approved):

PR #2205, opened 2023-03-15
Fix collecting of edges by merging them

indietyp:

Currently, there's a bug that only the entity types will be returned if entity edges and entity types are requested in the same query. This is because instead of merging the flattened adjacency lists, only the last list is collected (due to the use of `collect()`). This PR instead uses fold and properly merges duplicate entity ids.

> This is also possible if entity types and property types are requested.


TimDiekmann (approved):
Thanks for fixing this!

PR #2208, opened 2023-03-16
Support reversed edges in structural ontology queries

TimDiekmann:

To migrate the current implementation of looking up ontology-type references in memory to instead be from the reference tables, the filter should be pre-defined to avoid writing the filters in the traversal logic.
This PR adds all kind of allowed  edge filters.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204165139634979/f) _(internal)_

## 🚫 Blocked by

- #2202 
- #2207 

## 🔍 What does this change?

- Add filters for all kind of edges
- Replace old entity filters with new edge-based filters

## 🐾 Next steps

- [Migrate the current implementation of looking up ontology-type references in memory to instead be from the reference tables in DB](https://app.asana.com/0/1204000740778935/1204117847656666/f) _(internal)_

## 🛡 What tests cover this?

A test for each newly added filter was added:
- deserializable filters are tested against their JSON representation
- All filters (reversed and non-reversed) are tested with their resulting SQL query

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/query/filter.rs
+++ b/apps/hash-graph/lib/graph/src/store/query/filter.rs
@@ -644,7 +644,7 @@ mod tests {
             version: 1,
         };
 
-        let expected = json! {{
+        let expected = json!({
           "all": [
             { "equal": [
               { "path": ["baseUrl"] },
@@ -655,7 +655,7 @@ mod tests {
               { "parameter": url.version }
             ]}
           ]
-        }};
+        });
 
         test_filter_representation(
             &Filter::<DataTypeWithMetadata>::for_versioned_url(&url),
@@ -673,7 +673,7 @@ mod tests {
             revision_id: OntologyTypeVersion::new(1),
         };
 
-        let expected = json! {{
+        let expected = json!({
           "all": [
             { "equal": [
               { "path": ["baseUrl"] },
@@ -684,7 +684,7 @@ mod tests {
               { "parameter": url.revision_id }
             ]}
           ]
-        }};
+        });
 
         test_filter_representation(
             &Filter::<DataTypeWithMetadata>::for_ontology_type_vertex_id(&url),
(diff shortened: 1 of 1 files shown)


apps/hash-graph/lib/graph/src/store/query/filter.rs
```diff
@@ -518,6 +692,169 @@ mod tests {
...
+    fn for_property_type_constrains_properties_on() {
+        let url = PropertyTypeVertexId {
+            base_id: BaseUrl::new(
+                "https://blockprotocol.org/@blockprotocol/types/property-type/text/".to_owned(),
+            )
+            .expect("invalid base url"),
+            revision_id: OntologyTypeVersion::new(1),
+        };
+
+        let expected = json! {{
```
thehabbos007, line 737: small nit: does `rustfmt` add this space after `json!`?

apps/hash-graph/lib/graph/src/store/query/filter.rs
```diff
@@ -659,7 +1011,14 @@ mod tests {
           ]
         }};
 
-        test_filter_representation(&Filter::for_right_entity_by_entity_id(entity_id), &expected);
+        test_filter_representation(
+            &Filter::for_knowledge_graph_edge_by_entity_id(
+                entity_id,
+                KnowledgeGraphEdgeKind::HasRightEntity,
+                true,
+            ),
```
thehabbos007, line 1019: Just want to say that this seems to be more accurate to model the above query. In the old code `Filter::for_right_entity_by_entity_id` didn't really capture whether or not this was a reversed link or not. This new filter syntax is much more clear on that :+1:

thehabbos007 (commented):
Thank you, Tim! Looks good to me, couple quick comments :)

apps/hash-graph/lib/graph/src/store/query/filter.rs
TimDiekmann: Yes, it does, and it's annoying. I replaced it with `json!({` instead: [Replace json! braces](https://github.com/hashintel/hash/pull/2208/commits/c522386388c9a130d9b2ac27b508709456198b91)

thehabbos007 (approved):

PR #2222, opened 2023-03-17
Add Minesweeper block

CiaranMn:

Adds a Minesweeper block, which is a thin wrapper around https://github.com/kurehajime/minesweeper_tag allowing users to set the grid via and number of mines via entity properties.

The underlying component does not expose the state of the game and therefore game state is not persisted.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204065050321088/f) _(internal)_

## 🚫 Blocked by

- [x] Once PR is approved, publish the types to production and update this PR to use the production type
- [x] Ideally once https://github.com/blockprotocol/blockprotocol/pull/1191 is merged and a new version of `block-scripts` published, update the version of `block-scripts` here so that the correct `block-metadata.json` is generated (but not strictly necessary as it's cleaned by the BP API as part of publishing anyway)

## 🐾 Next steps

- Publish the block once finalised.

## ❓ How to test this?

1.  `yarn dev`

## 📹 Demo

Screenshot 2023-03-17 at 09 57 25


nonparibus (dismissed):
Legal UAT ✅

blocks/minesweeper/src/types/generated/block-entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  MinesweeperBlockOutgoingLinkAndTarget;
+
+export type MinesweeperBlock = Entity<MinesweeperBlockProperties>;
+
+export type MinesweeperBlockOutgoingLinkAndTarget = never;
+
+export type MinesweeperBlockOutgoingLinksByLinkEntityTypeId = {};
+
+/**
+ * The block entity of the "Minesweeper" block
```
Alfred-Mountfield, line 19:
```suggestion
 * The block entity of the "Minesweeper" block
 * 
 * See: https://blockprotocol.org/@hash/blocks/minesweeper
```

I know we can't edit this file but we should change the description to match this, it's consistent with the other block entity types

blocks/minesweeper/src/app.ts
```diff
@@ -0,0 +1,65 @@
...
+// eslint-disable-next-line import/extensions
+import { unsafeHTML } from "lit/directives/unsafe-html.js";
+import { mine_sweeper } from "mine-sweeper-tag";
+
+import { BlockEntity } from "./types/generated/block-entity";
+
+const colsKey: keyof BlockEntity["properties"] =
+  "https://blockprotocol-molpob88k.stage.hash.ai/@ciaranm/types/property-type/number-of-columns/";
+const minesKey: keyof BlockEntity["properties"] =
+  "https://blockprotocol-molpob88k.stage.hash.ai/@ciaranm/types/property-type/number-of-mines/";
```
luisbettencourt, line 13: Just a reminder to update these

blocks/minesweeper/src/app.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (
+    typeof value === "number" ||
+    (typeof value === "string" && !Number.isNaN(parseInt(value, 10)))
+  ) {
+    return value;
+  }
+  return defaultValue;
+};
+
+export class MinesweeperBlock extends BlockElementBase<BlockEntity> {
```
luisbettencourt, line 25: Interesting, why did you go with this approach?

luisbettencourt (approved):

blocks/minesweeper/src/app.ts
CiaranMn: Part of the motivation for this block was to use the custom element template. It's also easier to render custom elements using this template than React – React requires a special wrapper for custom elements.

blocks/minesweeper/src/types/generated/block-entity.ts
CiaranMn: https://github.com/hashintel/hash/pull/2222/commits/afca30260994fac67c35cca5eb0fd41f53dc9855

blocks/minesweeper/src/app.ts
CiaranMn: https://github.com/hashintel/hash/pull/2222/commits/afca30260994fac67c35cca5eb0fd41f53dc9855

luisbettencourt (approved):

indietyp:
Great work on the PR! Minor note, though, looking at the chosen implementation, they implemented Minesweeper incorrectly. The first move should never have a mine (the mines are generated after the first move). This is not the case in this implementation.

CiaranMn:
> Great work on the PR! However, looking at the chosen implementation, they implemented Minesweeper incorrectly. The first move should never have a mine (the mines are generated after the first move). After having a cursory look on the underlying library this is not the case 😞

@indietyp I believe that if you click a mine on the first move on Windows it is actually moved so that the click is safe, rather than mines being generated after the first move (there needs to be some assumption of where mines are in the first move to generate the size of the opening you get) – either way the library doesn't make it safe, you're right 😁

PR #2223, opened 2023-03-17
Use reference tables when looking up ontology references

TimDiekmann:

We're going to look up the references before looking up the actual records. This implies, that ontology references also needs to be looked up through the database. This also implies, that ontology types do not need to be looked up one-by-one.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204117847656666/f) _(internal)_

## 🚫 Blocked by

- #2202 
- #2207
- #2208

## 🔍 What does this change?

Replace lazy lookups based on the returned ontology reference iterators with direct lookups through reference tables

## 🐾 Next steps

The structure of looking up edges is now almost identical, which enables further work to [deduplicate the subgraph traversal logic](https://app.asana.com/0/1203363157432084/1203444301722127/f)

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
@@ -14,6 +14,7 @@ use crate::{
         time::{DecisionTime, Timestamp},
     },
     knowledge::{Entity, EntityLinkOrder, EntityMetadata, EntityProperties, EntityUuid, LinkData},
+    ontology::EntityTypeWithMetadata,
     provenance::{OwnedById, ProvenanceMetadata, UpdatedById},
     store::{
         crud::Read,
@@ -27,7 +28,7 @@ use crate::{
             EdgeResolveDepths, GraphResolveDepths, KnowledgeGraphEdgeKind,
             OutgoingEdgeResolveDepth, SharedEdgeKind,
         },
-        identifier::{EntityIdWithInterval, EntityTypeVertexId, EntityVertexId},
+        identifier::{EntityIdWithInterval, EntityVertexId},
         query::StructuralQuery,
         temporal_axes::QueryTemporalAxes,
         Subgraph,
@@ -41,30 +42,27 @@ impl<C: AsClient> PostgresStore<C> {
     #[tracing::instrument(level = "trace", skip(self, traversal_context, subgraph))]
     pub(crate) async fn traverse_entities(
         &self,
-        entity_vertex_ids: Vec<EntityVertexId>,
-        temporal_axes: QueryTemporalAxes,
-        graph_resolve_depths: GraphResolveDepths,
+        mut entity_queue: Vec<(EntityVertexId, GraphResolveDepths, QueryTemporalAxes)>,
         traversal_context: &mut TraversalContext,
         subgraph: &mut Subgraph,
     ) -> Result<(), QueryError> {
-        let time_axis = temporal_axes.variable_time_axis();
+        let time_axis = subgraph.temporal_axes.resolved.variable_time_axis();
 
-        let mut queue = entity_vertex_ids
-            .into_iter()
-            .map(|id| (id, graph_resolve_depths, temporal_axes.clone()))
-            .collect::<Vec<_>>();
+        let mut entity_type_queue = Vec::new();
 
-        while !queue.is_empty() {
+        while !entity_queue.is_empty() {
             // TODO: We could re-use the memory here but we expect to batch the processing of this
             //       for-loop. See https://app.asana.com/0/0/1204117847656663/f
-            for (entity_vertex_id, graph_resolve_depths, temporal_axes) in mem::take(&mut queue) {
-                let entity: &Entity = subgraph.get_vertex(&entity_vertex_id).unwrap_or_else(|| {
-                    // Entities are always inserted into the subgraph before they are resolved,
-                    // so this should never happen. If it does, it is a bug.
-                    unreachable!("entity should already be in the subgraph");
-                });
-
-                let entity_interval = entity
+            for (entity_vertex_id, graph_resolve_depths, temporal_axes) in
+                mem::take(&mut entity_queue)
+            {
+                let entity_interval = subgraph
+                    .get_vertex::<Entity>(&entity_vertex_id)
+                    .unwrap_or_else(|| {
+                        // Entities are always inserted into the subgraph before they are resolved,
+                        // so this should never happen. If it does, it is a bug.
+                        unreachable!("entity should already be in the subgraph");
+                    })
                     .metadata
                     .temporal_versioning()
                     .variable_time_interval(time_axis);
@@ -87,28 +85,38 @@ impl<C: AsClient> PostgresStore<C> {
                     });
 
                 if graph_resolve_depths.is_of_type.outgoing > 0 {
-                    let entity_type_id =
-                        EntityTypeVertexId::from(entity.metadata.entity_type_id().clone());
-                    subgraph.insert_edge(
-                        &entity_vertex_id,
-                        SharedEdgeKind::IsOfType,
-                        entity_type_id.clone(),
-                    );
-
-                    self.traverse_entity_type(
-                        vec![entity_type_id],
-                        temporal_axes.clone(),
-                        GraphResolveDepths {
-                            is_of_type: OutgoingEdgeResolveDepth {
-                                outgoing: graph_resolve_depths.is_of_type.outgoing - 1,
-                                ..graph_resolve_depths.is_of_type
-                            },
-                            ..graph_resolve_depths
-                        },
-                        traversal_context,
-                        subgraph,
+                    for entity_type in <Self as Read<EntityTypeWithMetadata>>::read(
+                        self,
+                        &Filter::for_shared_edge_by_entity_id(
+                            entity_vertex_id.base_id,
+                            SharedEdgeKind::IsOfType,
+                        ),
+                        &temporal_axes,
                     )
-                    .await?;
+                    .await?
+                    {
+                        let entity_type_vertex_id = entity_type.vertex_id(time_axis);
+
+                        subgraph.insert_edge(
+                            &entity_vertex_id,
+                            SharedEdgeKind::IsOfType,
+                            entity_type_vertex_id.clone(),
+                        );
+
+                        subgraph.insert_vertex(&entity_type_vertex_id, entity_type);
+
+                        entity_type_queue.push((
+                            entity_type_vertex_id,
+                            GraphResolveDepths {
+                                is_of_type: OutgoingEdgeResolveDepth {
+                                    outgoing: graph_resolve_depths.is_of_type.outgoing - 1,
+                                    ..graph_resolve_depths.is_of_type
+                                },
+                                ..graph_resolve_depths
+                            },
+                            temporal_axes.clone(),
+                        ));
+                    }
                 }
 
                 if graph_resolve_depths.has_left_entity.incoming > 0 {
@@ -143,7 +151,7 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph
                             .insert_vertex(&outgoing_link_entity_vertex_id, outgoing_link_entity);
 
-                        queue.push((
+                        entity_queue.push((
                             outgoing_link_entity_vertex_id,
                             GraphResolveDepths {
                                 has_left_entity: EdgeResolveDepths {
@@ -189,7 +197,7 @@ impl<C: AsClient> PostgresStore<C> {
                         subgraph
                             .insert_vertex(&incoming_link_entity_vertex_id, incoming_link_entity);
 
-                        queue.push((
+                        entity_queue.push((
                             incoming_link_entity_vertex_id,
                             GraphResolveDepths {
                                 has_right_entity: EdgeResolveDepths {
@@ -227,7 +235,7 @@ impl<C: AsClient> PostgresStore<C> {
                         let left_entity_vertex_id = left_entity.vertex_id(time_axis);
                         subgraph.insert_vertex(&left_entity_vertex_id, left_entity);
 
-                        queue.push((
+                        entity_queue.push((
                             left_entity_vertex_id,
                             GraphResolveDepths {
                                 has_left_entity: EdgeResolveDepths {
@@ -265,7 +273,7 @@ impl<C: AsClient> PostgresStore<C> {
                         let right_entity_vertex_id = right_entity.vertex_id(time_axis);
                         subgraph.insert_vertex(&right_entity_vertex_id, right_entity);
 
-                        queue.push((
+                        entity_queue.push((
                             right_entity_vertex_id,
                             GraphResolveDepths {
                                 has_right_entity: EdgeResolveDepths {
@@ -280,6 +288,9 @@ impl<C: AsClient> PostgresStore<C> {
                 }
             }
         }
+
+        self.traverse_entity_types(entity_type_queue, traversal_context, subgraph)
+            .await?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -87,34 +85,48 @@ impl<C: AsClient> PostgresStore<C> {
...
                 }
 
                 if graph_resolve_depths.has_left_entity.incoming > 0 {
                     for outgoing_link_entity in <Self as Read<Entity>>::read(
                         self,
-                        &Filter::for_outgoing_link_by_source_entity_id(entity_vertex_id.base_id),
+                        &Filter::for_knowledge_graph_edge_by_entity_id(
+                            entity_vertex_id.base_id,
+                            KnowledgeGraphEdgeKind::HasLeftEntity,
+                            true,
```
Alfred-Mountfield, line 128: Is it worth wrapping these in a struct and creating aliases kind of like we've done in TypeScript (although obviously easier there)?

Just thinking it might be too easy to accidentally put the wrong boolean in or the wrong edge-kind, and it'd be nice to just say `OutgoingLinkEdge` or something.

Could be done in follow-up if you think it's worth it

TimDiekmann: I replaced the boolean with an enumeration in #2226

Alfred-Mountfield: Yeah I noticed that :) 

I think it'd still be nice to know that an `OutgoingLink` is equivalent to `HasLeftEntity, Incoming`

Alfred-Mountfield: Similar to the aliases in this file:
https://github.com/blockprotocol/blockprotocol/blob/fc2153c3d86d53f10aa76108360a0ea35e98f16f/libs/%40blockprotocol/graph/src/shared/types/subgraph/edges/variants/knowledge.ts#L8C20-L15

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -25,197 +25,183 @@ impl<C: AsClient> PostgresStore<C> {
...
-                    for inherits_from_type_ref_url in inherits_from_type_ref_urls {
-                        let inherits_from_type_vertex_id =
-                            EntityTypeVertexId::from(inherits_from_type_ref_url);
+                if graph_resolve_depths.inherits_from.outgoing > 0 {
+                    for referenced_entity_type in <Self as Read<EntityTypeWithMetadata>>::read(
+                        self,
+                        &Filter::<EntityTypeWithMetadata>::for_ontology_edge_by_entity_type_vertex_id(
+                            &entity_type_vertex_id,
+                            OntologyEdgeKind::InheritsFrom,
+                            true,
```
Alfred-Mountfield, line 91: Why is this reversed?

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -25,197 +25,183 @@ impl<C: AsClient> PostgresStore<C> {
...
+                    }
+                }
+
+                if graph_resolve_depths.constrains_links_on.outgoing > 0 {
+                    for referenced_entity_type in <Self as Read<EntityTypeWithMetadata>>::read(
+                        self,
+                        &Filter::<EntityTypeWithMetadata>::for_ontology_edge_by_entity_type_vertex_id(
+                            &entity_type_vertex_id,
+                            OntologyEdgeKind::ConstrainsLinksOn,
+                            true,
```
Alfred-Mountfield, line 128: Why is this reversed?

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -25,197 +25,183 @@ impl<C: AsClient> PostgresStore<C> {
...
+                    .constrains_link_destinations_on
+                    .outgoing
+                    > 0
+                {
+                    for referenced_entity_type in <Self as Read<EntityTypeWithMetadata>>::read(
+                        self,
+                        &Filter::<EntityTypeWithMetadata>::for_ontology_edge_by_entity_type_vertex_id(
+                            &entity_type_vertex_id,
+                            OntologyEdgeKind::ConstrainsLinkDestinationsOn,
+                            true,
```
Alfred-Mountfield, line 169: Why is this reversed?

apps/hash-graph/lib/graph/src/store/postgres/ontology/property_type.rs
```diff
@@ -25,105 +25,92 @@ impl<C: AsClient> PostgresStore<C> {
...
-                    for property_type_ref_url in property_type_ref_urls {
-                        let property_type_vertex_id =
-                            PropertyTypeVertexId::from(property_type_ref_url);
+                if graph_resolve_depths.constrains_properties_on.outgoing > 0 {
+                    for referenced_property_type in <Self as Read<PropertyTypeWithMetadata>>::read(
+                        self,
+                        &Filter::<PropertyTypeWithMetadata>::for_ontology_edge_by_property_type_vertex_id(
+                            &property_type_vertex_id,
+                            OntologyEdgeKind::ConstrainsPropertiesOn,
+                            true,
```
Alfred-Mountfield, line 86: Why is this reversed?

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: Let's say we have an entity type `person` and `friendship` with the edge `person --> friendship`. `person` has the link `friendship` (the destination does not matter here but the idea is the same). To get the `friendship` entity type we need to query for the entity type, where the reversed edge (`friendship --< person`)has the vertex id of the person: "give me all entity types, where the entity types, which uses _this_ entity type as a link have the vertex id of `person`".

Alfred-Mountfield: This is about `Inherits` from but I presume the logic is the same.

Are the knowledge graph filters written the other way around?
https://github.com/hashintel/hash/blob/a30bc9e4ea3b8b83dc0b5cdefa9d7c16256391da/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs#L123-L130

Reads as if it were the other way around.

`outgoing_link_entity` is equivalent to a (kind=left_entity, reversed=true) relationship
- You are querying for `left_entity, true`
`inherits_from` here is equivalent to a (kind=inherits_from, reversed=false) relationship
- You are querying for `inherits_from, false`

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
TimDiekmann: I prefer to not add those aliases in this location/at this point. We don't consume the subgraph, currently, and it's only partially the truth. The subgraph does not imply meaning to link directions (thus, it even supports right-to-left links bi-directional links)
Wrapping those in a struct would be a step back as I need exactly this representation to unify the interface and make a subgraph, which can resolve only on edge kinds. If these would be in a struct again, this would need to be unwrapped or a trait would need to be defined.

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: > This is about InheritsFrom not links but I presume the logic is the same.

Sorry, I already moved the comment to the other section but GitHub probably has not updated for you since then and you were already replying.

> Are the knowledge graph filters written the other way around?

This is a mistake I'm sure is not present in the follow-up. I have noticed this issue when writing the follow-up but as I just rebased this branch I have not yet updated this branch to use the correct logic. Thanks for pointing out, though. finding this is hard :D

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
TimDiekmann: It's fixed now: [Fix traversal direction](https://github.com/hashintel/hash/pull/2223/commits/de548a6431d89fd2147ba9a6e9456f7326db7251)

Alfred-Mountfield: I think it would help readability a _**lot**_, otherwise we are going to easily run into problems exactly like the ones described below.

> make a subgraph, which can resolve only on edge kinds

In that case wouldn't we want something even more like this, to make sure that seeing some random edge kind and a direction would be understandable.

I guess the main issue is that we can't just wrap the two parameters (KnowledgeGraphEdgeKind, bool) into one generic struct and write aliases, because we can't constrain it to a specific enum variant? If there isn't an easy way around that then yea

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
Alfred-Mountfield: > Sorry, I already moved the comment to the other section

(Just for persistence) https://github.com/hashintel/hash/pull/2223#discussion_r1144646196

> This is a mistake I'm sure is not present in the follow-up. I have noticed this issue when writing the follow-up but as I just rebased this branch I have not yet updated this branch to use the correct logic. Thanks for pointing out, though. finding this is hard :D

Which is why I'm suggesting we might want to find a way to greatly improve the readability of these things https://github.com/hashintel/hash/pull/2223#discussion_r1144656382

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
Alfred-Mountfield: Considering these are all standalone functions (effectively aliases for specific filters) do we want to define them the other way around so they're more intuitive and more like how we'd think of writing a query?

"Get me all outgoing links of this entity"

Because now I'm going to look at this code:
https://github.com/hashintel/hash/blob/a30bc9e4ea3b8b83dc0b5cdefa9d7c16256391da/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs#L123-L130

And [Fix traversal direction](https://github.com/hashintel/hash/pull/2223/commits/de548a6431d89fd2147ba9a6e9456f7326db7251) means I'm going to be even more confused

As I'm going to look at the code for correctness and think, well
an `outgoing_link_entity` relationship is equivalent to `(kind=left_entity, reversed=true)`

And I'm going to look at your filter and see `left_entity, false` and think it's a bug

Alfred-Mountfield: Hmm okay, thinking more about how the filters work, I guess swapping it around would be confusing and inconsistent with other kinds of filters..

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: And this is exactly the reason why I want to abstract the logic away and don't have this logic in 11 places (each edge kind)

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
Alfred-Mountfield: Except something like this _would_ be more readable, I just appreciate we may not easily be able to achieve it.

```rust
 for (edge_kind, edge_direction) in &[ 
     OutgoingLinkEdge, 
     IncomingLinkEdge 
     LeftEntityEdge, 
     RightEntityEdge, 
 ] { 
     entity_queue.extend( 
         self.traverse_knowledge_graph_edge( 
             entity_vertex_id, 
             entity_interval, 
             *edge_kind, 
             *edge_direction, 
             graph_resolve_depths, 
             &temporal_axes, 
             traversal_context, 
             subgraph, 
         ) 
         .await?, 
     ); 
 } 
```

TimDiekmann: But what is an `OutgoingLinkEdge`? Either, we introduce such an edge or it would be confusing as soon as we have the non-default case, where the left entity is the target of the right entity.

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: The current end of the stack still have those, but I expect to completely remove these and only have `Filter::for_edge_by_vertex_id`

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
Alfred-Mountfield: As I expressed offline, we have an `OutgoingLinkEdge`, it is a thing in the official implementation of the spec. This is an implementation of a datastore that respects that spec.

Should the spec be updated to allow more things, we can always adjust the code.

> But what is an OutgoingLinkEdge

I don't really understand why this is so controversial when _you've already written it in the code_
https://github.com/hashintel/hash/blob/a30bc9e4ea3b8b83dc0b5cdefa9d7c16256391da/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs#L123

Alfred-Mountfield: If you feel really strongly that there shouldn't be such an alias as `OutgoingLink` then a separate discussion can be had about that within the context of the Block Protocol and the spec, but at the moment it _is_ a thing.

TimDiekmann: Because the above abstracts over this. Also, this does not align with the idea of completely abstracting away the logic. We **_don't_** have an `OutgoingLinkEdge` in the resolve depths and the traversal logic should only take the resolve depths into account when traversing it.

> Could be done in follow-up if you think it's worth it

I don't think it's worth it.

TimDiekmann: > If you feel really strongly that there shouldn't be such an alias as OutgoingLink then a separate discussion can be had about that within the context of the Block Protocol and the spec, but at the moment it is a thing.

Let's discuss this offline but please let us not block this PR on this as the code you commented on is already merged.

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
Alfred-Mountfield: > The current end of the stack still have those, but I expect to completely remove these and only have Filter::for_edge_by_vertex_id

Won't you need to write it the other way around at that point? If you're completely abstracting over just the edge by the vertex id, will you know the type of the destination you're querying?

I just find it very unintuitive that we're abstracting it to the point that we're doing every query in reverse. We're traversing a graph and we're trying to get all edges going from a thing, and instead we're querying for everything which has an edge _to_ it from the thing.. That just seems backwards

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
Alfred-Mountfield: Answered in https://github.com/hashintel/hash/pull/2223#discussion_r1144646196 and https://github.com/hashintel/hash/pull/2223#discussion_r1144652948

Alfred-Mountfield (approved):
I'm approving mostly because we've ended up hotfixing entity traversal which is completely the wrong way round at the moment in `main` and we should get that fixed ASAP. I still feel like we are not ending up with intuitive traversal code, and I'd like to see us improve it, but we can work on that towards the end of this stack of PRs when more of the generalisation is in place

PR #2224, opened 2023-03-17
Upgrade TypeScript to version 5

Alfred-Mountfield:

A new version of TypeScript [has been released](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0). This PR upgrades all of our `typescript` dependencies to `5.0.2`.

~~There apparently are no breaking changes for us, and no code changes were required.~~
There were a few minor changes that needed to be made, described below.

As a driveby I have downgraded the `shuffle` block's `typescript` dependency to a `devDependency` as it seems to work without it.

We can look at using new language features, module resolutions, etc. as the need becomes apparent.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1204203029155915/f) _(internal)_

## 🚫 Blocked by

- [x] ~~Fixing Storybook deployments~~

## 🔍 What does this change?

- [Upgrade typescript to 5](https://github.com/hashintel/hash/pull/2224/commits/018acf02181a6b8bc6125ce7c12e667afb8d707b)
- [Driveby: improve exclude patterns for changeset config](https://github.com/hashintel/hash/pull/2224/commits/53dddabee009bb3324972170153eb1336c9ed5d6)
- Fixed new errors based on stricter comparisons between numbers and strings
  - [Handle string/number comparisons](https://github.com/hashintel/hash/pull/2224/commits/b75ea1ef2f29f1e64c143931f7af96a85d9edfc9)
- And fixes recurring bugs that seemed to have resurfaced with the new version
 - [Fix errors in entity-store-plugin.ts](https://github.com/hashintel/hash/pull/2224/commits/d7f79f308f8c2ec88610a53b315189731b4c4369) 
 - We have encountered weird errors about `T is not assignable to WritableDraft<T>` before, and I found [some cast utilities recommended in the immer docs](https://immerjs.github.io/immer/typescript/#cast-utilities)

- No

## ⚠️ Known issues

- The current version of Storybook doesn't support TS 5, we could upgrade to the RC of 7 where the issue is apparently fixed
  - https://github.com/storybookjs/storybook/issues/21276
  - https://github.com/storybookjs/storybook/issues/21642

## 🐾 Next steps

- Use new TS features as they become apparently useful

## 🛡 What tests cover this?

- linting
- integration tests
- playwright tests

## ❓ How to test this?

- Check CI is green
- Try and run some blocks in dev, and build them
- Try and run the frontend and backend in dev, try and build them

## 📹 Demo

See CI

### DIFF
--- a/apps/hashdotdev/package.json
+++ b/apps/hashdotdev/package.json
@@ -61,6 +61,6 @@
     "@types/unist": "2.0.6",
     "babel-loader": "^9.1.0",
     "eslint": "8.33.0",
-    "typescript": "5.0.2"
+    "typescript": "5.0.4"
   }
 }
--- a/apps/hashdotdev/package.json
+++ b/apps/hashdotdev/package.json
@@ -61,6 +61,6 @@
     "@types/unist": "2.0.6",
     "babel-loader": "^9.1.0",
     "eslint": "8.33.0",
-    "typescript": "4.9.4"
+    "typescript": "5.0.2"
   }
 }
--- a/libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
+++ b/libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
@@ -1,5 +1,5 @@
 import { EntityId, EntityPropertiesObject } from "@local/hash-subgraph";
-import { Draft, produce } from "immer";
+import { castDraft, Draft, produce } from "immer";
 import { isEqual } from "lodash";
 import { Node } from "prosemirror-model";
 import { EditorState, Plugin, PluginKey, Transaction } from "prosemirror-state";
@@ -162,6 +162,9 @@ const updateEntitiesByDraftId = (
       // @ts-ignore
       const blockChildEntity = entity.blockChildEntity!;
       if (blockChildEntity.draftId && blockChildEntity.draftId === draftId) {
+        // This type is very deep now, so traversal causes TS to complain.
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
         entities.push(blockChildEntity as DraftEntity);
       }
     }
@@ -209,7 +212,7 @@ const setBlockChildEntity = (
       // updatedAt: targetEntity.updatedAt,
     };
 
-    draftEntityStore[targetEntityDraftId] = targetDraftEntity;
+    draftEntityStore[targetEntityDraftId] = castDraft(targetDraftEntity);
   }
 
   const draftBlockEntity = draftEntityStore[blockEntityDraftId];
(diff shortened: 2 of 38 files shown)


libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
```diff
@@ -162,6 +162,9 @@ const updateEntitiesByDraftId = (
       // @ts-ignore
       const blockChildEntity = entity.blockChildEntity!;
       if (blockChildEntity.draftId && blockChildEntity.draftId === draftId) {
+        // This type is very deep now, so traversal causes TS to complain.
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Alfred-Mountfield, line 167: @thehabbos007 this appeared again in another place :sob:

libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
```diff
@@ -209,7 +212,7 @@ const setBlockChildEntity = (
       // updatedAt: targetEntity.updatedAt,
     };
 
-    draftEntityStore[targetEntityDraftId] = targetDraftEntity;
+    draftEntityStore[targetEntityDraftId] = castDraft(targetDraftEntity);
```
Alfred-Mountfield, line 215: Was getting more stranger errors about `T` not being assignable to `WritableDraft<T>` again. But I discovered some utilities in the immer docs that _seem_ to get rid of this. https://immerjs.github.io/immer/typescript/#cast-utilities

nathggns (approved):
Thank you for this Alfie!

apps/hashdotdev/package.json
Alfred-Mountfield, line 1: Commenting here as `hashdotdesign` doesn't have changes..

Previously this PR was blocked as Storybook 6.x doesn't support TS 5, and Storybook 7.x hasn't had a stable release yet.
  - https://github.com/storybookjs/storybook/issues/21276
  - https://github.com/storybookjs/storybook/issues/21642

The changes made that split the storybook `app` from the `libs` seem to have hidden this issue. I am assuming because `typescript` is not an _explicit_ dependency of `hashdotdesign`, so I would guess it's actually using TS 4.x, but our version consistency lint isn't complaining so 🤷 

So:
- Do I make `typescript` and explicit dependency of `hashdotdesign` and then apply a hotfix (https://github.com/hipstersmoothie/react-docgen-typescript-plugin/issues/78#issuecomment-1409224863) to see if I can fix it
- Do I leave `typescript` as an implicit dependency, presumably outdated, but avoid needing to introduce and track the hotfix, as everything is working right now

thehabbos007: Given that the staging environment for `hashdotdesign` is passing/working https://vercel.com/hashintel/hashdotdesign/EM1BEmCEVRwKoeoHtFsztwnK862X I think it's fine to leave it as an implicit dependency. I have a feeling going with the explicit dependency would cause some annoying problems, and we can just prevent needing to add the hotfix as a patch

Alfred-Mountfield: I dislike the idea of us having mismatched TypeScript dependencies within the project, but I agree with this. I think let's not make it a problem until it becomes a problem 😅

thehabbos007 (approved):

PR #2225, opened 2023-03-17
Address Block - Configure button

luisbettencourt:

This PR adds a configure button to the header of the address tank.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204086167143843/f) _(internal)_

## 🔍 What does this change?

- Added a `Block settings` button to the design system;
- Used it in the Address block to toggle the settings visibility for mobile screens;
- Replaced the block editable fields with the component from the design system;
- Moved the `Get help` button to the design system to uniformize it accross blocks.

## 📹 Demo

demo


libs/@hashintel/design-system/src/configure-block-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+            color: palette.gray[40],
+          },
+          [`.${buttonClasses.endIcon}`]: {
+            transform: `rotate(${expanded ? 90 : 0}deg)`,
+          },
+          ...(Array.isArray(sx) ? sx : [sx]),
+        }),
+      ]}
+    >
+      Configure block
```
nonparibus, line 52:
```suggestion
      Configure block
```

@luisbettencourt Sorry for flip-flopping on this, but can we change this to read "Block settings"?

"Configure" is an action word, but it's quite possible the user will just want to _inspect_ the contents, and not actually modify it. "Block settings" makes it more neutral. Thank you/sorry!

We should probably rename the file, consts, etc too.

nonparibus: LGTM, thanks. Over to code review

nonparibus: Belatedly now realizing... hadn't we originally said it would only show as a dropdown on mobile, @luisbettencourt?

If there are _lots_ of items in block settings we could keep the dropdown on desktop, but provided they all fit on one row, we should just show them to make the options more discoverable, and save people the extra click.

Visibly highlighting the underlying services in use within any given block is almost certainly going to be encouraged as general best practice, so let's do our best to abide by it here in anticipation.

luisbettencourt: The `Block settings` button is being displayed when the block container width is less than 800px. 
However, blocks in the WP editor have a max-width of 650px, which means it will always be displayed there. 

responsiveness

nonparibus: Is there anything we can do make sure the `Block settings` button is _not_ always hidden in WordPress?

We want it to be visible as much of the time as possible and _only_ collapsed when it would otherwise bleed onto multiple lines.

It seems in the above example it is breaking overly-early. Instead, it should break dependent on the width of its contained content.

Likewise, it seems like we break for the "Enter an address" copy in the input overly prematurely.

luisbettencourt: Reduced the breakpoint to `620px` (which makes it not visible by default in WP):

responsiveness

Also made it so the input placeholder changes for a smaller breakpoint.

nonparibus: Looks better, thank you. You mean the full bar _is now_ visible by default in WP, though, or it's still not?

PR #2229, opened 2023-03-17
Update resolve depths based on edge kind

TimDiekmann:

Currently, we update resolve depths by hand. This can be done in a centralized place, so the functions to decrement and check the resolve depth per edge kind does not need to be inlined at the traversal logic.

## 🔗 Related links

## 🚫 Blocked by

- #2207 
- #2208 
- #2223 
- #2226 

## 🔍 What does this change?

- Add accessor for `GraphResolveDepths
- Use the function in traversal logic


thehabbos007 (approved):
This is a good addition to prevent accidental/incorrect decrements. Thank you! I checked through the diff to see if we were decrementing the right depth, looks good to me!

PR #2230, opened 2023-03-17
Deduplicate graph traversal for entities

TimDiekmann:

The logic for traversing entities is only different in its parameters, so it can be deduplicated

## 🔗 Related links

## 🚫 Blocked by

- #2207 
- #2208 
- #2223 
- #2226 
- #2229 

## 🔍 What does this change?

- Move traversal logic for entities to function
- Fix a bug to correctly resolve the edge interval. This is not necessarily the link's interval but the intersection of both endpoints. We did never encounter issues so far as without deletion it's not possible to construct this case.

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
@@ -11,7 +11,7 @@ use uuid::Uuid;
 use crate::{
     identifier::{
         knowledge::{EntityEditionId, EntityId, EntityRecordId, EntityTemporalMetadata},
-        time::{DecisionTime, Timestamp},
+        time::{DecisionTime, LeftClosedTemporalInterval, Timestamp},
     },
     knowledge::{Entity, EntityLinkOrder, EntityMetadata, EntityProperties, EntityUuid, LinkData},
     ontology::EntityTypeWithMetadata,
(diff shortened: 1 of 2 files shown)


apps/hash-graph/lib/graph/src/subgraph/edges/edge.rs
```diff
@@ -1,16 +1,48 @@
...
+use serde::{ser::SerializeStruct, Serialize, Serializer};
 use utoipa::{openapi, ToSchema};
 
-#[derive(Debug, Hash, PartialEq, Eq, Serialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[derive(Debug, Hash, PartialEq, Eq)]
 pub struct OutwardEdge<K, E> {
     pub kind: K,
     /// If true, interpret this as a reversed mapping and the endpoint as the source, that is,
     /// instead of Source-Edge-Target, interpret it as Target-Edge-Source
```
Alfred-Mountfield, line 8: Needs updating

TimDiekmann: Thanks, removed in [Remove outdated comment](https://github.com/hashintel/hash/pull/2230/commits/7a2c56ddc8fbb51f0baedeb2afaadd757fd3e7b4) and added a doc string in [Add comment to EdgeDirection](https://github.com/hashintel/hash/pull/2230/commits/ddfa958cd8ef57d14382f4f3a009d187a30b3392)

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -27,12 +27,81 @@ use crate::{
...
+                let edge_interval = edge_entity
+                    .metadata
+                    .temporal_versioning()
+                    .variable_time_interval(time_axis)
+                    .intersect(entity_interval)
+                    .unwrap_or_else(|| {
+                        unreachable!("The edge interval and the entity interval do not overlap")
+                    });
+
+                // reversed `HasLeftEntity` is equivalent to an outgoing link `Entity`
```
Alfred-Mountfield, line 81: Can we just delete this?

TimDiekmann:
```suggestion
```

Alfred-Mountfield (approved):

PR #2233, opened 2023-03-18
Upload clippy report to GitHub

TimDiekmann:

Instead of just failing the CI, it's possible to parse the output of `cargo clippy` into a SARIF file and upload it to GitHub.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1203358502199168/f) _(internal)_

## 🔍 What does this change?

- Installs the tools for converting clippy outputs to SARIF
- Post-process the SARIF file to make the paths relative to the repository root
- Upload the SARIF file to GH

## ⚠️ Known issues

Clippy may run multiple times for a crate, depending on the configurations. Thus, the output should be deduplicated.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,8 +105,6 @@ jobs:
             | tee clippy.sarif \
             | sarif-fmt
 
-          jq -e '.runs[].results == []' clippy.sarif> /dev/null
-
       - name: Process SARIF file
         working-directory: ${{ matrix.directory }}
         if: always()
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -100,11 +100,12 @@ jobs:
           cargo +${{ matrix.toolchain }} make clippy --message-format=json \
             | clippy-sarif \
             | jq '.runs[].results |= unique' \
-            | jq --arg pwd "$(pwd)" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .'  \
+            | jq --arg pwd "${{ matrix.directory }}" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .'  \
             > clippy.sarif
 
           if [[ $(jq '.runs[].results | length' clippy.sarif) -gt 0 ]]; then
-              sarif-fmt --input clippy.sarif
+            sarif-fmt --input clippy.sarif
+            exit 1
           fi
 
       - name: Upload SARIF file
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -110,8 +110,9 @@ jobs:
         if: always()
         run: |
           cat clippy.sarif \
-           | jq --arg pwd "${{ matrix.directory }}" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .' \
-           > clippy.sarif
+            | jq --arg pwd "${{ matrix.directory }}" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .' \
+            > clippy.sarif.tmp
+          mv clippy.sarif.tmp clippy.sarif
 
       - name: Upload SARIF file
         uses: github/codeql-action/upload-sarif@v2
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -97,12 +97,15 @@ jobs:
         working-directory: ${{ matrix.directory }}
         if: always()
         run: |
-          cargo +${{ matrix.toolchain }} make clippy --message-format=json \
+          cargo +${{ matrix.toolchain }} make clippy --message-format=json -- -D warnings \
             | clippy-sarif \
             | jq '.runs[].results |= unique' \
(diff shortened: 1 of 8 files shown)


apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -51,7 +51,7 @@
 
         let mut queue = entity_vertex_ids
             .into_iter()
-            .map(|id| (id, graph_resolve_depths, temporal_axes.clone()))
+            .map(|id| (id, graph_resolve_depths, temporal_axes))
```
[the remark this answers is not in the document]
TimDiekmann:
```suggestion
            .map(|id| (id, graph_resolve_depths, temporal_axes.clone()))
```

.github/workflows/rust.yml
```diff
@@ -96,7 +96,30 @@ jobs:
...
         if: always()
-        run: cargo +${{ matrix.toolchain }} make clippy -- -D warnings
+        run: |
+          cargo +${{ matrix.toolchain }} make clippy --message-format=json \
+            | clippy-sarif \
+            | jq '.runs[].results |= unique' \
+            | tee clippy.sarif \
+            | sarif-fmt
+
+          jq -e '.runs[].results == []' clippy.sarif> /dev/null
```
TimDiekmann, line 106: This ensures, that the results of the run are empty, so there are no warnings as we (most probably) want to continue require to have all warnings addressed.

.github/workflows/rust.yml
```diff
@@ -96,7 +96,30 @@ jobs:
...
+
+      - name: Process SARIF file
+        working-directory: ${{ matrix.directory }}
+        if: always()
+        run: |
+          cat clippy.sarif \
+            | jq --arg pwd "${{ matrix.directory }}" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .' \
+            > clippy.sarif.tmp
+          mv clippy.sarif.tmp clippy.sarif
```
TimDiekmann, line 115: We're running clippy in `matrix.directory`, so the files are relative to that location. By pretending this path to that location the SARIF file will be relative to the repository root.

.github/workflows/rust.yml
```diff
@@ -96,7 +96,30 @@ jobs:
       - name: Check clippy
         working-directory: ${{ matrix.directory }}
         if: always()
-        run: cargo +${{ matrix.toolchain }} make clippy -- -D warnings
+        run: |
+          cargo +${{ matrix.toolchain }} make clippy --message-format=json \
+            | clippy-sarif \
+            | jq '.runs[].results |= unique' \
```
indietyp, line 102: With https://github.com/psastras/sarif-rs/pull/429 (hopefully) landing soon, what would happen if `relatedLocations` is different but the error is the same? I imagine in that case, we would emit the error twice.

This scenario is more likely than one might imagine, e.g., if we have conditional code that emits a related error.

.github/workflows/rust.yml
indietyp, line 106:
```suggestion
          jq -e 'isempty(.runs[].results)' clippy.sarif > /dev/null
```

indietyp: To clarify the `-e` flag:

> Sets the exit status of jq to 0 if the last output values was neither false nor null, 1 if the last output value was either false or null, or 4 if no valid result was ever produced. Normally jq exits with 2 if there was any usage problem or system error, 3 if there was a jq program compile error, or 0 if the jq program ran.

* Exit Code 0: Success (if empty)
* Exit Code 1: Error (if `results` has at least one element)

.github/workflows/rust.yml
indietyp, line 115:
```suggestion
          tmp=$(mktemp)
          
          jq --arg pwd "${{ matrix.directory }}" '.runs[].results[].locations[].physicalLocation.artifactLocation.uri |= $pwd + "/" + .' clippy.sarif > "$tmp"
          
          mv "$tmp" clippy.sarif
```

.github/workflows/rust.yml
indietyp: How about leveraging `reduce` and merging if they have the same `ruleId`?

I think even better would be the use of [`@microsoft/sarif-multitool`](https://github.com/microsoft/sarif-sdk/blob/main/docs/multitool-usage.md) which already has a merge operation.

indietyp (changes requested):

.github/workflows/rust.yml
TimDiekmann: If the related locations are different it's a different error. This is purely to deduplicate errors that are exactly the same because e.g. clippy runs with a different permutation of features. I don't see how the tool you linked would help us here. The only real use case I see is adjusting the file paths, but it's not worth it for a one-liner to be replaced with a tool.

.github/workflows/rust.yml
TimDiekmann: This is not working as expected, I reverted the change.

indietyp: Good call, `isempty` did not work as I expected :/

indietyp (approved):

indietyp (approved):

PR #2234, opened 2023-03-19
Create `Kanban Board` block working with local data

yusufkinatas:

This PR creates initial version of the Kanban block. It only works with local data.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204160628104410/f) _(internal)_
- [Design](https://www.figma.com/file/T10d2P2iNb1LdRC5a9J7vA/Blocks?node-id=1740-5423&t=rj4rYDVuiRCmIEEg-4) _(internal)_

## ⚠️ Known issues

- If `updateEntity` call takes a bit longer (e.g. while testing the block inside WordPress), and a consecutive update happens right after the `updateEntity` gets called, the update made by second debounced `updateEntity` call gets overrided by the previous one.  (check the video below to see how it looks)

*This issue can be reproduced by setting `simulateDatastoreLatency = {{ min: 500, max: 500  }}`* on `MockBlockDock`

https://user-images.githubusercontent.com/39553853/226212142-5b85e698-36e3-411c-9190-0fb7636975d6.mov

https://github.com/hashintel/hash/blob/6082cee5b413364b04bfb4bba067726ff32f86e9/blocks/kanban/src/components/board/board.tsx#L103-L115

*ℹ️ This code runs on each render to make sure `Board.tsx` respects changes of `blockEntity` props coming from `App.tsx`. (like [this example from official React docs](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes))*

And commenting out the code above prevents this issue. This issue probably happens by these steps

- `updateEntity` runs, and updates the `blockEntity`
- Because `blockEntity` is updated, the code above works
- But  in the meantime, user triggered another `updateEntity` call, and because `updateEntity` calls are debounced at `syncLocalStateToEntity` function, it runs with a delay, and probably works with stale data somehow 
  - This step definitely needs debugging. Maybe the [hacky](https://github.com/hashintel/hash/blob/6082cee5b413364b04bfb4bba067726ff32f86e9/blocks/kanban/src/components/board/board.tsx#L70-L101) `useState` usage at `syncLocalStateToEntity` causes this issue

## 🐾 Next steps

- [x] Fix known issue mentioned above (solution discussion on [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1679390791856139) _(internal)_)
- [x] Update block preview image (cc @nonparibus)
- [x] Write titles and descriptions for block `entity type` & `property types`
- [x] Update `README.md` (cc @nonparibus)

## ❓ How to test this?

1.  Start Kanban block in development mode
1.  Confirm that it works as expected, and looks like the design.

## 📹 Demo

https://user-images.githubusercontent.com/39553853/226211794-9475705e-3bd4-46c6-9344-5c4119c25850.mov


blocks/kanban-board/src/components/board/card/card.tsx
```diff
@@ -0,0 +1,61 @@
...
+    >
+      <EditableCardContent
+        content={data.content}
+        onChange={(val) => updateCardContent(data.id, val)}
+        readonly={readonly}
+      />
+      {!readonly && (
+        <IconButton onClick={onDelete}>
+          <DiscardIcon />
+        </IconButton>
```
luisbettencourt, line 57: I would move this to inside `EditableCardContent` so there is no need to repeat it in `StaticCard`

blocks/kanban-board/src/components/board/column/column.tsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    <div
+      className={styles.wrapper}
+      ref={setNodeRef}
+      style={style}
+      {...attributes}
+    >
+      <div
```
luisbettencourt, line 62: Would it be possible to make this only a wrapper component for the drag and drop logic? That way `StaticColumn` could be a pure component and this could wrap it

blocks/kanban-board/src/components/board/card/editable-card-content/styles.module.scss
```diff
@@ -0,0 +1,11 @@
+.wrapper {
+  border: none;
```
luisbettencourt, line 2: According to the design this should also have `font-weight: 500` right?

luisbettencourt (changes requested):
Great work @yusufkinatas. Made some small suggestions but overall the code is looking good. The block works but slows down a bit when dragging cards between columns, is it possible something is causing unecessary re-renders for some reason?

blocks/kanban-board/src/components/board/card/card.tsx
yusufkinatas: https://github.com/hashintel/hash/pull/2234/commits/3afbe6abb587665457717a7cce0ad33f7e84d99c Moved it here, also renamed `EditableCardContent` -> `CardContent` since it also has button now.

blocks/kanban-board/src/components/board/card/editable-card-content/styles.module.scss
yusufkinatas: Right, updated it here https://github.com/hashintel/hash/pull/2234/commits/46cf9eef97f0e42071947a4d219fe3f46ddd9754

blocks/kanban-board/src/components/board/column/column.tsx
yusufkinatas: Hmm, `attributes` and `listeners` are spreaded into different elements, wouldn't that cause a problem?

blocks/kanban-board/src/components/board/column/column.tsx
```diff
@@ -0,0 +1,109 @@
...
+        />
+        {!readonly && (
+          <IconButton onClick={() => deleteColumn(data.id)}>
+            <DiscardIcon />
+          </IconButton>
+        )}
+      </div>
+      <div className={styles.body}>
+        <SortableContext
+          items={data.cards.map((card) => card.id)}
```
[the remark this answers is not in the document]
yusufkinatas: https://github.com/hashintel/hash/pull/2234/commits/a3b937870ed6a97f852acabbcf2746edf876ccd0 moved here!

blocks/kanban-board/src/components/board-title/board-title.tsx
```diff
@@ -0,0 +1,85 @@
+import clsx from "clsx";
```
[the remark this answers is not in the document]
nonparibus: Yes, please. Let's do that.

blocks/kanban-board/src/components/board/column/column.tsx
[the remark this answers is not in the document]
luisbettencourt: This change broke the drag preview.
bug

This is because the recalculation isn't happening every time the order of the cards change, you need to set the dependency array to `data.cards` instead

blocks/kanban-board/src/components/board/column/column.tsx
luisbettencourt: Just the 2 outside divs have those right? You can extract what's inside

blocks/kanban-board/src/components/board-title/board-title.tsx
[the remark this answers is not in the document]
luisbettencourt: Same for the text inside cards?

blocks/kanban-board/src/components/board/column/column.tsx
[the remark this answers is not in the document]
yusufkinatas: Thanks for noticing this, fixed here https://github.com/hashintel/hash/pull/2234/commits/4daed043eb7ef89b4403fc8f528ce4da83e3ce60

blocks/kanban-board/src/components/board-title/board-title.tsx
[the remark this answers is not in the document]
nonparibus: Yes, please!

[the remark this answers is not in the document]
yusufkinatas: Created a [task](https://app.asana.com/0/1204000740778938/1204256944614937/f) _(internal)_ for this, since it needs a bit fine-tuning with the styling of the `EditableField`.

blocks/kanban-board/src/components/board/column/column.tsx
yusufkinatas: Updated here https://github.com/hashintel/hash/pull/2234/commits/c5cdbc91eee25910996f76c1ae702c25f34305b4

luisbettencourt (changes requested):
Good work @yusufkinatas, thanks for addressing all the requested changes. Just need replacing the types now.

luisbettencourt (approved):

nonparibus (approved):

PR #2236, opened 2023-03-20
Add conflict behavior when batch-inserting ontology types

TimDiekmann:

We have a potential data race when inserting external types as a (direct or indirect) dependency either another ontology type or an entity to be inserted. Precisely, this can happen if two operations try to insert the same external dependency at the same time. One insertion will fail with a duplicate key violation.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1203958456055107/f) _(internal)_

## 🔍 What does this change?

- Adds `ConflictBehavior` to batch insertion methods_

## 🚫 Blocked by

- [x] This is currently a POC. The behavior for `ConflictBehavior` still needs to be implemented


apps/hash-graph/tests/integration/postgres/lib.rs
```diff
@@ -137,7 +137,9 @@ impl DatabaseTestWrapper {
 
             (data_type, metadata)
         });
-        store.create_data_types(data_types_iter).await?;
+        store
+            .create_data_types(data_types_iter, ConflictBehavior::Skip)
```
thehabbos007, line 141: Was `skip` chosen for the integration tests for a reason? I'm just curious :)

thehabbos007 (dismissed):
Looks good to me, thanks!

apps/hash-graph/tests/integration/postgres/lib.rs
TimDiekmann: Yes, it was quite annoying that I had to wipe the database every time I tried running the test. The possibility to skip those was quite nice.

thehabbos007: That's actually a pretty good reason, and something I've experienced as well. Thanks for explaining!

thehabbos007 (dismissed):

thehabbos007 (approved):

PR #2237, opened 2023-03-20
Reserve `sarif` crate

TimDiekmann:
## 🌟 What is the purpose of this PR?

As discussed offline, this reserves the `sarif` name on crates.io.

## 🔍 What does this change?

- Adds boilerplate to define a crate in HASH

## 🛡 What tests cover this?

- /s The crate currently has 💯% line coverage _and_ branch coverage

## ❓ How to test this?

- Verify the links in the README and potential typos as this will be published.
- /s Run the full test suite, this may take up to 0 seconds

PR #2238, opened 2023-03-20
Only run semver check if it's not an initial version

TimDiekmann:

The publish job fails if an initial version of a crate is published. This PR changes the behavior to only run the check, if the version is not `0.0.0-reserved`, which cover all current and planned cases

## 🔗 Related links

- #2237

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -336,7 +336,10 @@ jobs:
       - name: SemVer check
         if: ${{ github.event_name == 'pull_request' || github.event_name == 'merge_group' }}
         working-directory: ${{ matrix.directory }}
-        run: cargo +${{ matrix.toolchain }} semver-checks check-release
+        run: |
+          # Only run semver checks if this is not the initial release
+          cargo +${{ matrix.toolchain }} metadata --format-version 1 | jq -e '.packages[].version != "0.0.0-reserved"' > /dev/null || echo exit 0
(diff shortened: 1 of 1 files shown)


indietyp (changes requested):
This won't work
1. it includes all dependencies (use `--no-deps` to include them)
2. This includes all workspace members, `error-stack-macros` is `0.0.0-reserved`, meaning that checks will never run for `error-stack`.

The snippet below might make it a bit easier to decipher what we're doing, as the `echo exit 0` is a bit confusing. The snippet below also checks that all workspace members are **not** reserved.

```bash
has_reserved=$(cargo metadata --no-deps --format-version=1 | jq '[.packages[].version | select(. == "0.0.0-reserved")] | length > 0')

if [ $(has_reserved) = false ]; then
	 cargo +${{ matrix.toolchain }} semver-checks check-release
fi
```

(because the condition is untested, I did not add it as a snippet)

Another possibility would be to check that every workspace member is reserved:

```
cargo metadata --no-deps --format-version=1 | jq '([.packages[].version | select(startswith("0.0.0"))] | length) == (.packages | length)'
```

(`startswith` here due to `deer-desert` crate)

TimDiekmann:
Thanks, @indietyp, good flag!

We discussed offline to use `grep` instead as `setup.py` will determine the working directory by the `Cargo.toml` which is going to be published.

PR #2246, opened 2023-03-21
Support `version` and `semantic_version` in `ToolComponent`

TimDiekmann:

This further extends the `ToolComponent` struct

## 🚫 Blocked by

- #2243
- #2245

## 🔍 What does this change?

- Add `ToolComponent::version`
- Add `ToolComponent::semantic_version`
- Add builder methods and documentation
- Add to the `full` test

## 🛡 What tests cover this?

- This branch has 💯% test coverage

### DIFF
--- a/libs/sarif/src/schema/log.rs
+++ b/libs/sarif/src/schema/log.rs
@@ -312,8 +312,10 @@ pub(crate) mod tests {
         validate_schema(
             &SarifLog::new(SchemaVersion::V2_1_0).with_runs([
                 Run::new(
-                    Tool::new(ToolComponent::new("prettier"))
-                        .with_extension(ToolComponent::new("prettier-plugin-sql"))
+                    Tool::new(ToolComponent::new("prettier").with_version("2.8.2"))
+                        .with_extension(
+                            ToolComponent::new("prettier-plugin-sql").with_version("0.12.1"),
+                        )
                         .with_properties(|properties| {
                             properties
                                 .with_tag("format")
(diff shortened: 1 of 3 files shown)


libs/sarif/src/schema/log.rs
indietyp, line 1: I like how you're using a realist example for full! :D

libs/sarif/src/schema/tool.rs
```diff
@@ -149,6 +149,14 @@ impl Tool {
 pub struct ToolComponent {
```
indietyp, line 149: `3.19.2`: "At least one of version (§3.19.13) and semanticVersion (§3.19.12) SHOULD be present" this invariant is not upheld in the constructor (or by having publicly accessible fields)

indietyp (changes requested):

libs/sarif/src/schema/tool.rs
indietyp: This is wrong, I (once again), have mistaken `SHOULD` for `SHALL` (which while being very similar have a totally different meaning). So we do not need to ensure that at least one of them exists. My bad! I am fine with the current version!

indietyp (approved):

libs/sarif/src/schema/log.rs
TimDiekmann: Actually, I haven't found any examples, but would `prettier-plugin-sql` be an extension?

indietyp: I think so, yes.

indietyp (approved):

PR #2248, opened 2023-03-21
`antsi`: feature gate extensions

indietyp:

This PR adds several feature gates for commonly used extra extensions to the standard that are not supported by all terminals. These are:

* `rgba`
* `underline-color`
* `underline-variants`
* `overstrike`
* `script`

### DIFF
--- a/libs/antsi/src/font.rs
+++ b/libs/antsi/src/font.rs
@@ -112,11 +112,42 @@ pub enum FontScript {
     Super,
 }
 
+macro_rules! impl_builder {
+    (
+        $(#[$meta_set:meta])*
+        $set:ident($set_var:ident : $set_ty:ty);
+        $(#[$meta_with:meta])*
+        $with:ident($with_var:ident : $with_ty:ty);
+        $(#[$meta_getter:meta])*
+        $get:ident() -> $ret:ty
+    ) => {
+        $(#[$meta_set])*
+        pub fn $set(&mut self, $set_var: $set_ty) -> &mut Self {
+            self.$set_var = Some($set_var);
+            self
+        }
+
+        $(#[$meta_with])*
+        #[must_use]
+        pub const fn $with(mut self, $with_var: $with_ty) -> Self {
+            self.$with_var = Some($with_var);
+            self
+        }
+
+        $(#[$meta_getter])*
+        #[must_use]
+        pub const fn $get(&self) -> $ret {
+            self.$with_var
+        }
+    };
+}
+
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Default)]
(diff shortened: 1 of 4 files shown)


libs/antsi/src/font.rs
```diff
@@ -106,25 +109,57 @@ pub enum Blinking {
...
+        $(#[$meta_set])*
+        pub fn $set(&mut self, $set_var: $set_ty) -> &mut Self {
+            self.$set_var = Some($set_var);
+            self
+        }
+
+        $(#[$meta_with])*
+        #[must_use]
+        pub const fn $with(mut self, $with_var: $with_ty) -> Self {
+            self.$with_var = Some($with_var);
```
TimDiekmann, line 137: As proposed above, this should be implemented without a macro, but if we would keep the setters, can't you re-use those?
```rust
self.$set($with_ty);
self
```

libs/antsi/src/font.rs
```diff
@@ -106,25 +109,57 @@ pub enum Blinking {
...
+        $(#[$meta_with:meta])*
+        $with:ident($with_var:ident : $with_ty:ty);
+        $(#[$meta_getter:meta])*
+        $get:ident() -> $ret:ty
+    ) => {
+        $(#[$meta_set])*
+        pub fn $set(&mut self, $set_var: $set_ty) -> &mut Self {
+            self.$set_var = Some($set_var);
+            self
+        }
```
TimDiekmann, line 132: There may be situations where setters are useful, but we probably don't want to provide two different builder patterns at the same time. Also, it's unintuitive, that the setters implies `Some(value)`, so it's not possible to `unset()`.

TimDiekmann (changes requested):

indietyp:
Okay then I'll go with the pattern proposed in #2243, public fields as we have no invariants to uphold with `with_` methods.

TimDiekmann:
> Okay then I'll go with the pattern proposed in #2243, public fields as we have no invariants to uphold with `with_` methods.

Sounds good, thank you!

TimDiekmann:
I think we can just remove the setters/getters added in this PR and move to the other approach in a follow up, then it's easier to review/proceed

libs/antsi/src/font.rs
indietyp: macro removed in [9b2891a](https://github.com/hashintel/hash/pull/2248/commits/9b2891afbf6365b877c748d25cecf78e7f5ef90e)

indietyp:
getter changes have been removed, will open PR after this one to address

TimDiekmann (approved):

TimDiekmann:
Great, thank you very much!

PR #2250, opened 2023-03-21
`antsi`: rework getter

indietyp:

Remove all getter and setter (prefixed with `set_`), adhering to the builder-lite principle. Since we do not need to uphold any invariants on the struct, all fields have been marked pub, the only exception is `Font::style`, which is a bitfield and therefore needs special methods for access. These have been unified and made more readable.

### DIFF
--- a/libs/antsi/src/font.rs
+++ b/libs/antsi/src/font.rs
@@ -208,6 +208,7 @@ impl Font {
         self
     }
 
+    #[cfg(feature = "script")]
     #[must_use]
     pub const fn with_script(mut self, script: FontScript) -> Self {
         self.script = Some(script);
--- a/libs/antsi/src/font.rs
+++ b/libs/antsi/src/font.rs
@@ -131,13 +131,13 @@ impl FontStyle {
     #[must_use]
     const fn mask(self) -> u8 {
         match self {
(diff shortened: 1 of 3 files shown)


libs/antsi/src/font.rs
```diff
@@ -116,13 +116,40 @@ pub enum FontScript {
...
+    #[must_use]
+    const fn mask(self) -> u8 {
+        match self {
+            Self::Strikethrough => 0b0000_0001,
+            Self::Hidden => 0b0000_0010,
+            Self::Inverse => 0b0000_0100,
+            Self::Italic => 0b0000_1000,
+            #[cfg(feature = "overstrike")]
+            Self::Overstrike => 0b0001_0000,
+            Self::Overline => 0b0010_0000,
```
TimDiekmann, line 140:
```suggestion
            Self::Strikethrough => 1 << 0,
            Self::Hidden => 1 << 1,
            Self::Inverse => 1 << 2,
            Self::Italic => 1 << 3,
            #[cfg(feature = "overstrike")]
            Self::Overstrike => 1 << 4,
            Self::Overline => 1 << 5,
```

TimDiekmann (approved):

PR #2251, opened 2023-03-21
Add number of images to generate to `AI Image` block

luisbettencourt:

This PR adds the functionality to select the number of images to generate in the AI Image block and also to generate extra images after the prompt has been set.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204160628104433/f) _(internal)_
- [Asana task](https://app.asana.com/0/1204000740778935/1204086167143645/f) _(internal)_

## 🔍 What does this change?

- Made the model selector component in the AI Text block generic and moved it to the design system;
- Added a selector to the block to allow for selecting the number of images to initially generate;
- Added `Generate X more options` buttons to the image list;
- Added a skeleton before images load;
- Prevented images from re-rendering after being uploaded;
- Moved the error indicator in the Address block to the design system and used it for this block;
- Fixed animations for desktop and mobile when selecting an image.

## 📹 Demo

ezgif com-optimize (3)

Error states:
image

error
(color of the error is red, for some reason the GIF is picking it up as orange)


blocks/ai-image/src/app/generate-image/image-preview.tsx
```diff
@@ -212,352 +202,279 @@ export const ImagePreview = ({
...
+              <Typography
+                sx={{
+                  color: ({ palette }) => palette.gray[80],
+                  fontSize: 14,
+                  lineHeight: "18px",
+                  fontWeight: 500,
+                }}
+              >
+                {loading
+                  ? "Uploading images..."
```
nonparibus, line 462: When the text in the bottom left changes to read "Uploading images..." can we instead change it to read "Generating images..." and swap the up-arrow icon to its left for a small circular loading spinner similar to that used in https://github.com/hashintel/hash/pull/2267 (but a lot smaller)?

CiaranMn (commented):
Now that there are more cards in addition to generated images, I'm wondering if we should move the controls / instructions above the images, because you have to scroll down to figure out what to do (and also to know when images are still in the process of being uploaded to the app, during which time they are not clickable) – this would also make it more obvious what to do when a single image is selected, which is also off screen in WordPress. thoughts @nonparibus?

Screenshot 2023-03-24 at 08 49 34
Screenshot 2023-03-24 at 08 49 43

blocks/ai-image/src/app/generate-image/image-preview.tsx
nonparibus: This is perfect - thank you!

nonparibus:
Following @CiaranMn's comment above, single column designs have been produced @luisbettencourt: [AI Image Block in Figma](https://www.figma.com/file/T10d2P2iNb1LdRC5a9J7vA/Blocks?node-id=1961%3A5707&t=bCoHn8v8VKk9zW7N-1) (_internal link_)

luisbettencourt:
Made the changes to the designs, here's how it's looking like in WP:

image
image

Are we happy with this solution?

CiaranMn (approved):

PR #2256, opened 2023-03-22
Expose ontology transaction time to the graph

TimDiekmann:

The goal is to have proper testing of the returned subgraph. This implies, that a set of data is present in the database. To insert a set of data into the data the easiest way to achieve this is to import a previously exported set of data. As we, however, do want the data to be exactly the same after exporting, wiping, and importing the data, it implies, that the graph requires information about the time, when an ontology type was inserted into the database, which is recorded in `transactionTime`, a variable, which is not exposed yet at all.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204216809501000/f) _(internal)_

## 🔍 What does this change?

- Add `TransactionTime` to the ontology query paths
- Make the query compiler aware of `TransactionTime`

A doc-string has been added to the newly added query path parameters, but the field is still not exposed anywhere.

## ⚠️ Known issues

- Currently, this is only tested manually but not automatically.

## 🐾 Next steps

Please see the sibling tasks of the task, which is linked above

## 🛡 What tests cover this?

- none (yet)

## ❓ How to test this?

Add a transaction time query to the read implementation of the ontology types, print it, and make sure, it's the correct value (Should be `[now(), +∞)`)

## 📹 Demo

With the steps described above, this prints this to the terminal:
```
transaction_time: [2023-03-22 13:41:39.102656 +00:00:00, +∞)
```

### DIFF
--- a/apps/hash-graph/lib/graph/src/ontology/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/ontology/data_type.rs
@@ -83,6 +83,16 @@ pub enum DataTypeQueryPath<'p> {
     /// [`VersionedUrl`]: type_system::url::VersionedUrl
     /// [`DataType`]: type_system::DataType
     VersionedUrl,
+    /// The transaction time of the [`DataType`].
+    ///
+    /// It's not possible to query for the temporal axis directly, this has to be done via the
+    /// `temporalAxes` parameter on [`StructuralQuery`]. The transaction time is currently not part
+    /// of the [`OntologyElementMetadata`].
+    ///
+    /// [`EntityType`]: type_system::EntityType
+    /// [`OntologyElementMetadata`]: crate::ontology::OntologyElementMetadata
+    /// [`StructuralQuery`]: crate::subgraph::query::StructuralQuery
+    TransactionTime,
     /// The [`OwnedById`] of the [`OntologyElementMetadata`] belonging to the [`DataType`].
     ///
     /// ```rust
@@ -189,6 +199,10 @@ impl OntologyQueryPath for DataTypeQueryPath<'_> {
         Self::Version
     }
 
+    fn transaction_time() -> Self {
+        Self::TransactionTime
+    }
+
     fn updated_by_id() -> Self {
         Self::UpdatedById
     }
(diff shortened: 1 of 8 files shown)


apps/hash-graph/lib/graph/src/ontology/data_type.rs
```diff
@@ -83,6 +83,16 @@ pub enum DataTypeQueryPath<'p> {
     /// [`VersionedUrl`]: type_system::url::VersionedUrl
     /// [`DataType`]: type_system::DataType
     VersionedUrl,
+    /// The transaction time of the [`DataType`].
+    ///
+    /// It's not possible to query for the temporal axis directly, this has to be done via the
+    /// `temporalAxes` parameter on [`StructuralQuery`]. The transaction time is currently not part
+    /// of the [`OntologyElementMetadata`].
+    ///
+    /// [`EntityType`]: type_system::EntityType
```
Alfred-Mountfield, line 92: Should it be this?
```suggestion
    /// [`DataType`]: type_system::DataType
```

apps/hash-graph/lib/graph/src/ontology/property_type.rs
```diff
@@ -64,6 +64,16 @@ pub enum PropertyTypeQueryPath<'p> {
     /// [`PropertyType`]: type_system::PropertyType
     /// [`VersionedUrl`]: type_system::url::VersionedUrl
     VersionedUrl,
+    /// The transaction time of the [`PropertyType`].
+    ///
+    /// It's not possible to query for the temporal axis directly, this has to be done via the
+    /// `temporalAxes` parameter on [`StructuralQuery`]. The transaction time is currently not part
+    /// of the [`OntologyElementMetadata`].
+    ///
+    /// [`EntityType`]: type_system::EntityType
```
Alfred-Mountfield, line 73: Should it be this?
```suggestion
    /// [`PropertyType`]: type_system::PropertyType
```

apps/hash-graph/lib/graph/src/ontology/data_type.rs
TimDiekmann: Yes, thanks 🙃

Alfred-Mountfield (approved):

PR #2258, opened 2023-03-22
Ensure font size variables are set non-globally around type editor

nathggns:

This ensures consistent font sizing whereever the type editor is used

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204059902060151/f) _(internal)_

## 🚫 Blocked by

None

## 🔍 What does this change?

- Creates a new utility in the the design system called `fluidTypographyStyles`, which allows you to apply our fluid typography system variables to any given CSS selector, so that elements within it that refer to those variables are correctly styled. 
- New export called `fluidFontClassName` from the design system used on any container which need the elements within it to be correctly styled
- Add this class name to the top level wrapper around the type editor
- Add this class name to any popper (tooltips, modals, autocomplete, etc) used within the type editor – this is necessary as they will be rendered outside the top level wrapper around the type editor

None

## ⚠️ Known issues

We'll need discipline in applying the fluid wrapper class name to any future poppers added to the type system. 

## 🐾 Next steps

None

## 🛡 What tests cover this?

None

## ❓ How to test this?

There should be no visual changes to the type editor in HASH. This will take visual verification in Block Protocol, which you can see at https://github.com/blockprotocol/blockprotocol/pull/1209

### DIFF
--- a/.changeset/ninety-pandas-mate.md
+++ b/.changeset/ninety-pandas-mate.md
@@ -3,4 +3,4 @@
 "@hashintel/type-editor": patch
 ---
 
-Bug fixes
+Ensures consumers of the design system, and type editor, can apply global font variables using new `fluidFontClassName` export from the design system, and applies this to some elements which use portals. New `fluidTypographyStyles` export to actually set the styles for this class name


libs/@hashintel/type-editor/src/entity-type-editor/property-list-card/property-type-form/expected-value-selector/shared/delete-expected-value-modal.tsx
```diff
@@ -67,6 +68,7 @@ export const DeleteExpectedValueModal = ({
         border: 1,
         borderColor: theme.palette.gray[20],
       })}
+      classes={{ root: fluidFontClassName }}
```
CiaranMn, line 71: For `Modal`, since it's a component in the design system, could we just add the class there? Or would that mess something up in HASH?

nathggns: Done

PR #1875, opened 2023-01-18
`deer` implement `Deserialize` for built-in types (Part 2) 

indietyp:
## 🌟 What is the purpose of this PR?

Implements the remaining types from `core` (except if PR gets too big).

## 🔗 Related links

* Tracking Issue: hashintel/labs#69 

## 🚫 Blocked by

* hashintel/hash#1947 

## 🔍 What does this change?

Implementation of the standard library

## 🐾 Next steps

implement `alloc`

## 🛡 What tests cover this?

unit and integration tests that will be present

PR #2263, opened 2023-03-22
`deer`: `EnumVisitor`

indietyp:

This introduces a new visitor alongside `OptionalVisitor` and `Visitor`: `EnumVisitor`, which serves a similar purpose. The need for a specific visitor just for enums was discovered while trying to implement `Bounds` in #1875.

### Differences between 🦌 and serde

Like `OptionalVisitor` (and `FieldVisitor`), this specific visitor tries to expose the same functionality as their serde counterpart, but with a more ergonomic and easier-to-comprehend structure. This means:

* instead of a single `Visitor` trait, we split the trait into multiple categories that serve a single job
	* e.g.: `OptionalVisitor` instead of `visit_some`, `visit_none`
* trait based contracts
	* Access is done through traits (e.g. `FieldAccess`) these traits are designed in a way where it is impossible to make mistakes that are made in `serde`, e.g. calling `next_key` without `next_value`

In particular the difference between 🦌 and serde is:

calling `deserialize_enum` will do the following in `serde`:

0) user calls `deserialize_enum`
1) if possible (current value correct type) call `visit_enum` with `EnumAccess`
2) `EnumAccess` has one method `variant`, which when invoked will return `V` the discriminant as well as an implementation of `VariantAccess`
3) The caller then uses `V` to determine, which (and if) to call one of the methods on `VariantAccess`
3.1) `VariantAccess` will, if possible, process the next one in line
4) call that method, error handle and then return the value through `visit_enum`

calling `deserialize_enum` will do the following in deer:

0) user calls `deserialize_enum`
1) if possible (current value correct type) call `discriminant`, return if error
2) call `value` with the returned `discriminant` and a new deserializer, if there's no content for the key (e.g. `"UnitVariant"`) use a `NoneDeserializer` otherwise next in line
3) return the result

## 📋 ToDo

remove `visit_none` on `EnumVisitor`, we can just use the `NoneDeserializer` on the discriminant.

## 🐾 Next steps

* `FieldAccess` -> `FieldVisitor`
* explode #1875  into multiple PRs

## 🛡 What tests cover this?

Integration tests are supplied which ensure that _every_ single version that `serde` supports can be supported by 🦌 .


TimDiekmann (approved):
The code and description look good to me, thank you!

> `OptionalVisitor` instead of `visit_some`, `visit_none`

While not a huge fan of having different decisions than `serde` I really appreciate this change! IMO this was a mistake made by serde to special-case `Option`

TimDiekmann (approved):

PR #2264, opened 2023-03-23
Add `ReportingDescriptor` and `ToolComponent::rules`

TimDiekmann:

Adds a very basic `ReportingDescriptor` struct and uses it in `ToolComponent::rules`

## ⚠️ Known issues

The struct is neither complete nor is its documentation


libs/sarif/src/schema/multiformat_message_string.rs
```diff
@@ -0,0 +1,129 @@
...
+    ///
+    /// let message = MultiformatMessageString::new("mismatched types")
+    ///     .with_properties(|properties| properties.with_property("foo", "bar"));
+    ///
+    /// assert_eq!(message.properties.additional.get("foo").unwrap(), "bar");
+    /// ```
+    #[must_use]
+    pub fn with_properties(
+        mut self,
+        mut properties: impl FnMut(PropertyBag<'s>) -> PropertyBag<'s>,
```
indietyp, line 124: Why `FnMut`, shouldn't `FnOnce` be sufficient?

libs/sarif/src/schema/multiformat_message_string.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// ## Note
+    ///
+    /// This property is required to ensure that the message is viewable even in contexts that do
+    /// not support the rendering of formatted text
+    ///
+    /// See the [SARIF specification §3.12.3][spec] for more information.
+    ///
+    /// [spec]: https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317476
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub text: Cow<'s, str>,
```
indietyp, line 42: I am unsure how we want to ensure this, but the spec says it is required to be non-empty.

libs/sarif/src/schema/multiformat_message_string.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// See the [SARIF specification §3.12.4][spec] for more information.
+    ///
+    /// [GitHub-Flavored Markdown]: https://github.github.com/gfm/
+    /// [`text`]: Self::text
+    /// [spec]: https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317477
+    #[cfg_attr(
+        feature = "serde",
+        serde(borrow, default, skip_serializing_if = "Option::is_none")
+    )]
+    pub markdown: Option<Cow<'s, str>>,
```
indietyp, line 58: Same here, either `null` (`Option::None`) or a non-empty string.

libs/sarif/src/schema/reporting_descriptor.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    fn error_code_to_help(code: u32) -> io::Result<String> {
+        let explain_command = Command::new("rustc")
+            .arg("--explain")
+            .arg(format!("E{code:04}"))
+            .output()?;
+
+        Ok(String::from_utf8_lossy(&explain_command.stdout).to_string())
+    }
```
indietyp, line 386: I like this! :D

indietyp (changes requested):

libs/sarif/src/schema/multiformat_message_string.rs
TimDiekmann: I have noticed this, too, however, the JSON Schema does not require this. A non-empty string makes sense but I don't know if and how we want to ensure this. In my opinion, we should ignore the specs here for now and rely on the Schema for now, especially, as we don't have any concept of an invalid log, yet.

libs/sarif/src/schema/multiformat_message_string.rs
TimDiekmann: Good point, I fixed this in 19c7190b765577fb7f25d34bc1f15aa3754c701e

libs/sarif/src/schema/reporting_descriptor.rs
TimDiekmann: Sadly, this is nothing we can use at runtime as we don't necessarily have `rustc` in the environment, but the information should be there in the JSON output of the compiler, though.

libs/sarif/src/schema/multiformat_message_string.rs
indietyp: Could you please add a note referring to either this comment or that we don't ensure this yet? Just so we can keep track.

TimDiekmann: Added in ca908d2cff2f364e55d6a65efd7935a11e6bfa3b

indietyp (approved):

PR #2262, opened 2023-03-22
Add `<Skeleton/>` to `@hashintel/design-system` & use it for address block

yusufkinatas:
## 🌟 What is the purpose of this PR?

Implementing skeleton loading state for blocks which requires it.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204086167143851/f) _(internal)_

## 🔍 What does this change?

- Creates `Skeleton` component at `@hashintel/design-system`
- Implements loading state for `address` block

## 🐾 Next Steps

- [x] Check other external API blocks to see if they require skeleton loading state

## ❓ How to test this?

1.  Checkout the branch / view the deployment
1.  Run address block, visually confirm it has loading state as shown below

## 📹 Demo

https://user-images.githubusercontent.com/39553853/227010501-868738ef-7baf-4229-9bd5-872f5b7e1346.mov

PR #2267, opened 2023-03-23
Fix `Address` block error state when retrieving a suggestion

luisbettencourt:

This PR fixes an edge case introduced by #2262

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203801045329780/1204249591562078/f) _(internal)_

## 🔍 What does this change?

Changed the block to only reset the state in the `onClose` callback, instead of when the address card animates out. Because of the loading state introduced in #2262, the address card will always animate in when a suggestion is selected. If the block fails to retieve that suggestion's data it will display an error state and animate out. Because we were resetting the block after the card animated out, the error was being cleared as well.

## 📹 Demo

Before:
before

After:
after


yusufkinatas (approved):
Great catch!

PR #2272, opened 2023-03-23
Investigate GH runner out-of-device-memory error

TimDiekmann:

In #1810 the hash-graph tests are failing because we're hitting the amount of available disk space. Initially, this branch was based-off `main`, but the problem was not present there, so I based it off #1810 and investigate the error from here.

---

After some investigations it turned out that build folders were copied over as docker-context as the context path was changed from `/apps/hash-graph` to `/`. `.dockerignore` is a symlink to `.gitignore`. This PR adjust the `.gitignore` file to include also subdirectories. As a drive-by, this also adds more files, which should not be copied over.


.gitignore
```diff
@@ -64,27 +64,27 @@ blocks/**/.env
...
 # fleet IDEs config
-.fleet/
+**/.fleet/
 
 # Python virtual environment
-venv
+**/venv
 
 # Python compiled files and package metadata
 __pycache__
```
TimDiekmann, line 90:
```suggestion
**/__pycache__
```

Alfred-Mountfield (approved):
It feels pretty strange to have `**/` in front of all of those, especially ones we know _are_ only really at the top-level.. but if this works as a fix I think this is good for now. We can always revisit it if there's a reason to

TimDiekmann:
I think this is only required because we use the same file for both, `.gitignore` and `.dockerignore`. As this syntax work for both, I'm fine with it as it's better to have to keep two files in sync.

PR #2286, opened 2023-03-24
Use lifetimes for strings

TimDiekmann:

There is no real reason to propagate the lifetimes of the strings, so this exposes the lifetime `'s` for all string-fields

### DIFF
--- a/libs/sarif/src/schema/properties.rs
+++ b/libs/sarif/src/schema/properties.rs
@@ -12,19 +12,19 @@ use serde::{Deserialize, Serialize};
     derive(Serialize, Deserialize),
     serde(rename_all = "camelCase")
 )]
-pub struct PropertyBag {
+pub struct PropertyBag<'s> {
     /// A set of distinct strings that provide additional information.
     #[cfg_attr(feature = "serde", serde(skip_serializing_if = "BTreeSet::is_empty"))]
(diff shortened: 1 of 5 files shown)


libs/sarif/src/schema/properties.rs
```diff
@@ -12,19 +12,22 @@ use serde::{Deserialize, Serialize};
...
+        serde(skip_serializing_if = "BTreeSet::is_empty", borrow)
+    )]
+    pub tags: BTreeSet<Cow<'s, str>>,
 
     /// A dictionary, each of whose keys specifies a distinct additional information element and
     /// each of whose values provides the information.
     #[cfg(feature = "serde")]
-    #[serde(flatten)]
-    pub additional: BTreeMap<Cow<'static, str>, serde_json::Value>,
+    #[serde(flatten, borrow)]
```
indietyp, line 26: Any specific reason why additional properties are gated under the `serde`?

indietyp (changes requested):

libs/sarif/src/schema/properties.rs
TimDiekmann: It requires serde-json’s Value. I’m about to change this in a future PR

indietyp (approved):

indietyp (approved):

PR #2292, opened 2023-03-27
Use `just` instead of `cargo make`

TimDiekmann:

We encountered several issues or limitations with `cargo-make` in the past. In addition, we have serious problems to install a pre-built binary for `cargo-make` in CI. We chose to switch to `just` instead, which likely does not have these issues.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1203977361907417/f) _(internal)_

## 🔍 What does this change?

- Define a top-level `.justfile` (similar to `.github/scripts/rust/Makefile.toml`) to define jobs, which are used in the repository. This file also defines the jobs required for CI.
- Define `.justfile` on almost all crates:
    - `apps/hash-graph`: I chose to merge the different test recipes into one for two reasons:
        1. It was unexpected, that `just test` does not result in the same run as the CI. To quickly test the package, `just test` should be sufficient. This also implies, that `just test` will now also take care of the deployment.
        2. It's easier to "just" run ` test` instead of figuring out all different jobs, especially as we expect to increase the amount of different tests in the near 
    - All crates except `deer`: Either add a simple command to skip `miri` or re-use the logic from the `Makefile.toml`
    - `deer` does not need a `.justfile` as it can re-use the top-level one. This would also apply to `antsi` or `sarif`, but as they are `-F unsafe-code` we can avoid running `miri` there
- Remove the `docker` workflow as it was only used in `hash-graph`. Previously, I introduced this workflow to avoid compiling `hash-graph` in Docker and natively in the same workflow, but this is not longer possible anyway since we introduced he type-fetcher, which is now part of the deployment. Merging the REST tests into the `just test` recipe has the almost same runtime characteristics, as the Dockerfile was built anyway and the REST tests are very fast to run
- Removes all `Makefile.toml`s
- Fix several issues around the `component`s in `rust-toolchain.toml`
- Move the caching action to run before installing the tools to avoid overwriting tools

The documentation were adjusted to use `just` instead of `cargo make`

## ⚠️ Known Issues

`cargo-make` attempted to download missing components on demand. This was easier to set up but bad for two reasons:
- It has to check on each invocation, that the component exists
- The components should be installed via `rust-toolchain.toml`

`just` does not have this behavior. I ran into limitations when installing the toolchain within the composite action as with the override the components are not taken into account anymore. We use the override to test different toolchains for a crate (e.g. `error-stack` or `antsi`). I was not able to solve this, yet, but will take a look into this as a follow-up.

Update: Fixed in #2297 

---

The `cargo-script` command does not work as expected. With `cargo-make`, this always enforced a rebuilt, which slowed down the job by a fair amount of time. Luckily, we don't need this anymore but `cargo-script` does not properly see changes in `lint.rs`. I will take a look into this in a follow-up.

Update: Fixed in #2296


.github/scripts/rust/setup.py
```diff
@@ -37,14 +37,12 @@
...
-# Build a docker container for these crates
-DOCKER_PATTERNS = ["apps/hash-graph"]
-
 # Build a coverage report for these crates
 COVERAGE_EXCLUDE_PATTERNS = ["apps/engine**"]
 
 # We only run a subset of configurations for PRs, the rest will only be tested prior merging
-IS_PULL_REQUEST_EVENT = "GITHUB_EVENT_NAME" in os.environ and os.environ["GITHUB_EVENT_NAME"] == "pull_request"
+# IS_PULL_REQUEST_EVENT = "GITHUB_EVENT_NAME" in os.environ and os.environ["GITHUB_EVENT_NAME"] == "pull_request"
+IS_PULL_REQUEST_EVENT = False
```
TimDiekmann, line 45: Temporarily to test all possible jobs in the scope of a PR

.github/actions/install-rust-toolchain/action.yml
```diff
@@ -5,12 +5,31 @@ inputs:
...
+      env:
+        RUSTUP_TOOLCHAIN: ${{ inputs.toolchain }}
+      working-directory: ${{ inputs.working-directory }}
       run: |
-        rustup toolchain install ${{ inputs.toolchain }} --profile minimal --no-self-update
-        rustup default ${{ inputs.toolchain }}
+        rustup set profile minimal
+        rustup show
+
+        # Split the list of components by a comman and call `rustup component add` for each component
```
indietyp, line 28:
```suggestion
        # Split the list of components by a comma and call `rustup component add` for each component
```

.github/workflows/rust.yml
```diff
@@ -79,20 +78,19 @@ jobs:
...
-
-      - name: Install tools
-        uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-hack@0.5.26,cargo-script@0.2.8,clippy-sarif@0.3.7,sarif-fmt@0.3.7
-
-      - name: Install cargo-make
-        run: cargo install --version 0.36.3 cargo-make
+          working-directory: ${{ matrix.directory }}
+          components: rustfmt,clippy
```
indietyp, line 82: Should we maybe extract those from the toml via the python script instead? This way, we only have a single source of truth.

.justfile
```diff
@@ -0,0 +1,171 @@
...
+@install-cargo-script:
+  just install-cargo-tool 'cargo script' cargo-script 0.2.8
+
+[private]
+@install-llvm-cov:
+  just install-cargo-tool 'cargo llvm-cov' cargo-llvm-cov 0.5.9
+
+[private]
+@install-cargo-insta:
+  just install-cargo-tool 'cargo insta' cargo-insta 1.18.2
```
indietyp, line 112: Any specific reason why `just` is invoked instead of just referring to the recipe directly? (applies to most recipes in this file)

libs/@local/status/crate/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "nightly-2023-03-21"
-components = ['clippy', 'rustfmt']
+components = ['rustfmt', 'clippy']
```
indietyp, line 3: We're going for consistent ordering, I assume. What is the ordering based on? Personal preference, alphabetical ordering, or topical ordering?

libs/deer/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "nightly-2023-03-20"
-components = ['rust-src', 'clippy', 'rustfmt', 'miri']
+components = ['rustfmt', 'clippy', 'miri']
```
indietyp, line 3: note (to me): will require rust-src in the near future for integration tests of the derive macros.

.github/scripts/rust/setup.py
indietyp, line 1: Is deleting the docker-related code a drive-by?

indietyp: addressed in PR description

TimDiekmann: Yes, it was not used anywhere and the initial intention does not apply anymore:
> Remove the docker workflow as it was only used in hash-graph. Previously, I introduced this workflow to avoid compiling hash-graph in Docker and natively in the same workflow, but this is not longer possible anyway since we introduced he type-fetcher, which is now part of the deployment. Merging the REST tests into the just test recipe has the almost same runtime characteristics, as the Dockerfile was built anyway and the REST tests are very fast to run

.github/workflows/rust.yml
indietyp: partially addressed in #2297

libs/@local/status/crate/rust-toolchain.toml
TimDiekmann: I was looking at all `rust-toolchain.toml` files anyway. It does not deserve additional documentation as otherwise, it would fall out of sync. This really does not matter at all.

.justfile
TimDiekmann: This way it's possible to overwrite a recipe. E.g. the `lint` recipe calls a few other recipes. If these are overwritten by other justfiles, those are called instead, so the recipes in these files remain usable.

indietyp (approved):

.justfile
```diff
@@ -0,0 +1,171 @@
+set dotenv-load := true
```
thehabbos007, line 1: Should this file have a shebang?
```suggestion
#!/usr/bin/env just --justfile
set dotenv-load := true
```

apps/engine/.justfile
```diff
@@ -0,0 +1,28 @@
...
+  @just in-pr cargo clippy --profile {{profile}} --workspace --all-features --all-targets --no-deps {{arguments}}
+  @just not-in-pr cargo hack --workspace --optional-deps --feature-powerset clippy --profile {{profile}} --all-targets --no-deps {{arguments}}
+
+[private]
+test *arguments:
+  cargo build -p memory --profile {{profile}}
+  bash lib/execution/src/runner/python/setup.sh python3.10
+  @just --justfile {{repo}}/.justfile test {{arguments}}
+
+# Copied from `/.justfile` to ignore example scaping, which does not work properly on the Rust version we currently use.
```
thehabbos007, line 25:
```suggestion
# Copied from `/.justfile` to ignore example scraping, which does not work properly on the Rust version we currently use.
```

apps/hash-graph/Makefile.toml
thehabbos007, line 1: (Comment for `apps/hash-graph/package.json`)
I found `"build": "cargo make build",` on line 15 of the monorepo file, I suppose we want to change this?

thehabbos007 (commented):
Looks good to me. I have a couple comments :)

.github/actions/install-rust-toolchain/action.yml
```diff
@@ -5,12 +5,31 @@ inputs:
...
   steps:
     - name: "Install Rust"
+      env:
+        RUSTUP_TOOLCHAIN: ${{ inputs.toolchain }}
+      working-directory: ${{ inputs.working-directory }}
       run: |
-        rustup toolchain install ${{ inputs.toolchain }} --profile minimal --no-self-update
-        rustup default ${{ inputs.toolchain }}
+        rustup set profile minimal
+        rustup show
```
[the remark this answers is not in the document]
TimDiekmann: Yes, `rustup show` automatically installs the toolchain. Also, it's nice to have the information printed.

.justfile
TimDiekmann: Yes, thank you!
```suggestion
#!/usr/bin/env just --justfile

set dotenv-load := true
```

apps/hash-graph/Makefile.toml
TimDiekmann: I missed that, thanks!

However, I don't really know what we want to use here as `build` does not really make sense as the binary is not used anywhere. I'd say we `just lint` for now until we use Turbo to build, what do you think?

TimDiekmann: @Alfred-Mountfield do you have thoughts on this?

Alfred-Mountfield: I think for now we should build it, I might revisit this today/tomorrow when I look at our dependency graph as we're going to want a separation of 

1. "get some code ready so that it can be imported by other packages"
1. "get some code ready so that it can be ran"

These are possibly often the same, but we're most interested in `1.` for the dependency graph (we need to build dependents if we're importing them, etc.), but `2.` is also going to be interesting later on if we're looking at running things locally (we _could_ move away from docker compose for example, although I'm not suggesting that).

.github/actions/install-rust-toolchain/action.yml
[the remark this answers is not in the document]
Alfred-Mountfield: Are you able to put comments in this file? Might be nice to say `# automatically install the toolchain and output the one being used` or something along those lines.

I frequently find myself getting confused that `show` is actually how to install it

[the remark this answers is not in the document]
TimDiekmann: I added a comment in #2297: https://github.com/hashintel/hash/pull/2297/commits/750a2fb13b0334290da86db7acf6cf88c8b587cb

.env
```diff
@@ -51,7 +51,7 @@ HASH_TASK_EXECUTOR_PORT=5010
 LOG_LEVEL=info
 
 SYSTEM_USER_SHORTNAME=system-user
-SYSTEM_USER_PREFERRED_NAME=System User
+SYSTEM_USER_PREFERRED_NAME="System User"
```
Alfred-Mountfield, line 54: This seems problematic, why did we have to add quotes?

TimDiekmann: Without quotes, this is not a valid `.env` file. Many parsers may allow it, but the Rust crate for parsing `.env` files does not.

Alfred-Mountfield: I see, what sort of error message appears without it? Will CI pick it up?

Just trying to think about the cognitive overhead of how users define these, also this PR might break quite a few of people's current `.env.local` setups so it'd be good if the error message told them, otherwise we need to announce it 😅

TimDiekmann: CI will fail as `just` won't be able to parse it.
```
❯ just
error: Failed to load environment file: Error parsing line: 'System User', error at line index: 7
```

Alfred-Mountfield: Hm, that's probably fine, I do wish it would say something about spaces.

Thanks!

thehabbos007 (dismissed):

.github/workflows/rust.yml
```diff
@@ -79,20 +78,19 @@ jobs:
...
 
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
         with:
           workspaces: ${{ matrix.directory }}
 
+      - name: Install tools
```
Alfred-Mountfield, line 89: Can the tools be in the cache? And if so does this install action choose not to re-install them?

.github/workflows/rust.yml
```diff
@@ -102,13 +100,13 @@ jobs:
 
       - name: Check formatting
         working-directory: ${{ matrix.directory }}
-        run: cargo +${{ matrix.toolchain }} make format -- --check
+        run: just format
```
Alfred-Mountfield, line 103: I would have expected to still see a check flag passed through to this.

Is there a reason we've moved away from that? Formatting and checking formatting are different things in my mind

.github/workflows/rust.yml
```diff
@@ -173,21 +172,20 @@ jobs:
...
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
         with:
           workspaces: ${{ matrix.directory }}
           key: ${{ matrix.profile }}
 
+      - name: Install tools
+        uses: taiki-e/install-action@v2
+        with:
+          tool: just@1.13.0,cargo-hack@0.5.26,cargo-nextest@0.9.37
```
Alfred-Mountfield, line 187: Is there an easy way to centralize where the versions of tools are specified?

.github/workflows/rust.yml
```diff
@@ -203,12 +201,14 @@ jobs:
...
         working-directory: ${{ matrix.directory }}
-        run: cargo +${{ matrix.toolchain }} make --profile ${{ matrix.profile }} test --no-fail-fast
+        run: just test --no-fail-fast
 
       - name: Run miri
         if: ${{ startsWith(matrix.toolchain, 'nightly') }}
         working-directory: ${{ matrix.directory }}
-        run: cargo +${{ matrix.toolchain }} make --profile ${{ matrix.profile }} miri --no-fail-fast
+        run: |
+          rustup component add miri
```
Alfred-Mountfield, line 210: Not suggesting it's a problem, just seems strange that we have a manual `component add` command here where in other places we use the `components:` thing (although I appreciate that's removed in a follow-up PR)

.github/workflows/rust.yml
TimDiekmann: Not as I know of, no, unless we make everything a single action, which I would avoid. Even then we still have to specify the version in `.justfile`.

.github/workflows/rust.yml
TimDiekmann: `miri` is special here as it's never available in non-nightly toolchains.

.github/workflows/rust.yml
Alfred-Mountfield: How much prettier? (Just asking for your judgement, you don't have to supply a demo)

My instinct is that this isn't worth the change, there's a command designed to specifically check if things are formatted, and that's what we're doing here. Generally speaking, rewriting checked in source files in CI is not a good approach, it can cause problems with race conditions, and it's harder to debug. 

This isn't just a theoretical mind you, it's a genuine problem we're already seeing with our `codegen` steps in our blocks, where we need to format after codegen and because it's rewriting the source files in CI it's causing problems and we've had to `gitignore` them. 

I'm not saying that we _will_ run into issues here from it, I just don't see a reason to risk it. I also assume that actually formatting will be slower than checking? Although probably not significantly

.github/workflows/rust.yml
Alfred-Mountfield: Understood, thanks for clarifying :)

.justfile
```diff
@@ -0,0 +1,178 @@
...
+@list-repo-recipes:
+  echo "\nRepository recipes:"
+  just --list --unsorted --list-heading ''
+
+
+######################################################################
+## Helper to run a global command
+######################################################################
+
+# Runs yarn in the repository root
```
Alfred-Mountfield, line 44: Just so I understand, what about this forces it to be in the repository root?

.justfile
```diff
@@ -0,0 +1,178 @@
...
+  fi
+
+
+######################################################################
+## Install scripts
+######################################################################
+
+[private]
+@install-cargo-tool tool install version:
+  `{{tool}} --version | grep -q "{{version}}" || cargo install "{{install}}" --version "{{version}}" --locked --force`
```
Alfred-Mountfield, line 94: Does `cargo install` not already do this? If you rerun `cargo install` for the same thing does it override it? (Or is that why you have `--force`?)

.justfile
Alfred-Mountfield: Which way round does the overriding work sorry? I don't quite understand what's being described.

Are you suggesting a local justfile could override `install-cargo-tool` for example, and then if _within that local context_ you call `just install-cargo-script` it would use your overridden one? Or it would use the one defined in this file?

.justfile
TimDiekmann: This covers the case when you have installed the tool manually without cargo. But thinking about this I think maybe we should just rely on `cargo install`? It's at least easier to read and results in the same behavior in 99% of all cases.

.justfile
TimDiekmann: It does not have `[no-cd]`, so it will be run from the `.justfile` directory.

.justfile
TimDiekmann:
```
❯ cargo nextest --version
cargo-nextest 0.9.37
❯ cargo install cargo-nextest --version 0.9.37
     Ignored package `cargo-nextest v0.9.37` is already installed, use --force to override
```

TimDiekmann: I'd say I try it out and see if it will re-use the install-action tools. In theory it should.

.justfile
Alfred-Mountfield: I guess I'm unclear on how this differs from other commands within this `.justfile` and why this one is singled out as being the repository root

TimDiekmann: Other commands, such as `lint` or `format` has the `[no-cd]` attribute, which implies when invoking these jobs from a subdirectory it will not change the directory to the root. This implies, that when calling `just format` inside of `apps/hash-graph` it will not run `format` in the repository root but in `apps/hash-graph`. However, we want `yarn` to be executed in the workspace root, so it's possible to call `just yarn` from anywhere.

.justfile
```diff
@@ -0,0 +1,174 @@
...
+# Updates the compiler warnings in the `.cargo/config.toml` file.
+[private]
+[no-cd]
+lint-toml mode:
+  @cargo script "{{repo}}/.github/scripts/rust/lint.rs" {{mode}}
+
+# Runs all linting commands and fails if the CI would fail
+[no-cd]
+lint:
+  @just format -- --check
```
Alfred-Mountfield, line 126: This looks different to https://github.com/hashintel/hash/blob/b81c50a5c35fd9fcf73b5ef0ecc9994bd580f6bf/.github/workflows/rust.yml#L103

Does the `@just` syntax require `--`?

.justfile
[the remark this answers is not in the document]
TimDiekmann: If `just` is called (without specifying `--justfile <PATH>`) it will search for the next `.justfile` or `justfile` in this directory or parent directories. if a justfile was found it will terminate the search. However, if `set fallback` is set in the justfile and a command was not found (e.g. `lint`) it will search for more justfiles in the parent directories and apply the same logic. With this it's possible to basically overwrite some recipes like `test` in `apps/hash-graph/.justfile`. These are set to `[private]` to avoid duplicates when calling `just` (which prints the available recipes).

This is especially interesting when calling e.g. `just lint` as these will not `cd` into the workspace root but run every listed command in that folder, so `just test` will invoke the `test` recipe in `hash-graph` but the `format --check` recipe in the root.

.justfile
TimDiekmann: It does not work in CI:
```
error: binary `cargo-hack` already exists in destination
[23](https://github.com/hashintel/hash/actions/runs/4542938790/jobs/8007007610?pr=2292#step:9:24)
Add --force to overwrite
```

Also, it will always output to `stderr` if a binary was not installed as it already exists. I reverted the test.

apps/engine/.justfile
```diff
@@ -0,0 +1,28 @@
...
+set fallback
+
+repo := `git rev-parse --show-toplevel`
+profile := env_var_or_default('PROFILE', "dev")
+
+[private]
+@default:
+  just usage
+
+# Copied from `/.justfile` to ignore the `lint-toml` job.
```
Alfred-Mountfield, line 12: Any alternatives you can think of?

apps/hash-graph/.justfile
```diff
@@ -0,0 +1,60 @@
...
+[private]
+coverage *arguments:
+  @just deployment-up
+  cargo llvm-cov --workspace --all-features --all-targets {{arguments}}
+  @just deployment-down
+
+[private]
+bench *arguments:
+  @just deployment-up
+  @just --justfile {{repo}}/.justfile bench {{arguments}}
```
Alfred-Mountfield, line 55: Any possible alternatives to this?

Not really a fan of the implicit complexity we're adding in about reusing recipes, but I'm not sure we can easily get around it.

It seems that sometimes things reuse by just's parent resolution logic, and sometimes by explicitly calling one..

apps/hash-graph/README.md
```diff
@@ -38,25 +38,25 @@ If you're interested in just increasing the logs for the Graph, we recommend spe
...
 ```shell
-cargo make build
+just build
 ```
 
 In order to create an optimized build, run:
 
 ```shell
-cargo make --profile production build
+just --profile production build
```
Alfred-Mountfield, line 47: Does this work? It looked like the `.justfile` requires it to come from the env?

.justfile
Alfred-Mountfield: Thanks for the information :)

apps/engine/.justfile
Alfred-Mountfield: Agreed, but I more meant isn't there a better way to override this without copying it? Can we not locally redefine `lint-toml` to be a no-op like we do for `miri` elsewhere for example?

apps/hash-graph/.justfile
TimDiekmann: Actually, I like this solution. It's basically extending the default behavior. In comparison to `cargo make`, where we had two tasks for each step to make overwriting slightly easier, it was still hard to get the behavior as it only was partially overwriting the job.

One solution would be to have a general job, which would be run prior to and after every job, which can be overwritten, something like `pre-bench` and `post-bench`. Probably a solution we can discover but most likely not worth doing it right now, especially as we have to discover how we will invoke `turbo` in CI.

Alfred-Mountfield: I'm not arguing that the `cargo-make` approach was a good one. Just that it's proving quite hard to actually discover what is ran if I run a `just` command. I have to trace it through all of the parent `.justfiles` and then deal with logic around `[no-cd]` on _some_ of them, and now I **also** need to look at how some of them skips that whole process to directly call a specific `justfile`, which is going to create another set of edge-cases we need to handle when adjusting top-level definitions.

This is creating a very opaque and potentially convoluted dependency graph.

libs/error-stack/.justfile
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env just --justfile
+
+set fallback
+
+[private]
+@default:
+  just usage
+
+# TODO: add `--ignore-unknown-features` to `cargo hack` and pass `--workspace`
```
Alfred-Mountfield, line 9: What's the importance of this TODO? Is it tracked or handled in a future PR?

libs/error-stack/.justfile
```diff
@@ -0,0 +1,52 @@
...
+
+[private]
+clippy *arguments:
+  @just install-cargo-hack
+
+  @just in-pr cargo clippy --profile {{profile}} --workspace --all-features --all-targets --no-deps {{arguments}}
+  @just not-in-pr cargo hack --optional-deps --feature-powerset {{cargo-hack-groups}} --ignore-unknown-features clippy --profile {{profile}} --all-targets --no-deps {{arguments}}
+
+[private]
+test *arguments:
```
Alfred-Mountfield, line 21: Any reason why you don't have the other syntax for these?
```suggestion
@test *arguments:
```

apps/hash-graph/README.md
Alfred-Mountfield: Cheers, can you have a quick scan to see if there are any other similar ones. Maybe in `dockerfile`s or something

apps/engine/.justfile
TimDiekmann: Good idea: [Ignore lint-toml job in engine](https://github.com/hashintel/hash/pull/2292/commits/9ed789eeadb147c028298f2d45f099c304dc2ace)

libs/error-stack/.justfile
TimDiekmann: It's an annoyance and unimportant until we implement something in `error-stack` macros. It was just ported over.

libs/error-stack/.justfile
TimDiekmann: `@` in front of the job hides all `stderr` output emitted by `just`, so we don't need to apply it to every line. However, I'm fine avoiding this syntax.

TimDiekmann: I preferred `@` per line as this was more explicit but don't really have strong feelings.

Alfred-Mountfield: I don't feel strongly either, but we're inconsistent at the moment, that's what I'm highlighting.

Even for jobs where we're applying `@` to every line, we aren't using it on the top level command. But then on some other ones we do

apps/hash-graph/.justfile
TimDiekmann: The job is called the same, so just invoking `just bench` would be recursion. I don't know how we could re-use that behavior. Things like
```just
@`(cd .. && just bench {{arguments}})`
```
does not work either as the working directory would be different. One option is to implement a recipe, which does this recursion logic until it finds a parent `.justfile`, but this I consider out of scope of this PR.

Alfred-Mountfield: Yep, I can't think of an easy way either. I think the `pre-X` and `post-X` pattern might be a good idea to explore later as well, but it's probably got a horrible amount of boilerplate

TimDiekmann:
Thanks, @Alfred-Mountfield! 🎉 

I made the `@` more consistent and never use `@` on a recipe name.

I agree that we should try moving as much as possible to the `turbo` setup. forwarding `just` to `turbo` in that case? I.e. `just build` invokes `turbo` with the corresponding parameters? Nothing to discuss in this PR, just a few ideas.

Alfred-Mountfield (approved):

PR #2296, opened 2023-03-27
Use `rust-script` instead of `cargo-script`

TimDiekmann:

As mentioned in #2292, `cargo-script` is unmaintained, and `rust-script` is a maintained fork, which solves the issue we currently have.

## 🔍 What does this change?

- Replace `cargo-script` with `rust-script@0.23.0`
- Drive-by: allow running `just` in all directories within a workspace

### DIFF
--- a/.github/scripts/rust/lint.rs
+++ b/.github/scripts/rust/lint.rs
@@ -13,7 +13,7 @@ use std::{
     collections::{BTreeSet, HashMap},
     env, fs,
     iter::once,
-    path::Path,
+    path::{Path, PathBuf},
 };
 
 use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


apps/engine/.justfile
```diff
@@ -13,7 +13,6 @@ profile := env_var_or_default('PROFILE', "dev")
 [private]
 clippy *arguments:
   @just install-cargo-hack
-  @just install-cargo-script
```
indietyp, line 16: Specific reason why this was removed i stead of being replaced?

indietyp (changes requested):

indietyp (dismissed):

indietyp (approved):

PR #2297, opened 2023-03-27
Use `rust-toolchain.toml` file to install toolchains

TimDiekmann:

As noticed in #2292, the `rust-toolchain.toml` should be the location, where the components are specified. In #2292 I ran into limitations as `RUSTUP_TOOLCHAIN` would override the `rust-toolchain.toml`. This PR works around this by not setting `RUSTUP_TOOLCHAIN` if the toolchain does match with the one inside of `rust-toolchain.toml`

### DIFF
--- a/.github/actions/install-rust-toolchain/action.yml
+++ b/.github/actions/install-rust-toolchain/action.yml
@@ -23,7 +23,7 @@ runs:
         if grep -q "${{ inputs.toolchain }}" rust-toolchain.toml; then
           rustup show
         else
-          rustup toolchain install "${{ inputs.toolchain }}"
+          rustup default "${{ inputs.toolchain }}"
           rustup component add rustfmt
           rustup component add clippy
           rustup show
--- a/.github/actions/install-rust-toolchain/action.yml
+++ b/.github/actions/install-rust-toolchain/action.yml
@@ -15,8 +15,8 @@ runs:
   steps:
     - name: "Install Rust"
       working-directory: ${{ inputs.working-directory }}
+      shell: bash
       run: |
-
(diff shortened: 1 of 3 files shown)


indietyp (dismissed):

.github/actions/install-rust-toolchain/action.yml
```diff
@@ -9,27 +9,22 @@ inputs:
...
 
-        # Split the list of components by a comma and call `rustup component add` for each component
-        IFS=',' read -ra components <<< "${{ inputs.components }}"
-        for component in "${components[@]}"; do
-          rustup component add "$component"
-        done
-
-        echo "RUSTUP_TOOLCHAIN=$RUSTUP_TOOLCHAIN" >> $GITHUB_ENV
-      shell: bash
+        rustup show
```
TimDiekmann, line 29:
```suggestion
        # Installs the specified toolchains and prints it
        rustup show
```

.github/workflows/rust.yml
```diff
@@ -79,7 +79,6 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           working-directory: ${{ matrix.directory }}
-          components: rustfmt,clippy
```
Alfred-Mountfield, line 82: So why do we not require these any more?

TimDiekmann: We rely on `rust-toolchain.toml` for the required components, so the default toolchain will install them automatically. If we require another toolchain (e.g. `stable` for `error-stack`) these components are implied by `rustup set profile default`.

TimDiekmann: This enforces that the `rust-toolchain.toml` has the correct components specified and even fails to update to a toolchain without those components.

indietyp (approved):

PR #2298, opened 2023-03-28
Fix `variants` for the Heading Block

Alfred-Mountfield:

The `variants` of the Heading Block weren't updated to reflect the type system. This means that the demo on the hub is broken:

https://user-images.githubusercontent.com/25749103/228216172-33c3fbe7-4043-4c7f-b44e-a88d2ae2f66e.mov

This PR fixes this, when used in conjunction with https://github.com/blockprotocol/blockprotocol/pull/1223

## 📹 Demo

(After, with https://github.com/blockprotocol/blockprotocol/pull/1223)

https://user-images.githubusercontent.com/25749103/228216388-b4c54913-b697-4cda-a49d-accb00e3915a.mov

### DIFF
--- a/blocks/heading/package.json
+++ b/blocks/heading/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blocks/heading",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "private": true,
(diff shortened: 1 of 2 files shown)


nathggns (approved):
Good catch. Thanks @Alfred-Mountfield

PR #2299, opened 2023-03-28
Add `AI Chat` block (MVP)

benwerner01:

This PR adds an MVP version of the AI Chat block.

Remaining TODOs:
- [x] add block image to `package.json`
- [x] add block icon to `package.json`
- [x] give text-field a maximum width
- [x] publish the types in production and update the hardcoded values in this PR (blocked by first approval)
- [ ] publish the block 🎉

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204202623523142/f) _(internal)_

Not that I'm aware of.

## 🐾 Next steps

- improve error handling (currently there is none)
- consider unifying the `completeChatRequests` and `completeChatResponses` arrays into a single `messages` array to simplify the logic in this file. Potentially `RequestId` and `ResponseId` could be unified in a single `MessageId` as part of this.
- improve mobile responsiveness of the block
- allow users to select a message to display in readonly mode
- allow users to regenerate a response message
- allow users to rewrite a request message

TODO: create tasks for these

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch and install dependencies
2. Add an `.env` file to the `blocks/ai-chat` directory and add valid values for the `BP_HOST` and `BP_API_KEY` environment variables
3. Run the chat block using `yarn workspace @blocks/ai-chat dev`, and test its functionality

You can easily test whether the block is able to load an existing chat thread by uncommenting line 141 in `blocks/ai-chat/src/dev.tsx`. 

## 📹 Demo

https://user-images.githubusercontent.com/42802102/228319064-d2e831c9-b2e6-4d94-bac9-82c0728b4a6b.mov


blocks/ai-chat/src/types/generated.ts
Alfred-Mountfield, line 1: I assume this is called `generated.ts` to avoid the `.gitignore` for the sake of the review?

Just keen for us to stick to patterns where we can (and the templates have `types/generated/block-entity.ts` I believe)

Alfred-Mountfield, line 1: Could you format this file and push a commit? Might make it easier to review

benwerner01: Yes, this PR is not ready for review have only called `generated.ts` for now to easily distinguish from the other type files I have.

blocks/ai-chat/src/graph.ts
Alfred-Mountfield, line 1: Do we perhaps want to adopt the pattern that's proposed (in a draft PR atm, not merged) in this? https://github.com/blockprotocol/blockprotocol/tree/e0d02780463dbb22e1ee91b01a1967dfbce760ab/blocks/aio-demo/src/types

We already have a `types` directory here so `graph.ts` could be under that.

blocks/ai-chat/src/types/generated.ts
benwerner01: Do we have an existing practice for what to call these files when they contain multiple types? (guessing `types/generated/block-entity.ts` is not the most appropriate name for such a file)

Or would you prefer the types to be separated into multiple files?

blocks/ai-chat/src/graph.ts
```diff
@@ -0,0 +1,202 @@
...
+export const aiChatMessageLinkTypeId: keyof AIChatBlockOutgoingLinksByLinkEntityTypeId =
+  "http://localhost:3000/@alice/types/entity-type/ai-chat-message/v/1";
+
+export const rootAIChatRequestLinkTypeId: keyof AIChatBlockOutgoingLinksByLinkEntityTypeId =
+  "http://localhost:3000/@alice/types/entity-type/root-ai-chat-request/v/1";
+
+export const aiChatRequestResponseLinkTypeId: keyof AIChatRequestOutgoingLinksByLinkEntityTypeId =
+  "http://localhost:3000/@alice/types/entity-type/response/v/1";
+
+export const aiChatResponseRequestLinkTypeId: keyof AIChatResponseOutgoingLInksByLinkEntityTypeId =
```
Alfred-Mountfield, line 55:
```suggestion
export const aiChatResponseRequestLinkTypeId: keyof AIChatResponseOutgoingLinksByLinkEntityTypeId =
```

Typo it seems

blocks/ai-chat/src/types/generated.ts
Alfred-Mountfield: `block-entity.ts` is just the pattern for the templates because it's the most generic and requires the fewest changes for end-users when writing a block. I think _personally_ I'd prefer if we called them by the things we're generating in them, so `ai-chat-block.ts` or something.

If you're generating multiple top-level types (i.e. if you've put multiple entries under the file in the `codegen` part of the `package.json`) then yes maybe we should pick a different name for ones with multiple types.

> would you prefer the types to be separated into multiple files?

This seems cleanest, it will hoist shared types to `shared.ts` and re-export them (mostly, see [this Asana task](https://app.asana.com/0/1202805690238892/1204258875913354/f) for a known bug).

blocks/ai-chat/src/graph.ts
```diff
@@ -0,0 +1,202 @@
...
+export type AIChatRequestOutgoingLinksByLinkEntityTypeId =
+  AIChatRequestV4OutgoingLinksByLinkEntityTypeId;
+
+export type AIChatResponse = AIChatResponseV4;
+export type AIChatResponseOutgoingLInksByLinkEntityTypeId =
+  AIChatResponseV4OutgoingLinksByLinkEntityTypeId;
+
+/** Entity Type IDs */
+
+export const aiChatBlockEntityTypeId =
```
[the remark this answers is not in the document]
benwerner01: Nice, started doing this as part of https://github.com/hashintel/hash/pull/2299/commits/0b64e42373ca3ddb7e17bfdfc0d69cf4e059d836

blocks/ai-chat/package.json
```diff
@@ -64,19 +65,19 @@
...
           {
             "blockEntityType": true
           },
           {
-            "sourceTypeId": "http://localhost:3000/@alice/types/entity-type/ai-chat-request/v/4"
+            "sourceTypeId": "https://blockprotocol-imqnt3bj2.stage.hash.ai/@alfie/types/entity-type/request-message/v/2"
           },
           {
-            "sourceTypeId": "http://localhost:3000/@alice/types/entity-type/ai-chat-response/v/4"
+            "sourceTypeId": "https://blockprotocol-imqnt3bj2.stage.hash.ai/@alfie/types/entity-type/request-message/v/2"
```
Alfred-Mountfield: (these are the same, I think you meant response for one)

Do you need to specify these types? I'm assuming they're brought in by the `block-entity`

Alfred-Mountfield: If you do want to bring them in perhaps it'd be good to separete it into `request-message.ts` and `request-response.ts`

blocks/ai-chat/src/graph.ts
benwerner01: I've moved the file into `types` as part of https://github.com/hashintel/hash/pull/2299/commits/0b64e42373ca3ddb7e17bfdfc0d69cf4e059d836.

Would additionally you prefer it if I split the file into:
- `entity-types.ts` (and include the existing methods for creating `RequestMessage` and `ResponseMessage` entities)
- `property-types.ts`
- `link-entity-types.ts`

Alfred-Mountfield: Defer to your judgement and whatever seems maintainable, we can always refactor

blocks/ai-chat/src/types/generated.ts
benwerner01: Have moved them into distinct files in https://github.com/hashintel/hash/pull/2299/commits/cc7888d92006160894d3ba8e2fbc28b5b8a06889

blocks/ai-chat/package.json
benwerner01: I have fixed the typo and moved them into individual files in https://github.com/hashintel/hash/pull/2299/commits/cc7888d92006160894d3ba8e2fbc28b5b8a06889

blocks/ai-chat/src/complete-chat/chat-model-selector.tsx
```diff
@@ -0,0 +1,79 @@
...
+  [ModelGroupName.ChatGPT]: {
+    name: ModelGroupName.ChatGPT,
+    options: [
+      {
+        id: "gpt-3.5-turbo",
+        icon: <AbstractAiIcon sx={{ fontSize: "inherit" }} />,
+        title: "OpenAI GPT-3.5 Turbo",
+        helperText: "gpt-3.5-turbo",
+        /** @todo: improve description */
+        description: "The best model for many use cases.",
```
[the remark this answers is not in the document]
nonparibus: @benwerner01 `The fastest and cheapest ChatGPT model (default).` ...nevermind that it's the only one included, for now. GPT-4, Claude, and all the others soon!

blocks/ai-chat/src/assets.d.ts
luisbettencourt, line 1: We can remove this file, right?

blocks/ai-chat/src/complete-chat/chat-model-selector.tsx
luisbettencourt, line 33:
```suggestion
        description: "The fastest and cheapest ChatGPT model (default).",
```

libs/@hashintel/design-system/src/dropdown-selector.tsx
luisbettencourt, line 1: This should already be in the design system now.

luisbettencourt (changes requested):
Great job @benwerner01, the block is looking very polished! Have done a quick code review and left some comments. Will do a more in-depth review tomorrow morning

blocks/ai-chat/src/complete-chat/chat-model-selector.tsx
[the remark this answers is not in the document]
Alfred-Mountfield: (Leaving it to @nonparibus, but do we want to allude to something like this, getting ready for GPT-4)
"Recommended: Provides a good trade-off between cost and performance."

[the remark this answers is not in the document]
nonparibus: That would also work. Our descriptors largely mirror those on OpenAI's site and they frame it both ways.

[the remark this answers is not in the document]
Alfred-Mountfield: I think being directly told it's the cheapest might be even nicer than some opaque "good balance"

blocks/ai-chat/src/complete-chat.tsx
```diff
@@ -0,0 +1,486 @@
...
+
+export type RequestId = `req_${number}`;
+
+export const createRequestId = (): RequestId => `req_${requestCounter++}`;
+
+let responseCounter = 0;
+
+export type ResponseId = `res_${number}`;
+
+export const createResponseId = (): ResponseId => `res_${responseCounter++}`;
```
luisbettencourt, line 50: Suggestion: you can use `uuid` instead of the counter,

libs/@hashintel/design-system/src/dropdown-selector.tsx
benwerner01: Where is it located? I'm not able to see it in `main`: https://github.com/hashintel/hash/tree/main/libs/%40hashintel/design-system/src

blocks/ai-chat/src/complete-chat/header.tsx
```diff
@@ -0,0 +1,97 @@
...
+  setChatModel,
+  systemPromptId,
+  setSystemPromptId,
+}) => {
+  const [chatModelSelectorOpen, setChatModelSelectorOpen] = useState(false);
+  const [systemPromptSelectorOpen, setSystemPromptSelectorOpen] =
+    useState(false);
+
+  return (
+    <Box
```
luisbettencourt, line 31: We probably want to make this header fade in/out as we hover in/out of the block, like other blocks. Right @nonparibus?

Also, I noticed the selectors are aligned to the right-side of the block, whereas in other blocks everything is aligned to the start. Do we want to keep consistency here as well?

blocks/ai-chat/src/complete-chat/chat-model-selector.tsx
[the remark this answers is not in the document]
benwerner01: Have updated this to `The fastest and cheapest ChatGPT model (default).` in https://github.com/hashintel/hash/pull/2299/commits/614506c5f47f6db71e8b8f6e462c50e3bbeb6ae0

blocks/ai-chat/src/assets.d.ts
benwerner01: Correct, thanks for flagging. Have removed in https://github.com/hashintel/hash/pull/2299/commits/80d6d7cd417772d9c412a1130b89e96db5a00f46

luisbettencourt (changes requested):

blocks/ai-chat/src/complete-chat.tsx
```diff
@@ -0,0 +1,486 @@
...
+        setCompleteChatResponses((previousResponses) => {
+          const existingResponse = previousResponses.find(
+            ({ id }) => id === response.id,
+          );
+
+          return existingResponse
+            ? previousResponses
+            : [...previousResponses, response];
+        });
+      }, 500);
```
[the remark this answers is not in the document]
benwerner01: It's so that the response message skeleton isn't immediately displayed alongside the request message, so that their entrance animations are triggered sequentially rather than at the same time.

IMO this creates an effect that's closer to what users expect when having a conversation with another being, as it will not immediately start thinking about how to respond.

libs/@hashintel/design-system/src/dropdown-selector.tsx
luisbettencourt: Got a bit confused with what PRs had gone in already and what work was in them. It hasn't been merged yet. The PR to track is #2251. 
Sorry for wasting your time.

benwerner01: No worries, just want to flag that I made a few changes to the dropdown-selector so this will cause conflicts. I had to change how the positioning is determined for instance, as the dropdown was overflowing the page previously.

blocks/ai-chat/src/complete-chat.tsx
```diff
@@ -0,0 +1,486 @@
...
+      : [];
+  }, [rootChatRequestId, completeChatRequests, completeChatResponses]);
+
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const updateAiChatBlockEntity = useCallback(
+    async (params: {
+      updatedProperties: Partial<AIChatBlock["properties"]>;
+    }) => {
+      const { updatedProperties } = params;
```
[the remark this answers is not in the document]
benwerner01: Shouldn't the request fail anyway because the EA is enforcing whether or not they can be called? (cc @CiaranMn)

If this is inconsistent with a previous decision I'm happy to change this, but I feel like the block should be ensuring that an update method can't get called by a user in the UI in the first place, rather than abort in the function when it has gotten called. This could also help us catch when this is happening more easily, as an error should be thrown by the EA.

I also don't think it's the block's responsibility to enforce such a safety measure, it's the responsibility of the EA. But if my understanding is incorrect please let me know.

blocks/ai-chat/src/complete-chat.tsx
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+  if (aiChatBlockEntityChatAiModel !== chatModel) {
+    void updateAiChatBlockEntity({
+      updatedProperties: {
+        [propertyTypeBaseUrls.openAIChatModelName]: chatModel,
+      },
+    });
+  }
+
```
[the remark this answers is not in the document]
benwerner01: > Also, do we really need to set the values for the initial state of the block? We can just use the default values if there is none, right?

If we don't update the entity with the default settings, we wouldn't be capturing this data in the graph unless the user is not using the default settings, which seems sub-optimal to me.

I appreciate the way it is currently implemented `updateEntity` could be needlessly called twice in the same render, so have refactored this slightly to address this in https://github.com/hashintel/hash/pull/2299/commits/8d22a7afacc7aa7ef6d9b6bf2211fbd149ca070f

blocks/ai-chat/src/complete-chat/header.tsx
nonparibus: Yes and yes

blocks/ai-chat/src/complete-chat.tsx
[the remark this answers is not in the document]
benwerner01: I've tried to account for the race condition where the completed response may return under 500ms

blocks/ai-chat/src/complete-chat.tsx
benwerner01: I could, but this is an added dependency so decided to go without it. Any benefit for adding it now that you can think of?

blocks/ai-chat/src/complete-chat/header.tsx
benwerner01: I wasn't aware we were fading these in an out in other components, happy to do this. Regarding the alignment, this was my interpretation from [the designs](https://www.figma.com/file/T10d2P2iNb1LdRC5a9J7vA/Blocks?node-id=1880-6619&t=yx3HPCB4uXm42dz3-0) but am realising now it's a bit ambiguous whether they should be right-aligned or left-aligned.

blocks/ai-chat/src/complete-chat.tsx
luisbettencourt: Just code simplification, really. Happy to keep it as is

blocks/ai-chat/src/complete-chat.tsx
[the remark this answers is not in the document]
luisbettencourt: I see, thanks!

blocks/ai-chat/src/complete-chat.tsx
benwerner01: Fair enough, have addressed in https://github.com/hashintel/hash/pull/2299/commits/43f7e4ef5f39cde0c158b43f43bae256776cb2df

blocks/ai-chat/src/graph.ts
benwerner01: I'm going to stick to the `graph.ts` file for now, but agree its worth rethinking if we add to it down the line. Captured this in a small todo in https://github.com/hashintel/hash/pull/2299/commits/93941a48cb89ef87d4e34803f58df7381563f500

luisbettencourt (changes requested):
Approved functionality-wise. Just some responsiveness issues I've noticed when doing some final testing:

1. The block stretches to fill the container, but should probably only fill the needed width to accomodate the input and example prompts:
image

2. The example prompt buttons don't look great for smaller screen sizes:
image

3. The header overflows the container for smaller screen sizes. We can use the `BlockSettingsButton` in the design system for this, and also wrap the text when necessary (similar to what's been done in the address block).
image

Do we want to change these, @nonparibus? Happy to approve otherwise.

nonparibus:
Can't comment directly on Luís's comment, but yes, please. ^

benwerner01:
> 1. The block stretches to fill the container, but should probably only fill the needed width to accomodate the input and example prompts:

After [the internal conversation in slack](https://hashintel.slack.com/archives/C04SP1PGKC1/p1679706244764579) how to handle this, have added a max-width to the block container in https://github.com/hashintel/hash/pull/2299/commits/c291740f86e01628ab1abf02a016b10fbb79d253

> 2. The example prompt buttons don't look great for smaller screen sizes:

Have improved these in https://github.com/hashintel/hash/pull/2299/commits/dd6b48f8488686c6b4815d9070c358c28b50de38.

> 3. The header overflows the container for smaller screen sizes. We can use the `BlockSettingsButton` in the design system for this, and also wrap the text when necessary (similar to what's been done in the address block).

Have taken the approach from the address block, and added the `BlockSettingsButton` as you suggested in https://github.com/hashintel/hash/pull/2299/commits/1a4145aef2348cc0a737b441a764d28d08893c05.

On mobile sizes the block now looks like this:
image

luisbettencourt (approved):

nonparibus (approved):

PR #2306, opened 2023-03-28
Add several Rust crate groups to renovate.json

TimDiekmann:

Some crates are almost always updated at the same time or should be updated together

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -72,6 +72,45 @@
       "matchPaths": ["apps/**"],
       "groupName": "Rust applications toolchains",
       "commitMessageTopic": "Rust toolchains for `/apps`"
+    },
+    {
+      "matchManagers": ["cargo"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -72,6 +72,45 @@
       "matchPaths": ["apps/**"],
       "groupName": "Rust applications toolchains",
       "commitMessageTopic": "Rust toolchains for `/apps`"
+    },
+    {
```
indietyp, line 76: Don't they also need a `commitMessageTopic`? Otherwise, the PR name might be not as expected.

indietyp (changes requested):

.github/renovate.json
TimDiekmann: I don't think it's required, I like to try it out without specifying it explicitly, otherwise this should be an easy follow-up fix.

indietyp (approved):

PR #2313, opened 2023-03-29
Add `test_graph` module containing required and optional record fields

TimDiekmann:

For creating a snapshot, we cannot re-use the representation for the ontology types as we store more information in the database than exposed, e.g. the `transaction-time`. Also, we want to distinguish between fields specified by blockprotocol.org and custom fields we use internally in HASH.

This PR adds a module, which contains the structs required to deserialize a graph.

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1204000740778938/1204216809501005/f) _(internal)_

## 🔍 What does this change?

- Implement `Deserialize` for `EntityEditionId` and `EntityRecordId`
- Add structs to represent ontology types and entities with an optional field for metadata. Note, that I don't have added `AccountId`s (yet) as I don't expect it to be used in the short-term future. Querying those is complicated because that does not go through structural queries and when an account id is missing when inserting test data, the graph has to deal with that.

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/test_graph.rs
+++ b/apps/hash-graph/lib/graph/src/store/test_graph.rs
@@ -133,27 +133,16 @@ impl CustomGlobalMetadata {
     }
 }
 
-#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct BlockProtocolModuleVersions {
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub graph: Option<semver::Version>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub authorization: Option<semver::Version>,
-}
-
-impl BlockProtocolModuleVersions {
-    #[must_use]
-    const fn is_empty(&self) -> bool {
-        self.graph.is_none() && self.authorization.is_none()
-    }
+    pub graph: semver::Version,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct TestData {
-    #[serde(default, skip_serializing_if = "BlockProtocolModuleVersions::is_empty")]
-    pub block_protocol_spec: BlockProtocolModuleVersions,
+    pub block_protocol_module_versions: BlockProtocolModuleVersions,
     pub data_types: Vec<OntologyTypeRecord<DataType>>,
     pub property_types: Vec<OntologyTypeRecord<PropertyType>>,
     pub entity_types: Vec<OntologyTypeRecord<EntityType>>,
--- a/apps/hash-graph/lib/graph/src/store/test_graph.rs
+++ b/apps/hash-graph/lib/graph/src/store/test_graph.rs
@@ -133,10 +133,27 @@ impl CustomGlobalMetadata {
     }
 }
 
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockProtocolModuleVersions {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub graph: Option<semver::Version>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub authorization: Option<semver::Version>,
+}
+
+impl BlockProtocolModuleVersions {
+    #[must_use]
+    const fn is_empty(&self) -> bool {
+        self.graph.is_none() && self.authorization.is_none()
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct TestData {
-    pub block_protocol_spec: semver::Version,
+    #[serde(default, skip_serializing_if = "BlockProtocolModuleVersions::is_empty")]
+    pub block_protocol_spec: BlockProtocolModuleVersions,
     pub data_types: Vec<OntologyTypeRecord<DataType>>,
     pub property_types: Vec<OntologyTypeRecord<PropertyType>>,
     pub entity_types: Vec<OntologyTypeRecord<EntityType>>,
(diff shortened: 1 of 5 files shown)


apps/hash-graph/lib/graph/src/store/test_graph.rs
Alfred-Mountfield, line 1: Should these be defined in the `store` module? If they're going to be deserialized further up in the Graph's stack I think this should be defined closer to the root

apps/hash-graph/lib/graph/src/store/test_graph.rs
```diff
@@ -0,0 +1,146 @@
...
+impl CustomEntityMetadata {
+    #[must_use]
+    const fn is_empty(&self) -> bool {
+        self.provenance.is_none() && self.archived.is_none()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EntityMetadata {
```
Alfred-Mountfield, line 34: Guessing there isn't a way for us to easily reuse the existing `EntityMetadata` struct?

apps/hash-graph/lib/graph/src/store/test_graph.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl From<Entity> for EntityRecord {
+    fn from(entity: Entity) -> Self {
+        Self {
+            properties: entity.properties,
+            metadata: EntityMetadata {
+                record_id: entity.metadata.record_id(),
+                entity_type_id: entity.metadata.entity_type_id().clone(),
+                temporal_versioning: Some(entity.metadata.temporal_versioning().clone()),
```
Alfred-Mountfield, line 59: Not important, but I'd guess (hope?) the compiler would optimise out these clones? 🤔

apps/hash-graph/lib/graph/src/store/test_graph.rs
```diff
@@ -0,0 +1,146 @@
...
+    #[expect(clippy::unused_self)]
+    const fn is_empty(&self) -> bool {
+        true
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TestData {
+    pub block_protocol_spec: semver::Version,
```
Alfred-Mountfield, line 139: I'd potentially change this to a struct, perhaps something like
```rust
struct BlockProtocolModuleVersions {
  graph: Option<semver::Version>
}
```

which would make it forwards compatible if we ever had other modules we needed to consider, e.g.

```rust
struct BlockProtocolModuleVersions {
  graph: Option<semver::Version>,
  authorization: Option<semver::Version>
}
```

Maybe it doesn't make sense to have graph as `Option` at the moment though, I guess we could add that in the future without it breaking previous serialized formats

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: It most probably won't optimize them and this is the reason why I pushed so hard to have `pub` fields.

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: Nope, sadly this is not possible. But maybe we should always use this struct everywhere and bring the HASH backend closer to the BP spec (not in this PR but as a potential follow-up in the near future)

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: I was unsure about this. We need to do some restructuring in the Graph library and can probably find a better location. For now, this felt like an appropriate place.

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: Why would you make these optional?

Alfred-Mountfield: > Why would you make these optional?

So that it's forwards compatible.

apps/hash-graph/lib/graph/src/store/test_graph.rs
Alfred-Mountfield: > But maybe we should always use this struct everywhere and bring the HASH backend closer to the BP spec

That is more aligned with what I was thinking, it might be nice to actually explicitly label our extensions as being extensions / custom

apps/hash-graph/lib/graph/src/store/test_graph.rs
Alfred-Mountfield: I agree, wasn't asking for one 😁

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: Yes, absolutely!

apps/hash-graph/lib/graph/src/store/test_graph.rs
Alfred-Mountfield: ^ can we do this then?

TimDiekmann: Yes, I'm about to do this now, just didn't got to it, yet 😄

Alfred-Mountfield: No worries, only asked because my review was re-requested

apps/hash-graph/lib/graph/src/store/test_graph.rs
```diff
@@ -133,10 +133,27 @@ impl CustomGlobalMetadata {
     }
 }
 
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockProtocolModuleVersions {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub graph: Option<semver::Version>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub authorization: Option<semver::Version>,
```
Alfred-Mountfield, line 142: Sorry this doesn't exist, it was just an example of what we might want to do in the future 😅 

I'm unclear if `graph` should be optional, I think let's make it not optional for now, we can always change it in a compatible way later on

Alfred-Mountfield (approved):

PR #2316, opened 2023-03-29
Add a `snapshot` subcommand to the CLI

TimDiekmann:

In #2314 support to read the data was added, this PR adds an interface to dump the data to the terminal.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204216809501005/f) _(internal)_

## 🚫 Blocked by

- #2313 
- #2314 
- #2310 (Will fail at runtime until merged)
- #2285 
- #2304 (compilation will fail until merged)

## 🔍 What does this change?

This adds a simple subcommand to dump `test_graph::TestData` to the terminal as JSON.

## ❓ How to test this?

```sh
cargo run -- snapshot --database dev_graph dump | jq
```

### DIFF
--- a/apps/hash-graph/bin/cli/src/main.rs
+++ b/apps/hash-graph/bin/cli/src/main.rs
@@ -22,5 +22,6 @@ async fn main() -> Result<(), GraphError> {
             subcommand::completions(args);
             Ok(())
         }
+        Subcommand::Snapshot(args) => subcommand::snapshot(args).await,
     }
 }
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -103,7 +103,7 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
             serde_json::to_writer(&mut std::io::stdout(), &test_graph)
                 .into_report()
                 .change_context(GraphError)?;
-            writeln!(std::io::stdout(), "")
+            writeln!(std::io::stdout())
                 .into_report()
                 .change_context(GraphError)?;
         }
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -42,18 +42,18 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
     let pool = PostgresStorePool::new(&args.db_info, NoTls)
         .await
         .change_context(GraphError)
-        .map_err(|err| {
-            tracing::error!("{err:?}");
-            err
+        .map_err(|report| {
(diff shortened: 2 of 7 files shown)


apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
+    let _log_guard = init_logger(&args.log_config);
+
+    let pool = PostgresStorePool::new(&args.db_info, NoTls)
+        .await
+        .change_context(GraphError)
+        .map_err(|err| {
+            tracing::error!("{err:?}");
```
Alfred-Mountfield, line 46: Shouldn't we just be passing `err` into the tracing macro rather than putting it in a format string?

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -0,0 +1,113 @@
...
+            err
+        })?;
+
+    match args.command {
+        SnapshotCommand::Dump(_) => {
+            let data_types = Read::<test_graph::OntologyTypeRecord<DataType>>::read(
```
Alfred-Mountfield, line 61: Don't have to do as a drive-by right now, but I think we should look at that module structure soon as we discussed. `test_graph` doesn't seem like a great module name, maybe we just call it `snapshot` to be consistent with this or go with `raw_db` or something like that.

apps/hash-graph/bin/cli/src/main.rs
Alfred-Mountfield, line 1: I can't find it in the diff, so I assume it's not in this PR, but I could just be missing it. 

How long have we had the `--data-base` argument? 

It should be `--database`, any chance you could update that as a drive-by and fix any usages if we do have any?

TimDiekmann: Actually this was wrong in the description, it's `--database`

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
TimDiekmann: Yes, that would probably be better. I could do this as a drive-by but then I should do this for all subcommands (the code is copied over from there). What do you think?

Alfred-Mountfield: Drive-by seems fine, I'm wondering if there may be an issue here but I'm think the macro falls back to a display/debug implementation. (I'm unclear because of the `Valuable` stuff)

TimDiekmann: Fixed: [Drive-by: Properly print error when acquiring connection](https://github.com/hashintel/hash/pull/2316/commits/e029acc9e1e434febdcdc394010fbf6db6582f8f)

Alfred-Mountfield (approved):

PR #2317, opened 2023-03-29
Fix temporal axes not taken into account when selecting distinctly

TimDiekmann:

Entities are uniquely identified by their ID and their temporal data. For common operations, we can only return a pinned axis, so we haven't noticed this issue yet, however, as we need to read all temporal data with #2314, this would get problematic and remove non-duplicates from the output set.


apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -64,16 +64,28 @@ impl<C: AsClient> crud::Read<Entity> for PostgresStore<C> {
...
+            &EntityQueryPath::Uuid,
+            Distinctness::Distinct,
+            None,
+        );
+        let decision_time_index = compiler.add_distinct_selection_with_ordering(
+            &EntityQueryPath::DecisionTime,
+            Distinctness::Distinct,
+            None,
+        );
+        let transaction_time_index = compiler.add_distinct_selection_with_ordering(
```
thehabbos007, line 82: Do we know if the `entity_temporal_metadata_overlapping` GIST index would help alleviate some of the overhead of multi-column `DISTINCT` queries? I assume it would help here, but I can imagine the query plan for this being slightly different now. We could measure/look at the `EXPLAIN` for the two queries to compare

TimDiekmann: I was about to investigate this at some point but didn't do this as part of this PR as the previous implementation was wrong.

thehabbos007: Yeah, good point. It doesn't matter if the performance is worse if the query was wrong before :smile:

thehabbos007 (approved):

PR #2322, opened 2023-03-30
Add a more verbose `.dockerignore`

TimDiekmann:

Currently, the `.dockerignore` is the same file as `.gitignore`, but we don't need to copy over some directories like `.git`.

This uses `.gitignore` as base (https://github.com/hashintel/hash/commit/c18d6601d8eaab770b59c35e81483eca35573807) and only adds additional entries (https://github.com/hashintel/hash/commit/96fcb40a62ef6cdbd881e93c2c1800d5db730803)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1203761629260701/f) _(internal)_


.dockerignore
Alfred-Mountfield, line 1: I don't think we want users having to do a visual diff of these if the structures don't line up.  I'd suggest we add a break between the two sections and make it explicit what's shared between the `.gitignore` and what's not.

e.g. 
```
### -------- Common with other .ignore files --------
...

### -------- Extended Ignores (unique to this file) --------
...
```

Those heading namings are terrible but my brain is struggling

.gitignore
```diff
@@ -1,3 +1,6 @@
+### IMPORTANT ###
+# When adding new filters please also consider adding them to `.dockerignore` if desired
```
Alfred-Mountfield, line 2: We should change this to be a warning to keep _**all**_ ignore files in sync. (I _think_ there might be an existing warning in some README or something somewhere if you could search for it, would definitely be better within the file though)

TimDiekmann: I haven't noticed we have those section in other ignore files already and just the _most important_ ignore file does not have this information. I added a commit to unify them all.

.dockerignore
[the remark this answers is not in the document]
Alfred-Mountfield: Ah I see it's consistent with `.prettierignore`.. let's change that heading too for consistency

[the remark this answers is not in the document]
TimDiekmann: GitHub having issues as this was a soft link before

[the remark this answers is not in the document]
Alfred-Mountfield: > GitHub having issues as this was a soft link before

Interesting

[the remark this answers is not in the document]
TimDiekmann: I changed all ignore-files to use this instead, thanks for the suggestion!

Alfred-Mountfield (dismissed):

.markdownlintignore
```diff
@@ -15,11 +15,11 @@
 ########################
 
 # Logs
-logs
-*.log
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
+**/logs
```
Alfred-Mountfield, line 18: Do you remember why these `**/`s were required in the first place? (I see you just reverted them for `prettier`)

I know the inconsistency was introduced in a previous PR but it'd be good if we didn't have to have actual different ignore patterns between the shared portion of the ignore configs..

Alfred-Mountfield: And have you noticed a difference between the behaviour of these ignore files vs `.gitignore`? I believe the behaviour of `.gitignore` is to match at any subdirectory if the pattern is not a path 

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. **Otherwise the pattern may also match at any level below the .gitignore level.**

- https://git-scm.com/docs/gitignore

TimDiekmann: For `.gitignore` this is not required, however, for `.dockerignore` it's required. As these files should be in sync I kept the syntax for `.gitignore`

Alfred-Mountfield: > As these files should be in sync I kept the syntax for .gitignore

Why did you [Revert .prettierignore](https://github.com/hashintel/hash/pull/2322/commits/fff01aca94d6eb4680f80303c8bf5c596148c893) then?

TimDiekmann: Because we are generating `.eslintignore` based on that. Please see [this slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1680175990406509?thread_ts=1680167187.910809&cid=C02TWBTT3ED) for the context.

Alfred-Mountfield: Hm, it makes me quite uncomfortable to introduce even more inconsistency in this PR. Let's fix ASAP

Alfred-Mountfield (approved):
As expressed in the open thread, I'd like to see us resolve that inconsistency ASAP

I'm going to work on a fix locally

PR #2325, opened 2023-03-30
Fix ESLint Ignore

Alfred-Mountfield:

https://github.com/hashintel/hash/pull/2322 adjusted our `.ignore` files, and in the process attempted to drive-by improve them. Unfortunately we encountered a problem where our bespoke eslint ignore script failed when we added a `**` prefix which is required by `.dockerignore` (as it doesn't keep the same semantics as `.gitignore`).

This fixes that, the lines were incorrectly being filtered out as belonging to a different workspace.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1204291503420522/f) _(internal)_

## 🚫 Blocked by

- [x] #2322

## 🔍 What does this change?

- ~~Makes the one-line change to fix the incorrect filtering~~ Fixes filtering by checking if the path overlaps the current workspace
- Reverts the revert of making `.prettierignore` consistent
- Adds a callout in the `.prettierignore` to say it's used for ESLint filtering

- This added a callout in the prettier ignore but we probably want to be better at highlighting how tooling actually works within our repo, there's a lot of indirection.

## ⚠️ Known issues

N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

CI

## ❓ How to test this?

Check CI is green

## 📹 Demo

CI


.prettierignore
```diff
@@ -88,48 +95,62 @@ typings/
...
+**/pip-delete-this-directory.txt
+**/.mypy_cache
+**/.pytest_cache
+**/.hypothesis
 
 # generated files
 *.gen.*
-blocks/**/generated
 *.tsbuildinfo
+blocks/**/generated
```
Alfred-Mountfield, line 142: Ah hm, I don't think this will work as intended

Alfred-Mountfield: There are some unnecessary patterns (e.g. `**/minesweeper/src/**` is equivalent to `minesweeper/src/**`) but that should be fine

Alfred-Mountfield: Yep, but it would have been easy to accidentally add something like this in the future, so I wanted to check it behaves as expected (we don't want to ignore `/blocks/**/potato/src/**` either 😁 )

libs/@local/eslint-config/generate-ignore-patterns.cjs
```diff
@@ -29,25 +114,43 @@ module.exports = (workspaceDirPath) => {
     .relative(monorepoRoot, workspaceDirPath)
     .replace(/\\/g, "/")}/`;
 
-  const sharedPatternsFromPrettierignore = match
+  const sharedPatternsFromPrettierIgnore = match
```
TimDiekmann, line 117: I'm not sure about this rename as it's called `.prettierignore` not `.prettier-ignore` or something like that. I find the new name more readable, though.

TimDiekmann (approved):
I did some local testing and everything works as expected, thank you!

PR #2326, opened 2023-03-30
Prune docker builder cache after building

TimDiekmann:

We currently encountering the issue (again) that the GH runner runs out of disk space. The reasons for this (most likely) are
- Too much things copied over as docker context (solved)
- The docker build cache is huge (6GB and only 14GB are guaranteed to be available)

Since recently we never encountered this issue as we didn't had to run `yarn install`, which also adds ~6 GB. Given, that the rust compilation also takes about 3.2 GB (the sum is already > 14 GB) it's _*very*_ likely, that this will cause a device-out-of-memory error.

To handle this we have two options:
1. Move out the docker image creation to a separated job (captured in [this Asana task](https://app.asana.com/0/1203193968491715/1203893407788517/f) _(internal)_
2. Remove the build-cache after the container has been built

The first option obviously is the better way to handle this issue, it's also by far more complicated, especially as we currently run everything through a single `setup.py` (we want to revisit this setup, but also completely out of scope for such an urgent error, captured in [this Asana ticket](https://app.asana.com/0/1203193968491715/1203911750455808/f) _(internal)_) as not every job needs this docker container to be present. There are ways around but we still somehow need to depend on the other job.

So this PR _*hopefully*_ solves this issue by using the second approach. Instead of building the docker image on-demand, a `just` recipe has been added to build it first. Afterwards, the build-cache is removed. The docker-image is then expected to be reused for the rest of the workflow.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204291503420534/f) _(internal)_

## 🔍 What does this change?

- Add a `just in-ci` recipe, where the passed command is only run if the `CI` environment variable is set
- Move the build-step for the docker container to a separated recipe
- In CI only: prune the builder cache

## ⚠️ Known issues

This solution is not the optimal solution but most probably a solution we can use in the short- and mid-term future. Please see the discussion at the top.

## 🐾 Next steps

Currently, no direct follow-ups are planned. I will further investigate the issue if this will not resolve the issue.
We could, however, easily safe an additional 2.8 GB by just removing the provided docker images.

## 🛡 What tests cover this?

- The CI will hopefully highlight any issues

## ❓ How to test this?

- Carefully watch the CI jobs and make sure we don't run any unnecessary steps
- Alternative: Look at my summary below 💡🙃 👇🏻  

## 📹 Demo

| When                        | Disk space usage (rounded) | Usage in `apps/hash_graph` | Docker usage                     |
|-----------------------------|----------------------------|----------------------------|----------------------------------|
|Baseline         | 0 GB            | 4.0 KB                     |                                  |
|checking out          | 0 GB                       | **2.0 MB**                     |                                  |
|warming up            | **6 GB**                       | 2.0 MB                     |                                  |
|installing Rust       | **7 GB**                       | 2.0 MB                     |                                  |
|installing tools      | **7 GB**                       | 2.0 MB                     | 2.8GB Images / 0B Cache          |
|building Docker image | 13 GB                      | 2.0 MB                     | 3.194 GB Images / **5.545 GB** Cache |
|clearing build cache  | 7 GB                       | 2.0 MB                     | 3.194 GB Images / 0 GB Cache     |
|running tests         | 10 GB                      | **3.2 GB**                     |                                  |

From the observation it's clear, that we easily hit the 14 GB limit and why we never had this issue before (we didn't have the warmup-step)


apps/hash-graph/.justfile
```diff
@@ -17,9 +17,14 @@ default:
 run *arguments:
   cargo run --profile {{profile}} --bin hash-graph -- {{arguments}}
 
+[private]
+docker-build:
+  @just yarn external-services build graph
+  @just in-pr docker builder prune --force
```
TimDiekmann, line 23:
```suggestion
  @just in-ci docker builder prune --force
```

apps/hash-graph/.justfile
thehabbos007, line 1: (comment location doesn't matter)
Since we're pruning the builder, does that mean we don't share the cache layers in future builds? Or is the cache uploaded before we prune the builder?
I suppose our cache usage is getting very large now, I'll take a quick look

thehabbos007: Uploaded caches don't seem too huge. But still interested to know if this impacts any of that

TimDiekmann: We never have uploaded the cache for the docker build

thehabbos007: Oh, that must be only the production CI task that does that :+1:

thehabbos007 (dismissed):

apps/hash-graph/.justfile
TimDiekmann: That's possible 🙂

.justfile
```diff
@@ -84,6 +84,17 @@ not-in-pr +command:
     {{command}}
   fi
 
+# Runs the provided command in CI only
+[private]
+[no-cd]
+in-ci +command:
+  #!/usr/bin/env sh
```
TimDiekmann, line 91: `set -euo pipefail` is not available in `sh`
```suggestion
  #!/usr/bin/env bash
```

thehabbos007 (approved):

PR #2327, opened 2023-03-30
Provide a `snapshot restore` command

TimDiekmann:

After reading all data from the store we want to write it back. This adds the interface to write back the data to the graph.

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1204086167143889/1204216809501006/f) _(internal)_

## 🚫 Blocked by

- #2316

## 🔍 What does this change?

- Moves reading logic to the library
- Wrap reading logic in a function
- Add writing entry point to CLI
- Temporarily fail with error when attempting to write

This PR:

## 🐾 Next steps

Implement the logic to insert the data

## 🛡 What tests cover this?

This is supposed to be used in tests

## ❓ How to test this?

The first commit does contain a `println!` wich helps debugging. In order to extract a snapshot, run
```sh
hash-graph snapshot dump --database dev_graph > snapshot.json
```

To read it back, run
```
cat snapshot.json | hash-graph snapshot restore --database dev_graph
# or
hash-graph snapshot dump --database dev_graph | hash-graph snapshot restore --database dev_graph
```

### DIFF
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -75,8 +75,6 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
                 .into_report()
                 .change_context(GraphError)?;
 
-            println!("{:#?}", test_data);
-
             store
                 .write_test_graph(test_data)
                 .await
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -71,12 +71,12 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
                 .change_context(GraphError)?;
         }
         SnapshotCommand::Restore(_) => {
-            let test_data: test_graph::TestData = serde_json::from_reader(std::io::stdin())
+            let snapshot: test_graph::TestData = serde_json::from_reader(std::io::stdin())
                 .into_report()
                 .change_context(GraphError)?;
 
             store
-                .write_test_graph(test_data)
+                .write_test_graph(snapshot)
                 .await
                 .change_context(GraphError)?;
         }
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -3,15 +3,13 @@ use std::io::Write;
 use clap::Parser;
 use error_stack::{IntoReport, Result, ResultExt};
 use graph::{
-    knowledge::Entity,
     logging::{init_logger, LoggingArgs},
     store::{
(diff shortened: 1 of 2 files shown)


apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -58,54 +60,25 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
...
-            };
-
-            serde_json::to_writer(&mut std::io::stdout(), &test_graph)
+            writeln!(std::io::stdout())
                 .into_report()
                 .change_context(GraphError)?;
-            writeln!(std::io::stdout())
+        }
+        SnapshotCommand::Restore(_) => {
+            let test_data: test_graph::TestData = serde_json::from_reader(std::io::stdin())
```
thehabbos007, line 74: I think the `test_` prefix for the export is a little strange. Would it make more sense to just call it `snapshot`?

apps/hash-graph/lib/graph/src/store/test_graph.rs
```diff
@@ -150,3 +153,65 @@ pub struct TestData {
...
+    clippy::trait_duplication_in_bounds,
+    reason = "False positive: the generics are different"
+)]
+impl<S> TestStore for S where
+    S: Read<OntologyTypeRecord<DataType>>
+        + Read<OntologyTypeRecord<PropertyType>>
+        + Read<OntologyTypeRecord<EntityType>>
+        + Read<Entity>
+{
+}
```
thehabbos007, line 217: What's this for?

thehabbos007 (commented):
Thanks, I have a couple comments :)

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
TimDiekmann: Good point, I renamed it in [Rename variable to `snapshot`](https://github.com/hashintel/hash/pull/2327/commits/50eda2236b4634982c8636809d5df281c1238d22)

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann: This is just a blanket implementation as we only use the `Read` trait to get the data. Most likely I need to change this in my follow-up but for now, that's the easiest way to expose the functionality.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
thehabbos007: I just saw in the other thread Alfie also pointed out the module structure for `test_graph`. I would agree that `snapshot` would make more sense. Don't mind it being a follow-up

TimDiekmann: I like `snapshot` as module name as well. Will use that name in the follow-up then 🙂

PR #2346, opened 2023-04-01
Countdown Block improvements

luisbettencourt:

This PR fixes an issue with the Countdown block where the cursor would lose its position when changing the title value.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204270112397116/f) _(internal)_

## 🔍 What does this change?

- Made it so the value of the title's input isn't set every time it changes, which was causing the cursor to lose its position in some browsers (firefox and safari for example);
- Added a keydown event listener to submit the changes whenever Enter is pressed.

This PR:

- [ ] does not modify any publishable libraries

Are blocks libraries? They do need publishing

### DIFF
--- a/blocks/countdown/src/countdown-title.tsx
+++ b/blocks/countdown/src/countdown-title.tsx
@@ -16,13 +16,33 @@ export const CountdownTitle: FunctionComponent<CountdownTitleProps> = ({
   const textareaEl = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
-    if (!textareaEl.current) return;
-    textareaEl.current.textContent = value ?? "";
-  }, [value]);
+    const textarea = textareaEl.current;
+    if (!textarea) return;
+    textarea.textContent = value ?? "";
+
+    // Calling this whenever value changes makes it so the cursor position is lost in some browsers
+    // (firefox and safari for example), so we leave it out of the dependency array
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    const textarea = textareaEl.current;
+    if (!textarea) return;
+
+    const onKeyDown = ({ code }: KeyboardEvent) => {
+      if (code === "Enter") {
+        textarea.blur();
+        onBlur();
+      }
+    };
+
+    textarea.addEventListener("keydown", onKeyDown);
+    return () => textarea.removeEventListener("keydown", onKeyDown);
+  }, [onBlur]);
 
   const handleChange = (evt: ChangeEvent<HTMLDivElement>) => {
     if (!textareaEl.current) return;
-    textareaEl.current.textContent = evt.target.textContent;
+
     onChangeText(evt.currentTarget.textContent ?? "");
   };
 
--- a/blocks/countdown/src/countdown-title.tsx
+++ b/blocks/countdown/src/countdown-title.tsx
@@ -26,20 +26,6 @@ export const CountdownTitle: FunctionComponent<CountdownTitleProps> = ({
     }
   }, [value]);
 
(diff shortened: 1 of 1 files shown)


luisbettencourt:
Would like for this to be tested in safari please, I don't own an apple device

blocks/countdown/src/countdown-title.tsx
```diff
@@ -16,13 +16,33 @@ export const CountdownTitle: FunctionComponent<CountdownTitleProps> = ({
...
+  useEffect(() => {
+    const textarea = textareaEl.current;
+    if (!textarea) return;
+
+    const onKeyDown = ({ code }: KeyboardEvent) => {
+      if (code === "Enter") {
+        textarea.blur();
+        onBlur();
+      }
+    };
```
CiaranMn, line 37: This can be an `onKeyDown` handler on the element itself. You should also use `key` rather than `code`, because `key === "Enter"` will cover the numpad's enter as well (whereas its code is `NumpadEnter`)
```ts
        onKeyDown={({ currentTarget, key }) => {
          if (key === "Enter") {
            currentTarget.blur();
            onBlur();
          }
        }}
```

blocks/countdown/src/countdown-title.tsx
```diff
@@ -16,13 +16,33 @@ export const CountdownTitle: FunctionComponent<CountdownTitleProps> = ({
...
-    textareaEl.current.textContent = value ?? "";
-  }, [value]);
+    const textarea = textareaEl.current;
+    if (!textarea) return;
+    textarea.textContent = value ?? "";
+
+    // Calling this whenever value changes makes it so the cursor position is lost in some browsers
+    // (firefox and safari for example), so we leave it out of the dependency array
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
CiaranMn, line 26: This means that if `value` is changed from the outside, this element will not update. 

One quick option to prevent the cursor issue is to only update the value if it is different from the element's current content.

Although to be honest I'm unsure why we're using a `contentEditable` div rather than an actual `input` here, which would probably avoid some of these issues.

luisbettencourt: There's some differences from `input` and `textarea` elements, namely that they automatically expand to take a set width and don't really expand naturally as the value is changed.

luisbettencourt: Made it so the value reacts to external changes in bfcf717bea86ee97f89575593564186883cbc974

blocks/countdown/src/countdown-title.tsx
luisbettencourt: Changed to use `key` instead of `code` in bfcf717bea86ee97f89575593564186883cbc974

CiaranMn: Right – this linting rule is flagging a real problem, which is that we're using a `div` as a text input, and if we need to do that we should add additional handling for things that would otherwise be handled natively.

For example:
- add `role="textbox"` to the `div` ([docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role))
- if the user presses `"Escape"`, cancel input

CiaranMn: Having the key handler done imperatively in an effect doesn't remove these problems, it just makes it impossible for the linter to tell they exist

CiaranMn (approved):

blocks/countdown/src/countdown-title.tsx
luisbettencourt: Thank you for the help, I wasn't aware of `role="textbox"`

PR #2353, opened 2023-04-03
`intellij-plugin`: privot approach

indietyp:

This radically simplifies the plugin handling of modules using a new approach. Trying to unify the `.rs` and the directory was a real pain. While `ModuleFile` (earlier in the PR) partially solved the problem, some core problems are retained purely because smooshing together directories and files is not compatible with the IntelliJ Framework like we would like it to be.

Instead, this takes a similar but much more minimalistic approach, even if sorted by type, it will pair up the `.rs` file and the directory together, in a way where the `.rs` file will always be on top. This is a lot less invasive and(!) retains all the capabilities.

The next step then (in a follow up PR), would be to add functionality, first envisioned in this PR, this includes:

* rename of `.rs` file should also rename directory (with referential checks)
* deletion of `.rs` file should also delete directory (with referential checks)
* move of `.rs` file should also move directory (with referential checks)

I am unsure how possible this is without pulling in the rs framework, but I have hope; while researching for `ModuleFile` I found multiple ways (using handlers) in which we could potentially supplement the existing handler from the Rust plugin.

This PR:

## 📹 Demo

image

as you can see here the distinction between files and folders except for module folders is still there, to make it a bit easier on the eyes we might want to replace the icon of directories that are modules. (Tho I am no artist^^)


TimDiekmann:
This feels a little bit like "Folders only on top" not selected:
image

indietyp:
I consciously decided to sort the folders after the files for a specific reason legibility (actually having to sort the folder after the file is more complex than the other way around).

If I have the `core/` folder open and I'd like to navigate/find the `core.rs` file in your example, that is quite hard with an open file tree. I need to either (A) close the folder or (B) follow the indentation and then select the file. Especially with a large open file tree, this can be taxing and is definitely not fun (I speak from experience^^). This way, it is easy. I scroll up, follow the indentation, and at the top, I am guaranteed to find the folder and file grouped together. This also mirrors what the Rust plugin does with `mod.rs` files (they are always on the top).

I wholeheartedly agree that it is a bit confusing right now that the rest of the UI is folder -> file, but I think the solution for that is that we (1) color the folder differently (2) create an icon for these "modules" (or steal the icon of the `mod.rs` files).

What are your thoughts on this?

apps/intellij-plugin/build.gradle.kts
```diff
@@ -16,7 +16,7 @@ repositories {
 // Configure Gradle IntelliJ Plugin
 // Read more: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html
 intellij {
-    version.set("2022.3.2")
+    version.set("2023.1")
```
TimDiekmann, line 19: Is this everything, which is needed to update to the latest version? (just for clarification for the future)

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleDirectoryNode.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.intellij.ide.projectView.impl.nodes.PsiDirectoryNode
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiDirectory
+
+class ModuleDirectoryNode(project: Project?, value: PsiDirectory, viewSettings: ViewSettings?): PsiDirectoryNode(project, value, viewSettings) {
+    // This is needed, as otherwise the `GroupByTypeComparator` won't fall back to the `AlphaComparator`
+    override fun getTypeSortWeight(sortByType: Boolean): Int = 0
+
+    // this is the same weight of a file, allows us to group them together
+    override fun getWeight(): Int = 20
```
TimDiekmann, line 15: What is the specific reason for `20`? Seems like a magic number.

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleDirectoryNode.kt
```diff
@@ -0,0 +1,36 @@
...
+    // if the weights of both files are the same, the `AlphaComparator` uses the `FileNameComparator` with the `toString`
+    // representation of a node. For the node to be able to be displayed above the directory we "emulate" the name of
+    // the file (and add `.mod` to ensure that we're always last).
+    // AFAIK `toString()` is not used anywhere else.
+    override fun toString(): String {
+        val name = super.toString()
+
+        return "$name.rs.mod"
+    }
+}
```
TimDiekmann, line 36:
```suggestion
}

```

TimDiekmann (commented):
The previous approach was prettier but less functional. I wonder if this approach can be applied to more than just Rust (specifically I have TS in mind as we have the same structure there as well).

The reasoning to order the folders after the files makes sense.

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -25,9 +25,6 @@
     <!-- Extension points defined by the plugin.
          Read more: https://plugins.jetbrains.com/docs/intellij/plugin-extension-points.html -->
     <extensions defaultExtensionNs="com.intellij">
-        <treeStructureProvider implementation="ai.hash.rustextra.projectView.ModuleGroupTreeStructureProvider" order="last, after org.rust.lang"/>
+        <treeStructureProvider implementation="ai.hash.rustextra.projectView.ModuleGroupTreeStructureProvider" order="last, before org.rust.lang"/>
```
TimDiekmann, line 28: Why this dependency is needed at all?

apps/intellij-plugin/build.gradle.kts
indietyp: This is just the version for the debug run configuration, the actual version for the plugin is set down below and has not been adjusted. I can revert to the old version; this is a remnant of testing.

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleDirectoryNode.kt
indietyp: That's because it is, the value needs to be the same as `PsiFileNode` so that we're able to group them, `PsiFileNode` assigns a static weight of `20` so we mirror that.

apps/intellij-plugin/src/main/resources/META-INF/plugin.xml
indietyp:
```suggestion
        <treeStructureProvider implementation="ai.hash.rustextra.projectView.ModuleGroupTreeStructureProvider" />
```

no longer needed, this was something accidentally left behind from the previous approach

indietyp:
Definitely! It should be super easy to also support this for `.ts` and `.js` files. I can do that in a follow-up PR.

apps/intellij-plugin/build.gradle.kts
indietyp:
```suggestion
    version.set("2022.3.2")
```

apps/intellij-plugin/src/main/kotlin/ai/hash/rustextra/projectView/ModuleDirectoryNode.kt
TimDiekmann: Is it possible to directly use that weight? If not could you set a comment?

indietyp: Sadly no, because `getWeight` is an instance method, and we inherit from `PsiDirectoryNode` and not `PsiFileNode`. I will add a comment clarifying this.

TimDiekmann: Ah, that's unfortunate. I guess it's not possible to call `PsiDirectoryNode::getWeight()`?

indietyp: Nope, not in Java land I am afraid ._.

TimDiekmann (approved):

PR #2355, opened 2023-04-04
Change snapshots to completely rely on streams and sinks

TimDiekmann:

For snapshots, we're going for a full-stream approach. This changes the `snapshot` subcommand to use streams and sinks.

## 🔗 Related links

- [Slack thread](https://hashintel.slack.com/archives/C03F7V6DU9M/p1680514674567089) _(internal)_
- [Asana task](https://app.asana.com/0/1204086167143892/1204320436419776/f) _(internal)_

## 🔍 What does this change?

- Move the `test-graph` module to `snapshot` and split the file into submodules for better maintainability
- Change the entrypoint for reading/writing to take a `Sink`/`Stream`

This PR:

## ⚠️ Known issues

`serde_json` does not have native support for async so it's not straightforward to use it in an async context. Currently, to bypass this constraint, this approach uses channels to make `serde_json` available as `Sink`/`Stream`.

## 🐾 Next steps

Implement a `JsonLines` codec to avoid channels. `Encoder` and `Decoder` are two simple traits which allows usage in `Framed` reading/writing.

## ❓ How to test this?

Remove the `TODO` in `restore_snapshot` and pass the output from `snapshot dump` to `snapshot restore`:
```sh
cargo run --all-features -- snapshot --database dev_graph dump | RUST_LOG=trace cargo run --all-features -- snapshot --database dev_graph restore
```


apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -49,36 +48,66 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
...
-    let mut store = pool
-        .acquire()
-        .await
-        .change_context(GraphError)
-        .map_err(|report| {
-            tracing::error!(error = ?report, "Failed to acquire database connection");
-            report
-        })?;
+    let mut store = SnapshotStore::new(
+        pool.acquire_owned()
```
TimDiekmann, line 52: For simplicity, I changed this to `acquire_owned`. This will be reverted as soon as I removed channels again in the next PR.

apps/hash-graph/lib/graph/src/store/test_graph.rs
TimDiekmann, line 1: The majority of this file was just moved but `git` is not able to detect this.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
TimDiekmann: This was reverted in #2357

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -0,0 +1,156 @@
...
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum SnapshotEntry {
+    Snapshot(SnapshotMetadata),
+    DataType(OntologyTypeRecord<DataType>),
+    PropertyType(OntologyTypeRecord<PropertyType>),
+    EntityType(OntologyTypeRecord<EntityType>),
+    Entity(EntityRecord),
```
Alfred-Mountfield, line 32: In a future PR do we perhaps want to rename these to `EntitySnapshotRecord` or something?

Just to try avoiding confusion with other kinds of 'records' we have within the system (especially when we have something like `RecordId`)

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -0,0 +1,156 @@
...
+        )
+        .chain(
+            Read::<Entity>::read(&self.0, &Filter::All(vec![]), None)
+                .await?
+                .map_ok(|entity| SnapshotEntry::Entity(entity.into())),
+        )
+        .forward(sink.sink_map_err(|report| {
+            report
+                .change_context(QueryError)
+                .attach_printable("failed to write record into sink")
```
Alfred-Mountfield, line 99: Do we want to make this typed? 

I think we should lower our usage of plain strings, these sorts of messages don't end up crossing API boundaries

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -0,0 +1,156 @@
...
+    )]
+    pub async fn restore_snapshot(
+        &mut self,
+        snapshot: impl Stream<Item = SnapshotEntry> + Send,
+    ) -> Result<(), InsertionError> {
+        let mut snapshot = pin!(snapshot);
+        while let Some(entry) = snapshot.next().await {
+            match entry {
+                SnapshotEntry::Snapshot(global) => {
+                    assert_eq!(
```
Alfred-Mountfield, line 122: I don't think we want to be panicking here

TimDiekmann: No, I just put something here to see the different snapshot entries propagating

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -49,36 +48,66 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
...
+            .change_context(GraphError)
+            .map_err(|report| {
+                tracing::error!(error = ?report, "Failed to acquire database connection");
+                report
+            })?,
+    );
+
+    // TODO: Use a thin wrapper around serde-json which implements `Sink` and `Stream` to avoid the
+    //       need for a separate thread.
+    let (sender, mut receiver) = futures::channel::mpsc::channel::<SnapshotEntry>(8);
```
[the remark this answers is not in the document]
TimDiekmann: It's randomly chosen. Maybe it should be `4`? 😂 
Any number should work (but I didn't want to use `0` or an unbounded channel to have some backpressure), so I chose 8. As I knew I was about to remove it in a follow-up I didn't think much about it.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -49,36 +48,66 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
...
+                    .write(&[b'\n'])
+                    .into_report()
+                    .change_context(GraphError)?;
+            }
+
+            reader
+                .await
                 .into_report()
+                .change_context(GraphError)?
                 .change_context(GraphError)?;
```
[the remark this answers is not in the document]
TimDiekmann: Yes, the reader can fail and sending via the channel can fail. It will be removed again in #2357

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: Good point, I renamed it now: [Rename EntityRecord and OntologyTypeRecord](https://github.com/hashintel/hash/pull/2355/commits/5f56af83173caac42841f6d1f85aab42ae400844)
I find the name pretty long but it's better than having the confusion in place. If we replace the data we send to the backend, we may want to change it to `Entity`?

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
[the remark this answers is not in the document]
Alfred-Mountfield: If we're removing it then no worries, I was guessing we'd want quite a much more than `8` actually 😂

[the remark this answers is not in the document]
TimDiekmann: To explain this a little bit more: this is just the buffer the channel can hold until it's blocking any further item to be pushed into the channel. E.g. for dumping the snapshot it can hold 8 entries before they has to be written to `stdout`. Having a low number implies that everything needs to be printed to stdout immediately (or at least written the buffer held by `BufWriter`). A high number or unbounded channels could imply that the database creates a lot of entries, which have not been written to `stdout` yet, so they have to be kept in memory.

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: Yes, I think it's good to have it from the start, thanks for flagging! [Improve error reporting for snapshot store](https://github.com/hashintel/hash/pull/2355/commits/a1b25241a3fc0ace4a0462854c9b28e68cc20d26)

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: With the latest changes, I return an error.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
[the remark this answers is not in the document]
Alfred-Mountfield: I went and read the `docs.rs` page and somehow managed to convince myself the buffer was in bytes, as opposed to num entries.. Thanks for clarifying

apps/hash-graph/lib/graph/src/snapshot.rs
Alfred-Mountfield: Much better, thanks!

apps/hash-graph/lib/graph/src/snapshot/error.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Debug)]
+pub enum SnapshotRestoreError {
+    Unsupported,
+}
+
+impl fmt::Display for SnapshotRestoreError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Unsupported => write!(f, "The snapshot contains unsupported entries"),
```
Alfred-Mountfield, line 30: Do we log the entries that are unsupported? Or does this just happen whenever there's a `serde` error (in which case I guess we don't know if it's unsupported or malformed)

TimDiekmann: Currently, any value is unsupported. The logic is not implemented at all and I just did some outputs to the log to test the stream interface generally. Every call to `restore_snapshot` currently will `panic`.

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -65,49 +67,47 @@ where
...
-        }))
-        .await
+        });
+
+        metadata_stream
+            .chain(data_type_stream.map_ok(SnapshotEntry::DataType))
+            .chain(property_type_stream.map_ok(SnapshotEntry::PropertyType))
+            .chain(entity_type_stream.map_ok(SnapshotEntry::EntityType))
+            .chain(entity_stream.map_ok(|entity| SnapshotEntry::Entity(entity.into())))
+            .map_err(|report| report.change_context(SnapshotDumpError::Read))
```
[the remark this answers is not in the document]
TimDiekmann: At this point, we do not have information, on which record it has failed as we already errored at this point (so we don't have the record present). The information is only available either when creating the stream (i.e. in `Read::read` when the record is created) or when attempting to sink it, e.g. when it's not possible to write into the sink. For the latter case, it's the `sink`s responsibility to add more information (why it failed, what failed, etc.)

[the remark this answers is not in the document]
Alfred-Mountfield: That's unfortunate, would be good to look at if we could improve that situation in a follow-up

Alfred-Mountfield (dismissed):

Alfred-Mountfield (approved):

PR #2356, opened 2023-04-04
`ci`: use `yq` to install all necessary components

indietyp:

This changes the rust ci action to install components, not via the profile but individually from the `rust-toolchain.toml`.

We can only do this now because last week, the runner image shipped with version 1.33 of `yq`, landing read toml support, without needing to install additional tools.

The runner-image is available on 99.99% of all runners (only five are left, but those have no image version tag and are always "left behind").

### DIFF
--- a/.github/actions/install-rust-toolchain/action.yml
+++ b/.github/actions/install-rust-toolchain/action.yml
@@ -18,7 +18,7 @@ runs:
       shell: bash
       run: |
         # regex for components that are only available on nightly and should not be installed on stable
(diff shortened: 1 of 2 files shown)


TimDiekmann (changes requested):
Could you temporarily disable the check for the `GITHUB_EVENT_NAME` in `setup.py` so we can see the complete CI matrix?

TimDiekmann (commented):
Looks good to me, thank you!
Let's revert this change and merge it 🎉

TimDiekmann (approved):

PR #2357, opened 2023-04-04
Use a codec to read/write JSON Lines

TimDiekmann:

Currently, we rely on channels, which is suboptimal. This replaces the channels by a framed read/write by providing a very basic codec for JSON Lines. This is a direct follow-up of #2355.

## 🔗 Related links

- #2355 

## 🚫 Blocked by

- #2355 

## 🔍 What does this change?

- Implement `codec::JsonLines`, a thin wrapper around `serde_json`
- Replace the implementation of channels with framed reads/writes

This PR:

### DIFF
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -59,7 +59,7 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
             store
                 .dump_snapshot(FramedWrite::new(
                     io::BufWriter::new(io::stdout()),
-                    codec::JsonLines::default(),
+                    codec::JsonLinesEncoder::default(),
                 ))
                 .await
                 .change_context(GraphError)?;
@@ -70,7 +70,7 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
             store
                 .restore_snapshot(FramedRead::new(
                     io::BufReader::new(io::stdin()),
-                    codec::JsonLines::default(),
+                    codec::JsonLinesDecoder::default(),
                 ))
                 .await
                 .change_context(GraphError)?;
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -1,6 +1,5 @@
 use clap::Parser;
 use error_stack::{Result, ResultExt};
-use futures::{SinkExt, TryStreamExt};
 use graph::{
     logging::{init_logger, LoggingArgs},
     snapshot::{codec, SnapshotEntry, SnapshotStore},
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -1,8 +1,9 @@
 use clap::Parser;
 use error_stack::{Result, ResultExt};
+use futures::{SinkExt, TryStreamExt};
 use graph::{
     logging::{init_logger, LoggingArgs},
-    snapshot::{codec, SnapshotStore},
+    snapshot::{codec, SnapshotEntry, SnapshotStore},
     store::{DatabaseConnectionInfo, PostgresStorePool, StorePool},
 };
 use tokio::io;
@@ -38,6 +39,7 @@ pub struct SnapshotArgs {
 
 pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
     let _log_guard = init_logger(&args.log_config);
+    SnapshotEntry::install_error_stack_hook();
 
     let pool = PostgresStorePool::new(&args.db_info, NoTls)
         .await
@@ -57,23 +59,35 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
     match args.command {
         SnapshotCommand::Dump(_) => {
             store
-                .dump_snapshot(FramedWrite::new(
-                    io::BufWriter::new(io::stdout()),
-                    codec::JsonLinesEncoder::default(),
-                ))
+                .dump_snapshot(
+                    FramedWrite::new(
+                        io::BufWriter::new(io::stdout()),
+                        codec::JsonLinesEncoder::default(),
+                    )
+                    .sink_map_err(|report| {
+                        report.attach_printable("Failed to write snapshot to stdout")
+                    }),
+                )
                 .await
-                .change_context(GraphError)?;
+                .change_context(GraphError)
+                .attach_printable("Failed to read snapshot from store")?;
 
             tracing::info!("Snapshot dumped successfully");
         }
         SnapshotCommand::Restore(_) => {
             store
-                .restore_snapshot(FramedRead::new(
-                    io::BufReader::new(io::stdin()),
-                    codec::JsonLinesDecoder::default(),
-                ))
+                .restore_snapshot(
+                    FramedRead::new(
+                        io::BufReader::new(io::stdin()),
+                        codec::JsonLinesDecoder::default(),
+                    )
+                    .map_err(|report| {
+                        report.attach_printable("Failed to read snapshot from stdin")
+                    }),
+                )
                 .await
-                .change_context(GraphError)?;
+                .change_context(GraphError)
+                .attach_printable("Failed to write snapshot to store")?;
 
             tracing::info!("Snapshot restored successfully");
         }
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -69,21 +69,10 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
         }
         SnapshotCommand::Restore(_) => {
             store
-                .restore_snapshot(
-                    FramedRead::new(
-                        io::BufReader::new(io::stdin()),
-                        codec::JsonLines::default(),
-                    )
-                    .filter_map(|value| async move {
-                        match value {
-                            Ok(value) => Some(value),
-                            Err(report) => {
-                                tracing::error!(error = ?report, "Failed to read snapshot record");
-                                None
-                            }
-                        }
-                    }),
-                )
+                .restore_snapshot(FramedRead::new(
+                    io::BufReader::new(io::stdin()),
+                    codec::JsonLines::default(),
+                ))
                 .await
                 .change_context(GraphError)?;
 
--- a/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
+++ b/apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
@@ -59,35 +59,25 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
     match args.command {
         SnapshotCommand::Dump(_) => {
             store
-                .dump_snapshot(
-                    FramedWrite::new(
-                        io::BufWriter::new(io::stdout()),
-                        codec::JsonLinesEncoder::default(),
-                    )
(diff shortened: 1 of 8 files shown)


apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -113,10 +115,12 @@ where
     )]
     pub async fn restore_snapshot(
         &mut self,
-        snapshot: impl Stream<Item = SnapshotEntry> + Send,
+        snapshot: impl Stream<Item = Result<SnapshotEntry, impl Context>> + Send,
```
TimDiekmann, line 118: This was the easiest possibility I've seen to notify the stream, that it's canceled.

indietyp: I assume you mean notifying the consumer of the stream that the stream is about to be canceled? (-> having a fallible stream)

TimDiekmann: Yes, exactly. the stream could be abortable, but this requires the handle to be passed around.

indietyp: In that case, how about using a [`TryStream`](https://docs.rs/futures/latest/futures/prelude/stream/trait.TryStreamExt.html)?

indietyp: If you do not need to pass any information, there are also [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)s

TimDiekmann: This is a `TryStream` 😄 
```rust
impl<S, T, E> TryStream for S
where
    S: Stream<Item = Result<T, E>> + ?Sized;
```

indietyp: Ay, but I specifically meant the `try_next` method from `TryStreamExt`.

TimDiekmann: Yes, but this is exactly what `TryStream`s are, a `Stream` with a `Result`. `TryStreamExt` only provides some convenient functions on `Ok` or `Err` variants.

> If you do not need to pass any information, there are also [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)s

I was thinking about [`AbortHandle`](https://docs.rs/futures/latest/futures/stream/struct.AbortHandle.html), but I think `CancellationToken` should also work. I will take a look at these, thanks! (but not in this PR)

TimDiekmann: With `TryStream` you need to pass in some information but sometimes you don't even have an error type.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
```diff
@@ -48,64 +49,45 @@ pub async fn snapshot(args: SnapshotArgs) -> Result<(), GraphError> {
...
+                    .sink_map_err(|report| {
+                        report.attach_printable("Failed to write snapshot to stdout")
+                    }),
+                )
                 .await
-                .into_report()
-                .change_context(GraphError)?
-                .change_context(GraphError)?;
+                .change_context(GraphError)
+                .attach_printable("Failed to read snapshot from store")?;
```
Alfred-Mountfield, line 73: Is this error correct?

Alfred-Mountfield: I think it'd be pretty strange to see:
> Failed to write snapshot to stdout
> Failed to read snapshot from store

Next to each other. "Failed to write snapshot to stdout" sounds like there was a problem actually writing to stdout

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -35,6 +37,63 @@ pub enum SnapshotEntry {
...
+                    }
+                }
+            }
+            Self::Entity(entity) => {
+                let entity_id = format!(
+                    "{}%{}",
+                    entity.metadata.record_id.entity_id.owned_by_id,
+                    entity.metadata.record_id.entity_id.entity_uuid
+                );
+                context.push_body(format!("entity: {entity_id}"));
```
Alfred-Mountfield, line 86: Do you want to be calling two `format!`s in a row?

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -122,10 +181,14 @@ where
...
     pub async fn restore_snapshot(
         &mut self,
-        snapshot: impl Stream<Item = SnapshotEntry> + Send,
+        snapshot: impl Stream<Item = Result<SnapshotEntry, impl Context>> + Send,
     ) -> Result<(), SnapshotRestoreError> {
         let mut snapshot = pin!(snapshot);
         while let Some(entry) = snapshot.next().await {
+            let entry = entry
+                .change_context(SnapshotRestoreError::Canceled)
+                .attach_printable("reading the records resulted in an error")?;
```
Alfred-Mountfield, line 190: This doesn't seem like the most helpful message, is there even a point adding this if we're just using a `Canceled` error

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: I fixed this in [Implement Display for EntityId](https://github.com/hashintel/hash/pull/2357/commits/aa500094b05ee365ee5ad1e43a71f9812b6d9cae)

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
TimDiekmann: These two errors apply to different locations. `"Failed to write snapshot to stdout"` will be printed if `FramedWrite` is failing. `"Failed to read snapshot from store"` is printed when the `dump_snapshot` command is failing, which could be anything like a database error.

Alfred-Mountfield: So what happens if `FramedWrite` fails? Does it not propagate up to `dump_snapshot` failing?

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: This error will happen when the input stream (i.e. the records, which are being read) is an `Err` variant. Ideally, we don't pass in `Result` at all, this was discussed in https://github.com/hashintel/hash/pull/2357#discussion_r1157198072. I most likely go with a cancelation token but this is out of the scope of this PR as I didn't make the discoveries for that, yet.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
TimDiekmann: It is, but it's at a different context where it makes sense:
```
Error: the Graph query layer encountered an error during execution
├╴at bin/cli/src/subcommand/snapshot.rs:72:18
├╴Failed to read snapshot from store
│
├─▶ could not write a snapshot record into the sink
│   ╰╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot.rs:142:52
│
╰─▶ test error which I manually inserted as I'm not smart enough to let writing to `stdout` fail
    ├╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot/codec.rs:32:13
    ├╴entity: 63049908-fc50-43d4-81da-ae180e0f3b53%0004a6a9-0a9e-4fa9-874a-ad742d7dc5da
    ╰╴Failed to write snapshot to stdout
```

apps/hash-graph/lib/graph/src/snapshot.rs
Alfred-Mountfield: Fair enough, I just meant that the display/debug impl of `SnapshotRestoreError::Canceled` basically says the same as `"reading the records resulted in an error"` so there's not much point adding additional (but unhelpful) information here.

Don't feel strongly, if we don't have more useful information to add at this point that's fair enough

TimDiekmann: I'd be fine with removing it. Any further information should already be in the report. This is all information we have at this point.

apps/hash-graph/bin/cli/src/subcommand/snapshot.rs
Alfred-Mountfield: I don't feel strongly about this, we can merge as is, but I still think this:
```
├╴Failed to read snapshot from store
│
├─▶ could not write a snapshot record into the sink
```

Is confusing, we aren't erroring while trying to read the snapshot, we're erroring while trying to write the snapshot

TimDiekmann: Hmm, do you think this is more clear?
```
Error: the Graph query layer encountered an error during execution
├╴at bin/cli/src/subcommand/snapshot.rs:68:18
├╴Failed to read snapshot from store
│
├─▶ could not write a snapshot record into the sink
│   ├╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot.rs:166:56
│   ╰╴Failed to write snapshot to stdout
│
╰─▶ deserializing entity failed
    ├╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot/codec.rs:32:13
    ╰╴entity: 63049908-fc50-43d4-81da-ae180e0f3b53%0004a6a9-0a9e-4fa9-874a-ad742d7dc5da
```
Or should we just remove it entirely?

Alfred-Mountfield: I think we can improve this by just replacing `Failed to read snapshot from store` with  `Failed to produce snapshot dump`

TimDiekmann: Or just
```
Error: the Graph query layer encountered an error during execution
├╴at bin/cli/src/subcommand/snapshot.rs:72:18
├╴Failed to produce snapshot dump
│
├─▶ could not write a snapshot record into the sink
│   ╰╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot.rs:166:56
│
╰─▶ deserializing entity failed
    ├╴at /Users/ghost/Development/hash/apps/hash-graph/lib/graph/src/snapshot/codec.rs:32:13
    ╰╴entity: 63049908-fc50-43d4-81da-ae180e0f3b53%0004a6a9-0a9e-4fa9-874a-ad742d7dc5da
```

Alfred-Mountfield: Yep

TimDiekmann: the first or the latter? 😂

Alfred-Mountfield: I don't feel strongly about the `Failed to write snapshot to stdout` bit. If it's possible for that (as in the stdout bit) to error, and if we think the errors won't encapsulate the information, let's add the message. Otherwise we can drop it

TimDiekmann: I removed it. If we ever encounter errors here and the information is not helpful we can add more attachments in the appropriated places. E.g. when reading from a file we may want to attach other information. If writing to `stdout` does not work I'm sure we have some error message which is helpful. If we, however, end up with errors like `"Resource temporarily unavailable"` I don't see it as very helpful (I never encountered those issues, yet).

Alfred-Mountfield (approved):

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: Currently, the attachment is misleading, so I removed it again. As I want to revisit cancelation, anyway, I don't really see the point in just adding an attachment for the sake of an attachment. I expect, that the error message already contains better information.

PR #2359, opened 2023-04-05
Fix database dumps for big data sets

TimDiekmann:

While testing the snapshot dump and restore IO I noticed that this is not working for large datasets. After testing with a dataset of ~25k entities I encountered an issue when using the same connection for multiple parallel reads as the returned stream is not resolved.

This PR fixes this by fetching the ontology types first (sequentially) and the entities afterward.

## 🚫 Blocked by

- #2357 

## 🔍 What does this change?

- Add function to create a stream via the `Read` trait more conveniently
- Read the data sequentially

This PR:


apps/hash-graph/lib/graph/src/snapshot.rs
Alfred-Mountfield, line 1: These changes seem like a hotfix. Do we know why the stream was being resolved and why it wasn't working?

Could we still run into problems on different hardware and splitting it just happened to work for you?
Do we need to rate limit our pools or something?

I think this is probably worht investigating now

TimDiekmann: As agreed offline I added a TODO to the section: [Add TODO to revisit the behavior](https://github.com/hashintel/hash/pull/2359/commits/b1d69221033ce954f901c195b7002fe5946e7c9a)

Alfred-Mountfield (approved):

PR #2373, opened 2023-04-07
Write OpenAPI specs as file to generate client

TimDiekmann:

The current setup to generate the OpenAPI client is rather complicated. This simplifies the process by allowing to write the specs into a folder.

## 🔍 What does this change?

- Generate the OpenAPI specs into a folder
- Adjust the client generator to use the generated files
- adjust the `.justfile` to use the new flag

This PR:


apps/hash-graph/lib/graph/src/api/rest.rs
```diff
@@ -223,7 +223,79 @@ async fn serve_static_schema(Path(path): Path<String>) -> Result<Response, Statu
...
+                .attach_printable_lazy(|| schema_path_dir.display().to_string())?;
+
+            for (schema_path, schema) in components.schemas {
+                let schema_path = schema_path_dir.join(&schema_path);
+                let schema = serde_json::to_string_pretty(&schema)
+                    .map_err(io::Error::from)
+                    .into_report()?;
+                fs::write(schema_path, schema)?;
+            }
+        }
```
thehabbos007, line 272: I think we skip over this writing step as it should all be contained in the top level schema

thehabbos007 (approved):

PR #2374, opened 2023-04-07
Properly express the Turborepo dependency graph

Alfred-Mountfield:

During the process of setting up Turborepo, we reprioritised work and the repo was left in an inconsistent state. This has lead to a number of interim patches such as disabling caching in CI.

This PR attempts to correctly express dependencies between packages/workspaces within the repository. This was made complicated by a few things
- We have Rust applications/libraries which interact with TypeScript
- The `package.json` setups of blocks are massively inconsistent and involved a fair amount of manual work.
- Our `codegen` step is sometimes used to produce artifacts for normal usage of the package (e.g. running in development, linting, building), and sometimes used to produce artifacts for downstream consumptions of the package. This makes it difficult to express it at the top of the repo with simplistic definitions like "build depends on codegen"
- Our `build` step is somewhat inconsistently used to mean "be ready for production usage" or "be ready for any usage at all", and this affects how we want dependencies to be expressed, especially with the extremely common "^build" expression. We do not want to be running full production builds of applications if we are importing elements of their API surface for example (particularly relevant with the `status` package)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1203641339878924/f) _(internal)_

## 🚫 Blocked by

- N/A

## 🔍 What does this change?

- Modifies the top-level `turbo.json` to be more general
- Adds `turbo.json` files to packages that require custom script definitions
- A few drive-by cleanups

This PR:

- We should probably invest some effort into describing the monorepo structure adequately in repo documentation. As we do not have that, there are no changes required right now.

## ⚠️ Known issues

- We inconsistently describe codegen steps (especially for blocks, with `schema` or `codegen`, this is a known issue due to out of date dependencies)
- The `build` semantics inconsistency described above
- A number of blocks are missing `eslint`, `tsc`, or `format` scripts
- The semantics of some scripts have been changed (previously some blocks had `codegen` calling `format` as part of the script). We perhaps want to revert some of those changes. I have made the changes to better represent the dependencies in the turbo graph, but if people `cd` into the package directories and simply run the scripts (for example, `yarn codegen` within a block directory) then they will not benefit from turborepo's dependency graph. We will need to use `turbo run` for this.
  - _Edit: this was reverted in [Revert removal of format from codegen command in blocks](https://github.com/hashintel/hash/pull/2374/commits/cc319247da67c3d43d5dd22488fe55f71eb9230b) as it caused problems where running `yarn` would result in a git diff because generated files weren't being formatted, and we don't want to reformat the repo on every `yarn` run_

## 🐾 Next steps

Unknown at this time

## 🛡 What tests cover this?

This should mostly impact CI, all linting, tests, builds, etc. are important to be checked

## ❓ How to test this?

1. View CI
2. Try scripts from the top level (e.g. `yarn install`)
3. Try individual workspaces such as blocks by running with `turbo run`
4. Try building the dependency graph with `turbo run build --graph=graph.svg` (you can change the line ending to `.pdf` or `.dot` for other formats)
  - The root dependency graph is pretty unreadable (see demo), you can check the dependency graph of individual packages with a filter, e.g.: `turbo run build --filter=@apps/hash-api --graph=graph.svg` 

## 📹 Demo

graph


.github/workflows/ci.yml
Alfred-Mountfield, line 1: Just commenting on this file to have a place to thread a conversation. The `test:unit` command seems to be failing in CI at the moment and I have no idea why. I also can't reproduce it locally. There's no output at all:

apps/hash-graph/.justfile
```diff
@@ -31,7 +31,43 @@ deployment-up *arguments: docker-build
...
+[private]
+miri *arguments:
+  @echo 'miri is disabled for `hash-graph`'
+
 # Generates the OpenAPI client for the Graph REST API
+[private]
+generate-openapi-client-shared:
+  #!/usr/bin/env bash
+  set -euo pipefail
+
```
TimDiekmann, line 65:
```suggestion
```

.github/workflows/ci.yml
Alfred-Mountfield: Possibly the case! Most of the jobs hadn't finished running at the time 😅, still very confused there isn't an error message or anything though

apps/hash-graph/.justfile
Alfred-Mountfield: [Remove redundant shebang](https://github.com/hashintel/hash/pull/2374/commits/9623d56e369499cbebef9748be2e9dd059209147)

.github/workflows/ci.yml
Alfred-Mountfield: Oops, the `turbo` dependency was too old and it wasn't actually picking up the individual `turbo.json` files of the various workspaces. We'll see if it's fixed now that it's been bumped

apps/hash-graph/.justfile
```diff
@@ -31,7 +31,40 @@ deployment-up *arguments: docker-build
...
+[private]
+miri *arguments:
+  @echo 'miri is disabled for `hash-graph`'
+
 # Generates the OpenAPI client for the Graph REST API
+[private]
+generate-openapi-client-shared:
+  just yarn workspace @local/hash-graph-client-generator generate
+  just yarn workspace @local/hash-graph-client prettier --write .
+  just yarn workspace @local/hash-graph-client fix:eslint
```
TimDiekmann, line 65: Sorry, that I have not noticed that before...
```suggestion
  @just yarn workspace @local/hash-graph-client-generator generate
  @just yarn workspace @local/hash-graph-client prettier --write .
  @just yarn workspace @local/hash-graph-client fix:eslint
```

Alfred-Mountfield: Oops now that we removed the shebang, good point

Alfred-Mountfield: [Quiet generate-openapi-client-shared steps](https://github.com/hashintel/hash/pull/2374/commits/4084a8097aef3c840bf1223c5b845be906e2d49f)

apps/hash-external-services/docker-compose.yml
```diff
@@ -163,11 +163,6 @@ services:
       RUST_BACKTRACE: 1
     ports:
       - "${HASH_GRAPH_API_PORT}:4000"
-    healthcheck:
-      test: ["CMD", "/hash-graph", "server", "--healthcheck"]
-      interval: 2s
-      timeout: 2s
-      retries: 10
```
thehabbos007, line 170: For anyone wondering, this was moved to the dockerfile of the graph

apps/hash-graph/.justfile
```diff
@@ -31,7 +31,40 @@ deployment-up *arguments: docker-build
...
+  cargo test -p graph-benches --benches --profile {{profile}} {{arguments}}
+  @just deployment-up graph --wait
+  @just yarn httpyac send --all {{repo}}/apps/hash-graph/tests/rest-test.http
+  @just deployment-down
+  @just generate-openapi-client
+
+[private]
+coverage *arguments:
+  @just deployment-up
+  cargo llvm-cov --workspace --all-features --all-targets {{arguments}}
```
thehabbos007, line 47: Any reason why this is different from the root justfile? I see that the root rule uses `nextest`, is there a reason why this wasn't working with the graph?

blocks/address/turbo.json
```diff
@@ -0,0 +1,28 @@
+{
+  "extends": ["//"],
```
thehabbos007, line 2: I just noticed this, what does the "//" mean here? I guess it means root, but why is it '//'? :smile:

thehabbos007 (commented):
Comments from initial pass over the files. Will do some local testing. Thank you!

blocks/address/turbo.json
Alfred-Mountfield: No idea why 😅 
https://turbo.build/repo/docs/core-concepts/monorepos/configuring-workspaces

apps/hash-graph/.justfile
TimDiekmann: `nextest` does not play well with `criterion`

```
error: creating test list failed

Caused by:
  for `graph-benches::bench/read_scaling`, command `/Users/ghost/Development/hash/apps/hash-graph/target/llvm-cov-target/debug/deps/read_scaling-3150a3fb42776272 --list --format terse` exited with code 2
--- stdout:

--- stderr:
Gnuplot not found, using plotters backend
error: Found argument '--format' which wasn't expected, or isn't valid in this context

        If you tried to supply `--format` as a value rather than a flag, use `-- --format`

USAGE:
    read_scaling-3150a3fb42776272 --list

For more information try --help

---
error: process didn't exit successfully: `/Users/ghost/.rustup/toolchains/nightly-2023-04-04-aarch64-apple-darwin/bin/cargo nextest run --manifest-path /Users/ghost/Development/hash/apps/hash-graph/Cargo.toml --target-dir /Users/ghost/Development/hash/apps/hash-graph/target/llvm-cov-target --workspace --all-features --all-targets` (exit status: 104)
cargo llvm-cov nextest     220.75s user 33.28s system 773% cpu 32.861 total
```

thehabbos007 (approved):

apps/hash-external-services/docker-compose.yml
TimDiekmann: I'm sorry, I know that you specifically asked about this change and I approved this change but I just noticed that this has to be reverted. The healthcheck here is specifically for the `server` subcommand, not a general purpose healthcheck. Not all subcommands has a healthcheck. I'll open a PR for that.

Alfred-Mountfield: Good point, cheers

PR #2375, opened 2023-04-07
Allow Report to be thrown as a JS error with wasm

Fedeparma74:

This allows `Report` to be thrown as a JS error, allowing the use of this library with wasm.

## 🔍 What does this change?

- New `wasm` feature that adds the optional `wasm-bindgen` dependency.
- Implements `Into<JsValue>` for `Report<Context>`, which subsequently allows the implementation of the [`ReturnWasmAbi`](https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/trait.ReturnWasmAbi.html) trait for `error_stack::Result`.

- The docs for `error-stack` need updating to explain that users who would like to throw `Report` as a JS error need to enable the `wasm` feature

### DIFF
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "wasm")]
+use alloc::string::ToString;
 use alloc::{boxed::Box, vec, vec::Vec};
 #[cfg(nightly)]
 use core::error::Error;
@@ -11,6 +13,8 @@ use std::process::ExitCode;
 
 #[cfg(feature = "spantrace")]
 use tracing_error::{SpanTrace, SpanTraceStatus};
+#[cfg(feature = "wasm")]
+use wasm_bindgen::JsValue;
 
 #[cfg(nightly)]
 use crate::iter::{RequestRef, RequestValue};
@@ -724,5 +728,12 @@ impl<Context> Extend<Self> for Report<Context> {
         for item in iter {
             self.extend_one(item);
         }
+    }
+}
+
+#[cfg(feature = "wasm")]
+impl<Context> From<Report<Context>> for JsValue {
+    fn from(report: Report<Context>) -> Self {
+        JsValue::from_str(&report.to_string())
     }
 }
--- a/libs/error-stack/Cargo.toml
+++ b/libs/error-stack/Cargo.toml
@@ -23,6 +23,7 @@ anyhow = { version = "1.0.70", default-features = false, optional = true }
 eyre = { version = "0.6", default-features = false, optional = true }
 serde = { version = "1", default-features = false, optional = true }
 spin = { version = "0.9", default-features = false, optional = true, features = ['rwlock', 'once'] }
+wasm-bindgen = { version = "0.2", optional = true }
 
 [dev-dependencies]
 serde = { version = "1.0.159", features = ["derive"] }
@@ -51,6 +52,7 @@ std = ["anyhow?/std"]
 eyre = ["dep:eyre", "std"]
 serde = ["dep:serde"]
 hooks = ['dep:spin']
+wasm = ['dep:wasm-bindgen']
 
 [package.metadata.docs.rs]
 all-features = true


libs/error-stack/src/report.rs
```diff
@@ -726,3 +730,10 @@ impl<Context> Extend<Self> for Report<Context> {
         }
     }
 }
+
+#[cfg(feature = "wasm")]
+impl<Context> From<Report<Context>> for JsValue {
+    fn from(report: Report<Context>) -> Self {
+        JsValue::from_str(&report.to_string())
```
Alfred-Mountfield, line 737: This is likely to be a pretty poor experience in most cases for users as the `Display` impl really doesn't encapsulate all of the information within the Report.

If we did go about supporting a `From` impl natively in `error-stack` I think we'd need to think more about what sort of representation we'd want to send

Alfred-Mountfield (changes requested):
Thanks for your contribution!

Unfortunately, as I've highlighted in a comment, we think this form of the solution could be potentially problematic and probably isn't the direction we want to go with wasm support. 

We're really excited about you trying to use this at the wasm boundary (we're [fans of wasm too](https://github.com/blockprotocol/blockprotocol/tree/main/libs/%40blockprotocol/type-system/crate)) but we're not sure this is the right long term solution.

Have you thought about changing the return type of your `wasm_bindgen` functions to `JsValue` itself, and then following the guidance of the recommended approach nowadays of using `[serde-wasm-bindgen](https://crates.io/crates/serde-wasm-bindgen)`?

We'll look at trying to get an example working ourselves soon and publish that somewhere too! But I hope it helps unblock you for now.

Alfred-Mountfield:
We'll leave this PR open for a bit in case we've missed something that made a native `From` implementation necessary, but otherwise we'll likely close it. Thanks again for the suggestion, we'll try and get that example live.

Fedeparma74:
What about using `serde-wasm-bindgen` directly in the `From` trait implementation? Implementing such a trait would be useful in the case where a library using `error-stack` is meant to be used both within a Rust application and as a wasm package. In this case, returning `JsValue` isn't really feasible, as it would be nice only when compiling for wasm targets

libs/error-stack/src/report.rs
```diff
@@ -732,8 +730,8 @@ impl<Context> Extend<Self> for Report<Context> {
 }
 
 #[cfg(feature = "wasm")]
-impl<Context> From<Report<Context>> for JsValue {
-    fn from(report: Report<Context>) -> Self {
-        JsValue::from_str(&report.to_string())
+impl<C: Context> From<Report<C>> for JsValue {
+    fn from(report: Report<C>) -> Self {
+        serde_wasm_bindgen::to_value(&report).expect("Failed to serialize report")
```
Alfred-Mountfield, line 735: > What about using serde-wasm-bindgen directly in the From trait implementation? Implementing such a trait would be useful in the case where a library using error-stack is meant to be used both within a Rust application and as a wasm package. In this case, returning JsValue isn't really feasible, as it would be nice only when compiling for wasm targets

Threading a reply here. This is definitely a lot better than the previous approach, but I think we're still apprehensive about adding this due to the implications for package bloat.

In my mind there are two aspects here
- Using `error-stack` in wasm contexts (with `wasm-bindgen`) is possible _without_ us adding native support for it in the library here. 
  - You're able to just call `serde_wasm_bindgen::to_value` yourself on the `Report` and create a new type (or just change your wasm function signatures to `JsValue` and not even have a new type).
- Adding support for something like `wasm-bindgen` then opens questions about what other libraries we might want to support.
  - It's true that we do have some features related to external libraries already (`eyre` and `anyhow` for example), but those are more about compatibility layers between _error_ libraries and are used to _create_ `Report`s rather than _consuming_ them.
  - If we do want to add `wasm-bindgen`, then do we also want to add support for `axum`, or `pyo3`, or `actix-web`? We're not convinced we do want to do that (we're also not convinced we don't want to), but we do have limited resources and maintaining native support for these things is not going to come without investment. 

I think for now, we'd prefer not to add this. I believe we're centering on the general philosophy of "we'll extend `error-stack` with things _if people can't do it themselves_.

nonparibus:
Hi @Fedeparma74, we're going to close this for now, but if you'd like to continue the conversation please feel free to pick the thread back up. Thanks for your original suggestion, and for opening this PR in the first instance.

Issue #1700, opened 2022-12-20
`deer` type implementation


indietyp:
This is the tracking issue for all type implementations needed for `deer`, separated by crate

- [ ] `core`
	- [ ] `bool`
	- [ ] `f*`
	- [ ] `u*`
	- [ ] `i*`
	- [ ] `char`
	- [ ] `&str`
	- [ ] `&[u8]`
	- [ ] `array
	- [ ] `AtomicBool`
	- [ ] `AtomicU*`
	- [ ] `AtomicI*`
	- [ ] `NonZeroU*`
	- [ ] `NonZeroI*`
	- [ ] `()`
	- [ ] `tuples`
	- [ ] `LazyCell`
	- [ ] `OnceCell` (`nightly`)
	- [ ] `SyncUnsafeCell`
	- [ ] `Cell`
	- [ ] `RefCell`
	- [ ] `UnsafeCell`
	- [ ] `PhantomData`
	- [ ] `ManuallyDrop`
	- [ ] `Option`
	- [ ] `Result`
	- [ ] `Wrapping`
	- [ ] `Bound`
	- [ ] `Range`
	- [ ] `RangeInclusive`
	- [ ] `Bound`
	- [ ] `Ip*`
	- [ ] `Reverse`
	- [ ] `Duration`
- [ ] `alloc`
    - [ ] `Cow`
    - [ ] `Box`
    - [ ] `ThinBox` (`nightly`)
    - [ ] `BinaryHeap`
    - [ ] `BTreeMap`
    - [ ] `BTreeSet`
    - [ ] `LinkedList`
    - [ ] `VecDeque`
    - [ ] `Vec`
    - [ ] `CString`
    - [ ] `Box<CStr>`
    - [ ] `Rc`
    - [ ] `String`
    - [ ] `Arc`
- [ ]  `std`
	- [ ] `HashMap`
	- [ ] `&Path`
	- [ ] `PathBuf`
	- [ ] `Box<Path>`
	- [ ] `Mutex`
	- [ ] `RwLock`
	- [ ] `OsString`
	- [ ] `SystemType`

PR #2378, opened 2023-04-09
`deer`: implement `Deserialize` for `core::num`

indietyp:

Implements Deserialize for:
* `core::num::Wrapping`
* `core::num::Saturating`

Orginally taken from #1875 

## 🔗 Related links

* #1700 
* #1875 

## 👂 Open Questions

Should `Wrapping` and `Saturating` be considered at deserialization time? Currently we just deserialize the value with the underlying value and then map `Wrapping` onto it. We will still fail if we supply `256` to an u8. This is currently what `serde` does.

Alternative would be only to implement `Wrapping` on all primitive num types, and if a larger number is encountered we simply modulo it on wrapping or set it to the max.

This PR:


libs/deer/tests/test_impls_core_num.rs
```diff
@@ -106,3 +116,45 @@ fn usize_err_zero() {
...
+        assert_tokens(&expected, &[Token::Number(Number::from(value))]);
+    }
+
+    #[test]
+    fn saturating_ok(value in any::<u8>()) {
+        let expected = Saturating(value);
+
+        assert_tokens(&expected, &[Token::Number(Number::from(value))]);
+    }
+}
```
TimDiekmann, line 135: Could we add some more tests for larger numbers like `usize`, `u64`, and `u128`? Also, it would be good if we do the same for `Wrapping` and `Saturating`.

TimDiekmann (changes requested):
Looks good, but a few more test cases would be nice. I guess this won't be hard 🙂

libs/deer/tests/test_impls_core_num.rs
indietyp: I can also add them to `Wrapping` and `Saturating`, but because they are simply `T::deserialize.map(Wrapping)` I thought it was a bit unnecessary if we would go another route (as proposed in the open questions), I think separate tests to look at saturating and wrapping behavior will be necessary.

Integral values are tested in https://github.com/hashintel/hash/blob/7633fdd6be8501adf4023dcea732a73ecff3abde/libs/deer/tests/test_impls_core_integral.rs

TimDiekmann (dismissed):

TimDiekmann (approved):

TimDiekmann (approved):

TimDiekmann (approved):

PR #2389, opened 2023-04-10
`deer`: implement `Deserialize` for `&[u8]`

indietyp:

Implements deserialize for `&[u8]` to allow for borrowed byte slices, taken from #1875 

## 🔗 Related links

* #1700 
* #1875 

This PR:


libs/deer/src/impls/core/bytes.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn visit_borrowed_bytes(self, v: &'de [u8]) -> error_stack::Result<Self::Value, VisitorError> {
+        Ok(v)
+    }
+}
+
+impl Reflection for [u8] {
+    fn schema(_: &mut Document) -> Schema {
+        // this type does not really exist in json-schema :/
+        // TODO: correct valid schema?
+        Schema::new("bytes")
```
thehabbos007, line 28: Do you have any ideas for how we could handle this in JSON Schema? And would it be problematic for us to be using `&[u8]` for JSON?

thehabbos007 (approved):
I think we can merge as-is still regardless of question above, but would like thoughts on it still :)

libs/deer/src/impls/core/bytes.rs
indietyp: `&[u8]` is invalid JSON; depending on the library, it is either rejected (JustJson) or strings that are not valid utf8 are passed as bytes (`serde-json`).

In the future, I plan to introduce a more robust system for schemas (project blueprint™️) that would solve this ambiguity by being agnostic over the output schema and having JSON-Schema be one of the output schemas.

For JSON-schema, specifically bytes data is typically encoded via `contentEncoding` (https://json-schema.org/understanding-json-schema/reference/non_json_data.html) so you would specify as encoding base64 and then deserialize into bytes.

PR #2390, opened 2023-04-10
`deer`: implement `Deserialize` for `core::time`

indietyp:

Implements `Deserialize` for:
* `core::time::Duration` (taken from #1875, same format as serde)

Based on #2376, due to new assertion function used for relative equality.

## 🔗 Related links

* #1875 
* #1700 

This PR:


libs/deer/tests/test_impls_core_time.rs
```diff
@@ -0,0 +1,17 @@
...
+use proptest::prelude::*;
+
+#[cfg(not(miri))]
+proptest! {
+    #[test]
+    fn duration_ok(value in any::<Duration>()) {
+        let input = value.as_secs_f64();
+
+        assert_tokens_with_assertion(|received: Duration| {
+            assert_relative_eq!(received.as_secs_f64(), value.as_secs_f64());
```
thehabbos007, line 14: Is `approx` used in future implementations? Or can we simply do `a - b < f64:EPSILON` for f64 comparisons?

thehabbos007 (commented):
Implementation looks fine, just have a smol question :)

libs/deer/tests/test_impls_core_time.rs
indietyp: nope not at all, we can indeed just use `a - b < f64:EPSILON`. Good catch!

indietyp: fixed in [9048070](https://github.com/hashintel/hash/pull/2390/commits/904807051ec402a73d550893bb9cfd1aa33cbc7b)

thehabbos007 (approved):

PR #2392, opened 2023-04-10
Exclude `.config/lints.toml` from publishing

TimDiekmann:

Externally, the `.config/lints.toml` cannot be used.

Please note, that this PR is from a fork of the repository to test external contributions.


indietyp (approved):
Thank you for your external contribution! 🎉 

I jest, but this looks good to me; without the mono repo, only the library itself the `.config/lints.toml` doesn't make sense as it takes into account the lints from the parent directories.

PR #2394, opened 2023-04-11
Parse query macro attributes

indietyp:

This parses the enum definition (and attributes) into a couple of types, which are then used to create the generated types.

This PR:


TimDiekmann (changes requested):
Generally, the PR looks good but currently has merge conflicts and lints failing. 
Are you planning on do more testing and check if the output is the expected output? I'm fine with doing this in a follow-up.

indietyp:
Yes, the plan is to also test the expanded output, though I haven't been able to find a good similar library to expect-test that does that.

Will fix the lints and merge conflicts!

indietyp:
The plan in general is to 

(A) have pass UI tests (can I access the built type?, did everything compile?)
(B) observe behaviour (when using the derive macro, is the builder working, etc.)

TimDiekmann (approved):

PR #2396, opened 2023-04-12
Introduce dependency review workflow step

nonparibus:

This is experimental. We want to see if it catches dependency vulnerabilities introduced in new PRs that would otherwise go undetected.

This PR:

## 🐾 Next steps

Monitor the workflow step's status. If after some time we haven't noticed it active, deliberately open a PR that should cause the workflow to fail.


TimDiekmann (approved):
It appears that the action [has to be configured](https://github.com/actions/dependency-review-action#inline-configuration). This is also possible using a [configuration file](https://github.com/actions/dependency-review-action#configuration-file).

It seems that newly added dependencies are still checked, but for checking the license, a configuration should probably be added. I think it's fine to merge it as it is and tweak it in a follow-up.

nonparibus:
@TimDiekmann The default options mostly look fine, although:
- we should probable set `license-check` to `false` as we already use FOSSA for this and the check is duplicative
- we may want to expand `fail-on-scopes` to include `development` as well as `runtime` (it only includes the latter by default)

I agree we can do these in follow-up. Captured [here](https://app.asana.com/0/1203193968491715/1204386136612605).

TimDiekmann (approved):

PR #2398, opened 2023-04-12
Implement snapshot restore logic

TimDiekmann:

Currently, we don't have the logic implemented to restore a snapshot and write it into the store. This PR adds the logic to insert.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204216809501006/f) _(internal)_

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2381
- https://github.com/hashintel/hash/pull/2382

## 🔍 What does this change?

The data emitted by the stream is read in a separate thread and is sent to different channels for each record type. Each channel holds a buffer of `chunk_size` entries. The receivers of the channels are then used to insert the records into the store. When a write operation to the store succeeds, the next entry is read from the channel, even if the buffer of the channel is not full yet. This ensures, that the store is continuously writing to the database and does not wait for the buffer to be full.
    
Writing to the store happens in three stages:

1. The first stage is the `begin` stage. This stage is executed before any records are read from the stream. It is used to create a transaction, so a possible rollback is possible. For each data, which is inserted, a temporary table is created. This table is used to insert the data into the store without locking the store and avoiding yet unfulfilled foreign key constraints.
2. The second stage is the `write` stage. This stage is executed for each record type. It reads the batch of records from the channels and inserts them into the temporary tables.
3. The third stage is the `commit` stage. This stage is executed after all records have been read from the stream. It is used to insert the data from the temporary tables into the store and to drop the temporary tables. As foreign key constraints are now enabled, this stage might fail. In this case, the transaction is rolled back and the error is returned.
    
If the input stream contains an `Err` value, the snapshot restore is aborted and the error is returned.

This PR:

## ⚠️ Known issues

This basically duplicates the functionality we have to bulk-insert for benchmarks. This approach is a lot faster and more flexible and will replace the old approach somewhat soon.

## 🐾 Next steps

The functionality needs to be exposed to the typescript integration tests to actually being used for testing.

## 🛡 What tests cover this?

Currently, there are no tests covering this. it will, however, be used to write new test cases and benchmarks.

## ❓ How to test this?

Use a populated database and export the snapshot using `snapshot dump`. The resulting output (you should pipe it into a file) can be read using `snapshot restore`. Exporting a new snapshot from the restored data should result in the same output (when comparing, the key ordering is not deterministic).

## 📹 Demo

A database with about 25k entities was exported in less than a second and then got restored. This was the output:

```text
2023-04-10T18:29:24.156077Z  INFO graph::snapshot: snapshot restore started
2023-04-10T18:29:24.207779Z  INFO graph::snapshot::account::batch: Read 1 accounts
2023-04-10T18:29:24.216588Z  INFO graph::snapshot::ontology::batch: Read 66 ontology ids
2023-04-10T18:29:24.222246Z  INFO graph::snapshot::ontology::data_type::batch: Read 3 data type schemas
2023-04-10T18:29:24.227119Z  INFO graph::snapshot::ontology::property_type::batch: Read 43 property type schemas
2023-04-10T18:29:24.231863Z  INFO graph::snapshot::ontology::entity_type::batch: Read 20 entity type schemas
2023-04-10T18:29:24.254909Z  INFO graph::snapshot::entity::batch: Read 5747 entity ids
2023-04-10T18:29:24.264806Z  INFO graph::snapshot::ontology::batch: Read 62 ontology owned metadata
2023-04-10T18:29:24.268359Z  INFO graph::snapshot::ontology::property_type::batch: Read 36 property type value constrains
2023-04-10T18:29:24.271526Z  INFO graph::snapshot::ontology::entity_type::batch: Read 8 entity type inheritance
2023-04-10T18:29:24.420125Z  INFO graph::snapshot::entity::batch: Read 9027 entity editions
2023-04-10T18:29:24.452030Z  INFO graph::snapshot::ontology::batch: Read 4 ontology external metadata
2023-04-10T18:29:24.455289Z  INFO graph::snapshot::ontology::property_type::batch: Read 17 property type property type constrains
2023-04-10T18:29:24.458365Z  INFO graph::snapshot::ontology::entity_type::batch: Read 42 entity type property constrains
2023-04-10T18:29:24.621307Z  INFO graph::snapshot::entity::batch: Read 10000 entity temporal metadata
2023-04-10T18:29:24.629395Z  INFO graph::snapshot::ontology::entity_type::batch: Read 11 entity type link constrains
2023-04-10T18:29:24.669117Z  INFO graph::snapshot::entity::batch: Read 10000 entity links
2023-04-10T18:29:24.678366Z  INFO graph::snapshot::ontology::entity_type::batch: Read 12 entity type link destination constrains
2023-04-10T18:29:24.708138Z  INFO graph::snapshot::entity::batch: Read 10000 entity ids
2023-04-10T18:29:24.860429Z  INFO graph::snapshot::entity::batch: Read 9803 entity editions
2023-04-10T18:29:25.068576Z  INFO graph::snapshot::entity::batch: Read 10000 entity temporal metadata
2023-04-10T18:29:25.111708Z  INFO graph::snapshot::entity::batch: Read 9562 entity links
2023-04-10T18:29:25.139571Z  INFO graph::snapshot::entity::batch: Read 9267 entity ids
2023-04-10T18:29:25.228520Z  INFO graph::snapshot::entity::batch: Read 6184 entity editions
2023-04-10T18:29:25.339071Z  INFO graph::snapshot::entity::batch: Read 5014 entity temporal metadata
2023-04-10T18:29:25.339230Z  INFO graph::snapshot: snapshot reading finished, committing...
2023-04-10T18:29:29.142464Z  INFO hash_graph::subcommand::snapshot: Snapshot restored successfully
```


apps/hash-graph/lib/graph/src/snapshot/ontology/batch.rs
```diff
@@ -0,0 +1,125 @@
...
+#[async_trait]
+impl<C: AsClient> WriteBatch<C> for OntologyTypeMetadataRowBatch {
+    async fn begin(postgres_client: &PostgresStore<C>) -> Result<(), InsertionError> {
+        postgres_client
+            .as_client()
+            .client()
+            .simple_query(
+                r"
+                    CREATE TEMPORARY TABLE ontology_ids_tmp
+                        (LIKE ontology_ids INCLUDING ALL)
```
thehabbos007, line 28: Really awesome we can use `LIKE` here instead of manually creating the tables!

apps/hash-graph/lib/graph/src/snapshot/ontology/property_type/batch.rs
```diff
@@ -0,0 +1,144 @@
...
+                if !rows.is_empty() {
+                    tracing::info!("Read {} property type value constrains", rows.len());
+                }
+            }
+            Self::ConstrainsProperties(properties) => {
+                let rows = client
+                    .query(
+                        r"
+                            INSERT INTO property_type_constrains_properties_on_tmp
+                            SELECT DISTINCT * FROM UNNEST($1::property_type_constrains_properties_on_tmp[])
```
thehabbos007, line 96: How does this casting work? Is it the name of the table we cast as and `to_sql` knows how to convert records into the expected shape?

apps/hash-graph/lib/graph/src/snapshot/entity/batch.rs
```diff
@@ -0,0 +1,193 @@
...
+                r"
+                    CREATE TEMPORARY TABLE entity_ids_tmp
+                        (LIKE entity_ids INCLUDING ALL)
+                        ON COMMIT DROP;
+
+                    CREATE TEMPORARY TABLE entity_editions_tmp (
+                        entity_edition_id UUID PRIMARY KEY,
+                        properties JSONB NOT NULL,
+                        left_to_right_order INT4,
+                        right_to_left_order INT4,
```
thehabbos007, line 36: It's nice to have the size of the int in the type name here, but after looking at the docs, it seems that `integer` is more standard to SQL https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT

thehabbos007 (commented):
Thank you a lot, Tim! The code is pretty easy to follow, just some noise with the sink impl, but that's all good. 

I have some comments and will be testing it locally now :)

apps/hash-graph/lib/graph/src/snapshot/ontology/batch.rs
TimDiekmann: Yeah, this is really helpful. I did this for almost all tables, where it's possible. I, however, merged a few tables like the entity links to avoid having duplicated ids stored in two tables. I don't know if this is actually faster, but at least it's easier to read (less boilerplate)

apps/hash-graph/lib/graph/src/snapshot/restore/channel.rs
```diff
@@ -0,0 +1,150 @@
...
+
+impl Stream for SnapshotRecordReceiver {
+    type Item = SnapshotRecordBatch;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.stream.poll_next_unpin(cx)
+    }
+}
+
+pub fn channel(chunk_size: usize) -> (SnapshotRecordSender, SnapshotRecordReceiver) {
```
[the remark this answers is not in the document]
TimDiekmann: I don't have inspected the implementation but I guess they have something like `VecDeque::with_capacity(chunk_size)` internally. Even with a chunk size of `100_000` I never got above a few GB of memory usage at the same time. As we collect the channels in `ready_chunk`, which holds a `Vec` internally, we have a buffer of `20_000` per channel but this is still not very big considering we don't have a `Vec<Vec>` but only `N * Vec`.
I artificially increased my available dataset and had no issues at all. This is still a thing we should observe when inserting big datasets.

> though this was with tokio's mpsc, futures may handle this differently.

I think the implementation is more or less the same

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -176,57 +184,92 @@ where
...
+    /// operation to the store succeeds, the next entry is read from the channel, even if the
+    /// buffer of the channel is not full yet. This ensures, that the store is continuously writing
+    /// to the database and does not wait for the buffer to be full.
+    ///
+    /// Writing to the store happens in three stages:
+    ///   1. The first stage is the `begin` stage. This stage is executed before any records are
+    ///      read from the stream. It is used to create a transaction, so a possible rollback is
+    ///      possible. For each data, which is inserted, a temporary table is created. This table
+    ///      is used to insert the data into the store without locking the store and avoiding
+    ///      yet unfulfilled foreign key constraints.
```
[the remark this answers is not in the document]
TimDiekmann: I clarified this in [Clarify documentation on restore function](https://github.com/hashintel/hash/pull/2398/commits/f308cc6c31d99a97b932962b5e7d06fa6bc9f94f).
I like to keep the description as they reflect the steps of `WriteBatch`.

> Or are we dealing with two of temporary table for each kind of data?

No, we only use one temporary table per keind

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -176,57 +184,92 @@ where
...
+            .fold(
+                Ok(client),
+                |client, records: SnapshotRecordBatch| async move {
+                    let client = client?;
+                    records
+                        .write(&client)
+                        .await
+                        .change_context(SnapshotRestoreError::Write)?;
+                    Ok::<_, Report<SnapshotRestoreError>>(client)
+                },
```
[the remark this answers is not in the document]
TimDiekmann: Actually, I'm not 100% about the behavior. I haven't used `try_fold` as it requires the `TryStream` instead of `Stream` (`Item = Result<...>`). I agree, that `try_fold` is easier to read. By mapping the stream I was able to use `try_fold` instead: [Use try_fold instead of fold](https://github.com/hashintel/hash/pull/2398/commits/9695a312394e94775a8b962a252cdeca1304f462)

apps/hash-graph/lib/graph/src/snapshot/restore/channel.rs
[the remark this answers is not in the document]
thehabbos007: Okay, that's good. We should be a little cautious with this as GH runners have 7GB of memory, which could be a problem if we start using more and more. Just wanted to make sure :+1:

apps/hash-graph/lib/graph/src/snapshot.rs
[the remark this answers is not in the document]
TimDiekmann: Thinking about it again you were probably right, that it would iterate the whole stream before actually emitting the first error.

apps/hash-graph/lib/graph/src/snapshot/ontology/property_type/batch.rs
thehabbos007: Very cool, good to know!

apps/hash-graph/lib/graph/src/snapshot/entity/batch.rs
thehabbos007: Yeah my immediate thought was to use INT4 in the migrations too, but after checking the docs it seems non-standard :sweat_smile:

PR #2412, opened 2023-04-13
Tweak attachments for `restore` operation

TimDiekmann:

Currently, on failure, we attach the line to decode. This line may gets very long (in my test when I passed in the wrong format a single line was 27 MB long). We still have the information, in which column the error happens but the line is not attached anymore. In a follow-up I'm using the restore operation in a REST endpoint. As we want to return an appropriated status code, this attaches the status in the critical section of the restore logic.

This PR:


apps/hash-graph/lib/graph/src/snapshot/codec.rs
```diff
@@ -83,7 +83,6 @@ impl<T: DeserializeOwned> Decoder for JsonLinesDecoder<T> {
                 serde_json::from_str(&line)
```
thehabbos007, line 83: (comment for just above)
Could we perhaps attach a line number by using `enumerate`, or would the line number not be accurate?

thehabbos007 (dismissed):

apps/hash-graph/lib/graph/src/snapshot/codec.rs
TimDiekmann: That's a good suggestion, thank you! I added it in [Attach current line number](https://github.com/hashintel/hash/pull/2412/commits/fe4e27d11c9ec856d1877226ad07b227bca03261)

thehabbos007 (approved):

PR #2414, opened 2023-04-13
Fix `entity_link_edges_tmp` not being dropped on commit

TimDiekmann:

This was forgotten in #2398 

This PR:


thehabbos007 (approved):
Oh good you found this. I wonder if there is a transaction-wide option that can set this for all temporary tables for a transaction so we don't have to rely on the keyword (probably doesn't exist). 
Thank you!

PR #2415, opened 2023-04-13
Return a `Stream` instead of taking a `Sink` in `dump_snapshot`

TimDiekmann:

Eventually, we want to expose dumping snapshots to the REST endpoints as well. As `axum` has no supports for `Sink`s, this changes the return type for `dump_snapshot` to be a `Stream` instead of requiring a `Sink` to be passed in. Generally, this simplifies the usage as the function is not `async` anymore. In the CLI interface it's still possible to use a sink by just `forward`ing it into the `Sink`.


thehabbos007 (approved):
Thanks! seems more ergonomic this way as well

PR #2417, opened 2023-04-13
Fix blocks loading in HASH. Update dependencies. Delete toggle-item block

CiaranMn:

Makes changes (listed below) which have the effect of:
1. Blocks no longer loading multiple versions of `@blockprotocol/core` 
2. The HASH frontend's `yarn dev` mode working

It also deletes the `toggle-item` block as this is well out of date and unused.

Note that text blocks do not work as the hook service handling needs updating in the HASH frontend.

## 🔗 Related links

- [Asana task to update consistencies](https://app.asana.com/0/1204355839255039/1204355839255063/f) _(internal)_
- [Asana task to fix blocks loading in HASH (errors due to version inconsistency)](https://app.asana.com/0/1204355839255039/1204355839255060/f) _(internal)_

## 🔍 What does this change?

- Update `@blockprotocol/*` packages to be consistent across the repo, except for blocks which haven't been updated to `0.3` (and `hashdotdev`, which loads one such block, the calculation block) – stops blocks loading multiple versions of `@blockprotocol/core` by getting rid of nesting of it in some packages
   - this involves fixing some types up in the frontend 
- Fix circular imports in `@hashintel/design-system` and introduce a linting rule to avoid them in future (was causing the frontend to fail to run in dev mode)
- Transpile `react-syntax-highlighter` in the frontend (was causing the frontend to fail to run in dev mode)
- Makes the `dev.tsx` files in blocks more consistent (except outdated blocks):
  - add `simulateDatastoreLatency`
  - make use of environment variables consistent in blocks that need API keys

This PR:

## ❓ How to test this?

1.  Checkout the branch
2. Run `yarn dev`
3. Load a block


CiaranMn:
Missed a few updates in blocks (related to package.json), back to draft

nonparibus:
I had removed some of our old blocks in https://github.com/hashintel/hash/pull/2401 - originally I had left the `toggle-item` block as I assumed we wanted a toggle list, still, but have mirrored the removal there, too.

apps/hash-frontend/src/components/sandbox/framed-block/framed-block.tsx
```diff
@@ -27,13 +24,16 @@ export const FramedBlock: FunctionComponent = () => {
 
   const initialData = properties ? JSON.parse(properties) : undefined;
 
-  const [blockProperties, setBlockProperties] = useState<
-    BlockProtocolEntity | undefined
-  >(initialData);
+  const [blockProperties, setBlockProperties] = useState<Entity | undefined>(
```
Alfred-Mountfield, line 27: Should we rename this to `blockEntity`? As it's `Entity` not `EntityPropertyObject`

blocks/ai-chat/src/dev.tsx
```diff
@@ -144,8 +144,10 @@ const DevApp = () => {
...
       }}
-      blockProtocolApiKey={process.env.BP_API_KEY}
-      blockProtocolSiteHost={process.env.BP_HOST ?? "https://blockprotocol.org"} // update this to a recent staging deployment when testing
+      blockProtocolApiKey={process.env.BLOCK_PROTOCOL_API_KEY} // add this to an .env file in the block folder
+      blockProtocolSiteHost={
+        process.env.BLOCK_PROTOCOL_SITE_HOST ?? "https://blockprotocol.org"
+      } // update this to a recent staging deployment when testing
       // includeDefaultMockData // this seeds the datastore with sample entities and links, remove this to start with just the contents of `initialData`
       // hideDebugToggle <- uncomment this to disable the debug UI entirely
       // initialEntities={[]} <- customise the entities in the datastore (blockEntity is always added, if you provide it)
```
Alfred-Mountfield, line 153: This (and the 3 below but I can't comment) can be deleted

blocks/ai-image/src/dev.tsx
```diff
@@ -44,8 +44,10 @@ const DevApp = () => {
...
       }}
-      blockProtocolApiKey={process.env.BP_API_KEY}
-      blockProtocolSiteHost={process.env.BP_HOST ?? "https://blockprotocol.org"} // update this to a recent staging deployment when testing
+      blockProtocolApiKey={process.env.BLOCK_PROTOCOL_API_KEY} // add this to an .env file in the block folder
+      blockProtocolSiteHost={
+        process.env.BLOCK_PROTOCOL_SITE_HOST ?? "https://blockprotocol.org"
+      } // update this to a recent staging deployment when testing
       // includeDefaultMockData // this seeds the datastore with sample entities and links, remove this to start with just the contents of `initialData`
       // hideDebugToggle <- uncomment this to disable the debug UI entirely
       // initialEntities={[]} <- customise the entities in the datastore (blockEntity is always added, if you provide it)
```
Alfred-Mountfield, line 53: Same as other comment, I'll avoid making any more of these to avoid spam on the PR but we can remove all of the `initialEntities` args and such. They were accidentally left in the templates a while ago and have been removed since.

blocks/heading/package.json
```diff
@@ -25,15 +25,15 @@
...
-    "@blockprotocol/graph": "0.1.0",
-    "@blockprotocol/hook": "0.1.0"
+    "@blockprotocol/graph": "0.2.2",
+    "@blockprotocol/hook": "0.1.3"
   },
   "devDependencies": {
     "@local/eslint-config": "0.0.0-private",
     "@local/tsconfig": "0.0.0-private",
-    "block-scripts": "0.1.1",
+    "block-scripts": "0.2.3",
```
Alfred-Mountfield, line 34: Do you not need to fix (or rather introduce) the "codegen" object in these blocks?

Alfred-Mountfield (commented):
Changes look good to me aside from a few things
- Blocks seem to need a few more things done to update them properly (codegen for example)
- CI isn't passing, I think the late deletion of the para block has broken some of the flags
- The `turbo.json` files will need updating if we introduce `codegen` scripts rather than `schema` ones (as I think we should) like we have done in `ai-chat`, etc.

apps/hash-frontend/src/components/sandbox/framed-block/framed-block.tsx
CiaranMn: I _almost_ deleted these files becasue this sandbox isn't used, but it started to unpick some cross-frame considerations that probably are worth keeping in place for when we re-introduce it. I wouldn't spend any effort on updating the logic though beyond getting it to pass linting.

blocks/heading/package.json
CiaranMn: I will be doing this in immediate follow up because it will involve rewriting and moving all the type files, which will make for a large diff.

Alfred-Mountfield: image

I didn't think you could get away with leaving it to another PR :(

CiaranMn: https://github.com/hashintel/hash/pull/2417/commits/674b9ed9e5cf6e3e78f6250b57f43132ba74c068

blocks/address/turbo.json
```diff
@@ -1,24 +1,24 @@
...
     },
     "build": {
-      "dependsOn": ["^build", "schema", "format"],
+      "dependsOn": ["^build", "codegen", "format"],
       "outputs": ["./dist/**"]
     },
     "lint:tsc": {
-      "dependsOn": ["schema"]
+      "dependsOn": ["codegen"]
     },
```
Alfred-Mountfield, line 14: Not asking for a change, but we don't need this anymore as it's defined in the root. (Technically the root also says `"^build"` which might be important if we ever change how the eslint or tsconfig stuff is 'built'

Alfred-Mountfield (dismissed):

blocks/how-to/src/app.tsx
```diff
@@ -95,8 +100,8 @@ export const App: BlockComponent<RootEntity> = ({
         )
         .sort(
           (a, b) =>
-            (a.linkEntity.linkData?.leftToRightOrder ?? 0) -
-            (b.linkEntity.linkData?.leftToRightOrder ?? 0),
+            (a.linkEntity.linkData.leftToRightOrder ?? 0) -
```
Alfred-Mountfield, line 103: 🎉

Alfred-Mountfield (approved):
LGTM 🪨 🚀

blocks/address/turbo.json
CiaranMn: Did wonder about these. Will leave for now and we can amend when we next touch these

PR #2421, opened 2023-04-14
Only commit OpenAPI specs instead of graph client

TimDiekmann:

We are able to generate the OpenAPI specs offline without having the `hash-graph` running since #2373. As we also have the Turbo dependency graph in place, we can rely on Turbo to generate the hash-graph-client on demand.

## 🔍 What does this change?

- Ignore the hash-graph-client code
- Unignore the OpenAPI specs
- Merge the hash-graph-client-generator as codegen step into `hash-graph-client`
- Setup the correct dependencies

This PR:


libs/@local/hash-graph-client/turbo.json
```diff
@@ -1,10 +1,20 @@
 {
   "extends": ["//"],
   "pipeline": {
-    // @todo - this introduces a dependency on building the graph in docker to `yarn install` this causes problems in Vercel deployment builds
-    //    "build": {
-    //      "dependsOn": ["@apps/hash-graph#codegen:generate-openapi-client"],
-    //      "outputs": []
-    //    }
+    "build": {
```
Alfred-Mountfield, line 4: You need `"dependsOn": ["codegen"]` no?

libs/@local/hash-graph-client/package.json
```diff
@@ -17,9 +16,11 @@
     "@apps/hash-graph": "0.0.0-private",
     "@local/eslint-config": "0.0.0-private",
     "@local/tsconfig": "0.0.0-private",
+    "@openapitools/openapi-generator-cli": "2.5.2",
```
thehabbos007, line 19: We do have this as a dev-dependency, would it be required as a dependency when doing a "production only" yarn install in dockerfiles? I think it would fail in those cases (i.e. for the hash-api dockerfile)

thehabbos007 (commented):
Thanks! Runs well locally and lints pass after codegen. I have a couple comments

libs/@local/hash-graph-client/package.json
```diff
@@ -17,9 +16,11 @@
...
     "@types/node": "18.11.9",
     "@typescript-eslint/parser": "5.49.0",
     "eslint": "8.33.0",
+    "httpyac": "5.8.2",
```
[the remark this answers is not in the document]
TimDiekmann: Absolutely! I made it a `hash-graph` dev-dependency instead

libs/@local/hash-graph-client/package.json
TimDiekmann: Good point, I made it a normal dependency.

thehabbos007 (dismissed):
LGTM :rock:

nonparibus (dismissed):

nonparibus (approved):

PR #2423, opened 2023-04-15
`deer`:  improve how integrals are handled

indietyp:

This is the first in a (probably) three-step PR chain.

This PR resolves one of the long overdue TODOs in the code and makes the deserializer of all numbers less punishing. The end goal is to remove the `usize` and `isize` types in our data scheme.

This means that everything is more in line with `serde` and that a `u32` can be deserialized using `U32Deserializer` into `u64`.

Note: We determine the size of `*size` using `target_pointer_width`, `serde` internally handles all `usize` as `u64`. We correctly determine the size of the target and set conversion functions appropriately this means that conversions would truncate w/ `serde` on 16-bit or 32-bit systems, while 🦌 correctly errors out.

This PR:

### DIFF
--- a/libs/deer/src/impls/core/integral.rs
+++ b/libs/deer/src/impls/core/integral.rs
@@ -1,65 +1,246 @@
-use error_stack::ResultExt;
+use error_stack::{IntoReport, Report, Result, ResultExt};
+use num_traits::ToPrimitive;
 
 use crate::{
-    error::{DeserializeError, VisitorError},
-    Deserialize, Deserializer, Document, Reflection, Schema, Visitor,
+    error::{DeserializeError, ExpectedType, ReceivedValue, ValueError, Variant, VisitorError},
+    Deserialize, Deserializer, Document, Number, Reflection, Schema, Visitor,
 };
 
-macro_rules! impl_integral {
-    (@num $name:ident, $visit:ident, $deser:ident, $typ:ty) => {
-        struct $name;
-
-        impl<'de> Visitor<'de> for $name {
-            type Value = $typ;
-
-            fn expecting(&self) -> Document {
-                <$typ>::reflection()
-            }
-
-            fn $visit(self, v: $typ) -> error_stack::Result<Self::Value, VisitorError> {
-                Ok(v)
-            }
-        }
-
-        impl Reflection for $typ {
+macro_rules! impl_num {
+    (
+        $primitive:ident:: $deserialize:ident; $($method:ident !($($val:ident:: $visit:ident),*);)*
+    ) => {
+        impl Reflection for $primitive {
             fn schema(_: &mut Document) -> Schema {
                 Schema::new("integer")
-                    .with("minimum", <$typ>::MIN)
-                    .with("maximum", <$typ>::MAX)
+                    .with("minimum", Self::MIN)
+                    .with("maximum", Self::MAX)
             }
         }
 
-        impl<'de> Deserialize<'de> for $typ {
+        impl<'de> Deserialize<'de> for $primitive {
             type Reflection = Self;
 
-            fn deserialize<D: Deserializer<'de>>(
-                de: D,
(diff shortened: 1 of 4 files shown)


libs/deer/src/impls/core/integral.rs
```diff
@@ -1,65 +1,246 @@
...
+    num_self!(u32::visit_u32);
+    num_from!(u8::visit_u8, u16::visit_u16);
+    num_try_from!(i8::visit_i8, i16::visit_i16, i32::visit_i32, i64::visit_i64, i128::visit_i128, u64::visit_u64, u128::visit_u128);
+    num_number!(u32::to_u32);
+);
+
+impl_num!(
+    u64::deserialize_u64;
+    num_self!(u64::visit_u64);
+    num_from!(u16::visit_u16, u32::visit_u32);
```
thehabbos007, line 114: Should this have `u8::visit_u8` in `num_from`?

libs/deer/src/impls/core/integral.rs
```diff
@@ -1,65 +1,246 @@
...
+    num_try_from!(i8::visit_i8, i16::visit_i16, i32::visit_i32, i64::visit_i64, i128::visit_i128, u128::visit_u128);
+    num_number!(u64::to_u64);
+);
+
+impl_num!(
+    u128::deserialize_u128;
+    num_self!(u128::visit_u128);
+    num_from!(u16::visit_u16, u32::visit_u32, u64::visit_u64);
```
thehabbos007, line 122: as above, should this also have u8 here?

libs/deer/tests/test_impls_core_integral.rs
```diff
@@ -1,76 +1,154 @@
...
+
+proptest_integral!(
+    Token::Number(u32);
+    proptest_fit!(u8 :: U8Deserializer, u16 :: U16Deserializer);
+    proptest_try_fit!(i8 :: I8Deserializer, i16 :: I16Deserializer, i32 :: I32Deserializer, i64 :: I64Deserializer, i128 :: I128Deserializer, u64 :: U64Deserializer, u128 :: U128Deserializer);
+);
+
+proptest_integral!(
+    Token::Number(u64);
+    proptest_fit!(u8 :: U8Deserializer, u16 :: U16Deserializer, u32 :: U32Deserializer);
```
thehabbos007, line 113: Hmm, i see we are proptesting u8 here even through we didn't specify them in `num_from` for u64. Is the u8 impl defaulted somehow?

thehabbos007 (commented):
Thank you! I have a couple comments, but looks great!

libs/deer/src/impls/core/integral.rs
indietyp: You are completely right, don't know I missed that.

libs/deer/tests/test_impls_core_integral.rs
indietyp: huh, I mean it shouldn't so that is very interesting

indietyp: Oooo, this works because `visit_u8` default implementation defers to `visit_number`, which is implemented, but I wouldn't say that this was my intention

indietyp: fixed in [fc2d6bb](https://github.com/hashintel/hash/pull/2423/commits/fc2d6bb6adfc3c69e982b6723b3f3927c3cb0bff)

libs/deer/src/impls/core/integral.rs
indietyp: fixed in [fc2d6bb](https://github.com/hashintel/hash/pull/2423/commits/fc2d6bb6adfc3c69e982b6723b3f3927c3cb0bff)

thehabbos007 (dismissed):

PR #2424, opened 2023-04-15
`deer`: remove `isize` and `usize` from data-model

indietyp:

This PR removes `deserialize_isize` and `deserialize_usize` from the `Deserializer` and `visit_usize`, `visit_isize` from the `Visitor` trait.

With this change we basically move the burden of handling of `usize` from `Deserialize` and `Deserializer` into `deer` itself, this overall reduces complexity.

The reason for this change are twofold:
A) (by far the biggest) a `*size` is just another primitive number in disguise, depending on the pointer width of the target architecture. This means we can use `u64` on 64-bit machines, etc. to achieve the same. Implementations for `*size` still exist. They only do not exist in our `Deserializer` and `Visitor` implementation as they introduce ambiguity and are redundant.
B) serde does the same, except they always say `usize = u64`.

This PR:


TimDiekmann (approved):

thehabbos007 (dismissed):

thehabbos007 (approved):
Sorry approved too quickly when I saw the base was on `main`. Meant to await the rebase :smile:

PR #2426, opened 2023-04-16
Run the graph `test-server` from external services

TimDiekmann:

To allow the usage of the test-server in TS it's required that it can be run as an external service.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204367340604860/f) _(internal)_

## 🚫 Blocked by

- #2419 
- 
## 🔍 What does this change?

- Adjust `Dockerfile` to take a new build-argument
- Adjust docker-compose files to enable the test server
- Add a healthcheck for the testserver

This PR:

## ⚠️ Known issues

The configuration of the graph image is very hard as we have a set of permutations:
- default run (dev-profile + type-fetcher)
- offline (dev-profile)
- test (dev-profile + type-fetcher + test-server)
- production (dev-profile + type-fetcher)

This PR simplifies this and enables the test-server for all dev-profiles. The better option would be to decide first, how the graph is built and use that image.

## 🐾 Next steps

A follow-up PR will take the above issue into account


apps/hash-graph/bin/cli/src/subcommand/test_server.rs
```diff
@@ -20,13 +27,23 @@ pub struct TestServerArgs {
 
     /// The address the REST client is listening at.
```
thehabbos007, line 28: Perhaps there is another instance of this comment for the main graph server?
```suggestion
    /// The address the REST server is listening at.
```

thehabbos007 (dismissed):

thehabbos007 (approved):

PR #2433, opened 2023-04-17
Fix broken subpage show/hide button on sidebar.

yusufkinatas:

This PR fixes not working show/hide subpages button on sidebar.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203857507057423/f) _(internal)_

## 🔍 What does this change?

- Updates `rooks` version to fix this issue 
- Adds test for testing this functionality

`useLocalstorageState` function from `rooks` was not actually supporting calling the setState function like `setState(current => new)`. While debugging, I realized that setState runs twice. And because the specific setState here is the “Toggle dropdown” one, it was toggling it twice. Which means nothing changes.

`rooks` fixed the issue here: https://github.com/imbhargav5/rooks/releases/tag/rooks%407.6.1, so just updating the rooks version fixed the issue

This PR:

## 🛡 What tests cover this?

- `page-navigation.spec.ts`

## ❓ How to test this?

1.  Try expand/collapse sub-pages on sidebar.

## 📹 Demo

https://user-images.githubusercontent.com/39553853/232525258-58ead331-743a-4ddd-ac16-8fb927d6f722.mov

### DIFF
--- a/apps/hash-frontend/package.json
+++ b/apps/hash-frontend/package.json
@@ -79,7 +79,7 @@
     "react-modal-hook": "3.0.0",
     "react-responsive-carousel": "3.2.23",
     "react-transition-group": "4.4.5",
-    "rooks": "7.4.3",
+    "rooks": "7.11.1",
     "setimmediate": "1.0.5",
     "ts-node": "10.9.1",
     "url-regex-safe": "2.0.2",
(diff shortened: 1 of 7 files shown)


yusufkinatas:
@CiaranMn Not sure which checkbox to check for this one. It's just updating version of `rooks` library without any major changes. There is probably no benefit of publishing block versions just for updating `rooks` version. Should we just ignore those?

CiaranMn:
> @CiaranMn Not sure which checkbox to check for this one. It's just updating version of rooks library without any major changes. There is probably no benefit of publishing block versions just for updating rooks version. Should we just ignore those?

What are they using from rooks? Is there any chance they are affected by the same issue? same question for `type-editor`

If you think it has no functional effect it can be this

yusufkinatas:
All uses it for different purposes, and none benefits from the fix here. So checked that one.

luisbettencourt (approved):

PR #2435, opened 2023-04-17
`deer`: rename `v`, `de` to more meaningful names.

indietyp:

This PR is simply cleanup and adds no extra functionality, it renames `v` to `value` in the `Visitor` trait (and it's implementations), or to `object`/`array` for `visit_object` and `visit_array` respectively. It additionally changes `de` to `deserializer`.

Because IDEs often use the default parameter name in the trait as the default name changing the name makes the intent overall clearer.

As a drive-by it renamed the generic of `visit_object` and `visit_array` from `T` (which is very often used for generics in the struct) to `A` (for `A`ccess). This will avoid clashes in auto completion of IDEs and makes everything overall less confusing.

This PR:

> Looks like this is my 100th PR for HASH. Thank you so much for the continued support and just... everything. Thank you. You guys are awesome 💚 . Until the next milestone 😉


libs/deer/src/lib.rs
```diff
@@ -195,152 +195,153 @@ pub trait Visitor<'de>: Sized {
...
+    fn visit_string(self, value: String) -> Result<Self::Value, VisitorError> {
+        self.visit_str(&value)
     }
 
-    fn visit_bytes(self, v: &[u8]) -> Result<Self::Value, VisitorError> {
+    fn visit_bytes(self, value: &[u8]) -> Result<Self::Value, VisitorError> {
         Err(Report::new(TypeError.into_error())
             .attach(ReceivedType::new(visitor::BinarySchema::document()))
             .attach(ExpectedType::new(self.expecting()))
             .change_context(VisitorError))
```
thehabbos007, line 239: is `value` param unused? can't find it's usage in the fn body

libs/deer/src/lib.rs
```diff
@@ -195,152 +195,153 @@ pub trait Visitor<'de>: Sized {
...
     }
 
-    fn visit_bytes_buffer(self, v: Vec<u8>) -> Result<Self::Value, VisitorError> {
-        self.visit_bytes(&v)
+    fn visit_bytes_buffer(self, value: Vec<u8>) -> Result<Self::Value, VisitorError> {
+        self.visit_bytes(&value)
     }
 
-    fn visit_array<T>(self, v: T) -> Result<Self::Value, VisitorError>
+    fn visit_array<A>(self, array: A) -> Result<Self::Value, VisitorError>
```
thehabbos007, line 250: small nit: Just wondering about the consistency between the names here. I see most cases use `value` but array and object are `array` and `object` respectively. Any reason for this distinction?

thehabbos007 (commented):
Looks good to me, just have a question for the names

libs/deer/src/lib.rs
indietyp: That is correct. This is also why `#[allow(unused_variables)]` is used for this trait. The names are what IDEs suggest for the parameters when filling in the methods, so choosing a name allows us to make auto-complete better, even though we're not using the variable in the default implementation.

libs/deer/src/lib.rs
thehabbos007: Gotchu, that's actually an nice distinction, then

thehabbos007 (approved):
`#100` :tada: :rocket: :rock: :taco: 
Great stuff :)

thehabbos007 (approved):

PR #2434, opened 2023-04-17
GEN-34: `deer`: implement `Content` IR

indietyp:
## 🌟 What is the purpose of this PR?

This implements `Content` and related `Deserializer`s, which is one of the last steps needed before we can start on the derive macro.

## 🚫 Blocked by

* #2418 

## 🚀 Has this modified a publishable library?

This PR:

- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
- [x] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
- [ ] modifies a **block** that will need publishing via GitHub action once merged
- [ ] does not modify any publishable blocks or libraries, or modifications do not need publishing
- [ ] I am unsure / need advice

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204596705932088

PR #2437, opened 2023-04-17
`deer`: introduce `StructVisitor`

indietyp:

Implements `StructVisitor`, one of the few remaining key pieces, this allows non self-describing formats (like `bincode` or `msgpack`) to use arrays instead of objects to denote structs.

Important: our `json` implementation does not support `visit_array`, this is on purpose, we do not want self-describing formats to do the same as non self-describing formats. They should always prefer the object variant.

~~The visitor has 4 methods (instead of 2): `visit_none`, `visit_null`, `visit_object`, `visit_array`. `visit_none` and `visit_null` are of interest, as they allow us easily implement `Default` behaviour.~~

~~> TODO: ^ the above statement should not be true, combining the `OptionalVisitor` with the `StructVisitor` should be enough!~~

## 🔗 Related links

- unblocks #2422 

This PR:

## 🐾 Next steps

* `IdentifierVisitor` (allows for non-string identifiers for different formats)
* `ReceivedVariant` and `ExpectedVariant` support for non-str types
* `KeyError`, `MissingError` additions
* `Content` (#2434)
* `core::ops` (#2422 - unblocked by this PR)
* switch to `justjson` for json
* json tests (blocked by item above)
* alloc implementation
* derive macro (can happen simultaneously to other tasks once `IdentifierVisitor`, `MissingError`, `KeyError`, and `Content` are in)

### Experiments (slated for 0.2)

* blueprint (in local repo)
* switch from serde to value-bag (in local repo)


libs/deer/desert/src/token.rs
```diff
@@ -206,3 +206,43 @@ impl Display for Token {
...
+            Self::Object { .. } | Self::ObjectEnd => Document::new::<AnyObject>(),
+            Self::Null => Document::new::<<() as Deserialize>::Reflection>(),
+        }
+    }
+}
+
+// TODO: maybe number
+// TODO: IdentifierVisitor (u8, u64, str, borrowed_str, string,
+//  bytes, bytes_buf, borrowed_bytes)
+// TODO: test
```
thehabbos007, line 248: Do we want these todos to be more explicit? :)

thehabbos007 (commented):
Looks pretty comprehensive so far. Have a couple comments, but nothing stood out other than that. Thank you!

libs/deer/desert/src/token.rs
indietyp: that might be a good idea, I have those mirrored into my own personal todo list (/ project management system), but because I do not have access to Asana I cannot create those tasks there and I didn't want to pollute the GitHub issues for this sort of thing as they are mostly bug reports right now (tho I heard whispers that that might change in the future)

thehabbos007: I'm okay with this for now as a lot of this is in flux still :+1:

thehabbos007 (approved):

thehabbos007 (approved):

PR #2441, opened 2023-04-18
Correct Heading block name

CiaranMn:

We use the `name` field in `package.json["blockprotocol"]` to determine which API key to use when publishing blocks.

The `heading` block was missing the namespace prefix, which meant that automatic publishing was failing (see [here](https://github.com/hashintel/hash/actions/runs/4731993373/jobs/8397662623#step:5:15)).

This PR:

## ⚠️ Known issues

The `chart` block publishing is failing because it is published under a different namespace, for which we haven't entered an API key, but since it's outdated this is less important to address.

## 🐾 Next steps

I will re-run publishing for the `heading` block.

## ❓ How to test this?

1.  Look at a few `package.json`s for other blocks and see that this PR makes them consistent.

### DIFF
--- a/blocks/heading/package.json
+++ b/blocks/heading/package.json
@@ -51,7 +51,7 @@
     "blockType": {
       "entryPoint": "react"
     },
-    "name": "heading",
+    "name": "@hash/heading",
     "displayName": "Heading",
     "icon": "public/h1.svg",
     "image": "public/preview.svg",


nonparibus:
> The chart block publishing is failing because it is published under a different namespace, for which we haven't entered an API key, but since it's outdated this is less important to address.

@CiaranMn @benwerner01 We will want to move the chart block into the HASH namespace anyway as and when we update it for `0.3` -- whether that is in simplified (accepts CSV/TSV) form, or fully entity-linked and aggregate-rendering shape.

nonparibus (approved):

PR #2447, opened 2023-04-19
`deer`: rework how bounded access works

indietyp:

This completely reworks how bounded access works, by conceptually replacing `set_bounded` with `into_bounded`.

The reason for this is twofold:
* greater separation of concerns
* move implementation, which can be error prone, to the core, make it unified

With this move, only a single implementation of bounded and `Deserializer` specific implementation kinks or problems, like accidental contract violations, are no longer possible. This makes the whole behavior in the long term more predictable.

A completely typesafe version was prototyped in this PR but removed due to the fact that it relies on const expressions (e.g. `ConstObjectAccess<{N + 1}>`), which is not yet stable

This PR:


indietyp:
Disclaimer, diff is large because 538 lines are tests 😅

Issue #2302, opened 2023-03-28
[H-2] Make headings in hash.dev blog posts anchor-linkable


nonparibus:
We want the headings on HDEV blog posts (such as [https://hash.dev/blog/block-protocol-0-3-wordpress](https://hash.dev/blog/block-protocol-0-3-wordpress)) to be individually linkable.

We should render a 'link' icon next to them on-hover, similar to what we already do on [https://blockprotocol.org/docs](https://blockprotocol.org/docs), allowing a direct link to a heading/section to be copied to a user's clipboard. They should also function similarly when clicked, jumping a user down to that section.

<sub>From [SyncLinear.com](https://synclinear.com) | [H-2](https://linear.app/hash/issue/H-2/make-headings-in-hashdev-blog-posts-anchor-linkable)</sub>

PR #508, opened 2022-04-25
Final block design draft

MaggieAppleton:
## 🌟 What is the purpose of this PR?

Submit the final draft of the design guide for blocks

## 🚫 Blocked by

- [x] Images are still broken. I'm rendering them inline in `` tags as each images needs its own size declaration. But for inline images to work, I've found I have to also include the same image in a `` markdown format somewhere else in the post. For the moment I've put them all at the bottom of the post. These need to be removed once the image issue is fixed.
- [x] Some of the images render inconsistently - their sizes change on each server restart. It's unclear if this is related to the above issue.
- [x] Need to add alt text to images
- [x] Requires a cover image
- [x] Remove test blog post and assets in accompanying folder

PR #2448, opened 2023-04-19
[H-2] Make headings in hash.dev blog posts anchor-linkable

gitstart-blockprotocol:
## What does this PR do?
It provides anchor-linkable in the headings of the blog posts and clicking on it the heading link will be copied..

### Issue References
* Client: https://github.com/hashintel/hash/issues/2302

### Loom/Screenshot Demo
https://www.loom.com/share/7df6c053fcbc4c4abb7c524d5e5fb8ad

---
This code was written and reviewed by GitStart Community. Growing future engineers, one PR at a time.


nonparibus (dismissed):
Thanks for opening this PR! A few suggested refinements, having seen this live in action:

1. **We want to avoid the link icon ever appearing on a white line by itself.** To avoid this, we should ensure that the max-width for the final line of an `h[x]` header is slightly less than that of the container it is inside, ensuring there will always be enough room to display this link icon next to the final word on the final line, without breaking to a new one. You can see the bug currently occurring here:
Capture 2023-04-19 at 16 09 26@2x

2. Onhover over the 'link' icon, we'd like the words 'Copy link to this section' to appear in a tooltip to the right (or if no space, left) of the icon. This will make it is explicitly clear to a user that something is about to be copied to their clipboard, preventing the potentially unexpected overwriting of their current clipboard contents. Thanks for this PR's associated Loom video, which helped us in realizing the need for this.

Potential mockup attached based on tooltips we already use in this repo. @hashintel/frontend please can you advise where this component can be found?

Capture 2023-04-19 at 16 16 46@2x

TimDiekmann:
I updated the branch to make CI pass again (#2457)

nonparibus:
Fantastic. From a UX perspective this looks really well done. I'm tagging in @hashintel/frontend who'll provide code review and the approval necessary to merge this in.

(And thanks @TimDiekmann for updating the branch to fix CI)

apps/hashdotdev/src/components/mdx-page-content.tsx
```diff
@@ -1,16 +1,132 @@
-// @todo update from blockprotocol
+// import throttle from "lodash/throttle";
```
CiaranMn, line 1: What's this for?

apps/hashdotdev/src/components/mdx-page-content.tsx
```diff
@@ -1,16 +1,132 @@
...
+        // sessionStorage is not available
+      }
+
+      const shouldScrollToAnchor =
+        // if anchor is empty and we haven't scrolled, prevent it
+        (anchor === "" && scrolledOnce.current) ||
+        // if anchor is not empty, always allow scroll
+        anchor !== "" ||
+        // OR if previous path is either a docs or spec page
+        (previousRoute?.includes("/docs") && router.asPath.includes("/docs"));
```
CiaranMn, line 60: These routes do not exist in this site – worth a check when copying in code from elsewhere to see if it all remains relevant.

apps/hashdotdev/src/components/mdx-page-content.tsx
```diff
@@ -1,16 +1,132 @@
...
 export const MdxPageContent: FunctionComponent<MdxPageContentProps> = ({
   serializedPage,
 }) => {
-  // @ts-expect-error @todo fix this
-  return <MDXRemote {...serializedPage} components={mdxComponents} />;
+  const router = useRouter();
+
+  const [headings, setHeadings] = useState<Heading[]>([]);
+
+  const detectHeadingFromScroll = useRef<boolean>(true);
```
CiaranMn, line 22: What is the purpose of this ref? It gets set to `false` and `true`, including by using the timer above, but the value is never read for anything.

CiaranMn (changes requested):
Thank you for this – functionality looks good, but there is some code that doesn't seem to be doing anything.

apps/hashdotdev/src/components/mdx-page-content.tsx
```diff
@@ -1,8 +1,9 @@
-// @todo update from blockprotocol
```
nonparibus, line 1: @CiaranMn @nathggns Any idea what this comment was originally doing here, and whether we should persist it? I don't know if it's appropriate that we remove it (which implies it could've done with more detail).

nathggns: It comes all the way from my PR (recreated/merged by Maggie) which introduced the blog – https://github.com/hashintel/hash/pull/508/files#diff-0e23b4ea81c2ad7fa5a58d05355f073a4852616a21cd96f8fcbd69a25ab357bd 

I think what it was referring to was to update the MDX system to match the one used in blockprotocol, but I don't recall why. I don't think it's particularly worth keeping.

CiaranMn (approved):
Looks good to me, thanks very much!

PR #2429, opened 2023-04-16
Use caching for hash-graph docker image

TimDiekmann:
## 🌟 What is the purpose of this PR?

As mentioned in #2428, this uses caching for the `hash-graph` docker image, as this is currently one of the slowest part in CI.

This PR immensely speeds up CI times.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204367340604860/f) _(internal)_

## 🚫 Blocked by

- #2419 
- #2426 
- #2427 
- #2428

## 🔍 What does this change?

- Add a composite action to build the `hash-graph` image
- Add a composite action to load the image back into Docker
- Add a configuration to `setup.py` to filter by `hash-graph`
- Use the composite action to build the image in the `CI` and `Rust` workflows
- Use the cached artifact to speed up external services

## 🚀 Has this modified a publishable library?

This PR:

- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
- [ ] modifies a **block** that will need publishing via GitHub action once merged
- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing
- [ ] I am unsure / need advice

PR #2427, opened 2023-04-16
Use `turbo` in `Dockerfile` to prune unneeded content

TimDiekmann:
## 🌟 What is the purpose of this PR?

It's possible to better utilize the docker-layer-caching by using `turbo prune`. This command will remove unneeded clutter and write it into `out/`. Based on this the dependencies can be cached if the package.json, yarn.lock, and turbo.json did not change.

## 🔗 Related links

- [Turbo documentation: Deploying with Docker](https://turbo.build/repo/docs/handbook/deploying-with-docker)

## 🚫 Blocked by

- #2419 
- #2426

## 🚀 Has this modified a publishable library?

This PR:

- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
- [ ] modifies a **block** that will need publishing via GitHub action once merged
- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing
- [ ] I am unsure / need advice

PR #2453, opened 2023-04-19
Support docker layer caching for `hash-agents`

TimDiekmann:

This extends on #2429 and supports caching for docker layers for `hash-agents` as well

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1204431665478216/f) _(internal)_

## 🚫 Blocked by

- #2427
- #2428
- #2429

## 🔍 What does this change?

- Extend composite actions to select docker images to be built
- Adjust docker file creations (and use `buildx` directly)


.github/actions/build-docker-images/action.yml
```diff
@@ -1,13 +1,23 @@
...
 description: "Build docker images"
+inputs:
+  hash-agents:
+    description: "Build hash-agents image"
+    required: false
+    default: "false"
+  hash-graph:
+    description: "Build hash-graph image"
+    required: false
+    default: "false"
```
thehabbos007, line 11: Would this input be better as an array of images we want to build? Or does that give terrible error messages compared to the completion we could get from these bool options?

apps/hash-agents/package.json
```diff
@@ -6,20 +6,20 @@
   "main": "./src/agents.ts",
   "scripts": {
     "build": "yarn exe ./scripts/codegen.ts 'app/agents/**/*.ts'",
-    "build:docker": "DOCKER_BUILDKIT=1 docker build --tag hash-agents --file docker/Dockerfile ../../",
+    "build:docker": "docker buildx build --tag hash-agents --file docker/Dockerfile ../../",
```
thehabbos007, line 9: Is this included by default or do people need to install an extension?

.github/actions/build-docker-images/action.yml
TimDiekmann: An array with enum elements would be nice but I think this is not possible. With array of strings it’s easy to use the wrong string (e.g. `graph` or `hash_graph`)

apps/hash-agents/package.json
```diff
@@ -6,20 +6,20 @@
...
     "quicktype": "^16.0.43",
-    "ts-node": "10.9.1",
-    "typescript": "4.9.4",
     "yargs": "^17.7.1"
   },
   "devDependencies": {
     "@local/tsconfig": "0.0.0-private",
-    "@types/pascalcase": "1.0.1"
+    "@types/pascalcase": "1.0.1",
+    "ts-node": "10.9.1",
```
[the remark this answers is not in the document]
TimDiekmann: Yes, you're right. I copied that over from `hash-graph` but as we don't have `build-dependencies` in `yarn`, this should be a `dev-dependency` as well there.

apps/hash-agents/package.json
thehabbos007: https://docs.docker.com/build/install-buildx/ it seems to me that it may not be installed automatically for some folks. Should we take this into consideration? Perhaps adding the requirement to the readme next to the docker/docker compose versions

thehabbos007 (approved):

PR #2455, opened 2023-04-19
Avoid multiple CI runs for same job in same PR

TimDiekmann:

Similar to the Rust CI, this cancels CI runs on previous commits in the same PR

This PR:

- [X] does not modify any publishable blocks or libraries, or modifications do not need publishing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,10 @@ env:
   TURBO_TEAM: hashintel
   TURBO_REMOTE_ONLY: true
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   linting:
     name: Linting


.github/workflows/ci.yml
```diff
@@ -12,6 +12,10 @@ env:
   TURBO_TEAM: hashintel
   TURBO_REMOTE_ONLY: true
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
```
nonparibus, line 16: This doesn't mess with the merge queue in some unexpected way, does it?

TimDiekmann: We have this setup in the Rust CI and in the BP repository and I haven't noticed any issues with the merge queue. I don't think it should mess up the queue, as the queue has a branch per test.

CiaranMn (approved):

PR #2456, opened 2023-04-20
Log stacktrace when an exception is raised

thehabbos007:
This PR logs the exception info (stacktrace) when the agent runner runs into an issue


apps/hash-agents/app/__init__.py
```diff
@@ -37,7 +37,7 @@ def agent(agent_name):
         try:
             return call_agent(agent_name, **request.json)
         except Exception as e:
-            getLogger(__name__).error(e)
+            getLogger(__name__).critical(e, exc_info=True)
```
TimDiekmann, line 40: I would probably keep it as an `error` as this is not critical for the service itself, the service will still work.
```suggestion
            getLogger(__name__).error(e, exc_info=True)
```

TimDiekmann (approved):

PR #2457, opened 2023-04-20
Fix Rust workflow may cancel CI workflow

TimDiekmann:

The Rust workflow may cancel the CI workflow with the recent changes. This will fix it

## 🔍 What does this change?

Change the concurrency group:
- https://stackoverflow.com/a/72408109
The important change is to add `github.workflow`. Replacing `head_ref` with the PR number, however, makes it easier to read.

This PR:


thehabbos007 (approved):
This does seem more robust than using `head_ref`

PR #2461, opened 2023-04-20
Move Realtime to Redis Streams and adjust its DB schemas

thehabbos007:

This PR modifies `hash-realtime` to be up to date with `hash-graph`. After we changed our DB layouts, we broke the existing realtime implementation. I've adjusted the setup to use [Redis Streams](https://redis.io/docs/data-types/streams/) and changed the docker compose to use redis-stack (for a graphical Redis interface).

The adjustments are primarily to make sure we're allowing proper pub/sub in Redis and remove as much unnecessary locking as possible. We still have a technically highly-available realtime service that can support automatic failover, but this hasn't fully been utilized/tested yet - we may want to look at consumer groups for Redis Streams in the future.

This work is expected to help external services integrate with realtime data changes from HASH. @indietyp you may be interested in this as well, you can look at the types in `libs/@local/hash-backend-utils/src/pg-tables.ts` to see if you find them useful.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1202225566851516/f) _(internal)_ (makes this task redundant)
- [Asana task](https://app.asana.com/0/1204260760205594/1204398654562771/f) _(internal)_ (doesn't complete this)

## 🔍 What does this change?

- Revives realtime using new DB layout (we can't keep the schema in sync automatically)
- Adjusts infra with `turbo prune`
- Clean up environment variables for the services (If you have a `.env.local` please diff and check what has changed!)
- Add realtime to the prod docker-compose

This PR:

Nothing outward-facing has changed, existing docs have been updated

## 🐾 Next steps

- Indexing the realtime updates through another loader-service

## ❓ How to test this?

1.  Start the app (`yarn external-services up` and `yarn dev` in root)
1.  Go to http://localhost:8011/ and observe new update streams in Redis (should be a lot from starting the app)
1.  Sign in to the app and create a type and entities, observe these changes being reflected in their respective streams

## 📹 Demo

https://user-images.githubusercontent.com/6988668/233420579-fe29fcc6-49ef-4609-bcee-cb6063c71891.mp4


apps/hash-external-services/docker-compose.yml
```diff
@@ -35,6 +35,23 @@
     # volumes:
     #   - ../../var/hash-external-service/opensearch/data:/usr/share/opensearch/data
 
+  hash-redis:
```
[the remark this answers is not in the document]
thehabbos007: Fixed in 318748623313ae9b9443a58c78282975ad4bb192

infra/docker/realtime/prod/Dockerfile
```diff
@@ -1,36 +1,38 @@
...
 
-FROM node:16.18.1-alpine
+COPY --from=base /app/out/json/ .
+COPY --from=base /app/out/yarn.lock ./yarn.lock
+COPY --from=base /app/out/full/turbo.json turbo.json
 
-# Run as a non-root user
-RUN addgroup -S appgroup && adduser -S appuser -G appgroup
-COPY --from=builder --chown=appgroup:appuser /app /app
+RUN yarn install --frozen-lockfile --prefer-offline \
```
TimDiekmann, line 18: Please note, that this won't run `codegen` anymore. You may want to consider moving the `COPY` command of `turbo.json` below this and run `RUN yarn turbo codegen --filter '@apps/hash-realtime'`

TimDiekmann (commented):
I looked at the docker-related things and they're looking good. Only one thing I notice which may break in the future.

infra/docker/realtime/prod/Dockerfile
thehabbos007: Oh yes, thank you. Even though I think this doesn't need codegen, I will still add it :+1:

TimDiekmann: Maybe it's even sensible to call `build` here, but I'm unsure about that.

thehabbos007: Yeah I think that's better, should have codegen as a dep I suppose

TimDiekmann: yes, that dependency is global unless overwritten 🙂

thehabbos007: `turbo.json` seems to be required for the `yarn install` for now (fixed in #2463 perhaps?) do we want to wait for that to merge, or would the requirement be there regardless?

TimDiekmann: Let's wait for #2463, this will remove the dependency. 🙂

nonparibus:
[#2463](https://github.com/hashintel/hash/pull/2463) has now merged.

libs/@local/hash-backend-utils/src/realtime.ts
```diff
@@ -1,22 +1,43 @@
...
+  entityTables: ["entity_editions", "entity_temporal_metadata"],
+  entityTypeTables: ["entity_types"],
+  propertyTypeTables: ["property_types"],
+} as const;
+
+export const supportedRealtimeTables: SupportedRealtimeTable[] =
+  Object.values(supportedTables).flat();
+
+export type SupportedRealtimeTable =
+  (typeof supportedTables)[keyof typeof supportedTables][number];
```
Alfred-Mountfield, line 15: This seems to be `any` for me

Alfred-Mountfield: This is a bit ridiculous but the only way I found to solve it was like this:

```suggestion
type ArrayElementType<T> = T extends readonly (infer U)[] ? U : never;

export type SupportedRealtimeTable = ArrayElementType<
  (typeof supportedTables)[keyof typeof supportedTables]
>;
```

And unfortunately the weird `readonly` bit is needed as well because 
  
```ts
type ArrayElementType<T> = T extends (infer U)[] ? U : never;

export type SupportedRealtimeTable = ArrayElementType<
  (typeof supportedTables)[keyof typeof supportedTables]
>;
```

results in `never` due to the readonly tuple types.

You could perhaps move `ArrayElementType` alongside our other helper types

Alfred-Mountfield: We could also replace a lot of the `[number]` calls in the types in this file to be safe but I'm not sure if that's needed.

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -2,74 +2,106 @@
  * This module contains type definitions for tables in the Postgres database. Column
  * names are converted from snake_case to camelCase for consistency.
```
Alfred-Mountfield, line 3: This seems outdated

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -2,74 +2,106 @@
...
-     * @todo Update table definition when provenance info is added for updates
-     *   https://app.asana.com/0/1202805690238892/1202848989198291/f
-     */
-    public entityId: string,
-    public version: string,
-    public entityTypeVersionId: string,
-    public properties: Record<string, unknown>,
-    public createdBy: string,
-  ) {}
+export type EntityEdition = {
```
Alfred-Mountfield, line 8: Kinda want to insert :panic: emoji here.. How many times are we going to redefine these things and have separate sources of truth 😭

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -2,74 +2,106 @@
...
+  entity_edition_id: string;
+  properties: Record<string, unknown>;
+  left_to_right_order: number;
```
Alfred-Mountfield, line 11: Shouldn't this be optional? 🤔

Alfred-Mountfield: Where's the left entity ID and right entity ID?

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -2,74 +2,106 @@
...
-     *   https://app.asana.com/0/1202805690238892/1202848989198291/f
-     */
-    public entityId: string,
-    public version: string,
-    public entityTypeVersionId: string,
-    public properties: Record<string, unknown>,
-    public createdBy: string,
-  ) {}
+export type EntityEdition = {
+  entity_edition_id: string;
```
Alfred-Mountfield, line 9: Don't we want to reuse `EntityEditionId` here?

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -2,74 +2,106 @@
...
 
-  private static parseFromRow(row: Record<string, unknown>): Link {
-    return {
-      sourceEntityId: row.source_entity_id as string,
-      targetEntityId: row.target_entity_id as string,
-      linkTypeVersionId: row.link_type_version_id as string,
-      createdBy: row.created_by as string,
-    };
-  }
+export type Entity = EntityEdition & EntityTemporalMetadata;
```
Alfred-Mountfield, line 68: Do we perhaps want to name all of these types something else to avoid yet another competing `Entity` definition?

Alfred-Mountfield: (And why can we not just use the normal `Entity`?)

libs/@local/hash-backend-utils/src/queue/redis.ts
```diff
@@ -178,9 +171,9 @@ export class RedisQueueExclusiveConsumer implements QueueExclusiveConsumer {
     let item = await this.client.rpoplpush(processingName, processingName);
```
Alfred-Mountfield, line 171: Can't comment this far from the diff but there's a typo in the doc-comment above

 - timemoutMs === 0: blocks indefinitely until an item appears on the queue.

should be `timeoutMs`

libs/@local/hash-backend-utils/src/queue/redis.ts
```diff
@@ -178,9 +171,9 @@ export class RedisQueueExclusiveConsumer implements QueueExclusiveConsumer {
...
 
     // Otherwise, pop from the main queue and push onto the processing queue.
-    if (!item) {
+    if (item) {
```
Alfred-Mountfield, line 174: I don't quite understand why this has swapped. Why do we block if we already have an item?

libs/@local/hash-backend-utils/src/queue/redis.ts
Alfred-Mountfield, line 171: I appreciate you didn't write the comment above this in this PR, but what's the actual "check" referring to in
>  `Check if there's an item which wasn't processed correctly on the last call`

I don't see anything conditional

libs/@local/hash-backend-utils/src/stream/adapter.ts
```diff
@@ -0,0 +1,25 @@
+import { JsonObject } from "@blockprotocol/core/.";
+
+/** QueueProducer represents a resource which adds items to a queue. */
```
Alfred-Mountfield, line 3: `StreamProducer`?

libs/@local/hash-backend-utils/src/wal2json.ts
```diff
@@ -5,15 +5,15 @@
  * Note: We ignore transaction begin / end messages (i.e. those with action "B" / "C"),
  * and TRUNCATE messages (action "T").
  */
-export type Wal2JsonMsg = {
+export type Wal2JsonMsg<T = string> = {
```
Alfred-Mountfield, line 8: Should this be `T extends string`?

Alfred-Mountfield (commented):
Have some comments, excuse any ignorance, I've never looked at this code at all and it's been a bit of a hurdle to get into 😄

apps/hash-api/src/cache/redis.ts
```diff
@@ -18,14 +22,15 @@ export class RedisCache extends DataSource implements CacheAdapter {
...
   constructor(logger: Logger, cfg: RedisConfig) {
     super();
-    this.client = new AsyncRedisClient(logger, cfg);
-    this.get = this.client.get;
-    this.set = this.client.set;
-    this.setExpiring = this.client.setex;
-    this.rpush = this.client.rpush;
+    this.client = setupRedisClient(logger, cfg);
+    this.get = (key) => this.client.get(key);
+    this.set = (key, value) => this.client.set(key, value).then();
```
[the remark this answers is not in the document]
thehabbos007: Shouldn't be affected and we should still get a thrown error
image

apps/hash-external-services/turbo.json
```diff
@@ -6,6 +6,7 @@
     },
     "deploy": {
       "cache": false,
+      "persistent": true,
```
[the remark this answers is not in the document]
thehabbos007: Yeah, that is a fair point. I was thinking we shouldn't have any part of the Turbo graph depending on the system running like this. As you say it wouldn't be persistent with `--detach`, but unless we explicitly have a detached/attached task, I think we should assume this should be a leaf task

apps/hash-realtime/src/config.ts
```diff
@@ -1,38 +1,42 @@
...
+import { RedisStreamProducer } from "@local/hash-backend-utils/stream/redis";
 
 // The tables to monitor for changes
 export const MONITOR_TABLES = supportedRealtimeTables.map(
   (table) => `public.${table}`,
 );
 
+const redisConfig: RedisConfig = {
+  host: process.env.HASH_REDIS_HOST ?? "localhost",
+  port: parseInt(process.env.HASH_REDIS_PORT ?? "6379", 10),
```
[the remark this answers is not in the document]
thehabbos007: Yeah, it is needless. It does allow our defaults to be easier to discover, though. Perhaps we should have a `getRequiredEnv` equivalent called `getOptionalEnv` or something that can be defaulted and do the inner ternary there?

apps/hash-api/src/cache/redis.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Right sorry, `then` returns another `Promise` (I was thinking about how it generally takes a callback so it might continue executing)

With that in mind, what's the purpose of the `.then()` calls here?

apps/hash-external-services/turbo.json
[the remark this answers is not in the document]
Alfred-Mountfield: > I was thinking we shouldn't have any part of the Turbo graph depending on the system running like this.

Might be a bit hard with certain codegen utilities 😅

apps/hash-realtime/src/config.ts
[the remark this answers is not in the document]
Alfred-Mountfield: `getEnvWithDefault` or something 😅

apps/hash-realtime/src/index.ts
```diff
@@ -140,10 +153,62 @@ const releaseSlotOwnership = async (pool: PgPool, slotName: string) => {
   }
 };
 
-const pollChanges = async (pool: PgPool, slotName: string) => {
+type ChangeType = Wal2JsonMsg<SupportedRealtimeTable>;
+type EntityEditionId = string;
```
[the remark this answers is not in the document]
thehabbos007: Yes, we were using `string` before everywhere
Added in 8b7d01f10a366828c405319c05a380aa253978aa
Used in e855560dbaae3ce04d5cdd30f7646242ce44d4ef

apps/hash-realtime/src/index.ts
```diff
@@ -140,10 +153,62 @@ const releaseSlotOwnership = async (pool: PgPool, slotName: string) => {
...
+type EntityPollState = {
+  seenTemporalMetadata: Map<EntityEditionId, ChangeType>;
+  seenEntityEdition: Map<EntityEditionId, ChangeType>;
+};
+
+const findColumn = (columns: ChangeType["columns"], name: string): string =>
+  columns.find((col) => col.name === name)?.value as string;
```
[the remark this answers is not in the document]
thehabbos007: It is unsafe, I've changed the type away from `string` to a generic in e855560dbaae3ce04d5cdd30f7646242ce44d4ef

libs/@local/hash-backend-utils/src/realtime.ts
thehabbos007: Huh, how come you get `any` here?
image

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: Yes, 36bf50b179768275c97d5547d620d912ba8d02e9

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: Yeah, it's definitely not ideal. And as you've commented below, it's very easy to use the wrong types. Not sure how we can make this better without a codegen step from the SQL/DB to rust/ts/etc.

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: Yes they should. These types are only mapping the tables, and the table doesn't contain left and right entity ID. These are stored in `entity_has_left_entity` and `entity_has_right_entity` tables which i'm not joining currently

apps/hash-api/src/cache/redis.ts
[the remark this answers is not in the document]
thehabbos007: The class defines the operation to return `Promise<void>;` and the library returns `Promise<"OK">. It's to void the output

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: Hmm, I think that would be fine to do if I use the branded types now instead. Since these types map the PG tables, I'm not sure if we want to treat them as types given by `hash-subgraph`

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: Yeah, I'll prefix with `Pg` i think. We can't reuse `Entity` because we're not receiving the `Entity`, we're trying to reconstruct this from the DB change stream, so only a subset of the metadata is available

libs/@local/hash-backend-utils/src/queue/redis.ts
thehabbos007: This was not intentional. Thanks. I originally deleted the `queue` abstraction but decided to keep it for now. I don't really think we should keep this implementation around as it seems somewhat error-prone and based on application-level orchestration to work rather than leveraging the data structures provided by redis

libs/@local/hash-backend-utils/src/queue/redis.ts
thehabbos007: I gotta be honest, I have no clue what this means

libs/@local/hash-backend-utils/src/stream/adapter.ts
thehabbos007: Thanks, I thought I had changed them all :)
f824323ee87321a93f2dba9f4f0474c60fb356cb

libs/@local/hash-backend-utils/src/wal2json.ts
thehabbos007: Yes, thanks! 631acd495b532c5a0de962e09db9974d0fd1ca0e

libs/@local/hash-backend-utils/src/queue/redis.ts
thehabbos007: Thanks 31b2244fe3b18daf130d84f1c3e958304674b3ff

apps/hash-external-services/turbo.json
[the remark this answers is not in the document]
thehabbos007: OpenAPI? That isn't dependent on running the server anymore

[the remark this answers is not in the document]
Alfred-Mountfield: Yeah at present, just saying we might end up with a similar situation later on with some other REST service and client generator or whatever

apps/hash-realtime/src/index.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I think this is even less safe now? You're asserting away the `undefined`

libs/@local/hash-backend-utils/src/realtime.ts
Alfred-Mountfield: Hmmm, maybe it's just Jetbrain's type resolution 😠 

I'm not sure how we check what `tsc` resolves it as..

libs/@local/hash-backend-utils/src/pg-tables.ts
Alfred-Mountfield: Maybe I don't understand what this service actually does, but doesn't that mean we aren't correctly recreating entities if we're missing a good portion of their data?

apps/hash-api/src/cache/redis.ts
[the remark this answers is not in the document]
thehabbos007: yes 656e1a7a0cdf507ac99cf834eaeced3e6ba42436

apps/hash-realtime/src/config.ts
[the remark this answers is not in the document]
thehabbos007: After thinking about this, if we introduce `getEnvWithDefault` we'd still have a problem of parsing the string value as an int again since that wouldn't be done in `getEnvWithDefault` itself. Wouldn't change anything about the original comment :(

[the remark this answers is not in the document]
Alfred-Mountfield: You could pass a `convertType` function into `getEnvWithDefault` 😂...

definitely just theorycrafting amusing things here and we shouldn't change anything

libs/@local/hash-backend-utils/src/pg-tables.ts
thehabbos007: The full entity isn't recreated, no. We can do that, but that requires doing follow-up joins on the query, resulting in a lot of follow-up reads. 
I think the service should be considered a notification service that happens to provide some payload. I would be a little cautious to spend a lot of time making the change streams reflect every materialized result form the DB for now as it could blow up the scope a lot without needing the full view. And as we've pointed out, we're already duplicating the database tables, duplicating queries/join logic seems like something we shouldn't do (right now anyway)

libs/@local/hash-backend-utils/src/pg-tables.ts
```diff
@@ -1,75 +1,106 @@
...
+    record_created_by_id: obj.record_created_by_id as string,
+    archived: obj.archived as boolean,
+  };
+};
+
+export type PgEntityTemporalMetadata = {
+  owned_by_id: string;
+  entity_uuid: Record<string, unknown>;
+  entity_edition_id: string;
+  decision_time: string;
```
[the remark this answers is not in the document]
thehabbos007: We do need that to some degree. The consumers of the stream need to be aware of the decision time to know whether or not an actual newer version has been created. This type of notification service is really difficult because of our bitemporality, but we assume that the change stream is just a view into the table insertions, and services must find/use their own understanding of their entities to act accordingly

thehabbos007:
I changes the base branch as I don't like the permanence of merging this into main with all the duplicated DB mapping. Let's keep this scoped for the prototyping for now

PR #2460, opened 2023-04-20
Add `poetry.lock` file linting for `hash-agents`

TimDiekmann:
## 🌟 What is the purpose of this PR?

This adds the basic setup for linting for Python code. As a lint without adding any dependencies the lockfile is checked for validity.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204431665478215/f) _(internal)_

## 🚫 Blocked by

- #2458 

## 🚀 Has this modified a publishable library?

This PR:

- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
- [ ] modifies a **block** that will need publishing via GitHub action once merged
- [X] does not modify any publishable blocks or libraries, or modifications do not need publishing
- [ ] I am unsure / need advice

PR #2462, opened 2023-04-20
Add `black` as Python formatter

TimDiekmann:

This adds the `black` formatter for `hash-agents`' Python code

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204431665478215/f) _(internal)_

## 🚫 Blocked by

- #2458 
- #2460

## 🔍 What does this change?

- Add `black` as dev-dependency
- Add `fix:black` and `lint:black` as script
- Run `lint:black` from CI
- Add `fix:black` to lint-staged

This PR:

### DIFF
--- a/apps/hash-agents/pyproject.toml
+++ b/apps/hash-agents/pyproject.toml
@@ -23,8 +23,8 @@ production = ["gunicorn"]
 black = "^23.3.0"
 
 [tool.black]
-line-length = 100
 target-version = ['py311']
+preview = true
 
 [build-system]
 requires = ["poetry-core"]
--- a/apps/hash-agents/pyproject.toml
+++ b/apps/hash-agents/pyproject.toml
@@ -19,6 +19,13 @@ gunicorn = { version = "^20.1.0", extras = ["gevent"], optional = true }
 [tool.poetry.extras]
 production = ["gunicorn"]
 
+[tool.poetry.group.dev.dependencies]
+black = "^23.3.0"
+
+[tool.black]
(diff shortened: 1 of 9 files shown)


apps/hash-agents/pyproject.toml
```diff
@@ -15,6 +15,13 @@ openai = "^0.27.4"
 python-dotenv = "^1.0.0"
 tiktoken = "^0.3.3"
 
+[tool.poetry.group.dev.dependencies]
+black = "^23.3.0"
+
+[tool.black]
+line-length = 100
```
indietyp, line 22: A specific reason why a non-standard line length was chosen? The default line length of 88 is already higher than the 79 that PEP8 enforces.

indietyp: might make sense to also opt into [preview mode](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#preview-style), especially parentheses in with expressions is very nice.

indietyp (changes requested):

apps/hash-agents/pyproject.toml
TimDiekmann: We use `100` in other places, but it's probably good to stick with the default for now. It's easy to change it later.

I like the preview-mode, thanks for the suggestion!
Feel free to suggest more 🚀

indietyp (approved):

thehabbos007 (approved):

PR #2463, opened 2023-04-20
Use `turbo` to run scripts and avoid `codegen` in `postinstall` step

TimDiekmann:

We have a (hopefully) working dependency tree so it's possible to rely on that and avoid the codegen step when warming up the repository to run the codegen. This is problematic because of the reasons outlined in the linked slack thread.

## 🔗 Related links

- [Slack Thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1681838808146979) _(internal)_

## 🔍 What does this change?

- Run all `dev:*` variants using `turbo --filter...`
- Run the `test:*` variants using `turbo --filter...`
- Remove the `codegen` step as `postinstall` step
- Use `codegen` instead of `build` in `hash-agents`
- Drive-by: consistently use `--` at the end of `turbo` command in yarn scripts to allow appending values from the CLI

This PR:


package.json
```diff
@@ -13,41 +13,41 @@
...
-    "dev:backend:search-loader": "yarn workspace @apps/hash-search-loader dev",
-    "dev:frontend": "yarn workspace @apps/hash-frontend dev",
-    "external-services": "turbo run deploy --filter='@apps/hash-external-services' --",
-    "external-services:offline": "turbo run deploy:offline --filter='@apps/hash-external-services' --",
-    "external-services:test": "turbo run deploy:test --filter='@apps/hash-external-services' --",
+    "dev": "turbo run dev --filter '@apps/hash-api' --filter '@apps/hash-search-loader' --filter '@apps/hash-realtime' --filter '@apps/hash-frontend' --",
+    "dev:backend": "turbo run dev --filter '@apps/hash-api' --filter '@apps/hash-realtime' --filter '@apps/hash-search-loader' --",
+    "dev:backend:api": "turbo run dev --filter '@apps/hash-api' --",
+    "dev:backend:realtime": "turbo run dev --filter '@apps/hash-realtime' --",
+    "dev:backend:search-loader": "turbo run dev --filter '@apps/hash-search-loader' --",
```
thehabbos007, line 20: Do you think we should add `"persistent": true,` entries in these projects' `turbo.json` for their `dev` tasks to disallow other rules being dependent on them? The flags makes it so other tasks can't be dependent on the `dev` task (since they may run indefinitely)

apps/hash-agents/docker/Dockerfile
```diff
@@ -21,6 +21,8 @@ RUN yarn install --frozen-lockfile --prefer-offline && \
 
 COPY --from=base /app/out/full/ .
 
+RUN yarn turbo codegen --filter '@apps/hash-agents'
```
thehabbos007, line 24: Nice that this has been fixed for the agents :smile:

.github/workflows/ci.yml
```diff
@@ -256,7 +258,9 @@ jobs:
         uses: ./.github/actions/load-docker-images
 
       - name: Launch external services
-        run: yarn workspace @apps/hash-external-services deploy up --detach
+        run: |
+          yarn turbo codegen --filter '@apps/hash-external-services'
+          yarn workspace @apps/hash-external-services deploy up --detach
```
thehabbos007, line 263: Is this `codegen` step required because we're mounting the python files? Could we remove the need to explicitly call codgen by having a `codegen` dependency from the `deploy` task?

TimDiekmann: No, it will generate `.env.local` 😄

package.json
TimDiekmann: The global `turbo.json` contains `persistent: true` for `dev` because of this reason 🙂

.github/workflows/ci.yml
thehabbos007: I see, that makes sense, but a little strange :)

TimDiekmann: Yeah, but it was the only way I found to avoid building the images. Locally, we want the dependencies (and they should be cached in most cases), but in CI we have to work around this. 🙁

thehabbos007 (approved):

thehabbos007 (approved):

thehabbos007 (approved):

PR #2467, opened 2023-04-21
Add `ruff` as Python linter and a number of lints

TimDiekmann:

We want proper linting 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204431665478215/f) _(internal)_

## 🔍 What does this change?

- Add `ruff` as Python linter
- Add checks for CI for `ruff`
- Add various lint selections and apply fixes

This PR:


apps/hash-agents/app/__init__.py
```diff
@@ -1,29 +1,31 @@
...
-from logging import getLogger
-from flask import Flask, request
+from flask import Flask, request, Response, jsonify
 from dotenv import load_dotenv, find_dotenv
 
 from .logger import setup_logging
-from .agents import call_agent
+from .agents import call_agent, InvalidAgentNameError, InvalidAgentOutputError
+
+LOG = logging.getLogger(__name__)
```
indietyp, line 18: Bikeshedding, is fine with `LOG`, but I've often seen/used `logger` as a variable instead.

indietyp (dismissed):

apps/hash-agents/app/__init__.py
indietyp: if we go by the [python how-to docs](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial) it says:

> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
>
> ```python
> logger = logging.getLogger(__name__)
> ```

indietyp: fixed in [97862bc](https://github.com/hashintel/hash/pull/2467/commits/97862bc30314d335f5a95fb219f9ca5460969373)

indietyp (approved):

indietyp (approved):

PR #2482, opened 2023-04-22
`deer`: make char more tolerant

indietyp:

This achieves the roughly same thing as #2423 and #2425. If `char` receives a string with a single character as value, it will still pass, mirroring the behavior of serde (and making it more tolerant in general).

This PR:


libs/deer/tests/test_impls_core_string.rs
```diff
@@ -15,3 +15,10 @@ fn string_ok() {
     let value_str = "example";
     assert_tokens(&value_str, &[Token::BorrowedStr("example")]);
 }
+
+#[test]
+fn single_char_str_ok() {
+    assert_tokens(&'A', &[Token::BorrowedStr("A")]);
+    assert_tokens(&'A', &[Token::Str("A")]);
+    assert_tokens(&'A', &[Token::String("A")]);
+}
```
thehabbos007, line 24: Would it be useful to have a negative test for strings longer than 1 char?

thehabbos007 (dismissed):
Thank you! :sparkles:

thehabbos007 (approved):

PR #2484, opened 2023-04-24
Improve codegen of `hash-agents`

indietyp:

This improves codegen of hash-agents by not using quicktype directly but by utilizing [`datamodel-code-generator`](https://github.com/koxudaxi/datamodel-code-generator/). This change increases not only type safety (with better types), but also will validate the incoming and outgoing data!

This PR:


Alfred-Mountfield (changes requested):
We need to update the Dockerfile to install poetry within it so that codegen can be ran within the poetry env

indietyp:
Note: While technically working, in its current version, I'd very much not like to merge this PR. This PR requires every single user to install a poetry toolchain (or at least `datamodel-code-generator`), in a repository that is primarily python this might be okay, but with this added to the tooling required for TS and no interopt layer between them this is just chaos and too much bootstrap.

I still think that pydantic is the way to go. The benefits are non-negligible; we get better types, data validation, and integration into tools like FastAPI and CLI. Still, I don't think this is the way.

I will investigate different ways of achieving the same without the added dependency on `datamodel-code-generator` in the repository.

To indicate this dilemma, I have put the PR back into draft mode.

Alfred-Mountfield:
@indietyp 

> This PR requires every single user to install a poetry toolchain [..]

As discussed offline, the repo has been updated to introduce `poetry` as a dependency for running the HASH app anyway. We're likely to be expanding our usage of Python so if you do feel that this change to codegen is worthwhile to the current `hash-agents` implementation (or a future agent runner that looks somewhat similar) then I'm happy to review this when it's brought back in sync

apps/hash-agents/scripts/codegen.ts
Alfred-Mountfield, line 1: I know it's pretty time consuming, but it'd be helpful to provide test cases in the future, or at least instructions on how to test this appropriately. I appreciate that is a little mean to ask considering we don't presently have tests for this within the codebase, but especially regarding codegen I think we need to be checking edge-cases (at least relatively common ones)

I've tried creating a few test cases and have pushed them, and a revert, directly to this branch.
- [Try adding some test cases](https://github.com/hashintel/hash/pull/2484/commits/7b2b260ca66fc9c65fc11770e4de470161ccbdd0)
- [Check type name clashes](https://github.com/hashintel/hash/pull/2484/commits/7d28bb4b0b7abbf311958fd21e765f8e3ef3401b)

In doing so I've noticed a few limitations with our current codegen approach:

---

1. It does not seem possible to create a type with an empty array from TypeScript, `never[]` doesn't work `[]` doesn't work, and `any[] & { length: 0 }` produces something very strange.
  - _Note:_ on `main`,  `any[] & { length: 0}` produces an error "` Error: Trying to make an empty union - do you have an impossible type in your schema?.`"

2. `null` and `undefined` both get generated as `Any` in Python, as opposed to `None` (or being absent)
  - _Note:_ on `main`, `undefined` actually throws an error and `null` is **correctly** typed as `None`

---

Additionally, I wanted to know what would happen if someone tried to make an agent with either no `Input` or no `Output`. 

Codegen doesn't seem to error but I'm not sure what will happen with our dynamic resolution stuff, maybe you have an idea? (It's fine for us to just say for now that these are absolutely required and they can just return a boolean or flag or something)

Alfred-Mountfield (commented):
As always, thanks for your contributions! I'm a **huge** fan of us trying to get stronger types where we can. I think there are a few points we should discuss before this goes in. Perhaps we need to mitigate them, perhaps we do not.

Alfred-Mountfield (approved):

apps/hash-agents/scripts/codegen.ts
Alfred-Mountfield: Sweet, we can update the dependency when they publish a release.

PR #2487, opened 2023-04-24
`hash-agents`: rework the CLI

indietyp:

This reworks the CLI to use typer and typed arguments, this enables templates with different fields to properly be called and also upgrades the CLI in a major way.

This PR:

## 📹 Demo

image

image

image

### DIFF
--- a/apps/hash-agents/app/agents/__init__.py
+++ b/apps/hash-agents/app/agents/__init__.py
@@ -25,7 +25,7 @@ class InvalidAgentOutputError(ValueError):
 
 @beartype
 async def call_agent(name: str, **kwargs: dict) -> Coroutine[None, None, dict]:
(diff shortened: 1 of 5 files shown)


Alfred-Mountfield (changes requested):
As discussed offline, this seems really cool, and I love the output. Unfortunately, it breaks. It can handle some interestingly complex types, but fails on obvious ones which we will definitely need, such as objects. e.g.
```ts
type NonEmptyArray<T> = [T, ...T[]];

export type Input = {
  /** Math expression to evaluate */
  expression: string;
  array: NonEmptyArray<boolean | null>;
  obj: {
    objA: string;
    objB: number;
  };
};
```

Fails because it does not support `obj`

indietyp:
switching from `typer` to `argdantic` now allows for these types to be supported:

image

indietyp:
closing this in favor of #2512

PR #2493, opened 2023-04-26
Fixes and improved style in Agent Template

Alfred-Mountfield:

This PR adds a number of improvements to the Python Agent template, generally around styling and convention. In doing so it enables almost all of the `ruff` ruleset, however unfortunately `ruff` still does not pick up the pre-existing runtime crash that snuck in due to a missing parameter. It does not detect this as it does not implement Pylint's E1120. Nevertheless this is fixed in this PR.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204459915637147/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

See commits, a number of drive-bys were included too

This PR:

This fixed a mistake in the docs

## ⚠️ Known issues

- [ ] The docker container does not have `poetry` installed which makes it impossible for it to run `codegen` currently (which requires the Python environment to be setup)

## 🐾 Next steps

- Consider setting up `Pylint`
- Attempt to integrate the current linting and venv setup into the monorepo tooling better

## 🛡 What tests cover this?

- Unknown

## ❓ How to test this?

- Follow the instructions in the agent README

## 📹 Demo

N/A


apps/hash-agents/app/agents/__init__.py
```diff
@@ -11,35 +13,45 @@
 
 @beartype
 def find_allowed_agents() -> list[str]:
-    agents_dir = os.path.dirname(__file__)
```
indietyp, line 14: You did what Tim and I were too lazy to do, add docstrings 😅 Thanks!

apps/hash-agents/app/logger/http.py
```diff
@@ -41,9 +39,13 @@ async def logging_middleware(request: Request, call_next: Callable) -> Response:
 
         # Recreate the Uvicorn access log format,
         # but add all parameters as structured information
-        client = f"{client_host}:{client_port}"
         logger.info(
-            f"""{client} - "{http_method} {url} HTTP/{http_version}" {status_code}""",
+            """{%s:%s} - "{%s} {%s} HTTP/{%s}" {%s}""",
```
indietyp, line 43: Specific reason why? `f` strings are faster and AFAIK this is not really something you do w/ structlog


> Like in logging, positional arguments are [interpolated into the message string using %](https://docs.python.org/3/library/stdtypes.html#old-string-formatting). That might look dated, but it’s much faster than using [str.format](https://docs.python.org/3/library/stdtypes.html#str.format) and allows structlog to be used as drop-in replacement for [logging](https://docs.python.org/3/library/logging.html#module-logging). If you know that the log entry is always gonna be logged out, just use [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals) which are the fastest.

apps/hash-agents/pyproject.toml
```diff
@@ -38,8 +38,13 @@ target-version = ['py311']
 preview = true
 
 [tool.ruff]
-select = ["ANN", "B", "C901", "E", "F", "G201", "N", "UP", "S", "T20", "TRY"]
-ignore = ["ANN101"]
+select = ["ALL"]
```
indietyp, line 41: I am unsure if that is a good idea; for example, `COM812` directly influences the output of black and clashes with `COM819`, the `I` family of lints is already covered by `lint:isort`

indietyp (changes requested):
I know it's a draft, but these are just my comments on the current PR

apps/hash-agents/app/logger/http.py
Alfred-Mountfield: As it says, _If you know that the log entry is *always gonna be logged out*_. We do not, it's dependent on log level. Using f-strings in a logger is a codesmell (and a lint failure).

I agree it's not something you do with structlog, you generally want to actually structure them (e.g. `x=foo`) but this code seems to be trying to match a very specific format

apps/hash-agents/pyproject.toml
Alfred-Mountfield: I think we can get rid of `lint:isort` then, unless you have an argument against it. The lints that clash are automatically detected by `ruff`.

Everywhere else we have tried to be about as constrained as we can with lints (as in the opposite of permissive and just enable all of them until they prove annoying), I do not see why we wouldn't do that here.

The ruff guide does say that most people probably don't want `ALL` because it means lints will be implicitly enabled when you update versions, but that's actually the same strategy we employ with `clippy`.

Alfred-Mountfield: Do you have an argument for isort instead of `I`? I feel like reducing the tooling down would be better, and linting will fail until we resolve which one to use.

indietyp: we need isort anyway for `fix`, but I am fine with the `I` lint instead of `lint:isory`

Alfred-Mountfield: Why do we need `isort` for fix? I believe sorting is fixable through `ruff` is it not?

indietyp: you are completely right, yes

Alfred-Mountfield: Noice, I'll test out removing `isort` tomorrow

Alfred-Mountfield: I've removed it. [Remove isort in favour of ruff](https://github.com/hashintel/hash/pull/2493/commits/0bc3dde4f8654c1a1e7129db6bd797c380130dc6)

I tried editing the code artificially, screwing up imports all over the place, and `fix` and `lint` seems to still work quite well

indietyp (dismissed):

apps/hash-agents/package.json
```diff
@@ -6,20 +6,19 @@
...
-    "lint": "npm-run-all --continue-on-error \"lint:*\"",
     "lint:black": "poetry run black --check --diff --color app",
-    "lint:isort": "poetry run isort app --check --diff --color",
     "lint:lock-files": "poetry lock --check",
-    "lint:ruff": "poetry run ruff app"
+    "lint:ruff": "poetry run ruff app",
+    "poetry:install-dev": "poetry install --no-root --no-interaction --only main --only dev",
+    "poetry:install-lint-tools": "poetry install --no-root --no-interaction --only lint-tools",
+    "poetry:install-production": "poetry install --no-root --no-interaction --only main --only production",
+    "poetry:venv": "poetry install --no-root --no-interaction --only ,"
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yes, you are not able to provide nothing. I could not find a command to create the venv without installing some set of dependencies which was quite frustrating.

libs/@hashintel/design-system/README.md
```diff
@@ -13,3 +13,7 @@ turbo --filter '@apps/hashdotdesign' storybook
 ## Publishing
 
 See [`libs/README.md`](../../README.md#publishing)
+
+```
```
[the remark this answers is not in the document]
Alfred-Mountfield: Nope, this was removed in the parent branch and will be gone when this is rebased after it merges

indietyp (dismissed):

indietyp (approved):

thehabbos007 (approved):

PR #2495, opened 2023-04-27
Insert AI text based on given prompt into current page

nathggns:
This PR inserts text generated from the current prompt after the paragraph the cursor is currently in using a prompt entered into the command bar. You can also use cmd+G. This option only shows on the page editor (using a hacky method of the command bar checking the URL – we'll address this later by colocating the command with the page component, enabling us to remove use of global variables). 

You can enter topics like "World War 2" or "Google" to get back a one paragraph description of the topic. There is a basic safety net of no politics but we'd want to be more sophisticated in the future. 

This also includes changes to the command bar to make conditional options + options which use text input from the command bar itself. 

https://app.asana.com/0/1204260760205594/1204410864377820/f


apps/hash-agents/app/agents/generate-text-from-prompt/__main__.py
```diff
@@ -0,0 +1,51 @@
...
+
+from .io_types import Input, Output
+
+logger = structlog.stdlib.get_logger(__name__)
+
+SYSTEM_MESSAGE_CONTENT = (
+    "You are a text generator. Given a topic or a question, generate a"
+    "paragraph of text accurately describing the given topic or answering the"
+    " question. You must be concise. Do not comment on controversial issues"
+    " including politics."
```
nonparibus, line 15: I don't know that we want/need the "no controversy including politics" prompt, adding another layer beyond what's already in the OpenAI model itself. I'd rather we extricate ourselves from any debate around what's permissible and what's not until we have the ability to devote dedicated resource to determining what our parameters should be - and coming up with a conscious policy, rather than making normative judgements re: what is in/out of scope beyond those already determined by the AI ethicists at our model providers (in this case OpenAI).

I'll look into how other API consumers handle this, but unless there are severe objections let's drop this for now. We'll make sure there's Additional Product ToS language around our AI-enabled features to shield against legal risks (while appreciating reputational ones may still arise, but eating them for now).

apps/hash-agents/app/agents/generate-text-from-prompt/__main__.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+def main(agent_input: Input) -> Output:
+    chat = ChatOpenAI(model_name="gpt-3.5-turbo", temperature=0)
```
nonparibus, line 19: Did we experiment with the temperature here? Cold!

nathggns: No – @Alfred-Mountfield do you think I should do something different?

Alfred-Mountfield: I'd recommend we stick with 0 until there are reasons to increase it. Reasons I can think of (for the future):
- When it's a part of a chain and we are trying to retry things that aren't satisfactory (we want a bit of randomness then to be able to regenerate a broken response and hope it's not broken)
- When a user wants more random output (e.g. generating text, they may want multiple different suggestions they can pick from)

For the _current_ purposes of the branch and these PRs, I'd recommend we keep it low so it's (mostly) deterministic to avoid unexpected behaviour for a given set of working prompts right now.

PR #2496, opened 2023-04-27
Prototype: Generate react block

luisbettencourt:

This PR adds the prototype made for generating react blocks.

To accomplish the live preview I'm creating a react container and changing the contents of the `App.js` file inside it. 

Things to look out for:
- Installing dependencies takes a long time. It's installing the dependencies and re-initializing the block, which takes a while and probably can be massively improved. It also requires refreshing the iframe manually to preview the changes;
- ~~Typing on the code editor makes the cursor jump around;~~
- ~~The agent call times out after around 20 seconds, it only completes for short blocks of code;~~
- This prototype is not yet using the block-template-react for previewing the app;
- A lot of the changes in this PR are due to the inclusion of the react app boilerplate. In the future this should just be generated when needed and be ignored by git.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204405566357825/f) _(internal)_

## 📹 Demo

Generate an app:

demo1

Change and preview the code:

demo1 5

Iterate over the prompts:

demo2

Install missing dependencies:

demo3


nonparibus (commented):
Can we add an AGPL license to the folders that are triggering the legal CODEOWNERS review, treating this as we would any other part of the HASH app? Thanks.

nonparibus:
Scratch that, GitHub made me think the review was being requested for a weird reason. Now I see. Can we update the README in the react-app block dir so it's obvious what we're licensing? This uses the FAQ block text as a placeholder rn.

PR #2362, opened 2023-04-05
Introduce the block design system

luisbettencourt:
## 🌟 What is the purpose of this PR?

This PR adds the Block Design System to hash

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1204160628104490/f) _(internal)_

## 🔍 What does this change?

- Added the `block-design-system` to `libs/@hashintel/` and moved block-specific components from hash design-system to there;
- Replaced block import paths to reflect the changes above.

## 🔍 Has this modified a publishable library?

This PR:

- [x] modifies an **npm**-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
- [ ] does not modify any publishable libraries
- [ ] I am unsure / need advice

PR #2502, opened 2023-04-28
Uncover and Fix Cyclic Dependency, Version-Control Vercel Configs, Fix Turbo Graph.

Alfred-Mountfield:

This fixes a number of issues uncovered within the repository while trying to get #2493 in.

- There was a cyclic dependency between packages that share GraphQL types (`@apps/hash-api`, `@local/hash-graphql-shared`, etc.). This was largely hidden because they were within GraphQL configs, and relied on the file-system, which meant that the dependencies were not within the `package.json` and not visible to turborepo.
  - While a circular dependency between packages is implicitly a problem by itself, it also surfaces when we try to rely on things like [`turbo prune`](https://turbo.build/repo/docs/reference/command-line-reference#turbo-prune)
  - This was fixed by hoisting the types to a shared package as they should have been, rather than being defined within an 'app' and imported elsewhere
- The configuration for our Vercel deployments was managed purely through Vercel's UI. This means that the configuration (such as build commands) was not version-controlled, and importantly it is shared between **all** new deployments. Thus, if a PR requires a change to the configuration, that change will break all other in-flight PRs.
  - This is fixed by including a `vercel.json` in the individual projects. This `vercel.json` works _because the root directory is set within the project configuration_
  image

  - For the `hash-frontend`, additional shell scripts have been added in preparation of requiring some more complicated setup, including installing Python on the Vercel machine.
- Requirements were missing from the README such as `Python` and `Java`
- The changes introduced in #2362 accidentally missed some required updates to the Turbo graph. These have been fixed in this PR, and:
- The Pull Request template has been updated (to this one) to result in an updated checklist of things to look at pre-merge, which now includes a section on Turborepo.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204497155201681/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Please see commits.

This PR:

The changes in this PR:

  - the README has been updated to include missing dependencies
  - the pull request template has been updated to contain a bigger check-list

The changes in this PR:

## ⚠️ Known issues

- Our frontend build is eating up the cache on Vercel. The webpack cache is filling up with hundreds of `.pack` files that persist across runs instead of being overridden. This means we can have well over 8GB of disk being used up which results in failed deployments. This can be mitigated by manually redeploying the Vercel deployment through the UI and choosing not to re-use the cache. - [Slack Thread](https://hashintel.slack.com/archives/C022217GAHF/p1682615224959189) _(internal)_
- Attempting to use `turbo prune` within the Vercel build results in an internal server error. This is probably due to us deleting some of Vercel's files that they are putting in the same root directory as the source code. The attempt at using git to remove only git tracked files did not seem to work.
 
## 🐾 Next steps

- Update #2493 to remove changes superseded by this
- Investigate why `turbo prune` does not work for Vercel
- Investigate Vercel cache issues

## 🛡 What tests cover this?

- Tooling changes, installations, lints, etc. Basically all tests, so please look at CI

## ❓ How to test this?

- See CI and deployments
- Attempt various turbo graphs locally should you wish (such as running the storybook)

## 📹 Demo

This PR


.github/pull_request_template.md
```diff
@@ -25,25 +25,60 @@
...
 
 This PR:
 
-- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
-- [ ] modifies a **Cargo**-publishable library and **I have amended the version**
-- [ ] modifies a **Cargo**-publishable library, but **it is not yet ready to publish**
-- [ ] modifies a **block** that will need publishing via GitHub action once merged
-- [ ] does not modify any publishable blocks or libraries, or modifications do not need publishing
-- [ ] I am unsure / need advice
+- [ ] **⚠️ This is a placeholder, please uncomment at least one of the lines following this one and then delete this. ⚠️**
```
Alfred-Mountfield, line 37: I opted for this pattern to try and solve your bugbear, @CiaranMn, about the number of "incomplete tasks" that appear for given PRs. This way we should be able to make sure that every checkbox has been checked

apps/hash-frontend/vercel-install.sh
```diff
@@ -0,0 +1,24 @@
...
+#git ls-tree --name-only -d -r -z HEAD | sort -rz | xargs -0 rm -rf
+#
+#echo "Moving pruned repo back to root"
+#mv out/* .
+#rm out -r
+
+# Install the pruned dependencies
+
+echo "Installing yarn dependencies"
+HUSKY=0 yarn install --frozen-lockfile --prefer-offline --force --build-from-source
```
Alfred-Mountfield, line 24: The use for these files becomes a lot more apparent in #2493

CiaranMn (approved):

thehabbos007 (commented):
GH fooled me, I wanted to add these comments individually

thehabbos007 (approved):

.github/pull_request_template.md
indietyp: omg yes, this was bothering me so much

PR #2506, opened 2023-05-01
Fix logout runtime error

benwerner01:

This PR fixes a runtime error that was thrown when logging out of the application. The cause was that the `client.resetStore()` not only clears the apollo client cache, but also refetches all active queries. This means that any queries which require an authenticated user (such as the `getAccountPagesTree` query) would throw authentication erros.

What I've done to address this is to use `client.clearStore()` instead (which only clears the apollo client cache), and added a manual call to `refetchUser` to update the state in the `AuthInfo` context provider.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204355839255065/f) _(internal)_

This PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

1. Checkout the branch
2. Start the developer environment
3. Log in with a user, and log out
4. Ensure no runtime errors are thrown in this process

## 📹 Demo

https://user-images.githubusercontent.com/42802102/235488604-e8b834ee-7e56-4a52-8ada-b1a935c4da1a.mov

---
  - https://app.asana.com/0/0/1204502471509799

### DIFF
--- a/apps/hash-frontend/src/components/hooks/use-logout-flow.ts
+++ b/apps/hash-frontend/src/components/hooks/use-logout-flow.ts
@@ -3,10 +3,13 @@ import { AxiosError } from "axios";
 import { DependencyList, useEffect, useState } from "react";
 
 import { resetLocalStorage } from "../../lib/config";
+import { useAuthInfo } from "../../pages/shared/auth-info-context";
 import { oryKratosClient } from "../../pages/shared/ory-kratos";
 
 export const useLogoutFlow = (deps?: DependencyList) => {
   const client = useApolloClient();
+
+  const { refetch: refetchUser } = useAuthInfo();
 
   const [logoutToken, setLogoutToken] = useState<string>("");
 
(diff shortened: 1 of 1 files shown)


apps/hash-frontend/src/components/hooks/use-logout-flow.ts
```diff
@@ -32,9 +35,11 @@ export const useLogoutFlow = (deps?: DependencyList) => {
       if (logoutToken) {
         await oryKratosClient.updateLogoutFlow({ token: logoutToken });
 
+        await refetchUser();
+
         resetLocalStorage();
 
-        await client.resetStore();
+        await client.clearStore();
```
benwerner01, line 42: @CiaranMn @nonparibus when a user logs out, currently they stay on the current page they're on. That means that if they're on a page, and logout, they'll see the page in readonly mode (tried to showcase that in the demo video in the PR description).

Are you okay with this behaviour for now, or would you like me to redirect the user to the `/login` page on logout?

nonparibus: I'm fine with this for the timebeing. Thanks.

thehabbos007 (approved):
Thank you! This is working for me locally as expected. Maybe a good idea to get someone from frontend to take a look still :)

CiaranMn (approved):
Thanks @benwerner01 !

PR #2509, opened 2023-05-02
Fix infinite redirect bug on login page

benwerner01:

Previously there was an infinite redirect bug on the login page that occurs in an edge case scenario where:

- the user is authenticated (i.e. has a valid cookie in the browser),
- the kratos API is running and accessible by the FE, and
- the backend API is either not running, or not accessible by the FE.

It was possible to get into this state during development if for example your port for the backend API (`5001`) is already allocated when running `yarn dev`.

This PR fixes the infinite redirect by instead throwing an error, when the FE is able to authenticate with kratos, but not able to retrieve the authenticated user entity on the login page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204260760205594/1204502471509800/f) _(internal)_

## 🔍 What does this change?

- converts the existing kratos `createFlowErrorHandler` into a `useKratosFlowErrorHandler` hook, so that `useAuthInfo` can be be accessed in its implementation
- only redirects the user from the login flow to the homepage if the kratos session is valid, and if the `authenticatedUser` is defined

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch
2. Run the developer environment, and login to the workspace
3. In the terminal you used to run the backend API (i.e. the terminal in which you ran `yarn dev`), use `cmd + c` to cancel the process
4. In a new terminal run a server on port `5001` to allocate the port of the backend API (using `python3 -m http.server 5001` for example)
5. Go back to the other terminal window, and run `yarn dev` again. Observe in the output that all services are able to start, except the backend API (because its port is allocated to another process on your machine)
6. Go back to the browser tab. You should've been redirected to the `/login` page (because the `authenticatedUser` could not be fetched), and you should see a `The user is authenticated, but the user entity could not be fetched. Check your connection to the backend API, and ensure it is running.` error.

## 📹 Demo

https://user-images.githubusercontent.com/42802102/235639084-330aa2ba-89b9-481e-9b78-322574a9d892.mov

### DIFF
--- a/apps/hash-frontend/src/pages/login.page.tsx
+++ b/apps/hash-frontend/src/pages/login.page.tsx
@@ -126,7 +126,6 @@ const LoginPage: NextPageWithLayout = () => {
             }
 
             updateActiveWorkspaceAccountId(authenticatedUser.accountId);
-            // console.log("redirect to /");
             void router.push("/");
           })
           .catch(handleFlowError)


apps/hash-frontend/src/pages/login.page.tsx
```diff
@@ -140,7 +126,7 @@ const LoginPage: NextPageWithLayout = () => {
             }
 
             updateActiveWorkspaceAccountId(authenticatedUser.accountId);
-
+            // console.log("redirect to /");
```
thehabbos007, line 129:
```suggestion

```

apps/hash-frontend/src/pages/shared/use-kratos-flow-error-handler.ts
```diff
@@ -0,0 +1,126 @@
...
+          }
+          case "session_already_available":
+            // User is already signed in, if we're in the login flow let's redirect them home!
+            if (flowType === "login") {
+              if (!authenticatedUser) {
+                throw new Error(
+                  "The user is authenticated, but the user entity could not be fetched. Check your connection to the backend API, and ensure it is running.",
+                );
+              }
+              await router.push("/");
```
thehabbos007, line 47: Just for my own understanding, is this the main check that prevents the infinite login redirect?

thehabbos007 (commented):
Looks good to me, thank you Ben. I'm happy to approve after the comments have been addressed :)

apps/hash-frontend/src/pages/login.page.tsx
benwerner01: Oops, thanks for catching! Tracking this one down was a pain 😅

apps/hash-frontend/src/pages/shared/use-kratos-flow-error-handler.ts
benwerner01: Correct, essentially the loop that was occurring was:
1. Navigation to `/login` page
2. Existing kratos session can be used for authentication, so redirect to homepage
3. Navigation to `/`
4. Authenticated user could not be fetcehd, redirect to `/login` page
5. Navigation to `/login` page

And now the loop is broken because we only redirect from the `/login` page to `/` when the authenticated user has been successfully fetched.

thehabbos007 (approved):

PR #2511, opened 2023-05-03
Add script to create and configure a new block

CiaranMn:

When adding a new block to the repo, we need to amend some entries in `package.json` and modify some files (add licenses and eslint config, overwrite tsconfig.json).

This PR introduces a script which does that to save time

It also takes the opportunity to update block dependencies (since new blocks will be created with these updated dependencies).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204520547863397/f) _(internal)_

This PR:

The changes in this PR:

  -  The `README.md` in `blocks` has been updated to include mention of the new script

The changes in this PR:

## ❓ How to test this?

1.  Try `yarn create-block test-block` and check the outputted files are consistent with other blocks


blocks/faq/src/types/generated/block-entity.ts
```diff
@@ -30,16 +30,16 @@ export type DescriptionPropertyValue = Text;
 
 export type FAQBlock = Entity<FAQBlockProperties>;
 
-export type FAQBlockHasFrequentlyAskedQuestionLinks = {
+export type FAQBlockHasFrequentlyAskedQuestionLink = {
```
Alfred-Mountfield, line 33: I really do want to change codegen to not keep acronyms as capitals 😭

libs/@local/repo-chores/scripts/create-block.ts
Alfred-Mountfield, line 1: Commenting on the file as I'm not entirely sure where it happens (perhaps `create-block-app`), but do we want to be making/leaving `.gitignore` files within the individual block dirs?

libs/@local/repo-chores/scripts/create-block.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const packageJsonPath = path.join(newBlockFolder, "package.json");
+
+  const packageJsonString = fs.readFileSync(packageJsonPath, "utf-8");
+
+  const packageJson = JSON.parse(packageJsonString) as JsonObject;
+
+  packageJson.name = `@blocks/${blockName}`;
+  packageJson.license = "(MIT OR Apache-2.0)";
+  packageJson.author = "HASH";
```
Alfred-Mountfield, line 41: This doesn't seem to work for me, I've still got old fields within it:
```
  "license": "MIT",
  "author": "Alfred Mountfield",
```

libs/@local/repo-chores/scripts/create-block.ts
```diff
@@ -0,0 +1,77 @@
...
+  };
+  (packageJson.scripts as JsonObject)["fix:eslint"] = "eslint --fix .";
+  (packageJson.scripts as JsonObject)["lint:eslint"] =
+    "eslint --report-unused-disable-directives .";
+  (packageJson.scripts as JsonObject).format =
+    "prettier --write --ignore-unknown src/types/generated/*.ts";
+
+  (packageJson.devDependencies as JsonObject)["@local/eslint-config"] =
+    "0.0.0-private";
+
```
Alfred-Mountfield, line 55: Do we also want to update things such as `blockprotocol.displayName`, or `blockprotocol.name`?

Alfred-Mountfield (changes requested):
Really excited to get this in! Unfortunately it's not quite working for me locally at the moment 😁

libs/@local/repo-chores/scripts/create-block.ts
CiaranMn: Oh oops, I did notice that then somehow forgot it

libs/@local/repo-chores/scripts/create-block.ts
CiaranMn: Added in https://github.com/hashintel/hash/commit/725bd38352ad148282a61eb199430f3eb8518a87

Best we can do for `displayName` is capitalise the block name, not sure what else to do

libs/@local/repo-chores/scripts/create-block.ts
CiaranMn: Fixed in https://github.com/hashintel/hash/commit/725bd38352ad148282a61eb199430f3eb8518a87

libs/@local/repo-chores/scripts/create-block.ts
CiaranMn: We could do this but there's a slight chance that we introduce stuff to the `.gitignore` un the block templates in future that we miss. It'd _probably_ be fine but they're not doing any harm at the moment so I'd rather not add it into this PR.

Alfred-Mountfield: Is that not a problem anyway? If we update the block _templates_ we're still going to need to go through the blocks we've already made in the repo to migrate them. I don't think we can get around needing to remember changes we make upstream.

Having the `.gitignore` _in_ the package is problematic because it's inconsistent with other things we do such as `.dockerignore`, or `.prettierignore` (and the `eslintignore` is generated from the top-level one too).

Alfred-Mountfield, line 1: Do we want to try and handle dependency consistency? (e.g. if the block template contains the stable versions of the BP packages but we want to release canary ones)

Alfred-Mountfield: (Not a request, a lot of these things I'm asking we can defer, not do, or decide we don't want.)

libs/@local/repo-chores/scripts/create-block.ts
Alfred-Mountfield: Cheers

libs/@local/repo-chores/scripts/create-block.ts
Alfred-Mountfield: Yeah it behaves a bit weird when you make a block with kebab-case (e.g. `test-block` gets turned into `Test-block`), but I don't think it's worth us sinking time into

Alfred-Mountfield (approved):
A few open questions but none of them are blocking, glad we've got something like this in! We should _think_ about doing a similar thing to create a new TS lib, Rust lib, and Python lib

libs/@local/repo-chores/scripts/create-block.ts
CiaranMn: Let's address it when there's a change in it that introduces a practical problem

Alfred-Mountfield: I think it would be a good idea to hoist the `.gitignore` now though, because of the possible strange implications of the other `.ignore` files. And then worry about addressing the ergonomics of 

>there's a slight chance that we introduce stuff to the .gitignore in the block templates in future that we miss in the repo

if there's a change in it that introduces a practical problem

CiaranMn: I'd leave dependency consistency to be handled via linting / fixing as it is now, i.e. when introducing a package that has an inconsistent dependency either it's auto-fixed, we decide to introduce inconsistency and therefore ignore some package or another, or we decide to bring all other packages in line with the new, minority version. We could try and automate one of those approaches in this script but I don't think it's worth it and might hide the fact that a decision has been made which might not be the right one for the context.

Alfred-Mountfield: That way, we'd quite quickly notice if some files were being tracked when they shouldn't be, and then we'd be prompted to update _all_ the ignore files, whereas this could introduce git-ignored files but not ignore the others and we wouldn't have an obvious indication because of the local `.gitignore`.

Defer to your judgement, but just saying the current script creates packages that break our repo's guidelines re. this

CiaranMn: I have addressed this in https://github.com/hashintel/hash/pull/2511/commits/d0bede85af842607ea07d7394db3c923122772a3

Alfred-Mountfield (approved):

nonparibus (approved):

PR #2513, opened 2023-05-03
`deer`: introduce `IdentifierVisitor`

indietyp:
🌟 What is the purpose of this PR?

This implements a specialized visitor that is to be used in conjunction with things that are identifiers, like enum variant names or struct names. The difference between this `IdentifierVIsitor` and the existing `Visitor` is that the scope is very limited, prohibiting the use of borrowed data and only allowing a small subset of methods. The need for a specialized function like this has been identified after studying serdes implementation as well as other formats that use different identifiers. This method allows `Deserializers` to choose what counts as an identifier.

A follow-up PR will add a declarative helper macro that will allow the easy creation of such an `IdentifierVisitor`.

This PR:

The changes in this PR:

- [x] is in a state where docs changes are not _yet_ required but will be
  - this is tracked within: N/A

The changes in this PR:

## 🐾 Next steps

* declarative macro
* derive macro for `Deserialize` 🎉


indietyp:
Assigning @thehabbos007 directly, as I did an oopsie, and earlier this week, we merged the identifier macro PR (#2514) accidentally into this PR. This means the code is duplicate, and @thehabbos007 reviewed that PR, hope that's ok!

thehabbos007 (dismissed):
Looks good to me. Thank you!

thehabbos007 (approved):

PR #2516, opened 2023-05-03
Disable Edit property/link type buttons for types that belong to other users/orgs

luisbettencourt:

This PR disables the "edit property type" and "edit link type" buttons when the property type/link type doesn't belong to the user.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204081024577367/f) _(internal)_

## 🔍 What does this change?

- Added `canEditResource` to the ontologyFunctions passed to the entity type editor;
- Called it to check if the "Edit property type"/"Edit link type" buttons should be disabled;
- Added a tooltip to explain that the button is the disabled because the type doesn't belong to the user.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Apply changes to the entity type editor in Blockprotocol.

## ❓ How to test this?

1.  Create an entity type;
2. Create a new property type and check if you can edit it;
3. Add an existing property type from another namespace and check if the edit button is disabled;
4. Do the same for link types.

## 📹 Demo

image

image


luisbettencourt:
@nonparibus @CiaranMn Are you happy with the tooltips in the screenshots above?

nonparibus:
@luisbettencourt Yes

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts
```diff
@@ -123,11 +126,26 @@ export const useEditorOntologyFunctions = (
     [generateTypeUrlsForUser, getEntityType, getPropertyType],
   );
 
+  const canEditResource = useCallback(
+    (resourceNamespace: string) => {
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts:130]: "namespace" is not a concept that the BP or type editor needs to know about. We can pass the entire resource here, or its metadata.

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts
```diff
@@ -123,11 +126,26 @@ export const useEditorOntologyFunctions = (
...
+      if (!resourceNamespace || !user) {
+        return false;
+      }
+
+      return !!(
+        resourceNamespace === user.shortname ||
+        user.memberOf.find((org) => org.shortname === resourceNamespace)
+      );
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts:138]: There's an existing function in the codebase `canUserEditResource` which we could use in place of this.

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx
```diff
@@ -24,7 +25,12 @@ export const DefinitionTab = ({
   ownedById,
   readonly,
 }: DefinitionTabProps) => {
-  const ontologyFunctions = useEditorOntologyFunctions(ownedById ?? null);
+  const { authenticatedUser } = useAuthInfo();
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab.tsx:28]: This hook can be inside `useEditorOntologyFunctions`

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card.tsx
```diff
@@ -140,6 +143,12 @@ const LinkTypeRow = ({
     editModalPopupState.close();
   };
 
+  const editDisabledReason = !canEditResource(linkNamespace)
+    ? "Can't edit link types that belong to other users or organizations you aren't a member of"
```
CiaranMn, line 147: This error message is specific to HASH and we might want to have multiple reasons for denying edit access in future. I would return the reason from the function, as `validateTitle` does.

CiaranMn (changes requested):
Thanks @luisbettencourt – I think this could be changed to make the type editor not need to know how editability is determined (in this case, namespace) and to not assume the reason for denial.

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts
Comment (luisbettencourt) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts]: Changed it so `canEditResource` takes the whole resource and also the `kind` of the resource (property type or link type), similarly to `validateTitle`.

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts
Comment (luisbettencourt) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts]: Thanks for the suggestion, I didn't use that at first because I didn't really have access to the accountId of the property types and link types, but @Alfred-Mountfield pointed out that checking against the namespace was flawed using the previous solution for types with the same namespace and different domains (`blockprocol.org/@bob01/types/entity-type/test and hash.ai/@bob01/types/entity-type/test would have the same namespace, even though the domain is different).

So I had to make changes to `LatestPropertyTypesContextProvider` and `useEntityTypeValue` to also get the types' metadata and pass it to `useEditorOntologyFunctions` to then check against the resource that `canEditResource` was called with.

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card.tsx
luisbettencourt: Done in 12c5824633ed9506136aae6ab04ff85ea6333c8a

CiaranMn (commented):
One small comment on copy, but otherwise have tested and this looks good

apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts
```diff
@@ -123,11 +136,45 @@ export const useEditorOntologyFunctions = (
...
+      return resourceAccountId &&
+        canUserEditResource(resourceAccountId, authenticatedUser)
+        ? {
+            allowed: true,
+            message: "ok",
+          }
+        : {
+            allowed: false,
+            message: `Can't edit ${
+              kind === "link-type" ? "link" : "property"
```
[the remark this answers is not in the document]
Comment (luisbettencourt) [reply apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/definition-tab/use-editor-ontology-functions.ts]: Changed it to "link entity types" as I think it's clearer (4a9d8e8db58431066dd4900ae41b55f7a9544791)

CiaranMn (approved):

PR #2517, opened 2023-05-04
Handle external types in HASH pages

thehabbos007:

This PR assigns blocks on pages their actual block entity type ID when creating block links and block instances. This leverages the work done with the type fetcher to automatically fetch the type of the block without needing to explicitly create the block before creating the block entities.

I had to change some adjust some of the existing code around the `blockMap` in the frontend code and the type fetcher transport for it to work end-to-end. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203771979583945/f) _(internal)_

## 🔍 What does this change?

- Adds the component ID to BlockMeta map as a dependency to the prosemirror save step
- Change type fetcher's transport from Messagepack to JSON as we were seeing strange behavior with Messagepack (see https://hashintel.slack.com/archives/C03F7V6DU9M/p1683203634322339)
- Correctly implement the componentId to block entity type id mapping function
- Remove the createEntityType page action
- Bumps Rust BP Type System git dependency
- Bumps rooks to `7.14.1`

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Checkout the branch, start the HASH app (external services and `yarn dev`)
1.  Sign in with any test user and create a page. Add a BP block to the page such as the kanban board or the code block
1.  Confirm that the blocks are working
1.  Confirm that the DB now contains new entity types that reflect the entity type IDs of the blocks
I used the following PG query:
```sql
select base_url, version, additional_metadata, properties from entity_editions ee
inner join entity_is_of_type et on et.entity_edition_id = ee.entity_edition_id
inner join ontology_id_with_metadata om on et.entity_type_ontology_id = om.ontology_id
where not starts_with(base_url, 'http://localhost')
```

## 📹 Demo

image


apps/hash-api/src/graphql/resolvers/knowledge/page/update-page-actions.ts
```diff
@@ -37,11 +36,6 @@ export const createEntityWithPlaceholdersFn =
           draft.existingEntityId,
         ) as EntityId;
       }
-      if (draft.entityTypeId) {
-        draft.entityTypeId = placeholderResults.get(
-          draft.entityTypeId,
-        ) as VersionedUrl;
-      }
```
thehabbos007, line 44: The `placeholderResults` and placeholder map object was previously used to store placeholders for entity type IDs as well. Now that we can pass the exact entity type ID from BP, the placeholder map is solely for entity ids.

apps/hash-api/src/graphql/resolvers/knowledge/page/update-page-actions.ts
```diff
@@ -90,7 +84,7 @@ const isPlaceholderId = (value: unknown): value is `placeholder-${string}` =>
   typeof value === "string" && value.startsWith("placeholder-");
 
 export class PlaceholderResultsMap {
-  private map = new Map<string, string>();
+  private map = new Map<string, EntityId>();
```
thehabbos007, line 87: see above

apps/hash-api/src/graphql/resolvers/knowledge/page/update-page-actions.ts
```diff
@@ -158,7 +152,9 @@ export const handleCreateNewEntity = async (params: {
       throw new UserInputError(`action ${params.index}: ${error}`);
     }
     throw new Error(
-      `createEntity: Could not create new entity: ${JSON.stringify(error)}`,
+      `createEntity: Could not create new entity: ${JSON.stringify(
+        error,
+      )}, trying to create ${JSON.stringify(params)}`,
```
thehabbos007, line 157: I added this for debugging purposes, but it turned out to be quite informative. Can remove if we don't want this in the error.

libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/page.typedef.ts
```diff
@@ -165,29 +165,6 @@ export const pageTypedef = gql`
     ownedById: OwnedById!
   }
 
-  """
-  Create an entity type, which you can then reference in future CreateEntityActions
-  """
-  input CreateEntityTypeAction {
```
thehabbos007, line 171: This action has been made redundant by the type fetcher, so I've removed it from the save action

libs/@local/hash-isomorphic-utils/src/blocks.ts
```diff
@@ -14,6 +14,8 @@ export type HashBlock = {
   meta: HashBlockMeta;
 };
 
+export type ComponentIdHashBlockMap = Record<string, HashBlock>;
```
thehabbos007, line 17: This used to be a type def in the frontend code, but is now used in the `save.ts` file, so i hoisted the type def and gave it a more precise name here.

libs/@local/hash-isomorphic-utils/src/save.ts
```diff
@@ -32,21 +33,17 @@ const generatePlaceholderId = () => `placeholder-${uuid()}`;
 const flipMap = <K, V>(map: Map<K, V>): Map<V, K> =>
   new Map(Array.from(map, ([key, value]) => [value, key] as const));
 
-type EntityTypeForComponentResult = [VersionedUrl, UpdatePageAction[]];
-
 /**
  * Given the entity 'store', the 'blocks' persisted to the database, and the PromiseMirror 'doc',
  * determines what changes are needed to persist changes to the database.
  */
-const calculateSaveActions = async (
```
thehabbos007, line 41: No network calls here anymore :taco:

libs/@local/hash-isomorphic-utils/src/save.ts
```diff
@@ -141,12 +136,11 @@ const calculateSaveActions = async (
           throw new Error("Cannot find parent entity");
         }
 
-        const [assumedEntityTypeId, newTypeActions] =
-          await getEntityTypeForComponent(blockEntity.componentId ?? "");
+        const assumedEntityTypeId = getEntityTypeForComponent(
+          blockEntity.componentId ?? "",
```
thehabbos007, line 140: This is still not great as the `componentId` is optional, not sure if we can do anything about this without investing more time cleaning up the entity store

libs/@local/hash-isomorphic-utils/src/save.ts
```diff
@@ -170,8 +164,7 @@ const calculateSaveActions = async (
        *       being handled here, so that's okay for now.
        */
       if (dependedOn) {
-        const idx = actions.findIndex((item) => !item.createEntityType);
-        actions.splice(idx === -1 ? 0 : idx, 0, action);
```
thehabbos007, line 174: This added `dependedOn` entities to the front of the queue after `createEntityType` actions. These actions don't exist anymore, so we just 'unshift' the entity to the front now

libs/@local/hash-isomorphic-utils/src/save.ts
```diff
@@ -390,18 +384,19 @@ export const save = async (
 
   // const entityTypeForComponentId = new Map<string, string>();
 
-  const [actions, placeholderToDraft] = await calculateSaveActions(
+  const [actions, placeholderToDraft] = calculateSaveActions(
     store,
     ownedById,
     /** @todo This type ID should *not* be hardcoded as is here. */
     TEXT_ENTITY_TYPE_ID,
```
CiaranMn, line 391: Does anyone know what this `@todo` is for? Not clear if it's still relevant, or why we don't just use `TEXT_ENTITY_TYPE_ID` in `calculateSaveActions` directly

probably one for @thehabbos007 @nathggns

apps/hash-graph/lib/graph/src/store/fetcher.rs
```diff
@@ -226,6 +226,7 @@ where
                 break;
             }
 
+            dbg!(&ontology_urls);
```
TimDiekmann, line 229: I wonder if there is a lint for this.

```suggestion
```

thehabbos007: Something is weird with the Rust lints for sure

libs/@local/hash-isomorphic-utils/src/save.ts
CiaranMn: What is the problem in hardcoding it like this? What is the thing we want to be doing instead?

CiaranMn: Thank you – I will update this todo to a comment explaining that in #2503

PR #2523, opened 2023-05-08
Ignore `clippy::too-many-lines` on `FilterSchemaAddon`

TimDiekmann:

https://github.com/hashintel/hash/pull/1844 added four lines to the `modify` method, which triggers the clippy lint. Generally, the lint is reasonable, but in this case, having it all in one function is more readable as the builders of `utoipa` are pretty verbose and boiler platy

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest.rs
@@ -422,6 +422,7 @@ impl Modify for OperationGraphTagAddon {
 struct FilterSchemaAddon;
 
 impl Modify for FilterSchemaAddon {
+    #[expect(clippy::too_many_lines)]
     fn modify(&self, openapi: &mut openapi::OpenApi) {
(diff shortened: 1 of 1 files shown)


Alfred-Mountfield (approved):
Did we figure out why this wasn't caught in CI? (I guess we should see a _failure_ in CI if there's a tooling mismatch or something now that we've added an `expect`)

thehabbos007 (approved):

TimDiekmann:
> Did we figure out why this wasn't caught in CI? (I guess we should see a _failure_ in CI if there's a tooling mismatch or something now that we've added an `expect`)

No, sadly not. Somehow, `clippy` was not running for `graph` (the library), however, the other PR was checking the CI. As the PR was very old I can imagine that it has something to do with caching, but I'm unsure. The command being run for `clippy` contains `--workspace`, so it should have been checked. I need to look into the merge queue job for this PR to properly track this down as this also does not seem to run the feature-powerset.

PR #2530, opened 2023-05-08
Provide initial setup to restore snapshots from BE tests

TimDiekmann:

This adds a very minimal set of functions to load a snapshot from the Backend Integration test suite.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204549651666220/f) _(internal)_

## 🔍 What does this change?

- Add a `subgraph` directory in the test suite, were I plan to insert new subgraph tests
- Provide a function to easily restore a snapshot (using the `status` library for errors)
- Add a simple test case to insert an empty snapshot, which only contains the metadata.

This PR:

The changes in this PR:

The changes in this PR:

- [x] affected the execution graph as it adds a dependency for the BE test suite on `hash-graph`, however, `turbo.json` does not need to be adjusted.


tests/hash-backend-integration/src/tests/setup.ts
```diff
@@ -16,3 +19,24 @@ export const recreateDbAndRunSchemaMigrations = async () => {
...
+      fetch("http://localhost:4001/snapshot", {
+        method: "POST",
+        body: snapshot,
+      }),
+    )
+    .then(async (response) => {
+      const status: GraphStatus = await response.json();
+      if (status.code !== "OK") {
+        if (status.message) {
+          throw new Error(`Snapshot restoration error: ${status.message}`);
```
Alfred-Mountfield, line 36: Simply printing the message will probably lose a lot of information, such as accompanying `ErrorInfo` structs.

tests/hash-backend-integration/src/tests/subgraph/simple.jsonl
Alfred-Mountfield, line 1: Do we perhaps want to call this `empty.jsonl`?

TimDiekmann: Actually, I wanted to add `empty.jsonl`, which is literally empty and ensure that this is failing but that could be named `empty-misformed.jsonl` or something. Ideas?

Alfred-Mountfield: Perhaps `valid/empty-state.jsonl`, and the other could be `invalid/empty.jsonl` and we could have other variants of `invalid` kinds

tests/hash-backend-integration/src/tests/setup.ts
Alfred-Mountfield: We do have `GraphApiError` you could reuse from the backend, but I agree about doing that in follow-up. I just think we should opt to `JSON.stringify` in the interim

TimDiekmann: I added a logging output: [Log full status object on failure](https://github.com/hashintel/hash/pull/2530/commits/9f4bdeb77fec167f2beda7d72fbc454abe045068)

Alfred-Mountfield: Can we also just put it in the error message for now please. We've had a few reports recently of devs not being able to find the information associated with an error and we don't really want them having to dig through a history of the logs. Especially as this is a test and sometimes messages get swallowed.

Alfred-Mountfield (approved):

tests/hash-backend-integration/src/tests/subgraph/simple.jsonl
TimDiekmann: I went with `pass` and `fail` as these are common terms for testing.

Alfred-Mountfield (dismissed):
LGTM 🪨 🚀

Alfred-Mountfield (approved):

PR #2538, opened 2023-05-09
Allow resetting database from BE integration tests

TimDiekmann:

To test various snapshots and to avoid conflicts between them and other test data the BE integration test require the ability to reset the Graph. This PR adds functions to remove existing entities, types, and accounts and provide a convenient function to reset the Graph to the provided Snapshot.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204549651666223/f) _(internal)_

## 🔍 What does this change?

- Move test-server function to `test-server.ts`
- Provide functions to delete entities, types, and accounts
- Provide convenient function to wipe the Graph and restore a provided snapshot

This PR does not modify any publishable blocks or libraries, or modifications do not need publishing.

The changes in this PR is internal and does not require a docs change, however, documentation for the functions has been added.

The changes in this PR does not affect the execution graph.

## 🐾 Next steps

- Provide functions to read data from the Graph
- Write actual tests

### DIFF
--- a/tests/hash-backend-integration/src/tests/test-server.ts
+++ b/tests/hash-backend-integration/src/tests/test-server.ts
@@ -1,0 +1,98 @@
+import { promises as fs } from "node:fs";
+
+import { GraphStatus } from "@apps/hash-graph/type-defs/status";
+import fetch from "node-fetch";
+
+const deleteRecords = async (endpoint: string) => {
+  await fetch(`http://127.0.0.1:4001/${endpoint}`, { method: "DELETE" }).then(
+    async (response) => {
+      const status: GraphStatus = await response.json();
+      if (status.code !== "OK") {
+        throw new Error(
+          `Could not remove ${endpoint}: ${JSON.stringify(status)}`,
+        );
+      }
+    },
+  );
+};
+
+/**
+ * Delete all accounts from the Graph.
+ *
+ * This will fail if there are is any data in the Graph.
+ */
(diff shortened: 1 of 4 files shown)


tests/hash-backend-integration/src/tests/test-server.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Restore a snapshot from a file.
+ */
+export const restoreSnapshot = async (snapshotPath: string) => {
+  await fs
+    .readFile(snapshotPath)
+    .then((snapshot) =>
+      fetch("http://127.0.0.1:4001/snapshot", {
+        method: "POST",
+        body: snapshot,
```
thehabbos007, line 71: Just pointing out that this would read the whole snapshot from disk to memory before passing it and not stream it as the body. This is likely fine for now, but would be problematic if we do tests on very large snapshots

thehabbos007 (approved):

tests/hash-backend-integration/src/tests/test-server.ts
TimDiekmann: Yes, thanks for flagging this explicitly. I'm aware of this issue but didn't want to put more effort into this, yet. For BE tests this probably is fine for a long time as I expect to have big snapshots mainly being used for benchmarks, but it probably wouldn't hurt to run some additional checks here.

PR #2542, opened 2023-05-09
Provide constants for `temporalAxes`

TimDiekmann:

We manually specify the `temporalAxes` quite a few time and this is pretty boiler-platy. With the expected new tests this is going to be worse, so this PR adds three small constants to define the most commonly used constants for `temporalAxes`.

## 🔗 Related links

## 🔍 What does this change?

- Provide 
    - `currentTemporalAxes`: Only read the current timestamp
    - `fullDecisionTimeAxis`: variable decision time with `unbounded` as start (currently unused)
    - `fullTransactionTimeAxis`: variable transaction time with `unbounded` as start (currently unused)
- Use constant in the whole codebase

This PR does not modify any publishable blocks or libraries, or modifications do not need publishing 

The changes in this PR is internal and does not require a docs change

The changes in this PR [x] does not affect the execution graph


apps/hash-api/src/graph/index.ts
```diff
@@ -47,6 +48,52 @@ export const zeroedGraphResolveDepths: GraphResolveDepths = {
   hasRightEntity: { incoming: 0, outgoing: 0 },
 };
 
+export const currentTemporalAxes: QueryTemporalAxesUnresolved = {
```
Alfred-Mountfield, line 51: Maybe call this `currentTimeInstantTemporalAxes` or something?

apps/hash-api/src/graph/index.ts
```diff
@@ -47,6 +48,52 @@ export const zeroedGraphResolveDepths: GraphResolveDepths = {
...
+    interval: {
+      start: {
+        kind: "unbounded",
+      },
+      end: null,
+    },
+  },
+};
+
+export const fullTransactionTimeAxis: QueryTemporalAxesUnresolved = {
```
Alfred-Mountfield, line 81: Is this used anywhere? if it's not I'd suggest we remove it

apps/hash-api/src/graph/index.ts
Alfred-Mountfield, line 51: Maybe we want to add some doc-comments on these to quickly explain them in English, we should remember that this stuff isn't super intuitive and others (or us in the future) might come back and really struggle to quickly grasp what these do.

Maybe we say something like "Slices the datastore across this instant of time", and "According to the database's most up to date knowledge (transaction time), return the full history of entities and the times at which those decisions were made"

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -130,7 +134,7 @@ export const queryEntitiesResolver: Extract<
   const { data: entitySubgraph } = await graphApi.getEntitiesByQuery({
     filter,
     graphResolveDepths: {
-      inheritsFrom: { outgoing: 0 },
+      ...zeroedGraphResolveDepths,
```
Alfred-Mountfield, line 137: Upon seeing this diff I'm not _sure_ I like this pattern. If we added a new depth to the query, or originally had forgotten to pass one of these (`constrainsValuesOn` for example) then there wouldn't be an error here.

Also, looking at the code, it's now unclear what _is_ zeroed, we have to mentally apply the difference between the full object and the relatively long list of things here..

Maybe the code ergonomics are worth it, but I'm not sure I lean that way right now

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -204,23 +196,21 @@ export const getEntityResolver: ResolverFn<
...
-        },
-      },
-    },
+    temporalAxes: entityVersion
+      ? {
+          pinned: {
+            axis: "transactionTime",
+            timestamp: null,
+          },
+          variable: {
```
Alfred-Mountfield, line 205: I know you didn't add this, but maybe we want to add a comment saying "Constrain the results so that they overlap the given instant specified by the entityVersion timestamp" or something

apps/hash-api/src/graph/index.ts
TimDiekmann: It's not used, yet, but I'm going to use it in the tests I'm about to write. (stated that they are unused in the OP)

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
TimDiekmann: Do you suggest to remove it from this code section, remove the constant everywhere, or just leave it in? It feels like `Partial<GraphResolveDepths>` generally is an issue.

Alfred-Mountfield: I'm not sure I'm proposing anything, we can leave it as is, just wanted to express some thoughts

apps/hash-api/src/graph/index.ts
TimDiekmann: I was struggling with the name. I think yours is at least better 😄

TimDiekmann: [Rename currentTemporalAxes to currentTimeInstantTemporalAxes](https://github.com/hashintel/hash/pull/2542/commits/22828cd1ad004368546fa11b86c177cc20ea3271)

TimDiekmann: I slightly refined the comments and added them, thanks for the suggestions! [Add documentation to constants](https://github.com/hashintel/hash/pull/2542/commits/892db700c238e01b3b4741e1420c4be6ec3d761d)

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
TimDiekmann: I moved that statement to a variable to make it more readable and added a comment on it: [Add comment to entity resolver](https://github.com/hashintel/hash/pull/2542/commits/663a1d1f94b400e672470c2228458dada5ceb401)

Alfred-Mountfield (approved):

PR #2543, opened 2023-05-09
Add ability to upgrade the version of expected entity types in links

luisbettencourt:

This PR adds the ability to upgrade the version of expected entity types in links in the entity type editor.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204080201525406/f) _(internal)_

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  Create two entity types;
2.  Create a link in the first entity type and add the second entity type as an expected entity type;
3.  Make changes to the second entity type and publish it to create a new version;
4. Go back to the first entity type (make sure you reload the page) and check if you can upgrade the version of the expected entity type in the link.

## 📹 Demo

demo


nonparibus:
@luisbettencourt Please can you screencap the behavior when a chip is at the end of a row, and the slide-out-to-expand action that reveals "update" would cause it to bleed onto the next line? Curious how this is being handled. IIRC the original Figma designs averted a need for slide-out behavior.

luisbettencourt:
I accounted for other chips bleeding over to other lines (which looks fine) but forgot to account for the chip that's expanding jumping around.
jump

Should we keep it as is but always be expanded? Or should we shrink it to just a button?

nonparibus:
@luisbettencourt Let's shrink to a button, and have 'Update' appear onhover.

nathggns:
@nonparibus I've done the update to try and get this merged – let me know if this is suitable. 

image

nonparibus:
@nathggns By default I would have the tooltip appear above, not beneath, and a little closer to the chip itself (but I would have thought we'd have a consistent approach to these anyway? A bit confused)

nathggns:
@nonparibus We have to specify the placement whereever we use tooltips. I'll put it above instead like this.

image

Spacing is handled by MUI, and I think this is the standard spacing used with most tooltips. I can reduce it if you want?

nonparibus:
@nathggns Thanks -- a slight reduction would be great. Poking around I see the need for this in other contexts, too, so I think the change will generalize well. Thank you.

nathggns (approved):

CiaranMn (approved):

PR #2551, opened 2023-05-11
Python Temporal worker with dev infrastructure

thehabbos007:

This PR adds a python worker for the Temporal cluster that is able to pick up tasks off the `aipy` task queue. This can be considered a scaffold/template for now, and will eventually take over the `hash-agents` app.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204521571620724/f) _(internal)_

## 🔍 What does this change?

- Adds Temporal Python worker app
- Changes dev compose to start the worker
- Adds python worker build steps to CI
- Adjusts documentation

This PR:

The changes in this PR:

    - Documentation for running the python worker 

The changes in this PR:

## ⚠️ Known issues

The workers operate in separate work queues so that we don't run unnecessary retries. We should in theory be able to add them to the same task queue and have them work identically as their implementation is more or less the same

## ❓ How to test this?

1.  Checkout the branch 
1.  Ensure you have an OpenAI key set in your `.env.local`
1.  Start external services with `yarn external-services up`
1.  Using the temporal CLI, run `temporal workflow execute -t aipy --type DemoWorkflow -i '"say hi"` which should execute the workflow defined in the python runner

## 📹 Demo

image


apps/hash-ai-worker-py/README.md
```diff
@@ -0,0 +1,44 @@
+# HASH AI worker (Python)
+
+## Requirements
+
+- Python 3.x > 3.10 (the version used in HASH is 3.11)
```
Alfred-Mountfield, line 5: Can you change this, and the other one in `hash-agents` to say 3.11? I think we actually just set that in poetry and it's what we use in CI and such so..

apps/hash-ai-worker-py/app/__main__.py
```diff
@@ -0,0 +1,64 @@
...
+from aiohttp import web
+
+from app.workflows import DemoWorkflow
+from app.activities import complete
+
+from dotenv import find_dotenv, load_dotenv
+
+
+load_dotenv()
+load_dotenv(dotenv_path=find_dotenv(filename=".env.local"), override=True)
```
Alfred-Mountfield, line 14: I think this `override` is what I've seen causing me problems in our code elsewhere, I think it actually overrides your environment using the file. Whereas I think the hierarchy we want is `.env` < `.env.local` < environment.

If we remove `override` does it just work?

apps/hash-ai-worker-py/app/workflows.py
```diff
@@ -0,0 +1,17 @@
+from datetime import timedelta
+from temporalio import workflow
+
+with workflow.unsafe.imports_passed_through():
```
Alfred-Mountfield, line 4: Do they say to use `unsafe` in the docs? if so.. pretty strange naming

apps/hash-ai-worker-py/app/workflows.py
```diff
@@ -0,0 +1,17 @@
...
+    from .activities import complete
+
+
+@workflow.defn
+class DemoWorkflow:
```
Alfred-Mountfield, line 9: Not suggesting we should change this, but having them both be the same workflow (or rather them having the same WorkflowID, even if our workers are subscribing to different task queues) is slightly concerning to me, because it might mislead people's intuition in ways we don't want to right now. 

When we have some form of statically defining workflows (codegen the signatures or whatever) then I think it would be a great example to have the same workflow in both, but until then, I think it'd be too easy to change the signature of one, or something else without realising the implications.. I think I'd just opt for distinct workflow names at this time, and look to have a cohesive demo when we have a codegen thingy

apps/hash-ai-worker-py/app/workflows.py
thehabbos007: Yes it seems so.
https://github.com/temporalio/sdk-python/blob/main/README.md#implementing-a-workflow
I believe the naming is like this to make it clear that you're passing through their "sandbox" for any imports defined underneath

apps/hash-ai-worker-py/app/__main__.py
thehabbos007: It does seem to work fine without the `override`. It correctly loads the `.env` and the `.env.local` while preserving the environment passed in explicitly.
The docs specify that the env will be overridden https://saurabh-kumar.com/python-dotenv/reference/

nonparibus (dismissed):
Legal ✅ Code review pending.

apps/hash-ai-worker-py/app/__main__.py
TimDiekmann: From [the docs](https://pypi.org/project/python-dotenv/):

> With `load_dotenv(override=True)` or `dotenv_values()`, the value of a variable is the first of the values defined in the  > following list:
> - Value of that variable in the .env file.
> - Value of that variable in the environment.
> - Default value, if provided.
> - Empty string.
> With `load_dotenv(override=False)`, the value of a variable is the first of the values defined in the following list:
> - Value of that variable in the environment.
> - Value of that variable in the .env file.
> - Default value, if provided.
> - Empty string.

I think we should remove the override
```suggestion
load_dotenv(dotenv_path=find_dotenv(filename=".env.local"))
```

TimDiekmann (commented):
Let's remove the override and fix `ruff`, then it should be good to be merged 🙂

TimDiekmann (approved):

nonparibus (approved):

PR #2554, opened 2023-05-12
Fix production Docker compose and images

thehabbos007:

This PR fixes problems with the production docker compose and the images used by it.

The CD pipeline should be fixed by this as well

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204588404823948/f) _(internal)_

## 🔍 What does this change?

- Fixes API prod docker compose (using turbo graph and adding Python for the agent typings)
- Fixes Frontend prod docker compose (also using turbo graph and addign Python for the agent typings)
- Adjusts the docker-compose to properly make use of the graph image (changing the `command`)
- Adds the type fetcher to the docker compose

This PR:

The changes in this PR:

The changes in this PR:

- [x] Does change the execution graph, but not because of the change. I have a suspicion that the graph was misrepresenting the entry that was changed for API codegen

## 🛡 What tests cover this?

- Manual running

## ❓ How to test this?

1.  Checkout the branch
1.  Run `docker compose --file apps/hash-external-services/docker-compose.prod.yml --env-file .env --env-file .env.local up --build` from the root of the repo
1.  Confirm that the service eventually starts at http://localhost:3000


apps/hash-api/turbo.json
```diff
@@ -4,7 +4,7 @@
     "codegen": {
       // Outputs of GraphQL generation
       "outputs": ["./src/**/*.gen.*"],
-      "dependsOn": ["^build"]
+      "dependsOn": ["^codegen"]
```
thehabbos007, line 7: This change may be incorrect, but I couldn't tell why this was depending on `build` for its dependencies

infra/docker/api/prod/Dockerfile
```diff
@@ -1,69 +1,73 @@
...
+#   seems to work with just these commands, so we opt to do this rather than needing to build it from source or
+#   install it from an external source such as https://github.com/indygreg/python-build-standalone (see frontend vercel-install)
+#   if this causes problem, or we update dependencies, we can change approach.
 
+COPY --from=python-base /usr/local/bin/python3 /usr/local/bin/python3
+COPY --from=python-base /usr/local/bin/python3.11 /usr/local/bin/python3.11
+COPY --from=python-base /usr/local/lib/python3.11 /usr/local/lib/python3.11
+COPY --from=python-base /usr/local/lib/libpython3.so /usr/local/lib/libpython3.so
+COPY --from=python-base /usr/local/lib/libpython3.11.so /usr/local/lib/libpython3.11.so
+COPY --from=python-base /usr/local/lib/libpython3.11.so.1.0 /usr/local/lib/libpython3.11.so.1.0
```
thehabbos007, line 47: Not too happy that we had to copy this so many times. Perhaps there's a way we can make this more extensible when we need python for the turbo build graph?

indietyp (commented):
Thanks for the heads-up 🎉

infra/docker/api/prod/Dockerfile
thehabbos007: The python3 in bullseye is not 3.11

TimDiekmann (approved):
Tested locally and works, thank you 🙂

PR #2555, opened 2023-05-12
Add very basic snapshot tests to BE integrations

TimDiekmann:

This is the initial PR to add the first BE integration tests using a snapshot as the data source.

## 🔗 Related links

## 🚫 Blocked By

- #2569 

## 🔍 What does this change?

- Adds the testdata set we use in the REST endpoint test
- Modify test data set to make it more readable
- Write a bunch of *very* simple tests

This PR:

The changes in this PR:

- [x] requires changes to docs. 
    - The documentation on how to create test-data and store it in a snapshot will be covered in a follow-up PR as described in [this Asana task](https://app.asana.com/0/1204274556967336/1204584047986189/f)

The changes in this PR:


tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -0,0 +1,184 @@
...
+          "2002-01-01T00:00Z" as Timestamp,
+          "2001-01-01T00:00Z" as Timestamp,
+        ),
+      },
+    }).then(getRoots);
+
+    expect(entities.length).toBe(0);
+  });
+
+  it("read entities at 2002-02-01 as of now", async () => {
```
thehabbos007, line 164: Having these "readable" date in the test data snapshot seems to have required quite a bit of manual tuning. Is this something we're going to do for all test data? It could be quite time intensive for us to clean the data. (similar question for the UUIDs). Or is it mostly just because this test snapshot is so "small" so it was easy to do?

thehabbos007 (approved):
Looks good to me, thank you. I have a question regarding the "clean" IDs we have here :)

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
TimDiekmann: I was thinking about writing a script for this but for now I just did it with this data set as it's small enough

PR #2556, opened 2023-05-12
Provide yarn script to run production environment

TimDiekmann:

While testing https://github.com/hashintel/hash/pull/2554 it was very annoying to run the production images as we have no script for it. This adds scripts similar to what we have in the dev-setup but without the test-server running and with an optimized build.

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-external-services/package.json
+++ b/apps/hash-external-services/package.json
@@ -7,6 +7,7 @@
     "codegen": "touch ../../.env.local",
     "deploy": "docker compose --project-name hash-external-services --file docker-compose.yml --file docker-compose.type-fetcher.yml --env-file ../../.env --env-file ../../.env.local",
     "deploy:offline": "docker compose --project-name hash-external-services --file docker-compose.yml --env-file ../../.env --env-file ../../.env.local",
+    "deploy:prod": "docker compose --project-name hash-external-services --file docker-compose.prod.yml --env-file ../../.env --env-file ../../.env.local",
     "deploy:prototype": "docker compose --project-name hash-prototype-services --file docker-compose.prototype.yml --env-file ../../.env --env-file ../../.env.local",
     "deploy:test": "docker compose --project-name hash-external-services --file docker-compose.yml --file docker-compose.type-fetcher.yml --file docker-compose.test.yml --env-file ../../.env --env-file ../../.env.local"
   },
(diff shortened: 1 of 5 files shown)


apps/hash-external-services/package.json
```diff
@@ -7,6 +7,7 @@
     "codegen": "touch ../../.env.local",
     "deploy": "docker compose --project-name hash-external-services --file docker-compose.yml --file docker-compose.type-fetcher.yml --env-file ../../.env --env-file ../../.env.local",
     "deploy:offline": "docker compose --project-name hash-external-services --file docker-compose.yml --env-file ../../.env --env-file ../../.env.local",
+    "deploy:prod": "docker compose --project-name hash-external-services --file docker-compose.prod.yml --env-file ../../.env --env-file ../../.env.local",
```
thehabbos007, line 10: I think this is fine, but one small detail is that we're layering the `env-file`s which forces the user to have a newer docker compose version. I know we set a repo-wide minimum version, but I'm thinking for hte production deployment script we should perhahps have a `.env.prod` that is an exact env file that can be used on its own? cc @indietyp

TimDiekmann: I was thinking about a `.env.prod` (and `.env.test`) as well but I'd like us not to duplicate every entry as this will be a pretty big maintenance overhead and they will fall out of sync. Similarly, I also like to make the `docker-compose.prod.yml` layered, so the maintenance cost there will be minimal as well.

indietyp: We currently have docker compose `2.16.0`/`2.12.0` deployed. I'd need to schedule a downtime, but that wouldn't pose a problem

indietyp: We currently have docker compose `2.16.0`/`2.12.0` deployed. I'd need to schedule a downtime, but that wouldn't pose a problem

apps/hash-external-services/turbo.json
```diff
@@ -31,6 +31,15 @@
         "codegen"
       ]
     },
+    "deploy:prod": {
+      "cache": false,
+      "dependsOn": [
+        "@apps/hash-graph#build:docker:prod",
+        "@apps/hash-agents#build:docker",
+        "@apps/hash-ai-worker-ts#build:docker",
```
thehabbos007, line 39: We don't use the TS workers in the production image :)
```suggestion
```

TimDiekmann: Good flag, thanks 😄

thehabbos007 (approved):

PR #2561, opened 2023-05-14
`intellij-plugin`: explicitely unset `until-version`

indietyp:

Previously `until-version` was specified implicitly, leading to the case where build invocations on different machines would lead to different `until-version` versions, which would lead to some being unable to use the plugin.

This fixes it by explicitly unsetting the value, through `untilBuild.convention()`

This PR:

The changes in this PR:

The changes in this PR:


nonparibus (approved):

TimDiekmann (commented):
Works now as expected, thanks!

PR #2569, opened 2023-05-15
Reset databases after running BE tests

TimDiekmann:

Some BE tests requires a clean database to avoid tests interfering each other. This PR cleans up the databases after running the tests.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204596705932035/f) _(internal)_

## 🔍 What does this change?

- Add function to remove a kratos identifier
- Add `afterAll` call to BE tests

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Currently, each user has to be deleted manually. Ideally, the database would just be reset.

### DIFF
--- a/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
+++ b/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
@@ -1,3 +1,4 @@
+import { deleteKratosIdentity } from "@apps/hash-api/src/auth/ory-kratos";
 import {
   currentTimeInstantTemporalAxes,
   ensureSystemGraphIsInitialized,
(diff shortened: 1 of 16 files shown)


tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
```diff
@@ -149,6 +152,20 @@ describe("Entity CRU", () => {
     });
   });
 
+  afterAll(async () => {
+    await deleteKratosIdentity({
+      kratosIdentityId: testUser.kratosIdentityId,
+    });
+    await deleteKratosIdentity({
+      kratosIdentityId: testUser2.kratosIdentityId,
+    });
```
TimDiekmann, line 161: Resetting the test users is not strictly needed but it's cleaner.

thehabbos007: Probably not necessary for us to do right now, but I assume manually keeping track of created/deleted users may be somewhat annoying/error prone. Perhaps we should have `createTestUser` persist a list of created users per-test so we can just loop through the `kratosIdentityId`s in these `afterAll`s

TimDiekmann: Yeah, 100% agreed!

PR #2576, opened 2023-05-16
Adjust ontology creation methods to load external types

TimDiekmann:

This changes the endpoints to allow to either pass an ontology-type schema URI (without the owner id) or to pass the schema (with the owner id). If in the former case a type is passed which matches the current domain, it will fail, otherwise the type will be fetched and inserted as external type.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204596705932067/f) _(internal)_

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2573
- https://github.com/hashintel/hash/pull/2574
- https://github.com/hashintel/hash/pull/2575

## 🔍 What does this change?

- Change the creation request parameters
- Adjust the OpenAPI spec (all changes are completely type safe)
- Adjust the `.justfile` as the `type-fetcher` feature has to be enabled

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Use these changes in the `hash-api` to insert external types if they don't exist

## 🛡 What tests cover this?

- A small endpoint test was added

## ❓ How to test this?

Use the rest endpoint test file and test various queries with external types.

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest/data_type.rs
@@ -94,7 +94,7 @@ struct CreateOwnedDataTypeRequest {
 #[serde(rename_all = "camelCase")]
 struct CreateExternalDataTypeRequest {
     #[schema(value_type = String)]
-    schema: VersionedUrl,
+    data_type_id: VersionedUrl,
     actor_id: RecordCreatedById,
 }
 
@@ -139,7 +139,7 @@ where
                 store
                     .load_external_type(
                         &domain_validator,
-                        OntologyTypeReference::DataTypeReference((&request.schema).into()),
+                        OntologyTypeReference::DataTypeReference((&request.data_type_id).into()),
                         request.actor_id,
                     )
                     .await?,
--- a/apps/hash-graph/lib/graph/src/api/rest/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest/data_type.rs
@@ -10,6 +10,8 @@ use type_system::{url::VersionedUrl, DataType};
 use utoipa::{OpenApi, ToSchema};
 
 use super::api_resource::RoutedResource;
+#[cfg(feature = "type-fetcher")]
+use crate::ontology::OntologyTypeReference;
 use crate::{
     api::rest::{
         json::Json,
@@ -70,11 +72,26 @@ impl RoutedResource for DataTypeResource {
 }
 
 #[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase", untagged)]
+enum CreateDataTypeRequest {
+    Owned(CreateOwnedDataTypeRequest),
+    #[cfg(feature = "type-fetcher")]
+    External(CreateExternalDataTypeRequest),
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
-struct CreateDataTypeRequest {
+struct CreateOwnedDataTypeRequest {
     #[schema(inline)]
     schema: MaybeListOfDataType,
     owned_by_id: OwnedById,
+    actor_id: RecordCreatedById,
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+struct CreateExternalDataTypeRequest {
+    schema: VersionedUrl,
     actor_id: RecordCreatedById,
 }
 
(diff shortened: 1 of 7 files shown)


apps/hash-graph/lib/graph/src/api/rest/data_type.rs
```diff
@@ -65,15 +73,31 @@ impl RoutedResource for DataTypeResource {
...
     schema: MaybeListOfDataType,
     owned_by_id: OwnedById,
     actor_id: RecordCreatedById,
 }
 
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+struct CreateExternalDataTypeRequest {
+    #[schema(value_type = String)]
+    schema: VersionedUrl,
```
thehabbos007, line 97: Not sure how I feel about this being called `schema`. Since we're using untagged anyway, couldn't we call this `data_type_id` or something? (relevant for all ontology types)

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -105,12 +129,70 @@ async fn create_entity_type<P: StorePool + Send>(
...
+) -> Result<Json<ListOrValue<OntologyElementMetadata>>, Response>
+where
+    for<'pool> P::Store<'pool>: RestApiStore,
+{
+    let mut store = pool.acquire().await.map_err(|report| {
+        tracing::error!(error=?report, "Could not acquire store");
+        status_to_response(Status::new(
+            hash_status::StatusCode::Internal,
+            Some(
+                "Could not acquire store. This is an internal error, please report to the \
```
thehabbos007, line 141: Any reason for this pool acquisition using `hash_status` when the `data_type` and `property_type` equivalent changes don't?

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -105,12 +129,70 @@ async fn create_entity_type<P: StorePool + Send>(
...
-    }) = body;
+    } = match body.0 {
+        CreateEntityTypeRequest::Owned(request) => request,
+        #[cfg(feature = "type-fetcher")]
+        CreateEntityTypeRequest::External(request) => {
+            return Ok(Json(ListOrValue::Value(
+                store
+                    .load_external_type(
+                        &domain_validator,
+                        OntologyTypeReference::DataTypeReference((&request.schema).into()),
```
thehabbos007, line 171:
```suggestion
                        OntologyTypeReference::EntityTypeReference((&request.schema).into()),
```
Would this have resulted in errors in other places?

apps/hash-graph/lib/graph/src/api/rest/property_type.rs
```diff
@@ -96,12 +119,35 @@ async fn create_property_type<P: StorePool + Send>(
...
-    }) = body;
+    } = match body.0 {
+        CreatePropertyTypeRequest::Owned(request) => request,
+        #[cfg(feature = "type-fetcher")]
+        CreatePropertyTypeRequest::External(request) => {
+            return Ok(Json(ListOrValue::Value(
+                store
+                    .load_external_type(
+                        &domain_validator,
+                        OntologyTypeReference::DataTypeReference((&request.schema).into()),
```
thehabbos007, line 144:
```suggestion
                        OntologyTypeReference::PropertyTypeReference((&request.schema).into()),
```
Similar to above, or is this intentional?

thehabbos007 (commented):
I realized kind of late that you haven't rebased yet. So this is the preliminary review. May as well post it :)

apps/hash-graph/lib/graph/src/api/rest/data_type.rs
TimDiekmann: I think calling it `xxx_type_id` is better, yes. Going to address this after rebasing (forgot that, sorry 😞 )

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
TimDiekmann: When the status library was introduced it seems that these methods were not adjusted, most probably due to the reason that the status lib in its current form is extremely boiler-platy and hard to use. As flagged offline I'm going to revisit error handling in the Rest API soon, so I chose to leave the behavior as it is right now.

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
TimDiekmann: Thanks for finding this, this wasn't intentional. It didn't break anything because I only used the `.url()` method.

apps/hash-graph/lib/graph/src/api/rest/data_type.rs
TimDiekmann: [`8debcbd` (#2576)](https://github.com/hashintel/hash/pull/2576/commits/8debcbd8bf171ed2317ff3c827f97367cfcfa002)

thehabbos007 (approved):
Awesome, thanks!

PR #2577, opened 2023-05-16
Jump to start of document when pressing Enter in page's title

luisbettencourt:

This PR makes it so the start of a document is focused when pressing the Enter key inside the Title of a page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056209/1203623179643222/f) _(internal)_

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

[Paragraph immediately loses focus in Chrome](https://app.asana.com/0/1203179076056209/1204610234070526/f)

## ❓ How to test this?

1.  Navigate to a page;
2.  Focus the title and press Enter;
3. Check that the first block in the document is focused.


luisbettencourt:
@nonparibus When the first block of a page isn't a text block, it still focus it, but non-text blocks don't really have any visual indicator that they are focused (only for elements focused inside it). Should it skip non-text blocks and focus the first available text block?

nonparibus:
> @nonparibus When the first block of a page isn't a text block it still focus it, but non-text blocks don't really have any visual indicator that they are focused (only for elements focused inside it). Should it skip non-text blocks and focus the first available text block?

No, let's insert a new paragraph block immediately beneath the header in this case (above the existing non-text block as the first element in the page), instead.

nathggns (dismissed):

luisbettencourt:
> > @nonparibus When the first block of a page isn't a text block it still focus it, but non-text blocks don't really have any visual indicator that they are focused (only for elements focused inside it). Should it skip non-text blocks and focus the first available text block?
> 
> No, let's insert a new paragraph block immediately beneath the header in this case (above the existing non-text block as the first element in the page), instead.

This only applies when hitting Enter, not when pressing Arrow Down, right?

nonparibus:
@luisbettencourt Yes. When pressing the down arrow key the block beneath should be inserted into (or selected/highlighted if it is non-text).

nathggns (dismissed):
linting needs to be addressed i think

nathggns (approved):

PR #2578, opened 2023-05-17
Add entity type base URL filter

thehabbos007:

This PR adds a `queryEntities` filter to HASH that allows checking against a non-versioned `EntityTypeId` a `EntityTypeBaseUrl`. I am aware we have multiple names for this, but since the query syntax is explicitly for querying entities, I wanted to be explicit about the fact that `EntityTypeBaseUrl` queries for the entity type of the entity.

We should likely not keep adding these cases to the query translator, as it can turn into quite a messy function, this one is quite useful for parts of the frontend.

## 🔍 What does this change?

- Adds `EntityTypeBaseUrl` clause to the `queryEntities` translation function
- Update the frontend entity type 'entities' tab to make use of the new filter (fixing the todo)
- Add warning when querying with empty `queryEntities` `operation`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

We probably do not just to just warn when an empty query is given. Perhaps we should error?

## ❓ How to test this?

1.  Checkout the branch 
1.  Start external services + `yarn dev`
1.  Create new entity type, create entities for that type and observe that they appear for the entities

## 📹 Demo

image
image

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
+++ b/apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
@@ -116,7 +116,7 @@ export const queryEntitiesResolver: Extract<
     hasLeftEntity,
     hasRightEntity,
   },
-  { dataSources },
+  { logger, dataSources },
   __,
 ) => {
   const { graphApi } = dataSources;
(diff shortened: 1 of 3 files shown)


apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -131,6 +131,12 @@ export const queryEntitiesResolver: Extract<
     ? bpMultiFilterToGraphFilter(operation.multiFilter)
     : { any: [] };
 
+  if ("any" in filter && filter.any.length === 0) {
+    logger.warn(
+      "QueryEntities called with empty filter, which means returning an empty subgraph. This is probably not what you want.",
+    );
+  }
```
TimDiekmann, line 138: Thanks for adding this!

TimDiekmann (approved):
Thank you!

PR #2579, opened 2023-05-17
Require a `'static` lifetime for transpiling

TimDiekmann:

We want to avoid user-inserted strings to be directly inserted into the database. Up until now we errored as soon as the compiler tried to transpire a string, which was user provided. With this change, it's becoming a compile time error as soon as an implementation is added to transpire a non-static string.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1203821263193164/f) _(internal)_

## 🔍 What does this change?

Most notably:
```diff
- pub trait Transpile
+ pub trait Transpile: 'static
```

In order to achieve this, a `into_owned` method was added to `Column` which makes an expression `'static` by moving out the dynamic part and storing an index into the parameters array. With this change, the `'c` lifetime could be removed from the compiler!

This PR:

The changes in this PR:

The changes in this PR:

- [x] does not affect the execution graph -

## 🛡 What tests cover this?

There are a bunch of tests for the compiler (probably the area in the Graph which is most tested). Also, every integration test which is querying the graph will implicitly test this.

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, collections::HashSet, fmt::Display, marker::PhantomData};
+use std::{collections::HashSet, fmt::Display, marker::PhantomData};
 
 use postgres_types::ToSql;
 use tokio_postgres::row::RowIndex;
@@ -6,10 +6,7 @@ use tokio_postgres::row::RowIndex;
 use crate::{
     store::{
         postgres::query::{
-            table::{
-                DataTypes, EntityEditions, EntityTemporalMetadata, EntityTypes, JsonField,
-                OntologyIds, PropertyTypes,
-            },
+            table::{EntityTemporalMetadata, OntologyIds},
             Alias, AliasedColumn, AliasedTable, Column, Condition, Distinctness, EqualityOperator,
             Expression, Function, JoinExpression, OrderByExpression, Ordering, PostgresQueryPath,
             PostgresRecord, SelectExpression, SelectStatement, Table, Transpile, WhereExpression,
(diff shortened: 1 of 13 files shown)


apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -341,38 +335,16 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
...
-                    JsonField::JsonPathParameter(self.artifacts.parameters.len()),
-                )))
-            }
-            column => column,
+        let (column, parameter) = path
+            .terminating_column()
+            .into_owned(self.artifacts.parameters.len() + 1);
+
+        if let Some(parameter) = parameter {
+            self.artifacts.parameters.push(parameter);
```
TimDiekmann, line 347: I'm particularly happy that this is now finally removed and statically checked.

apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
```diff
@@ -341,38 +335,16 @@ impl<'c, 'p: 'c, R: PostgresRecord> SelectCompiler<'c, 'p, R> {
...
-            )))) => {
-                self.artifacts.parameters.push(field);
-                Column::EntityEditions(EntityEditions::Properties(Some(
-                    JsonField::JsonPathParameter(self.artifacts.parameters.len()),
-                )))
-            }
-            column => column,
+        let (column, parameter) = path
+            .terminating_column()
+            .into_owned(self.artifacts.parameters.len() + 1);
```
thehabbos007, line 344: Why is the `+1` required here? wouldn't the `.len()` be 1 higher than the current index automatically? I would assume that `len() + 1` would cause the current index to be one higher than the current index?

thehabbos007 (commented):
Thank you, Tim. This is a welcome simplification! Just have one question :)

apps/hash-graph/lib/graph/src/store/postgres/query/compile.rs
thehabbos007: Ah of course. Makes sense :+1:

thehabbos007 (approved):

PR #2581, opened 2023-05-18
H-29: Refactor the `Command Bar` (k-bar v2)

nathggns:

This rewrites the Command Bar with the same functionality before (as well as the functionality implemented by https://github.com/hashintel/hash/pull/2508 to have async resolution of screens within the Command Bar). The purpose of the rewrite is to enable the implementation of a "cheat sheet", where hot keys are centrally registered and it is possible to be globally aware of every option / whether it is active in the current particular context. We can also now colocate the functionality of activating an option in the command bar with the component it acts on. 

This PR also implements a very basic cheat sheet. 

Originally defined constraints:

- We need to ensure each item in the key map is “exclusive” – i.e, the same hotkey is never registered to do two different things in the same context
- Do we also want it to be globally exclusive – i.e, the same hotkey is never registered to do two different things in any context?
- We want to ability to colocate at least some command bar functionality with the component it acts on – i.e, to put the code of “insert text into page” functionality which interacts with Prosemirror within the react component, not centralised within the definition of the command bar
- We want the ability to show a “cheat sheet” of all hot keys registered – including the ones available within the current context, but also the ones registered globally (and maybe ones available within any context)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204501027002758/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Rewrite of the command bar to be based on a global store of command menus / options, with a new `useCommandBarOption` hook to "active" / "deactivate" options when components are visible or not. 

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

N/A

## 🐾 Next steps

There will be work to actually set up the commands we want and to make this look pretty. 

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Use command+K and ? in different parts of the app


apps/hash-frontend/src/shared/command-bar.tsx
luisbettencourt, line 1: This should probably be split into multiple files

luisbettencourt (changes requested):
The cheatsheet is looking pretty rough with the outline and bullet points on each item, but probably not something to be fixed in this PR?

apps/hash-frontend/src/shared/command-bar.tsx
nonparibus: As part of this PR, @nathggns has agreed to split the shortcuts themselves out into their own standalone file.

apps/hash-frontend/src/shared/layout/plain-layout.tsx
```diff
@@ -1,14 +1,18 @@
...
+import { FunctionComponent, ReactNode, useContext } from "react";
 
 import { isProduction } from "../../lib/config";
+import { WorkspaceContext } from "../../pages/shared/workspace-context";
 import { CommandBar } from "../command-bar";
 
 export const PlainLayout: FunctionComponent<{
   children?: ReactNode;
 }> = ({ children }) => {
+  const { activeWorkspaceAccountId } = useContext(WorkspaceContext);
```
CiaranMn, line 14: Couldn't this be done in the CommandBar itself? Saves passing it down

apps/hash-frontend/src/shared/command-bar.tsx
```diff
@@ -197,14 +203,28 @@ const doesIdentifierMatchKeyboardEvent = (
...
+  const { data: accountPages } = useAccountPages(namespaceOwnedById);
+
+  const lastRootPageIndex = useMemo(() => {
+    const rootPages = accountPages
+      .filter(({ parentPageEntityId }) => parentPageEntityId === null)
+      .map(({ index }) => index)
+      .sort();
+
+    return rootPages[rootPages.length - 1] ?? null;
+  }, [accountPages]);
```
CiaranMn, line 226: Can we add `lastRootPageIndex` to the return of `useAccountPages`? 

It could be calculated in the same `map` that's already happening there and saves repeating this in multiple places

CiaranMn (commented):
Thanks @benwerner01 (and Nate!) – a couple of comments to reduce code

apps/hash-frontend/src/shared/layout/plain-layout.tsx
benwerner01: That makes sense, have moved it in https://github.com/hashintel/hash/pull/2581/commits/3da971ad5b2fe8fee46b9ac3f01bfa3fd0fbc24b

apps/hash-frontend/src/shared/command-bar.tsx
benwerner01: Sure, have moved it into the hook in https://github.com/hashintel/hash/pull/2581/commits/c6a68eb58ada6d69197a664263517ce1f2bb5b15

CiaranMn (dismissed):

nonparibus (approved):

PR #2582, opened 2023-05-18
Create external entity types and property types on demand

TimDiekmann:

The Graph API is now able to directly load types from the type fetcher. The expected behavior at GQL is when querying an type it should be fetched if it's external and does not exist already.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204596705932069/f) _(internal)_

## 🚫 Blocked by

- #2576 

## 🔍 What does this change?

- Add function to `hash-api` to handle the logic
- Use those functions in GQL

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Use these changes in the `hash-api` to insert external types if they don't exist

## 🛡 What tests cover this?

- A test for each ontology type has been added to test the behavior


apps/hash-api/src/graph/ontology/primitive/data-type.ts
```diff
@@ -120,6 +122,48 @@ export const getDataTypeById: ImpureGraphFunction<
   return dataType;
 };
 
+/**
+ * Get an data type rooted subgraph by its versioned URL and if it's not available in the Graph and is an external type it will load this type into the Graph.
```
Alfred-Mountfield, line 126:
```suggestion
 * Get a data type rooted subgraph by its versioned URL.
 *
 * If the type does not already exist within the Graph, and is an externally-hosted type, this will also load the type into the Graph.
```

Perhaps something like this for the docs on these

apps/hash-api/src/graph/ontology/primitive/property-type.ts
```diff
@@ -121,6 +123,50 @@ export const getPropertyTypeById: ImpureGraphFunction<
   return propertyType;
 };
 
+/**
+ * Get an property type rooted subgraph by its versioned URL and if it's not available in the Graph and is an external type it will load this type into the Graph.
```
Alfred-Mountfield, line 127: See suggestion above

Alfred-Mountfield (approved):

PR #2588, opened 2023-05-20
Add `Message` struct and pave the path for new object approach

TimDiekmann:

@indietyp and I agreed offline that the first priority will be to add all available structs as a POD. We will build a builder pattern around these at a later point when proper validation can happen as many structs (i.e. `Message`) require other objects for true validation (e.g. if `id` instead of `text` is set).

This PR starts by adjusting the `MultiformatMessageString` and adding the `Message` struct.

## 🔗 Related links

- [Specs for `Message`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html#_Toc34317459)
- [Specs for `MultiformatMessageString`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html#_Toc34317473)

## 🔍 What does this change?

- Add a `serde` helper to allow true optional values (fail on `null`, only allow absent values or the inner value)
- Adjust documentation and remove builder for `MultiformatMessageString`
- Add very extensive documentation (from the specs) for `Message` and the `Message` struct itself.

This PR:

## ⚠️ Known issues

With the building pattern being removed the code coverage may drop below 100%. This is expected.

## 🐾 Next steps

Adjust other existing structs and add missing structs

## 🛡 What tests cover this?

None

## ❓ How to test this?

Read the documentation using `just doc --open`

### DIFF
--- a/libs/sarif/src/schema/message.rs
+++ b/libs/sarif/src/schema/message.rs
@@ -1,0 +1,482 @@
+#![expect(
+    rustdoc::broken_intra_doc_links,
+    reason = "Some structs are not implemented yet and will be added in a follow-up PR"
+)]
+
+use alloc::{borrow::Cow, vec::Vec};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::schema::PropertyBag;
+
+/// Encapsulates a message intended to be read by the end user ([§3.11]).
+///
+///
+/// # General
+///
+/// Certain objects define messages intended to be viewed by a user. SARIF represents such a message
+/// with a message object, which offers the following features:
+///
+/// - Message strings in plain text (“[plain text messages]”) ([§3.11.3]).
+/// - Message strings that incorporate formatting information (“[formatted messages]”) in [GitHub
+///   Flavored Markdown][GFM] ([§3.11.4]).
+/// - Message strings with [placeholders] for variable information ([§3.11.5]).
+/// - Message strings with [embedded links] ([§3.11.6]).
+///
+///
+/// # Constraints
+///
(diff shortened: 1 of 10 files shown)


libs/sarif/src/schema/message.rs
```diff
@@ -0,0 +1,482 @@
...
+/// messages **shall not** attempt to render them and **shall** instead fall back to the
+/// corresponding plain text messages.
+///
+///
+/// # Messages with placeholders
+/// [placeholders]: #messages-with-placeholders
+///
+/// A message string **may** include one or more “placeholders". The syntax of a placeholder is:
+///
+/// ```text
```
indietyp, line 109:
```suggestion
/// ```ebnf
```

indietyp: SARIF uses ISO EBNF (https://www.iso.org/standard/26153.html) in contrast to RFC ABNF (https://datatracker.ietf.org/doc/html/rfc5234)

libs/sarif/src/schema/message.rs
```diff
@@ -0,0 +1,482 @@
...
+/// >
+/// > The GFM link syntax is very flexible. Since a SARIF viewer that renders formatted messages
+/// > will presumably rely on a full-featured GFM processor, there is no need to restrict the
+/// > embedded link syntax in SARIF formatted messages.
+///
+/// Within a [plain text message][plain text messages] ([§3.11.3]), an embedded link **shall**
+/// conform to the following syntax (which is a greatly restricted subset of the GFM link syntax)
+/// before JSON encoding:
+///
+/// ```text
```
indietyp, line 170:
```suggestion
/// ```ebnf
```

indietyp (changes requested):
Otherwise great stuff, absolutely in love with the documentation!

indietyp (approved):

PR #2592, opened 2023-05-22
Add diagram to make test snapshot more readable

TimDiekmann:

It's hard to read a snapshot. This PR adds a small text file to draw the graph for the given snapshot.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204584047986189/f) _(internal)_

This PR:

The changes in this PR:

The changes in this PR:


tests/hash-backend-integration/src/tests/subgraph/pass/friendship.txt
thehabbos007, line 1: Thanks for this. I'm wondering if we can somehow reason about this in a more automated way in the future? As I understand it these diagrams are "manual" currently. Perhaps we could create a subgraph/snapshot -> dot visualization at some point to make this error prone :)

thehabbos007 (approved):
Thank you. Looks good to me. Cross checked jsonl and diagram :)

tests/hash-backend-integration/src/tests/subgraph/pass/friendship.txt
TimDiekmann: I thought about this, too, but haven't invested time yet. As soon as the snapshots get bigger it might be useful but I worry that the diagrams would also be less readable.

PR #2593, opened 2023-05-22
Add snapshot to BE tests with circular links

TimDiekmann:

For a more complex snapshot this introduces a snapshot which has a circular dependency and a double linked list.

## 🔗 Related links

## 🔍 What does this change?

- Add `.http` file to generate a graph
- Export snapshot and make it readable
- Create diagram to better understand the snapshot

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/tests/circular-links.http
thehabbos007, line 1: I thought we wanted to find a better testing layer for the HTTP. Are we content with httpyac for these since we're making them more structured now?

TimDiekmann: I have mainly put them here to have an easy way to create a snapshot. I don't want to really rely on this as a test case it's still the easiest way to quickly insert data into the graph.

tests/hash-backend-integration/src/tests/subgraph/circular.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import { resetGraph, restoreSnapshot } from "../test-server";
+
+jest.setTimeout(60000);
+
+afterAll(async () => {
+  await resetGraph();
+});
+
+describe("Circular Snapshot", () => {
+  it("can upload snapshot", async () => {
```
thehabbos007, line 12: Should we call this restore instead of upload?
```suggestion
  it("can restore snapshot", async () => {
```

apps/hash-graph/tests/circular-links.http
```diff
@@ -0,0 +1,308 @@
...
+POST http://127.0.0.1:4000/entities
+Content-Type: application/json
+Accept: application/json
+
+{
+  "ownedById": "{{account_id}}",
+  "actorId": "{{account_id}}",
+  "properties": {},
+  "entityTypeId": "http://localhost:3000/@snapshot/types/entity-type/object/v/1"
+}
```
thehabbos007, line 61: Just for my understanding, the `ciruclar.jsonl` snapshot has been manually cleaned up (the identifiers changed too?) Could we use `entity_uuid` as a param to alleviate some of the manual labor required after dumping the snapshot? 
Of course that means that we can't run the file multiple times, but that was the case already for `friendship.http`

thehabbos007 (dismissed):
Thank you, Tim! I have some comments, but they aren't strictly blocking merge

apps/hash-graph/tests/circular-links.http
TimDiekmann: I added the `entityUuid` parameter, good flag!

thehabbos007 (approved):

PR #2600, opened 2023-05-23
Split REST endpoints for loading external types

TimDiekmann:

We want to split the endpoints of the REST API to have a better distinguishment of loading external types and creating owned types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204643165960949/f) _(internal)_
- https://github.com/hashintel/hash/pull/2582
- https://github.com/hashintel/hash/pull/2576

## 🔍 What does this change?

- Split the endpoints
- Generate OpenAPI specs
- Adjust `hash-api` to use new endpoints

## 🚫 Blockers

The CI is currently failing as the `load_external_X_type` is not available when compiling without the type fetcher. It could be solved by duplicating the `OpenAPI` attributes or making the body conditionally, but neither solution is satisfying. See the [relevant Slack thread](https://hashintel.slack.com/archives/C03F7V6DU9M/p1684409097217459) _(internal)_.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Creating errors using the `status` library is very unergonomic and will be improved soon. This PR copies the existing code to avoid sinking too much time into this as this will be changed in the near future.


Alfred-Mountfield (dismissed):

apps/hash-graph/lib/graph/src/api/rest/data_type.rs
```diff
@@ -200,6 +172,58 @@ where
...
+) -> Result<Json<OntologyElementMetadata>, StatusCode>
+where
+    for<'pool> P::Store<'pool>: RestApiStore,
+{
+    let mut store = pool.acquire().await.map_err(|report| {
+        tracing::error!(error=?report, "Could not acquire store");
+        StatusCode::INTERNAL_SERVER_ERROR
+    })?;
+
+    #[allow(clippy::infallible_destructuring_match)]
```
Alfred-Mountfield, line 210: I've seen this before, do you know why it's complaining? I don't see a `match`, also do we perhaps want to use `expect` instead?

TimDiekmann: Previously, with the type fetcher being a compile-time option, the second arm of the match was was compiled out. As this lint only happened without the type fetcher enabled I couldn't use `expect` because otherwise it would complain if the type fetcher is enabled.

That I still had it here was a copy-paste error, but as the type fetcher is not compiled conditionally it's not neither regardless. Thanks for spotting! [`f58d430` (#2600)](https://github.com/hashintel/hash/pull/2600/commits/f58d43023bc81dffb621dd39b8e2c0fd98169f4e)

PR #2604, opened 2023-05-25
Return an empty subgraph if ontology type queried by its ID does not exist

TimDiekmann:

I accidentally changed the behavior of the GQL resolvers to error if a type does not exist.

## 🔗 Related links

- [Slack thread 1](https://hashintel.slack.com/archives/C022217GAHF/p1684912726686829)
- [Slack thread 2](https://hashintel.slack.com/archives/C022217GAHF/p1685010273602719)


nathggns (approved):
Checked it works and it does – thanks @TimDiekmann!

PR #2607, opened 2023-05-25
H-85: Auto-increment slugs when type names clash

nathggns:

This was an aborted attempt at adding auto suffixing with an incrementing number when types clashed. The only relevant commit is the last one. Leaving here as an example of how to do it.


CiaranMn:
Closing this, can pick work from it as needed in future – thank you @nathggns!

PR #2608, opened 2023-05-26
Use Vault for HASH AWS credentials

thehabbos007:

This PR switches from our secrets-based hardcoded AWS credentials to using our vault instance to get temporary credentials for pushing images.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1204675418698206/f) _(internal)_

## 🔍 What does this change?

- Changes actions and workflows related to pushing to ECR
- Driveby todo fixes

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Migrate to STS based credentials in terraform
- Automate Terraform plan/apply

## 🛡 What tests cover this?

- CI/CD

## ❓ How to test this?

Hard to test manually, This run contains the result from this change https://github.com/hashintel/hash/actions/runs/5089339009 (no effective change)

### DIFF
--- a/.github/workflows/hash-backend-cd.yml
+++ b/.github/workflows/hash-backend-cd.yml
@@ -5,14 +5,12 @@ on:
     branches:
       - main
     paths:
+      - "apps/hash-graph/**"
       - "apps/hash-api/**"
       - "apps/hash-external-services/kratos/**"
       - "libs/@local/hash-backend-utils-utils/**"
       - "libs/@local/hash-isomorphic-utils/**"
       - "libs/@local/status/**"
-      # TODO: remove/rewrite these as we remove the package directory.
-      - "apps/hash-graph/**"
-      - "apps/hash/**"
 
 env:
   VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
--- a/.github/workflows/hash-backend-cd.yml
+++ b/.github/workflows/hash-backend-cd.yml
@@ -15,8 +15,7 @@ on:
       - "apps/hash/**"
 
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/hash-backend-cd.yml
```diff
@@ -5,18 +5,15 @@ on:
     branches:
       - main
     paths:
+      - "apps/hash-graph/**"
```
TimDiekmann, line 8: Do we want to sort them by their name?

.github/workflows/hash-backend-cd.yml
```diff
@@ -45,9 +42,27 @@ jobs:
   build-graph:
     name: Build and push HASH graph image
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
TimDiekmann, line 47: We should probably add `permissions` to all actions at some point.

TimDiekmann (approved):
CHanges look good to me and seem to work, thank you!

Regarding the failing dependency check: We spoke offline about it and agreed that it's safe to use `v2` as `v2` points to the `v2.5.0` tag and the issue was [fixed in version 2.2](https://github.com/hashicorp/vault-action/blob/v2/CHANGELOG.md#220-may-6th-2021)

.github/workflows/hash-backend-cd.yml
thehabbos007: Probably yes, but I started out just grouping them by their path at least

.github/workflows/hash-backend-cd.yml
thehabbos007: Yes we should. These are necessary specifically for the id-token vault auth though

PR #2609, opened 2023-05-26
Automate HASH infrastructure deployment

thehabbos007:

This PR automates deployment of the HASH infrastructure. 
Pieces of this PR is quite repetitive and can be consolidated (also similar to our internal API deployment infra.)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1204675418698198/f) _(internal)_

## 🔍 What does this change?

- Moves to STS based Vault user generation in Terraform
- Adds two workflows for apply/plan of the terraform script

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Consolidate our scripts together (both GHA and Terraform)
- Perhaps make apply part of the CD workflow we have

## 🛡 What tests cover this?

- CI/CD

## ❓ How to test this?

It's hard to test this locally, although you should be able to see https://github.com/hashintel/hash/actions/runs/5092036652?pr=2609 succeed

### DIFF
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -33,7 +33,7 @@ jobs:
           DISPATCH_ENV: ${{ github.event.inputs.env }}
         run: |
           # Only apply the prod environment on merge to main or manual dispatch
-          if [ "$DISPATCH_ENV" == "prod" || GITHUB_REF_NAME == "main" ]; then
+          if [ "$DISPATCH_ENV" == "prod" || "$GITHUB_REF_NAME" == "main" ]; then
             echo "matrix=[\"prod\"]" >> $GITHUB_OUTPUT
           else
             echo "matrix=[]" >> $GITHUB_OUTPUT
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -93,18 +93,18 @@ jobs:
         env:
           TF_WORKSPACE: ${{ matrix.env }}
         working-directory: ${{ env.working-directory }}
-        run: terraform init -no-color
+        run: terraform init -no-color 2>&1
 
       - name: Terraform Validate
         id: validate
-        run: terraform validate -no-color
+        run: terraform validate -no-color 2>&1
 
       - name: "[${{ matrix.env }}] Terraform Apply"
         id: apply
         env:
           TF_WORKSPACE: ${{ matrix.env }}
         working-directory: ${{ env.working-directory }}
-        run: terraform apply -auto-approve -no-color -var-file=${{ matrix.env }}-usea1.tfvars -refresh=false
+        run: terraform apply -auto-approve -no-color -var-file=${{ matrix.env }}-usea1.tfvars -refresh=false 2>&1
 
       - uses: actions/github-script@v6
         id: set-result
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -30,13 +30,13 @@ jobs:
       - name: Set matrix
         id: set-matrix
         env:
-          ENV: ${{ github.event.inputs.env }}
+          DISPATCH_ENV: ${{ github.event.inputs.env }}
         run: |
-          # If no env is specified, run both dev and prod
-          if [ "$ENV" == "prod" ]; then
-            echo "matrix=[\"$ENV\"]" >> $GITHUB_OUTPUT
-          else
+          # Only apply the prod environment on merge to main or manual dispatch
+          if [ "$DISPATCH_ENV" == "prod" || GITHUB_REF_NAME == "main" ]; then
             echo "matrix=[\"prod\"]" >> $GITHUB_OUTPUT
+          else
+            echo "matrix=[]" >> $GITHUB_OUTPUT
           fi
 
   terraform:
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -60,6 +60,11 @@ jobs:
       - uses: actions/checkout@v3
       - uses: hashicorp/setup-terraform@v2
 
+      - name: Prepare SSH key
+        working-directory: ${{ env.working-directory }}
+        run: |
+          echo "${{ secrets.BASTION_SSH_KEY }}" > h-hash-prod-usea1-bastionkey.pem
+
       - name: Authenticate Vault
         id: secrets
         uses: hashicorp/vault-action@v2
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -98,7 +98,7 @@ jobs:
         id: validate
         run: terraform validate -no-color
 
-      - name: "[${{ matrix.env }}] Terraform Plan"
+      - name: "[${{ matrix.env }}] Terraform Apply"
         id: apply
         env:
           TF_WORKSPACE: ${{ matrix.env }}
@@ -123,7 +123,7 @@ jobs:
                 "Validation Output 🤖 (${{ steps.validate.outcome }})",
                 `<pre><code>${{ steps.validate.outputs.stdout }}</code></pre>`
               )
-              .addDetails("Plan Output 📖 (${{ steps.apply.outcome }})", 
+              .addDetails("Apply Output 📖 (${{ steps.apply.outcome }})", 
                 `<pre><code lang="diff">${apply}</code></pre>`
               )
               .write();
--- a/.github/workflows/tf-apply-hash.yml
+++ b/.github/workflows/tf-apply-hash.yml
@@ -1,0 +1,129 @@
+name: HASH Terraform Apply
+
+# TODO: While this workflow is specific to HASH, we should consider
+#  making this a reusable workflow, parameterized on the directory and environment.
+
+on:
+  workflow_dispatch:
+    inputs:
+      env:
+        type: choice
+        options: ["prod"]
+        description: "Environment to deploy"
+        required: true
+        default: "prod"
+
+  push:
+    branches:
+      - main
+    paths:
+      - "infra/internal-api/**"
+
+env:
+  VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
+
+jobs:
+  matrix-eval:
+    name: Prepare environments to be applied
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Set matrix
+        id: set-matrix
+        env:
+          ENV: ${{ github.event.inputs.env }}
+        run: |
+          # If no env is specified, run both dev and prod
+          if [ "$ENV" == "prod" ]; then
+            echo "matrix=[\"$ENV\"]" >> $GITHUB_OUTPUT
+          else
+            echo "matrix=[\"prod\"]" >> $GITHUB_OUTPUT
+          fi
+
+  terraform:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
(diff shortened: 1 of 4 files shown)


.github/workflows/tf-apply-hash.yml
```diff
@@ -0,0 +1,129 @@
...
+
+# TODO: While this workflow is specific to HASH, we should consider
+#  making this a reusable workflow, parameterized on the directory and environment.
+
+on:
+  workflow_dispatch:
+    inputs:
+      env:
+        type: choice
+        options: ["prod"]
```
TimDiekmann, line 11: What is the reason we don't allow dispatching `dev`?

.github/workflows/tf-apply-hash.yml
```diff
@@ -0,0 +1,129 @@
...
+          fi
+
+  terraform:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      working-directory: infra/terraform/hash
+      TF_VAR_in_ci: true
```
TimDiekmann, line 51: Is this only to work around parsing environment variables from Terraform by e.g. calling a bash script that outputs the environment as JSON?

.github/workflows/tf-apply-hash.yml
```diff
@@ -0,0 +1,129 @@
...
+        env:
+          TF_WORKSPACE: ${{ matrix.env }}
+        working-directory: ${{ env.working-directory }}
+        run: terraform apply -auto-approve -no-color -var-file=${{ matrix.env }}-usea1.tfvars -refresh=false
+
+      - uses: actions/github-script@v6
+        id: set-result
+        with:
+          script: |
+            const apply = `${{ steps.apply.outputs.stdout }}`.replaceAll(/^~/g, "!");
```
TimDiekmann, line 112: Does GitHub CI have the concept of `stdout` and `stderr`? I wonder if `stderr` output would be omitted.

.github/workflows/tf-apply-hash.yml
```diff
@@ -0,0 +1,129 @@
...
+        options: ["prod"]
+        description: "Environment to deploy"
+        required: true
+        default: "prod"
+
+  push:
+    branches:
+      - main
+    paths:
+      - "infra/internal-api/**"
```
TimDiekmann, line 20: I think these needs adjustment
```suggestion
      - "infra/terraform/hash/**"
      - "infra/terraform/modules/**"
```

.github/workflows/tf-plan-hash.yml
TimDiekmann, line 1: The plan and apply workflows are very similar. Is it possible to put this into one file and use a variable in the matrix to determine if it's a "plan" or an "apply" run?

.github/workflows/tf-plan-hash.yml
```diff
@@ -0,0 +1,130 @@
...
+        description: "Environment to plan"
+        required: true
+        default: "prod"
+
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "infra/terraform/hash/**"
+      - "infra/terraform/modules/**"
```
TimDiekmann, line 21: These will prevent making this a required job. I think we want to make it required at some point?

.github/workflows/tf-plan-hash.yml
```diff
@@ -0,0 +1,130 @@
...
+      env:
+        type: choice
+        options: ["prod"]
+        description: "Environment to plan"
+        required: true
+        default: "prod"
+
+  pull_request:
+    branches:
+      - main
```
TimDiekmann, line 18: Why do we only run it against main?

TimDiekmann (commented):
Thanks, @thehabbos007, the PR looks good so far. I have a couple of comments that I like to resolve before approving.

.github/workflows/tf-apply-hash.yml
```diff
@@ -0,0 +1,129 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Set matrix
+        id: set-matrix
+        env:
+          ENV: ${{ github.event.inputs.env }}
+        run: |
+          # If no env is specified, run both dev and prod
```
CiaranMn, line 37: Either this comment needs correcting or the matrix below does

.github/workflows/tf-plan-hash.yml
```diff
@@ -0,0 +1,130 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Set matrix
+        id: set-matrix
+        env:
+          ENV: ${{ github.event.inputs.env }}
+        run: |
+          # If no env is specified, run both dev and prod
```
CiaranMn, line 38: Either this comment needs correcting or the matrix below does

.github/workflows/tf-plan-hash.yml
```diff
@@ -0,0 +1,130 @@
...
+          if [ "$ENV" == "prod" ]; then
+            echo "matrix=[\"$ENV\"]" >> $GITHUB_OUTPUT
+          else
+            echo "matrix=[\"prod\"]" >> $GITHUB_OUTPUT
```
CiaranMn, line 42: I'm wondering if it's wise to run the production deployment if no environment is specified (regardless of if we also run the dev or not if the variable is missing) – it might be better to force people to be explicit about deploying to prod if that's their intention, to avoid unintended deployments. Or is there some technical reason we need to be able to deploy to prod if the environment variable isn't specified?

CiaranMn: I guess this comment is better applied to the 'apply' file but you get the point

.github/workflows/tf-apply-hash.yml
thehabbos007: We don't have a prod HASH environment :)

TimDiekmann: I assume you mean `dev` but then the comment below in `matrix-eval` needs to be adjusted 🙂

.github/workflows/tf-apply-hash.yml
TimDiekmann: I mean the way how we pass in the `is_ci` environment variable. The `CI` variable is set by GH Actions automatically, I wonder if we want to rely on that instead of manually setting it.

.github/workflows/tf-apply-hash.yml
thehabbos007: You're right, thanks for catching!

.github/workflows/tf-apply-hash.yml
thehabbos007: Yes sorry a dev env. I will adjust the comments!

.github/workflows/tf-apply-hash.yml
TimDiekmann: There is a small risk of a data race condition and the two outputs will be mixed up, generally, this is fine I think. In the (probably unlikely case) this will happen the GH Action output will still have everything in the correct order.

.github/workflows/tf-apply-hash.yml
thehabbos007: Yeah i was looking into that and having TF look for the `CI` env var, which may be possible. I don't think we can do `TF_VAR_in_ci: ${{ env.CI }}` as this is defined for the job and the `env` isn't set at this point as far as I can tell

.github/workflows/tf-plan-hash.yml
thehabbos007: Yeah you're right. I'll try to look into this in https://app.asana.com/0/1204610208832639/1204691860848733/f
as this may be a reusable workflow/action that both the internal API and HASH uses

.github/workflows/tf-plan-hash.yml
thehabbos007: For `pull_request` this means that it only runs on PRs targetting `main`

.github/workflows/tf-plan-hash.yml
thehabbos007: For the `apply` workflow, it wouldn't have any `$ENV` unless it's executed through a `workflow_dispatch`, so we need the default to do merged to main. 

When we dispatch the workflow you have to choose your env, and if we did have a `dev` env we could default to that (we default to `dev` for the internal API apply workflow dispatch)

.github/workflows/tf-plan-hash.yml
TimDiekmann: I mentioned a workaround in [this slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1684844804416769?thread_ts=1684844296.022789&cid=C02TWBTT3ED): It's possible by either having an empty job running or by using the `if:` conditions: [Handling skipped but required checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks). A filtered out path will result in the `Pending` state, while `if: skipped` will result in a `Success` state.

.github/workflows/tf-apply-hash.yml
thehabbos007: I suppose that could work, I just didn't want to introduce a `data` shell script like that. What if I look into doing this in the next-up asana task https://app.asana.com/0/1204610208832639/1204691860848733/f
and update
https://github.com/hashintel/hash/blob/70cd4979dbcae33947e7172f739a36641396cea5/infra/terraform/modules/variables/main.tf
to do that CI check?

TimDiekmann: I'm fine with deferring it or not using it at all in this case, just wanted to check if we evaluated all possible solutions 🙂

.github/workflows/tf-plan-hash.yml
CiaranMn: We could maybe read from a [GitHub environment variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) to determine if this is a merge to main (e.g. `GITHUB_REF_NAME == "main"`) rather than having it be the default

thehabbos007: I've addressed this in f21e71353927c3f6af984d7b4111acf9fd9d3db7

.github/workflows/tf-apply-hash.yml
thehabbos007: I've removed the conditional as it is superfluous 3e654bb45e31fd64c490ea1fe30ff6ed005a3186

.github/workflows/tf-plan-hash.yml
thehabbos007: Addressed in f21e71353927c3f6af984d7b4111acf9fd9d3db7

.github/workflows/tf-apply-hash.yml
thehabbos007: Addressed in bb59888e13bbcb5d5536c9768b144cdc9b073b10

.github/workflows/tf-plan-hash.yml
thehabbos007: Removed the path condition in 9d8175a35d867d98359137e6d3f8636d68a4a1a3 as it wasn't really worth the hassle to do path matching when we can just run the (quick) workflow

.github/workflows/tf-plan-hash.yml
thehabbos007: Removed in ef00b5449bb9f2b455fd2c2f2065afacef697c0c, thanks :)

.github/workflows/tf-plan-hash.yml
TimDiekmann: I think that's fair, thank you!

TimDiekmann (approved):

PR #2612, opened 2023-05-29
Move reading vertices until after traversal is finished

TimDiekmann:

The traversal logic is currently in the process of being optimized. The first goal in order to make optimization possible the reading of vertices is going to be deferred to the end. With this, it's possible to only read the edge information while traversing instead of reading the whole record.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1200211978612931/1204117845662853/f) _(internal)_

## 🔍 What does this change?

- Store entity edition ids and ontology vertex ids in `TraversalContext`
- Create a new traversal context when starting a traversal
- Read vertices after traversing is finished

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- This will currently double the amount of queries being done to the graph as each record is read one by one. This will be addressed in a direct follow-up.

## 🐾 Next steps

As mentioned above, the direct follow-up will add functionality to read all records at once. Further follow-ups will also enable batched reading of edges, which is more complicated as this requires either a manual SQL statement or decent changes to the structural queries. 

## 🛡 What tests cover this?

- Directly:
- Indirectly:

## ❓ How to test this?

- Make sure the subgraph returned is the same as before

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/traversal_context.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/traversal_context.rs
@@ -1,0 +1,188 @@
+use std::collections::HashSet;
+
+use error_stack::Result;
+
+use crate::{
+    identifier::knowledge::EntityEditionId,
+    knowledge::{Entity, EntityQueryPath},
+    ontology::{DataTypeWithMetadata, EntityTypeWithMetadata, PropertyTypeWithMetadata},
+    store::{
+        crud::Read,
+        query::{Filter, FilterExpression, Parameter},
+        AsClient, PostgresStore, QueryError, Record,
+    },
+    subgraph::{
+        identifier::{DataTypeVertexId, EntityTypeVertexId, PropertyTypeVertexId},
+        Subgraph,
+    },
+};
+
+impl<C: AsClient> PostgresStore<C> {
+    async fn read_data_types_by_ids(
+        &self,
+        vertex_ids: impl IntoIterator<Item = &DataTypeVertexId, IntoIter: Send> + Send,
+        subgraph: &mut Subgraph,
+    ) -> Result<(), QueryError> {
+        for vertex_id in vertex_ids {
+            for data_type in <Self as Read<DataTypeWithMetadata>>::read_vec(
+                self,
+                &DataTypeWithMetadata::create_filter_for_vertex_id(vertex_id),
+                Some(&subgraph.temporal_axes.resolved),
+            )
+            .await?
+            {
+                subgraph.insert_vertex(vertex_id, data_type);
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn read_property_types_by_ids(
+        &self,
+        vertex_ids: impl IntoIterator<Item = &PropertyTypeVertexId, IntoIter: Send> + Send,
+        subgraph: &mut Subgraph,
+    ) -> Result<(), QueryError> {
+        for vertex_id in vertex_ids {
+            for property_type in <Self as Read<PropertyTypeWithMetadata>>::read_vec(
+                self,
+                &PropertyTypeWithMetadata::create_filter_for_vertex_id(vertex_id),
+                Some(&subgraph.temporal_axes.resolved),
+            )
+            .await?
+            {
+                subgraph.insert_vertex(vertex_id, property_type);
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn read_entity_types_by_ids(
+        &self,
+        vertex_ids: impl IntoIterator<Item = &EntityTypeVertexId, IntoIter: Send> + Send,
+        subgraph: &mut Subgraph,
+    ) -> Result<(), QueryError> {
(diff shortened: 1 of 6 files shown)


apps/hash-graph/lib/graph/src/store/postgres/traversal_context.rs
```diff
@@ -0,0 +1,188 @@
...
+        for vertex_id in vertex_ids {
+            for enttity_type in <Self as Read<EntityTypeWithMetadata>>::read_vec(
+                self,
+                &EntityTypeWithMetadata::create_filter_for_vertex_id(vertex_id),
+                Some(&subgraph.temporal_axes.resolved),
+            )
+            .await?
+            {
+                subgraph.insert_vertex(vertex_id, enttity_type);
+            }
```
Alfred-Mountfield, line 75:
```suggestion
        for vertex_id in vertex_ids {
            for entity_type in <Self as Read<EntityTypeWithMetadata>>::read_vec(
                self,
                &EntityTypeWithMetadata::create_filter_for_vertex_id(vertex_id),
                Some(&subgraph.temporal_axes.resolved),
            )
            .await?
            {
                subgraph.insert_vertex(vertex_id, entity_type);
            }
```

apps/hash-graph/lib/graph/src/store/postgres/traversal_context.rs
```diff
@@ -0,0 +1,188 @@
...
+#[derive(Debug, Default)]
+pub struct TraversalContext {
+    data_types: OntologyTraversalContext<DataTypeWithMetadata>,
+    property_types: OntologyTraversalContext<PropertyTypeWithMetadata>,
+    entity_types: OntologyTraversalContext<EntityTypeWithMetadata>,
+    entities: EntityTraversalContext,
+}
+
+impl TraversalContext {
+    pub async fn load_vertices<C: AsClient>(
```
Alfred-Mountfield, line 152: Can we think of another name here? I'm not entirely sure I love `load_` but I don't have a good suggestion right now

TimDiekmann: What do you think about `read_traversed_vertices`?

TimDiekmann: [`867e3c4` (#2612)](https://github.com/hashintel/hash/pull/2612/commits/867e3c44e9aa0b7a494a5b9a296b983bf04fa88f)

Alfred-Mountfield: Are the vertices traversed yet? It sounds like they're candidates to be traversed, or in the queue, but haven't been until this function has completed. I'd also probably opt to just use normal nomenclature like `visit_x`

TimDiekmann: Yes, the traversal context is populated when traversing the graph. `visit_x` sounds like you're going to traverse until `x`.

Alfred-Mountfield: I don't think that's true.. Let's say I have a super basic graph which is just a linked list `a -> b -> c`

If I'm doing a BFS/DFS (somewhat unimportant right now so ignore the post-order/pre-order whatever)

Then 
- I _visit_ `a`, and I look at its neighbors (`b`). I add `b` to the queue (note I have not _traversed_ `b`), I do whatever, and then
- I _visit_ `b`, and I look at its neighbors (`c`)...
- I _visit_ `c`...

In traversal you visit something when you read it, not when you add it to the queue. I don't think I've ever heard `visit_x` as being traversing _until_ `x`, is that how it's used in libraries like `serde`? 🤔 

(See ChatGPT's description for more support)
image

Alfred-Mountfield: Ah I'm realising now that this operation doesn't actually add anything to the queue, it merely reads the records (which we don't need to do _during_ traversal?)

Alfred-Mountfield: Assuming we're not going to extend this to add more things to the explore 'queue', I don't feel strongly about the naming of this, happy to approve

Alfred-Mountfield (approved):

apps/hash-graph/lib/graph/src/store/postgres/traversal_context.rs
TimDiekmann: Yes, we don't need to read the records during the traversal. The purpose of this PR is to defer the reading of the vertices until after traversal so it's only required to look at the edges during traversal.

Later this week I'm going to change the `HashSet` to track which vertices were already visited so visiting the same vertex can be avoided (given that the resolve depths and time intervals are matching). The context, however, will only be used as the data which has been traversed, it will never hold any records of edges or vertices, which are not yet traversed so to directly answer your assumption: this will never be an "explore queue" but always an "explored set" 🙂

PR #2615, opened 2023-05-29
Support batched reading for ontology-ontology edges

TimDiekmann:

This is the first PR which actively improves performance on the traversal logic. This starts by batching together edge lookups for ontology-to-ontology edges based on the edge kind.

Note, that I have not precisely measured the performance improvement. In general the number of queries is reduced which directly impacts performance. A summary will be posted in a follow-up PR.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204117847656663/f) _(internal)_

## 🚫 Blocked by

- #2612
- #2614

## 🔍 What does this change?

- Add function to read ontology-to-ontology edges based on a reference table
- Use that function to read as many ontology edges withing the same Query as possible

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Do the same for
- knowledge -> ontology (different due to temporal axes)
- kowledge -> knowledge (complicated due to temporal axes)

## 🛡 What tests cover this?

- Directly:
- Indirectly:

## ❓ How to test this?

Make sure the subgraph looks exactly as before

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/ontology/property_type.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/ontology/property_type.rs
@@ -1,23 +1,26 @@
-use std::{borrow::Borrow, mem};
+use std::{borrow::Borrow, collections::HashMap};
 
 use async_trait::async_trait;
 use error_stack::{IntoReport, Result, ResultExt};
-use type_system::PropertyType;
+use type_system::{url::VersionedUrl, PropertyType};
 
 use crate::{
-    ontology::{DataTypeWithMetadata, OntologyElementMetadata, PropertyTypeWithMetadata},
+    ontology::{OntologyElementMetadata, PropertyTypeWithMetadata},
     provenance::RecordCreatedById,
     store::{
         crud::Read,
         error::DeletionError,
-        postgres::{ontology::OntologyId, TraversalContext},
-        query::Filter,
+        postgres::{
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/store/postgres/ontology/property_type.rs
```diff
@@ -35,87 +38,82 @@ impl<C: AsClient> PostgresStore<C> {
...
-                            EdgeDirection::Incoming,
-                        ),
-                        Some(&temporal_axes),
+            if let Some(traversal_data) =
+                edges_to_traverse.get(&OntologyEdgeKind::ConstrainsPropertiesOn)
+            {
+                property_type_queue.extend(
+                    self.read_ontology_edges::<PropertyTypeVertexId, PropertyTypeVertexId>(
+                        traversal_data,
+                        ReferenceTable::PropertyTypeConstrainsValuesOn,
```
thehabbos007, line 100: Is this the wrong ReferenceTable?
```suggestion
                        ReferenceTable::PropertyTypeConstrainsPropertiesOn,
```

TimDiekmann: Oh, good catch. I just noticed that we never use that edge kind in tests.

thehabbos007 (approved):

PR #2625, opened 2023-05-30
Support batched reading for entity->ontology edges

TimDiekmann:

Similar to #2615 this will add the batched traversal for knowledge-to-ontology edges.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204117847656663/f) _(internal)_

## 🚫 Blocked by

- #2612
- #2614
- #2615

## 🔍 What does this change?

- Add function to read knowledge-to-ontology edges based on a reference table
- Use that function to read as many shared edges within the same Query as possible

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Do the same for
- kowledge -> knowledge

## 🛡 What tests cover this?

- Directly:
- Indirectly:

## ❓ How to test this?

Make sure the subgraph looks exactly as before


apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -190,3 +192,123 @@ impl<C: AsClient> crud::Read<Entity> for PostgresStore<C> {
...
+    ) {
+        self.owned_by_ids.push(vertex_id.base_id.owned_by_id);
+        self.entity_uuids.push(vertex_id.base_id.entity_uuid);
+        self.entity_revision_ids.push(vertex_id.revision_id);
+        self.intervals.push(interval);
+        self.resolve_depths.push(resolve_depth);
+    }
+}
+
+pub struct SharedEdgeTraversal {
```
thehabbos007, line 234: Could it be useful to have the direction of this shared edge as part of the name? It's not immediately clear that it's entity -> ontology when reading the name :)

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -190,3 +192,123 @@ impl<C: AsClient> crud::Read<Entity> for PostgresStore<C> {
...
+            .client
+            .as_client()
+            .query(
+                &format!(
+                    r#"
+                        SELECT
+                             filter.idx,
+                             ontology_ids.base_url,
+                             ontology_ids.version,
+                             filter.interval * source.{variable_axis}
```
thehabbos007, line 261: What is the `*` operation here?

thehabbos007 (commented):
Thank you, Tim. This looks good to me. Just have a comment & question :)

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: That's [the intersection](https://www.postgresql.org/docs/15/functions-range.html) of two ranges

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: Yes, that's a good point. I'm struggling to find a name, do you have a suggestion?

thehabbos007: Not sure other than `EntityToOntologySharedEdge` which is quite verbose, and potentially used somewhere else

thehabbos007: Sure that's fine with me!

thehabbos007 (approved):

PR #2618, opened 2023-05-30
Refactor Terraform scripts to use shared modules

thehabbos007:
## 🌟 What is the purpose of this PR?

This PR refactors our HASH Terraform config to use reusable component we've defined in another repo (yet to be a public repo). The modules we've refactored here are ones that we're using in other Terraform configs and most general.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204610208832639/1204691860848733/f) _(internal)_ (TF part of the task)

## 🔍 What does this change?

- Refactors TF modules

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] is internal and does not require a docs change 

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] does not affect the execution graph

## 🐾 Next steps

- Remove `moved` clauses from the code after the `apply` has been carried out. This will make the statements obsolete.

## 🛡 What tests cover this?

- CI/CD

## ❓ How to test this?

1.  Checkout the branch
1.  With an authenticated Vault CLI, run `terraform init` and `terraform plan --var-file prod-usea1.tfvars  --refresh=false` from `/infra/terraform/hash`
1. The plan should be _unchanged_

PR #2630, opened 2023-05-31
Remove `moved` statements from refactored infra

thehabbos007:
As a follow up to #2618 this PR removes the `moved` blocks that were necessary for the initial refactor. These are now redundant since the TF apply was successful https://github.com/hashintel/hash/actions/runs/5132276830/jobs/9233363476 

Running `terraform plan --var-file prod-usea1.tfvars  --refresh=false` from the `/infra/terraform/hash` path locally should still result in zero changes.

### DIFF
--- a/infra/terraform/hash/networking/main.tf
+++ b/infra/terraform/hash/networking/main.tf
@@ -5,58 +5,6 @@ module "base_network" {
   region_az_names = var.region_az_names
 }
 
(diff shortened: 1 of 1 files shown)


TimDiekmann (approved):
Thank you!

> No changes. Your infrastructure matches the configuration.

PR #2636, opened 2023-06-02
Change the `EntityId` separator to `'~'`

TimDiekmann:

The % as separator causes problems as it has to be escaped in URLs. Instead, we chose to change the separator to a ~, which does not has to be escaped.

## 🔗 Related Links

- [Asana Task](https://app.asana.com/0/1204610208832639/1204723730564659/f) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C03TWBSRT8B/p1685630264122289) _(internal)_

This PR:

The changes in this PR:

The changes in this PR:


indietyp (dismissed):
Thank you! LGTM 🎉

Alfred-Mountfield (commented):
I think it looks good, and if CI passes I'm _fairly_ certain we shouldn't have missed any (at least important ones). 

Can we call it `delimiter` rather than `separator`? Just seems to be more idiomatic for this sort of thing, although it's not super important

Alfred-Mountfield (approved):

PR #2637, opened 2023-06-02
Fix roots not being added to the traversal context

TimDiekmann:

#2629 introduced a bug by not adding the ontology id roots to the traversal context. This fixes the issue

## 🔗 Related links

- #2629
- [Slack thread](https://hashintel.slack.com/archives/C022217GAHF/p1685715544856149) _(internal)_

## 🔍 What does this change?

- Add the roots to the traversal context

## ⚠️ Known issues

Sadly, there is no `TryStreamExt::try_flat_map` function, so I have to use `map_ok` followed by a `try_flatten`.

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/ontology/data_type.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/ontology/data_type.rs
@@ -1,8 +1,8 @@
 use std::borrow::Borrow;
 
 use async_trait::async_trait;
-use error_stack::{IntoReport, Result, ResultExt};
-use futures::TryStreamExt;
+use error_stack::{IntoReport, Report, Result, ResultExt};
+use futures::{stream, TryStreamExt};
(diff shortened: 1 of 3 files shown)


Alfred-Mountfield (commented):
Can we please add a couple of tests that would have caught this? Otherwise it looks good, thanks for fixing so quickly

TimDiekmann:
> Can we please add a couple of tests that would have caught this? Otherwise it looks good, thanks for fixing so quickly

I planned to do this as a follow-up but provide the fix ASAP

TimDiekmann:
I added tests in #2641. That PR is expected to fail until this PR is merged.

Alfred-Mountfield (approved):
Sorry on the delay, was trying to decide if we should merge #2641 into this branch, so that we could directly see it passing in CI. (Or if we should merge this branch into it).

I'm happy as is, thanks for the fix 🚀

PR #2638, opened 2023-06-05
Make the type-fetcher a runtime option

TimDiekmann:

Currently, the type-fetcher is enabled via the "type-fetcher" feature flag. The Graph behaves differently with the type-fetcher is enabled:
- data types and link entity type is seeded if the type-fetcher is disabled
- type-fetcher will (obviously) not run if it's disabled
- It's not possible to load external types on demand at HASH if it's disabled

The DX with the `type-fetcher` being a compile time feature is very bad (as seen in https://github.com/hashintel/hash/pull/2600). This changes the implementation to add an `--offline` flag to the `server` subcommand which loads a snapshot of the built-in types.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1204630029959178/f) _(internal)_

## 🔍 What does this change?

- Add an `--offline` CLI argument
- Add a snapshot to load if `--offline` is specified

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -94,207 +98,35 @@ pub struct ServerArgs {
...
+            FramedRead::new(
+                io::BufReader::new(Cursor::new(include_str!("../../built-in-types.jsonl"))),
+                codec::JsonLinesDecoder::default(),
+            ),
+            10_000,
+        )
+        .await;
+
+    if let Err(err) = restore_result {
+        if err.contains::<VersionedUrlAlreadyExists>() {
```
Alfred-Mountfield, line 119: What happens if the snapshot that's being used is updated. As in, let's say they change the timestamps, or update the types to point to newer versions.. I think it'll still _work_ fine, but might have some unexpected behaviour..

I don't think we want to try and work around that right now, but we might want to find a way to capture that possibility somewhere, in documentation of the option or something.

apps/hash-graph/bin/cli/built-in-types.jsonl
```diff
@@ -0,0 +1,8 @@
...
+{"type":"dataType","schema":{"$schema":"https://blockprotocol.org/types/modules/graph/0.3/schema/data-type","kind":"dataType","$id":"https://blockprotocol.org/@blockprotocol/types/data-type/number/v/1","title":"Number","description":"An arithmetical value (in the Real number system)","type":"number"},"metadata":{"recordId":{"baseUrl":"https://blockprotocol.org/@blockprotocol/types/data-type/number/","version":1},"custom":{"provenance":{"recordCreatedById":"00000000-0000-0000-0000-000000000000"},"temporalVersioning":{"transactionTime":{"start":{"kind":"inclusive","limit":"2023-06-02T17:20:56.196069000Z"},"end":{"kind":"unbounded"}}},"fetchedAt":"2023-06-02T17:20:56.193720000Z"}}}
```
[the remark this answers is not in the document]
TimDiekmann: Actually, I think the `fetchedAt` should precisely the time when they were fetched. If you look at the database compare them to the ontology types hosted at blockprotocol.org and see some inconsistency (for whatever reason) it's known that at the `fetchedAt` date the schema looked differently. Why it would cause strange behavior? That's the date of retrieval, not the date of when the types were uploaded to blockprotocol.org.

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -94,207 +98,35 @@ pub struct ServerArgs {
...
-            } else {
-                return Err(error.change_context(GraphError));
-            }
-        } else {
-            tracing::info!(%root_account_id, "inserted the primitive {title} data type");
+async fn restore_builtin_types<C: AsClient>(mut store: SnapshotStore<C>) -> Result<(), GraphError> {
+    let restore_result = store
+        .restore_snapshot(
+            FramedRead::new(
+                io::BufReader::new(Cursor::new(include_str!("../../built-in-types.jsonl"))),
```
Alfred-Mountfield, line 111: We would perhaps want to consider reusing this logic for generally seeding a dataset on graph startup? Perhaps users could provide their own snapshots or something

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -94,207 +98,35 @@ pub struct ServerArgs {
...
-    {
-        if error.contains::<VersionedUrlAlreadyExists>() {
-            tracing::info!(%root_account_id, "tried to insert {title} entity type, but it already exists");
-        } else {
-            return Err(error.change_context(GraphError));
-        }
-    } else {
-        tracing::info!(%root_account_id, "inserted the {title} entity type");
-    }
+    tracing::info!("restored built-in type");
```
Alfred-Mountfield, line 129:
```suggestion
    tracing::info!("restored built-in types");
```
?

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -94,207 +98,35 @@ pub struct ServerArgs {
...
-            .change_context(GraphError)
-        {
-            if error.contains::<VersionedUrlAlreadyExists>() {
-                tracing::info!(%root_account_id, "tried to insert primitive {title} data type, but it already exists");
-            } else {
-                return Err(error.change_context(GraphError));
-            }
-        } else {
-            tracing::info!(%root_account_id, "inserted the primitive {title} data type");
+async fn restore_builtin_types<C: AsClient>(mut store: SnapshotStore<C>) -> Result<(), GraphError> {
```
Alfred-Mountfield, line 107: `built_in` I think would be more correct

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -343,19 +175,29 @@ pub async fn server(args: ServerArgs) -> Result<(), GraphError> {
...
-            args.type_fetcher_address.type_fetcher_host,
-            args.type_fetcher_address.type_fetcher_port,
-        ),
-        DomainValidator::new(args.allowed_url_domain.clone()),
-    );
+        FetchingPool::new_offline(pool)
+    } else {
+        // `store` is bound to the lifetime of `pool`, which is moved into `FetchingPool`.
+        // The acquire is done regardless to check for a valid database connection.
+        drop(store);
```
Alfred-Mountfield, line 191: I don't quite understand why we drop the store here but we don't do that in the offline branch

apps/hash-graph/lib/graph/src/snapshot.rs
```diff
@@ -257,9 +260,9 @@ impl<C: AsClient> SnapshotStore<C> {
                         SqlState::FOREIGN_KEY_VIOLATION => {
                             report.attach_printable(StatusCode::NotFound)
                         }
-                        SqlState::UNIQUE_VIOLATION => {
-                            report.attach_printable(StatusCode::AlreadyExists)
-                        }
+                        SqlState::UNIQUE_VIOLATION => report
+                            .attach_printable(StatusCode::AlreadyExists)
+                            .attach(VersionedUrlAlreadyExists),
```
Alfred-Mountfield, line 265: Is this only for types? Can snapshots not return entities too?

Alfred-Mountfield (commented):
Looks mostly good to me :)

Have a few comments/questions though

apps/hash-graph/bin/cli/built-in-types.jsonl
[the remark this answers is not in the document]
Alfred-Mountfield: Hm this is a complicated topic, and one that depends on the semantics of `fetchedAt` and the temporal axes. Let's say I start my production instance of my application today, and I start to heavily utilise temporal versioning, creating dashboards over provenance, etc. If I _**know**_ my application's history starts today, then having records that were somehow 'fetched' before my application started is pretty... strange.

I hear your point about inconsistencies.. I mean, according to the spec the types _**must**_ be immutable, but there is always a possibility, and if this ends up used for more than the primitive types it becomes more problematic. I'd almost prefer if there wasn't a `fetchedAt` at all if they are built-in, but that'd introduce a tonne of unnecessary application complexity regarding the external ontology type shape, so let's not do that.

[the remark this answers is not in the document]
TimDiekmann: But if you would have started the application prior this date, the built-in types should be in the database already.

apps/hash-graph/bin/cli/src/subcommand/server.rs
```diff
@@ -94,207 +98,35 @@ pub struct ServerArgs {
...
+                io::BufReader::new(Cursor::new(include_str!("../../built-in-types.jsonl"))),
+                codec::JsonLinesDecoder::default(),
+            ),
+            10_000,
+        )
+        .await;
+
+    if let Err(err) = restore_result {
+        if err.contains::<VersionedUrlAlreadyExists>() {
+            tracing::info!("tried to restore built-in type, but they already exist");
```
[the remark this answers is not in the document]
TimDiekmann: This is the same behavior as before:
```rust
tracing::info!(%root_account_id, "tried to insert primitive {title} data type, but it already exists");
```

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: They would be ignored as the data already exists in the database. That's the same behavior as before.

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: This was my idea with snapshots from the get go, but it's definitely out of scope of this PR and probably not worth doing it currently. Having a general seeding script would be useful in any case.

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: `restore_built_in_types` consumes the store.

apps/hash-graph/bin/cli/built-in-types.jsonl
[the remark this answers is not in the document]
Alfred-Mountfield: Yes but I mean the first time you start up the app, not restarting it during a migration or crash or whatever :) anyway I'm not sure we have any easy avenues in front of us for how we would mitigate this even if we wanted to, so perhaps it doesn't matter

apps/hash-graph/bin/cli/src/subcommand/server.rs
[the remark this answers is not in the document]
Alfred-Mountfield: I don't feel strongly about this, just this is a condition we'd expect to be hit every single time you start up the server (aside from the _**very**_ first), so it seems a bit noisy. And the log message sounds a bit like an error, perhaps we'd want to say something like this instead:
"skipped persisting built-in type X as it is already present in the datastore"

The big difference I guess here is what happens if _one_ of the types already exists but the others don't? Does the logic above throw an error and fail early?

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: [`bf5ab0c` (#2638)](https://github.com/hashintel/hash/pull/2638/commits/bf5ab0c2acead679ca9b9285f9bf3f8bfb0f2fcc)

apps/hash-graph/lib/graph/src/snapshot.rs
TimDiekmann: Good point, I forgot that it's may be possible to insert an entity with the same UUID in snapshots.
[`09e51d5` (#2638)](https://github.com/hashintel/hash/pull/2638/commits/09e51d5e59d13e19ca9c8155470a6f0327101295)

apps/hash-graph/bin/cli/src/subcommand/server.rs
Alfred-Mountfield: To my understanding, this _isn't_ the same behaviour as before because we're allowing the hard-coded control over more information about the type here. We're also inserting temporal metadata and `fetchedAt` and such.

To re-emphasise a part of my point:
> As in, **let's say they change the timestamps**, or update the types to point to newer versions.. I think it'll still work fine, but might have some unexpected behaviour..

You would have a snapshot that you try and insert, there's a happy log message saying "didn't insert it because it already exists" and no warning that the data you _think_ you're inserting is actually distinct from the data that's already in the store.

apps/hash-graph/bin/cli/src/subcommand/server.rs
Alfred-Mountfield: Ah it takes ownership, makes sense

apps/hash-graph/bin/cli/built-in-types.jsonl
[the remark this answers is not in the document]
TimDiekmann: Sorry, but I think we're talking past each other. Why would the behavior be different when rerunning the application?

apps/hash-graph/bin/cli/src/subcommand/server.rs
TimDiekmann: The offline mode is not designed for being used in production but _**really**_ only for testing. The only point why I kept the offline mode is because you asked for it, removing it entirely would have been a whole lot easier. Honestly, I don't like to sink any more into this if not strictly required because the whole snapshot loading here is only as a workaround if you need to wipe your database often locally and you don't have a stable internet connection.

apps/hash-graph/bin/cli/built-in-types.jsonl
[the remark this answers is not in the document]
Alfred-Mountfield: The behaviour is not different, I thought you were getting confused about that, because of your comment here:
> But if you would have started the application prior this date, the built-in types should be in the database already.

Which seemed pretty unrelated to what I was saying 😅 

> Let's say I start my production instance of my application today, and I start to heavily utilise temporal versioning, creating dashboards over provenance, etc. If I know my application's history starts today, [...]

Would it make more sense to you if I add this?

> Let's say I start my production instance of my application today _for the first time ever_ [...]

I don't think this is purely philosophical by the way, although it is very much an edge-case. When dealing with timestamps and such it's important to know the limit points of your data, if any. That's part of why we ran into problems with the time crate, because it was trying to account for years before `0` (i.e. years BCE), which is meaningless in the use-cases we'd encounter or support in our app. Disagreements on the start of time is problematic, because you can set various invariants in your code, and it affects your ability to reason about things in the datastore.

If I launch my application today, therefore starting its history today, it could be problematic if I have things in my database that claim to have entered the database even before it even existed. What if I write migration logic or something to rollback elements that happened after timestamp X, etc.

apps/hash-graph/bin/cli/src/subcommand/server.rs
Alfred-Mountfield: In the conversation about this, I suggested I don't disagree with us removing it if it adds too much burden, which you seem to be suggesting here
> removing it entirely would have been a whole lot easier.

As I've expressed before, an offline mode is _not_ only useful for development or testing, it is entirely reasonable to think someone might want to run the graph in a ring-fenced environment without access to the external web. We are not actively prioritising that use-case right now (which is why I acknowledge we may want to abandon the capability for the moment) but those use-cases have been part of the reasoning from the very first time we discussed offline support.

If we are changing the code, I'd like us to change it to code that works in the way we would want it to work, rather than something with inconsistencies that can catch us out later on. If we do not have the capacity to implement this properly at the moment, then we should remove it altogether.

Alfred-Mountfield (approved):
I think the snapshot approach is still worth revisiting, but we might want to look at error handling and such a bit more beforehand (e.g. an option on the restore command like `exists_ok`)

PR #2641, opened 2023-06-05
Add tests to validate ontology types to exist in subgraph

TimDiekmann:

This adds test cases to retrieve ontology types if the resolve depths are not zeroed. Zeroed resolve depths are covered in other tests.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204610208832639/1204749865544745/f) _(internal)_

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2637 (tests are expected to fail until the fix is merged)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
+++ b/tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
@@ -204,9 +204,16 @@ describe("Ontology queries", () => {
     });
     expect(subgraph.roots.length).toEqual(3);
     expect(Object.keys(subgraph.edges).length).toEqual(0);
-    const data_types = getRoots(subgraph);
 
-    expect(data_types.length).toBe(3);
+    expect(
+      getRoots(subgraph)
+        .map(({ schema }) => schema.$id)
+        .sort(),
+    ).toStrictEqual([
+      "http://localhost:3000/@alice/types/data-type/number/v/1",
+      "http://localhost:3000/@alice/types/data-type/text/v/1",
+      "http://localhost:3000/@alice/types/data-type/text/v/2",
+    ]);
   });
 
   it("read property types", async () => {
@@ -224,9 +231,15 @@ describe("Ontology queries", () => {
     });
     expect(subgraph.roots.length).toEqual(2);
     expect(Object.keys(subgraph.edges).length).toEqual(0);
-    const property_types = getRoots(subgraph);
 
-    expect(property_types.length).toBe(2);
+    expect(
+      getRoots(subgraph)
+        .map(({ schema }) => schema.$id)
+        .sort(),
+    ).toStrictEqual([
+      "http://localhost:3000/@alice/types/property-type/name/v/1",
+      "http://localhost:3000/@alice/types/property-type/name/v/2",
+    ]);
   });
 
   it("read entity types", async () => {
@@ -244,9 +257,17 @@ describe("Ontology queries", () => {
     });
     expect(subgraph.roots.length).toEqual(4);
     expect(Object.keys(subgraph.edges).length).toEqual(1);
-    const entity_types = getRoots(subgraph);
 
-    expect(entity_types.length).toBe(4);
+    expect(
+      getRoots(subgraph)
+        .map(({ schema }) => schema.$id)
+        .sort(),
+    ).toStrictEqual([
+      "http://localhost:3000/@alice/types/entity-type/friendship/v/1",
+      "http://localhost:3000/@alice/types/entity-type/person/v/1",
+      "http://localhost:3000/@alice/types/entity-type/person/v/2",
+      "https://blockprotocol.org/@blockprotocol/types/entity-type/link/v/1",
+    ]);
   });
 });
 
--- a/tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
+++ b/tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
@@ -190,84 +190,135 @@ afterAll(async () => {
 
 describe("Ontology queries", () => {
   it("read data types", async () => {
-    const subgraph = await getDataTypes(graphContext, {
-      query: {
-        filter: {
-          all: [],
-        },
-        graphResolveDepths: {
-          ...zeroedGraphResolveDepths,
-          inheritsFrom: { outgoing: 1 },
-        },
-        temporalAxes: fullDecisionTimeAxis,
+    const resolve_depths = [
+      zeroedGraphResolveDepths,
+      {
+        ...zeroedGraphResolveDepths,
+        inheritsFrom: { outgoing: 1 },
       },
-    });
-    expect(subgraph.roots.length).toEqual(3);
-    expect(Object.keys(subgraph.edges).length).toEqual(0);
+      {
+        ...zeroedGraphResolveDepths,
+        inheritsFrom: { outgoing: 255 },
+      },
+    ];
 
-    expect(
-      getRoots(subgraph)
-        .map(({ schema }) => schema.$id)
-        .sort(),
-    ).toStrictEqual([
-      "http://localhost:3000/@alice/types/data-type/number/v/1",
-      "http://localhost:3000/@alice/types/data-type/text/v/1",
-      "http://localhost:3000/@alice/types/data-type/text/v/2",
-    ]);
+    for (const resolve_depth of resolve_depths) {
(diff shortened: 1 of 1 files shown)


tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -183,6 +188,68 @@ afterAll(async () => {
...
+describe("Ontology queries", () => {
+  it("read data types", async () => {
+    const subgraph = await getDataTypes(graphContext, {
+      query: {
+        filter: {
+          all: [],
+        },
+        graphResolveDepths: {
+          ...zeroedGraphResolveDepths,
+          inheritsFrom: { outgoing: 1 },
```
Alfred-Mountfield, line 200: Is there a reason we're doing this rather than leaving them all zeroed?

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -183,6 +188,68 @@ afterAll(async () => {
...
+        },
+        temporalAxes: fullDecisionTimeAxis,
+      },
+    });
+    expect(subgraph.roots.length).toEqual(3);
+    expect(Object.keys(subgraph.edges).length).toEqual(0);
+    const data_types = getRoots(subgraph);
+
+    expect(data_types.length).toBe(3);
```
Alfred-Mountfield, line 209: Do we want to check equality (or at least the `$id`) of the types in here to be what we expect? 

Not suggesting it is worth the extra code, but having an error about a specific vertex being missing (or making sure we don't accidentally put the wrong vertices in the wrong places, etc.) might be more helpful than just checking a length. Happy to merge otherwise

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
TimDiekmann: Described in the OP: The error only happens if at least one resolve depth is not zero (this is why this case wasn't covered before)

TimDiekmann: https://github.com/hashintel/hash/pull/2629 short circuited the traversal if no traversal is needed.

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
TimDiekmann: A few extra lines don't hurt here: [`1d7c84c` (#2641)](https://github.com/hashintel/hash/pull/2641/commits/1d7c84c1ed389871bec4f5e81d0bcfaaf4ce1164)

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
Alfred-Mountfield: Ah of course, sorry. I think this is somewhat of a confusing test at the moment then, and it'd be quite easy to accidentally remove our coverage of this if we go and implement inheritance for data types and update the test.

Perhaps we should rewrite the test as one that checks that the number of roots of the subgraph is the same regardless of the depths, so one with `0` depths and one with `255`? (And maybe we pick something that does have edges, such as property types)

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
Alfred-Mountfield: Thanks, I prefer this :)

Although we probably want to improve this a bit later to avoid the magic strings and have them as a constant that's referable somewhere, or something we can import from the underlying test data. Could see this being a bit annoying to update if we change the tests

TimDiekmann: I don't know yet how we could achieve this as we cannot easily use the strings from the snapshot, but I agree that these should be improved at some point.

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
TimDiekmann: I added a wider range of resolve depths: [`27bf7af` (#2641)](https://github.com/hashintel/hash/pull/2641/commits/27bf7afb2edbc1b7b477f4a64dc3efdde402baff)

Alfred-Mountfield: One final comment https://github.com/hashintel/hash/pull/2641#discussion_r1218154126

As suggested I think `0` and `255` would be better as it checks our short circuit logic too

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -190,84 +190,135 @@ afterAll(async () => {
...
-        graphResolveDepths: {
-          ...zeroedGraphResolveDepths,
-          inheritsFrom: { outgoing: 1 },
-        },
-        temporalAxes: fullDecisionTimeAxis,
+    const resolve_depths = [
+      zeroedGraphResolveDepths,
+      {
+        ...zeroedGraphResolveDepths,
+        inheritsFrom: { outgoing: 1 },
```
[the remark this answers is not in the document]
Alfred-Mountfield: Ah apologies, I didn't notice the one above, was slightly confusing because there were spread operators on each other entry.

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
Alfred-Mountfield: ^ got confused by the syntax, disregard, thanks for clarifying :)

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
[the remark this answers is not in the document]
TimDiekmann: I agree that they are not necessarily readable. I could change it it
```ts
[
      {
        ...zeroedGraphResolveDepths,
      },
      {
        ...zeroedGraphResolveDepths,
        constrainsValuesOn: { outgoing: 1 },
        constrainsPropertiesOn: { outgoing: 1 },
      },
]
```

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -183,6 +188,140 @@ afterAll(async () => {
...
+        ...zeroedGraphResolveDepths,
+        constrainsValuesOn: { outgoing: 255 },
+      },
+      {
+        ...zeroedGraphResolveDepths,
+        constrainsPropertiesOn: { outgoing: 255 },
+      },
+    ];
+
+    for (const resolve_depth of resolve_depths) {
```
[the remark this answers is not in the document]
TimDiekmann: I will take a look...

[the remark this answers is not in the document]
TimDiekmann: [`7f045c4` (#2641)](https://github.com/hashintel/hash/pull/2641/commits/7f045c4e7d9cbf097c29d78b2599f2b421160471)

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
```diff
@@ -189,136 +189,124 @@ afterAll(async () => {
...
+    zeroedGraphResolveDepths,
+    {
+      ...zeroedGraphResolveDepths,
+      inheritsFrom: { outgoing: 1 },
+    },
+    {
+      ...zeroedGraphResolveDepths,
+      inheritsFrom: { outgoing: 255 },
+    },
+  ])("read data types %#", async (resolve_depths) => {
```
Alfred-Mountfield, line 202: I'd perhaps go for `%j` or `%o` or something rather than printing the index here, but happy that we have identifiable output at least

Alfred-Mountfield (approved):

tests/hash-backend-integration/src/tests/subgraph/friendship.test.ts
TimDiekmann: Spoke about this offline, and we agreed that the full object notation would be _a little bit_ too long.

Alfred-Mountfield: (see this)
image

If indices prove unclear later on we might swap to adding some brief descriptions alongside the test cases or something

PR #567, opened 2022-05-13
Migrate JS internal packages to TS

leudz:
## 🌟 What is the purpose of this PR?

The PR aims to provide support for Typescript.
This means having Typescript packages and behaviors.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1199548034582004/1202239593614677/f) _(internal)_

## 🚫 Blocked by

- [ ] [Investigate memory leak](https://app.asana.com/0/1199548034582004/1202479511750422/f) (_internal_)

## 🔍 What does this change?

- Allow packages with the extensions "js", "jsx", "ts" and "tsx"

## 📜 Does this require a change to the docs?

- The docs for x need updating to explain that y

## 🛡 What tests cover this?

- All tests involving Javascript behaviors

## ❓ How to test this?

1.  Checkout the branch / view the deployment
2.  Try `cargo test`
3.  Confirm that all tests pass

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203358502199055

PR #2651, opened 2023-06-08
H-86: Add minor hEngine compatibility with TypeScript for behaviors and initialization

rivertam:
<details>
<summary><h2>Preamble</h2></summary>

I'm really interested in this project, both as a user and as a potential contributor. I have a bespoke simulation framework that I've been toying with and would be interested in trying to merge some of the work I've done on it to Hash engine to see if it's useful. I'd love to take some conversations offline and I plan on messaging in the Discord, but just fyi I'm open to hopping on a video call to have higher throughput.

This PR isn't so much "I want to merge this" as it is "I wanted to get familiar enough with the internals to be able to do this", and I've spent the last couple of weeks reading through and trying to figure out how the project works from the inside. I've certainly made progress, but I'm sure guidance would help me speed up! In terms of this PR, I do need to do more thorough testing and add automated tests and I'm sure other things that are part of the process, but I wanted to submit for feedback first to avoid wasting my time. Think of this as a very involved Discussion rather than a PR, maybe.

I have a couple of ideas on things I want to work on after this PR (see next steps), but the long-and-short is that I'm interested in using [`swc_bundler`](https://github.com/swc-project/swc/blob/main/crates/swc_bundler) to make it possible to split up complex logic (and maybe get rid of some temporary hacks on the worker).
</details>

This uses the swc compiler to strip the TypeScript types from behavior files named `<something>.ts` and initialization files named `init.ts`

## 🔍 What does this change?

This adds a compilation step to the simulation executor for initialization and for behavior execution

> Confirm you have taken the necessary action to record a changeset or publish a change, as appropriate

I absolutely have not and am not sure how to do this! TBD

This PR:

The changes in this PR:

  - I'm not sure if this should even be merged considering all it does is strip types (and not check them)
  - If it were to be merged, I think the docs could reflect that TypeScript is supported in these places

The changes in this PR:

  - I added swc to Cargo.toml, but I'm not sure if that's involved with the Turbo Graph or if that's just npm

## ⚠️ Known issues

- No type checking
- No exported types that are useful

## 🐾 Next steps

Great question

- Export types that can be used in TypeScript (somehow)
- Check types using `tsc` (is this out of scope? might be better to just leave this to editors/the web client)
- Personally, the next thing _I_ want to work on is adding `swc_bundler` to the manifest stage so we can split the `init` and behavior files into multiple files, which I think is currently unsupported (please correct me if I'm wrong!). While this wouldn't much help close out the "add TypeScript support" checkbox, it's what I want next as a user.

## 🛡 What tests cover this?

- None! I am definitely interested in adding tests but I wanted to get feedback before wasting my time!

## ❓ How to test this?

- So far, I've just converted the most basic project (the one with 2 agents with ages 0 and 5 going up) to TypeScript. It's so minimal it's bad! I've attached the project I'm using.

[ageing-agents.ts.tar.gz](https://github.com/hashintel/hash/files/11691171/ageing-agents.ts.tar.gz)

## 📹 Demo

I guess here ya go!

https://asciinema.org/a/ee9BQceUMV0t5OTAiWw7icy76

### DIFF
--- a/apps/engine/lib/execution/Cargo.toml
+++ b/apps/engine/lib/execution/Cargo.toml
@@ -21,6 +21,8 @@ rand = "0.8.5"
 rayon = "1.5.3"
 serde = { version = "1.0.138", features = ["derive"] }
 serde_json = "1.0.82"
+swc = "0.261.29"
+swc_common = "0.31.11"
 thiserror = "1.0.31"
 tokio = { version = "1.19.2", features = ["macros", "rt", "sync", "process", "time"] }
 tracing = "0.1.35"
--- a/apps/engine/lib/execution/src/package/simulation.rs
+++ b/apps/engine/lib/execution/src/package/simulation.rs
@@ -15,7 +15,7 @@
 //!
 //! ## Package types
 //!
-//! The the simulation [`Package`] System consists of 4 types of interfaces for 4 types of
+//! The simulation [`Package`] System consists of 4 types of interfaces for 4 types of
 //! simulation packages:
 //! - [`init`]alization
 //! - [`context`]
--- a/apps/engine/lib/execution/src/package/simulation/init/task.rs
+++ b/apps/engine/lib/execution/src/package/simulation/init/task.rs
@@ -1,5 +1,5 @@
 use crate::{
-    package::simulation::init::js_py::{JsInitTask, PyInitTask},
+    package::simulation::init::js_py::{TsInitTask, JsInitTask, PyInitTask},
     task::{
         StoreAccessValidator, TargetedTaskMessage, Task, TaskDistributionConfig, TaskSharedStore,
     },
@@ -8,10 +8,12 @@ use crate::{
     Error, Result,
 };
 
+
 /// All init package tasks are registered in this enum
 #[derive(Clone, Debug)]
 pub enum InitTask {
     JsInitTask(JsInitTask),
+    TsInitTask(TsInitTask),
     PyInitTask(PyInitTask),
 }
 
@@ -19,6 +21,7 @@ impl Task for InitTask {
     fn name(&self) -> &'static str {
         match self {
             Self::JsInitTask(task) => task.name(),
+            Self::TsInitTask(task) => task.name(),
             Self::PyInitTask(task) => task.name(),
         }
     }
@@ -26,6 +29,7 @@ impl Task for InitTask {
     fn distribution(&self) -> TaskDistributionConfig {
         match self {
             Self::JsInitTask(task) => task.distribution(),
+            Self::TsInitTask(task) => task.distribution(),
             Self::PyInitTask(task) => task.distribution(),
         }
     }
@@ -47,6 +51,7 @@ impl WorkerHandler for InitTask {
     fn start_message(&self) -> Result<TargetedTaskMessage> {
         match self {
             Self::JsInitTask(inner) => inner.start_message(),
+            Self::TsInitTask(inner) => inner.start_message(),
             Self::PyInitTask(inner) => inner.start_message(),
         }
     }
--- a/apps/engine/lib/execution/src/runner/javascript/modules.rs
+++ b/apps/engine/lib/execution/src/runner/javascript/modules.rs
@@ -42,6 +42,7 @@ pub(in crate::runner::javascript) fn import_module<'s>(
     let source_code = read_file(path).map_err(|err| {
         JavaScriptError::AccessJavascriptImport(path.to_string(), err.to_string())
     })?;
+
     let js_source_code = new_js_string(scope, &source_code);
     let js_path = new_js_string(scope, path);
     let source_map_url = v8::undefined(scope);
--- a/apps/engine/lib/execution/src/runner/javascript/modules.rs
+++ b/apps/engine/lib/execution/src/runner/javascript/modules.rs
@@ -42,7 +42,6 @@ pub(in crate::runner::javascript) fn import_module<'s>(
     let source_code = read_file(path).map_err(|err| {
         JavaScriptError::AccessJavascriptImport(path.to_string(), err.to_string())
     })?;
-
     let js_source_code = new_js_string(scope, &source_code);
     let js_path = new_js_string(scope, path);
     let source_map_url = v8::undefined(scope);
--- a/apps/engine/lib/execution/src/runner/language.rs
+++ b/apps/engine/lib/execution/src/runner/language.rs
@@ -2,7 +2,7 @@ use core::fmt;
 
 use serde::{Deserialize, Serialize};
 
-use crate::{Error, Result};
+use crate::{runner::JavaScriptError, Error, Result};
 
 /// Supported languages
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
@@ -10,6 +10,7 @@ pub enum Language {
     JavaScript = 0,
     Python = 1,
     Rust = 2,
+    TypeScript = 3,
 }
 
 impl fmt::Display for Language {
@@ -19,9 +20,13 @@ impl fmt::Display for Language {
 }
 
 impl Language {
-    pub const NUM: usize = 3;
-    pub const ORDERED: [Language; Self::NUM] =
-        [Language::JavaScript, Language::Python, Language::Rust];
+    pub const NUM: usize = 4;
+    pub const ORDERED: [Language; Self::NUM] = [
+        Language::JavaScript,
+        Language::Python,
+        Language::Rust,
+        Language::TypeScript,
+    ];
 
     pub fn as_index(self) -> usize {
         self as usize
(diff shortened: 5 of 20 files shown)


apps/engine/lib/execution/src/runner/language.rs
```diff
@@ -43,8 +48,76 @@ impl Language {
...
+    use std::sync::Arc;
+
+    use swc::config::Options;
+    use swc_common::{errors::Handler, Globals};
+
+    eprintln!(
+        "Stripping typescript from {filename} (source code: {})",
+        source_code.len()
+    );
+
```
rivertam, line 91:
```suggestion
```

apps/engine/lib/execution/src/package/simulation.rs
```diff
@@ -15,7 +15,7 @@
 //!
 //! ## Package types
 //!
-//! The the simulation [`Package`] System consists of 4 types of interfaces for 4 types of
+//! The simulation [`Package`] System consists of 4 types of interfaces for 4 types of
```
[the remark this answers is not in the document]
nonparibus: @rivertam Hi Ben, thanks for opening this PR! Would you be willing to split these out? We'll be able to approve and merge these in much sooner than we will the overall PR. Super grateful for your efforts here.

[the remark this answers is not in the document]
rivertam: np gimme a couple minutes1

apps/engine/lib/execution/src/runner/javascript/modules.rs
```diff
@@ -42,6 +42,7 @@ pub(in crate::runner::javascript) fn import_module<'s>(
     let source_code = read_file(path).map_err(|err| {
         JavaScriptError::AccessJavascriptImport(path.to_string(), err.to_string())
     })?;
+
```
rivertam, line 45:
```suggestion
```

apps/engine/lib/execution/src/package/simulation.rs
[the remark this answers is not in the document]
rivertam: https://github.com/hashintel/hash/pull/2652

apps/engine/lib/execution/src/package/simulation/init/task.rs
```diff
@@ -8,24 +8,28 @@ use crate::{
...
+
 /// All init package tasks are registered in this enum
 #[derive(Clone, Debug)]
 pub enum InitTask {
     JsInitTask(JsInitTask),
+    TsInitTask(TsInitTask),
     PyInitTask(PyInitTask),
 }
 
 impl Task for InitTask {
```
rivertam, line 20: Overall I think the traits for `Task` could do with some overhaul, but one thing in particular I've considered is applying [this technique](https://github.com/hashintel/hash/pull/2653) to `InitTask`.

apps/engine/lib/experiment-structure/src/manifest.rs
```diff
@@ -615,7 +618,7 @@ fn get_behavior_from_dependency_projects(
         } else {
             name_root.replace('_', "-")
         };
-        let full_name = "@hash/".to_string() + &dir + "/" + file_name;
+        let full_name = format!("@hash/{dir}/{file_name}");
```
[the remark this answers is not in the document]
TimDiekmann: Thanks for spotting this ugly beast 😄

apps/engine/lib/execution/src/package/simulation.rs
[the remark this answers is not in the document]
nonparibus: Thank you! ❤️

apps/engine/lib/execution/Cargo.toml
```diff
@@ -21,6 +21,8 @@ rand = "0.8.5"
 rayon = "1.5.3"
 serde = { version = "1.0.138", features = ["derive"] }
 serde_json = "1.0.82"
+swc = "0.261.29"
+swc_common = "0.31.11"
```
TimDiekmann, line 25: Is this also possible using `v8`? `swc` is a pretty big dependency with very many (and outdated) dependencies.

TimDiekmann (commented):
Thanks for the PR, this seems like a very useful addition.

I did a quick look through the PR. Generally, the code looks good, however, I have one major concern about the newly added dependency.

It would also be good if you could add a few small test cases. The easiest way to achieve this is probably to just change a few `init.js` to `init.ts` in the integration test suite.

apps/engine/lib/execution/Cargo.toml
TimDiekmann: There was a [related PR](https://github.com/hashintel/hash/pull/567), which added TS support for another location in the code. I haven't looked at transpiling TS to JS code in Rust, yet, but it might worth taking a look 🙂

rivertam: No, v8 does not have TypeScript compatibility. Unfortunately, the options to compile TypeScript to JavaScript are pretty limited and all pretty heavy in this context. swc is used all throughout the TS ecosystem, perhaps most notably by Next.js, and is often touted as the fastest/lightest production compiler. It's also the only compiler I know of that doesn't require installing it externally and shelling out to it.

I'm not a huuuge fan of swc either, but it's kind of the best for what it does right now. Competitors include esbuild (which includes bundling), babel, and the default TypeScript compiler (`tsc`). Eventually, `tsc` needs to be in the stack somewhere to do actual type-checking, but the compiler is written in JS, it's quite slow (seconds where swc is milliseconds), and it isn't easily invoked directly through Rust (unlike swc). Additionally, the next stop here was to add `swc_bundler`, the best bundling solution I can see (can turn two TS files into one; this would allow importing files from other files, including JS) that's accessible by Rust.

Another solution I've considered is transitioning the JS runner to use [`bun`](https://bun.sh/) which can do a lot of stuff "automatically". Downsides include lack of Windows support, it's supported by a single developer, and we'd definitely have to shell out to invoke it (rather than being able to inline it). However, I think there'd be a lot of upsides including automatically included ffi. There's potentially a lot to discuss here.

I'm definitely open to other solutions, though. I just think they're mostly actually heavier and more cumbersome than `swc`. I also don't want to keep investing in swc (through `swc_bundler`) if you don't want to merge it here, so definitely worth figuring out what the optimal final setup would look like. In particular, the alternatives to `swc_bundler` would all require shelling out as well, so I think we could want a totally different approach.

I'd boil it down to the two approaches of

1. Do everything inline/in memory (the way this PR does). If we want this, there aren't really any good alternatives to swc right now, but if they ever were developed (e.g. Rome), we could pretty easily swap it in place of swc.
2. Shell out to another project that uses the file system itself. This would be something like `esbuild` or `tsc`. I view this as the "heavier" option, but one nice aspect is that I think it will be more naturally conducive to bundling as one step (`esbuild` in particular can do both).

Please feel free to take your time with this. I'm in no rush to get this merged in, and I don't think you are either. I'm pretty happy either way, but I would of course eventually like to be able to write simulations using TypeScript, so that's my only real goal.

rivertam:
@TimDiekmann Definitely gonna add test cases before recommending a merge. Do you think it should be merged soon? I was expecting it to wait on other stuff (actual type-checking, documentation, perhaps frontend support; I'm not sure what the processes are here)

apps/engine/lib/execution/Cargo.toml
nonparibus: Cheers for the detailed and thoughtful reply.

FWIW, I know Jarred @ Oven raised $7m+ for [`bun`](https://bun.sh/) and friends quite a while back now, so if it is genuinely just supported by him at the moment I think that likely that won't be the case for all that long. But there is certainly more adoption risk w/ it than `swc` in my mind, and I would consider lack of native Windows support to be a blocker given our existing/historical user-base and desire to remain widely accessible.

I'll leave actual follow-up re: next steps and a decision here to Tim, but I can see the argument for `swc` and the benefit this PR provides beyond what we have in place already. Like you say, updates to this version of hEngine are not a high priority for us at present, but this seems like an obvious win if we can bag it in some form, and we're very grateful for the contribution here. We'll do what we can to see it reviewed in a timely fashion.

rivertam: Yep, I'm very excited about bun in the future. 

For my edification, what are some of your top priorities at the moment? Is there a future version of the engine in the works?

nonparibus: @rivertam We're presently evaluating the best role for hEngine to play in simulation pipelines that more heavily make use of generative agents. For the avoidance of doubt, we aim to continue work on this version of the engine (as opposed to another), but the future roadmap is currently in flux.

TimDiekmann:
There is no need to merge this soon, no.

To what extent do you expect to have type-checking in place? Documentation would be nice, yes, but hEngine does not have a frontend, it's only driven through the CLI currently. Just to make sure we're talking about the same thing here: hEngine is _not_ the backend of https://core.hash.ai, it's a standalone engine written from scratch.

rivertam:
Ahh, right, I forgot that this is the "alpha" version of the next iteration. That's perfectly fine with me -- I've been using this standalone.

In terms of what I'm expecting, for now I would expect the engine just to crash in the same way I'd expect it to if you provide a bad behavior path or no initial state. It's certainly all at start-up time, nothing dynamic about it. Let me add in a couple of tests here, but given that it's not in production, I see how there's no great reason to block merging it.

rivertam:
Sorry for the delay on this y'all. I had tested this just by slapping a `const age: number = 5` or something like that in `age.js` (and turned it into `age.ts`.

However, to use TypeScript effectively here, we'll want to insert code similar to this:

```typescript
import * as globals from '../globals.json'

type Context<Globals extends object> = {
  globals(): Globals
}

// ... redacted

// ignore `state` for now
function behavior(state: State<typeof stateSchema>, context: Context<typeof globals>) { /* redacted */ }
```

Unlike the "slap a `: number` somewhere in the code" approach, this uses an `import`. We're only using it for types (not values, which would definitely necessitate a bundler), but this causes the `strip_typescript` function to emit a module, which doesn't seem to be immediately resolvable by V8.

swc has multiple module output modes, including ES Modules which v8 is ostensibly somewhat compatible with, but they all include some notion of importing/exporting unlike the way behaviors are structured right now where the variables are global a la `<script>` tag.

The way I see it, there is basically no point in having TypeScript support if we can't import at least things like the `Context` template class and use it in our editor. The first is to eagerly add `swc_bundler` to this PR (or make a chain PR, or something), which would add a lot of complexity to the preconditions of merging. The second is to work with v8/the runners to get them to execute modules properly to avoid having to do _too_ much work. However, I'm less confident this is work that will actually be useful in the future (as we may have no need to execute modules in this way in the future).

Not sure exactly which I'm gonna explore next. Let me know if you have any opinions. I responded to @nonparibus on Discord 3 days late that I'd love to sync and get to know more about the state of this project and how it fits into your business; I think that would obviously help a lot in deciding next steps here. You can reach me there (and I'll try to be more responsive) or just email me at ben.m.berman@gmail.com.

apps/engine/lib/execution/Cargo.toml
nonparibus: Yes, exactly. :) (And sorry, missed this comment somehow!)

nonparibus:
Hey @rivertam, thanks for your patience! Just to note that we've now moved this project across into our `labs` repo. You can find its new home @ https://github.com/hashintel/labs/tree/main/apps/sim-engine

TimDiekmann:
Closing in favor of https://github.com/hashintel/labs/pull/41

PR #2653, opened 2023-06-08
Simplify some trait boilerplate by using Deref

rivertam:

I noticed an opportunity to remove some boilerplate. This pattern is used a number of times throughout the repo, and I find it to be somewhat cumbersome, so this is just a suggested way to get around having this kind of indirection. I also recognize this is probably a matter of opinion, which is why I only gave this treatment to one enum rather than overzealously applying it across the repo (also, I'm not familiar enough with the files to remember all the places I've seen it). What do you think?

On a technical level, it swaps enum-oriented dynamicism for vtable lookups. I haven't profiled it or anything, but it seems like it would be a very minor penalty if a penalty at all.

This PR:

## 🐾 Next steps

We could apply this to other similar enums

## 🛡 What tests cover this?

Dunno 😬 I assume all of them

## ❓ How to test this?

I just tried it on a couple of projects; I'd expect it to fail instantly on any project if there were some kind of issue. Please advise if there's an easy set of tests I could run or where I should put unit/integration tests that might be relevant.

### DIFF
--- a/apps/engine/lib/execution/src/worker.rs
+++ b/apps/engine/lib/execution/src/worker.rs
@@ -495,7 +495,7 @@ impl Worker {
         source: Language,
     ) -> Result<()> {
         if let Some(pending) = self.tasks.inner.get_mut(&msg.task_id) {
-            let next = WorkerHandler::handle_worker_message(&mut pending.task, msg.payload)?;
+            let next = WorkerHandler::handle_worker_message(&mut *pending.task, msg.payload)?;
             match next.target {
                 MessageTarget::Rust => {
                     let inbound = InboundToRunnerMsgPayload::TaskMsg(RunnerTaskMessage {
@@ -626,7 +626,7 @@ impl Worker {
     /// [`PartialSharedState::split_into_individual_per_group()`]: crate::task::PartialSharedState::split_into_individual_per_group
     async fn spawn_task(&mut self, sim_id: SimulationId, task: WorkerTask) -> Result<()> {
         let task_id = task.task_id;
-        let msg = WorkerHandler::start_message(&task.task)?;
+        let msg = WorkerHandler::start_message(&*task.task)?;
 
         let context = task.shared_store.context().clone();
         let shared_stores = match task.shared_store.state {
--- a/apps/engine/lib/execution/src/package/simulation/comms.rs
+++ b/apps/engine/lib/execution/src/package/simulation/comms.rs
@@ -2,7 +2,7 @@ use stateful::field::PackageId;
 
 use crate::{
     package::simulation::{PackageTask, SimulationId},
-    task::{ActiveTask, StoreAccessValidator, TaskId, TaskSharedStore},
+    task::{ActiveTask, TaskId, TaskSharedStore},
     worker_pool,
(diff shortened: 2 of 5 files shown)


apps/engine/lib/execution/src/worker.rs
```diff
@@ -495,7 +495,7 @@ impl Worker {
         source: Language,
     ) -> Result<()> {
         if let Some(pending) = self.tasks.inner.get_mut(&msg.task_id) {
-            let next = WorkerHandler::handle_worker_message(&mut pending.task, msg.payload)?;
+            let next = WorkerHandler::handle_worker_message(&mut *pending.task, msg.payload)?;
```
rivertam, line 498: It is kinda ugly to have to put the `*` here. There might be a way around that, but I couldn't think of it.

TimDiekmann (changes requested):
Thanks for the PR @rivertam!

I hear your concern about the boilerplate. Previously, we used the [`enum-dispatch`](https://docs.rs/enum_dispatch/latest/enum_dispatch/) crate for exactly these functions to avoid exactly this boilerplate, however, this made development kind of a nightmare as (at least back then) IDEs were not able to properly parse proc macros.

I'm happy to have a common trait to be implemented on these types, however, I feel quite strongly about `Deref` and `DerefMut` as this is a misuse of these traits. They are meant to be used for smart pointers and similar (this is even explicitly mentioned [in the documentation](https://doc.rust-lang.org/std/ops/trait.Deref.html):

> Implementing `Deref` for smart pointers makes accessing the data behind them convenient, which is why they implement `Deref`. On the other hand, the rules regarding `Deref` and `DerefMut` were designed specifically to accommodate smart pointers. Because of this, **`Deref` should only be implemented for smart pointers** to avoid confusion.

I remember that we used `enum-dispatch` in a few locations and `Task` was one of them, out of my head I cannot remember the other locations either 😅 

I'm happy to simplify the code, but not with a `Deref`. 

Regarding the performance I'm unsure if this has any implications, probably not, so boxing the values would be another option. What do you think?

rivertam:
I agree about Deref/DerefMut. I didn't know about `enum_dispatch`. If it's not an issue with `rust-analyzer`, probably worth moving back. Otherwise, maybe it should just be a method like `as_task()` and `as_task_mut()`?

TimDiekmann:
Yes, I think adding `as_task()` and `as_task_mut()` is a sensible addition, thanks!

rivertam:
I tried to do this with `enum_dispatch` and couldn't get it to work right (I think the traits and enums have to be in the same file for it to work?), and I ended up not liking putting `task.task.as_task()` everywhere, so I implemented `as_task` and `as_task_mut` as private and just use them to implement the traits. I don't really like how it turned out -- feel free to either close or merge, tbh.

TimDiekmann:
Thanks for the update. I don't think this is a real improvement. While it does deduplicate some code lines, it's not really helping readability. Because of this, I prefer to keep it as it was before. We really appreciate the PR, though and I'm also looking forward to your other PR 🙂

rivertam:
agreed!

PR #2664, opened 2023-06-13
Remove unneeded PrivateLinks from HASH infrastructure

thehabbos007:

This PR removes AWS PrivateLinks from the HASH TF infra. These are unnecessary as our ECS cluster requires outbound network, and we handle that by putting the ECS service on a subnet with an internet gateway. Unlike the subnet for RDS/Redis, the ECS service has internet access and does not require PrivateLink connections to communicate with any AWS service data plane.

This PR also adds modules to move to a VPC Hub and VPC Spoke architecture for our PrivateLinks in the case we require ingress-only networks. These new modules are not used, but lightly tested (and functioning) for the type of ECS workload we run.

The VPC Hub module is self-contained has would have to be deployed on its own. Any VPC that needs PrivateLink connection would peer to the Hub throuhg the `vpc_spoke_peer` module, which would attach a VPC to the relevant routing tables and DNS rules to allow routing through the VPC Hub.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204610208832639/1204766818588447/f) _(internal)_

## 🔍 What does this change?

- Adds module for VPC Hub and Spoke
- Remove PrivateLink in HASH
 
## 🐾 Next steps

- Apply the same changes to our internal infrastructure.

## 🛡 What tests cover this?

- CI/CD

## ❓ How to test this?

Confirm CI passes and the subsequent apply passes in CD

Go to production HASH instance and verify it's working (changes should already be deployed)


infra/terraform/modules/privatelink_endpoints/main.tf
```diff
@@ -0,0 +1,74 @@
...
+    # Used for various Fargate related operations and Fargate Exec
+    ec2messages = {
+      name        = "com.amazonaws.${var.region}.ec2messages"
+      type        = "Interface"
+      private_dns = false
+      phz_name    = "ec2messages.${var.region}.amazonaws.com"
+      alias       = [""]
+    }
+    # Used for logging to CloudWatch
+    ec2messages = {
```
TimDiekmann, line 34: Is this supposed to be called `ec2logs`?
```suggestion
    ec2logs = {
```

infra/terraform/modules/vpc_spoke_peer/main.tf
```diff
@@ -0,0 +1,60 @@
...
+locals {
+  endpoints = module.endpoints.endpoints
+}
+
+data "aws_route53_zone" "endpoints" {
+  for_each     = local.endpoints
+  name         = each.value.phz_name
+  private_zone = true
+  tags         = { Group = "vpc-hub" }
+}
```
TimDiekmann, line 22: That's a pretty nice way to define the endpoints

infra/terraform/modules/privatelink_endpoints/main.tf
thehabbos007: Yes, thanks!

infra/terraform/vpc_hub/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+    H -- PrivateLink <--> S
+    V1 -- VPC Peering Connection <--> H
+    V2 <--> H
+
+```
+
+## Considerations for using the VPC Hub
+
+When a VPC Spoke is to be connected to the Hub, it's essential that theres a non-ambiguous network route to the Spoke. Essentially this means our infrastructure should have unique CIDR blocks for each environment in every project. If this is not the case, the VPC Hub will not know which Spoke to route traffic back to. I suggest reading this [AWS documentation site](https://docs.aws.amazon.com/vpc/latest/peering/peering-configurations-partial-access.html#peering-incorrect-response-routing) to understand why this is the case.
```
CiaranMn, line 22: What do we mean by 'environment' here? I understand that 'project' corresponds to a VPC (I think)

infra/terraform/vpc_hub/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+If we were to use this VPC Hub, we have to make an active decision to use unique CIDR blocks for each environment in every project in AWS. A common CIDR delegation scheme such as the following would be beneficial to stick to:
+
+```text
+10.0.0.0/14 ~ [Project Root] ~ e.g. 10.0.0.0 - 10.3.255.255
+    This is a logical part of the network and never actually allocated in AWS.
```
CiaranMn, line 28: Didn't quite follow this comment – does it mean that this will be the implicit CIDR block of all environments in a project, but we never actually write `10.x.x.x/14` anywhere in a Terraform file/AWS config? wouldn't we need `10.0.0.0/14` as the CIRD block for a Project that contains 4 environment, where each environment is taking up `/16`? I may have misunderstood all this

CiaranMn (commented):
Thanks @thehabbos007 – great resource usage reduction!

I've a few questions to help my own understanding.

thehabbos007:
@CiaranMn I'll merge this now as I don't want state drift to happen. I'll address you comments and fixes this in a follow up

infra/terraform/vpc_hub/README.md
thehabbos007: An environment would be `dev`, `prod`, `test`, etc. We currently only use prod/dev, but the given delegation scheme allows up to 4 environments per project. For example, `HASH` would be the project, and `prod` would be an environment we host in a VPC

infra/terraform/vpc_hub/README.md
thehabbos007: Yes, you got it right. We can't delegate that /14 because AWS doesn't allow more than a `/16` for a VPC. So we'd have to manage our numbering separately. The /14 doesn't need to be allocated anywhere because each VPC would just take on `10.0../16`, `10.1../16`, `10.2../16`, `10.3../16` for each environment.

Ideally we would be able to actually allocate a `/14`, but at this point it is more complicated than keeping track of the numbers per project. I think ideally we'd be able to have some shared Terraform module that has all project names so we can at least statically define the delegation. Maybe it should be an array of project names and calculate the project `/14` from the index in the list

infra/terraform/vpc_hub/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+10.0.0.0/16 ~ [Environment]  ~ e.g. 10.0.0.0 - 10.0.255.255
```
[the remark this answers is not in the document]
thehabbos007: Yep correct, COuld be something like 
HASH  – VPC
prod: 10.0.0.0/16
staging: 10.1.0.0/16
dev: 10.2.0.0/16
test: 10.3.0.0/16

PR #2691, opened 2023-06-27
Use Advanced Visibility in Postgres for Temporal

thehabbos007:

This PR enables Advanced Visibility for our local Temporal setup in anticipation for using the a similar setup in production.
From Temporal version 1.20+ it is possible to use Postgres 12+ as an Advanced Visibility store, which means we don't need to run Cassandra/OpenSearch for Temporal.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204521571620726/f) _(internal)_

## 🔍 What does this change?

- Changes the setup/migrate scripts slightly to switch to pg12+
- Use `TEMPORAL_ADDRESS` instead of `TEMPORAL_CLI_ADDRESS` as the latter is deprecated.
- Parameterize the Temporal docker container version to allow us to easily update the version we're running

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Prod infra for temporal [Asana task](https://app.asana.com/0/1204274556967336/1204521571620726/f) (_internal_)

## 🛡 What tests cover this?

Not directly tested, but local dev/ci would fail if the compose is wrong.

## ❓ How to test this?

1.  Checkout the branch
1.  Start external services with an OpenAI key configured
1.  Run `temporal workflow execute -t ai -w foo --type DemoWorkflow -i '"How many eggs can you fit in a basket"'` to submit a workflow
1.  Verify that the advanced visibility store is working by running `temporal workflow list --query "TaskQueue = 'ai' AND ExecutionStatus = 2"`. This command fails if advanced visibility is not configured for the operator

## 📹 Demo

image


apps/hash-external-services/temporal/setup.sh
```diff
@@ -106,30 +61,32 @@ setup_server() {
...
         # ipv4
-        export TEMPORAL_CLI_ADDRESS="${BIND_ON_IP}:7233"
+        export TEMPORAL_ADDRESS="${BIND_ON_IP}:7233"
     fi
 fi
 
+# TEMPORAL_CLI_ADDRESS is deprecated and support for it will be removed in the future release.
+if [[ -z "${TEMPORAL_CLI_ADDRESS:-}" ]]; then
+    export TEMPORAL_CLI_ADDRESS="${TEMPORAL_ADDRESS}"
+fi
```
TimDiekmann, line 90: As we use `TEMPORAL_ADDRESS` above why we need to set the deprecated environment variable?

TimDiekmann (approved):
Changes look good and it's working locally 🎉 🪨 

I have a single question but I don't consider it as blocking the PR from being merged.

apps/hash-external-services/temporal/setup.sh
thehabbos007: We don't really need it. It does allow us to downgrade temporal in case we need to use a previous version, though. It was useful for me when I tested the migration path from 1.20 to 1.21 to ensure that our system was still oeprational. The official scripts also do this and I thought it would be good to follow their example.

PR #2693, opened 2023-06-29
Move logic to ignore conflicting ontology type to Postgres

TimDiekmann:

When loading multiple external types at the same time which uses the same underlying ontology types one connection will fail to insert that datatype.

To avoid this we need to provide the on-conflict behavior to Postgres.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204727813960182/1204871439592713/f) _(internal)_

## 🔍 What does this change?

- Remove old ontology creation functions (required as the parameters change and we don't want to overload the functions but replace then(
- Add three new functions:
    - Create ontology id from a BaseURL and a version, taking into account if the type is external or failure is allowed
    - Create metadata (owned + external), calls the above function and insert the metadata afterwards

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/postgres_migrations/V8__ontology_create_functions.sql
+++ b/apps/hash-graph/postgres_migrations/V8__ontology_create_functions.sql
@@ -1,0 +1,159 @@
+DROP FUNCTION
+  create_ontology_id (
+    "ontology_id" UUID,
+    "base_url" TEXT,
+    "version" BIGINT,
+    "record_created_by_id" UUID
+  );
+
+DROP FUNCTION
+  create_owned_ontology_id (
+    "base_url" TEXT,
+    "version" BIGINT,
+    "record_created_by_id" UUID,
+    "owned_by_id" UUID
+  );
+
+DROP FUNCTION
+  create_external_ontology_id (
+    "base_url" TEXT,
+    "version" BIGINT,
+    "record_created_by_id" UUID,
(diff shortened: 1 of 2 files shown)


apps/hash-graph/postgres_migrations/V8__ontology_create_functions.sql
```diff
@@ -0,0 +1,159 @@
...
+  );
+
+CREATE FUNCTION
+  create_ontology_id (
+    "base_url" TEXT,
+    "version" BIGINT,
+    "record_created_by_id" UUID,
+    "resume_on_conflict" BOOLEAN,
+    "is_external" BOOLEAN
+  ) RETURNS TABLE (ontology_id UUID) AS $create_owned_ontology_id$
```
thehabbos007, line 32: Do you want to rename `create_owned_ontology_id` to `create_ontology_id`?

thehabbos007 (commented):
Looks good to me, just that one comment above :)

apps/hash-graph/postgres_migrations/V8__ontology_create_functions.sql
TimDiekmann: Thanks, fixed it in a16c09b152ec54bfdc80a9ce2763aeab4d9553c1 (but GitHub does not show the commit...)

thehabbos007 (approved):

thehabbos007 (approved):

PR #2699, opened 2023-06-30
Make Python worker aware of the type system

TimDiekmann:

We need the type system information in the temporal worker and preferable in a strongly typed fashion so we (a) can rely on Pydantic to parse it correctly and (b) will immediately get notified about incompatibilities.

## 🔗 Related links

- Part of [this Asana task](https://app.asana.com/0/1204727813960182/1204904869169432/f) _(internal)_

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2693
- https://github.com/hashintel/hash/pull/2695
- https://github.com/hashintel/hash/pull/2696 (CI will fail until merged)
- https://github.com/hashintel/hash/pull/2702 (CI is likely to fail until merged)

## 🔍 What does this change?

- Add `pydantic` and configuration to the temporal worker
- Implement the type system in Python so the schemas are strongly typed
- Add simple workflow to test the conversions

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

This is currently not tested automatically

## ❓ How to test this?

1.  Start external services (`temporal-setup`, `hash-temporal-worker-ts`, `hash-temporal-worker-py`, `graph`)
1.  Create an account or use an existing one (here: `00000000-0000-0000-0000-000000000000`)
1.  Confirm the following workflow returns the (recursive) person entity type:

    ```
    temporal workflow execute -t aipy --type getEntityType -i '{ "entityTypeId": "https://blockprotocol.org/@examples/types/entity-type/person/v/1", "actorId": "00000000-0000-0000-0000-000000000000" }'
    ```

1. Test other URLs and ontology types

## 📹 Demo

```
Running execution:
  WorkflowId  086ec237-66a6-40be-9096-75d57848a167                                                                                                    
  RunId       af1a7f0b-41c8-4379-bc17-72690a0860d1                                                                                                    
  Type        getEntityType                                                                                                                           
  Namespace   default                                                                                                                                 
  TaskQueue   aipy                                                                                                                                    
  Args        [{"actorId":"00000000-0000-0000-0000-000000000000","entityTypeId":"https://blockprotocol.org/@examples/types/entity-type/person/v/1"}]  

Progress:
  ID          Time                          Type                  
   1  2023-06-30T12:13:08Z  WorkflowExecutionStarted              
   2  2023-06-30T12:13:08Z  WorkflowTaskScheduled                 
   3  2023-06-30T12:13:08Z  WorkflowTaskStarted                   
   4  2023-06-30T12:13:08Z  WorkflowTaskCompleted                 
   5  2023-06-30T12:13:08Z  StartChildWorkflowExecutionInitiated  
   6  2023-06-30T12:13:08Z  ChildWorkflowExecutionStarted         
   7  2023-06-30T12:13:08Z  WorkflowTaskScheduled                 
   8  2023-06-30T12:13:08Z  WorkflowTaskStarted                   
   9  2023-06-30T12:13:08Z  WorkflowTaskCompleted                 
  10  2023-06-30T12:13:08Z  ChildWorkflowExecutionCompleted       
  11  2023-06-30T12:13:08Z  WorkflowTaskScheduled                 
  12  2023-06-30T12:13:08Z  WorkflowTaskStarted                   
  13  2023-06-30T12:13:08Z  WorkflowTaskCompleted                 
  14  2023-06-30T12:13:08Z  WorkflowExecutionCompleted            

Result:
  Run Time: 1 seconds
  Status: COMPLETED
  Output: [{"$id":"https://blockprotocol.org/@examples/types/entity-type/person/v/1","$schema":"https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type","description":"An extremely simplified representation of a person or human being.","kind":"entityType","links":{"https://blockprotocol.org/@examples/types/entity-type/employed-by/v/1":{"items":{"oneOf":[{"$ref":"https://blockprotocol.org/@examples/types/entity-type/company/v/1"}]},"minItems":0,"type":"array"}},"properties":{"https://blockprotocol.org/@blockprotocol/types/property-type/name/":{"$ref":"https://blockprotocol.org/@blockprotocol/types/property-type/name/v/1"},"https://blockprotocol.org/@examples/types/property-type/e-mail/":{"$ref":"https://blockprotocol.org/@examples/types/property-type/e-mail/v/1"}},"required":["https://blockprotocol.org/@blockprotocol/types/property-type/name/"],"title":"Person","type":"object"}]
```


apps/hash-ai-worker-py/app/__main__.py
```diff
@@ -7,8 +7,8 @@
 from temporalio.client import Client
 from temporalio.worker import Worker
 
-from app.activities import complete
-from app.workflows import DemoWorkflowPy
+from .activities import complete
+from .workflows import DataTypeWorkflow, EntityTypeWorkflow, PropertyTypeWorkflow
```
Alfred-Mountfield, line 11: Can I ask why you've moved to relative imports?

TimDiekmann: This must be due to some autoimports. I don't know if this is configurable but if it's not I'd prefer to keep it relative as otherwise it will become inconsistent. I'll do some research

apps/hash-ai-worker-py/app/typesystem.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class PropertyTypeSchema(OntologyTypeSchema, PropertyValueOneOf):
+    """A property type schema as defined by the Block Protocol.
+
+    see https://blockprotocol.org/types/modules/graph/0.3/schema/property-type
+    """
+
+
+class EntityTypeSchema(OntologyTypeSchema):
```
Alfred-Mountfield, line 125: Did you consider utilising TypeScript (perhaps as an activity or different workflow) and using the type system package in order to benefit from the existing validation functions?

It seems potentially problematic to fragment the implementation further and introduce a maintenance burden on another language implementation. We could also look at generating Python bindings into a compiled lib from Rust (rather than using Wasm).

I'm not suggesting these should be done now, but I would like to know if we considered the other approaches first

apps/hash-ai-worker-py/app/__main__.py
TimDiekmann: Seems to depend where VSCode thinks the root is. Made it consistent in [`cc45602` (#2699)](https://github.com/hashintel/hash/pull/2699/commits/cc4560250baa6258c56a2fddb8b5b14f6c736c6b)

apps/hash-ai-worker-py/app/typesystem.py
Alfred-Mountfield: I appreciate this isn't _just_ validation, it's also to have nicer typed objects.. so perhaps we'd end up doing something like this anyway. 

If these models were autogenerated from the schemas, do we want to add a comment saying what tool was used to manually do that, with a TODO to look at automating it perhaps? I'm guessing they were manually done though as `OntologyTypeSchema` seems like a manual simplification

TimDiekmann: Yes, I considered the Typescript package directly but this would not give me typed objects which is the main motivation for these changed. I was considering generating the types from the OpenAPI specs or from PyO3 but decided to not sink time into that - yet. I'm not sure if I need these schemas as I have to create a model based on the content in any case (will be a follow-up PR). That dynamic model will be a dynamic Pydantic model which also is able to emit a similar schema to the implemented schema above, which then can be used in OpenAI functions directly for both, inputs and outputs.

The models added in this PR are all hand-crafted.

Alfred-Mountfield: You're using types, schemas, and models a bit interchangeably and I can't entirely follow what you're talking about. For example, "I'm not sure if I need these schemas " — are you saying you're not sure if you need the models you've introduced in this PR? If so, why are we adding them currently?

> That dynamic model will be a dynamic Pydantic model which also is able to emit a similar schema to the implemented schema above

I don't really understand this, I'm guessing you're suggesting it's capable of generating JSON Schema (or some other schema) that you would be able to insert into the prompt (or functions interface) of an OpenAI call? 

If I'm correct in that, what do you mean "the implemented schema above", are you just on about the entity/property/data type meta-schemas?

TimDiekmann: >  If so, why are we adding them currently?

Because my current follow-up code relies on it and it was far easier to simply add a few model definitions and have the full type-safety than spending days to work around adding these and fail at runtime. Also, I can imagine that these models are useful to have if we want the LLM to generate ontology types.

> I'm guessing you're suggesting it's capable of generating JSON Schema [...] that you would be able to insert into the prompt (or functions interface) of an OpenAI call?

Yes, these schemas are then self-contained without external `$ref`s so OpenAI can properly understand them. Also, it enables reusing the OpenAI API by wrapping the API calls in a function which takes `BaseModel`.

> what do you mean "the implemented schema above"

The dynamic Pydantic models will emit a similar schema (almost identical) as the schemas we pass in to the models I implemented.

Alfred-Mountfield: Makes sense as to why you chose to implement them now to move forwards. 

However, I think it's potentially a dangerous route to go down for long, it means a lot of the prompt engineering will be tailored towards Pydantic (both in terms of the schema representation, and the validation errors), which seems to be in conflict with the design we were working towards with these workers having language-agnostic interfaces that could easily be swapped between TypeScript or Python or otherwise. Not suggesting you _shouldn't_ do it this way, perhaps aiming for generality is too slow, but something to be aware of as we had previously talked about the benefits of being forced to have language-agnostic definitions.

I'm also a little concerned about the layers of indirection that are now happening, and what that means for maintainability and future development. We have the Type System spec and meta-schemas which are manually implemented as Pydantic models, which are then used to produce JSON schemas again (generally a lossy process), which we then are crafting LLM flows towards, and then we get outputs of those, Pydantic validates them, we then transform them to JSON, and they hit our other representation which is validated in a different way. That makes the Pydantic models the source of truth, which is an issue if they don't adequately capture all of the necessary validations which will then break when they hit our universal implementation (either in the Rust side of the graph, or the TypeScript side of the HASH API, etc.).

Something worth noting is that our TypeScript types for these do not mix concerns with _validation_ (like Pydantic does), they instead provide a nice typed interface if you want to do things like `foo.title` or `foo.allOf`, but the validation is still elsewhere. Perhaps something to consider for this, if the main goal right now is to be able to do things like `foo['title']` in Python and have it typed for using when you're trying to insert it into a prompt or something. You could consider a `TypedDict` or similar as the JSON representation _is_ consistent across all languages. If you really want to use it for more than that (e.g. validation and or schema generation) that's fair enough, I'd just be aware of what that could imply later on.

Alfred-Mountfield: > The dynamic Pydantic models will emit a similar schema (almost identical) as the schemas we pass in to the models I implemented.

I still don't understand what you mean here. Are you saying that you pass in some entity type, say `Person`, and then after you have `EntityTypeSchema(person)` you can get it to give you the JSON back, and it's "similar"? Doesn't it _**have**_ to be exactly the same?

TimDiekmann: I agree that the amount of indirection is not ideal but also not a concern currently. Striving for perfection and generalize everything will just not work in time. We need to get this LLM stuff to work _now_. The reason I did it this way is that OpenAI _understands_ the schemas so they must not be exactly the same. `allOf = []` for example will immediately fail with a `400` error in OpenAI so there needs to be a lossy process. Also things like `$schema` should not appear as they don't give any benefits but only increase the token limit (they could even confuse the AI). And, obviously, `$ref` needs to be resolvable to be either inlined or by having a `#/definitions/` reference instead of a URL. In addition, using another schema for the LLM interface is required to set `additional_properties = false`, allow `required` to be omitted, or at some point to insert more properties which the LLM should generate alongside.

This does not make Pydantic the source of truths, it's just defining the API interface for OpenAI.

TimDiekmann: To avoid confusion: the models implemented in this PR _are_ a native implementation of the typesystem but just for the sake of simplicity.

apps/hash-ai-worker-py/app/typesystem/data_type.py
```diff
@@ -0,0 +1,20 @@
...
+    ref: str = Field(..., alias="$ref")
+
+
+class DataTypeSchema(OntologyTypeSchema, extra=Extra.allow):
+    """A data type schema as defined by the Block Protocol.
+
+    see https://blockprotocol.org/types/modules/graph/0.3/schema/data-type
+    """
+
+    ty: str = Field(..., alias="type")
```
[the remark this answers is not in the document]
TimDiekmann: Yes, good spot!

indietyp (dismissed):

indietyp (approved):
Thank you for adding the comment!

PR #2702, opened 2023-06-30
Disable caching of `.venv`

TimDiekmann:

We have a caching setup for `.venv`, however, it does not seems to work properly with different entries for the same output. At least [a recent Action failed due to this](https://github.com/hashintel/hash/actions/runs/5422658218/jobs/9859681702?pr=2699). To avoid spending much time on figuring out this, this PR removes caching of `.venv`.

## 🔗 Related links

- #2699

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-ai-worker-py/turbo.json
+++ b/apps/hash-ai-worker-py/turbo.json
@@ -24,32 +24,25 @@
       "cache": false
     },
     "fix:black": {
-      "dependsOn": ["poetry:install-lint-tools"],
       "inputs": ["./**/*.py", "pyproject.toml"]
     },
     "fix:ruff": {
-      "dependsOn": ["poetry:install-lint-tools"],
       "inputs": ["./**/*.py", "pyproject.toml"]
     },
     "fix:lock-files": {
-      "dependsOn": ["poetry:install-lint-tools"],
       "inputs": ["poetry.lock", "pyproject.toml"],
       "outputs": ["poetry.lock"]
     },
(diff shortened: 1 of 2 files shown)


apps/hash-ai-worker-py/turbo.json
```diff
@@ -6,24 +6,19 @@
       "persistent": true
     },
     "poetry:venv": {
-      "outputs": ["./.venv/**"],
-      "inputs": ["pyproject.toml", "poetry.lock"]
+      "cache": false
```
Alfred-Mountfield, line 9: I understand wanting to avoid debugging this issue, but I think this will cause a reinstall on every single Python turbo command (such as `black`, `ruff`, etc.)

Do we perhaps want to remove the `dependsOn` from those, or remove the `cache` here and rely on it checking the whole workspace for cache invalidation? etc.

TimDiekmann: Removing the `dependsOn` might be sensible, yes, but I don't understand what you mean by checking the whole workspace for cache invalidation. Do you mean add `pyproject.toml` and `poetry.lock` to the global invalidation logic? This _could_ work as well.

TimDiekmann: This does not work reliably, I removed the dependencies on the install script. For `yarn` we have the same behavior: we have to call `yarn install`.

CiaranMn (approved):

PR #2711, opened 2023-07-03
[H-41] Save less frequently

CiaranMn:

Closes https://github.com/hashintel/hash/pull/2606 in favour of cherry-picking the commit there Nate introduced to reduce save frequency.

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing -->

The changes in this PR:

- [x] are internal and do not require a docs change -->

The changes in this PR:

- [x] do not affect the execution graph -->

## ❓ How to test this?

1.  Check that saving in a page still works

### DIFF
--- a/apps/hash-frontend/src/blocks/page/create-editor-view.ts
+++ b/apps/hash-frontend/src/blocks/page/create-editor-view.ts
@@ -15,7 +15,7 @@ import { save } from "@local/hash-isomorphic-utils/save";
 import { AccountId, EntityId, OwnedById } from "@local/hash-subgraph";
 import { debounce } from "lodash";
 // import applyDevTools from "prosemirror-dev-tools";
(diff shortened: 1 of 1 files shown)


nonparibus:
@TimDiekmann This appears to have deployed `hashdotdev` when it didn't need to, while not deploying `hash` (which it should have). Any idea why?

TimDiekmann:
The first commit on a branch will always try to deploy (except `hcore-plugin-pm` which uses a different rule). Any further commit should ignore the deployment.

TimDiekmann:
Investigated further: The first commit triggered a HASH deployment. The second was canceled because it didn't change the code to trigger another deployment. So everything's good 👍

benwerner01 (approved):
Thanks Ciaran, looks good to me and saving is working for me locally 👍

PR #2716, opened 2023-07-04
Reduce number + size of queries

CiaranMn:
## 🔗 Related links

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

This PR:

- [ ] **⚠️ This is a placeholder, please uncomment at least one of the lines following this one and then delete this. ⚠️**

The changes in this PR:

- [ ] **⚠️ This is a placeholder, please uncomment at least one of the lines following this one and then delete this. ⚠️**

The changes in this PR:

- [ ] **⚠️ This is a placeholder, please uncomment at least one of the lines following this one and then delete this. ⚠️**

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Try X
1.  Confirm that Y


apps/hash-frontend/src/components/hooks/use-users.ts
```diff
@@ -37,9 +37,9 @@ export const useUsers = (
       constrainsPropertiesOn: { outgoing: 0 },
       constrainsLinksOn: { outgoing: 0 },
       constrainsLinkDestinationsOn: { outgoing: 0 },
-      isOfType: { outgoing: 1 },
-      hasLeftEntity: { incoming: 1, outgoing: 1 },
-      hasRightEntity: { incoming: 1, outgoing: 1 },
```
TimDiekmann, line 42: Just to check if this is the correct way: For a given entity this will traverse the ***outgoing*** link. So in the following example, it will also discover `B` when querying for `A`:
```
Entity A --Link-> Entity B
```

apps/hash-frontend/src/components/hooks/use-orgs.ts
```diff
@@ -42,9 +42,9 @@ export const useOrgs = (
       constrainsPropertiesOn: { outgoing: 0 },
       constrainsLinksOn: { outgoing: 0 },
       constrainsLinkDestinationsOn: { outgoing: 0 },
-      isOfType: { outgoing: 1 },
-      hasLeftEntity: { incoming: 1, outgoing: 1 },
-      hasRightEntity: { incoming: 1, outgoing: 1 },
```
TimDiekmann, line 47: Just to check if this is the correct way: For a given entity this will traverse the ***incoming*** link. So in the following example, it will also discover `A` when querying for `B`:
```
Entity A --Link-> Entity B
```

TimDiekmann (commented):
PR looks good to me, I just have two questions to make sure we get the depths parameters in the correct way. 

It would be good if @benwerner01 could take a look as well.

apps/hash-frontend/src/components/hooks/use-users.ts
CiaranMn: Yes, that's correct. For users we want to retrieve orgs they are a member of. This is modelled as follows:

```
User -> OrgMembership -> Org
```

apps/hash-frontend/src/components/hooks/use-orgs.ts
CiaranMn: Yes, that's correct. For orgs we want to retrieve members (users). This is modelled as follows:

```
User -> OrgMembership -> Org
```

apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx
```diff
@@ -200,29 +194,22 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async ({
   return {
     props: {
       pageWorkspace,
-      blocks: fetchedBlocks,
+      blocks: [],
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx:197]: Does this mean that `UserBlocksProvider` is populated client side in a child component? Or is the value no longer populated at all?

benwerner01 (dismissed):
Thanks Ciaran, these changes look good and are working for me 👍

apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx:197]: There's an effect inside `UserBlocksProvider` that fetches the list of blocks from the BP API and adds it to the initial list.

Since the server-side code was fetching default blocks (paragraph etc) which will appear in the BP API anyway, they will be fetched client-side in this effect.

I did notice while getting the link for this that it was making requests for block metadata serially. I've updated this in dba74666218679ed183e1cf5860f1e32b1e8a51a

benwerner01 (approved):

CiaranMn:
@benwerner01 I had to fix a bug that the tests flagged – did so in https://github.com/hashintel/hash/pull/2716/commits/4962a3e79945d649d4630f536d76e82e5cb5a19b – can I get a re-review please?

benwerner01 (approved):

PR #2743, opened 2023-07-10
H-100: Add scaffolding for `@local/hash-graph-types`

TimDiekmann:

This adds the basic scaffolding for the implementation of schema validation, where the majority of the logic of `hash-ai-worker-py` will live.

## 🔍 What does this change?

- Add basic setup for the Python library
- Add a dependency for `hash-ai-worker-py`
- Rework some script invocations to make it work with internal dependencies

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-graph-types/LICENSE.md
```diff
@@ -0,0 +1,3 @@
+# License
```
nonparibus, line 1: @TimDiekmann Shouldn't this be AGPL?

libs/@local/hash-graph-types/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "hash-graph-types"
 version = "0.0.0"
 description = "Type definitions for the type system and entity validation based on schemas"
-license = "AGPL-3.0-only"
+license = "MIT OR Apache-2.0"
```
nonparibus, line 5: Why is this not AGPL like the HASH application?

libs/@local/hash-graph-types/LICENSE.md
TimDiekmann: I was assuming as we want it at the awesome-blocks page and this is only a small library that "only" transforms some JSON schemas we were about to publicize it under MIT/Apache, especially as the [type-system is also MIT/Apache](https://github.com/blockprotocol/blockprotocol/blob/main/libs/%40blockprotocol/type-system/crate/LICENSE.md). Still, I was about to check with you as soon as the PR is ready for review.

libs/@local/hash-graph-types/pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+description = "Type definitions for the type system and entity validation based on schemas"
+license = "MIT OR Apache-2.0"
+authors = ["HASH"]
+readme = "README.md"
+packages = [{ include = "graph_types" }]
+
+[tool.poetry.dependencies]
+python = "^3.11"
+
+pydantic = "^2.0"
```
TimDiekmann, line 13: Added in this PR to add the `pydantic.mypy` plugin

libs/@local/hash-graph-types/LICENSE.md
TimDiekmann: You want me to change it to AGPL?

nonparibus: Let's make this AGPL for now, as it's only at the point of being `@local` anyway (even if we are planning to link from `awesome-blocks`). We can can examine this in more depth with a view to liberalizing it later. Thanks!

libs/@local/hash-graph-types/README.md
```diff
@@ -0,0 +1,3 @@
+# HASH Graph types
+
+Type definitions for the type system and entity validation based on schemas.
```
nonparibus, line 3: I wonder if we can make this a bit more descriptive. @CiaranMn for thoughts.

libs/@local/hash-graph-types/graph_types/__init__.py
```diff
@@ -0,0 +1 @@
+"""Replicates the Block Protocol type system to be used in Python."""
```
nonparibus, line 1: "to be used" -> "for use"?

```suggestion
"""Replicates the Block Protocol type system for use in Python."""
```

nonparibus (approved):
Legal ✅ but a couple of comments (CCd @CiaranMn)

libs/@local/hash-graph-types/README.md
nonparibus: We should also update the package metadata to reflect the revised description here when added.

indietyp (dismissed):
LGTM, thank you! 🎉

indietyp (approved):

indietyp (approved):

indietyp:
The build simplifications look great btw!

nonparibus (approved):

libs/@local/hash-graph-types/README.md
TimDiekmann: I propose to do this in a follow-up so we can get this PR out of the way.

TimDiekmann:
> The build simplifications look great btw!

I would have preferred to do this in a separate PR but it was kind of required. Thanks, though!

nonparibus (approved):
Thanks, @TimDiekmann!

PR #2746, opened 2023-07-10
H-101: Move type system schema creation from `@apps/hash-ai-worker-py` to `@local/hash-graph-types`

TimDiekmann:

#2743 introduced the `hash-graph-types` package, this moves the schema definitions to that PR.

## 🚫 Blocked by

- #2743

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graph-types/graph_types/__init__.py
+++ b/libs/@local/hash-graph-types/graph_types/__init__.py
@@ -4,11 +4,11 @@ from uuid import UUID
 
 # This solution is not great as it _could_ lead to circular imports. However,
 # it is the best solution we have for now. The alternative is to have a single
-# file that contains all of the schemas, but that is not very maintainable.
+# file that contains all the schemas, but that is not very maintainable.
 #
 # If we run into issues with circular imports, we can refactor this to use
 # direct imports. For example, instead of importing `DataTypeSchema` from
-# `typesystem`, we can import it from `typesystem.data_type`.
+# `graph_types`, we can import it from `graph_types.data_type`.
 from .data_type import DataTypeReference as DataTypeReference
 from .data_type import DataTypeSchema as DataTypeSchema
 from .entity_type import EntityTypeReference as EntityTypeReference
@@ -17,7 +17,7 @@ from .property_type import PropertyTypeReference as PropertyTypeReference
 from .property_type import PropertyTypeSchema as PropertyTypeSchema
 
 
-class GraphApiProtocol(Protocol):
+class GraphAPIProtocol(Protocol):
     """Defines the interface for a graph API."""
 
     async def get_data_type(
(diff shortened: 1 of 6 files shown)


libs/@local/hash-graph-types/graph_types/__init__.py
```diff
@@ -1 +1,60 @@
...
+from typing import Protocol
+from uuid import UUID
+
+# This solution is not great as it _could_ lead to circular imports. However,
+# it is the best solution we have for now. The alternative is to have a single
+# file that contains all of the schemas, but that is not very maintainable.
+#
+# If we run into issues with circular imports, we can refactor this to use
+# direct imports. For example, instead of importing `DataTypeSchema` from
+# `typesystem`, we can import it from `typesystem.data_type`.
```
indietyp, line 11: comment needs adjustment

libs/@local/hash-graph-types/graph_types/_schema.py
```diff
@@ -13,13 +14,19 @@
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import CoreSchema, core_schema
 
+if typing.TYPE_CHECKING:
+    from . import GraphApiProtocol
+
+G = TypeVar("G", bound="GraphApiProtocol")
```
indietyp, line 20: If you're using a protocol, you don't need a TypeVar; you can use the class name as the type hint.

Both approaches are correct and valid.

libs/@local/hash-graph-types/graph_types/__init__.py
```diff
@@ -1 +1,60 @@
...
+from .data_type import DataTypeReference as DataTypeReference
+from .data_type import DataTypeSchema as DataTypeSchema
+from .entity_type import EntityTypeReference as EntityTypeReference
+from .entity_type import EntityTypeSchema as EntityTypeSchema
+from .property_type import PropertyTypeReference as PropertyTypeReference
+from .property_type import PropertyTypeSchema as PropertyTypeSchema
+
+
+class GraphApiProtocol(Protocol):
```
indietyp, line 20: Naming, Python customs are to not lowercase abbreviations, so:

```
GraphAPIProtocol
```

Examples: https://docs.python.org/3/library/http.html, https://fastapi.tiangolo.com/, (and there are plenty more out there)

To quote the styling bible PEP8 (https://peps.python.org/pep-0008/#descriptive-naming-styles)

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

indietyp (changes requested):
some minor comments

indietyp (dismissed):

indietyp (approved):

PR #2756, opened 2023-07-12
H-133: Fix circular import for `hash-graph-types` package

TimDiekmann:

The GraphAPIProtocol type is defined in __init__ after the schemas are imported but the schema requires the protocol so it results in a circular dependency. By conditionally import it this will solve running the app.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-ai-worker-py/poetry.lock
+++ b/apps/hash-ai-worker-py/poetry.lock
@@ -418,9 +418,7 @@ version = "0.0.0"
 description = "Type definitions for the type system and entity validation based on schemas"
 optional = false
 python-versions = ">=3.11,<4.0"
-files = [
-    {file = "hash_graph_types-0.0.0-py3-none-any.whl", hash = "sha256:c6f1eb929e115f93a585a86eff6a0821112c6e3d93053decbeb886e3854d7595"},
-]
+files = []
 
 [package.dependencies]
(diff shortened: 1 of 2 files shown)


apps/hash-ai-worker-py/poetry.lock
```diff
@@ -418,9 +418,7 @@ version = "0.0.0"
 description = "Type definitions for the type system and entity validation based on schemas"
 optional = false
 python-versions = ">=3.11,<4.0"
-files = [
-    {file = "hash_graph_types-0.0.0-py3-none-any.whl", hash = "sha256:c03c391cf843e36bc392c812fb9924286377c523ccd294b6d3eeff08119c2e42"},
-]
+files = []
```
indietyp, line 421: Why is that here? I assume you forgot to lock again?

indietyp (changes requested):
Small thing otherwise LGTM

(Python circular dependency hell here we go^^)

apps/hash-ai-worker-py/poetry.lock
TimDiekmann: Because this way it's working without messing around with the lock file. 🤷🏻‍♂️ 
I will most likely remove the wheel, anyway as it's super annoying to deal with.

indietyp (approved):

PR #2757, opened 2023-07-12
Reduce frontend shared JS size. Upgrade Ory clients

CiaranMn:

In starting to look at frontend performance, I noticed that every page shared a common JS bundle of almost 1MB. 

This is because the root `_app` page imported from some places it shouldn't, including the main Page page.

This PR moves the small functions it needed from those places into a shared location, halving the shared JS from `990kB` to `487kB`. There is more that could be done here, but these are the quick wins.

It also disables some Sentry stuff to cut how much it contributes to the bundle

Also upgrades Ory clients as a drive-by (I was doing this to debug Ory issues, although it doesn't seem to have fixed them)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  Check things still work. Hopefully this will have an impact on perceived page load times once deployed – it's hard to test locally.

### Before
Screenshot 2023-07-12 at 18 04 27

### After
Screenshot 2023-07-12 at 18 02 52

### DIFF
--- a/apps/hash-api/package.json
+++ b/apps/hash-api/package.json
@@ -40,7 +40,7 @@
     "@opentelemetry/resources": "1.0",
     "@opentelemetry/sdk-trace-base": "1.0",
     "@opentelemetry/sdk-trace-node": "1.0",
-    "@ory/client": "1.1.33",
+    "@ory/client": "1.1.41",
     "@ory/kratos-client": "0.13.1",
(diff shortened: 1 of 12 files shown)


benwerner01 (approved):
Thanks Ciaran, looks good and is working for me locally as expected 👍

PR #2726, opened 2023-07-07
H-87: Upgrade Ory Kratos

CiaranMn:
## 🌟 What is the purpose of this PR?

As part of investigating performance of the deployed app, I noticed that calls to Ory's `whoami` can be very slow (sometimes 6-7s). 

The `0.13` release of Ory Kratos [specifically mentions](https://github.com/ory/kratos/releases/tag/v0.13.0) optimising `whoami` calls, so I'm upgrading to see what benefit this brings.

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🐾 Next steps

- If this does not solve the `whoami` performance issues, my next target will be looking at [the post-registration hook](https://github.com/hashintel/hash/blob/main/apps/hash-external-services/kratos/hooks/after.registration.jsonnet) in case that is the bottleneck somehow.

## ❓ How to test this?

1.  Checkout the branch
1. Rebuild `external-services` fully (i.e. get rid of the underlying Ory Kratos image at least)
1. Check login / logout still works

PR #2761, opened 2023-07-14
H-98: Fix user registration, and treat incomplete signups properly

CiaranMn:

This PR fixes user creation, which was broken by the upgrade to Ory Kratos `0.13` in #2726: the behaviour for post-registration hooks changed, as described in the linked issue (see diff).

This revealed another error, which was the Command Bar being loaded if an 'authenticated user' was present, even if the authenticated user hadn't completed signup, and was therefore missing some other properties the command bar relied on. I've also fixed a few places where we weren't checking if a user had completed signup before setting readonly mode, although in practice we redirect users to `/signup` if they are authenticated but need to complete signup.

## 🔍 What does this change?

- Update the Ory Kratos configuration – I've removed a deprecated field (`can_interrupt`). I'm not sure what it was needed for but since it will be removed in a future Kratos version we will need another way of doing it if we were relying on it (@benwerner01 any ideas?)
- Check if a user has completed signup before showing the command bar or showing editing UI.

This PR:

The changes in this PR:
- [x] is internal and do not require a docs change -->

The changes in this PR:

- [x] do not affect the execution graph -->

## ❓ How to test this?

1.  Check that you can create a new user


apps/hash-external-services/kratos/kratos.dev.yml
```diff
@@ -53,7 +53,10 @@ selfservice:
           hooks:
             - hook: web_hook
               config:
-                can_interrupt: true
```
benwerner01, line 56: I think this may have been added in the hopes to somehow interrupt the flow if an error occurs in the webhook. Not sure what exactly kratos would've done in this circumstance with the field set to try, but if the field is being deprecated anyway we can worry about handling this properly down the line.

benwerner01 (approved):
Thanks Ciaran for fixing this, registration is working for me locally again 👍

PR #2765, opened 2023-07-17
H-123: Fix loading an entity into a block

CiaranMn:

Fixes a few bugs / broken features – one commit each.

## 🔍 What does this change?

- Makes the 'add / swap entity in block' context menu item functional. Currently limited to the same entity type as the block (any version)
- Sorts the block list alphabetically in the block selector
- Fixes the z-index of the block selector (it could be covered by the page header)
- Use 'Untitled' as a fallback in the HTML title

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  Load two blocks of the same type in a page. Put data in one. Try loading that entity into the other

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/34a94d56-b88b-40c7-a4b7-6dac50bf81bd


benwerner01 (approved):

PR #2766, opened 2023-07-17
H-138: Add GraphQL resolver stub for inferring entities from text

CiaranMn:

This PR adds a GraphQL mutation for inferring entities from text input.

It does not actually do anything yet – it is a placeholder for @TimDiekmann to call his 'infer entities' implementation inside once that work is complete.

Some points to note:
- We already use `DraftEntity` as a type elsewhere which is why I went for `ProposedEntity` here
- I've returned the proposed entities under an `entities` key in case we want to add other metadata in the return

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

1. @TimDiekmann adds the implementation inside this
2. @CiaranMn adds a way of using it from the frontend

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1.  Visit `http://localhost:5001/graphql` with the app running
2. Try this 
```graphql
mutation {
  inferEntities(textInput: "Hello") {
    entities {
      entityTypeId
      properties
    }
  }
}
```


libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
```diff
@@ -39,6 +39,28 @@ export const entityTypedef = gql`
...
   }
 
+  """
+  An entity proposed for creation. The suggested data can be used in further calls, e.g. to createEntity
+  """
+  type ProposedEntity {
+    """
+    The suggested type of the proposed entity.
+    """
+    entityTypeId: VersionedUrl!
```
TimDiekmann, line 49: The entity type id is not going to be suggested. Entities are going to be suggested based on entity types.

TimDiekmann (changes requested):

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -306,3 +308,25 @@ export const archiveEntityResolver: ResolverFn<
 
   return true;
 };
+
+// @todo replace this with the actual implementation
+const inferEntitiesPlaceholder = async (_textInput: string) => {
```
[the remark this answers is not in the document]
CiaranMn: I've updated this in https://github.com/hashintel/hash/pull/2766/commits/7a9f8e112540e6259ab7d67970cd5af0dd84b7bd

libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
CiaranMn: I would say it's still a suggested entityTypeId because a user may choose to assign a different type to an entity, where there are multiple valid types for it. But I've removed 'suggested' from this comment anyway

TimDiekmann (approved):
Thanks, Ciaran, this looks good and I'm looking forward to implement the functionality 🚀

PR #2767, opened 2023-07-17
H-89: Speed up comment request resolution

CiaranMn:

This PR reduces the time it takes to resolve requests to create, edit and resolve comments, by:
1. Removing unnecessary requests: where we are given an `entityId` by the caller, we were fetching the entity, and then passing the entity into another function, where the other function only used the `entityId` in the first place. This updates such functions to only require the `entityId`.
2. Parallelising requests where they can be – creating a comment previously involved running 5 requests serially. These can be done in 2 batches.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- There's a backend integration test for creating a comment.

## ❓ How to test this?

1.  Visit a page
2. Check that comment operations still work: creating, replying, editing, deleting, resolving a thread

### DIFF
--- a/apps/hash-api/src/graph/knowledge/primitive/entity.ts
+++ b/apps/hash-api/src/graph/knowledge/primitive/entity.ts
@@ -435,21 +435,19 @@ export const updateEntityProperty: ImpureGraphFunction<
  */
 export const getEntityIncomingLinks: ImpureGraphFunction<
   {
-    entity: Entity;
+    entityId: EntityId;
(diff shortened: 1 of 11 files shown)


TimDiekmann (approved):
Thanks, Ciaran! I also noted these locations when I touched the files a few months ago but didn't want to do it as a drive-by and then forgot to do that as a follow-up. I don't think it will be a huge speed-up but it will definitely add up.

PR #2769, opened 2023-07-18
H-149: Hide hotkey cheat sheet

CiaranMn:

There's a command bar which appears if you hit cmd/ctrl + K. There's also a hotkey cheat sheet which appears if you press ?. This happens even if you press ? inside contenteditables, which is not the desired behaviour.

This PR fixes that bug, but also hides the cheat sheet entirely, as it doesn't serve much purpose at the moment, since the hot keys are already shown on the command bar. Its use would be to be able to display _all_ hot keys, even ones which are not applicable in the current context, but since we don't have many at the moment and the cheat sheet requires styling, it's not worth it.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
+++ b/apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
@@ -68,7 +68,6 @@ export const PageSidebar: FunctionComponent = () => {
         icon={faHome}
         title="Home"
         href="/"
-        tooltipTitle="View your inbox and latest activity"
         active={router.pathname === "/[shortname]"}
       />
       {/* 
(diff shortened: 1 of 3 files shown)


apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
```diff
@@ -68,7 +68,6 @@ export const PageSidebar: FunctionComponent = () => {
         icon={faHome}
         title="Home"
         href="/"
-        tooltipTitle="View your inbox and latest activity"
```
CiaranMn, line 71: Driveby: remove inaccurate tooltip

nonparibus (dismissed):

nonparibus (approved):

PR #2776, opened 2023-07-19
H-167: `hash-graph-types`: Add marker classes and unify interface

indietyp:

Using some black-magic type jujitsu, I was able to add marker classes to all generated types; the only recognizable change is that `PropertyType`s are now also RootModels, the type information has been unified as well, and the different `create_model` methods now return `*Type` classes instead of any, making it a lot easier to work with them.

## 🔍 What does this change?

* add base marker classes
* unify interface

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

* I want to add human usable names to the different properties, instead of the current ones.
* Add tests


libs/@local/hash-graph-types/graph_types/_schema.py
```diff
@@ -91,23 +105,16 @@ async def create_model(
         *,
         actor_id: UUID,
         graph: "GraphAPIProtocol",
-    ) -> Annotated[Any, ...]:
-        ty = await self.items.create_model(actor_id=actor_id, graph=graph)
+    ) -> type[list[T]]:
```
indietyp, line 108: Reminder for myself. This annotation is incorrect

libs/@local/hash-graph-types/graph_types/data_type.py
```diff
@@ -96,57 +95,27 @@ async def create_data_type(
...
-                return json_schema
-
-            @classmethod
-            def validate_const(cls, v: DataType) -> None:
-                if v != const:
-                    msg = f"Value must be {const}"
-                    raise ValueError(msg)
-
-        return Annotated[self._type(), DataTypeAnnotation]
+        type_ = Literal[const] if const is not None else self._type()
```
TimDiekmann, line 110: `const` is not literal. This is also why Pylance is complaining on this line:
image
In fact, I tried using `Literal` before but this is not allowed as the value is not a literal but could literally (🙃) be anything.
See https://github.com/pydantic/pydantic/issues/6346

libs/@local/hash-graph-types/graph_types/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    identifier: str
+    schema_url: str
+    title: str
+    description: str | None = None
+    kind: str
+
+
+class Type(BaseModel, ABC):
+    """Base class for all graph types."""
```
TimDiekmann, line 24: This is only true for ontology types.

libs/@local/hash-graph-types/graph_types/_schema.py
TimDiekmann: What would be the correct annotation?

libs/@local/hash-graph-types/graph_types/data_type.py
```diff
@@ -49,15 +46,15 @@ async def create_model(
...
 
         schema = await graph.get_data_type(self.ref, actor_id=actor_id)
+        model = await schema.create_data_type(actor_id=actor_id, graph=graph)
+
+        # Potential race condition here, another coroutine could have created the model
+        # before we did on the previous await statement.
+        # Just check if we have it in the cache again, this way we won't
+        # accidentally create the same model twice.
+        if cached := self.cache.get(self.ref):
+            return cached
```
TimDiekmann, line 56: I tried this before and this does not work as expected. For example for the entity type `https://blockprotocol.org/@examples/types/entity-type/person/v/1` this just generated this schema:
```json
{
  "$defs": {
    "https_blockprotocol_org_blockprotocol_types_property_type_name_v_1": {
      "$id": "https://blockprotocol.org/@blockprotocol/types/property-type/name/v/1",
      "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/property-type",
      "allOf": [
        {
          "$ref": "#/$defs/pydantic__main__https_blockprotocol_org_blockprotocol_types_data_type_text_v_1_4578735440"
        }
      ],
      "description": "A word or set of words by which something is known, addressed, or referred to.",
      "kind": "propertyType",
      "title": "Name"
    },
    "https_blockprotocol_org_examples_types_property_type_e_mail_v_1": {
      "$id": "https://blockprotocol.org/@examples/types/property-type/e-mail/v/1",
      "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/property-type",
      "allOf": [
        {
          "$ref": "#/$defs/pydantic__main__https_blockprotocol_org_blockprotocol_types_data_type_text_v_1_4572375792"
        }
      ],
      "description": "An e-mail address.",
      "kind": "propertyType",
      "title": "E-Mail"
    },
    "pydantic__main__https_blockprotocol_org_blockprotocol_types_data_type_text_v_1_4572375792": {
      "$id": "https://blockprotocol.org/@blockprotocol/types/data-type/text/v/1",
      "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
      "description": "An ordered sequence of characters",
      "kind": "dataType",
      "title": "Text",
      "type": "string"
    },
    "pydantic__main__https_blockprotocol_org_blockprotocol_types_data_type_text_v_1_4578735440": {
      "$id": "https://blockprotocol.org/@blockprotocol/types/data-type/text/v/1",
      "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
      "description": "An ordered sequence of characters",
      "kind": "dataType",
      "title": "Text",
      "type": "string"
    }
  },
  "$id": "https://blockprotocol.org/@examples/types/entity-type/person/v/1",
  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
  "additionalProperties": false,
  "description": "An extremely simplified representation of a person or human being.",
  "kind": "entityType",
  "properties": {
    "https://blockprotocol.org/@examples/types/property-type/e-mail/": {
      "$ref": "#/$defs/https_blockprotocol_org_examples_types_property_type_e_mail_v_1"
    },
    "https://blockprotocol.org/@blockprotocol/types/property-type/name/": {
      "$ref": "#/$defs/https_blockprotocol_org_blockprotocol_types_property_type_name_v_1"
    }
  },
  "required": [
    "https://blockprotocol.org/@blockprotocol/types/property-type/name/"
  ],
  "title": "Person",
  "type": "object"
}
```

The `text` data type appears twice here.

libs/@local/hash-graph-types/graph_types/data_type.py
```diff
@@ -96,57 +95,27 @@ async def create_data_type(
...
-        return Annotated[self._type(), DataTypeAnnotation]
+        type_ = Literal[const] if const is not None else self._type()
+
+        base = type("Base", (DataTypeBase,), {"info": self.type_info()})
+
+        model = create_model(
+            slugify(self.identifier, regex_pattern=r"[^a-z0-9_]+", separator="_"),
+            __base__=(base, RootModel),
+            root=(type_, Field(...)),
+        )
```
TimDiekmann, line 118: Is it intended that two models are created now?

TimDiekmann (changes requested):

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: This simply moved the `create_model` call from `DataTypeReference` to `DataType`

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: I unified the cache logic in [9d30d65](https://github.com/hashintel/hash/pull/2776/commits/9d30d651ed2a11e9333d3d57913227bf72d764e8), could you try again?

libs/@local/hash-graph-types/graph_types/base.py
indietyp: fixed in [d2b08ad](https://github.com/hashintel/hash/pull/2776/commits/d2b08adb062fdb15c189843dbe796342ebe79776)

libs/@local/hash-graph-types/graph_types/_schema.py
indietyp: `type[list[BaseModel]]` as we don't know what the call `create_model` will return

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: I tend to disagree. See this snippet:

```
from typing import Literal

from pydantic import TypeAdapter

a = "abc"

x = Literal[a]
print(x)

adapter = TypeAdapter(x)
adapter.validate_python("abcd")
```

```
Traceback (most recent call last):
  File "/Users/bmahmoud/Library/Application Support/JetBrains/PyCharm2023.1/scratches/scratch_41.py", line 11, in <module>
    adapter.validate_python("abcd")
  File "/Users/bmahmoud/Sync/projects/scionics/hash/client/.venv/lib/python3.11/site-packages/pydantic/type_adapter.py", line 206, in validate_python
    return self.validator.validate_python(__object, strict=strict, from_attributes=from_attributes, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for literal['abc']
  Input should be 'abc' [type=literal_error, input_value='abcd', input_type=str]
    For further information visit https://errors.pydantic.dev/2.0.3/v/literal_error
```

indietyp: but it will fail if the value is a non-primitive during construction. Which is a bug anyway

indietyp: fixed in 9d30d651ed2a11e9333d3d57913227bf72d764e8

libs/@local/hash-graph-types/graph_types/_const.py
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError(msg)
+
+        return v
+
+
+_Undefined = object()
+
+
+@overload
+def constant(const_: T, /) -> type[T]:
```
TimDiekmann, line 50: What does the `/` mean?

libs/@local/hash-graph-types/graph_types/_const.py
```diff
@@ -0,0 +1,70 @@
...
+    ...
+
+
+@overload
+def constant(type_: type[T], const_: T, /) -> type[T]:
+    ...
+
+
+def constant(type_: type[T] | T, const_: T | object = _Undefined, /) -> type[T]:
```
TimDiekmann, line 59: I don't really understand this signature. Why `type_` can be `type[T]` **or** `T`? Why `const_` is allowed to be `object`?
What was wrong with the previous approach - what is the difference?

libs/@local/hash-graph-types/graph_types/_const.py
```diff
@@ -0,0 +1,70 @@
+from typing import Annotated, Any, TypeVar, cast, overload, Literal
```
TimDiekmann, line 1: For what `Literal` is used here?

libs/@local/hash-graph-types/graph_types/_schema.py
```diff
@@ -1,17 +1,22 @@
...
-    GetCoreSchemaHandler,
     GetJsonSchemaHandler,
+    conlist,
+    create_model,
 )
 from pydantic.json_schema import JsonSchemaValue
-from pydantic_core import CoreSchema, core_schema
+from pydantic_core import CoreSchema
+
+from .base import OntologyType, OntologyTypeInfo
```
TimDiekmann, line 19: Is `OntologyType` required for some types? Otherwise I think we can remove the import?
```suggestion
from .base import OntologyTypeInfo
```

libs/@local/hash-graph-types/graph_types/data_type.py
```diff
@@ -96,57 +96,27 @@ async def create_data_type(
...
-
-            @classmethod
-            def validate_const(cls, v: DataType) -> None:
-                if v != const:
-                    msg = f"Value must be {const}"
-                    raise ValueError(msg)
-
-        return Annotated[self._type(), DataTypeAnnotation]
+        type_ = self._type()
+        if const:
```
TimDiekmann, line 108: `const` could be `False`
```suggestion
        if const is not None:
```

TimDiekmann: or even better check if `const` is in `model_extra`

libs/@local/hash-graph-types/graph_types/data_type.py
TimDiekmann: This is probably worth moving into a separated PR.

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: or `[]` or an empty string or anything else. Yea major oversight. Thanks for the flag.

indietyp: > or even better check if `const` is in `model_extra`

I'll do that, I did that before but mypy yelled at me because `model_extra` can be `None`

libs/@local/hash-graph-types/graph_types/_const.py
indietyp: Every argument before `/` is required to be positional

libs/@local/hash-graph-types/graph_types/_const.py
indietyp: Python has no native support for sentinel values; this typedef is not what the end user will see, only the overloaded one. The `| object` is to allow for the sentinel value `_Undefined`. Otherwise we would need to add a default value to `None` or similar and that could lead to clashes and false positives.

libs/@local/hash-graph-types/graph_types/data_type.py
TimDiekmann:
```python
if "const" in (self.model_extra or {})
```
😂

libs/@local/hash-graph-types/graph_types/_const.py
indietyp: The previous approach has nothing wrong with it (although the validate function lacked a `return`) meaning no value was returned after validation.

indietyp: I took the issue advice and then transformed it into a function to make it more composable in the future (and make it clear what's happening)

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: I have that exact line in my working copy right now xD

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: This works for me (both did) but I think the issue is that the temporal implementation may use threads and asyncio primitives are not thread safe™

TimDiekmann: But I think this could be the case. Even though it's likely not possible to do the things in parallel, it may be done concurrently.

indietyp: I mean, this is why I tried to use the synchronization primitives for `Event`, but getting this right in Python is a lot harder than say... Rust.

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: fixed in [4523167](https://github.com/hashintel/hash/pull/2776/commits/45231679687f7011718077219d96d8dd473f13f7)

libs/@local/hash-graph-types/graph_types/_schema.py
indietyp: fixed in [8c97d8b](https://github.com/hashintel/hash/pull/2776/commits/8c97d8b9711344e7b703a28c2083e6ceeb4c66ad)

libs/@local/hash-graph-types/graph_types/_const.py
indietyp: fixed in [8c97d8b](https://github.com/hashintel/hash/pull/2776/commits/8c97d8b9711344e7b703a28c2083e6ceeb4c66ad)

libs/@local/hash-graph-types/graph_types/_const.py
TimDiekmann: We use `*` for this in other locations, is it the same as `*, const_: T`?

libs/@local/hash-graph-types/graph_types/data_type.py
indietyp: > it may be done concurrently.

To guard against that exact thing no await point is before checking, so that shouldn't happen in a non-threaded environment. Anyway. More investigation is needed 🔍

libs/@local/hash-graph-types/graph_types/data_type.py
```diff
@@ -104,7 +104,8 @@ async def create_data_type(
 
         type_ = self._type()
         if "const" in (self.model_extra or {}):
-            const = self.model_extra["const"]
+            # `const` can only be in `model_extra`, therefore it is safe to index!
+            const = self.model_extra["const"]  # type: ignore[index]
```
TimDiekmann, line 108: If we would just have `if let Some(...)` 🙃

libs/@local/hash-graph-types/graph_types/_const.py
indietyp: Nope,

Arguments before `/` are required positional, arguments after `*` are required keyword. Arguments in between them are positional-or-keyword.

So it's `(positional-only *)? positional-or-keyword (* keyword-only)?`

TimDiekmann (approved):

TimDiekmann (approved):

PR #2780, opened 2023-07-20
H-102, H-122, H-150: Show older type versions, and type editor bug fixes

CiaranMn:

This PR mainly updates the type editor and the preview slideover to handle showing earlier versions of entity types. Previously, they would each only show the latest version of the type. 

It also fixes a few bugs:
- multiple versions of a type appearing in the sidebar when more versions were created
- don't lose the description when creating an entity type
- fix a slug
- make the 'convert to link type' button look less out of place

Fixes h-102, h-122, h-150

## 🔍 What does this change?

- See commit list

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This PR updates the entity types fetched as part of app context to fetch all versions of all types, rather than the latest version of all types. This will be too much data with any real usage – but so will fetching all types in any case. I've created H-176 to track this. It suggests that we could stop fetching all types and instead:

- have some initial clever list of types in the sidebar based on the user/usage/popular types
- for type selectors, have the same 'recommended' list and then do dynamic queries to the API if the user types some search string in the selector
- for the type editor page, fetch the requested type from the API rather than expecting it to be in context (I thought about doing that as part of this PR but it would have involved more API work so didn't yet)

## 🛡 What tests cover this?

- There is an entity type creation test but this mainly requires manual testing.

## ❓ How to test this?

1.  Create a type, check the description from the initial form is carried into the main form (it isn't in `main`)
2. Create another version of the type, check that works
3. Try opening the earlier version's URL

## 📹 Demo

Screenshot 2023-07-20 at 15 20 08


benwerner01 (approved):
Thanks Ciaran, changes look good to me and are working for me locally as expected 👌

PR #2786, opened 2023-07-20
H-141: Expose `labelProperty` to GraphQL

TimDiekmann:

The HASH Backend is aware of the label property. In order to make it available to the frontend, it has to be exposed to GraphQL.

## 🚫 Blocked by

- #2775 
- #2778 
- #2782 
- #2783 
- #2784 
- #2785 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graphql-shared/src/graphql/type-defs/ontology/entity-type.typedef.ts
+++ b/libs/@local/hash-graphql-shared/src/graphql/type-defs/ontology/entity-type.typedef.ts
@@ -3,6 +3,7 @@ import { gql } from "apollo-server-express";
 export const entityTypeTypedef = gql`
   scalar ConstructEntityTypeParams
   scalar EntityTypeWithMetadata
+  scalar BaseUrl
 
   extend type Query {
     """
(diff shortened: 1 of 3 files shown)


libs/@local/hash-graphql-shared/src/graphql/type-defs/ontology/entity-type.typedef.ts
```diff
@@ -37,6 +38,10 @@ export const entityTypeTypedef = gql`
       """
       ownedById: OwnedById
       entityType: ConstructEntityTypeParams!
+      """
+      The label which is used as the label property for the entity type.
```
CiaranMn, line 42:
```suggestion
      The property which should be used as the label for entities of this type.
```

libs/@local/hash-graphql-shared/src/graphql/type-defs/ontology/entity-type.typedef.ts
```diff
@@ -51,6 +56,10 @@ export const entityTypeTypedef = gql`
       New entity type schema contents to be used.
       """
       updatedEntityType: ConstructEntityTypeParams!
+      """
+      The label which is used as the label property for the entity type.
```
CiaranMn, line 60:
```suggestion
      The property which should be used as the label for entities of this type.
```

CiaranMn (approved):

PR #2787, opened 2023-07-21
H-181: Implement logic to run a workflow from HASH

TimDiekmann:

To invoke a temporal workflow we require the API to connect to the temporal server. This PR introduces the `TemporalClient` and uses that to call the `inferEntities` workflow (which is not implemented yet).

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/package.json
+++ b/apps/hash-api/package.json
@@ -44,6 +44,7 @@
     "@ory/kratos-client": "0.13.1",
     "@rudderstack/rudder-sdk-node": "2.0.3",
     "@snowplow/node-tracker": "3.3.1",
+    "@temporalio/client": "1.8.1",
     "agentkeepalive": "4.2.1",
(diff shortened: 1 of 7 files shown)


apps/hash-api/package.json
```diff
@@ -44,6 +44,7 @@
     "@ory/kratos-client": "0.13.1",
     "@rudderstack/rudder-sdk-node": "2.0.3",
     "@snowplow/node-tracker": "3.3.1",
+    "@temporalio/client": "1.8.1",
```
CiaranMn, line 47: This is non-blocking, but we should update `@temporalio/worker` to `1.8.1` in a future PR – the `@temporalio/*` packages are part of an SDK and seem to get the same version bumps. Aligning them is probably best to avoid type mismatches or weird behaviour

CiaranMn (approved):

PR #2788, opened 2023-07-21
H-170: Enable org creation

CiaranMn:

Allows users to create an org via the + menu in the top right.

[Design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&mode=design&t=5iQdRP5HESgWGeLs-0)

## 🔍 What does this change?

- Adds some more properties to orgs (description, location, website)
- Update the `createOrg` mutation to allow anyone to use it, and to automatically join the user to the org on creation
- Adds a `/new/org` page which allows users to create an org, and redirects to the profile page afterwards
- Stop the entity type search in the sidebar being autofocused, which was interferring with the org form focus
- Driveby: rename `shortName` to `shortname` in one object for consistency
- Driveby: delete unused `entity-types.ts` file

This PR:

The changes in this PR:

  - we will explain org creation in any public-facing docs

The changes in this PR:

## ⚠️ Known issues

The newly created org does not appear in the user's workspace switcher (top left) until a refresh. I will fix this in follow up, tracking in H-184.

## 🐾 Next steps

- Re-use the org form to allow editing of an org's data (location and description fields will be introduced here)
- Fix the workspace switcher issue
- Add org membership management
- Add org avatars

## ❓ How to test this?

1.  Try creating an org


apps/hash-frontend/src/shared/layout/layout-with-header/actions-dropdown.tsx
```diff
@@ -116,6 +116,9 @@ export const ActionsDropdownInner: FunctionComponent<{
               >
                 <ListItemText primary="Create Entity Type" />
               </MenuItem>,
+              <MenuItem key="org" href="/new/org" onClick={popupState.close}>
+                <ListItemText primary="Create Organization" />
+              </MenuItem>,
```
benwerner01, line 121: Nit: maybe we can change this to something like this:
```Typescript
{activeWorkspace
          ? [
              { href: "/new/entity", label: "Create Entity" },
              { href: "/new/types/entity-type", label: "Create Entity Type" },
              { href: "/new/org", label: "Create Organization" },
            ].map(({ href, label }) => (
              <MenuItem key={href} href={href} onClick={popupState.close}>
                <ListItemText primary={label} />
              </MenuItem>
            ))
          : null}
```

That way we don't have to remember to add a `key` property every time we add another action item, whilst reducing the code a bit.

apps/hash-frontend/src/pages/shared/org-form.tsx
```diff
@@ -0,0 +1,235 @@
...
+          label="Website"
+          hint="Provide a link to help others identify your org"
+          htmlFor="website"
+        />
+        <TextField
+          id="name"
+          placeholder="https://acme.com"
+          sx={{ width: 400 }}
+          {...register("website", { required: false })}
+        />
```
benwerner01, line 216: Maybe adding some basic validation such as `pattern="https?://.+"` could help ensure this URL to be in the expected format, not sure this is a big deal though.

benwerner01 (approved):
Thanks Ciaran, looks good to me and is working for me locally. Have two non-blocking thoughts whilst looking through the code.

apps/hash-frontend/src/shared/layout/layout-with-header/actions-dropdown.tsx
CiaranMn: Good idea – I've done this in https://github.com/hashintel/hash/pull/2788/commits/3d46ff72d911d94fad250188bf019958764afe29

apps/hash-frontend/src/pages/shared/org-form.tsx
CiaranMn: We should do this – I'll do it in a follow up when I add the company website to their profile page. Tracking in H-182

benwerner01 (approved):

benwerner01 (approved):

PR #2790, opened 2023-07-21
H-185: Implement Workflow stub which is called from the API

TimDiekmann:

#2787 calls the `inferEntities` workflow from the `hash-api` package. This implements the endpoint on the `aipy` task queue.

Currently, the function does always error but in order to report back the error the status library has to be mirrored to Python. It would be possible to invent a codegen utility for Python as well similar as we do for Typescript and Rust, but quickly implement a few types let us move forward quickly.

## 🔍 What does this change?

- Implement basic `Status` and `StatusCode`
- Define parameter and return values for `inferEntities`
- Define workflow

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-py/app/workflows.py
```diff
@@ -199,3 +201,46 @@ async def get_entity_type(
...
+
+class InferEntitiesWorkflowResult(BaseModel, extra=Extra.forbid):
+    """Result of entity inference workflow."""
+
+    entities: list[ProposedEntity]
+
+
+@workflow.defn(name="inferEntities")
+class InferEntitiesWorkflow:
+    """A workflow that reads an entity type from the provided URL."""
```
CiaranMn, line 229: I'm not sure this is quite right... maybe something like this?
```suggestion
    """A workflow that infer entities of the specified type(s) from the provided text input"""
```

TimDiekmann: Python has this incredibly stupid guideline that the first sentence must be less than 88 characters and end with a period.
What about this?
```suggestion
    """Infers entities of the specified type(s) from the provided text input."""
```

CiaranMn (approved):

PR #2797, opened 2023-07-25
H-189: Implement activities and workflows to read Linear entities

TimDiekmann:

To read data from linear we're going to use their node SDK.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-integration-worker/src/activities.ts
+++ b/apps/hash-integration-worker/src/activities.ts
@@ -26,72 +26,72 @@ const readNodes = async <T>(connection: Connection<T>): Promise<T[]> => {
   return nodes;
 };
 
-export const createLinearIntegrationActivities = (createInfo: {
+export const createLinearIntegrationActivities = ({
+  linearClient,
+}: {
   linearClient: LinearClient;
 }) => ({
   async me(): Promise<User> {
-    return await createInfo.linearClient.viewer;
+    return await linearClient.viewer;
   },
 
   async organization(): Promise<Organization> {
-    return await createInfo.linearClient.organization;
+    return await linearClient.organization;
   },
 
   async teams(): Promise<Team[]> {
-    return await createInfo.linearClient.teams().then(readNodes);
+    return await linearClient.teams().then(readNodes);
   },
 
   async issues(filter?: { teamId?: string }): Promise<Issue[]> {
     const issuesQueryVariables: LinearDocument.IssuesQueryVariables = {
       filter: {},
     };
-    if (filter && filter.teamId) {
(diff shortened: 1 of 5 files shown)


CiaranMn (approved):

apps/hash-integration-worker/src/activities.ts
```diff
@@ -1,6 +1,120 @@
...
+    nodes.push(...connection.nodes);
+  }
+  return nodes;
+};
+
+export const createLinearIntegrationActivities = (createInfo: {
+  linearClient: LinearClient;
+}) => ({
+  async me(): Promise<User> {
+    return await createInfo.linearClient.viewer;
```
CiaranMn, line 33: I would make the return of all these functions either just `return x` or `return await x`.

The main difference is that `return await` means that this function will show in the stack trace, whereas if you do not await the return here it won't. `return await` also technically adds a tiny bit more work although it will make no practical difference here

You can read a bit more here: https://eslint.org/docs/latest/rules/no-return-await

apps/hash-integration-worker/src/activities.ts
```diff
@@ -1,6 +1,120 @@
...
+  const nodes = connection.nodes;
+  while (connection.pageInfo.hasNextPage) {
+    // eslint-disable-next-line no-param-reassign
+    connection = await connection.fetchNext();
+    nodes.push(...connection.nodes);
+  }
+  return nodes;
+};
+
+export const createLinearIntegrationActivities = (createInfo: {
```
CiaranMn, line 29: I would deconstruct this to save repeating `createInfo.linearClient` below
```suggestion
export const createLinearIntegrationActivities = ({ linearClient }: {
```

apps/hash-integration-worker/src/activities.ts
```diff
@@ -1,6 +1,120 @@
...
+
+  async teams(): Promise<Team[]> {
+    return await createInfo.linearClient.teams().then(readNodes);
+  },
+
+  async issues(filter?: { teamId?: string }): Promise<Issue[]> {
+    const issuesQueryVariables: LinearDocument.IssuesQueryVariables = {
+      filter: {},
+    };
+    if (filter && filter.teamId) {
```
CiaranMn, line 48: [The optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
```suggestion
    if (filter?.teamId) {
```

CiaranMn (approved):
A few non-blocking comments on the code

CiaranMn (approved):

PR #2802, opened 2023-07-25
H-199: Add Linear connection ontology types

benwerner01:

This PR adds an `@linear` HASH user, with the `User`, `Organization` and `Issue` ontology types.

## 🔗 Related links

- ...

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Set the `LINEAR_CLIENT_ID` environment variable
2. Run the local developer environment, which should seed the types in the database


apps/hash-api/src/graph/linear-user.ts
```diff
@@ -0,0 +1,55 @@
+import { Logger } from "@local/hash-backend-utils/logger";
+
+import { createKratosIdentity } from "../auth/ory-kratos";
+import { ImpureGraphContext } from "./index";
+import { createUser, getUserByShortname } from "./knowledge/system-types/user";
+import { systemUserAccountId } from "./system-user";
+
+const linearUserShortname = "linear";
```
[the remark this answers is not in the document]
benwerner01: /ignore reason This username will always be `linear` so can be hardcoded

TimDiekmann (dismissed):

apps/hash-api/src/graph/linear-types.ts
```diff
@@ -0,0 +1,1436 @@
...
+      {
+        propertyType: identifierPropertyType,
+        required: true,
+      },
+      {
+        propertyType: numberPropertyType,
+        required: true,
+      },
+      {
+        propertyType: previousIdentifierPropertyType,
```
TimDiekmann, line 1166:
```suggestion
        propertyType: previousIdentifiersPropertyType,
```

benwerner01: This is intentional so that the property type is reusable on entity types whether in an array or not

TimDiekmann: Discussed offline, the property types should be singular

TimDiekmann (approved):

PR #2804, opened 2023-07-26
H-200: Make Graph API available in integration worker

TimDiekmann:

To allow creating/reading entities in/from the Graph, the API needs to be available in the integration worker.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-integration-worker/package.json
+++ b/apps/hash-integration-worker/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": "./src/main.ts",
     "./workflows": "./src/workflows.ts",
-    "./activities": "./src/activities.ts"
+    "./activities": "./src/activities.ts",
+    "./graph": "./src/graph.ts"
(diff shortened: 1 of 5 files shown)


apps/hash-integration-worker/package.json
```diff
@@ -7,7 +7,8 @@
   "exports": {
     ".": "./src/main.ts",
     "./workflows": "./src/workflows.ts",
-    "./activities": "./src/activities.ts"
+    "./activities": "./src/activities.ts",
+    "./graph": "./src/graph.ts"
   },
   "typesVersions": {
```
CiaranMn, line 13: You can remove these two objects entirely – they are to declare what exports are available from this package for other apps/packages to use, but this is not designed to be imported by other packages. Apologies for not spotting this sooner.

CiaranMn (approved):
One non-blocking comment

PR #2811, opened 2023-07-27
H-207: Allow creating and updating from Linear webhooks

TimDiekmann:

This adds functionality to update issues and connects both, creation and updating, to webhooks

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The data passed in webhooks differs from the data passed from the linear node sdk, for example the web hook data does not contain the issue identifier (e.g. `H-207`). 
- the web hook data is in plain JSON, so only primitive datatypes are used, such as ISO strings for dates instead of the `Date` type. 

I will address these issues in a follow-up PR.

### DIFF
--- a/apps/hash-api/src/integrations/linear/webhook.ts
+++ b/apps/hash-api/src/integrations/linear/webhook.ts
@@ -2,6 +2,10 @@ import crypto from "node:crypto";
 
 import { RequestHandler } from "express";
 
+import { logger } from "../../logger";
+import { createTemporalClient } from "../../temporal";
+import { genId, getRequiredEnv } from "../../util";
+
 type LinearWebhookPayloadBase = {
   action: string;
   createdAt: string; // ISO timestamp when the action took place.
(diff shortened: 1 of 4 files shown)


apps/hash-api/src/integrations/linear/webhook.ts
```diff
@@ -45,9 +49,32 @@ export const linearWebhook: RequestHandler<{}, "ok", string> = async (
...
+  ) {
+    const workflow = `${payload.action}${payload.type}`;
+    await temporalClient.workflow.execute(workflow, {
+      taskQueue: "integration",
+      args: [
+        {
+          // @todo Use correct account IDs
+          actorId: "00000000-0000-0000-0000-000000000000",
+          ownedById: "00000000-0000-0000-0000-000000000000",
+          issue: (payload as LinearWebhookPayloadChange).data,
```
CiaranMn, line 72: It might be a user
```suggestion
          payload: (payload as LinearWebhookPayloadChange).data,
```

TimDiekmann: Good point, thank you. Fixed it in [`87b9e19` (#2811)](https://github.com/hashintel/hash/pull/2811/commits/87b9e192eb1d6ffaff6d76c97faa8e101f626660)

CiaranMn (approved):

PR #2812, opened 2023-07-27
H-209: Link Linear integration with HASH workspaces

benwerner01:

This PR adds a `syncLinearIntegrationWithWorkspace` mutation which links a linear integration to HASH users and orgs that it will sync data to. It also updates the FE to call this mutation when the user has selected the relevant linear teams to sync with a workspace.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The table isn't populated with the previous state.

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Go to `/account/integrations` and click "Connect to Linear"
2. Complete the OAuth flow, and select some teams to add to workspaces in the table dropdown
3. Click "Save and continue"
4. Once redirected, the `syncLinearDataWith` links should have been created. You can check the datastore to verify this

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
+++ b/apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
@@ -1,0 +1,39 @@
+import { Entity } from "@local/hash-subgraph";
+
+import {
+  getLinearIntegrationById,
+  syncLinearIntegrationWithWorkspace,
+} from "../../../../graph/knowledge/system-types/linear-integration-entity";
+import {
+  MutationSyncLinearIntegrationWithWorkspacesArgs,
+  ResolverFn,
+} from "../../../api-types.gen";
+import { LoggedInGraphQLContext } from "../../../context";
+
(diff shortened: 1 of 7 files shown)


apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
```diff
@@ -0,0 +1,41 @@
...
+  MutationSyncLinearIntegrationWithWorkspacesArgs
+> = async (
+  _,
+  { linearIntegrationEntityId, syncWithWorkspaces },
+  { dataSources, user },
+) => {
+  const linearIntegration = await getLinearIntegrationById(dataSources, {
+    entityId: linearIntegrationEntityId,
+  });
+
```
CiaranMn, line 26: Can we add a check that this integration is owned by the requesting user?

CiaranMn: We can do this in follow up to get this in sooner

CiaranMn (approved):

PR #2816, opened 2023-07-28
H-193: Trigger workflow when synchronizing Linear with workspace

TimDiekmann:

When synchronizing a linear account with the workspace we want to trigger the temporal workflow.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
+++ b/apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
@@ -1,9 +1,11 @@
-import { Entity } from "@local/hash-subgraph";
+import { Entity, extractOwnedByIdFromEntityId } from "@local/hash-subgraph";
 
 import {
   getLinearIntegrationById,
-  syncLinearIntegrationWithWorkspace,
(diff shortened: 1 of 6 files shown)


apps/hash-api/src/graphql/resolvers/integrations/linear/sync-workspaces-with-teams.ts
```diff
@@ -18,24 +20,55 @@ export const syncLinearIntegrationWithWorkspacesMutation: ResolverFn<
...
+    syncWithWorkspaces.map(async ({ workspaceEntityId, linearTeamIds }) => {
+      await linkIntegrationToWorkspace(dataSources, {
         linearIntegrationEntityId,
         workspaceEntityId,
         linearTeamIds,
         actorId: user.accountId,
-      }),
-    ),
+      });
+      return linearClient.syncWorkspace({
```
CiaranMn, line 65: I think we shouldn't return this, because it could take a long time and the frontend waits for this return.

We'll need some way of telling the user the status of their sync, which we can add later
```suggestion
      linearClient.syncWorkspace({
```

CiaranMn (approved):

PR #2819, opened 2023-07-28
H-216: Make sure the same format is used for dates in linear integration

TimDiekmann:

The Node SDK of linear returns a Date object while the webhooks passes an ISO string. Depending where the data comes from the correct conversion needs to be done.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-integration-worker/src/mappings.ts
+++ b/apps/hash-integration-worker/src/mappings.ts
@@ -22,6 +22,13 @@ export type PartialEntity = {
   entityTypeId: VersionedUrl;
 };
 
+const toIsoString = (date: string | Date): string => {
+  if (typeof date === "string") {
+    return date;
+  }
+  return date.toISOString();
+};
+
 export const userToEntity = (user: User): PartialEntity => {
   return {
     entityTypeId: linearTypes.entityType.user.entityTypeId,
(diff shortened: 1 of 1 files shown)


apps/hash-integration-worker/src/mappings.ts
```diff
@@ -22,6 +22,13 @@ export type PartialEntity = {
   entityTypeId: VersionedUrl;
 };
 
+const toIsoString = (date: string | Date): string => {
+  if (typeof date === "string") {
+    return date;
+  }
+  return date.toISOString();
+};
```
CiaranMn, line 30: You could save the tenaries below by updating it to this
```suggestion
const toIsoStringIfDefined = (date?: string | Date): string => {
  if (!date || typeof date === "string") {
    return date;
  }
  return date.toISOString();
};
```

CiaranMn (approved):
One non-blocking comment on saving code

apps/hash-integration-worker/src/mappings.ts
TimDiekmann: We agreed offline that we don't want to do this as it's weird to pass in a maybe-undefined value just to get another undefined back out.

PR #2824, opened 2023-07-28
H-192: Return an optional error message from workflows

TimDiekmann:

We want some hints if a workspace did succeeded or errored.

## 🚫 Blocked by

- #2821 

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann:
We don't use `.execute` for workflows and returning the status from the workflows is probably a bad design. Instead, we should capture the ids of the workflows and track them independently.

PR #2826, opened 2023-07-28
H-221: Add Linear integrations page, and improve linking Linear teams to workspaces

benwerner01:

This PR adds the `/account/integrations/linear` page, where existing linear integrations can be managed (workspaces added and removed from linear teams). It also makes other FE improvements, and fixes the `syncLinearIntegrationWithWorkspaces` mutation to allow for removing synced workspaces from an existing integration.

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

Link a linear integration in the app, which will redirect you to `/account/integrations/linear` page on completion. Or directly access the page if you already have a linked linear integration.

## 📹 Demo

image


apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
```diff
@@ -1,100 +1,153 @@
...
-    extractBaseUrl(types.propertyType.linearOrgId.propertyTypeId)
-  ] as string | undefined;
+  const handleSaveAndContinue = useCallback(async () => {
+    if (linearIntegration && linearOrganization) {
+      /** @todo: add proper error handling */
+
+      await syncLinearIntegrationWithWorkspaces({
+        variables: {
+          linearIntegrationEntityId:
+            linearIntegration.entity.metadata.recordId.entityId,
```
CiaranMn, line 106: I think these must be equivalent since you find one from the other
```suggestion
          linearIntegrationEntityId,
```

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
```diff
@@ -1,100 +1,153 @@
...
 
   const linearIntegrationEntityId = useMemo(() => {
     return router.query.linearIntegrationEntityId as EntityId;
   }, [router]);
 
-  const [linearIntegrationEntities, setLinearIntegrationEntities] =
-    useState<Entity[]>();
+  const { linearIntegrations } = useLinearIntegrations();
+
+  const linearIntegration = useMemo(() => {
```
CiaranMn, line 59: I think you could move this inside the fetching effect since you don't need it apart from to get the organization (the save callback can just use the `linearIntegrationEntityId` directly)

apps/hash-frontend/src/pages/account/integrations/select-linear-teams-table.tsx
```diff
@@ -0,0 +1,332 @@
...
+                      .filter(
+                        (entityId) =>
+                          teams.length ===
+                          teams.filter(({ workspaceEntityIds }) =>
+                            workspaceEntityIds.includes(entityId),
+                          ).length,
+                      )}
+                    possibleWorkspaces={possibleWorkspaces}
+                    setSelectedWorkspaceEntityIds={(entityIds) =>
+                      setLinearOrganizations((prev) => {
```
CiaranMn, line 181: probably not much of a difference but if we ended up syncing lots of orgs this would be a good candidate for moving to a callback somewhere higher up to avoid creating this function on every render as many times as there are Linear orgs synced

apps/hash-frontend/src/pages/account/integrations/use-linear-integrations.ts
```diff
@@ -0,0 +1,97 @@
...
+        data: {
+          operation: {
+            multiFilter: {
+              filters: [
+                /** @todo: figure out how to make this work */
+                // {
+                //   field: ["ownedById"],
+                //   operator: "EQUALS",
+                //   value: authenticatedUser.accountId,
+                // },
```
CiaranMn, line 37: I think this should work now

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
benwerner01: That's true, but `linearIntegrationEntityId` is populated form a query param so I felt it's safer to use the one from the `linearIntegration` as it's fetched from the server.

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
benwerner01: See [this response](https://github.com/hashintel/hash/pull/2826#discussion_r1279043708) on why I'd prefer using `linearIntegration` than `linearIntegrationEntityId` directly. If you disagree happy to change this.

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
CiaranMn: `linearIntegration` is the result of this:
```ts
linearIntegrations.find(
  ({ entity }) =>
    entity.metadata.recordId.entityId === linearIntegrationEntityId,
);
```
unless I'm missing something, it is logically impossible for them to be different

apps/hash-frontend/src/pages/account/integrations/use-linear-integrations.ts
benwerner01: Yes, thanks for reminding. Have made use of it in https://github.com/hashintel/hash/pull/2826/commits/636dff47faad60a18951019aba7e863afdc6f0b0

apps/hash-frontend/src/pages/account/integrations/select-linear-teams-table.tsx
benwerner01: That's a good point, I've moved them into callbacks in https://github.com/hashintel/hash/pull/2826/commits/680c4c03fce014db7a03fb02dc066248320e562c

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
benwerner01: If the `linearIntegrationEntityId` query param is something incorrect, then `linearIntegration` will remain undefined and the `useEffect` won't try to fetch the organizations with an incorrect ID. If I make use of `linearIntegrationEntityId` directly in the `useEffect` will try to fetch the organizations with the incorrect ID. 

But I suppose neither scenario is great anyway, we'll want to handle that scenario more gracefully perhaps by redirecting back to the `/account/integrations` page for example.

If I modify the `useMemo` hook to do that would you be happy with it?
```TypeScript
  const linearIntegration = useMemo(() => {
    if (linearIntegrations && linearIntegrationEntityId) {
      const integration = linearIntegrations.find(
        ({ entity }) =>
          entity.metadata.recordId.entityId === linearIntegrationEntityId,
      );
      
      if (!integration) {
        void router.push("/account/integrations");
      }

      return integration;
    }
  }, [linearIntegrations, linearIntegrationEntityId, router]);
```

CiaranMn: I'm suggesting that you could do the find for `linearIntegration` inside the same effect where you fetch the org, since it's only required there.
```ts
   useEffect(() => {
    void (async () => {
      if (!linearIntegrations || !linearIntegrationEntityId) {
        return;
      }

      const linearIntegration = linearIntegrations.find(
        ({ entity }) =>
          entity.metadata.recordId.entityId === linearIntegrationEntityId,
      );

      if (!linearIntegration) {
        void router.push("/account/integrations");
        return;
      }

      const linearOrgId = linearIntegration.entity.properties[
        extractBaseUrl(types.propertyType.linearOrgId.propertyTypeId)
      ] as string;

      const { data } = await getLinearOrganization({
        variables: { linearOrgId },
      });

      if (data) {
        const organization = data.getLinearOrganization;

        setLinearOrganization(
          mapLinearOrganizationToLinearOrganizationTeamsWithWorkspaces({
            linearIntegrations,
          })(organization),
        );
      } else {
        throw new Error("Could not get linear organization");
      }
    })();
  }, [
    linearIntegrations,
    linearIntegrationEntityId,
    getLinearOrganization,
    router,
  ]);
```

benwerner01: I see how that's cleaner now, thank for explaining. Have addressed in https://github.com/hashintel/hash/pull/2826/commits/0afd9f20d7568e28b60e92a2fa125ea628fec88a

CiaranMn: For the save callback, I think you'd need to change the `EntityId` you're importing to satisfy the type
```ts
import { EntityId } from "@local/hash-subgraph";

...

  const linearIntegrationEntityId = useMemo(() => {
    return router.query.linearIntegrationEntityId as EntityId | undefined; // add | undefined since it might be
  }, [router]);

...

  const handleSaveAndContinue = useCallback(async () => {
    if (linearOrganization && linearIntegrationEntityId) {
      /** @todo: add proper error handling */

      await syncLinearIntegrationWithWorkspaces({
        variables: {
          linearIntegrationEntityId,
          syncWithWorkspaces:
            mapLinearOrganizationToSyncWithWorkspacesInputVariable({
              linearOrganization,
              possibleWorkspaces,
            }),
        },
      });

      void router.push("/account/integrations/linear");
    }
  }, [
    syncLinearIntegrationWithWorkspaces,
    linearIntegrationEntityId,
    linearOrganization,
    possibleWorkspaces,
    router,
  ]);
```

apps/hash-frontend/src/pages/account/integrations/linear/new.page.tsx
CiaranMn: Resolved in https://github.com/hashintel/hash/commit/0afd9f20d7568e28b60e92a2fa125ea628fec88a

apps/hash-frontend/src/pages/account/integrations/use-linear-integrations.ts
CiaranMn: Was there any client-side filtering that's now redundant?

CiaranMn (approved):

apps/hash-frontend/src/pages/account/integrations/use-linear-integrations.ts
benwerner01: There wasn't a client side filter in this instance, no. I think I removed it when moving the query to the hook, or may have left it out entirely.

PR #2828, opened 2023-07-28
H-225: Finalize two-way Linear issue synchronization

CiaranMn:

This PR makes the 'sync issue changes back to Linear' work.

It also updates the Linear type seeding script – this previously included duplicate types named Organization and User (they appeared as both entity types and link types). I've kept a single link between Orgs and Users, `hasMember`, as we don't need to create links in both directions. I've also removed a cyclical dependency between Users and Issues because our seeding script doesn't handle this – I will fix it properly next week.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Fill out more actions beyond 'update issue properties'
## ❓ How to test this?

1.  Link a Linear account
2. Find a Linear issue in the HASH entity editor (e.g. visit `/@linear/types/entity-type/issue` and pick an entity from the entities tab)
3. Edit its title and check that it updates in Linear

### DIFF
--- a/apps/hash-api/src/integrations/linear/sync-back.ts
+++ b/apps/hash-api/src/integrations/linear/sync-back.ts
@@ -1,25 +1,89 @@
+import { GraphApi } from "@local/hash-graph-client";
 import { linearTypes } from "@local/hash-isomorphic-utils/ontology-types";
-import { Entity } from "@local/hash-subgraph";
+import {
+  Entity,
+  entityIdFromOwnedByIdAndEntityUuid,
+  EntityUuid,
+  extractOwnedByIdFromEntityId,
+  OwnedById,
+  Uuid,
+} from "@local/hash-subgraph";
+import { extractBaseUrl } from "@local/hash-subgraph/type-system-patch";
+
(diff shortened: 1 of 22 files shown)


apps/hash-api/src/integrations/linear/sync-back.ts
```diff
@@ -70,20 +70,11 @@ export const processEntityChange = async (
 
   switch (entityTypeId) {
     case linearTypes.entityType.issue.entityTypeId: {
-      // eslint-disable-next-line no-console
-      console.log("Change to Linear Issue detected");
-
-      // @todo check this works
-      const result = await temporalClient.workflow.start("updateLinearIssue", {
+      await temporalClient.workflow.start("updateLinearIssue", {
```
TimDiekmann, line 73: Do we want to wait for completion instead? Then we could check if the sync was successful.

CiaranMn: I think we should handle that via visibility into Temporal job failures, not returning it to the API – there's no user request to pass the error back to so the most we could do is log an error somewhere

TimDiekmann (approved):

PR #2834, opened 2023-07-31
H-230: Share Temporal types with API

CiaranMn:

We want to share the types of Temporal workflows with the API, so that the arguments (and where relevant) return types have type safety.

One of way of doing this is to import the workflow functions themselves when starting a workflow, but this is undesirable because:
1. It requires evaluating all the code in the workflows and their supporting activities, just to get their types – potentially leading to errors
2. The design of the repo is that shared code should go in `libs/` rather than apps importing from one another

This PR introduces shared types in `@local/hash-backend-utils`

## 🔍 What does this change?

- Adds shared types
- Renames some workflow functions to make clear whether they're updating a HASH resource or a Linear resource
- Ensures that the webhook uses a correct workflow name, and simplifies the webhook payload type to make this easier
- Ensures that the functions exported from the workflows file match the expected ones from the types 

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  Rebuild `external-services`
2. Check the integration still works

### DIFF
--- a/libs/@local/hash-backend-utils/src/temporal-workflow-types.ts
+++ b/libs/@local/hash-backend-utils/src/temporal-workflow-types.ts
@@ -1,6 +1,11 @@
 import { VersionedUrl } from "@blockprotocol/type-system";
 import { Issue, Team, User } from "@linear/sdk";
-import { Entity, EntityPropertiesObject } from "@local/hash-subgraph";
+import {
+  AccountId,
+  Entity,
+  EntityPropertiesObject,
+  OwnedById,
+} from "@local/hash-subgraph";
 
 export type PartialEntity = {
   properties: Partial<Entity["properties"]>;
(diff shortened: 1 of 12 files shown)


libs/@local/hash-backend-utils/src/temporal-workflow-types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type PartialEntity = {
+  properties: Partial<Entity["properties"]>;
+  entityTypeId: VersionedUrl;
+};
+
+export type CreateHashIssueWorkflow = (params: {
+  payload: Issue;
+  ownedById: string;
+  actorId: string;
```
benwerner01, line 13: I'm wondering why we're not using `OwnedById` and `ActorId` to type these? Is this to simplify calling them in temporal where they are typed as strings?

benwerner01 (approved):
Looks good to me, and is working for me locally - thanks Ciaran!

libs/@local/hash-backend-utils/src/temporal-workflow-types.ts
CiaranMn: No reason – done in https://github.com/hashintel/hash/pull/2834/commits/53c90c0af5ab9007c65cf8670f571585de51c695

PR #2837, opened 2023-08-01
H-227, H-229: Update Vault config and path construction

CiaranMn:

This PR does the following:
- Updates the path construction for Vault secrets to avoid users of the Vault client attempting to use engines / paths we don't want used from the API (fixes H-229)
- Adds the Vault environment variables to the API Terraform config, ~and removes `ROOT` from the Vault token environment variable name, since we want to use a more specific token with a limited policy (being introduced to the Vault config in https://github.com/hashintel/internal-api/pull/133) – fixes H-227 once that linked PR is merged and we have created a token for that role for use in the API~ – we should use the `iam` auth method instead, see 'blocked by'
- Driveby: Removes a redundant environment variable for the Linear OAuth callback URL (it will never differ from `${apiOrigin}/oauth/linear/callback`)
- Driveby: associates the user secret with the user always, and the integration with the owning workspace (rather than the user secret potentially being owned by an org as it was previously)

## 🚫 Blocked by

- [ ] https://github.com/hashintel/internal-api/pull/133 must be applied to create the Vault role for the API, ~which can then have a token created for it. Once the token is available we can merge this so that the API is deployed with it~
- [ ] implement the `iam` auth method with Vault (tokens have a time limit) and allow the API's execution role to use it – [docs](https://developer.hashicorp.com/vault/docs/auth/aws)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  Check that creating a Linear integration still works


CiaranMn:
Paths were updated in #4141 and environment variables added in #4174 

There is still an outstanding task to use IAM profiles to connect to Vault (H-227).

PR #2838, opened 2023-08-01
H-151: Update OpenAPI specification

indietyp:

This PR updates the OpenAPI specification, by basically translating the `0.2` to `0.3` types for querying data.

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
This PR also changes from `$defs` to `definitions`, as it seems to be special-cased.

Otherwise you get errors like:

```
[[ts] ../../../apps/hash-graph/openapi/openapi.json] [main] WARN  o.o.codegen.utils.ModelUtils - Failed to get the schema name: #/$defs/PropertyValues
[[ts] ../../../apps/hash-graph/openapi/openapi.json] [main] WARN  o.o.codegen.utils.ModelUtils - #/$defs/PropertyValues is not defined
[[ts] ../../../apps/hash-graph/openapi/openapi.json] [main] WARN  o.o.codegen.utils.ModelUtils - Failed to get the schema name: #/$defs/PropertyValues
```

during generation.

indietyp:
Summary of the changes as Diff to the TS client:

screenshot 2023-08-01T18-49-55

screenshot 2023-08-01T18-50-31

screenshot 2023-08-01T18-50-57

screenshot 2023-08-01T18-51-14

screenshot 2023-08-01T18-51-27

screenshot 2023-08-01T18-51-46

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
...
+      "type": "object",
+      "properties": {
+        "$ref": {
+          "$ref": "#/definitions/VersionedUrl"
+        }
+      },
+      "required": ["$ref"],
+      "additionalProperties": false
+    },
+    "PropertyTypeReference": {
```
TimDiekmann, line 112: This is not a `PropertyTypeReference`. We could call it `PropertyObjectReference` but I would prefer to define it inline.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
...
+      "type": "string"
+    },
+    "allOf": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/EntityTypeReference"
+      }
+    },
     "examples": {
-      "$comment": "Example Entity instances",
```
TimDiekmann, line 25: Why removing the comments?

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
...
       }
     },
-    "links": { "$ref": "#/$defs/linkTypeObject" }
+    "links": {
+      "type": "object",
+      "propertyNames": {
+        "$ref": "#/definitions/VersionedUrl"
+      },
+      "patternProperties": {
+        "^.*$": {
```
TimDiekmann, line 56: This is different from the upstream spec.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
...
+        }
+      }
+    }
   },
   "additionalProperties": false,
-  "required": ["$schema", "kind", "$id", "type", "title", "properties"],
-  "$defs": {
-    "propertyTypeObject": {
+  "required": ["$schema", "kind", "type", "$id", "title", "properties"],
+  "definitions": {
```
TimDiekmann, line 87: Can we please keep using `$defs`?

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
...
+      "type": "object",
+      "properties": {
+        "$ref": {
+          "$ref": "#/definitions/BaseUrl"
+        }
+      },
+      "required": ["$ref"],
+      "additionalProperties": false
+    },
+    "OneOfEntityTypeReference": {
```
TimDiekmann, line 122: Splitting this out does not help readability. I'd prefer to define it inline.

TimDiekmann (changes requested):
Very happy to see these schemas updated 🎉 

Partial review, for now, will review the rest after the review is addressed. Is there a specific reason why you changed some previous approaches?

apps/hash-graph/lib/graph/src/api/rest/json_schemas/data_type.json
```diff
@@ -10,13 +10,29 @@
       ]
     },
     "kind": {
+      "type": "string",
```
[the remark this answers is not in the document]
indietyp: Yes, otherwise the generator assumes `object`.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -1,6 +1,6 @@
 {
-  "$id": "https://blockprotocol.org/type-system/0.2/schema/meta/entity-type",
-  "description": "Specifies the structure of an Entity Type",
+  "title": "Entity Type",
```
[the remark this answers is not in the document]
indietyp: `$schema` wasn't there to begin with as it is incompatible with the OpenAPI spec. `$id` was removed (can add again), but needs to not be a URL and just an identifier, otherwise the generator generates garbage.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: I can do that, defining it outside has the benefit that generators will be able to generate distinct types (with distinct names) for EntityType references, etc.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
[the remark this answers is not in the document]
TimDiekmann: What do you mean by "generates garbage"? Didn't it work before?

apps/hash-graph/lib/graph/src/api/rest/json_schemas/data_type.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://blockprotocol.org/type-system/0.2/schema/meta/data-type",
+  "title": "Data Type",
```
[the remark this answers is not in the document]
indietyp: See other comment.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
TimDiekmann: We definitely need to rename it as in other parts of the code `PropertyTypeReference` is a `VersionedUrl`. I don't feel strongly about making it inline

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: I'd like that, but `$defs` has the problem that types that are not referenced explicitly get omitted and then result in `object`. `definitions` is due to OpenAPI special cased and will always be included.
I can create a repro to show it of if needed.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: Sure, essentially just copied it from the original.

TimDiekmann:
OpenAPI spec 3.1 [claims to be conform to JSON Schema](https://www.openapis.org/blog/2021/02/18/openapi-specification-3-1-released), can't we just use that?

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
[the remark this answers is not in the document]
indietyp: It works here (as openapi-generator ignores it on the top level), but if `$id` is in any of the sub-definitions they will error out and generate as `object`. To stay consistent I removed it from the top too (although I can add it again)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: and because of that I decided to remove them (as there are no tools that recognize `x-$comment`), when we switch to OpenAPI 3.1, we could simply embed the schema.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/data_type.json
```diff
@@ -10,13 +10,29 @@
...
     },
     "kind": {
+      "type": "string",
       "enum": ["dataType"]
     },
-    "$id": { "type": "string", "format": "uri" },
-    "title": { "type": "string" },
-    "description": { "type": "string" },
-    "type": { "type": "string" }
+    "$id": {
```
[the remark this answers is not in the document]
indietyp: I first had a `$ref` and `oneOf` but that was invalid, and with an outer `oneOf` the generator generated the wrong stuff.

I could either add the VersionedUrl ref and keep it as a string or keep it an enum (or format: uri)

indietyp:
Yes it is, but the tooling around it is.... lacking at best. openapi-generator just started like a month ago to support 3.1 and utoipa doesn't support 3.1 AFAIR.

indietyp:
otherwise I would've suggested it a long time ago :/

indietyp:
see https://github.com/juhaku/utoipa/issues/531#issue-1631861455 (also look at the issue author^^)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: My apologies, I was also thinking if simply removing them (as you can see the generator doesn't support them.)
`patternProperties` and `propertyNames` are not recognized by OpenAPI 3.0, there is `x-patternProperties` as an unofficial substitute, but I wasn't able to ascertain if OpenAPI-generator supports it (with our current configuration certainly not)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
```diff
@@ -10,63 +10,150 @@
       ]
     },
     "kind": {
+      "type": "string",
```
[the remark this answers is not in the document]
indietyp: I am unsure, but `openapi-codegenerator` does, if not provided it will default to `object`.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: fixed in [474fe88](https://github.com/hashintel/hash/pull/2838/commits/474fe8823c5f74cba4856564c8c7474172e7bd02)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: fixed in [9ccbd86](https://github.com/hashintel/hash/pull/2838/commits/9ccbd8673e92eede8b25ed19917a6f23ec116aa0)

indietyp:
> Is there a specific reason why you changed some previous approaches?

I essentially copied the example from the BP repository and applied the transformations outlined in https://github.com/openapi-contrib/json-schema-to-openapi-schema. As OpenAPI 3.0 is essentially draft-04 with some caveats.

Most of the other changes are to introduce consistency and also aid use in `datamodel-code-generator`.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/data_type.json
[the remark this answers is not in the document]
indietyp: fixed in [8ebe3e8](https://github.com/hashintel/hash/pull/2838/commits/8ebe3e8d2bc145ac90c7025f12726689e5e7dcf2)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: No, see https://github.com/hashintel/hash/pull/2838#issuecomment-1660700773; this error only happens if we use `$defs` not `definitions`. `PropertyValues` is not directly referenced in the "parent" schema, only in `$defs`, and therefore not included (because `definitions` is special-cased), but if it is in `definitions`, it is included regardless.

indietyp: OpenAPI 3.0 is based on `draft-04`, which doesn't yet have `$defs` and instead uses `definitions` (https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.5.7)

apps/hash-graph/lib/graph/src/api/rest/json_schemas/entity_type.json
indietyp: So include `patternProperties` (or in this case, `x-patternProperties`)?

TimDiekmann: If `x-patternProperties` is part of the spec let's use that but we should keep the information I think.

indietyp: fixed in [bf1dcfc](https://github.com/hashintel/hash/pull/2838/commits/bf1dcfcdb3ff3a99dad62943e7deb8df9fdf8adf)

(`patternProperties` and `propertyNames` are not supported by `OpenAPI 3.0`, but there are tools that understand the `x-` prefix versions, `x-` is the namespace for any property that isn't strictly OpenAPI 3.0 compliant, also known as extensions)

TimDiekmann:
> I am unable to replicate the issue in `CI / build-hash-ai-worker-ts` locally :/

I haven't tried reproducing it locally, but the enumeration field on `$id` causes problems. I'd say we revert that change. Also, the `additionalProperties: true` removal _will_ cause problems in the future.

indietyp:
I am not a huge fan of removing accuracy and information, as this version is more correct, but I get that with https://github.com/blockprotocol/blockprotocol/pull/355 on the horizon we will need to make that change sooner or later.

apps/hash-graph/lib/graph/src/api/rest/json_schemas/data_type.json
```diff
@@ -10,13 +10,31 @@
...
+    "description": {
+      "type": "string"
+    },
+    "type": {
+      "type": "string"
+    }
   },
   "required": ["$schema", "kind", "$id", "title", "type"],
-  "additionalProperties": true
+  "additionalProperties": false,
```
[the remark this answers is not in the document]
indietyp: Whoops, sorry! I muscle memory must've gotten the better of me.

TimDiekmann (approved):

PR #2839, opened 2023-08-01
H-249, H-256, H-272: Fix entity type entities tab table lazy loading, add basic loading indicator

CiaranMn:

The entity type editor has an 'entities' tab which shows a table of entities assigned to that type.

This table's height was unconstrained, meaning that all elements were rendered at once, causing a crash in Firefox and poor performance in other browsers.

This PR introduces a cap on the table height which allows Glide's `DataEditor` component to lazily render rows when they are in view, as it is supposed to.

I've also taken this opportunity to introduce a basic loading indicator in place of the entities count, while the data is loading. It was previously misleadingly showing `0` entities when the loading was taking place. See demo below.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Later on, add better empty state on the entities tab while data is loading, e.g. a loading skeleton (H-256). 

Also ideally we would:
1. Get a count of entities before the entities themselves 
2. Paginate the data

## ❓ How to test this?

1.  You can make a bunch of data load in the table by editing line 87 in `use-entities-table.tsx`
```ts
      new Array(500_000).fill(entities[0]).map((entity) => {
```

## 📹 Demo

Demo shows loading state and 2 million entities in table

https://github.com/hashintel/hash/assets/37743469/9845c182-56a0-41a9-b108-275e09cdbc5e

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entities-tab.tsx
+++ b/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entities-tab.tsx
@@ -153,8 +153,8 @@ export const EntitiesTab: FunctionComponent = () => {
           <FontAwesomeIcon icon={faCircleQuestion} sx={{ fontSize: 14 }} />
         }
       >
-        <Paper sx={{ overflow: "hidden" }}>
-          {isEmpty ? (
+        {isEmpty ? (
+          <Paper sx={{ overflow: "hidden" }}>
             <SectionEmptyState
               title="There are no entities of this type visible to you"
               titleIcon={
@@ -162,7 +162,13 @@ export const EntitiesTab: FunctionComponent = () => {
               }
               description="Assigning this type to an entity will result in it being shown here"
             />
-          ) : (
+          </Paper>
+        ) : (
+          <Box
+            sx={{
+              height: "50vh",
+            }}
+          >
             <Grid
               showSearch={showSearch}
               onSearchClose={() => setShowSearch(false)}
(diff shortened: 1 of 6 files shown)


TimDiekmann:
> Demo shows loading state and 2 million entities in table

After loading, the demo shows `Entities 1` instead of `Entities 2 000 000`

nonparibus:
> > Demo shows loading state and 2 million entities in table
> 
> After loading, the demo shows `Entities 1` instead of `Entities 2 000 000`

I think this is just because it's showing the same entity over and over (artificially), but I might be wrong.

CiaranMn:
> > > Demo shows loading state and 2 million entities in table
> > 
> > 
> > After loading, the demo shows `Entities 1` instead of `Entities 2 000 000`
> 
> I think this is just because it's showing the same entity over and over (artificially), but I might be wrong.

@TimDiekmann @nonparibus that's right, I used a type which only had 1 entity but duplicated it artificially.

Here's a demo with the updated styling following a suggestion from @nonparibus that shows the accurate entity count

https://github.com/hashintel/hash/assets/37743469/4f4ccd3f-90ce-4db1-9b9d-c23bb1d7ad29

benwerner01 (approved):
Thanks Ciaran, code looks good to me and is working locally as expected 👍

CiaranMn:
> What happens to the loading indicator if the number has multiple digits, like 100? Does it automatically transition the width?

@indietyp depending on what you mean by transition the width, the answer is:

1. Does the chip become the correct width when the number loads in? **Yes, see video below**
2. Is the loading indicator the same width as it will be once the number is loaded? **No, because we don't yet know the count – but we should add a count method so that we can load the count quicker than the table**
3. Is there an animation which transitions the width of the chip? **No, we could add it but it seems fairly low priority**

https://github.com/hashintel/hash/assets/37743469/65bcb95a-152b-48d6-b9fd-6dc7c98b2eef

CiaranMn:
Thanks @indietyp – let me know if it looks bad moving from the loading indicator to many digits and we can add an animation.

PR #2841, opened 2023-08-02
H-291: Add Python package for `hash-graph-client`

indietyp:

Create the package skeleton for a Python equivalent of `hash-graph-client` for a semi-automatically generated Python client from the OpenAPI specification.

The first attempt (not in history) was to merge both the typescript and python language in the same directory, while possible it was **very** confusing and would be hard to maintain. To combat this, the typescript client has been moved into `typescript/`, while the `python` client lives in the `python/` subdirectory.

The `typescript` package remains with the `@local/hash-graph-client` package name, while the python package uses `@local/hash-graph-client-python`, but(!) the official name (in `pyproject.toml`) is `hash-graph-client`.

As discussed with @nonparibus offline, the package is to be licensed w/ AGPL.

> The move and other possibilities of package arrangements have been discussed offline with @TimDiekmann.

## 🔍 What does this change?

* add `hash-graph-client` python package.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

* generate client, and move client over from internal repo


libs/@local/hash-graph-client/python/turbo.json
```diff
@@ -0,0 +1,41 @@
+{
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build", "codegen"],
+      "inputs": ["./**/*.py", "pyproject.toml", "poetry.lock", "LICENSE*"],
+      "outputs": ["dist/**"]
+    },
```
TimDiekmann, line 8: Does this not depend on the OpenAPI?

libs/@local/hash-graph-client/typescript/openapitools.json
```diff
@@ -1,14 +1,14 @@
...
   "spaces": 2,
   "generator-cli": {
     "version": "6.6.0",
     "generators": {
       "ts": {
         "generateAliasAsModel": true,
         "generatorName": "typescript-axios",
-        "output": "#{cwd}/../hash-graph-client",
-        "inputSpec": "../../../apps/hash-graph/openapi/openapi.json",
+        "output": "#{cwd}/../typescript",
```
TimDiekmann, line 10: Isn't this the same as `"#{cwd}"`?

package.json
```diff
@@ -8,6 +8,7 @@
       "libs/*",
       "libs/@hashintel/*",
       "libs/@local/*",
+      "libs/@local/*/typescript",
```
TimDiekmann, line 11: We want the python package to be part of the workspace as well.
```suggestion
      "libs/@local/*/typescript",
      "libs/@local/*/python",
```

libs/@local/hash-graph-client/typescript/openapitools.json
indietyp: I thought the same but was anxious to change the previously configured value (which also points to `#{cwd}`). I'll change it an see what happens.

libs/@local/hash-graph-client/python/turbo.json
indietyp: Not yet, #2842 adds that dependency

libs/@local/hash-graph-client/typescript/openapitools.json
indietyp: fixed in [4367f05](https://github.com/hashintel/hash/pull/2841/commits/4367f052c76d88b83b2be7386cc6db1d5abe57e2)

TimDiekmann (dismissed):

nonparibus (approved):

libs/@local/hash-graph-client/typescript/turbo.json
TimDiekmann, line 1: The input was supposed to be changed here 😢

PR #2842, opened 2023-08-02
Add `hash-graph-client` functionality

indietyp:

Adds all the required code (and codegen commands) required.

## 👂 Open Questions

To be able to consume the client externally (which I need to do^^) `models.py` cannot be generated on-demand. It would need to be generated beforehand (just like `openapi.json`) and then optionally on a PR, we'd need to check if the file changed (the git repo is dirty).

## 🚫 Blocked by

- #2841 
- #2838 

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

* generate client, and move client over from internal repo


libs/@local/hash-graph-client/python/graph_client/models.py
TimDiekmann, line 1: Why these files are committed instead of being generated on the client as part of a `turbo` step?

indietyp: > To consume the client externally (which I need to do^^) models.py cannot be generated on-demand. It would need to be generated beforehand (just like openapi.json).

Unlike `yarn`, `poetry` (or any other python tool for that matter) executes post-install steps (or any steps), meaning that the client would be usable internally, but external users (e.g. me), would not be able to consume the package.

.github/workflows/openapi.yml
```diff
@@ -0,0 +1,23 @@
...
+        if: ${{ success() || failure() }}
+
+      - name: Setup python
+        run: yarn poetry:install
+
+      - name: Generate python client
+        run: yarn codegen --filter @local/hash-graph-client-python
+
+      - name: Ensure empty git diff
+        run: git --no-pager diff --exit-code --color libs/@local/hash-graph-client/python
```
TimDiekmann, line 23: We can enforce an empty diff everywhere. Running those steps should not change anything.
```suggestion
        run: git --no-pager diff --exit-code --color
```

libs/@local/hash-graph-client/python/graph_client/models.py
TimDiekmann: Sorry, missed that in the OP 🫣

As the least thing we should add this file to `.gitattributes` to mark it as generated then. It's not blocking this PR but we really need to solve this issue, though. Any suggestions are highly appreciated.

TimDiekmann: In any case, this still requires a proper `turbo` setup as otherwise the local development suffer from this issue. If, for example, the `openapi.json` is changed, `turbo` should automatically update this file as a result.

indietyp: One solution _could_ be to upload the package to a registry (like PyPi), then in the publishing process, we could generate the `models.py`. I was about to suggest [GitHub Packages](https://docs.github.com/en/packages/working-with-a-github-packages-registry), but turns out, they do not support Python just yet :/

TimDiekmann: This implies that we need proper versioning for every package. However, this would probably solve the dependency issue we currently encounter with Python, though.

indietyp: > In any case, this still requires a proper turbo setup as otherwise the local development suffer from this issue. If, for example, the openapi.json is changed, turbo should automatically update this file as a result.

That's already done, the file is in the `turbo.json`, `codegen` exists with the required inputs and outputs and runs before build.

TimDiekmann: We need a more general solution as we don't want a separate solution for every package, which has internal dependencies such as `hash-graph` depending on the status library.

indietyp: as a stop-gap, `models.py` has been marked as a generated file in [40cada6](https://github.com/hashintel/hash/pull/2842/commits/40cada6877a6bf7005fd88d02c9b024d49c1db48), I'll think and explore other alternatives that will potentially solve our current issues and post them in the internal Slack.

TimDiekmann: I will also give the issue more thought. Thanks for adding the commit! 🎉

TimDiekmann (approved):

TimDiekmann (approved):

PR #2847, opened 2023-08-02
Make PR template checkbox completion easier

CiaranMn:

Our PR template has some checkboxes, under 'Pre-merge checklist', that contributors are required to delete and check as appropriate.

Most of the time all three sub-sections are likely to require the 'no action required' checkbox only to be present and checked, which was previously fiddly to do.

This PR makes it easier by:
1. Checking all boxes by default, so you only need to delete the inappropriate ones, not additionally check the remaining ones
2. Moves the 'no action required' boxes to the top of each sub-section, so that it's easier to select the remainder and delete them

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  You could open the amended file in an editor and check that it seems easier to end up with only the 'no action required' boxes than it was before

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -34,15 +34,15 @@
 
 This PR:
 
-- [ ] modifies an **npm**-publishable library and **I have added a changeset file(s)**
+- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing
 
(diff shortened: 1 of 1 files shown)


TimDiekmann (approved):
I like this approach more, should be quick to remove unwanted parts.

indietyp (approved):
👍

benwerner01 (approved):

PR #2852, opened 2023-08-03
H-299: Generate TypeScript types from system entity types

CiaranMn:

We have a number of 'system' ontology types, e.g. `User`, `Org`, which are generated from code but are not available as TypeScript types. That means are we manually asserting the types of their properties, e.g. [here](https://github.com/hashintel/hash/blob/9849da725c6e93e128207b496274c2fae8da9bbf/apps/hash-frontend/src/lib/user-and-org.ts#L207). This is slower and less safe than having their TS types available.

This PR adds a script to generate TypeScript types from the system types. It requires the frontend and API to be running locally as it relies on fetching the JSON Schema from its URL.

In enabling this it also makes requesting the JSON of types from their URL better in a couple of ways:
1.  The switches the default from 'serve as HTML unless JSON is accepted' to 'serve as JSON unless HTML is requested', meaning that a simple request for a type's URL will return the JSON unless loaded as a browser page.
2. For property and data types, which do not have any HTML representation, it will always serve JSON, so that visiting a property or data type URL in the browser will also show the JSON.

## 🔍 What does this change?

As well as the changes described above:
- allows unauthenticated requests for property and entity types. These would previously crash because they relied on a user being present in order to pass an `actorId` if an external type was requested (to be able to create it if unavailable). Unauthenticated requests for external types not already in the db will fail, but unaunthenticated requests for existing types will succeed rather than crashing. 
- drive-by: moves the logic to acquire a Redis queue for watching database changes to write back to third-party integrations, because the previous approach wasn't reacquiring the queue on some hot reloads, causing the API to crash.
- updates `@blockprotocol/graph` to bring in necessary fixes from https://github.com/blockprotocol/blockprotocol/pull/1311 (to avoid type name clashes between `Text` and `Text` entity and data types) and https://github.com/blockprotocol/blockprotocol/pull/1310 (to be able to call `codegen` from our non-ESM files)

This PR:

    - blocks are technically modified but not in any way that requires republishing them

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

1. This currently requires manually running the generation script when system types are changed to update their type. We could alternatively make generating the types part of the `yarn dev` command.
2. We are using a canary version of `@blockprotocol/graph` because integration tests are failing on https://github.com/blockprotocol/blockprotocol/pull/1310, probably as a result of upgrading the GitHub action runner to Node 18. I will return to this.
3. By design of the codegen utility, types which are shared across multiple of the entity types to be generated are stored in a `shared.ts` file, including entity types which are referenced by other entity types. This means that many entity types end up in `shared.ts`, given that they reference each other, which is not the most intuitive for finding them, especially where autocomplete would offer many options (e.g. `User`).. The alternative would be to generate each file individually and accept that there would be many `UserProperties` in the repo. Because I expect to mainly use `XProperties` types, autocomplete should be okay, but we may want to revisit this.

## 🐾 Next steps

- I will be using the generated types to improve how we deal with `User` and `Org` entity types in the frontend, e.g.
    - remove the need to manually declare what type each property is
    - create a utility to be able to acccess system type properties by simpler keys than their URLs (better DX as url keys are more fiddly to write/autocomplete and more verbose)

## ❓ How to test this?

1.  Run `yarn generate-system-types`. The output is deterministic and should not cause a diff.

## 📹 Demo

See the generated types in the diff.


libs/@local/hash-isomorphic-utils/src/system-types/block.ts
TimDiekmann, line 1: Could you add this and other generated file to a `.gitattributes` file to mark them as generated? E.g.
```
block.ts linguist-generated=true
comment.ts linguist-generated=true
...
```

TimDiekmann (commented):
I'm probably the wrong person to give a proper review to this but the changes are very exciting, good job! 🎉

apps/hash-api/src/graph/system-types.ts
```diff
@@ -10,6 +10,11 @@ import { logger } from "../logger";
 import { ImpureGraphContext } from "./index";
 import { entityTypeInitializer, propertyTypeInitializer } from "./util";
 
+/**
+ * IF YOU EDIT THIS FILE in a way which affects the number or structure of system types,
+ * run `yarn generate-system-types` to update their TypeScript representation
```
benwerner01, line 15: Ideally we would have a CI check that this has been run (similar to the CI check where we check the diff between `yarn.lock` files after `yarn install`). Maybe we can add a `@todo` for this?

libs/@local/hash-isomorphic-utils/.eslintrc.cjs
```diff
@@ -29,5 +29,12 @@ module.exports = {
         ],
       },
     },
+    {
+      files: ["./src/system-types/*.ts"],
+      rules: {
+        "simple-import-sort/exports": "off",
+        "simple-import-sort/imports": "off",
+      },
```
benwerner01, line 37: Is this to avoid eslint errors after running `yarn generate-system-types`? If so maybe we can just run `yarn fix:eslint` as part of that script, as these should be auto-fixable errors?

libs/@local/hash-isomorphic-utils/src/generate-system-types.ts
```diff
@@ -0,0 +1,40 @@
+// We need to import this file to load the environment variables –– ontology-types.ts relies on SYSTEM_USER_SHORTNAME.
+// This script is used in the backend only, but the outputs are used in the frontend, thus the following eslint-disable.
+// eslint-disable-next-line no-restricted-imports
+import "@local/hash-backend-utils/environment";
+
+import { codegen, CodegenParameters } from "@blockprotocol/graph/codegen";
+import slugify from "slugify";
+
+import { types } from "./ontology-types";
```
benwerner01, line 9: Just want to check you're aware we're not generating the types for the linear entity types here yet, is that something we're planning to do?

libs/@local/hash-isomorphic-utils/src/generate-system-types.ts
benwerner01, line 1: Running this script locally results in a diff, because prettier is not applied to the generated files as part of the script. Perhaps it's worth running this as part of the script, otherwise I for-see devs only realising this when the linting check fails in CI.

benwerner01 (commented):
Thanks Ciaran for your work on this, these types will be really useful 👍

The script is working for me locally, but had a couple of thoughts some of which may be worth addressing as part of this PR. Let me know what you think 🙂

apps/hash-api/src/graph/system-types.ts
TimDiekmann: For the Graph API's OpenAPI specs we do a similar thing. It's part of our testing suite and after all tests have run we enforce an empty `git diff`.

libs/@local/hash-isomorphic-utils/src/generate-system-types.ts
CiaranMn: Good idea – I've done so in https://github.com/hashintel/hash/pull/2852/commits/eb32f8c6e2e7ed6f76d88520d1b1fd58e5d0edf3

libs/@local/hash-isomorphic-utils/src/generate-system-types.ts
CiaranMn: I forgot these were separate – added in https://github.com/hashintel/hash/pull/2852/commits/eb32f8c6e2e7ed6f76d88520d1b1fd58e5d0edf3

These have to be generated in a separate step, because if we try and do it all at once we'll get clashes between Linear `User` and the system `User`, and they end up with `1`s added to them, which seems worse than just putting them in separate files.

libs/@local/hash-isomorphic-utils/.eslintrc.cjs
CiaranMn: Done in https://github.com/hashintel/hash/pull/2852/commits/eb32f8c6e2e7ed6f76d88520d1b1fd58e5d0edf3

apps/hash-api/src/graph/system-types.ts
CiaranMn: Todos added in https://github.com/hashintel/hash/pull/2852/commits/1ac1e2d4eebf4ea9f88090177ee4f3ff39a140a3, H-308

libs/@local/hash-isomorphic-utils/.eslintrc.cjs
CiaranMn: The ordering of properties is still non-deterministic. I'll worry about this in follow up.

benwerner01 (approved):
Thanks Ciaran, happy for this to be merged in. I'm still getting a diff when running the script on the order of the property types, but guessing this is not something we can easily fix as the order of the keys in the object aren't deterministic.

This is probably something we'd need to address as part of H-308 if we want to run a simple diff on the files as part of CI.

image

libs/@local/hash-isomorphic-utils/src/system-types/block.ts
CiaranMn: Good idea – I will do this in the next PR

PR #2862, opened 2023-08-04
H-180: Allow editing profile of an existing org

CiaranMn:

This PR adds a page at `/account/workspaces/[workspace-slug]` to edit org settings. The layout doesn't yet match t[he design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=8022-265030&mode=design&t=c85wXXHN3DxNVU5d-4) because we need to introduce a sidebar to switch between different account settings screens.

It also:
- introduces `next-seo` to manage meta tags on pages (without it there was race conditions between which `title` tag would take precedence, the one defined in `plain-layout` or the one defined in a child component)
- introduces a `simplifyProperties` function and uses it in `user-and-orgs.ts` so that we get type safety when accessing user and org entity properties. this revealed a couple of bugs which are fixed: `preferredName` should be optional and `email` is an array, not a single value

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

1. ~The output of `yarn generate-system-types` is not deterministic, because the order of (1) properties in an object and (2) members of a union can change. This might be because one or both of (1) the codegen utility (H-314) and (2) the JSON saved in the API (H-316) are non-deterministic.~ https://github.com/hashintel/hash/pull/2863 
2. Generated types cannot be used to auto-suggest/complete or check the correct value for property keys or type ids, because these both rely on an environment variable. This means constructing the property object for an update still requires using the `SYSTEM_TYPES` object. It would be possible to have a runtime function that converted back from the simple properties object to a complete one, generated as part of the codegen utility, with some caveats (an object, when simplified, might not contain all possible keys, and if a value for a missing key was added later its key would need to be 'recomplicated' by assuming that its domain was the same as other keys passed in initially, since the correct domain is only available at runtime). 

## 🐾 Next steps

- Fix the above issues
- H-175: allow adding / removing org members

## ❓ How to test this?

1.  Visit `/account/workspaces/example-org` and edit something

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/0ac539d4-d15c-4147-9a20-d28885fea3eb


apps/hash-frontend/src/pages/shared/org-form.tsx
```diff
@@ -202,23 +206,63 @@ export const OrgForm = ({
...
+            hint="Provide a brief description of your organization"
+            htmlFor="description"
+          />
+          <TextField
+            id="description"
+            sx={{ width: 400 }}
+            {...register("description", { required: false })}
+          />
+        </InputGroup>
+      )}
```
benwerner01, line 222: Maybe we should add a `@todo` to make it possible to set the description/location on creation as well?

apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { authenticatedUser, refetch } = useAuthenticatedUser();
+
+  const org = authenticatedUser.memberOf.find(
+    (orgOption) => orgOption.shortname === shortname,
+  );
+
+  if (!org) {
+    void router.push("/");
+    return null;
+  }
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx:35]: Might be nice to have a 404 page instead at some point, maybe we can add a `@todo` for now

benwerner01 (approved):
Thanks Ciaran, looks good and is working for me locally. Have left a couple of small thoughts which shouldn't block merging.

apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx
```diff
@@ -373,23 +373,21 @@ const Page: NextPageWithLayout<PageProps> = ({
 
   return (
     <>
-      <Head>
-        <title>{pageTitle || "Untitled"} | HASH</title>
-
-        {/*
-          Rendering favicon.png again even if it's already defined on _document.page.tsx,
-          because client-side navigation does not fallback to the default icon when visiting a page without an icon
-        */}
```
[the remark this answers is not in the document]
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/[page-slug].page.tsx:382]: I removed it because we're no longer doing what is described (including the default `favicon.png` link if no `icon` is set for the page) – the issue with the browser not refreshing the favicon properly is handled in the layout instead, [here](https://github.com/hashintel/hash/blob/e2a0cdecb2791b5b8b5d7ef2a17540714937e04f/apps/hash-frontend/src/shared/layout/plain-layout.tsx#L48)

apps/hash-frontend/src/pages/shared/org-form.tsx
CiaranMn: These are not in the design for the create org screen ([here](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=8022-264328&mode=design&t=8TUgf14VHztIChEf-4)) so I believe they are intentional omissions

apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx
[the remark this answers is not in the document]
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx:1]: Yes, I'm thinking we will have the tab name as part of the URL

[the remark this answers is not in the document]
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx:1]: Although without the tab name will still work for arriving here

apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/general.page.tsx]: Done in https://github.com/hashintel/hash/pull/2862/commits/c53b568a046274c2c8d5c794ca492a1e2e08d86e

PR #2864, opened 2023-08-07
Add python package for `hash-graph-sdk`

indietyp:

As discussed offline, this creates the `hash-graph-sdk` package, which offers an ergonomic high-level interface to interact with the API.

The same strategy as outlined in #2841 has been adopted to reduce churn, meaning that the project is nested under an additional `python` directory to stay consistent.

For now, this PR does not implement _any_ functionality.

As discussed with @nonparibus offline, the package is to be licensed under the AGPL.

## 🔍 What does this change?

- add new python package

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann (approved):
Looks good to me, thank you. I wonder if we want functionality like we have in `hash-subgraph` into this package at some point (not now or direct follow-up, just a consideration).

nonparibus (approved):

TimDiekmann (approved):

nonparibus (approved):

PR #2868, opened 2023-08-07
H-346: Add `hash-graph-sdk` Python functionality

indietyp:

This ports over the code from the internal repo into the public repo and implements a high-level convenient python interface.

This PR:

The changes in this PR:

The changes in this PR:

- [x] affect the execution graph

## ⚠️ Known issues

* Before this goes out of draft the examples need to be tested

## 🐾 Next steps

* Tests

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Try X
1.  Confirm that Y


libs/@local/hash-graph-sdk/python/graph_sdk/options.py
```diff
@@ -0,0 +1,372 @@
...
+                        start=start.to_temporal_bound() if start else None,
+                        end=end.to_limited_temporal_bound() if end else None,
+                    ),
+                ),
+            ),
+        )
+
+    def latest(self) -> QueryTemporalAxesUnresolved:
+        """Return all entities that are valid at the current time."""
+        return self.between(start=None, end=None)
```
TimDiekmann, line 203:
```suggestion
    def current(self) -> QueryTemporalAxesUnresolved:
        """Return all entities that are valid at the current time."""
        return self.between(start=None, end=None)
```

libs/@local/hash-graph-sdk/python/graph_sdk/options.py
```diff
@@ -0,0 +1,372 @@
...
+                    axis=TransactionTime(root="transactionTime"),
+                    interval=UnresolvedRightBoundedTemporalInterval(
+                        start=start.to_temporal_bound() if start else None,
+                        end=end.to_limited_temporal_bound() if end else None,
+                    ),
+                ),
+            ),
+        )
+
+    def latest(self) -> QueryTemporalAxesUnresolved:
```
TimDiekmann, line 242:
```suggestion
    def current(self) -> QueryTemporalAxesUnresolved:
```

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
```diff
@@ -0,0 +1,352 @@
...
+    FilterExpressionItem1,
+    FilterItem,
+    FilterItem1,
+    FilterItem2,
+    FilterItem3,
+    FilterItem4,
+    FilterItem5,
+    FilterItem6,
+    FilterItem7,
+)
```
TimDiekmann, line 36: Similar to above, this is horrific. Any way to provide aliases to it?

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @classmethod
+    def always(cls) -> "NaryFilter[Never]":
+        """Returns a query that always returns true."""
+        return NaryFilter(NaryOperation.ALL, [])
+
+    @classmethod
+    def never(cls) -> "NaryFilter[Never]":
+        """Returns a query that always returns false."""
+        return NaryFilter(NaryOperation.ANY, [])
```
TimDiekmann, line 127: Why you would ever use this? I'd just delete it.

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
```diff
@@ -0,0 +1,352 @@
...
+        return UnaryFilter(UnaryOperation.NOT, self)
+
+    def __invert__(self) -> "UnaryFilter[Self]":
+        """Negates the query."""
+        return self.not_()
+
+    @classmethod
+    def always(cls) -> "NaryFilter[Never]":
+        """Returns a query that always returns true."""
+        return NaryFilter(NaryOperation.ALL, [])
```
TimDiekmann, line 122: Probably not possible now but I like to remove this ASAP.

libs/@local/hash-graph-sdk/python/scripts/generate_blocking.py
```diff
@@ -0,0 +1,225 @@
+# type: ignore
+"""Script to generate the blocking implementation of the client."""
```
TimDiekmann, line 2: Why do we need a sync client in the first place?

yarn.lock
TimDiekmann, line 1: Would be good if you could deduplicate the versions.

TimDiekmann (commented):
Just skimmed through the PR as it's pretty big and still in draft.

libs/@local/hash-graph-sdk/python/graph_sdk/filter.py
TimDiekmann, line 1: Is this autogenerated? If not, is it possible to autogenerate it?

indietyp: Right now, `filter.py` is hand-crafted artisan code, but I can write something similar to `scripts/generate_blocking.py` (which was basically for me like a playground to test if it works).

libs/@local/hash-graph-sdk/python/scripts/generate_blocking.py
TimDiekmann: Discussed offline.

libs/@local/hash-graph-sdk/python/graph_sdk/client/_compat.py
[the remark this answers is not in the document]
indietyp: I agree 110%; this is bad, _really bad_, but at least it's consistent across `datamodel-code-generator` invocations.

I wanted to tackle this in an upcoming PR. To be able to do so, we need to do a couple of things:

A) move out any duplicate definitions into their own schema file and then reference them (that way no duplicates are generated) - this will get us there ~50%
B) Give everything a unique `$id`, that is **not** a URL


With both of them, it should be a lot cleaner.

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
indietyp: It's the only way to ever get all entities, regardless of query, so I'd like to keep it, to make it clear that that's the case.

libs/@local/hash-graph-sdk/python/graph_sdk/filter.py
indietyp: I'd like to get the code in first, then tackle this and https://github.com/hashintel/hash/pull/2868#discussion_r1286171146 in two separate PRs, otherwise this PR gets very big.

indietyp: #2872 has already been opened; the generation PR needs to wait _until_ this PR has been merged

indietyp: Addressed in #2878

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
TimDiekmann: This is why I'm asking, returning all entities is discouraged as this will become pretty big.

libs/@local/hash-graph-sdk/python/graph_sdk/types.py
```diff
@@ -0,0 +1,207 @@
...
+        self,
+        data_type_id: str,
+        *,
+        actor_id: UUID,
+        is_after_load: bool = False,
+    ) -> DataTypeSchema:
+        """Returns the data type schema for the given data type ID.
+
+        If the data type is not found it will attempt to fetch it and use
+        the actor ID to authenticate the request.
```
TimDiekmann, line 58: This is supposed to be implemented in the Graph API. Could you add a TODO to remove this once [H-136](https://linear.app/hash/issue/H-136) is done? (Same for property types and entity types)

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
indietyp: That is true, but for now (as you noted above), it is still needed in some places.

We could add something to the docs that it is discouraged, would that help? (and remove `never`)

TimDiekmann: Yeah, let's remove `never` and add a note. We can remove it once it's not required anymore 🚀

libs/@local/hash-graph-sdk/python/graph_sdk/types.py
indietyp: addressed in [577146f](https://github.com/hashintel/hash/pull/2868/commits/577146f5ba49ea34e9db3af7ae67a7fc9a47c267)

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
indietyp: addressed in [577146f](https://github.com/hashintel/hash/pull/2868/commits/577146f5ba49ea34e9db3af7ae67a7fc9a47c267)

libs/@local/hash-graph-sdk/python/graph_sdk/query.py
indietyp: removed in [577146f](https://github.com/hashintel/hash/pull/2868/commits/577146f5ba49ea34e9db3af7ae67a7fc9a47c267)

TimDiekmann (approved):

PR #2877, opened 2023-08-08
H-175: Org membership management page

CiaranMn:

Introduces a new page where org members:
- are listed
- can be removed
- can be added
- can have their 'responsibility' edited (this is a string, not an application role)

## 🔍 What does this change?

- Moves `zeroedGraphResolveDepths` and the time axis helpers to a shared lib (we were repeating `zeroedGraphResolveDepths` in the frontend and we will want the time axis helpers there at some point also)
- Adds the table at `/account/workspaces/[org-shortname]/members`
- Patch `getEntities` in the API to filter out archived vertices – the current approach to archiving only filters on the roots, and so without this additional filtering if you archive a link, querying for something that has that link will still return the link.
- Updates a couple of entity resolvers so that they go through that `getEntities` function and don't use the graph API client directly (so that archived things also get filtered out)
- Adds a new `MainContainer` component for constraining the width of the main content, which is currently wider than desired – more to be done on layout in a follow up

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

1. There is still no way of reaching these pages – the next PR will add the settings sidebar, and also rename 'Workspaces' to 'Organizations', and 'Account' to 'Settings'
2. As with the rest of the app, there is no permissioning, and anyone can add or remove people from any org, edit their responsibility, etc.

## 🐾 Next steps

- Add the sidebar for the settings page

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1.  Visit `/account/workspaces/example-org/members` (assuming you have kept `example-org` as the system user shortname locally)

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/755985e6-2b24-4968-8159-d06da1c54e14


nonparibus:
@CiaranMn Not sure where the `Responsibility` column came from, but I think we can lose it (a port of the existing hIndex member system? It's absent from the HASH designs)

CiaranMn:
> @CiaranMn Not sure where the `Responsibility` column came from, but I think we can lose it (a port of the existing hIndex member system? It's absent from the HASH designs)

@nonparibus it was in the previous org system for HASH, see e.g. [this design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=3663-177961&mode=dev). we can get rid of it if we're not going to want anything like it

nonparibus:
> 

@CiaranMn Ah, I see. That particular question in the sign up/join an org flow is intended to be user self-selection into some 'function' within an organization (e.g. Marketing, Sales, HR, Engineering, etc.) so that we can pre-populate their workspace with more tailored suggested pages and integrations, from the get-go. It's not intended to be displayed to users anywhere in the interface, and should be an enum we set (and ask for only at the point of sign-up/joining their org, with users unable to edit it after the fact).

nonparibus:
@CiaranMn Actual designs for the members list are in Figma [here](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=8023%3A265974&mode=design&t=cEXN2JzPHD4sjLHT-1) (internal link).

apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx
Comment (benwerner01) [inline apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx:1]: @CiaranMn I've seeded and rebuilt my local environment, but am getting an error when trying to remove a user from an org (the second time the button is clicked). Is this is something you can reproduce? Perhaps this is caused by the table not updating correctly for some reason, and then an error is thrown when you try to remove a user from an org they're not a member of?

https://github.com/hashintel/hash/assets/42802102/dc569406-0848-480d-bd6e-c461871fa68b

benwerner01 (changes requested):

nonparibus:
@CiaranMn We should update the code and URL slugs to read `orgs` instead of `workspaces` btw, and in-app copy to read `Organizations` (which can later be localized to `Organisations` for International/British English folks).

apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx:1]: It looks like the API is still returning archived entities for some reason, which causes an error when you try and re-archive something that is already archived. I cannot recreate this – can you check if the API is returning archived things in the return (see video)? If it is can you try killing and restarting the API? not sure why it is happening. it'll be `me` directly after removing someone, or if you want to check on initial load have a look for `queryEntities` with a filter to an org type


https://github.com/hashintel/hash/assets/37743469/3f46fd38-78d1-451b-8de2-2b35b93d0426

CiaranMn:
@nonparibus I've removed the `responsibility` property in https://github.com/hashintel/hash/pull/2877/commits/b93e50cfe0a52d87cef9c53e2df2f3cb5c112d06

> We should update the code and URL slugs to read orgs instead of workspaces btw, and in-app copy to read Organizations 

Agreed, I will do this in the next PR which will introduce the sidebar and sort the layout of settings out

apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx:1]: After deleting all my `node_modules` directories and reinstalling it's working now, guessing the patch wasn't applying correctly previously. Thanks!

apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    a.user.preferredName?.localeCompare(
+                      b.user.preferredName ?? "ZZZ",
+                    ) ?? 1,
+                )
+                .map((membership) => (
+                  <MemberRow
+                    key={membership.membershipEntity.metadata.recordId.entityId}
+                    membership={membership}
+                    self={
+                      membership.user.shortname === authenticatedUser.shortname
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx:105]: Just a thought: I wonder if it'd be slightly safer to compare on `accountId` here, as `shortname` can change and in an extreme edge case it might be possible for the `authenticatedUser` to have an updated value whilst the `membership` still has an outdated one.

benwerner01 (approved):
Thanks Ciaran, code looks good to me and is working for me locally now. Still not sure why it was broken for me previously, apologies for that.

apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/account/workspaces/[shortname]/members.page.tsx]: Done in https://github.com/hashintel/hash/pull/2877/commits#:~:text=update%20%27self%27%20check%20in%20membership%20table

benwerner01 (approved):

Issue #1968, opened 2023-01-29
H-357: Allow `change_context` to call `into_report`


bjchambers:
### Related Problem

I find myself often writing code that looks like:

```
some_call()
  .into_report()
  .change_context(Error::A)?;
some_other_call()
  .into_report()
  .change_context(Error::B)?;
...
```

This arises when calling libraries that don't return errors using error stack (or worse -- do so using `anyhow` since I'm in the middle of migrating).

### Proposed Solution

I think it should be possible to implement `change_context` for types that are not already `Report<E>`. For instance, if reports and `anyhow::Error` could be used with `IntoReport`, then the `ResultExt` trait could be implemented for any result containing an error that implements `IntoReport`. This would allow the original code to be written as:

```
some_call()
  .change_context(Error::A)?;
some_other_call()
  .change_context(Error::B)?;
...

### Alternatives

_No response_

### Additional context

_No response_

TimDiekmann:
The situation is a bit tricky. `into_report()` or `Report::new()` does more than just put an error `E` into `Report<E>`. If necessary, a `Backtrace` or even a `SpanTrace` will be captured. `Report::change_context` (or the corresponding method on the `Result`) will “only” capture a new `Frame` and put it in the report, which should be pretty cheap. So calling `change_context()` directly on a non-`Report`-`Result` would hide from the user that possibly more happens than putting the error into a new frame. We have assumed `into_report` so far because it is more explicit to predict the behavior. In fact, `into_report()` should not be called at all when in the hot path, so this precaution is perhaps a bit too verbose.

What I personally would like to see is a method that does both, i.e. an `into_report_with_context<C>(C)`, but I have not been able to find a name for this method that is sufficiently short (obviously, the above is too long for a method that is needed relatively often), but still describes the behavior well enough, since this method is defined on `Result<T, E>` and that is an enormously frequently used type. Also, the interface is getting even bigger.

Personally, I don't have a strong opinion on this and I know that `into_report().change_context(C)` is quite annoying. However, from personal experience, I have found that the migration itself is quite difficult as many types rely on non-`error-stack` types so you have to call `into_report()` quite often, but this is better once you can rely on your own functions that rely on `error-stack` (and of course any other crate, which using `error-stack`) and I quite like the compile-time checked validation, that the error is wrapped in a `Report`.

Happy to hear your thoughts on that!

---

[...]

PR #2883, opened 2023-08-09
GEN-37: Combine `IntoReport` and `ResultExt`

TimDiekmann:

After a lot of back and forth and various discussions (internally and on GitHub and other public platforms), we decided to merge the functionality of `IntoReport` into `ResultExt`.

## 🔗 Related links

- Closes https://github.com/hashintel/hash/issues/1968

## 🔍 What does this change?

- Extend `ResultExt` by a `type Context: Context` associated type (technically a breaking change likely has no impact) - requires a minor version bump
- Implement `ResultExt` for `core::result::Result<T, C> where C: Context`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- We still have `IntoReportCompat` but due to the Orphan rule it's not possible to implement `ResultExt` on `eyre` or `anyhow` errors.

### DIFF
--- a/libs/error-stack/src/compat.rs
+++ b/libs/error-stack/src/compat.rs
@@ -13,13 +13,12 @@ mod eyre;
 
 /// Compatibility trait to convert from external libraries to [`Report`].
 ///
-/// *Note*: It's not possible to implement [`IntoReport`] or [`Context`] on other error libraries'
-/// types as both traits have blanket implementation relying on [`Error`]. Thus, implementing either
-/// trait would violate the orphan rule; the upstream crate could implement [`Error`] and this would
-/// imply an implementation for [`IntoReport`]/[`Context`].
+/// *Note*: It's not possible to implement [`Context`] on other error libraries' types as the trait
+/// has a blanket implementation relying on [`Error`]. Thus, implementing the trait would violate
+/// the orphan rule; the upstream crate could implement [`Error`] and this would imply an
+/// implementation for [`Context`].
 ///
 /// [`Report`]: crate::Report
-/// [`IntoReport`]: crate::IntoReport
 /// [`Context`]: crate::Context
 /// [`Error`]: core::error::Error
 pub trait IntoReportCompat: Sized {
--- a/libs/error-stack/src/compat.rs
+++ b/libs/error-stack/src/compat.rs
@@ -13,12 +13,14 @@ mod eyre;
 
 /// Compatibility trait to convert from external libraries to [`Report`].
 ///
-/// *Note*: It's not possible to implement [`Context`] on other error libraries' types as the trait
-/// has a blanket implementation relying on [`Error`]. Thus, implementing the trait would violate
-/// the orphan rule; the upstream crate could implement [`Error`] and this would imply an
-/// implementation for [`Context`].
+/// **Note**: It's not possible to implement [`Context`] on other error libraries' types from within
+/// `error-stack` as the trait has a blanket implementation relying on [`Error`]. Thus, implementing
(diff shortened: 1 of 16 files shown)


libs/error-stack/src/report.rs
```diff
@@ -430,7 +427,7 @@ impl<C> Report<C> {
     /// # #[cfg(all(feature = "std", not(miri)))] {
     /// use std::{fmt, fs};
     ///
-    /// use error_stack::{IntoReport, ResultExt};
+    /// use error_stack::{ResultExt};
```
indietyp, line 430:
```suggestion
    /// use error_stack::ResultExt;
```

libs/error-stack/src/result.rs
```diff
@@ -67,38 +71,123 @@ pub trait ResultExt {
...
+    type Ok = T;
+
+    #[track_caller]
+    fn attach<A>(self, attachment: A) -> Result<T, C>
+    where
+        A: Send + Sync + 'static,
+    {
+        match self {
+            Ok(value) => Ok(value),
+            Err(error) => Err(Report::from(error).attach(attachment)),
```
indietyp, line 123: Out of curiosity: in your mind: is `Report::new` or `Report::from` preferred? For me it's the former (`Report::new`)

libs/error-stack/src/compat.rs
```diff
@@ -13,13 +13,12 @@ mod eyre;
 
 /// Compatibility trait to convert from external libraries to [`Report`].
 ///
-/// *Note*: It's not possible to implement [`IntoReport`] or [`Context`] on other error libraries'
-/// types as both traits have blanket implementation relying on [`Error`]. Thus, implementing either
-/// trait would violate the orphan rule; the upstream crate could implement [`Error`] and this would
-/// imply an implementation for [`IntoReport`]/[`Context`].
+/// *Note*: It's not possible to implement [`Context`] on other error libraries' types as the trait
```
indietyp, line 16: Do we maybe want to make this a heading instead?

libs/error-stack/src/lib.rs
```diff
@@ -234,7 +228,7 @@
 //! ```rust
 //! # #![cfg_attr(not(feature = "std"), allow(dead_code, unused_variables, unused_imports))]
 //! # use std::{fs, path::Path};
-//! # use error_stack::{IntoReport, Report};
+//! # use error_stack::{Report};
```
indietyp, line 231:
```suggestion
//! # use error_stack::Report;
```

indietyp (changes requested):
Minor things, otherwise looking good. Thank you!

indietyp:
Also: We still need a changelog entry!

libs/error-stack/src/result.rs
TimDiekmann: I don't really have an opinion. Maybe we should remove `new`?

libs/error-stack/src/compat.rs
TimDiekmann: This is the second sentence (and only description) on a trait that is only used for compatibility. I would leave it as it is. I, however, generally improved the documentation a bit.

libs/error-stack/src/result.rs
TimDiekmann: Keeping it does not hurt, that's true.

TimDiekmann: Actually, I think `from` is more intuitive in this context as an error already exists and we create a Report from something which already exists. For simplicity, we can keep `new`, though.

indietyp (dismissed):

indietyp (approved):

PR #2885, opened 2023-08-10
GEN-38: Minor improvements of code quality

TimDiekmann:

There are a bunch of locations in `error-stack` which could be improved slightly. The amount of changes adds up and it's worth doing this in a separate PR.

## 🔍 What does this change?

- Clean up imports
- Remove unneeded `repr(C)` (we don't rely on it anymore and if we would we should use `repr(transparent)`
- Remove feature flags in doc test except for `nightly` checks (we only run doc tests for stable+nightly with `--all-features`; no `miri` and no feature permutations
- Some other minor changes

This PR:

The changes in this PR:

The changes in this PR:


indietyp (approved):
Big QoL, especially the doc-tests have been a pain previously.

PR #2887, opened 2023-08-11
GEN-39: Update Renovate config

vilkinsons:

Updates our Renovate configuration file.

Additionally makes drive-by changes to:

- the roadmap page we link to from the root README, as well, to direct GtiHub users to our development roadmap at [hash.dev/roadmap](https://hash.dev/roadmap) over the current (and less up-to-date) [hash.ai/roadmap](https://hash.ai/roadmap) page
- the `libs` README's description of our _internal libraries_ published to npm.

## 🔗 Related links

- Renovate's docs detailing its [Configuration Options](https://docs.renovatebot.com/configuration-options/)

## 🚫 Blocked by

- #2891 merging. We will try re-running the CI lint fix script once it's in, using this PR as the test one for that.

## 🔍 What does this change?

- **Branch prefixes:** renovate shortened to deps, and env-specific prefixes appended
- **Groups:** new groups added

Previously had intended to add self-labeling here, but we have decided to keep this centralized in the `labeler.yml` file to avoid splitting where this occurs.

- [x] **No.** This PR does not modify any publishable blocks or libraries, or modifications do not need publishing

- [x] **No.** This PR is internal and does not require a docs change

- [x] **I am unsure and need a bit of an explainer.** I doubt it does, but I've never worked with the Turbo Graph before and don't know how it works.

## ⚠️ Known issues

- None ✅

## 🐾 Next steps

- Ideally find a way to test this out before merging, or subject it to proper code review.

## 🛡 What tests cover this?

- None that I am aware of.

## ❓ How to test this?

- Unsure how we can without merging to `main` 🫣

## 📹 Demo

- _n/a_

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,7 +12,7 @@
   "rebaseWhen": "conflicted",
   "semanticCommits": "disabled",
   "updateInternalDeps": "true",
-  
+
   "packageRules": [
     {
       "matchManagers": ["npm"],


.github/renovate.json
```diff
@@ -11,6 +13,24 @@
...
+      "groupName": "Block Protocol",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": [
+        "^@blockprotocol/",
+        "^block-",
+        "blockprotocol",
+        "create-block-app",
+        "mock-block-dock"
+      ],
+      "addLabels": "area: blocks"
```
TimDiekmann, line 32: Isn't this label added by the labeler flow anyway? I'd rather keep it in one location.
```suggestion
      ]
```

nonparibus: Let me confirm. Stupidly didn't really think about the `labeler` as I implemented this, but I 100% agree it would be best to have a single source of truth.

TimDiekmann: The list is not as big as your new list but we can update that instead:
https://github.com/hashintel/hash/blob/12fecc40c71cf21350af50e198d58c8baadaadfc/.github/labeler.yml#L27-L28

TimDiekmann:
> Unsure how we can without merging to main 🫣

Not worth it. It's possible by forking it and setting up renovate for the fork. I wouldn't do it.

.github/renovate.json
nonparibus: @TimDiekmann I agree with your principle of removing all the `addLabels` from this PR, where we can achieve the same thing through better rules in `labeler.yml`. Would you mind taking the sister task at some point in the future? Captured as GEN-40 (extremely low prio... I was only picking this low-prio ticket up in the first place as I needed to clear my mind with something mundane for ~20 before getting back to writing).

TimDiekmann: Sure! I added a back-link to this issue so I'm able to find it again when picking it up.

.github/renovate.json
```diff
@@ -20,35 +38,72 @@
         "^@typescript-eslint/",
         "^eslint-",
         "^eslint$"
-      ]
+      ],
```
TimDiekmann, line 41:
```suggestion
      ]
```

.github/renovate.json
```diff
@@ -20,35 +38,72 @@
...
+      "groupName": "GraphQL",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": ["^@graphql/", "^graphql$", "^graphql-"]
     },
     {
       "enabled": true,
       "groupName": "Jest",
       "matchManagers": ["npm"],
-      "matchPackagePatterns": ["^@types/jest$", "^jest", "^ts-jest$", "^jest-"]
+      "matchPackagePatterns": ["^@types/jest$", "jest", "^jest-", "-jest$"],
```
TimDiekmann, line 53:
```suggestion
      "matchPackagePatterns": ["^@types/jest$", "jest", "^jest-", "-jest$"]
```

.github/renovate.json
```diff
@@ -20,35 +38,72 @@
...
+      "groupName": "OpenTelemetry",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": ["^@opentelemetry/"]
     },
     {
       "enabled": true,
       "groupName": "Playwright",
       "matchManagers": ["npm"],
-      "matchPackagePatterns": ["^@playwright/", "^playwright$", "^playwright-"]
+      "matchPackagePatterns": ["^@playwright/", "^playwright$", "^playwright-"],
```
TimDiekmann, line 65:
```suggestion
      "matchPackagePatterns": ["^@playwright/", "^playwright$", "^playwright-"]
```

.github/renovate.json
```diff
@@ -20,35 +38,72 @@
...
+    },
+    {
+      "enabled": true,
+      "groupName": "Prettier",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": ["^prettier"],
```
TimDiekmann, line 71:
```suggestion
      "matchPackagePatterns": ["^prettier"]
```

.github/renovate.json
```diff
@@ -20,35 +38,72 @@
...
     },
     {
       "enabled": true,
       "groupName": "ProseMirror",
       "matchManagers": ["npm"],
-      "matchPackagePatterns": ["^prosemirror-"]
+      "matchPackagePatterns": ["^prosemirror$"]
```
TimDiekmann, line 77: Was this change intentional or did you revert the wrong pattern?

.github/renovate.json
```diff
@@ -22,33 +40,70 @@
...
+      "groupName": "GraphQL",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": ["^@graphql/", "^graphql$", "^graphql-"]
+    },
     {
       "enabled": true,
       "groupName": "Jest",
       "matchManagers": ["npm"],
-      "matchPackagePatterns": ["^@types/jest$", "^jest", "^ts-jest$", "^jest-"]
+      "matchPackagePatterns": ["^@types/jest$", "jest", "^jest-", "-jest$"]
```
TimDiekmann, line 53: What was the reason for this change? `"jset"` is a very generic pattern (which will actually cover all the other patterns in this line as well).

nonparibus: I was hoping to avoid potential false positives, but am realizing that the logic here could be simplified in any case.

And actually, upon further reflection, if we end up with false positives we can just refine it then. Thanks for flagging. Amended in https://github.com/hashintel/hash/commit/198b1c5e5b7dcecfb70d12e7a06d4661d1850708

.github/renovate.json
nonparibus: We have some packages that end in `prosemirror`, as well as start, so I thought this would be an improvement.

.github/renovate.json
```diff
@@ -22,33 +40,70 @@
         "^eslint$"
       ]
     },
+    {
+      "enabled": true,
+      "groupName": "GraphQL",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": ["^@graphql/", "^graphql$", "^graphql-"]
```
nonparibus, line 47:
```suggestion
      "matchPackagePatterns": ["^@graphql/", "^graphql$"]
```

It's redundant, right? @TimDiekmann

.github/renovate.json
TimDiekmann: I'm pretty sure the amended version is not what we want. With this we won't capture `@types/jest`, `ts-jest`, `eslint-plugin-jest`, `babel-jest`, `jest-junit`, `jest-message-util` (various others).
```suggestion
"matchPackagePatterns": ["^@types/jest$", "^jest-", "-jest$"]
```

nonparibus: My bad. Actually fixed, not broken more, in https://github.com/hashintel/hash/pull/2887/commits/cb8bc7a73ef70dc0cedfdde7071cd6e2413cbaf0

TimDiekmann (approved):
LGTM, let's test it soon-ish 🙂

.github/renovate.json
```diff
@@ -19,36 +37,90 @@
...
+      "matchManagers": ["poetry"],
+      "commitMessageTopic": "Python package `{{depName}}`",
+      "additionalBranchPrefix": "py/"
+    },
+    {
+      "enabled": true,
+      "groupName": "Pydantic",
+      "matchManagers": ["poetry"],
+      "matchPackagePatterns": ["^pydantic"]
+    },
```
nonparibus, line 118: @TimDiekmann Could I have combined these into one? I think possibly yes, right?

TimDiekmann: Which one do you mean, the python related groups? I don't see how we could combine them as you add a group name for `Pydantic`. I don't know if we need that, though 🤷🏻‍♂️

nonparibus:
@TimDiekmann For testing:
- opening a Python PR to test branch prefixing, team assignment and commit/PR titling 
- opening a Rust PR to test branch prefixing and team assignment
- opening a JS/TS PR to test branch prefixing (could do one or more of new/changed groups specifically: `Block Protocol`, `ESLint`, `GraphQL`, `Jest`, `Playwright`, `Prettier`, `ProseMirror` and `Signia`, although we'll see their groupings in the dependency dashboard anyway)

.github/renovate.json
nonparibus: My brain is literally not working this morning. I'm going to take a nap.

TimDiekmann (approved):

TimDiekmann (dismissed):

.github/renovate.json
```diff
@@ -11,6 +12,23 @@
...
+      "groupName": "Block Protocol",
+      "matchManagers": ["npm"],
+      "matchPackagePatterns": [
+        "^@blockprotocol/",
+        "^block-",
+        "^blockprotocol$",
+        "^create-block-app$",
+        "^mock-block-dock$"
+      ]
+    },
```
TimDiekmann, line 31: I just noted that `groupName` does not mean the packages are put in the same section but a PR for all these files will be created at once. In this case this is likely not what as we for example don't want to update all `@blockprotocol` at once.
```suggestion
```

TimDiekmann: The PR is still approved, just hit `merge` if you disagree with me 🙂

nonparibus: Why wouldn't we want to merge all Þ packages imported into the # repo at once? cc @CiaranMn

CiaranMn: If we keep on top of it, it should be fine to upgrade all BP packages at once, because they're mostly interconnected, and new releases of some will often just be upgrading the version of another one, in which case it makes sense to do them together.

If there are releases of independent packages (e.g. `@blockprotocol/hook` and `@blockprotocol/events`) then it becomes a pain because you have to do unrelated work in the same PR to fix breaking changes or whatever is required, and you might end up splitting it out.

I don't think it particularly matters either way. We can pick one approach and change it later if it does.

TimDiekmann: This means, that we cannot use renovate anymore to bump the type-system package as this would require updating all other BP packages. Yes, most packages are connected, but not all.

.github/renovate.json
```diff
@@ -19,36 +39,74 @@
...
+      "groupName": "Pydantic",
+      "matchManagers": ["poetry"],
+      "matchPackagePatterns": ["^pydantic"]
+    },
     {
       "matchManagers": ["cargo"],
-      "commitMessageTopic": "Rust crate `{{depName}}`"
+      "commitMessageTopic": "Rust crate `{{depName}}`",
+      "additionalBranchPrefix": "rs/",
+      "reviewers": ["team:Rust"]
```
TimDiekmann, line 109: Do we want to add this to the toolchain groups as well?

TimDiekmann: (The branch prefix and the reviewers)

nonparibus: @TimDiekmann I think this is done in https://github.com/hashintel/hash/commit/b8d865a7105bd4cb721672d9d6ad3b2321a6a4af -- if you could check I've caught all Rust crates as well as the toolchain updates which use the custom regex manager approach that'd be great.

TimDiekmann: Looks good, let's try it 🙂

TimDiekmann (approved):

nonparibus:
@TimDiekmann I'm re-running lint fix on this now ([job](https://github.com/hashintel/hash/actions/runs/5855308430/job/15872853929)). Will add to the merge queue when done.

.github/renovate.json
```diff
@@ -57,6 +118,8 @@
       "separateMajorMinor": false,
       "fetchReleaseNotes": false,
       "commitMessageExtra": "to nightly-{{newVersion}}"
```
TimDiekmann, line 120:
```suggestion
      "commitMessageExtra": "to nightly-{{newVersion}}",
```

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #2880, opened 2023-08-09
H-264: Implement type archival/unarchival in the frontend

benwerner01:
## 🌟 What is the purpose of this PR?

This PR implements type archival in the FE for entity types.

## 🚫 Blocked by

- [x] merging https://github.com/hashintel/hash/pull/2876
- [x] merging https://github.com/hashintel/hash/pull/2888

## 🔍 What does this change?

- refactors the `TopContextBar` component to make it re-usable for archival across entity types, pages, and entities (entity archival not yet implemented)
- moves the archive banner into the `TopContextBar` to make it re-usable across entity types, pages and entities
- refactors `queryEntityTypesResolver` to accept an optional `includeArchived` param which will include archived types in the response
- refactors global `EntityTypesContext` to get all latest types including archived types

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Checkout the branch, login to the FE
2. Create a type, and archive it via the context bar. Observe the type being removed from the sidebar
3. Unarchive the type via the context bar, and observe the type re-appearing in the sidebar

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/aacc6361-6f14-4c52-a3ae-8d9766613d26

PR #2888, opened 2023-08-11
H-372: Work around that subgraph can only contain a single ontology type per versioned URL, 2nd Attempt

TimDiekmann:

#2879 did not cover all possible paths. Instead of manually comparing the time it will now rely on the ordering returned from the database. The latest record will always be returned first.

## 🔗 Related links

- Reverts #2879

## 🔍 What does this change?

Correctly pick the ontology type when traversing

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/ontology.rs
+++ b/apps/hash-graph/lib/graph/src/ontology.rs
@@ -393,22 +393,6 @@ pub enum CustomOntologyMetadata {
     },
 }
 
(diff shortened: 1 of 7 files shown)


benwerner01 (approved):
Thanks Tim for your work on this, this fixes the type archival FE in [my branch](https://github.com/hashintel/hash/pull/2880) :)

PR #2889, opened 2023-08-11
H-121: Define Tables required for inheritance

TimDiekmann:

To allow entity type inheritance in queries it's useful to have inheritance information stored in the database. This consists of:

- Closure schemas for types, which contains the collapsed schema of a type
- reference tables to also link to indirect references

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/2881
## 🔍 What does this change?

- Drop old closure tables - they can be integrated in existing tables
- Add new columns to edges-tables and entity type table
- Initialize the tables with default values. This works as we currently don't have a way to insert inheritance other than the link entity type which does not inherit any edges.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Make sure only the correct edges are queried when traversing the tables
- Populate the edges with higher values


CiaranMn (dismissed):

CiaranMn (dismissed):

TimDiekmann:
Disabled auto-merge for now in case further changes have to be made to the migration files. I don't expect it but just in case this would avoid another new migration file in the future.

CiaranMn (approved):

PR #2893, opened 2023-08-14
Use `workspace` to define some common dependencies

TimDiekmann:

When changing dependencies in Cargo.toml every Cargo.toml in a workspace has to be adjusted to use the correct dependencies. This should be moved to the top-level manifest instead.

## 🔍 What does this change?

- With some revision bumps for the `type-system` package in the next hours this will simplify the changes required for it.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/Cargo.toml
+++ b/apps/hash-graph/Cargo.toml
@@ -13,6 +13,11 @@ license-file = "LICENSE.md"
 publish = false
 repository = "https://github.com/hashintel/hash/tree/main/apps/hash-graph"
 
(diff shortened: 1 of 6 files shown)


indietyp:
Very much in favor of this change. For anyone curious, renovate supports workspace dependencies since 34.53.0 (2022-12-09).

TimDiekmann:
Thanks for verifying @indietyp. Would you mind approving this PR?

indietyp (approved):

PR #2894, opened 2023-08-14
H-363: Adjust queries in Graph to only return direct references from edge tables

TimDiekmann:

With the edges-tables containing inheritance information, by default only direct references should be returned by default. Structural queries are going to be adjusted later to support the inheritance depth information.

## 🚫 Blocked by

- #2889 

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (dismissed):

apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
```diff
@@ -886,11 +975,11 @@ pub enum Column<'p> {
     EntityEditions(EntityEditions<'p>),
     PropertyTypeConstrainsValuesOn(PropertyTypeConstrainsValuesOn),
     PropertyTypeConstrainsPropertiesOn(PropertyTypeConstrainsPropertiesOn),
-    EntityTypeConstrainsPropertiesOn(EntityTypeConstrainsPropertiesOn),
-    EntityTypeInheritsFrom(EntityTypeInheritsFrom),
-    EntityTypeConstrainsLinksOn(EntityTypeConstrainsLinksOn),
-    EntityTypeConstrainsLinkDestinationsOn(EntityTypeConstrainsLinkDestinationsOn),
-    EntityIsOfType(EntityIsOfType),
+    EntityTypeConstrainsPropertiesOn(EntityTypeConstrainsPropertiesOn, u32),
```
indietyp, line 978: I know this is the inheritance depth, but this might not be so clear to a person new to the codebase. Maybe switch to a struct variant or employ the newtype pattern?

indietyp (changes requested):
Minor comment otherwise LGTM

apps/hash-graph/lib/graph/src/store/postgres/query/table.rs
indietyp: Sounds good to me

indietyp (approved):

PR #2908, opened 2023-08-15
H-265: Add `Type` list pages

benwerner01:

This PR adds a `/types` page, where entity types, link types, property types and data types are listed in a table.

## 🔍 What does this change?

- adjusts the sidebar styling for the "Types" section to be closer to the updated figma designs
- uses the `millify` package to abbreviate the number displayed in the `TabLink` component, similar to how it can be seen in the designs (`1k` instead of `1000`, `1M` instead of `1000000`, etc)

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

Checkout the branch, login and visit the `/types` page (or click on "View All Types" in the sidebar)

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/a552fb51-6bb9-47b8-bc85-fa785d30e818


apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const latestLinkEntityTypes = useMemo(
+    () =>
+      latestEntityTypes?.filter((entityType) =>
+        isLinkEntityType(entityType.schema),
+      ),
+    [latestEntityTypes],
+  );
+
+  const { queryPropertyTypes } = useBlockProtocolQueryPropertyTypes();
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:82]: There's a `useLatestPropertyTypesContextValue` hook you could possibly replace this with, as it does a similar thing.

It might be even better to move the `<LatestPropertyTypesContextProvider>` up from the type page to surround the app, so that we're sharing the same set of property types rather than refetching them on different pages

apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
```diff
@@ -0,0 +1,229 @@
...
+            ...latestDataTypes,
+          ]
+        : undefined,
+    [
+      latestNonLinkEntityTypes,
+      latestLinkEntityTypes,
+      latestPropertyTypes,
+      latestDataTypes,
+    ],
+  );
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:137]: Nit: this'd be a bit shorter – are there behaviour changes it causes that are undesirable?

Saves the checks for `allTypes` being defined further down also
```suggestion
  const allTypes = useMemo(
    () => [
      ...(latestNonLinkEntityTypes ?? []),
      ...(latestLinkEntityTypes ?? []),
      ...(latestPropertyTypes ?? []),
      ...(latestDataTypes ?? []),
    ],
    [
      latestNonLinkEntityTypes,
      latestLinkEntityTypes,
      latestPropertyTypes,
      latestDataTypes,
    ],
  );
```

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
```diff
@@ -0,0 +1,83 @@
...
+      sx={{
+        minHeight: 0,
+        overflow: "visible",
+        alignItems: "flex-end",
+        flex: 1,
+        [`.${tabsClasses.scroller}`]: {
+          overflow: "visible !important",
+        },
+      }}
+    >
```
CiaranMn, line 52: This looks identical to the equivalent component in `entity-type-tabs.tsx`, can we share it also? e.g. some kind of `TabsContainer`. Ideally all these tabs would look identical and only be defined in one place

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
```diff
@@ -0,0 +1,83 @@
...
+      {tabIds.map((tabId) => (
+        <TabLink
+          key={tabId}
+          value={tabId}
+          href={tabId === "all" ? "/types" : `/types/${tabId}`}
+          label={tabTitles[tabId]}
+          count={numberOfTypesByTab[tabId]}
+          loading={numberOfTypesByTab[tabId] === undefined}
```
CiaranMn, line 60: This doesn't work to show the loading indicator because there's a check for `typeof count === "number"` in `TabLink` before showing the loading indicator. 

It looks like we could remove that check and the `?? 0` from `count={entities?.length ?? 0}` on the entity type entities tab to get the desired behaviour (might be wrong)

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
```diff
@@ -0,0 +1,83 @@
+import { Tabs, tabsClasses, typographyClasses } from "@mui/material";
```
CiaranMn, line 1: Strictly speaking this file and `types-table.tsx` (and therefore the `types-table` folder) should be in the `[[...type-kind]].page` folder, because they are its private dependencies

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/search-input.tsx
```diff
@@ -0,0 +1,98 @@
...
+          color: palette.gray[50],
+          svg: { fontSize: 14 },
+        })}
+        onClick={() => showSearchInput()}
+      >
+        <MagnifyingGlassLightIcon />
+      </IconButton>
+    </Tooltip>
+    <Fade in={searchVisible}>
+      <TextField
```
CiaranMn, line 39: Can we add autofocus of this? It wasn't autofocusing before because it was omnipresent, but now that it appears on button click I think it should autofocus

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/search-input.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { MagnifyingGlassLightIcon } from "../../../icons/magnifying-glass-light";
+
+type SearchInputProps = {
+  searchVisible: boolean;
+  showSearchInput: () => void;
+  hideSearchInput: () => void;
+  onChangeText: (text: string) => void;
+};
+
+export const SearchInput: FunctionComponent<SearchInputProps> = ({
```
CiaranMn, line 19: UX question to consider (not for this PR) for @nonparibus – now that we have the 'view all types', and this search input appears by clicking a button next to 'view all types', it's unexpected that searching in the input just searches the list of account types above. We should think about what to do with this search bar, whether that's to launch into the table with a filter, or do a dynamic query to the db and show the results above, or something else

In this row I can click 'view all types', or 'search types'
Screenshot 2023-08-15 at 17 40 10

'search types' doesn't search all types, but just the list of types in my account above
Screenshot 2023-08-15 at 17 40 19

libs/@hashintel/design-system/src/button.tsx
```diff
@@ -4,6 +4,7 @@ import {
   ButtonProps as MuiButtonProps,
   useTheme,
 } from "@mui/material";
+import Link from "next/link";
```
CiaranMn, line 7: We shouldn't do this because it means the design system is making assumptions about the application it is used in – we have `shared/ui/button.tsx` in the frontend app which wraps this and should be modified as necessary instead, and also has additional behaviour like when to openInNew that this doesn't. 

We could lint against this using the following rule added to `.eslintrc.cjs`, although there would probably be failures unrelated to your PR to fix so you may not want to add it
```js
{
    "no-restricted-imports": [
      "error",
      {
        paths: [
          {
            name: "@hashintel/design-system",
            importNames: [
              "Button",
              "IconButton",
              "Link",
              "MenuItem",
            ],
            message: "Please use the custom wrapper component in src/shared/ui component instead"
          },
       ]
    }
  ]
}
```

CiaranMn (commented):
Thanks @benwerner01, this is looking good and working well – a few comments, mainly about code re-use

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/search-input.tsx
nonparibus: Will respond separately. Tracking this UX Q as H-418.

libs/@hashintel/design-system/src/button.tsx
nonparibus: If we don't add the rule above in this PR, let's add it in a small follow-up.

apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:137]: I had this previously, but I wanted `allTypes` to be `undefined` when it's items have not yet been fetched.

apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:82]: Hmm yes I can move this up so I can make use of the context. Do you know why we return the property types in an object rather than an array by chance?

Comment (CiaranMn) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:82]: The entity editor asks for them that way, I believe it was done like this to make it easier to find relevant property types easily

apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx:137]: Fair enough, thanks!

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list.tsx
```diff
@@ -196,74 +158,61 @@ export const AccountEntityTypeList: FunctionComponent<
...
+                    }}
+                  />
                 </Typography>
-              </Link> */}
-
+              </Link>
               <SearchInput
                 searchVisible={searchVisible}
                 showSearchInput={() => setSearchVisible(true)}
                 hideSearchInput={() => setSearchVisible(false)}
```
CiaranMn, line 211: Appreciate this is pre-existing but the button is marked 'Clear search' and currently doesn't
```suggestion
                hideSearchInput={() => {
                  setSearchQuery("");
                  setSearchVisible(false);
                }}
```

apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx]: Alright I suppose it doesn't make much difference, may have been for a slight performance gain when looking up property types by versioned URL. Have made it a global context, and re-used the value in https://github.com/hashintel/hash/pull/2908/commits/e638dee00ac1ae9e00db280fc204740fd6d039b2

Comment (CiaranMn) [reply apps/hash-frontend/src/pages/types/[[...type-kind]].page.tsx]: Cool, looks good

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
benwerner01: Yeah thanks for reminding me to make this a re-usable component, have done so in https://github.com/hashintel/hash/pull/2908/commits/f732973c3c3369daaaa65a5b8bf14976d79e34ce and also moved the style overrides for `Tabs` and `Tab` into the theme rather than defining them inline.

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
```diff
@@ -0,0 +1,83 @@
...
+                [`.${typographyClasses.root}`]: {
+                  color: `${
+                    currentTab === tabId
+                      ? palette.primary.main
+                      : palette.blue[60]
+                  } !important`,
+                },
+              },
+            }),
+          ]}
```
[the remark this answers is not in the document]
benwerner01: This is to prevent the typography from changing to a lighter blue on hover. 

https://github.com/hashintel/hash/assets/42802102/83bbe691-fa32-4357-bda5-022747e6dcda

I've added this to the theme in https://github.com/hashintel/hash/pull/2908/commits/aea05d3b2fcd4c8a4a0017e4d3801884171fc2a6 to make it re-usable and so we can remove the `!important`.

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
benwerner01: Ah thanks for catching, thought the easiest fix would to just add the `loading` boolean to the conditional. Have done so in https://github.com/hashintel/hash/pull/2908/commits/dd6059a9bc7ce8833451b21c3ab68c872611855f

apps/hash-frontend/src/pages/types/types-page-tabs.tsx
benwerner01: Ah yes, fair enough. If I'm understanding right this also applies to the other components defined in that directory, have moved them in https://github.com/hashintel/hash/pull/2908/commits/116df25ed6a56f457e8c53368ae844dffcf8de9b

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/search-input.tsx
benwerner01: Good idea, have addressed in https://github.com/hashintel/hash/pull/2908/commits/d99e2819e61f3c6cf0c5ed09461ef4f5b6068cd2

libs/@hashintel/design-system/src/button.tsx
benwerner01: Thanks Ciaran, that makes sense - forgot we had that local `Button`. Have removed the `next` dependency from the design system, added the rule to the `.eslintrc.cjs` (note that we currently don't have a custom `IconButton`), and have fixed the lint errors in https://github.com/hashintel/hash/pull/2908/commits/7d83b69d778a5f67ac1f016fd0f5da7f3344b43c

PR #2909, opened 2023-08-15
H-355: Implement inheritance in the type editor

CiaranMn:

Implements the UI for entity type inheritance, allowing users to set (and remove) one or more types which another entity type 'extends', and displaying the inherited properties and links in the relevant tables.

[Design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=8377-289907&mode=design&t=cDUo38X3kTkfW0Q3-4) (internal link)

## 🔍 What does this change?

- Add `inheritsFrom` as a depth argument to entity type queries in the GraphQL API and relevant queries in the frontend
- Adds inheritance management to `@hashintel/design-system` via a new 'Extends' section at the top
- Adds a hook in the type editor to get inherited properties/links
- Adds new 'inherited property/link rows' to the properties and links tables
- **(most of the diff)** Extracts various components so that they can be used for displaying information only in the inherited rows, and aren't relying on retrieving data from the form context, since inherited properties and links are not part of the editable form data
    - this is the bulk of the diff – any deleted files and big chunks of components are just moved elsewhere, and I will comment where this is so
- Updates `use-entity-type-value` in the frontend so that it also gets properties from the inheritance chain
- Adds a 'secondary quiet' button variant to the design system (this is the 'Add type +' in the extends section)

This PR:

The changes in this PR:

   - we should document inheritance 

The changes in this PR:

## ⚠️ Known issues

- Inherited properites / links will not appear in the entity editor

## 🐾 Next steps

- Handle inheritance in the entity editor - H-356
- Add an 'extend this type' button to existing type pages to create a new type which extends another - H412

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1.  Visit a type page, try adding a parent
2. Try adding a parent to the parent
3. Try releasing a new version of a parent
4. Check that inherited properties/links display okay, aren't editable

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/95fa27d6-3b3c-42c9-a7e3-148791548fea

N.B. at one point towards the end of the video two 'Created At' appears – I fixed this in https://github.com/hashintel/hash/pull/2909/commits/ee7020e38a91310449abc63fbf197ddaf1c701e2


apps/hash-frontend/src/graphql/queries/ontology/entity-type.queries.ts
```diff
@@ -46,27 +50,6 @@ export const queryEntityTypesQuery = gql`
   ${subgraphFieldsFragment}
 `;
 
-export const getEntityTypeRootedSubgraphQuery = gql`
```
CiaranMn, line 49: This is equivalent to `getEntityTypeQuery` above, and unused

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/types-section/type-card.tsx
```diff
@@ -1,85 +0,0 @@
-import { faArrowsRotate, faAsterisk } from "@fortawesome/free-solid-svg-icons";
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/types-section/type-card.tsx:1]: Moved to `design-system`

libs/@hashintel/design-system/src/type-card.tsx
```diff
@@ -0,0 +1,142 @@
...
+  Typography,
+} from "@mui/material";
+import { ElementType, FunctionComponent, useState } from "react";
+
+import { FontAwesomeIcon } from "./fontawesome-icon";
+import { IconButton } from "./icon-button";
+import { WhiteCard } from "./white-card";
+
+interface TypeCardProps {
+  onDelete?: () => void;
```
CiaranMn, line 17: This is mostly just moved from the type editor, but this `onDelete` is new (it's the 'x' that appears if you hover over an inherited type)

libs/@hashintel/design-system/src/white-card.tsx
```diff
@@ -12,13 +12,17 @@ import { ElementType } from "react";
 
 export const WhiteCard = ({
   onClick,
+  onMouseEnter,
+  onMouseLeave,
```
CiaranMn, line 16: Needed to handle the new `onDelete` in the file above

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card.tsx
```diff
@@ -181,7 +183,7 @@ const LinkTypeRow = ({
           </EntityTypeTableTitleCellText>
         </TableCell>
         <TableCell sx={{ py: "0 !important" }}>
-          <LinkEntityTypeSelector linkIndex={linkIndex} />
+          <DestinationEntityTypeSelector linkIndex={linkIndex} />
```
CiaranMn, line 186: Renamed a few components for clarity – there are two 'link entity type selectors' in the link table, the one where you add a new link to the entity, and the one where you add a new link as a destination for a link type. this is the latter

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card/destination-entity-type-selector.tsx
```diff
@@ -1,129 +1,25 @@
...
 
 import { useEntityTypesOptions } from "../../shared/entity-types-options-context";
 import { EntityTypeEditorFormData } from "../../shared/form-types";
 import { useIsReadonly } from "../../shared/read-only-context";
-import { Link } from "../shared/link";
 import { useFilterTypeOptions } from "../shared/use-filter-type-options";
-import { useTypeVersions } from "../shared/use-type-versions";
-import { VersionUpgradeIndicator } from "../shared/version-upgrade-indicator";
-
-const TypeChipLabel = ({
```
CiaranMn, line 31: Only moved to its own file

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card/destination-entity-type-selector.tsx
```diff
@@ -1,129 +1,25 @@
...
-              })}
-            />
-          </IconButton>
-        </Link>
-      )}
-    </Stack>
-  );
-};
-
-const ChosenEntityType = ({
```
CiaranMn, line 91: Renamed to `DestinationEntityType` and moved to its own file (no changes)

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card/destination-entity-type.tsx
```diff
@@ -0,0 +1,48 @@
+import { VersionedUrl } from "@blockprotocol/type-system";
+import { EntityType } from "@blockprotocol/type-system/slim";
+import { Chip } from "@hashintel/design-system";
+
+import { useEntityTypesOptions } from "../../shared/entity-types-options-context";
+import { useTypeVersions } from "../shared/use-type-versions";
+import { TypeChipLabel } from "./type-chip-label";
+
+export const DestinationEntityType = ({
```
CiaranMn, line 9: A straight extraction of `ChosenEntityType` from the file above

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card/destination-type-container.tsx
```diff
@@ -0,0 +1,30 @@
+import { Stack, StackProps } from "@mui/material";
+
+export const DestinationTypeContainer = ({
```
CiaranMn, line 3: No changes, straight extraction of a `Box` with styling

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card/type-chip-label.tsx
```diff
@@ -0,0 +1,72 @@
...
+} from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@hashintel/design-system";
+import { Box, IconButton, Stack } from "@mui/material";
+import { ReactNode } from "react";
+
+import { useIsReadonly } from "../../shared/read-only-context";
+import { Link } from "../shared/link";
+import { VersionUpgradeIndicator } from "../shared/version-upgrade-indicator";
+
+export const TypeChipLabel = ({
```
CiaranMn, line 14: No changes, straight extraction

libs/@hashintel/type-editor/src/entity-type-editor/property-list-card.tsx
```diff
@@ -40,286 +29,37 @@ import { EntityTypeEditorFormData } from "../shared/form-types";
...
-          </Table>
-        </Collapse>
-      </TableCell>
-    </TableRow>
-  );
-};
-
 export const REQUIRED_CELL_WIDTH = 100;
 
-const PropertyRow = ({
```
CiaranMn, line 122: This has been moved to its own file, `property-row.tsx`, no changes

libs/@hashintel/type-editor/src/entity-type-editor/property-list-card/inherited-property-row.tsx
```diff
@@ -0,0 +1,29 @@
+import { usePropertyTypesOptions } from "../../shared/property-types-options-context";
```
CiaranMn, line 1: This file is mostly a straight extraction from `property-list-card`, with the exception of `InheritanceData` added as props, which affects the readonly error message shown

libs/@hashintel/type-editor/src/entity-type-editor/property-list-card/property-row.tsx
```diff
@@ -0,0 +1,260 @@
+import { PropertyType, VersionedUrl } from "@blockprotocol/type-system/slim";
```
CiaranMn, line 1: This file is mostly a straight extraction from property-list-card, with the exception of InheritanceData added as props, which affects the readonly tooltip message shown

libs/@hashintel/type-editor/src/entity-type-editor/shared/multiple-values-cell.tsx
```diff
@@ -54,7 +55,60 @@ const Frozen: FunctionComponent<
   return <>{frozenChildren}</>;
 };
 
-const MultipleValuesCellSummary = ({
+export const MultipleValuesControlContainer = ({
```
CiaranMn, line 58: This just moves the deleted `Box`s below into a sharable wrapping component

benwerner01 (commented):
Thanks Ciaran, this is working for me locally and the changes look good to me. I made two observations whilst testing:

1. The constraints of an inherited property aren't displayed (i.e. when the parent has a required property, the required checkbox is not checked in the child for the inherited property). Was this intentional?
2. It's possible to create a cycle (i.e. the child inherits from the parent, and the parent inherits from the child). Granted you never end up in a state where the latest version of both the child and the parent inherit from another, but I'm wondering whether this is something we are thinking of preventing on the BE?

benwerner01:
> This is not the case for me – I can see required/array checks on inherited properties (directly and indirectly inherited) – could you provide an example please? Is it possible that you made a property required on a new version of a parent but didn't upgrade the child's version of the parent?

Ah apologies I just realised this is only occurring for me when I create a cycle. But here you can see `Test 2 v12` inheriting from `Test 3 v15`, not indicating that the property is multiple, or required:
image
And here is `Test 3 v15`, which has the property set to multiple and required:
image

But since this is a state we want to prevent users from getting into, I don't think this is something worth worrying about.

CiaranMn:
> > This is not the case for me – I can see required/array checks on inherited properties (directly and indirectly inherited) – could you provide an example please? Is it possible that you made a property required on a new version of a parent but didn't upgrade the child's version of the parent?
> 
> Ah apologies I just realised this is only occurring for me when I create a cycle. But here you can see `Test 2 v12` inheriting from `Test 3 v15`, not indicating that the property is multiple, or required: But since this is a state we want to prevent users from getting into, I don't think this is something worth worrying about.

@benwerner01 understood, thank you, and agreed re. fixing this by preventing cycles rather than dealing with the consequences of them

libs/@hashintel/type-editor/src/entity-type-editor/inheritance-row.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const directParentEntityTypeIds = useWatch({
+    control,
+    name: "allOf",
+  });
+
+  const { entityTypes } = useEntityTypesOptions();
+  const entityTypesArray = Object.values(entityTypes);
+  const directParents = entityTypesArray.filter((type) =>
+    directParentEntityTypeIds.includes(type.$id),
+  );
```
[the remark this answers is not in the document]
CiaranMn: Done in https://github.com/hashintel/hash/commit/0b4c5e7b8645eeca4af3dbcec724d2e7e020a6e4

libs/@hashintel/type-editor/src/entity-type-editor/shared/inherited-icon.tsx
```diff
@@ -0,0 +1,24 @@
+import { SvgIcon, SvgIconProps } from "@mui/material";
+import { FunctionComponent } from "react";
+
+// FontAwesome 'arrow-turn-down-right' / ArrowTurnDownRight
+export const InheritedIcon: FunctionComponent<SvgIconProps> = ({
```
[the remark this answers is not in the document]
CiaranMn: Done in https://github.com/hashintel/hash/pull/2909/commits/3ee1907d38b4f388c031b2bc109cd98751ff8374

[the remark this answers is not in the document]
CiaranMn: I forgot `Icon`, I added it in https://github.com/hashintel/hash/pull/2909/commits/3c71c4c54238729c77113edc0374c428be09c023

libs/@hashintel/design-system/src/type-card.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const { newVersion, onUpdateVersion } = newVersionConfig ?? {};
+
+  const [hovered, setHovered] = useState(false);
+
+  return (
+    <WhiteCard
+      href={url}
+      LinkComponent={LinkComponent}
+      onMouseEnter={() => (onDelete ? setHovered(true) : null)}
+      onMouseLeave={() => (onDelete ? setHovered(false) : null)}
```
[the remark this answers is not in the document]
CiaranMn: The only issue is that it causes React to rerender the component pointlessly. This won't cause a DOM repaint since nothing will change in the component (the browser is repainting it though since there's a hover effect) - but since we don't need it to happen at the moment I think it's fine to leave as-is

libs/@hashintel/design-system/src/selector-autocomplete.tsx
```diff
@@ -28,7 +28,7 @@ export const TYPE_SELECTOR_HEIGHT = 57;
 export type TypeListSelectorDropdownProps = {
   query: string;
   createButtonProps: Omit<ButtonProps, "children" | "variant" | "size"> | null;
-  variant: "entityType" | "propertyType" | "entity" | "linkType";
+  variant: "entity type" | "property type" | "entity" | "link type";
```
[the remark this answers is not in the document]
CiaranMn: I used it like this because these are only used to populate label values. I think it's fine to use kebab case things elsewhere, this is pretty localised

libs/@hashintel/type-editor/src/entity-type-editor/link-list-card.tsx
CiaranMn: Honestly I think this is more immediately understandable than 'right entity type', and until we introduce bidirectional links it's accurate

benwerner01 (dismissed):
Thanks Ciaran 👌

benwerner01 (dismissed):

PR #2910, opened 2023-08-15
H-413: Use inheritance for query parameters

TimDiekmann:

As a sensible default we chose to take into account inheritance when querying data from the Graph. This PR adjusts the structural queries to take inheritance into account. In the query parameter `Some(depth)` specifies the inheritance depths, which is explored, `0` means no inheritance is taken into account. `None` means there is no limit on the depth.

## 🚫 Blocked by

- #2894 

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -407,14 +413,17 @@ impl fmt::Display for EntityTypeQueryPath<'_> {
             Self::PropertyTypeEdge {
                 edge_kind: OntologyEdgeKind::ConstrainsPropertiesOn,
                 path,
+                ..
```
indietyp, line 416: I would prefer a wildcard here instead (so `inheritance_depth: _`) while virtually the same thing, doing this instead of the rest pattern means that this will error out when a new field is added aiding in Compiler Driven Development ™️

apps/hash-graph/lib/graph/src/knowledge/entity/query.rs
```diff
@@ -300,6 +302,7 @@ impl fmt::Display for EntityQueryPath<'_> {
             Self::EntityTypeEdge {
                 edge_kind: SharedEdgeKind::IsOfType,
                 path,
+                ..
```
indietyp, line 305: same as other comment

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -407,14 +413,17 @@ impl fmt::Display for EntityTypeQueryPath<'_> {
...
             } => write!(fmt, "properties.{path}"),
             #[expect(
                 clippy::use_debug,
                 reason = "We don't have a `Display` impl for `OntologyEdgeKind` and this should \
                           (a) never happen and (b) be easy to debug if it does happen. In the \
                           future, this will become a compile-time check"
             )]
-            Self::PropertyTypeEdge { edge_kind, path } => write!(fmt, "<{edge_kind:?}>.{path}"),
+            Self::PropertyTypeEdge {
+                edge_kind, path, ..
```
indietyp, line 425: same as other comment

apps/hash-graph/lib/graph/src/store/postgres/query/condition.rs
```diff
@@ -27,6 +31,7 @@ pub enum EqualityOperator {
 }
 
 impl Transpile for Condition {
+    #[expect(clippy::too_many_lines)]
```
indietyp, line 34: Maybe a sign to refactor? I could imagine a function like `fn transpile_binary_op(lhs: impl Transpile, op: &'static str, rhs: impl Transpile)` would remove quite a lot of repetitive code.

apps/hash-graph/tests/friendship.http
```diff
@@ -1275,6 +1275,77 @@ Accept: application/json
...
+          "type",
+          "versionedUrl"
+        ]
+      },
+      {
+        "parameter": "https://blockprotocol.org/@blockprotocol/types/entity-type/link/v/1"
+      }
+    ]
+  },
+  "graphResolveDepths": {
```
indietyp, line 1296: any plans to expose the inheritance depth parameter in these arguments?

indietyp (changes requested):
Hope its okay that I took the liberty of looking through the code. Overall looks good to me (although I do **not yet** have intricate knowledge about the graph database, so take with a grain of salt), but have some comments.

apps/hash-graph/lib/graph/src/store/postgres/query/condition.rs
```diff
@@ -12,6 +12,10 @@ pub enum Condition {
     Not(Box<Self>),
     Equal(Option<Expression>, Option<Expression>),
     NotEqual(Option<Expression>, Option<Expression>),
+    Less(Expression, Expression),
+    LessOrEqual(Expression, Expression),
+    Greater(Expression, Expression),
+    GreaterOrEqual(Expression, Expression),
```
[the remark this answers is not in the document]
TimDiekmann: You cannot compare `null <= 10` as an example, but you could check for equality. `null` is not `PartialOrd`.

apps/hash-graph/lib/graph/src/store/postgres/query/condition.rs
TimDiekmann: I'm about to refactor this function at some point but it's currently pretty straightforward to read.

apps/hash-graph/tests/friendship.http
indietyp: Keeping this open to think about this, but is a non-blocker.

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
TimDiekmann: Typically I would agree but the edges on the query paths are not really well designed in the code. We have far too many wildcards around these variants, the `..` is not really a deal-breaker. My goal is to remove the generic edge variants with the actual edge names so it's possible to maybe enable `wildcard_enum_match_arm`. Sadly, there is not a lint such as `match_field_pattern` lint.
Adding all variants to the edges would be pretty verbose as some edges have up to four parameters and only adding the wildcard in a few locations seems inconsistent.

TimDiekmann:
I tried to apply the suggestion of the previous PR and decided to not go with the suggestion as the diff would be 285 lines. Instead, I added a comment to the enumeration itself. Similary, I also won't destructure all field in the edges table as they are subject to change and does not add anything currently besides a big chunk of lines which effectively makes the code less readable.

indietyp (dismissed):
With previous comment and offline clarification: LGTM

indietyp (approved):

PR #2912, opened 2023-08-15
GEN-53: Upgrade `error-stack` toolchain and use `Error::provide`

TimDiekmann:

The Error::provide API has changed. To avoid conflicts with the recent Rust toolchain and maintain compatibility we need to update the toolchain. This reverts #2906 (GEN-52).

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -205,6 +205,11 @@
       "matchPackagePatterns": ["^tracing[-_]?"],
       "excludePackageNames": ["tracing-opentelemetry"],
       "groupName": "`tracing` crates"
+    },
+    {
+      "matchManagers": ["cargo"],
+      "matchFileNames": ["libs/error-stack/Cargo.toml"],
+      "excludeDepNames": ["anyhow"]
     }
   ],
   "regexManagers": [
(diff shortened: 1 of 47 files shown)


indietyp (approved):
Even tho our bot overlords have blessed this PR here you go. Overall I dislike that the provide API has been removed, but o well 🤷 it's a nightly API after all.

.github/renovate.json
```diff
@@ -205,6 +205,11 @@
       "matchPackagePatterns": ["^tracing[-_]?"],
       "excludePackageNames": ["tracing-opentelemetry"],
       "groupName": "`tracing` crates"
+    },
```
indietyp, line 208: Specific reason why anyhow is being singled out into it's own group?

indietyp: (Non blocking just curious)

TimDiekmann: Renovate tries updating `anyhow` all the time but we only have a patch version specified here for nightly compatibility. After all this is a library and `anyhow` is an optional feature and we should not put stricter requirements on it than needed.

indietyp (approved):

PR #2914, opened 2023-08-16
H-423: Make BE tests and PW tests depend on the integration worker

TimDiekmann:

The external services require the integration worker. Until now it never happened (to my knowledge) that the BE tests or PW tests started before the integration worker completed but it just happened for the first time and to avoid issues in the future, it should be made sure that the tests are waiting for the worker to be built.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,7 +265,13 @@ jobs:
 
   backend-integration-tests:
     name: Backend integration tests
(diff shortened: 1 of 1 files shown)


indietyp (approved):
@TimDiekmann is cleaning up dependencies today 🧹, good stuff 👍

PR #2915, opened 2023-08-16
H-426: Extend structural queries to allow specifying inheritance depth

TimDiekmann:

Currently, by default, the maximum inheritance depth is used for query parameters. To allow specifying it a parameter may be passed to the query token, e.g. instead of `["type", "versionedUrl"]` it can be specified as `["type(inheritanceDepth = 5)", "versionedUrl]`.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/ontology/entity_type.rs
+++ b/apps/hash-graph/lib/graph/src/ontology/entity_type.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{fmt, str::FromStr};
 
 use serde::{
     de::{self, Deserializer, SeqAccess, Visitor},
@@ -9,7 +9,9 @@ use utoipa::ToSchema;
 use crate::{
     knowledge::EntityQueryPath,
     ontology::{property_type::PropertyTypeQueryPathVisitor, PropertyTypeQueryPath, Selector},
-    store::query::{JsonPath, OntologyQueryPath, ParameterType, PathToken, QueryPath},
+    store::query::{
+        parse_query_token, JsonPath, OntologyQueryPath, ParameterType, PathToken, QueryPath,
+    },
     subgraph::edges::{EdgeDirection, OntologyEdgeKind, SharedEdgeKind},
 };
 
@@ -226,6 +228,28 @@ pub enum EntityTypeQueryPath<'p> {
     /// });
     /// # Ok::<(), serde_json::Error>(())
     /// ```
+    ///
+    /// ### Specifying the inheritance depth
+    ///
+    /// By passing `inheritanceDepths` as paramteer it's possible to limit the searched depth:
+    ///
+    /// ```rust
+    /// # use serde::Deserialize;
+    /// # use serde_json::json;
+    /// # use graph::ontology::{EntityTypeQueryPath, PropertyTypeQueryPath};
+    /// # use graph::subgraph::edges::OntologyEdgeKind;
+    /// let path = EntityTypeQueryPath::deserialize(json!([
+    ///     "properties(inheritanceDepth=10)",
+    ///     "*",
+    ///     "baseUrl"
+    /// ]))?;
+    /// assert_eq!(path, EntityTypeQueryPath::PropertyTypeEdge {
+    ///     edge_kind: OntologyEdgeKind::ConstrainsPropertiesOn,
+    ///     path: PropertyTypeQueryPath::BaseUrl,
+    ///     inheritance_depth: Some(10),
+    /// });
+    /// # Ok::<(), serde_json::Error>(())
+    /// ```
     PropertyTypeEdge {
         edge_kind: OntologyEdgeKind,
         path: PropertyTypeQueryPath<'p>,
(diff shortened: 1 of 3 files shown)


apps/hash-graph/lib/graph/src/store/query.rs
```diff
@@ -13,6 +18,39 @@ pub trait QueryPath {
...
+    token: &'de str,
+) -> Result<(T, HashMap<&str, &str>), E> {
+    if let Some((token, parameters)) = token.split_once('(') {
+        let parameters = parameters
+            .strip_suffix(')')
+            .ok_or_else(|| E::custom("missing closing parenthesis"))?;
+
+        let parameters = parameters
+            .split(',')
+            .filter(|parameter| !parameter.trim().is_empty())
```
TimDiekmann, line 39: Required for `"".split(',')` resulting in `[""]`. That `",,"` will be allowed here as well is a side effect but also not really an issue.

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -226,6 +228,28 @@ pub enum EntityTypeQueryPath<'p> {
     /// });
     /// # Ok::<(), serde_json::Error>(())
     /// ```
+    ///
+    /// ### Specifying the inheritance depth
+    ///
+    /// By passing `inheritanceDepths` as paramteer it's possible to limit the searched depth:
```
indietyp, line 234:
```suggestion
    /// By passing `inheritanceDepth` as a parameter it's possible to limit the searched depth:
```

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -272,6 +296,30 @@ pub enum EntityTypeQueryPath<'p> {
     /// # Ok::<(), serde_json::Error>(())
     /// ```
     ///
+    /// ### Specifying the inheritance depth
+    ///
+    /// By passing `inheritanceDepths` as paramteer it's possible to limit the searched depth:
```
indietyp, line 301:
```suggestion
    /// By passing `inheritanceDepth` as a parameter it's possible to limit the searched depth:
```

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -298,6 +346,26 @@ pub enum EntityTypeQueryPath<'p> {
...
```
indietyp, line 351:
```suggestion
    /// By passing `inheritanceDepth` as a parameter it's possible to limit the searched depth:
```

apps/hash-graph/lib/graph/src/store/query.rs
```diff
@@ -13,6 +18,39 @@ pub trait QueryPath {
...
 
+/// Parses a query token of the form `token(key=value)`.
+///
+/// Whitespaces are ignored and multiple parameters are supported.
+///
+/// # Errors
+///
+/// - If the token is not of the form `token`, `token()`, or `token(key=value)`
+/// - If `token` can not be deserialized into `T`
+pub fn parse_query_token<'de, T: Deserialize<'de>, E: de::Error>(
```
indietyp, line 29: Currently works, nothing major, but I see some caveats I'd like to see documented:

A) quoted strings do not currently work (what if - in the future - I have whitespace or `,` in my value?)
B) documentation does not specify that multiple key-value pairs are supported.

I would usually go with something like `nom`, `winnow`, `chumsky` as a parser combinator, to be future proof and have more readable code, but I understand that adding a new dependency for this might not be warranted.

apps/hash-graph/lib/graph/src/store/query.rs
```diff
@@ -13,6 +18,39 @@ pub trait QueryPath {
...
+                    .ok_or_else(|| E::custom("missing parameter value, expected `key=value`"))?;
+                Ok((key.trim(), value.trim()))
+            })
+            .collect::<Result<_, _>>()?;
+
+        T::deserialize(token.into_deserializer()).map(|token| (token, parameters))
+    } else {
+        T::deserialize(token.into_deserializer()).map(|token| (token, HashMap::new()))
+    }
+}
```
indietyp, line 52:
```suggestion
    let Some((token, parameters)) = token.split_once('(') else {
          return T::deserialize(token.into_deserializer()).map(|token| (token, HashMap::new()));
    };

    let parameters = parameters
        .strip_suffix(')')
        .ok_or_else(|| E::custom("missing closing parenthesis"))?;

    let parameters = parameters
        .split(',')
        .filter(|parameter| !parameter.trim().is_empty())
        .map(|parameter| {
            let (key, value) = parameter
                .split_once('=')
                .ok_or_else(|| E::custom("missing parameter value, expected `key=value`"))?;
            Ok((key.trim(), value.trim()))
        })
        .collect::<Result<_, _>>()?;

    T::deserialize(token.into_deserializer()).map(|token| (token, parameters))
}
```

indietyp (changes requested):

apps/hash-graph/lib/graph/src/store/query.rs
TimDiekmann: I would also go with an external parser for proper parsing but as we only split by `,` to detect well-defined strings and numbers this is by far the easiest solution.

apps/hash-graph/lib/graph/src/store/query.rs
TimDiekmann: Still not used to the `let ... else` syntax. Maybe because I always forget that we finally have a formatting for it 😂

apps/hash-graph/lib/graph/src/store/query.rs
TimDiekmann: I also wouldn't pass `Deserialize` here but a `FromStr` instead.

indietyp (approved):

PR #2916, opened 2023-08-16
H-432: Support `children` query path for entity types

TimDiekmann:

It's useful to be able to search for parents given an entity type.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- New tests have been added


apps/hash-graph/lib/graph/src/ontology/entity_type.rs
```diff
@@ -296,6 +296,23 @@ pub enum EntityTypeQueryPath<'p> {
     /// # Ok::<(), serde_json::Error>(())
     /// ```
     ///
+    /// It's also possible to create a query path for the reversed direction:
```
indietyp, line 299: How does this interact with external types? I assume that because there isn't a central repository, they won't show up until loaded in explicitly?

(non-blocking, just clarification)

libs/@local/hash-graph-client/python/graph_client/models.py
```diff
@@ -105,6 +105,7 @@ class EntityTypeQueryToken(Enum):
     label_property = "labelProperty"
     links = "links"
     inherits_from = "inheritsFrom"
+    children = "children"
```
indietyp, line 108: We'll also need to update the filter; I'll do that in a follow-up PR. I wrote the code for it but forgot to add it to the codegen pipeline 🤦

indietyp (approved):

apps/hash-graph/lib/graph/src/ontology/entity_type.rs
TimDiekmann: Yes, you cannot know if a type externally referencing this type, so it's not possible to read that.

PR #2924, opened 2023-08-17
H-67: Add ability to view archived pages

benwerner01:

This PR adds a "View all pages" page at `/pages`, which the list of page entities a user can see. It adds a page at `/entities` which either displays all entities a user can see, or all entities of an entity type version or entity type if the optional `entityTypeIdOrBaseUrl` query param is set.

Note that `/pages` is essentially a mask of `/entities?entityTypeIdOrBaseUrl=http://localhost:3000/@system-user/types/entity-type/page/v/1`.

## 🔍 What does this change?

- makes the entities table re-usable between the `/entities` page and the entities tab on the entity type page via a `EntitiesTable` component

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Currently the list of entities a user can see is all the entities in the instance of HASH, as authorization has not yet been implemented.
- Currently the "Last Edited" column for page entities represents when the page entity was last edited, which does not account for edits to nested text/block entities in the page  ([internal ticket](https://linear.app/hash/issue/H-498))
- The `/entities` page is not discoverable via the UI, as we aren't linking to it from anywhere

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Checkout the branch and login
2. Go to the `/pages` route, or click on "View all pages" in the sidebar. Check the table lists page entities as expected, and use the filters to discover archived and external pages
3. Go to the `/entities` route (I don't think we are currently linking to this anywhere). Check the table lists all entities, and use the filters to discover archived/external entities. You can also try out the `entityTypeIdOrBaseUrl` query param to transform the page into a view of that type.
4. Go to to an entity type page, and click on the "Entities" tab. Check the table still works as previously, with the addition of filters, updated styling, and the "Last Edited"/"Last Edited By" columns.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/68f40ee1-24fd-48fd-be07-d87a270a5b5a


apps/hash-frontend/src/pages/entities.page.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+type ParsedQueryParams = {
+  entityTypeIdOrBaseUrl?: string;
+};
+
+const EntitiesPage: NextPageWithLayout = () => {
+  const router = useRouter();
+  const { activeWorkspace } = useContext(WorkspaceContext);
+
+  const { entityTypeId, entityTypeBaseUrl } = useMemo(() => {
```
CiaranMn, line 49: Why not have `entityTypeId` and `entityTypeBaseUrl` as separate query params? 

I think we only really need `entityTypeBaseUrl` at first (we could be looking for `Page`s by their base URL) but no harm in having `entityTypeId`

apps/hash-frontend/src/pages/entities.page.tsx
```diff
@@ -0,0 +1,220 @@
...
+      <NextSeo title={pageTitle} />
+      <TopContextBar
+        defaultCrumbIcon={null}
+        crumbs={[
+          {
+            title: pageTitle,
+            href: isViewAllPagesPage
+              ? "/pages"
+              : `/entities${
+                  entityTypeId ? `?entityTypeId=${entityTypeId}` : ""
```
CiaranMn, line 119: This won't work because `entityTypeId` is not a query param we're checking for at the moment

apps/hash-frontend/src/pages/entities.page.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  const entityTypeEntitiesValue = useEntityTypeEntitiesContextValue({
+    entityTypeBaseUrl,
+    entityTypeId,
+  });
+
+  const isViewAllPagesPage =
+    entityType?.schema.$id === types.entityType.page.entityTypeId;
+
+  const pageTitle = entityType ? `${entityType.schema.title}s` : "Entities";
```
CiaranMn, line 102: I'd suggest appending the version number if we've filtered to a specific version number by using a versioned instead of base URL as the filter, e.g. `Issue v1`

apps/hash-frontend/next.config.js
```diff
@@ -62,6 +62,24 @@ process.env.NEXT_PUBLIC_SENTRY_DSN = process.env.SENTRY_DSN;
...
+    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
+    : process.env.FRONTEND_URL ?? "http://localhost:3000");
+
+const systemUserShortname =
+  process.env.SYSTEM_USER_SHORTNAME ??
+  // you cannot access process.env in NextJS by variable, thus the repetition of keys in this section
+  process.env.NEXT_PUBLIC_SYSTEM_USER_SHORTNAME ??
+  "example-org";
+
+const pageEntityTypeId = `${frontendUrl}/@${systemUserShortname}/types/entity-type/page/v/1`;
```
CiaranMn, line 81: I have suggested elsewhere that if we filter to a specific version we could make that clear in the page title – in that case it's probably worth doing this because we're not actually filtering to a specific version of Page in order to exclude others, there just happens to be only one of them
```suggestion
const pageEntityTypeId = `${frontendUrl}/@${systemUserShortname}/types/entity-type/page/`;
```

CiaranMn (commented):
Thanks @benwerner01, page looks good and works well – only a couple of comments around handling the specific version vs base version query params

apps/hash-frontend/src/pages/entities.page.tsx
benwerner01: My thinking was since it shouldn't be possible to allow for both `entityTypeId` and `entityTypeBaseUrl` to be provided, that unifying them in a single query param would make more sense. But I'm happy to change this if you'd prefer it.

apps/hash-frontend/src/pages/entities.page.tsx
benwerner01: Thanks for catching, have addressed in https://github.com/hashintel/hash/pull/2924/commits/b6aaf595bf7bd1e7252272a4299a0f206ef23b6c

apps/hash-frontend/src/pages/entities.page.tsx
benwerner01: Good idea, have addressed in https://github.com/hashintel/hash/pull/2924/commits/6d69c61601d502942819ed3f70b639ffbfd0d9b7

apps/hash-frontend/src/pages/entities.page.tsx
CiaranMn: I don't think it's a big deal. Let's see how the filters evolve

CiaranMn (dismissed):

CiaranMn (approved):

PR #2936, opened 2023-08-18
H-351: Use `hash-graph-types` to create entities

indietyp:

Implement higher-level operations on the graph by using `hash-graph-types` on entities.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Before this gets ready for tests I need to first write an example to verify the code actually works.

There is also no truly convenient way to get all entities from the graph yet, this will need to be fixed before this is ready.

Next PR will introduce real tests.

### DIFF
--- a/libs/@local/hash-graph-sdk/python/graph_sdk/client/concurrent.py
+++ b/libs/@local/hash-graph-sdk/python/graph_sdk/client/concurrent.py
@@ -45,10 +45,12 @@ from pydantic_core._pydantic_core import Url
 from yarl import URL
 
 from graph_sdk.client._compat import recast
+from graph_sdk.entity import Entity
 from graph_sdk.options import Options
 from graph_sdk.query import BaseFilter
 
 T = TypeVar("T")
+U = TypeVar("U", bound=GraphEntityType)
 
 
 def assert_not_none(value: T | None) -> T:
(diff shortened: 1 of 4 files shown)


libs/@local/hash-graph-sdk/python/graph_sdk/client/concurrent.py
```diff
@@ -255,3 +262,60 @@ async def query_entities(self, query: BaseFilter, options: Options) -> Subgraph:
...
+        # by directly calling the magic method we can bypass this.
+        # Annoying? Yes. Necessary? Also yes.
+        properties_type: type[U] = type(properties)
+        entity_type = cast(type[Entity[U]], Entity.__class_getitem__(properties_type))
+
+        return entity_type.from_create(metadata, link, properties)
+
+    async def update_entity(
+        self,
+        entity: Entity[U],
```
indietyp, line 299: I am still a bit unsure about this one, we create the types dynamically, so I don't know if `U` will ever be of use, as you cannot statically refer to any entity type.

libs/@local/hash-graph-sdk/python/graph_sdk/entity.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class Link(BaseModel):
+    """Aggregated link information for an entity."""
+
+    left: EntityId | "Entity" | None = None
+    right: EntityId | "Entity" | None = None
+
+    left_to_right_order: LinkOrder | None = None
+    right_to_left_order: LinkOrder | None = None
```
TimDiekmann, line 31: This should not be added as this is going to be removed in the Graph as well.

indietyp: Thought so, thanks for the flag!

PR #2943, opened 2023-08-19
H-516: Split out `temporal-versioning` from `hash-graph` into crate

TimDiekmann:

The hash-graph contains two major subcrates: graph and type-fetcher. The graph crate really should be split up into multiple parts do improve DX and improve reusability.

## 🔍 What does this change?

- Moves temporal versioning related structures into separated workspace crate
- Split the `interval::tests` module into a `tests/` folder
- This does not change any logic, it's purely moving code.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Ideally we would move this to `libs/` but we currently cannot have multiple Rust workspaces depending on each other as otherwise the CI would not be able to determine the dependency order. This would result in dependants not being tested in CI if the dependency was changed.


apps/hash-graph/bench/Cargo.toml
```diff
@@ -13,6 +13,7 @@ autobenches = false
 [dev-dependencies]
 graph = { path = "../lib/graph" }
 graph-test-data = { path = "../tests/test_data" }
+temporal-versioning = { workspace = true }
```
indietyp, line 16: Of course, this is of personal taste, but I usually to `temporal-versioning.workspace = true` (seen in some other projects), to mimic `temporal-versioning = "0.1" or something similar.

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
```diff
@@ -78,7 +78,7 @@ pub async fn type_fetcher(args: TypeFetcherArgs) -> Result<(), GraphError> {
             channel.execute(server.serve())
         })
         .buffer_unordered(255)
-        .for_each(|_| async {})
+        .for_each(|()| async {})
```
indietyp, line 81: non blocking, just out of curiosity, which is the empty async block needed here?

indietyp (approved):
LGTM, thank you! 🎉

apps/hash-graph/bench/Cargo.toml
TimDiekmann: I don't really mind either way. Personally I think I like the `{}` notation more as it's more consistent with `path = `, but this could also be written as `.path`. Maybe we can change it globally at some point but I prefer not having it in a few places only.

apps/hash-graph/bin/cli/src/subcommand/type_fetcher.rs
TimDiekmann: For iterators you would typically use `.for_each(drop)`, but I'm not aware of a similar approach for streams. We need to consume the iterator but `collect` seems wrong here (even though it's `()`).

TimDiekmann: I changed that line btw because `clippy` complained.

indietyp (approved):

PR #2960, opened 2023-08-21
GEN-71: Fix `deer` CI failure

indietyp:

This fixes the CI failure exhibited because `similar-assert` `1.5` has been released; the change made it so that the macro now invokes another macro (instead of a function), leading to `similar-assert` calling the `format!` macro in client code instead of library code.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/deer/desert/src/assert.rs
+++ b/libs/deer/desert/src/assert.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "pretty")]
-use alloc::format;
 use core::fmt::Debug;
 
 use deer::{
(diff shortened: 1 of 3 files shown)


libs/deer/desert/src/assert.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "pretty")]
+use alloc::format;
```
TimDiekmann, line 2: This does not seem like an error the client is supposed to solve but the upstream library. What do you think?

TimDiekmann (dismissed):

TimDiekmann (approved):

libs/deer/desert/src/assert.rs
indietyp: I want to fix the CI as soon as possible, but I will create a PR upstream soon to fix it instead. This will be a tiny bit more complicated, however.

PR #2963, opened 2023-08-22
H-528: Fix bugs when navigating between types

CiaranMn:

This PR fixes two bugs related to the entity type form: a crash could occur when switching between viewing two different types in the form, either navigating via the sidebar (moving between pages for two different types), or navigating between types in the preview slide.

The crash was caused by a mismatch between the data loaded into the form, which contains references to property types, and the property type options available to the form: it was possible for the options to be missing the property types the form data referenced. The same is true for link types.

There were two variants of this:
**1.** In the preview slide, navigating from a type with any properties to another type which was missing at least one of those properties caused a crash because the property type options were changed before the form data. The PR fixes this by resetting the slide's state when the URL loaded into it changes, by adding a `key` to the component. This also has a UX benefit of causing the slideover animation to repeat when moving between types, making it clear to the user when the type in the slide has changed (rather than values changing, which can be hard to spot depending on how many values change). 

**2.** If viewing a type page which refers to a property type below that property type's latest version, navigating to another type's page via a link in the sidebar would cause a crash. This PR fixes that by changing when the form data resets so that the previous type's data isn't left in the form when the property type options have been changed (since are then missing the non-latest property type).

**Drive-by:** I noticed that the button to remove a type from the 'Extends' section was present in the preview slide, and have fixed that also.

## 🔗 Related links

- [Slack report](https://hashintel.slack.com/archives/C03TWBSRT8B/p1692691432748659) (internal)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

You can recreate the bugs in `main` by doing the following, and then switching to the branch and confirming there is no crash:

**Bug 1**
1. Create Type A that has at least one property, and a link to another type Whatever (doesn't matter which)
2. Create Type B that links to Type A
3. From Type B's page, click Type A in the links to open the preview slide, and then from Type A click the link to Whatever – it should crash in `main` but not in the branch

**Bug 2**
1. Have any type which refers to a non-latest property type (e.g. go to Type B, add whatever property you added to Type A, and amend it so that it is no longer the latest on Type A)
2. Be on that type's page, and then click another type in the sidebar – it should crash in `main` but not in the branch

**Bug 3**
1. Load a type which extends another type in the preview slide and hover over the type in 'Extends' – the 'x' should appear in `main` but not in the branch

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
+++ b/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
@@ -379,6 +379,7 @@ const Page: NextPageWithLayout = () => {
 
       {previewEntityTypeUrl ? (
         <TypePreviewSlide
+          key={previewEntityTypeUrl}
           onClose={() => setPreviewEntityTypeUrl(null)}
           onNavigateToType={onNavigateToType}
           typeUrl={previewEntityTypeUrl}
--- a/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/use-entity-type-value.tsx
+++ b/apps/hash-frontend/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/use-entity-type-value.tsx
@@ -243,15 +243,18 @@ export const useEntityTypeValue = (
 
   const completedRef = useRef<VersionedUrl | null>(null);
 
-  useLayoutEffect(() => {
-    if (
-      stateEntityType &&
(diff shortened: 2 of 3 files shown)


CiaranMn:
> Finding this hard to test because all of the sidebar links to types in the deployment at https://hash-gv8y9jmkr.stage.hash.ai/ point to https://app.hash.ai/ instead -- FYI. Don't think introduced in this PR but we should fix and then I can test properly.

@nonparibus I don't think this is fixable without a staging backend, because the backend assumes all types it's asked to create are for `app.hash.ai`. The only thing we could do is create a type with whatever URL the frontend request, but this then leaves it open to people to request any URL they want for the type.

PR #2977, opened 2023-08-24
H-182, H-184, H-369: Organization management tweaks and bug fixes

CiaranMn:

Makes a few small improvements to organization management:
1. Moves the 'Create organization' page to settings, and adds a button to get there from `/settings/organizations`  (H-369)
2. For user-provided org website values, displays them on the profile page (`/@org-shortname`) and sanitize them there. Also client-side validation (sticking plaster, server to follow later) (H-182).
3. Makes a newly created organization appear in the user's workspace list as soon as it is created – previously required a refresh. This is because there's a server-side fetch for the user, which is cached after the initial load (by cookie value), but  there's also code in the auth context provider to reset the authenticated user with whatever data the server provides each time, meaning that the cached server data could override the updated data just refetched by the client. This PR fixes that by only taking the server data on the first load (since it's not going to change after that, and if it needs to change the client knows when). This doesn't cover times when the user is updated by _other_ clients, but handling this this requires polling or a subscription anyway, and wouldn't be covered by a cached server value. H-184
4. Drive-by: removes a tooltip offset that was causing the tooltip to interfere with the clickability of the items it was attached to

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Next up for organizations is the ability to upload an avatar

## 🛡 What tests cover this?

- None really.

## ❓ How to test this?

1.  Try creating an org via the settings menu, check it shows up in the workspace switcher immediately
2. Try editing the website of the org via its settings page

## 📹 Demo

Screenshot 2023-08-24 at 16 31 17
Screenshot 2023-08-24 at 17 34 46
Screenshot 2023-08-24 at 17 35 09


nonparibus (approved):
Follow-ups in H-583, H-584 and H-585.

PR #3014, opened 2023-08-28
H-469: Add basic Zanzibar structs and simple check functionality

TimDiekmann:

This is the first PR of adding basic Zanzibar functionality to the Graph backend.

As discussed, this will be based on a common `AuthorizationBackend` trait to maybe add other backends later. Currently, this adds the functionality for `SpiceDB`. In this PR only the `check` functionality is implemented.

Care has been taken to implement the interface more generically as the Authorization service is expected to be a backbone of the Graph and extremely likely a bottleneck performance-wise if not done correctly.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

There are no automated tests yet as no schema is defined. These will be added as a follow-up. I expect an extensive test suite for permissions as this is a backbone of the whole graph.


apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -0,0 +1,56 @@
...
+type StringTuple = Tuple<
+    GenericResource<String, String>,
+    GenericAffiliation<String>,
+    GenericSubject<GenericResource<String, String>, GenericAffiliation<String>>,
+>;
+
+#[derive(Debug)]
+pub struct CheckError {
+    pub tuple: StringTuple,
+}
```
TimDiekmann, line 39: Errors are likely to be moved out to a different folder at a later point. I'm not 100% sure about the design of the errors but I wanted the tuple to be associated with the error itself.

apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl fmt::Display for CheckError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(fmt, "failed to check permission: `{}`", self.tuple)
```
TimDiekmann, line 43: I expect that printing the tuple will help debugging

apps/hash-graph/lib/authorization/src/backend.rs
indietyp: Good idea. I think maybe having the tuple as well as a reason enum (if there are multiple) would be appropriate in that case.

`AuthorizationBackend` could be generic over the `reason` enum, which must implement `Display`; that way different backends can have different errors.

apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -0,0 +1,47 @@
...
+        &self,
+        subject: &S,
+        affiliation: &A,
+        resource: &R,
+        consistency: Consistency<'_>,
+    ) -> Result<CheckResponse, Report<CheckError>>
+    where
+        S: Subject + Sync,
+        A: Affiliation<R> + Sync,
+        R: Resource + Sync;
```
indietyp, line 24: Wait, isn't this automatically `!Send`? You can force that a trait implementation must be `Send` by using:

```suggestion
    fn check<S, A, R>(
        &self,
        subject: &S,
        affiliation: &A,
        resource: &R,
        consistency: Consistency<'_>,
    ) -> impl Future<Output = Result<CheckResponse, Report<CheckError>>> + Send + '_
    where
        S: Subject + Sync,
        A: Affiliation<R> + Sync,
        R: Resource + Sync;
```

Implementors can still use `async fn`, but are forced to be `Send`.

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
```diff
@@ -0,0 +1,303 @@
...
+        match self {
+            Self::Request(error) => fmt::Display::fmt(error, fmt),
+            Self::Rpc(status) => write!(fmt, "Error {}: {}", status.code, status.message),
+        }
+    }
+}
+
+impl Error for InvocationError {}
+
+enum InvocationResult<T> {
```
indietyp, line 171: why is `InvocationResult` not just a `Result`?

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
```diff
@@ -0,0 +1,303 @@
...
+    Success(T),
+    Failure(RpcStatus),
+}
+
+impl SpiceDb {
+    async fn request<R: DeserializeOwned>(
+        &self,
+        path: &'static str,
+        body: &(impl Serialize + Sync),
+    ) -> Result<InvocationResult<R>, reqwest::Error> {
```
indietyp, line 181: I do not particularly like that we essentially return `Result<Result<R, RpcStatus>, request::Error>` here, IMO only a single `Result` should be returned, e.g.

```rust
enum RequestError {
	Request(reqwest::Error),
	Remote(RpcStatus)
}

Result<R, RequestError>
```

indietyp: I mean `InvocationError` already exists.

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
```diff
@@ -0,0 +1,303 @@
...
+            #[serde(rename = "PERMISSIONSHIP_CONDITIONAL_PERMISSION")]
+            Conditional,
+        }
+
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct RequestResponse {
+            checked_at: schema::ZedToken<'static>,
+            permissionship: Permissionship,
+        }
```
indietyp, line 258: specific reason why all of these were implemented in the function body instead of outside? (just a subjective style thing, I try to avoid declaring types in function bodies as it tends to decrease readability quite a bit IMO)

apps/hash-graph/lib/authorization/src/zanzibar.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Note that the snapshot used will be loaded at the beginning of the API call, and that new
+    /// data written after the API starts executing will be ignored.
+    ///
+    /// > ## Warning
+    /// >
+    /// > Use of `FullyConsistent` means little caching will be available, which means performance
+    /// > will suffer. Only use if a [`Zookie`] is not available or absolutely latest information
+    /// > is required.
+    FullyConsistent,
+}
```
indietyp, line 264: Didn't miss this file, @TimDiekmann and I talked about this in detail offline already, so no comments :P

indietyp (changes requested):

apps/hash-graph/lib/authorization/src/backend.rs
TimDiekmann: The function is designed to not require `Send`, the values are never passed into an `await` block, only the reference.
This is the first time I use the `async trait` feature, if it turns out to be required I'll change it later

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
TimDiekmann: The structs are only expected to be used here so the scope is not polluted. As these are only relevant here the definitions are closer to the location where they are used.

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
TimDiekmann: Do’h, the enum was more complex previously, I'll overhaul this section.

apps/hash-graph/lib/authorization/src/backend.rs
TimDiekmann: A generic `reason` would try to invent a new error-handling. For that, I'd use error-stack. It is, however, always known what tuple is checked.

apps/hash-graph/lib/authorization/src/zanzibar.rs
TimDiekmann: Did you know that you can comment on a file? 🙂 
image

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
TimDiekmann: Removed in [`0a2ba99` (#3014)](https://github.com/hashintel/hash/pull/3014/commits/0a2ba99b9a56408bb5bceede1bf323f27e759310)

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
TimDiekmann: Changed in [`0a2ba99` (#3014)](https://github.com/hashintel/hash/pull/3014/commits/0a2ba99b9a56408bb5bceede1bf323f27e759310)

apps/hash-graph/lib/authorization/src/zanzibar.rs
indietyp: d'oh, didn't think about that when (I just woke up when I wrote the review^^) :P

indietyp (approved):

PR #3026, opened 2023-08-29
H-469: Support creating new relationships

TimDiekmann:

New relations should be able to be added from the Graph.

## 🔍 What does this change?

- Adds new function `update_relationship` and expose it as `create_relation`
- Make ordering of parameters consistent: 1. resource, 2. affiliation, 3. subject. This makes it consistent with a `Tuple` and across different functions 
- Make it possible to pass unsized types to the functions

## 🚫 Blocked By

- #3014 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/authorization/src/backend.rs
+++ b/apps/hash-graph/lib/authorization/src/backend.rs
@@ -6,22 +6,47 @@ use std::error::Error;
 use error_stack::Report;
 
 pub use self::spicedb::{SpiceDb, SpiceDbConfig};
-use crate::zanzibar::{Affiliation, Consistency, Resource, StringTuple, Subject, Zookie};
+use crate::zanzibar::{Affiliation, Consistency, Relation, Resource, StringTuple, Subject, Zookie};
 
 /// A backend for interacting with an authorization system based on the Zanzibar model.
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -6,22 +6,47 @@ use std::error::Error;
...
+    async fn create_relation<R, A, S>(
         &self,
-        subject: &S,
+        resource: &R,
         affiliation: &A,
+        subject: &S,
+    ) -> Result<CreateRelationResponse, Report<CreateRelationError>>
+    where
+        R: Resource + ?Sized + Sync,
+        A: Relation<R> + ?Sized + Sync,
```
indietyp, line 27: Shouldn't this be `Affiliation` instead?

indietyp (changes requested):

apps/hash-graph/lib/authorization/src/backend.rs
TimDiekmann: No, you cannot add a permission, only a relation.

indietyp (approved):

indietyp (approved):

PR #3029, opened 2023-08-29
H-649: Support deletion of relationships

TimDiekmann:

Similar to the creation of relationships, the deletion is fundamental for a permission system.

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -11,22 +11,39 @@ use crate::zanzibar::{Affiliation, Consistency, Relation, Resource, StringTuple,
...
         R: Resource + ?Sized + Sync,
         A: Relation<R> + ?Sized + Sync,
         S: Subject + ?Sized + Sync;
 
+    /// Deletes a new relation between a [`Subject`] and an [`Resource`] with the specified
+    /// [`Relation`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the relation already exists or could not be created.
```
indietyp, line 35:
```suggestion
    /// Returns an error if the relation does not exist or could not be deleted.
```

indietyp (changes requested):

indietyp (approved):

PR #3034, opened 2023-08-29
H-607: Allow user upload of avatars for organizations

CiaranMn:

Allows users to upload an avatar for an organization via its setting page.

## 🔍 What does this change?

- Adds a new `RemoteImageFile` type (hosted on blockprotocol.org) to use for image files
- Removes the obsolete HASH-specific file type which we stopped using in #3011 
- Adds `hasAvatar` links to user and org system entity types
- Adds an image field component, supported by an image dropzone component
- Regenerates system file types

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Generic file upload feature
- Use the avatar in the workspace switcher and org's profile page

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1.  Try adding an avatar to an org

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/4f0da6ec-aa2f-4b57-84ee-797d12e31edf


nonparibus:
@CiaranMn Dropped a Q via Slack.

nonparibus (approved):
@TimDiekmann LGTM; would appreciate you taking a look as well to the extent it makes sense. Cheers CM!

apps/hash-frontend/src/components/hooks/block-protocol-functions/knowledge/use-block-protocol-file-upload.ts
```diff
@@ -16,6 +17,7 @@ import {
 import { UploadFileRequestCallback } from "./knowledge-shim";
 
 export const useBlockProtocolFileUpload = (
+  ownedById?: OwnedById,
```
TimDiekmann, line 20: What is the reason we cannot make `ownedById` required?

apps/hash-frontend/src/graphql/queries/user.queries.ts
```diff
@@ -8,11 +8,14 @@ export const isShortnameTaken = gql`
   }
 `;
 
+// We follow links rightwards twice in order to get the orgs that the user is a member of, and the orgs' avatars
+// This may lead to significant overfetching when the user has a lot of e.g. pages
+// @todo consider fetching the org data separately (e.g. in useOrgs)
```
TimDiekmann, line 13: Not suggesting to do it now, just wondering how much work this would be? Fetching at a depth of two seems bad (most likely not an issue in the short-term future)

apps/hash-frontend/src/lib/user-and-org.ts
```diff
@@ -248,10 +255,30 @@ export const constructOrg = (params: {
...
+
   const org = constructMinimalOrg({
     orgEntity,
   }) as Org;
 
+  org.hasAvatar = hasAvatar
+    ? {
+        // these are each arrays because each entity can have multiple revisions
+        linkEntity: hasAvatar.linkEntity[0] as LinkEntity,
+        rightEntity: hasAvatar.rightEntity[0] as unknown as RemoteImageFile,
```
TimDiekmann, line 278: Just to improve my TS skills: whe `unknown` is needed here? Shouldn't the `Entity` type and `RemoteImageFile` type overlap?

TimDiekmann (approved):
Looks good to me, just a few questions, nothing blocking. 🎉

apps/hash-frontend/src/components/hooks/block-protocol-functions/knowledge/use-block-protocol-file-upload.ts
CiaranMn: In various places we call it we don't reliably have the active workspace's account id – e.g. in pages where the viewer might be anonymous (viewing a public page). Ideally we'd rework the logic of pages so that we're not attempting to use this unless we have an authenticated user, but that's a separate piece of work

apps/hash-frontend/src/graphql/queries/user.queries.ts
CiaranMn: Not much – tracking in H-657. Ideally we'd be able to filter the types of links which are followed by that's much more work

apps/hash-frontend/src/lib/user-and-org.ts
CiaranMn: They do overlap, but they don't "sufficiently" overlap to satisfy the compiler that we should be doing this. The problem here is that `RemoteImageFile` has specific properties which a generic `Entity` doesn't, and therefore it is unsafe to do this cast, since there's no indication that `rightEntity` has any of the specific properties required

apps/hash-frontend/src/components/hooks/block-protocol-functions/knowledge/use-block-protocol-file-upload.ts
nonparibus: Captured as H-660.

PR #3042, opened 2023-08-31
H-666: Hide 'General' from use case list

CiaranMn:

On the hash.dev [roadmap page](https://hash.dev/roadmap), there's a 'General' category of feature area – this is meant to be a method of categorising and filtering roadmap items, but not listed as a 'Use case'. This PR fixes that by hiding it from the 'Use case' list.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1.  View the deployment

## 📹 Demo

Screenshot 2023-08-31 at 09 28 30

### DIFF
--- a/apps/hashdotdev/src/pages/roadmap.page.tsx
+++ b/apps/hashdotdev/src/pages/roadmap.page.tsx
@@ -129,43 +129,45 @@ const UseCases: FunctionComponent = () => (
       flexWrap="wrap"
       sx={{ justifyContent: { xs: "space-between", sm: "flex-start" } }}
     >
-      {useCases.map(({ name, icon }) => (
-        /** @todo: make clickable when docs pages exist for each item */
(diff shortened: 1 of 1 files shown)


nonparibus:
@CiaranMn @TimDiekmann Any idea re: these spurious playwright test failures here?

TimDiekmann:
Sadly, this happens sometimes. For some reason the backend is slow to start and PW does not wait long enough. I recently increased the timeout even further (it was one minute before, which should have been way more than enough). We could just raise it again and hope it will solve it.

TimDiekmann:
I did some further investigation. The error was not a slow backend to start (the timeout still happened because it didn't start at all) but the Type Fetcher was not able to get a Block Protocol entity type. The request failed with a "the request exceeded its deadline" error.

PR #3043, opened 2023-08-31
H-654, H-664: Implement deletion for relations based on a filter

TimDiekmann:

The current authorization types have a few issues:
- They are not object safe
- Operations which do not require to have a type present needs to be typed anyway (such as filters)
- Operations who takes multiple relations requires the same type (especially as the types are not object safe so we cannot use `dyn Trait`

To properly implement batch deletion of relations I took the opportunity to do both things at once, refactor some types to make `Affiliation` and `Resource` object safe. `Subject` is still not object safe as it's using `Self` in its definition. I will revisit this issue later when allowing to insert multiple relations at once, probably it's not needed to make it object safe.

## 🔗 Related links

- https://doc.rust-lang.org/reference/items/traits.html#object-safety

## 🔍 What does this change?

- Make the types more approachable by splitting them into a `model` and `api` module
- Use `AsRef<str>` instead of `Serialize`. `AsRef` is a little bit more annoying to implement but in comparison to `Serialize` it can be made into an object
- Implement deletion of relations based on a filter (batch deletion)

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/authorization/src/backend.rs
indietyp, line 1: Just to clarify, changes in the `SubjectFilter` and `RelationshipFilter` have been discussed in great length offline, which is why there won't be any comments on implementation.

apps/hash-graph/lib/authorization/src/backend/spicedb.rs
```diff
@@ -1,509 +1,16 @@
...
-    pub struct SubjectFilter<'a> {
-        pub subject_type: &'a str,
-        #[serde(skip_serializing_if = "Option::is_none", rename = "optionalSubjectId")]
-        pub subject_id: Option<&'a str>,
-        #[serde(skip_serializing_if = "Option::is_none", rename = "optionalRelation")]
-        pub relation: Option<&'a str>,
-    }
-}
+mod api;
+mod model;
```
indietyp, line 4: I really like the split into two different files 👍

indietyp (approved):

PR #3044, opened 2023-08-31
H-680: Write initial test suite for authorization

TimDiekmann:

Authorization is a backbone of the Graph and needs to be very well tested. A proper testing strategie is required, which includes an automatic cleanup after running.

## 🚫 Blocked by

- #3043

## 🔍 What does this change?

- Adds base logic for a test suite
- Tests the current API with a very simple schema

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/authorization/tests/schemas/simple.zed
+++ b/apps/hash-graph/lib/authorization/tests/schemas/simple.zed
@@ -1,0 +1,9 @@
+definition graph/account {}
+
+definition graph/entity {
+    relation writer: graph/account
+    relation reader: graph/account
+
+    permission edit = writer
+    permission view = reader + edit
+}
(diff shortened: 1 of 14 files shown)


apps/hash-graph/lib/authorization/tests/schemas/simple.zed
indietyp, line 1: man, it would be so cool (and way to overkill) to generate the rust schema from the `zed` schema.

indietyp (approved):
great to see some tests!

apps/hash-graph/lib/authorization/tests/schemas/simple.zed
TimDiekmann: Yes, absolutely. Maybe some day 😄

TimDiekmann: The other way around would also be neat 😂

indietyp (approved):

PR #3051, opened 2023-09-01
H-642: Allow mass creation/deletion of relationships

TimDiekmann:

SpiceDB allows to modify (touch, create, delete) relations. For simplicity we only expose a non-batched version yet. This PR improves some DX issues we had previously (discussed offline) and allows multiple tuples to be inserted/deleted at the same time.

## 🔍 What does this change?

- Support preconditions in modification endpoints
- Simplify interface to improve DX and probably compile time
- Allow multiple tuples to be passed in modification endpoints

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

The authorization integration tests uses the new batch creation/deletion methods

### DIFF
--- a/apps/hash-graph/lib/authorization/src/backend.rs
+++ b/apps/hash-graph/lib/authorization/src/backend.rs
@@ -6,9 +6,7 @@ use std::error::Error;
 use error_stack::Report;
 
 pub use self::spicedb::{SpiceDb, SpiceDbConfig};
-use crate::zanzibar::{
-    Affiliation, Consistency, Permission, Relation, Resource, StringTuple, Subject, Zookie,
-};
+use crate::zanzibar::{Affiliation, Consistency, Relation, Resource, Tuple, UntypedTuple, Zookie};
 
 /// A backend for interacting with an authorization system based on the Zanzibar model.
 pub trait AuthorizationApi {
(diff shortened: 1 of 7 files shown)


apps/hash-graph/lib/authorization/src/zanzibar.rs
```diff
@@ -5,234 +5,223 @@ use std::borrow::Cow;
 
 use serde::{Deserialize, Serialize};
 
-/// The relation or permission of a [`Subject`] to an [`Resource`].
+/// The relation or permission of a [`Resource`] to another [`Resource`].
 pub trait Affiliation<R: Resource + ?Sized>: AsRef<str> {}
 
-/// A computed set of [`Subject`]s for a particular [`Resource`].
+impl<A: Affiliation<R>, R: Resource> Affiliation<R> for &A {}
+// impl<A: Affiliation<R>, R: Resource> Affiliation<&R> for A {}
```
indietyp, line 12: Why is this commented out, and if it is not needed why is it still here?

TimDiekmann: I forgot to remove that, thank you

apps/hash-graph/lib/authorization/src/backend.rs
```diff
@@ -33,73 +31,58 @@ pub trait AuthorizationApi {
...
 
-    /// Creates a new relation between a [`Subject`] and an [`Resource`] with the specified
-    /// [`Relation`].
+    /// Creates a new relation specified by the [`Tuple`].
     ///
     /// # Errors
     ///
     /// Returns an error if the relation already exists or could not be created.
-    async fn create_relation<R, A, S>(
+    async fn create_relation<'p, 't, T>(
```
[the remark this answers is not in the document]
TimDiekmann: Good point, thanks!

apps/hash-graph/lib/authorization/src/zanzibar.rs
TimDiekmann: Removed in da0abcf0a3792aff1edcd21cd699fa2dbb623886. I expect to remove all reference-based parameters as all types should implement `Clone` anyway and this avoids issues like passing reference to tuples and such.

apps/hash-graph/lib/authorization/src/backend.rs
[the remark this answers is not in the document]
TimDiekmann: If you don't mind I'll do that in my follow up to avoid conflicts.

indietyp (approved):

PR #3053, opened 2023-09-01
H-696: Add tests to test all API endpoints

TimDiekmann:

Adds various test cases for the currently implemented API

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/authorization/tests/api.rs
+++ b/apps/hash-graph/lib/authorization/tests/api.rs
@@ -33,7 +33,7 @@ impl TestApi {
     ///
     /// After disconnecting every created relation will be deleted again.
     #[must_use]
-    #[allow(clippy::missing_panics_doc)] // Only the cleanup thread may panic
+    #[allow(clippy::missing_panics_doc, clippy::print_stderr)] // Only the cleanup thread may panic
     pub fn connect() -> Self {
         let host =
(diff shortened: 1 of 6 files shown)


apps/hash-graph/lib/authorization/tests/api.rs
```diff
@@ -68,10 +68,16 @@ impl TestApi {
 
                     let tuples: Vec<UntypedTuple> =
                         tuple_receiver.await.expect("failed to receive tuples");
-                    client
-                        .delete_relation(&tuples, [])
-                        .await
-                        .expect("failed to delete relations");
+
+                    if let Err(error) = client.delete_relations(&tuples, []).await {
+                        eprintln!(
```
indietyp, line 73: Shouldn't this be tracing::error instead?

indietyp (changes requested):

apps/hash-graph/lib/authorization/tests/api.rs
TimDiekmann: We don't have `tracing` set up in the tests and I don't want to mess around with that here.

indietyp: 👍

indietyp (approved):

indietyp (approved):

indietyp (approved):

PR #3058, opened 2023-09-02
GEN-55: Move `sim-engine` from `hashintel/hash` repo into `/labs`

TimDiekmann:
## 🔗 Related links

- https://github.com/hashintel/labs/pull/29

This PR:

The changes in this PR:

The changes in this PR:


.github/scripts/rust/setup.py
```diff
@@ -43,7 +43,7 @@
 ]
 
 # Build a coverage report for these crates
-COVERAGE_EXCLUDE_PATTERNS = ["apps/engine**"]
+COVERAGE_EXCLUDE_PATTERNS = []
```
nonparibus, line 46:
```suggestion
COVERAGE_EXCLUDE_PATTERNS = []
```

@TimDiekmann Do we need this line at all?

TimDiekmann: This is still checked in the `setup.py` script so we could just add patterns here. I avoided any other changes than the removal.

nonparibus (approved):
I expect we'll find follow-ups, but this LGTM right now. Thanks!

PR #3077, opened 2023-09-02
GEN-57: Run BE-Test and PW-Test only if needed in CI

TimDiekmann:

The current `CI` workflow always runs regardless of the changes. improves the behavior to only run it if needed. This can be optimized further such as only lint and run the unit tests if needed but the BE and PW tests requires a big setup this this is the first step to improve the behavior

## 🔍 What does this change?

- Adds a dependency for the PW tests on the frontend and make the API a dependency of the frontend which implies a chain of dependencies
- Check for changes in CI
- Run the bespoken jobs only if required

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,16 +38,22 @@ jobs:
         id: packages
         run: |
           if [ "${{ github.event_name }}" = "pull_request" ]; then
+            # `github.event.pull_request.base.sha` does not give the SHA of the last common ancestor but the most recent
+            # commit on the base branch. This is not what we want, so we use `git merge-base` instead.
             TURBO_REF_FILTER="$(git merge-base ${{ github.event.pull_request.base.ref }} ${{ github.event.pull_request.head.ref }})"
           else
+            # `github.event.before` gives back the most recent commit on the base branch.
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -17,8 +17,52 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  setup:
+    runs-on: ubuntu-22.04
+    outputs:
+      be-tests: ${{ steps.packages.outputs.be-tests }}
+      pw-tests: ${{ steps.packages.outputs.pw-tests }}
```
nonparibus, line 24: When I read `pw-tests` I think "password" not "Playwright". We have an outstanding `@todo` to rename that test directory anyway, as it's not semantically descriptive (e.g. "FE integration tests"). But we can address this in follow-up (remembering to change this then, too).

nonparibus (dismissed):

.github/workflows/ci.yml
TimDiekmann: I 100% agree

nonparibus (approved):

PR #3080, opened 2023-09-02
GEN-99: Create a global Test/Lint CI

TimDiekmann:

The current CI is (a) slow and (b) runs various configuration which are not needed. This PR provides a new approach for the CI to only run required scripts.

## 🔍 What does this change?

- Create an action to prune the repository for a given scope so only required packages needs to be installed.
- Create a workflow to run tests for only changed packages
- Fix various dependencies (many dependencies were missing as they were installed from other packages)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- `patch-packages` will always fail on CI and it's not possible to change this behavior. The prune-step will change the post-install step to allow it to fail

## 🛡 What tests cover this?

This PR changes the global `turbo.json` so every possible package will run thus every affected package is tested.


nonparibus (approved):

.github/workflows/package.yml
```diff
@@ -0,0 +1,373 @@
...
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        with:
+          fetch-depth: 2
+
+      - name: Install turbo
+        run: yarn global add turbo
+
+      - name: Determine changed packages
+        id: packages
+        run: |
```
indietyp, line 38: Would it make sense for these large `run` statements to put them into separate `.sh` script files? I feel like that would increase maintainability quite drastically as we can debug/run/edit them locally before pushing and avoiding the painful push -> check -> change workflow.

TimDiekmann: I was planning on having a separate script file for this but will do that in a follow-up.  I'd prefer to have the script either in another language or adding shellckeck to the CI.

PR #3091, opened 2023-09-04
H-705: Enable Sentry integration in `hash-graph`

indietyp:

This enables Sentry integration into the hash-graph. Like OTLP, Sentry is essentially disabled if the `HASH_GRAPH_SENTRY_DSN` is unset.

To properly spawn sentry the configuration of the runtime has been delegated, as the sentry guard needs to be initialized before tokio starts (it even works if that's not the case, but then measurements can be a bit misleading).

This PR:

The changes in this PR:

The changes in this PR:


.env
```diff
@@ -64,6 +64,7 @@ HASH_REDIS_HOST=localhost
 HASH_REDIS_PORT=6379
 
 HASH_OTLP_ENDPOINT=http://localhost:4317
+HASH_GRAPH_SENTRY_DSN=
```
TimDiekmann, line 67: We should not set an empty default in `.env`

apps/hash-graph/bin/cli/src/args.rs
```diff
@@ -6,6 +6,9 @@ use crate::subcommand::Subcommand;
 #[derive(Debug, Parser)]
 #[clap(version, author, about, long_about = None)]
 pub struct Args {
+    #[arg(long, env = "HASH_GRAPH_SENTRY_DSN")]
+    pub sentry_dsn: Option<Option<sentry::types::Dsn>>,
```
TimDiekmann, line 10: Why this is a nested `Option`?

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -5,27 +5,33 @@ mod args;
...
-        Subcommand::Completions(ref args) => {
-            subcommand::completions(args);
-            Ok(())
-        }
-        Subcommand::Snapshot(args) => subcommand::snapshot(args).await,
-        #[cfg(all(hash_graph_test_environment, feature = "test-server"))]
-        Subcommand::TestServer(args) => subcommand::test_server(args).await,
-    }
+    let _sentry = sentry::init(sentry::ClientOptions {
+        dsn: args.sentry_dsn.clone().flatten(),
```
TimDiekmann, line 21: Why the `sentry_dsn` has to be cloned? You're destructuring `args`and only call `execute` on another variable below.

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -5,27 +5,33 @@ mod args;
...
-        Subcommand::TypeFetcher(args) => subcommand::type_fetcher(args).await,
-        Subcommand::Completions(ref args) => {
-            subcommand::completions(args);
-            Ok(())
-        }
-        Subcommand::Snapshot(args) => subcommand::snapshot(args).await,
-        #[cfg(all(hash_graph_test_environment, feature = "test-server"))]
-        Subcommand::TestServer(args) => subcommand::test_server(args).await,
-    }
+    let _sentry = sentry::init(sentry::ClientOptions {
```
TimDiekmann, line 20: For what the `_sentry` variable is used? If this is a guard we should name it like this and/or add a comment.

apps/hash-graph/bin/cli/src/subcommand.rs
```diff
@@ -33,3 +36,30 @@ pub enum Subcommand {
...
+    }
+
+    pub(crate) fn execute(self) -> Result<(), GraphError> {
+        // create a runtime
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("failed to create runtime");
+
+        runtime.block_on(self.delegate())
```
TimDiekmann, line 63: Why do we block on `delegate` if not every subcommand in `delegate` is asynchronous? The `completions` subcommand does not require a runtime at all. I know that we previously had a similar behavior with `#[tokio::main]` but as we don't have this anymore we also don't need to spin up a runtime for that.

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -5,27 +5,33 @@ mod args;
...
 
+use std::sync::Arc;
+
 use error_stack::Result;
 use graph::load_env;
 
-use self::{args::Args, error::GraphError, subcommand::Subcommand};
+use self::{args::Args, error::GraphError};
 
-#[tokio::main]
```
TimDiekmann, line 13: With this change it's possible to remove the `macros` feature from `tokio` I think.

apps/hash-realtime/src/index.ts
```diff
@@ -66,8 +66,8 @@ const acquireReplicationSlot = async (
             -- Second argument to this calls specifies whether or not the replication
             -- slot is temporary.
             select 1 from pg_create_logical_replication_slot(${slotName}, 'wal2json', ${
-              options.temporary ?? true
-            })
+    options.temporary ?? true
+  })
```
TimDiekmann, line 70: Do you have a different `prettier` version installed or why was this changed?

TimDiekmann (changes requested):

apps/hash-graph/bin/cli/src/args.rs
indietyp: I thought (from the clap documentation) that `Option<Option<T>>` would allow for empty values. I was wrong. Removed in [8e62433](https://github.com/hashintel/hash/pull/3091/commits/8e624331872d7bb6331f7e7bfd70a877aa182e66)

apps/hash-graph/bin/cli/src/main.rs
indietyp: Nice catch! Fixed in [8e62433](https://github.com/hashintel/hash/pull/3091/commits/8e624331872d7bb6331f7e7bfd70a877aa182e66)

apps/hash-graph/bin/cli/src/main.rs
indietyp: Added documentation in [8e62433](https://github.com/hashintel/hash/pull/3091/commits/8e624331872d7bb6331f7e7bfd70a877aa182e66)

apps/hash-graph/bin/cli/src/subcommand.rs
indietyp: made optional in [8e62433](https://github.com/hashintel/hash/pull/3091/commits/8e624331872d7bb6331f7e7bfd70a877aa182e66)

apps/hash-graph/bin/cli/src/main.rs
indietyp: let's try it! removed in [c413faf](https://github.com/hashintel/hash/pull/3091/commits/c413faf59488886df56f0a4abc5162c3cf3412b3)

apps/hash-realtime/src/index.ts
indietyp: I was wondering about that myself, but I have yarn installed repeatedly (with `just yarn install`) and I always get the same result, 4 files in actuality changed. This one slipped through, the other three I correctly stashed.

apps/hash-graph/bin/cli/src/args.rs
TimDiekmann: For what an empty value would be used?

indietyp: The idea was to set in the .env (and in the production `docker-compose.prod.yml`) the `HASH_GRAPH_SENTRY_DSN` to an empty value (so `HASH_GRAPH_SENTRY_DSN=`), which would then render to `None` because it is an empty value, so sentry isn't activated (I mean an empty string will stop startup anyway as it isn't a valid DSN)

apps/hash-realtime/src/index.ts
indietyp: Yea, if we look into the CI we can see:

```
Checking formatting...
[warn] apps/hash-realtime/src/index.ts
[warn] apps/hash/README.md
[warn] tests/hash-playwright/tests/shared/login-using-ui.ts
[warn] Code style issues found in 3 files. Run Prettier to fix.
```

Those are the other three files. Weird stuff.

TimDiekmann: Well, it's good to have it fixed then I guess 😄

.env
indietyp: commented out in [8e62433](https://github.com/hashintel/hash/pull/3091/commits/8e624331872d7bb6331f7e7bfd70a877aa182e66)

apps/hash-graph/bin/cli/src/args.rs
```diff
@@ -1,11 +1,16 @@
...
 
 /// Arguments passed to the program.
 #[derive(Debug, Parser)]
 #[clap(version, author, about, long_about = None)]
 pub struct Args {
+    // we need to qualify `Option` here, as otherwise `clap` tries to be too smart and only uses
+    // the `value_parser` on the internal `sentry::types::Dsn`, failing.
+    #[arg(long, env = "HASH_GRAPH_SENTRY_DSN", value_parser = OptionalSentryDsnParser)]
+    pub sentry_dsn: core::option::Option<sentry::types::Dsn>,
+
```
TimDiekmann, line 13: I wonder if this is supposed to be a global argument. However, the only service where it does not matter is the `completions` endpoint I'd assume. (non-blocking comment)

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -3,29 +3,48 @@
...
-        Subcommand::TestServer(args) => subcommand::test_server(args).await,
+    let Args {
+        sentry_dsn,
+        subcommand,
+    } = Args::parse_args();
+
+    if let Some(dsn) = &sentry_dsn {
+        let client = sentry::Client::from_config(dsn);
+
+        assert!(client.is_enabled(), "Invalid Sentry DSN has been provided");
```
TimDiekmann, line 27: Do we want to return an error instead? I see why you use `assert!` as this is not recoverable at this point, however, it's directly next to `main()` and an error has the IMO better output.

TimDiekmann (commented):
A few minor comments, looks good otherwise 🎉

apps/hash-graph/bin/cli/src/subcommand.rs
```diff
@@ -33,3 +38,28 @@ pub enum Subcommand {
     #[cfg(all(hash_graph_test_environment, feature = "test-server"))]
     TestServer(TestServerArgs),
 }
+
+fn spawn_runtime(future: impl Future<Output = Result<(), GraphError>>) -> Result<(), GraphError> {
```
TimDiekmann, line 42: Minor suggestion: Rename to `block_on`

apps/hash-graph/bin/cli/src/parser.rs
```diff
@@ -0,0 +1,29 @@
...
+            let Some(value) = value.to_str() else {
+                return Err(Error::new(ErrorKind::InvalidValue).with_cmd(cmd));
+            };
+
+            sentry::types::Dsn::from_str(value)
+                .map(Some)
+                .map_err(|_error| Error::new(ErrorKind::InvalidValue))
+        }
+    }
+}
```
TimDiekmann, line 29: For some reason the CI fails as it interprets this parameter as required.

indietyp: This is because we use `core::option::Option<T>`, instead of `Option<T>`, which means that clap does not set it to `required = false`. We need to do this manually.

fixed in [6e591ab](https://github.com/hashintel/hash/pull/3091/commits/6e591ab5bb600a6b7881131b437fb0e044584f83)

apps/hash-graph/bin/cli/src/subcommand.rs
indietyp: renamed in [6e591ab](https://github.com/hashintel/hash/pull/3091/commits/6e591ab5bb600a6b7881131b437fb0e044584f83)

apps/hash-graph/bin/cli/src/main.rs
indietyp: fixed in [6e591ab](https://github.com/hashintel/hash/pull/3091/commits/6e591ab5bb600a6b7881131b437fb0e044584f83)

apps/hash-external-services/docker-compose.prod.yml
```diff
@@ -8,6 +8,7 @@ services:
   graph:
     image: hash-graph:prod
     environment:
+      HASH_GRAPH_SENTRY_DSN: "${HASH_GRAPH_SENTRY_DSN-}"
```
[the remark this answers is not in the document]
indietyp: fixed in [6e591ab](https://github.com/hashintel/hash/pull/3091/commits/6e591ab5bb600a6b7881131b437fb0e044584f83)

apps/hash-graph/bin/cli/src/args.rs
indietyp: Yes, you are right, but (as you noted), I have decided to move it to a global flag, as the only command that does not benefit is the completions command, and it would lead to more code duplication (and make the guard more complicated)

> In my mind it also makes sense when you `--help`, as something like sentry or tracing should likely be a global argument and come before any subcommand. That way it is easier to switch out commands, etc.

TimDiekmann: That's fair. Do you know if it's possible to exclude this argument for certain subcommands? (Just wondering, doesn't block this PR)

> something like sentry or tracing should likely be a global argument and come before any subcommand

It's possible to achieve this with `clap` when making this a `global` argument.

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -3,29 +3,51 @@
...
 
-    match args.subcommand {
-        Subcommand::Server(args) => subcommand::server(args).await,
-        Subcommand::Migrate(args) => subcommand::migrate(args).await,
-        Subcommand::TypeFetcher(args) => subcommand::type_fetcher(args).await,
-        Subcommand::Completions(ref args) => {
-            subcommand::completions(args);
-            Ok(())
+        if !client.is_enabled() {
+            return Err(Report::new(SentryError::InvalidDsn).change_context(GraphError));
```
TimDiekmann, line 29: Let's attach the DSN. These kinds of attachments makes life so much easier with `error-stack` and this should be easy enough I think.

TimDiekmann (dismissed):

apps/hash-graph/bin/cli/src/main.rs
indietyp: Good call! ([a5b4bcb](https://github.com/hashintel/hash/pull/3091/commits/a5b4bcbb75caaf586b4c96766d4ce6d094380973))

apps/hash-graph/bin/cli/src/main.rs
```diff
@@ -3,29 +3,53 @@
...
-        Subcommand::Migrate(args) => subcommand::migrate(args).await,
-        Subcommand::TypeFetcher(args) => subcommand::type_fetcher(args).await,
-        Subcommand::Completions(ref args) => {
-            subcommand::completions(args);
-            Ok(())
+        if !client.is_enabled() {
+            return Err(Report::new(SentryError::InvalidDsn)
+                .attach_printable(dsn.clone())
+                .change_context(GraphError));
         }
```
TimDiekmann, line 32:
```suggestion
        ensure!(
            client.is_enabled(),
            Report::new(SentryError::InvalidDsn)
                .attach_printable(dsn.clone())
                .change_context(GraphError)
        );
```

TimDiekmann (approved):
Whatever style you prefer 😉

TimDiekmann (approved):

TimDiekmann (approved):

PR #3093, opened 2023-09-04
H-708, H-709: Avoid hardcoding instances of the FE domain, redundant type version increments

CiaranMn:

1. Stop hardcoding 'hash.ai' as the domain for display in certain paths, and use the `frontendDomain` instead, which is derived as follows:
```ts
export const frontendUrl =
  process.env.NEXT_PUBLIC_FRONTEND_URL ??
  (process.env.NEXT_PUBLIC_VERCEL_URL
    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
    : process.env.FRONTEND_URL ?? "http://localhost:3000");

export const frontendDomain = new URL(frontendUrl).hostname;
```

2. Prevent creation of new versions of an entity type unless the form is dirty, as a safety fallback – mysterious unprompted form submission has been witnessed in the past. We should figure out what could have caused this (H-706), but in the meantime this might prevent whatever the underlying bug is from causing versions to be created when they shouldn't

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
Thank you! ❤️

nonparibus (approved):

TimDiekmann (approved):

PR #3094, opened 2023-09-04
H-253: Introduce instance-level setting to disable pages

CiaranMn:

Adds an option which can be set on the `hash-instance` entity to hide page functionality.

This does not disable any API resolvers etc to do with pages, but instead stops the frontend from attempting to use them, or showing any UI which would trigger them.

## 🔍 What does this change?

- adds a new `Pages Are Enabled` boolean property to the `hashInstance` entity
- Replaces some code we were using to shorten and camelCase properties of the `hashInstance` entity with a utility function that does the same thing
- disables the following if pages are disabled:
   - the account page list in the sidebar
   - the 'Create Page' option in the command bar
   - the 'Create Page' option in the top-right dropdown

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Because the HASH Instance entity is fetched client-side, when pages are enabled there is a brief delay before pages appear in the sidebar, while that entity is being fetched. We should fetch that on the server instead (H-710).

## 🐾 Next steps

- At some point we will expose a UI to amend the instance settings.

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1. View the deployment to see an instance with pages disabled (since the backend has not had the 'pages enabled: true' property set yet)
2. To test 'pages enabled' functionality locally, either rebuild the backend so that the HASH Instance entity is re-seeded with the 'pages enabled: true' property value, or update the entity's properties (either in the db directly or via the API) to set it true. Probably `"http://localhost:3000/@system-user/types/property-type/pages-are-enabled/": true` unless you have changed the frontend URL or system user shortname.

## 📹 Demo

1. See deployment


nonparibus:
Just to flag we do have designs for the instance admin UI (see [Figma](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=5546%3A243899&mode=design&t=VflID0XCEyIeVr1P-1); _internal link_). I don't think we should fuss about implementing this now, but if we can make it easy for ourselves in the future (documenting or centralizing these things intended to be toggleable in the future then great).

apps/hash-frontend/src/components/hooks/use-account-pages.ts
```diff
@@ -18,13 +19,23 @@ export const useAccountPages = (
...
+  const { hashInstance } = useHashInstance();
+
+  useEffect(() => {
+    if (!hashInstance?.properties.pagesAreEnabled) {
+      return;
+    }
+
+    void getAccountPages();
+  }, [getAccountPages, hashInstance?.properties.pagesAreEnabled]);
+
```
benwerner01, line 38: I think you could also just use the `skip` parameter of `useQuery` to simplify this a bit. I think it's something like:
```TypeScript
  const { data, loading } = useQuery<
    GetAccountPagesTreeQuery,
    GetAccountPagesTreeQueryVariables
  >(getAccountPagesTree, {
    variables: { ownedById, includeArchived },
    skip: !hashInstance?.properties.pagesAreEnabled
  });
```

benwerner01 (approved):
Thanks Ciaran, this looks good and is working for me.

One thing that might be worth considering to also do as part of this PR is maybe removing the `Create Page` button on the `/pages` page when pages are disabled? But I'm not sure it's required as it shouldn't be possible to navigate there in the UI, but the page remains accessible nonetheless:
image

nonparibus:
@CiaranMn I think it is worth implementing Ben's suggestion (while leaving the `/pages` route visible in case an instance already has legacy pages within it at the point they are later disabled).

apps/hash-frontend/src/components/hooks/use-account-pages.ts
CiaranMn: Ahh much simpler, thank you – https://github.com/hashintel/hash/pull/3094/commits/bcf9fd912379c6497e55bb4c62e889ef6b7db4a5

CiaranMn:
> One thing that might be worth considering to also do as part of this PR is maybe removing the `Create Page` button on the `/pages` page when pages are disabled

Thanks @benwerner01, done in https://github.com/hashintel/hash/pull/3094/commits/bcf9fd912379c6497e55bb4c62e889ef6b7db4a5

benwerner01 (approved):
Thanks Ciaran 👍

benwerner01 (dismissed):

benwerner01 (approved):

PR #3096, opened 2023-09-04
H-711: Make `docker-compose.*.yml` files composable

TimDiekmann:

We currently have a bunch of docker-compose files:

- `docker-compose.yml `- our "main" file
- `docker-compose.test.yml` - A file built on top of the main 
- `docker-compose.type-fetcher.yml` - A configuration to additionally provide the type fetcher
- `docker-compose.prod.yml` - A completely independent file mirroring the above files, doing some minor adjustments, and adding the `hash-api` and `hash-frontend` containers.

For authorization, it's required to have two different `spicedb` services, one backed by a database for `dev` and `prod` and one backed by memory for `test` so the configuration does not match the above files and making another (two) purely additive files is not really a good solution either. Instead, it's good to use the opportunity and clean up the above files by:

1. Renaming `docker-compose.yml` to `docker-compose.dev.yml`
2. Create a new `docker-compose.yml` holding all shared configuration
3. build `docker-compose.type-fetcher.yml` and `docker-compose.prod.yml` on top of that

Some further improvements can be made:

- Instead of defining three different variables in the files for the database names use the same variable name but provide them by `.env` files
- Don't expose all ports in `prod` - that's a security concern

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

GitHub does not really detect movements of files if the file is recreated afterwards, I'm sorry for the diff...

## 🛡 What tests cover this?

The CI will cover most of the configuration


.env
```diff
@@ -100,7 +95,6 @@ KRATOS_COOKIE_DOMAIN=localhost
 KRATOS_SECRETS_COOKIE=VERY-INSECURE-AND-SHOULD-ONLY-BE-USED-IN-DEV
 KRATOS_SECRETS_CIPHER=32-LONG-SECRET-NOT-SECURE-AT-ALL
 
-
```
TimDiekmann, line 103: Two newlines are not allowed in `.env` files

.env.production
TimDiekmann, line 1: Maybe useful in the future

.env.test
TimDiekmann, line 1: `spicedb` does not have a `test`-database (the reason for this PR)

apps/hash-external-services/docker-compose.dev.yml
TimDiekmann, line 1: This was largely copied from `docker-compose.yml` but GitHub does not recognize moved files if a new file is created in the old location.

apps/hash-realtime/src/index.ts
```diff
@@ -221,7 +221,7 @@ const main = async () => {
      * @todo: update how the database is set once realtime if realtime is run in the testing environment.
      *   See https://app.asana.com/0/0/1203046447168483/f
      */
-    database: getRequiredEnv("HASH_GRAPH_PG_DEV_DATABASE"),
+    database: getRequiredEnv("HASH_GRAPH_PG_DATABASE"),
```
TimDiekmann, line 224: The `dev` scripts has been adjusted that the correct environment variable will be loaded (`dev_graph` by default)

apps/hash-graph/lib/graph/src/lib.rs
```diff
@@ -24,3 +26,36 @@ pub mod store;
...
+    [
+        ".env.local",
+        &environment_path_local,
+        &environment_path,
+        ".env",
+    ]
+    .into_iter()
+    .filter_map(|path| {
+        dotenv_flow::from_filename(
+            Path::new(concat!(env!("CARGO_MANIFEST_DIR"), "/../../../..")).join(path),
```
TimDiekmann, line 56: Points to the repository root and joins the specified path

.justfile
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env just --justfile
 
-set dotenv-load := true
```
indietyp, line 3: assuming this has been removed in favor or `dotenv_flow`?

apps/hash-external-services/docker-compose.yml
```diff
@@ -256,216 +164,34 @@ services:
...
-      POSTGRES_SEEDS: "postgres" # the hostname of the postgres container
     security_opt:
       - no-new-privileges:true
 
-  temporal:
-    container_name: temporal
-    image: "temporalio/server:${HASH_TEMPORAL_VERSION}"
+  redis:
+    image: redis:6.2
+    read_only: true
```
indietyp, line 187: what does `read_only` do here? I thought it only applies to volumes?

indietyp (approved):
just some minor comments/clarifications

apps/hash-external-services/docker-compose.yml
TimDiekmann: No, `read_only` applies to the container itself

.justfile
TimDiekmann: Yes. If this were kept, the `.env` variables would be passed to the Graph and the graph wouldn't use the `.env.*` files as the environment variables would already be specified.

apps/hash-external-services/docker-compose.yml
TimDiekmann: Note, that the `.prod` file specified this before as well

PR #3098, opened 2023-09-05
H-718: Change Subgraph traversal logic to check for permissions to see entities

TimDiekmann:

Regardless of the specific implementation of authorization the interface for entities should not change. It's expected, that the API can return information if an actor has the permission to view an entity.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest.rs
@@ -18,9 +18,8 @@ mod entity;
 mod entity_type;
 mod property_type;
 
-use std::{fs, io, sync::Arc};
+use std::{fs, future::Future, io, sync::Arc};
 
-use async_trait::async_trait;
 use axum::{
     extract::Path,
     http::StatusCode,
@@ -81,17 +80,15 @@ use crate::{
     },
 };
 
-#[async_trait]
 pub trait RestApiStore: Store + TypeFetcher {
-    async fn load_external_type(
+    fn load_external_type(
         &mut self,
         domain_validator: &DomainValidator,
         reference: OntologyTypeReference<'_>,
         actor_id: RecordCreatedById,
-    ) -> Result<OntologyElementMetadata, StatusCode>;
+    ) -> impl Future<Output = Result<OntologyElementMetadata, StatusCode>> + Send;
 }
 
-#[async_trait]
 impl<S> RestApiStore for S
 where
     S: Store + TypeFetcher + Send,
--- a/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
@@ -2,7 +2,6 @@ mod read;
 
 use std::collections::HashMap;
 
-use async_trait::async_trait;
 use error_stack::{Report, Result, ResultExt};
 use graph_types::{
     knowledge::{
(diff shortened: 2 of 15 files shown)


apps/hash-graph/lib/authorization/src/lib.rs
```diff
@@ -1,8 +1,78 @@
...
+            ),
+            CheckError,
+        >,
+    > + Send;
+}
+
+#[derive(Debug, Default)]
+pub struct NoAuthorization;
+
+// #[async_trait]
```
indietyp, line 50:
```suggestion
```

apps/hash-graph/lib/graph/src/store/postgres/knowledge/entity.rs
```diff
@@ -503,8 +554,12 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
             .collect())
     }
 
-    #[tracing::instrument(level = "info", skip(self))]
-    async fn get_entity(&self, query: &StructuralQuery<Entity>) -> Result<Subgraph, QueryError> {
+    // #[tracing::instrument(level = "info", skip(self, authorization_api))]
```
indietyp, line 557: uncomment or remove

apps/hash-graph/lib/authorization/src/lib.rs
```diff
@@ -1,8 +1,78 @@
...
+use crate::{
+    backend::CheckError,
+    zanzibar::{Consistency, Zookie},
+};
 
 pub mod backend;
 
 pub mod zanzibar;
+
+pub trait AuthorizationApi {
```
indietyp, line 23: (not strictly a review) just for anyone wondering. The earlier `AuthorizationApi` has been renamed into `ZanzibarBackend`, and a new `AuthorizationApi` for use in the REST-API has been created.

apps/hash-graph/lib/graph/src/api/rest.rs
```diff
@@ -194,6 +200,7 @@ where
     // The `/api-doc` endpoints are nested as we don't want any layers or handlers for the api-doc
     merged_routes
         .layer(Extension(dependencies.store))
+        .layer(Extension(dependencies.authorization_api))
```
indietyp, line 203: I've seen this a couple of times, but axum now recommends the use of a single `State` using `with_state` instead of an extension (https://docs.rs/axum/latest/axum/index.html#sharing-state-with-handlers) which also allows for substrates: https://docs.rs/axum/latest/axum/extract/struct.State.html#substates.

It might be worth to switch to a `State` object instead.

indietyp (changes requested):

apps/hash-graph/lib/graph/src/api/rest.rs
TimDiekmann: I agree that we should switch but I'm not going to do this as a drive-by. There are a lot things which require improvements but if I would do this every time I touch the specific code it would take much longer. Thank you for pointing this out, though!

indietyp: Yea might have come across this way as a drive-by but was more like a note. Sorry!

indietyp (dismissed):

indietyp (approved):

TimDiekmann:
Reset the branch to pre-merge as the other changes have to be reverted. See #3101

TimDiekmann:
While resetting I pushed the wrong branch. Will try to reset the branch now.

PR #3099, opened 2023-09-05
H-721: Replace almost all `#[async_trait]` usages with native `async trait`

TimDiekmann:

On `nightly` it's possible to replace the amazing `#[async_trait]` by the native implementation.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Replacing `Read` - the last remaining `trait` using `#[async_trait]` - is left as an exercise for the reader/reviewer.


TimDiekmann:
I'm pretty sure only a single lifetime is required but as the trait is rather complex and implemented quite often it's really time consuming.

PR #3101, opened 2023-09-06
Revert "H-721: Replace almost all `#[async_trait]` usages with native `async trait` (#3099)"

TimDiekmann:
This reverts commit 52dfa7ed50426e6c7799fbf771cc6fecb43cb018.

It's amazing that `async trait` is there, now, but we encounter a huge impact on compile times. A build on my machine built in 15 seconds previously, with H-721 merged, takes 1m 6s.


indietyp (approved):
"I am inevitable" - the `Box` type right now.

PR #3103, opened 2023-09-06
H-719: Require the `X-Authenticated-User-Actor-Id` Header to be present in the Graph API endpoints

TimDiekmann:

To authorize a user the user information has to be passed to the graph.

## 🔍 What does this change?

- Require the header on _all_ REST endpoints
- Remove the `actorId` body parameter and use the value of the header instead
- Generate the OpenAPI specs from the changes and adjust both sides, the caller side and the server side.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The user could pass any account id, however, the Graph service is currently expected to not be accessible to end users. In the future we will add authentication to the Graph as well.


apps/hash-api/src/graph/index.ts
```diff
@@ -20,6 +21,9 @@ import {
   ensureSystemUserExists,
 } from "./system-user";
 
+export const publicUserAccountId: AccountId =
+  "00000000-0000-0000-0000-000000000000" as AccountId;
```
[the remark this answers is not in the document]
TimDiekmann: This is a well-defined UUID constant.

apps/hash-api/src/graphql/resolvers/ontology/data-type.ts
```diff
@@ -41,17 +44,17 @@ export const getDataType: ResolverFn<
...
-    dataTypeId,
-    actorId: user ? user.accountId : systemUserAccountId,
-    /** @todo - make these configurable once non-primitive data types are a thing https://app.asana.com/0/1200211978612931/1202464168422955/f */
-    graphResolveDepths: {
-      ...zeroedGraphResolveDepths,
-      constrainsValuesOn,
+> = async (_, { dataTypeId, constrainsValuesOn }, { dataSources, user }) =>
+  getDataTypeSubgraphById(
+    dataSourcesToImpureGraphContext(dataSources),
+    { actorId: user?.accountId ?? publicUserAccountId },
```
CiaranMn, line 50: I don't think we want to have to remember to do this for any request that can be made by a public person – that means we have two places where we are tracking 'this action is public' instead of one central permission store.

I suggest we add `actorId` to the context in `create-apollo-server.ts`:
```ts
  return new ApolloServer({
    schema: combinedSchema,
    dataSources: getDataSources,
    context: (ctx): Omit<GraphQLContext, "dataSources"> => ({
      ...ctx,
      actorId: ctx.req.user?.accountId ?? publicUserAccountId,
      user: ctx.req.user,
      // etc
```
then we can use `actorId` from the context in all the resolvers, rather than having to make any decisions about which should have public and which not.

CiaranMn (commented):
One proposed change for discussion which I think will save us work / permissions becoming out of sync

apps/hash-api/src/graphql/resolvers/ontology/data-type.ts
TimDiekmann: I think that's a good point. Will do the changes. thank you!

apps/hash-api/src/graphql/context.ts
```diff
@@ -1,13 +1,21 @@
...
 import { EmailTransporter } from "../email/transporters";
 import { GraphApi } from "../graph";
 import { User } from "../graph/knowledge/system-types/user";
 import { UploadableStorageProvider } from "../storage";
 import { TemporalClient } from "../temporal";
-import { VaultClient } from "../vault/index";
+import { VaultClient } from "../vault";
+
+export const publicUserAccountId: AccountId =
+  "00000000-0000-0000-0000-000000000000" as AccountId;
```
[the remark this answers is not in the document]
TimDiekmann: This is a well-defined UUID constant.

apps/hash-api/src/graphql/resolvers/ontology/data-type.ts
TimDiekmann: I moved the context in https://github.com/hashintel/hash/pull/3103/commits/30f4213dafb4318a229985fe0d83d7dbe3fe7656

CiaranMn: Great stuff!

CiaranMn (approved):

CiaranMn (approved):

PR #3108, opened 2023-09-07
H-587: Sidebar styling improvements and fixes

benwerner01:

This PR makes improves the styling and makes some fixes to the sidebar of the HASH app.

Additional ticket addressed in this PR: [Make font weight of types match those of pages, and make whitespace between them and "View all" equal between both.](https://linear.app/hash/issue/H-738/make-font-weight-of-types-match-those-of-pages-and-make-whitespace)

## 🔍 What does this change?

- changes the “Add subpages, delete, duplicate and more” tooltip to simply read “Options”
- always display the right chevron of pages in the tree, displaying "No sub-pages inside" when un-collapsing a page which has no children
- prevents the emoji picker from repositioning when it is used on non-top-level pages
- fixes a bug which set incorrect fractional indexes for top-level pages when they were repositioned. Previously if you rearranged a top-level page it would become `a0` no matter where it was being placed, meaning that you could end up in a state where multiple top-level pages had the same fractional index leading to their order to change on re-renders.
- adjusts the font-weight of pages in the page tree to be consistent with types in the sidebar
- adjusts the spacing between the page tree and the "View all pages" button to be consistent with the spacing before the "View all types" button

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- [Improve drag-and-drop behaviour in HASH App sidebar](https://linear.app/hash/issue/H-740/improve-drag-and-drop-behaviour-in-hash-app-sidebar)

## 🛡 What tests cover this?

None

## ❓ How to test this?

Go to the deployment, and try out the affected functionality in the sidebar. Note that the fractional indexes of top-level pages may be incorrect if you've tried to previously rearrange them when the bug was unaddressed. It may be possible to fix this by rearranging all top-level pages at least once.

## 📹 Demo

Updated tooltip:
image

Un-collapsible pages with no children:
image

### DIFF
--- a/apps/hash-frontend/src/components/emoji-picker/emoji-picker.tsx
+++ b/apps/hash-frontend/src/components/emoji-picker/emoji-picker.tsx
@@ -28,11 +28,11 @@ export const EmojiPicker = ({
       {...bindPopover(popupState)}
       anchorOrigin={{
         vertical: "bottom",
-        horizontal: "center",
+        horizontal: "left",
       }}
       transformOrigin={{
         vertical: "top",
-        horizontal: "center",
+        horizontal: "left",
       }}
       elevation={4}
       sx={{ mt: 1 }}
--- a/apps/hash-frontend/src/components/page-icon.tsx
+++ b/apps/hash-frontend/src/components/page-icon.tsx
@@ -28,6 +28,7 @@ export const PageIcon = ({ icon, size = "medium", sx = [] }: PageIconProps) => {
           width: sizes.container,
           height: sizes.container,
           fontSize: sizes.font,
+          fontFamily: "auto",
           display: "flex",
           justifyContent: "center",
           alignItems: "center",
--- a/apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list.tsx
+++ b/apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list.tsx
@@ -141,7 +141,6 @@ export const AccountEntityTypeList: FunctionComponent<
               alignItems: "center",
               mx: 0.5,
               minHeight: 36,
-              my: 0.25,
               borderRadius: "4px",
               // "&:hover": {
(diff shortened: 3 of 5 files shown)


nonparibus:
@benwerner01 Small bug: the `:relaxed:` emoji doesn't appear properly in the sidebar, when chosen, even though it successfully renders in the emoji selector. Any idea why?

CiaranMn (commented):
Style changes look good – a couple of things:

1. On @nonparibus 's comment
> @benwerner01 Small bug: the `:relaxed:` emoji doesn't appear properly in the sidebar, when chosen, even though it successfully renders in the emoji selector. Any idea why?

The font-family needs setting on that element, then it will work

2. Re-ordering is still broken when attempting to move something from a different depth, e.g. move something from a sub-page to the top level in a position other than first, and it will jump to first.

benwerner01:
> The font-family needs setting on that element, then it will work

@CiaranMn thanks for the suggestion, that indeed seems to have fixed it. Have addressed in https://github.com/hashintel/hash/pull/3108/commits/8eca932229252500c73f4d98db9bbda021356a2a.

> Re-ordering is still broken when attempting to move something from a different depth, e.g. move something from a sub-page to the top level in a position other than first, and it will jump to first (this seems to be true of all moves between different depths, not just 'to root')

Thanks for spotting this, I think I've fixed this in https://github.com/hashintel/hash/pull/3108/commits/bd917c6b051bd59652f54f42ff83085cdeb6823d and taken the opportunity to clean up the `handleDragEnd` a bit. I'm still not entirely happy with it, but don't think it's worth over-investing before [this follow-up ticket](https://linear.app/hash/issue/H-740/improve-drag-and-drop-behaviour-in-hash-app-sidebar) in which its functionality may need to change further.

https://github.com/hashintel/hash/assets/42802102/c8f6495c-b1ff-42d4-9ea4-f1328872f159

PR #3121, opened 2023-09-09
H-744: Adjust database to store accounts and account groups

TimDiekmann:

In order to provide the concept of account groups the groups has to be stored in the database.

## 🔍 What does this change?

- Add a `owners` table - All accounts and account groups are "owners"
- Change `accounts` table to reference `owners` table
- Add `account_groups` table
- Add API endpoint to create an account group
- TBD: Adjust Graph to take account groups into account (follow-up)
- TBD: Adjust snapshot logic (partially done, rest in follow-up)
- Drive-by: Fix Postgres healthcheck spamming the postgres log

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/graph/src/snapshot/account.rs
TimDiekmann, line 1: This file was moved, not deleted, but git didn't recognize that.

apps/hash-graph/lib/graph/src/snapshot/account/channel.rs
TimDiekmann, line 1: This file was moved, not deleted, but git didn't recognize that.

apps/hash-graph/lib/graph/src/snapshot/account/table.rs
TimDiekmann, line 1: This file was moved, not deleted, but git didn't recognize that.

apps/hash-graph/lib/graph/src/snapshot/entity/channel.rs
```diff
@@ -64,15 +63,15 @@ impl Sink<EntitySnapshotRecord> for EntitySender {
...
+        self.owner
+            .start_send_unpin(OwnerId::Account(
                 entity
                     .metadata
-                    .record_id
-                    .entity_id
-                    .owned_by_id
+                    .custom
+                    .provenance
+                    .record_created_by_id
```
CiaranMn, line 72: I'm not familiar with the context of this function – here we've switched from taking an `owned_by_id` (a user or an org) to the record creator (a user, not necessarily the owner) – can you confirm that's correct for its usage?

TimDiekmann: This function is called when restoring a snapshot. The snapshot currently does not export accounts but instead creates account IDs in the Graph based on required accounts in the snapshots.

The previous behavior was to use the `owned_by_id` to create accounts that were referred to by entities. This is no longer possible as `owned_by_id` could be both, an account id or an account group id.  Ideally, this snippet would be removed entirely, but as long as we don't make accounts and account groups a part of the snapshot (I'm currently doing exactly this) there is no other easy way to seed the Graph with accounts. For the time being, I use the `record_created_by_id` to seed accounts and tested that with all available test sets I have.

I also left this snippet in to maintain compatibility between snapshots in the short term to be able to import older snapshots. With the future PR, which will export accounts and account groups as well, we can remove this.

CiaranMn: Understood, thank you.

CiaranMn (approved):

PR #3122, opened 2023-09-09
H-745: Apply changes to python packages

indietyp:

This applies the recent change of #3103 to the `hash-graph-client` (and subsequently `hash-graph-sdk`) packages and _significantly_ improves the ergonomics of setting and using an actor.

As a drive-by, an issue with the `turbo.json` that was discovered in #3119 has been fixed by requiring that `lint` is a dependency of `build`.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graph-client/python/graph_client/__init__.py
+++ b/libs/@local/hash-graph-client/python/graph_client/__init__.py
@@ -72,7 +72,10 @@ async def _send_request(
     return response_t.model_validate(json, strict=False)
 
 
-def _assert_not_none(value: T | None) -> T:
+U = TypeVar("U")
+
+
+def _assert_not_none(value: U | None) -> U:
     """Assert that the value is not None."""
     if value is None:
         msg = "value cannot be None"
--- a/libs/@local/hash-graph-client/python/graph_client/__init__.py
+++ b/libs/@local/hash-graph-client/python/graph_client/__init__.py
@@ -98,7 +98,10 @@ class GraphClient:
         return _assert_not_none(actor)
 
     async def query_entity_types(
-        self, query: EntityTypeStructuralQuery, *, actor: UUID | None = None
+        self,
+        query: EntityTypeStructuralQuery,
+        *,
+        actor: UUID | None = None,
     ) -> Subgraph:
         """Query the HASH API for entity types."""
         endpoint = self.base / "entity-types" / "query"
@@ -231,7 +234,10 @@ class GraphClient:
         )
 
     async def query_data_types(
-        self, query: DataTypeStructuralQuery, *, actor: UUID | None = None
+        self,
+        query: DataTypeStructuralQuery,
+        *,
+        actor: UUID | None = None,
     ) -> Subgraph:
         """Query the HASH API for data types."""
         endpoint = self.base / "data-types" / "query"
@@ -296,7 +302,10 @@ class GraphClient:
         )
 
     async def query_entities(
-        self, query: EntityStructuralQuery, *, actor: UUID | None = None
+        self,
+        query: EntityStructuralQuery,
+        *,
+        actor: UUID | None = None,
     ) -> Subgraph:
         """Query the HASH API for entities."""
         endpoint = self.base / "entities" / "query"
@@ -310,7 +319,10 @@ class GraphClient:
         )
 
     async def create_entity(
-        self, request: CreateEntityRequest, *, actor: UUID | None = None
+        self,
+        request: CreateEntityRequest,
+        *,
+        actor: UUID | None = None,
     ) -> EntityMetadata:
         """Create an entity."""
         endpoint = self.base / "entities"
@@ -324,7 +336,10 @@ class GraphClient:
         )
 
     async def update_entity(
-        self, request: UpdateEntityRequest, *, actor: UUID | None = None
+        self,
+        request: UpdateEntityRequest,
+        *,
+        actor: UUID | None = None,
     ) -> EntityMetadata:
         """Update an entity."""
         endpoint = self.base / "entities"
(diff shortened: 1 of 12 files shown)


libs/@local/hash-graph-sdk/python/turbo.json
```diff
@@ -19,7 +19,7 @@
       "dependsOn": ["codegen:blocking", "codegen:filter"]
     },
     "build": {
-      "dependsOn": ["^build", "codegen"],
+      "dependsOn": ["^build", "codegen", "lint:mypy"],
```
TimDiekmann, line 22: It's the other way around. It's not required to lint a package to build it but the linting may have a different output if the dependencies failed.

TimDiekmann (changes requested):
The code changes look good but the new turbo config causes issues. With that config the package is linted every-time a downstream package is being built. A package can be built without linting it but the linting should run if a dependency changed. For now we can use `^build` for this.

libs/@local/hash-graph-sdk/python/turbo.json
indietyp: mh yes, well, my thought about this was/is the following: `mypy` is the build step for Python as it ensures that the code will not be malformed (this is also why only `lint:mypy` is included, not `lint:black` or `lint:ruff`) as mypy is our best chance to verify that something _will_ run in production, even if the build step (which doesn't exist) successfully runs. Hope that makes sense?

TimDiekmann: I get your point, but this is precisely what linting is about. There is no point in lint every project in your dependency chain each time, it is required only once. Other packages require a package to be build but they don't care how it ensures it's correctness.

Saying it from another perspective: you wouldn't want to require `clippy` to pass to run your app. Especially you don't want to require upstream linting to pass.

TimDiekmann: > `mypy` is the build step for Python 

No, it's a linter. Building would be packing it into a wheel.

indietyp: That's not what I meant, and you know it... and I wouldn't put `clippy` on the same level as `mypy`, `clippy` is more in line with `ruff`. **anyway** this doesn't really matter and as requested I have inverted the relationship in [f471b3e](https://github.com/hashintel/hash/pull/3122/commits/f471b3ea6d8e8e30d9b7e438f91bebf398b8f82d).

`mypy` is a lot more in line with `tsc` from typescript and even though one can only transpile, `hash-ai-worker-ts` for example chooses to type-check on build as well ([see here](https://github.com/hashintel/hash/blob/main/apps/hash-ai-worker-ts/package.json))

TimDiekmann: That's true, `clippy` is probably more comparable to `ruff`, yet, `mypy` (just like `tsc`) is _not_ required to build a package. We don't use the build-step for many things at the moment, but we have defined the dependencies via it. If I run `yarn lint:eslint` for example, I don't want `yarn lint:mypy` to be called in a few packages as well, because it's not needed for that.

TimDiekmann (dismissed):

TimDiekmann:
Great thank you! Linting is still failing but the changes so far look good. Just re-request a review when needed 🙂

libs/@local/hash-graph-sdk/python/turbo.json
indietyp: Yes, `mypy` is not required just like `tsc`, but often just a sanity check^^ in any case if in future the `build` step gets to be more important one should maybe think about unifying this. Either by removing `tsc` (and just transpiling) which would be a lot faster, or including `mypy`.

libs/@local/hash-graph-client/python/graph_client/__init__.py
```diff
@@ -67,29 +72,56 @@ async def _send_request(
...
+
+
 class GraphClient:
     """Low-level implementation of the client for the HASH API."""
 
     base: URL
+    actor: UUID | None
 
-    def __init__(self, base: URL) -> None:
+    def __init__(self, base: URL, *, actor: UUID | None = None) -> None:
```
TimDiekmann, line 93: Just so you're aware of it: In `hash-api` we always set the `actor` on creation. I like the idea of being able to override it when calling it, I just don't know a good reason why this is useful (for `actor`).

TimDiekmann (approved):

libs/@local/hash-graph-client/python/graph_client/__init__.py
indietyp: This is to support the two modes of operation: ad-hoc `actor` (per request) and persistent `actor`.

Both have their use cases. Let's say you have a long-running session with a single actor; in that case, you can specify the actor when creating (or setting it with the property setter). Another application might favor putting it on a per-request basis but requires the same client (the graph type protocol, for example). Both are valid use cases, so I included both in a simple API, enabling the overrides.

To bring it into some more practical terms: setting via `__init__` is something a script might do, while overrides per request basis would be done in e.g. a REST API that reuses the same client.

TimDiekmann: Yeah, I get the two modes, but I don't see the ad-hoc mode as very useful. e.g. a REST API should better create a new client per session (= tied to a user) than having a global variable to call it every time. Passing the actor in an ad-hoc application is error-prone as it's not required and might be forgotten easily. 🤷🏻‍♂️

indietyp: To be fair, I thought you found the non-ad-hoc mode problematic :P tbf I mostly added it to accommodate the protocol used in the `hash-graph-types` which needs the ad-hoc mode. (I initially planned to have no ad-hoc mode).

TimDiekmann: If you'd also prefer the non-ad-hoc mode let's try to remove the ad-hoc mode at some point 🙂

PR #3127, opened 2023-09-11
H-748: Export accounts and account groups in snapshots

TimDiekmann:

Previously, accounts were created on demand. This has the disadvantage, that after a snapshot is restored there might be fewer accounts than before. Also, with the concept of account groups, it's no longer possible to create accounts or account groups based on the `ownedById` as this could be either.

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/graph/src/snapshot/ontology/metadata/channel.rs
```diff
@@ -86,12 +83,6 @@ impl Sink<(Uuid, OntologyElementMetadata)> for OntologyTypeMetadataSender {
             }
         };
 
-        self.owner
-            .start_send_unpin(OwnerId::Account(
-                provenance.record_created_by_id.as_account_id(),
-            ))
-            .attach_printable("could not send account")?;
-
```
TimDiekmann, line 94: See https://github.com/hashintel/hash/pull/3121#discussion_r1321621388

CiaranMn (approved):

PR #3136, opened 2023-09-12
H-758: Don't require SpiceDB database parameters to be passed in postgres init script

TimDiekmann:

We don't need the database everywhere (e.g. in test and currently in prod). It should be possible to ignore it.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Try to start the different external service configurations locally


indietyp (approved):
Thank you! 🎉

PR #3147, opened 2023-09-13
H-588: Style cleanup in the 'view all' pages/types/entities tables

benwerner01:

This PR makes several styling and related fixes to the "view all" pages/entities/types tables.

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

View the deployment and visit the `/types`, `/pages` and `/entities` pages.


benwerner01:
@nonparibus with the types table now having global types included by default, I'm wondering whether we want to add an empty state for the table when a user hasn't created any types yet? If you agree maybe worth adding a low priority follow-up design ticket for what the empty states should look like.

image

apps/hash-frontend/src/components/grid/grid.tsx
```diff
@@ -51,6 +51,14 @@ export type GridProps<T> = Omit<
   sortRows?: (rows: T, sort: ColumnSort<string>) => T;
 };
 
+const gridHeaderHeight = 42;
+
+export const gridHeaderHeightWithBorder = gridHeaderHeight + 1;
+
+export const gridRowHeight = 42;
+
+export const gridHorizontalScrollbarHeight = 17;
```
CiaranMn, line 60: This will vary across browsers, themes etc, but worst case scenario is that a table without enough data to fill the screen is not quite tall enough to fit the rows, causing an unnecessary vertical scrollbar, so I think it's fine to keep this assumption

CiaranMn (commented):
Code looks good, will leave it to @nonparibus to approve the visuals

apps/hash-frontend/src/components/grid/grid.tsx
benwerner01: That's a fair point, I could increase it by a few pixels to preemptively account for that but that will create a gap between the last row and the scrollbar, so not sure it's a great trade-off.

CiaranMn: I don't think it's worth it. I'd expect entities tables will have enough rows to fill the max vertical height, in which case this number is irrelevant anyway.

nonparibus:
> @nonparibus with the types table now having global types excluded by default, I'm wondering whether we want to add an empty state for the table when a user hasn't created any types yet? If you agree maybe worth adding a low priority follow-up design ticket for what the empty states should look like.
> 
> image

@benwerner01 Agreed. Captured in H-772.

nonparibus (dismissed):

CiaranMn (approved):

TimDiekmann (approved):

PR #3148, opened 2023-09-13
H-767: Fix bugs in type editor form when moving between types

CiaranMn:

This PR fixes a couple of bugs related to moving from an existing entity type to a draft entity type without leaving the type editor page, which has only been possible since the introduction of the 'Extend type' button
1. Resets the edit bar state when moving between types (fixes click extend type, edit bar shows incorrect version)
2. Update `completedRef.current` when tracking which type we last loaded to take account of moving from a type to a draft and then back to the same type (fixes click extend type, click parent in sidebar, form state does not reset)
3. Wait for entity types to load before checking ancestors (fixes click extend type, then refresh page)
4. Add hacky timeout when resetting form state with a draft (fixes click extend type where the parent has at least one of its own properties, then click 'publish' – will crash, see longer comment). Not happy with this but lost a lot of time looking for a better solution, might be being dumb.

It also

5. Fixes a bug which was introduced by overly strict validation in #3093, which prevented publishing a type if the user hadn't touched the form. Users may wish to publish _new_ types (at v1) without adding any properties or links.
6. Updates `react-hook-form` (did this trying to debug the above, might as well keep it)
7. Correct an incorrect variable name

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. See bug recreation steps in description.

### DIFF
--- a/apps/hash-frontend/package.json
+++ b/apps/hash-frontend/package.json
@@ -85,7 +85,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-dropzone": "14.2.3",
-    "react-hook-form": "7.45.4",
+    "react-hook-form": "7.46.1",
     "react-responsive-carousel": "3.2.23",
     "react-transition-group": "4.4.5",
     "rooks": "7.14.1",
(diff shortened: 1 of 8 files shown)


benwerner01 (approved):
Thanks Ciaran, the fixes are working for me. I'm also not sure why react form state is not resetting properly without the timeout, strange that it won't work in `useEffect` either.

PR #3153, opened 2023-09-13
H-760: Check `labelProperty`s when generating an entity's label

CiaranMn:

We have a custom `labelProperty` which can be set on entity types, which indicates which property should be used when labelling an entity in the UI. 

This PR starts using it to label entities, where it is defined on the entity's type or on its type's parents (preferring closer ones).

I've also simplified how we check for fallbacks based on some property names that should make for reasonable labels.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- H-142: Allow setting a type's label property in the UI (currently only configurable via the API/db)

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1. Try setting a `labelProperty` on a type or its parents (directly in the db locally is easiest), then creating an entity with that property value

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/919ea905-dd9a-48e5-ae74-5293fb77f6e4

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/shared/use-draft-entity-subgraph.ts
+++ b/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/shared/use-draft-entity-subgraph.ts
@@ -18,7 +18,7 @@ export const useDraftEntitySubgraph = (
   Dispatch<SetStateAction<Subgraph<EntityRootType> | undefined>>,
   boolean,
 ] => {
-  const [loading, setLoading] = useState(false);
+  const [loading, setLoading] = useState(true);
   const [draftEntitySubgraph, setDraftEntitySubgraph] =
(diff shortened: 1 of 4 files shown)


apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/shared/use-draft-entity-subgraph.ts
```diff
@@ -18,7 +18,7 @@ export const useDraftEntitySubgraph = (
   Dispatch<SetStateAction<Subgraph<EntityRootType> | undefined>>,
   boolean,
 ] => {
-  const [loading, setLoading] = useState(false);
+  const [loading, setLoading] = useState(true);
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/shared/use-draft-entity-subgraph.ts:21]: drive-by: without this, you could get a flash of error state when viewing the `create-entity-page', since both `loading` and `draftEntitySubgraph` start off false.

PR #3164, opened 2023-09-14
H-720: Check for permissions when modifying an entity

TimDiekmann:

The Graph is expected to check for a permission before doing any modification. This adds checks for entities.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/authorization/src/api.rs
+++ b/apps/hash-graph/lib/authorization/src/api.rs
@@ -67,7 +67,6 @@ pub trait AuthorizationApi {
         consistency: Consistency<'_>,
     ) -> impl Future<Output = Result<CheckResponse, CheckError>> + Send;
 
-    #[must_use]
     fn can_view_entities(
         &self,
         actor: AccountId,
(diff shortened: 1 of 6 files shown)


apps/hash-graph/lib/authorization/src/api.rs
```diff
@@ -1,36 +1,95 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum VisibilityScope {
+    Public,
+    Account(AccountId),
+    AccountGroup(AccountGroupId),
+}
+
 pub trait AuthorizationApi {
-    async fn view_entity(
+    fn add_account_group_member(
```
CiaranMn, line 25: I assume these functions will take more parameters when fully implemented... e.g. the id of the member to add/remove

CiaranMn (approved):

apps/hash-graph/lib/authorization/src/api.rs
TimDiekmann: Yup, they already do in the next PR 🙂

PR #3166, opened 2023-09-14
H-373: Replace Next Ory API middleware with hash-api endpoint

benwerner01:

This PR removes the Next.js API endpoint for proxying Ory kratos requests to the public Ory Kratos API endpoint, and replaces it with a proxy in the HASH API to eliminate lambda cold start as a potential cause of app slowness.

## 🔗 Related links

- H-373

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Checkout the branch locally, and start the developer environment
2. Try to login, and check the user remains authenticated across refreshes. This tests both server-side Ory Kratos requests made by the FE (the `/whoami` request made to check if the user logged in), and browser side Ory Kratos requests (the `/login` request made to log the user in.

Unfortunately we don't have a way of testing this out in a staging environment, because it requires backend changes to be made and we don't have deployments for those in the staging environment. This is why the deployments attached to this PR will not work.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/c1864cb5-dd6d-4199-a544-7d722a19d4d8


apps/hash-api/src/index.ts
```diff
@@ -283,6 +285,55 @@ const main = async () => {
...
   });
 
+  /**
+   * Add a proxy for requests to the Ory Kratos public API, to be consumed
+   * by the frontend for authentication related requests made in the
+   * browser. Note that server-side frontend authentication requests
+   * can be sent the the Ory Kratos public URL directly, because the
+   * CORS requirements are not as strict as the one from the browser.
+   */
+  app.use("/auth/*", cors(CORS_CONFIG), (req, res, next) => {
```
benwerner01, line 295: @nonparibus @CiaranMn it was my understanding from the ticket we do not want to have the `ory` name in the path for requests made to Ory Kratos in the browser, so I'm using the `/auth/*` prefix for these routes instead of something like `/ory-kratos/*`. 
I do think this is slightly less intuitive for developers to understand that this is a direct proxy to the Ory Kratos API, and not auth endpoints that are maintained by us, but I suppose that's not a huge deal.

indietyp: As a side note, even though this is less intuitive, it also prevents vendor lock-in, as we no longer name by the service that provides it, but instead the function that the underlying proxied service provides.

CiaranMn (approved):
Cool, let's see if it helps!

PR #3167, opened 2023-09-14
H-725, H-726, H-771, H-782: Improve labeling and display of linked entities

CiaranMn:

Improvements to the display and functionality of the 'linked entities' row in the entity editor, i.e. a row which displays which entities are already linked to the entity, and allows you to select from an list of other entities you may add (the 'entity selector'):

- Fixed an issue with labelling entities where they appeared as options in the entity selector (H-771) and when they appeared as linked entities (H-782), where their type wasn't already associated with the entity in some way – the change is to pass the labelling function the subgraph the linked entity was found in, rather than the entity-being-edited's subgraph, which isn't guaranteed to contain the type of the linked entity.
- For linked entities with a file URL and an image mime type, show a little preview image. (H-725)
- Update chip colours for linked entities and their expected type
- In the entity selector, use a chip with their type title instead of showing the versioned URL (chips don't fully reflect the design yet, tracked in H-781)
- Added the type chip to the image preview popup (when you hover over an image in the entity selector) – there is additional metadata in the design for this, tracked in H-530
- I've updated the caching policy for all of queries to `cache-and-network`, which services cache content immediately while also refetching the query, which will update the cache if the response has changed. Our previous `no-cache` policy everywhere meant that dropdowns such as the entity selector were in 'loading' state every time they were opened, while waiting for the query response. 
- Fixed being able to click on the scrollbar in the entity selector (and other dropdowns)
- the type selector in the 'new entity page' was showing all versions of all types. I've made it show only the latest of each

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1. See demo

### Previously

https://github.com/hashintel/hash/assets/37743469/5d268716-d1d9-4cb8-a748-cc9cb34be118

### Now

https://github.com/hashintel/hash/assets/37743469/7c6335da-0d48-45c2-97ff-e7482b36cde6


libs/@hashintel/design-system/src/selector-autocomplete/selector-autocomplete-option.tsx
```diff
@@ -65,6 +73,8 @@ export const SelectorAutocompleteOption = ({
   const onMouseEnter = () => (imageUrl ? setShowPreviewPane(true) : null);
   const onMouseLeave = () => (imageUrl ? setShowPreviewPane(false) : null);
 
+  const typeTitle = slugToTitleCase(typeId.split("/").slice(-3, -2)[0]!);
```
benwerner01, line 76: I think this will only work for types created in HASH, if we're rendering external types here hosted on an arbitrary URL we would need to lookup the title in the type itself. Not how soon we'll encounter this, but maybe worth leaving a `@todo` or just rendering the URL using the `OntologyChip` component when `slugToTitleCase` returns `""` here.

benwerner01 (commented):
Thanks Ciaran.

When trying to use the entity selector locally in the links table of the entity editor I get a runtime error:
image

Is this just happening for me, or are you able to reproduce this from the latest version of the branch?

Also I spotted that the link type selector in the entity type editor is currently broken, but I appreciate that wasn't introduced in this PR so have created [a ticket](https://linear.app/hash/issue/H-786/fix-link-type-selector-in-the-entity-type-editor):
image
image

CiaranMn:
> When trying to use the entity selector locally in the links table of the entity editor I get a runtime error: image 
> Is this just happening for me, or are you able to reproduce this from the latest version of the branch?

This is also happening in `main` so I suspect it's related to the other crash – must be something to do with already merged recent changes to the selectors or underlying text fields.

benwerner01:
> This is also happening in main so I suspect it's related to the other crash – must be something to do with already merged recent changes to the selectors or underlying text fields.

@CiaranMn I've had a look and discovered this was introduced by my usage of `inputProps` in [this PR](https://github.com/hashintel/hash/pull/3155). Have found a fix, and opened https://github.com/hashintel/hash/pull/3183 - sorry for not catching this earlier!

benwerner01:
Thanks Ciaran, I can now see the changes locall yand they look good to me - one bug I spotted whilst testing is that the first time you upload a file in the HASH App, there is a runtime error. After a refresh further uploads work as expected. Not sure if this was introduced by the changes in this PR, but might be worth taking a look at whilst we're here?

https://github.com/hashintel/hash/assets/42802102/801cbd01-13aa-446a-abe1-2dcf40a96e25

libs/@hashintel/design-system/src/selector-autocomplete/selector-autocomplete-option.tsx
CiaranMn: Thanks, done in https://github.com/hashintel/hash/pull/3167/commits/e43ca7d4c7d46a877faef955b6e5e120bc765f9b

CiaranMn:
> Thanks Ciaran, I can now see the changes locall yand they look good to me - one bug I spotted whilst testing is that the first time you upload a file in the HASH App, there is a runtime error. After a refresh further uploads work as expected. Not sure if this was introduced by the changes in this PR, but might be worth taking a look at whilst we're here?

Thanks @benwerner01, fixed in https://github.com/hashintel/hash/pull/3167/commits/dfb788a2f614ba0a27ec10dc2731b8dc29c83ed9

It was introduced by this PR

benwerner01 (approved):
Working for me now, thanks Ciaran 👍

PR #3171, opened 2023-09-15
H-779(a): Connect authorization API to SpiceDB

TimDiekmann:

This is the first part out of two PRs to connect the Authorization API to SpiceDB. The connection is basically made but the actual functions are not propagated yet. The `hash-graph` binary is changed to connect to the provided SpiceDB server (the defaults are set in the `.env` file and should just work). Connecting to the Server requires a feature flag to be passed to the Graph which, by default, is not passed when building the docker images. This ensures that this will not throw unexpected errors as long as the permission system is under development.

## 🚫 Blocked by

- #3165 

## 🔍 What does this change?

- Provide CLI arguments to connect to SpiceDB (disabled by default, automatically passed by `.env`)
- Improve creation DX of a SpiceDB client
- Provide a `ZanzibarClient` struct which implements `AuthorizationAPI` and wraps a Zanzibar backend (such as the SpiceDB OpenAPI)

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/authorization/src/zanzibar/api.rs
TimDiekmann, line 1: Moved from `../zanzibar.rs`

apps/hash-graph/lib/authorization/src/zanzibar/types.rs
TimDiekmann, line 1: Moved from `../zanzibar.rs`

CiaranMn (approved):

CiaranMn (approved):

PR #3179, opened 2023-09-15
H-695: Provide an initial SpiceDB schema and the corresponding code representations

TimDiekmann:
## 🔗 Related links

- ...

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Try X
1. Confirm that Y


apps/hash-graph/lib/authorization/schemas/v1__initial_schema.zed
```diff
@@ -1,9 +1,27 @@
...
+	relation direct_admin: graph/account | graph/account_group#member
+	relation direct_membership: graph/account | graph/account_group#member | graph/account:*
+
+	permission add_admin = direct_admin
+	permission remove_admin = direct_admin
+	permission add_member = direct_admin
+	permission remove_member = direct_admin
+
+	permission member = direct_admin + direct_membership
+}
```
CiaranMn, line 13: I think it would be helpful to add some comments to this file to make it quicker to understand what some of the things mean – e.g. here are my questions that maybe future people will have and we can add comments to explain. It's probably easier to discuss this in person first.
1. Why `direct_x`? What is an indirect x?
2. We have a relation here, `graph/account_group#member`, saying that a `direct_admin` of an account group can be a `graph/account_group#member`, but that looks like a permission on the account group... it seems a bit circular... I thought `direct_admin` and `direct_membership` of an account group would each just be `graph/account` (i.e. users), and `graph/account_group#member` would only appear on other resources
3. What does `graph/account:*` mean? I looked in a couple of SpiceDB docs for `:*` but couldn't find it
4. `member` sounds more like a relation than a permission... maybe we can be more explicit here, e.g. `edit_resources`... maybe even split it further, `edit_types`, `edit_entities`, `view_types`, `view_entities`? Once we have more role levels in an org we'll definitely want roles that can edit the ontology, and roles that can only manage the knowledge data... might be better to separate from the outset?
5. I think the `relation` should be `member` rather than `membership` because that's how we describe the other relations (owner, viewer, admin, [member])

I'm assuming account group owners will be added in a future PR

CiaranMn (commented):
Thanks @TimDiekmann – I left a comment which I think will be easier to discuss in person first and then add the conclusion/advice to the thread (early thoughts/answer in text welcome also if you like)

apps/hash-graph/lib/authorization/schemas/v1__initial_schema.zed
TimDiekmann: I agree, let's add some comments in the file. Initial thoughts:

> Why `direct_x`? What is an indirect x?

I chose "direct" because if you would check if "Bob" is a "direct member" it would not return `true` if Bob is only a `direct_admin`.

> We have a relation here, graph/account_group#member, saying that a direct_admin of an account group can be a graph/account_group#member, but that looks like a permission on the account group...

It is circular, yes. That way it's possible to make a whole account group the admins of another account group. If you would for example have the organization "Big Org" and a role called "Admin" you can assign all "Admin"s to the "Big Org" as `direct_admins`.

> What does `graph/account:*` mean?

It's called [Wildcard](https://authzed.com/docs/reference/schema-lang#wildcards) and allows to make a relation public

> member sounds more like a relation than a permission...

I agree. We should probably rename `direct_membership` to `direct_member`. and rename this to `membership` or something else. 
An account group is not responsible for granting permissions such as `create_entity`, that is what namespaces are for. This is because, otherwise you would need the same permissions for both, accounts and account groups. Similary, you assign the permissions to e.g. view an entity by setting a relation between the members of an account group to that entity. An account group should only collect multiple accounts so you can assign the members of an account group as a whole, so e.g. you assign all members of the "Engineering" group to allow viewing entity "styleguide".

> Once we have more role levels in an org we'll definitely want roles that can edit the ontology, and roles that can only manage the knowledge data...

You're talking about roles here. A user is either a member of a group or it is not. A group may or may not have the permission to edit the ontology or not. If an organization needs finer-grained permissions for who can edit ontologies and who should allow knowledge data a role such as "ontology guys" or "knowledge people" should be added, which then will have the permissions assigned. I have not encoded any sub-groups, yet, but have a prototype which is working like this.

> I'm assuming account group owners will be added in a future PR

Yes, as part of H-816

TimDiekmann: Related: https://authzed.com/docs/reference/schema-lang#naming-permissions

CiaranMn (approved):
Thanks for the explanation in the comment thread @TimDiekmann, and for planning to add comments in a future PR – let's get this in

PR #3183, opened 2023-09-16
H-786: fix `TextField` ref not being passed to component correctly

benwerner01:

This PR fixes a bug that was introduced in https://github.com/hashintel/hash/pull/3155, where the usage of `inputProps` on the `TextField` component to disable 1Password on inputs by default prevented the `ref` from being passed to the component correctly for some reason.

This broke several inputs in the application which relied upon the `ref` being set, for example the link type selector in the entity type editor, the link selector in the entity editor, etc.

After some trial and error, I discovered that using the `slotProps` prop to achieve the override does not introduce this issue.

## 🔗 Related links

- H-786

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None.

## ❓ How to test this?

Try adding a link to an entity type, and confirm this is working again.

Go to `/settings/organizations/example-org/members` and inspect the "Add member" input to see the 1Password overlay is not being displayed to confirm the override still works.

## 📹 Demo

Adding link types to an entity type works again:

https://github.com/hashintel/hash/assets/42802102/bf5b6b8e-9004-4015-9c97-7532f5e3e252

The unwanted 1Password overlay is not displayed on the "Add member" input thanks to the override:

image

### DIFF
--- a/libs/@hashintel/design-system/src/text-field.tsx
+++ b/libs/@hashintel/design-system/src/text-field.tsx
@@ -163,7 +163,9 @@ export const TextField: FunctionComponent<TextFieldProps> = forwardRef(
           error,
           autoResize,
           multiline: textFieldProps.multiline,
-          inputProps: theme.components?.MuiInputBase?.defaultProps?.inputProps,
+          slotProps: {
+            input: theme.components?.MuiInputBase?.defaultProps?.inputProps,
+          },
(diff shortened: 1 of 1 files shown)


CiaranMn (approved):
Thanks for finding and fixing this!

PR #3191, opened 2023-09-18
H-656, H-657: Improve how we fetch users and orgs, show orgs' avatars

CiaranMn:

The feature change this PR introduces is that organization avatars are now shown on the org's profile page, and in the workspace switcher in the top left.

The bulk of the PR is a refactoring of how we fetch users and orgs, to reduce overfetching, by:
1. Exposing the Graph API's query syntax to the frontend so that we can make requests for an org or user with a given shortname, rather than having to fetch all orgs/users and filter them client-side
2. Introduces a hook to fetch a user or org by either shortname or accountId, and deletes various other hooks that were doing similar things
3. Refactors how users and orgs are loaded when the app loads to reduce request size:
   - for the 'all users' and 'all orgs' requests, stop fetching anything linked to them (we want to stop fetching all users and all orgs altogether, later)
   - for the authenticated user, only fetch the orgs they're a member of and no further at first, and fetch more information about the orgs (members, avatar) in a subsequent request, to speed up the initial user load and avoid bringing in too much of the graph in one request
4. Deletes unused context `RouteWorkspaceInfo`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues / next steps

1. Requests for a specific user or org by shortname are slow. H-802 tracks optimizations to be implemented this week
2. On the first page load, an org's avatar won't be available for a second and will load in afterwards. To avoid this we'd have to move org avatar fetching back to the server load – we can see how jarring this remains after further optimizations.

## 🛡 What tests cover this?

- Existing tests cover user login / construction as part of their flow
- No tests for org profile page / workspace switcher

## ❓ How to test this?

1. Set an org avatar via its profile page, and check it appears in the workspace switcher and on its public page (`/@[shortname]`)
3. Check that other features affected work, e.g. page creation, mentioning users within a page

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/5f848442-856a-470f-955c-4120e36a5ddc

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
+++ b/apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
@@ -9,6 +9,7 @@ import {
   splitEntityId,
   Subgraph,
 } from "@local/hash-subgraph";
+import { LinkEntity } from "@local/hash-subgraph/type-system-patch";
 import {
   ApolloError,
   ForbiddenError,
@@ -26,7 +27,6 @@ import { bpMultiFilterToGraphFilter } from "../../../../graph/knowledge/primitiv
 import {
   createLinkEntity,
   isEntityLinkEntity,
-  LinkEntity,
   updateLinkEntity,
 } from "../../../../graph/knowledge/primitive/link-entity";
 import { getEntityTypeById } from "../../../../graph/ontology/primitive/entity-type";
--- a/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/link-entity.test.ts
+++ b/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/link-entity.test.ts
@@ -12,7 +12,6 @@ import {
   createLinkEntity,
   getLinkEntityLeftEntity,
   getLinkEntityRightEntity,
-  LinkEntity,
 } from "@apps/hash-api/src/graph/knowledge/primitive/link-entity";
 import { User } from "@apps/hash-api/src/graph/knowledge/system-types/user";
 import { createEntityType } from "@apps/hash-api/src/graph/ontology/primitive/entity-type";
@@ -30,6 +29,7 @@ import {
   linkEntityTypeUrl,
   OwnedById,
 } from "@local/hash-subgraph";
+import { LinkEntity } from "@local/hash-subgraph/type-system-patch";
 
 import { resetGraph } from "../../../test-server";
 import { createTestImpureGraphContext, createTestUser } from "../../../util";


nonparibus:
@CiaranMn **Can we use the same edit/replace icon background that we use on the file upload interface more generally?** This way it'll work better if the avatar color happens to be the same or similar to the edit pencil's.

We can keep the icon as the edit pencil instead of the 'swap' icon used on the file interface, though.

apps/hash-frontend/src/components/hooks/use-orgs-with-links.ts
```diff
@@ -0,0 +1,82 @@
...
+    QueryEntitiesQuery,
+    QueryEntitiesQueryVariables
+  >(queryEntitiesQuery, {
+    variables: {
+      operation: {
+        multiFilter: {
+          filters: orgAccountGroupIds.map((entityId) => ({
+            field: ["metadata", "recordId", "uuid"],
+            operator: "EQUALS",
+            value: entityId,
```
benwerner01, line 41: Should we try and refer to `entityUuid` explicitly in our code? I think it could lead to confusion otherwise.
```suggestion
          filters: orgAccountGroupIds.map((entityUuid) => ({
            field: ["metadata", "recordId", "uuid"],
            operator: "EQUALS",
            value: entityUuid,
```

apps/hash-frontend/src/lib/user-and-org.ts
```diff
@@ -66,202 +82,216 @@ export const constructMinimalUser = (params: {
...
 
-export type User = MinimalUser & {
-  memberOf: Org[];
+export type Org = MinimalOrg & {
+  hasAvatar?: {
+    linkEntity: LinkEntity;
+    imageEntity: Image;
+  };
+  memberships: {
+    linkEntity: Entity<OrgMembershipProperties>;
```
benwerner01, line 91: Maybe we should update the `LinkEntity` type to accept a generic as well, so we can use it here:
```suggestion
    linkEntity: LinkEntity<OrgMembershipProperties>;
```

apps/hash-frontend/src/pages/settings/organizations/new/index.page/create-org-form.tsx
```diff
@@ -28,6 +30,7 @@ export const CreateOrgForm = () => {
     }
 
     await refetchUser();
+    void refetchOrgs();
```
benwerner01, line 33: Don't think it makes a huge difference, but I don't think there's a reason `refetchOrgs` has to happen after `refetchUser`:
```suggestion
    await Promise.all([refetchUser, refetchOrgs]);
```
Or if we don't want the redirect to be blocked:
```suggestion
    void refetchUser();
    void refetchOrgs();
```

benwerner01 (approved):
Thanks Ciaran, these are great improvements. Whilst testing I had one idea for an improvement, perhaps we can introduce a cancel button/icon of some sort when editing an existing image? This could make it more clear to the user that they haven't just deleted their avatar when clicking the `edit` button:

https://github.com/hashintel/hash/assets/42802102/e3e025d5-933a-4e7d-8ddf-05fb2607ab9e

But that shouldn't block merging, the PR looks good to me 👍

apps/hash-frontend/src/pages/settings/organizations/new/index.page/create-org-form.tsx
CiaranMn: Thanks, done in https://github.com/hashintel/hash/pull/3191/commits/ecc9450d89b7cfe08ef8497ad9db9649a4c56967

Since we just fetch the org on that page we don't rely on the user/full orgs list

apps/hash-frontend/src/components/hooks/use-orgs-with-links.ts
CiaranMn: Done in https://github.com/hashintel/hash/pull/3191/commits/677831f0abf9e7f3eebb8d6d4243515633702627

CiaranMn:
@nonparibus @benwerner01 I've updated the edit button and added a back button in https://github.com/hashintel/hash/pull/3191/commits/a03da6d4b9889d8ef63eddd8e0f6aecee94a367c


https://github.com/hashintel/hash/assets/37743469/b82cb64f-71fc-414f-9c61-0abb86713fab

benwerner01 (approved):
Thanks Ciaran, looks good to me 👍

PR #3193, opened 2023-09-18
H-759: Add bulk actions dropdown on entities and types tables

benwerner01:

This PR adds a bulk actions dropdown on the various entities and types tables in the HASH app, which currently allows for bulk archiving/un-archiving entity/property/link types, and page entities.

## 🔗 Related links

- H-759

## 🔍 What does this change?

- updates the `queryPropertyTypes` GQL resolver to accept `includeArchive` and `latestOnly` arguments, to be consistent with the `queryEntityTypes` GQL resolver and enable the FE to retrieve archived property types

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- [Allow bulk actions to be applied to a subset of the current selection, and update the tooltip to explain this](https://linear.app/hash/issue/H-824/allow-bulk-actions-to-be-applied-to-a-subset-of-the-current-selection)

## 🛡 What tests cover this?

None.

## ❓ How to test this?

1. Go to `/pages` page, and ensure pages can be bulk selected and archived/un-archived in the entities table
2. Go to the `/entities` page, and ensure entities can be bulk selected. Note that only pages can be archived, as general entity archival is not yet supported.
3. Go to the `/types` page and ensure entity types, link types and property types can be bulk selected and archived/un-archived. Note that data type archival is not yet supported, as we don't yet have custom data types.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/ebde0289-403f-4373-94a1-b8c631cdd052


apps/hash-frontend/src/shared/util.ts
benwerner01, line 1: I know having these very broad `util` files is not ideal, but the methods it contains are being used by unrelated components. @CiaranMn would you prefer this file to be split out into `type-utils.ts` and `archive-utils.ts` in the same directory?

apps/hash-frontend/src/shared/table-header/bulk-actions-dropdown.tsx
```diff
@@ -0,0 +1,298 @@
...
+  });
+
+  const refetchPropertyTypes = useFetchLatestPropertyTypes();
+
+  const [archivePropertyType] = useMutation<
+    ArchivePropertyTypeMutation,
+    ArchivePropertyTypeMutationVariables
+  >(archivePropertyTypeMutation, {
+    onCompleted: refetchPropertyTypes,
+  });
```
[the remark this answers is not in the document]
benwerner01: (Same applies for `queryEntitytypes`)

apps/hash-frontend/src/components/hooks/block-protocol-functions/ontology/use-block-protocol-query-property-types.ts
```diff
@@ -45,6 +45,7 @@ export const useBlockProtocolQueryPropertyTypes = (): {
           constrainsValuesOn: { outgoing: 255 },
           constrainsPropertiesOn: { outgoing: 255 },
           ...graphResolveDepths,
+          includeArchived: true,
```
CiaranMn, line 48: I think we want to make this an argument passed to the hook, rather than true – at the moment, archived property types will show up as options in the dropdowns for the entity type selector, which I don't think we want (otherwise there's no real point in archiving them). This hook can be called with `includeArchived: true` from the types table page.

apps/hash-frontend/src/shared/table-header/bulk-actions-dropdown.tsx
```diff
@@ -0,0 +1,298 @@
...
+  ]);
+
+  const menuItems = useMemo(() => {
+    return [
+      {
+        icon: <BoxArchiveIcon />,
+        label: "Archive",
+        onClick: archiveItem,
+        disabled: !canArchiveSelectedItems,
+        tooltipTitle: "Cannot archive one or more of the selected items.",
```
CiaranMn, line 220:
```suggestion
        tooltipTitle: canArchiveSelectedItems
          ? undefined
          : "Cannot un-archive one or more of the selected items.",
```

apps/hash-frontend/src/shared/table-header/bulk-actions-dropdown.tsx
```diff
@@ -0,0 +1,298 @@
...
+    onCompleted: refetchPropertyTypes,
+  });
+
+  const popupState = usePopupState({
+    variant: "popover",
+    popupId: "table-header-bulk-actions-dropdown-menu",
+  });
+
+  // Whether or not the selected items can be archived or un-archived
+  const itemsAreArchiveable = useMemo(
```
CiaranMn, line 105: I would also check here whether the items belong either to the user or an org they are a member of.

apps/hash-frontend/src/shared/table-header/bulk-actions-dropdown.tsx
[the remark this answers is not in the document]
CiaranMn: I have had problems with `refetchQueries` in the past, I can't remember what the solution was though. I think this is fine

apps/hash-frontend/src/shared/util.ts
CiaranMn: I would rather have more descriptive file names. In this case, it could be something like `is-of-type.ts` and `is-archived.ts`. I think `util` files should be a last resort

CiaranMn (commented):
Thanks @benwerner01 – this is working well for me, just a couple of suggestions for enhancements

apps/hash-frontend/src/shared/util.ts
benwerner01: Have done so in https://github.com/hashintel/hash/pull/3193/commits/32a9a2122ae3404f1900631c9451c0abf750e65b

apps/hash-frontend/src/shared/table-header/bulk-actions-dropdown.tsx
benwerner01: Good idea, have addressed in https://github.com/hashintel/hash/pull/3193/commits/fc766b9cc22219b34dfdcf06b798a6d312c9c74c

apps/hash-frontend/src/components/hooks/block-protocol-functions/ontology/use-block-protocol-query-property-types.ts
benwerner01: Currently this context is global, so this hook is being used in a single place to populate that context, where we now want to also get the archived property types.

But I appreciate your point regarding the property type selector now displaying archived property types, so I've updated the `useLatestPropertyTypesContextRequired` hook to accept an `includeArchived` parameter, which by default filters any archived entities from the response, in https://github.com/hashintel/hash/pull/3193/commits/fdfacaa9d8e64d36134cffdc1c837c634096e1f0. This should prevent archived property types from being included in things by default, and allow the property types table to continue accessing the archived property types via `includeArchived: true`.

I've also removed a `useLatestPropertyTypes` hook which was used previously, which didn't require the context to be set. However this didn't make much sense to me, as the context provider is located in `_app.page.tsx` so no matter where you require the context in the app it should be available. I've replaced all usages of `useLatestPropertyTypes` with `useLatestPropertyTypesContextRequired`.

CiaranMn: I would just call the hook `useLatestPropertyTypes`, we can add `useLatestPropertyTypesOptional` if we ever need it – I think `useLatestPropertyTypesContextRequired` is too verbose. I appreciate we have `useEntityTypesContextRequired` elsewhere but I don't think it's necessary, my default expectation would be that a hook relying on context would crash if the context wasn't available.

---

> Currently this context is global, so this hook is being used in a single place to populate that context, where we now want to also get the archived property types.

I don't think we need to fix this now but we will want to change this at some point.

The drawback of this is that now the global context for property types which loads on every page is fetching more types than it needs to, and if any else uses this hook they are also getting archived types by default. I think it would be better if the types page did its own querying – it doesn't need to go via the `useBlockProtocol` hook either, which adds a level of indirection around just using Apollo Client directly, e.g. 

```ts
  const { data, loading } = useQuery<
    QueryPropertyTypesQuery,
    QueryPropertyTypesQueryVariables
  >(queryPropertyTypesQuery, {
    fetchPolicy: "cache-and-network",
    variables: {
      ...zeroedGraphResolveDepths,
      constrainsValuesOn: { outgoing: 255 },
      constrainsPropertiesOn: { outgoing: 255 },
      includeArchived: true,
    },
  });

  const latestPropertyTypes = useMemo(
    () =>
      data
        ? Object.fromEntries(
            getRoots<PropertyTypeRootType>(
              data.queryPropertyTypes as Subgraph<PropertyTypeRootType>,
            ).map((propertyType) => {
              return [propertyType.schema.$id, propertyType];
            }),
          )
        : null,
    [data],
  );
```

This would have the advantage of also being able to vary `includeArchived` based on user choice on the type page, rather than filtering client-side, which we'll also want to do with other filters and pagination on this page once we need to worry about there being lots of data.

By using a `cache-and-network` policy we ensure that any new property types added during the session would be fetched when visiting the page again, so we don't have to worry about refetching if we change things elsewhere (unless we happen to be on this page somehow, which isn't possible at the moment).

CiaranMn (approved):
Thanks @benwerner01 – I've further thoughts on the hook, which needn't block merging.

nonparibus (changes requested):
**Style changes:**

- Remove the column divider from the right of the checkboxes. It doesn't need the border between it and the icons.
- We can decrease the space between the checkbox and the element on its right.
- We should increase the space to the left of the checkbox.
- We should use the new style checkboxes, ensuring they're smaller than those shown in the mockup video here ([Figma file](https://www.figma.com/file/YPwCWSIf124J6YkMaMpXFE/hash.ai?type=design&node-id=1371%3A8686&mode=design&t=m3EeESbdRpDC7TIy-1) - _internal link_).

apps/hash-frontend/src/components/hooks/block-protocol-functions/ontology/use-block-protocol-query-property-types.ts
benwerner01: Agreed, have renamed the hook to `useLatestPropertyTypes`, and have also added the `includeArchived` prop to  preparation for moving the context out of `_app.tsx` have added the `includeArchived` prop to the `LatestPropertyTypesContextProvider` component incase we want to move it out of `_app.page.tsx`, and not include archived property types, as part of https://github.com/hashintel/hash/pull/3193/commits/674ad4cb0fcc1ba69a4b5c5cdbdeead2ecf1a5f9

nonparibus:
@benwerner01 Can we also change ‘Un-Archive’ to ‘Unarchive’, if we haven't already?

benwerner01:
@nonparibus as discussed in standup yesterday, I've made the styling adjustments you've suggested which are straightforward to add without implementing the checkbox selection functionality from scratch ourselves (or patching the package to achieve similar results). Here's what I managed to do:

- remove the divider between the checkbox column and first column
- made the "Title" column sticky on the left (if I recall correctly you liked this from the example data grid component we looked at)
- removed the extra left padding of the title column
- fixed the copy in the dropdown to read "Unarchive" instead of "Un-archive"
- verified the checkbox color is the same as the one from Figma (`blue 70`)

What I could not figure out to do using the out-of-the-box checkbox implementation:

- increase the left padding of the checkbox column (neither the theme or the `drawCell` prop can influence the checkbox column padding as far as I can tell)
- use the custom checkbox styling from our figma

I've captured these inconsistencies in [this follow-up ticket](https://linear.app/hash/issue/H-836/implement-custom-checkbox-selection-column-for-glide-grid-tables).

Here's what it looks like now:

https://github.com/hashintel/hash/assets/42802102/93acac62-0c9d-42a1-bb71-a786decf1860

nonparibus:
@benwerner01 Has the column divider between subsequent cols (e.g. cols 3/4) not been accidentally removed? This should not have been removed. We only wanted the divider next to the checkboxes to be removed.

benwerner01:
@nonparibus you're right, sorry having missed that. Should've been fixed in https://github.com/hashintel/hash/pull/3193/commits/cbf1534fdd0e6d1f513b584ed4ee90de04e6081f

Here's what it looks like now:

https://github.com/hashintel/hash/assets/42802102/0afab813-f623-47f1-b044-3b66c05c36f3

nonparibus:
@benwerner01 Is there a way we can reduce the padding between the checkbox and the first-col further (when there are no expandable rows in a table, necessitating whitespace be preserved for placement of the expand/collapse chevron/caret? The gap between the checkbox and the first col still seems rather large to me.

benwerner01:
> @benwerner01 Is there a way we can reduce the padding between the checkbox and the first-col further (when there are no expandable rows in a table, necessitating whitespace be preserved for placement of the expand/collapse chevron/caret? The gap between the checkbox and the first col still seems rather large to me.

@nonparibus I've reduced the spacing further in https://github.com/hashintel/hash/pull/3193/commits/1e26ae630023f59a02d42053b851519a1cf24511

image

nonparibus:
@benwerner01 About half way between what it was and this (i.e. more similar to all of the other columns' distances from their left-dividers) would be ideal. Thank you.

benwerner01:
> 
@benwerner01 It's now a bit too tight. About half way between what it immediately prior to this and the last commit (to make it approximately the same as all of the other columns' distances from their left-dividers) would be ideal. Sorry for the lack of guiding design here, but this should do it. Thank you.

@nonparibus no worries, have adjusted it to as best I could to match the other columns, might be off by 1 pixel or so though as it's a manual adjustment.

image

nonparibus (dismissed):
Thanks for the changes!

CiaranMn (approved):

PR #3194, opened 2023-09-18
H-802: Skip checking type parents when looking up system entities by type

CiaranMn:

Where we query for entities and match against the type of the entity or of some other entity linked to them, the queries can currently be slow, mainly because of type inheritance.

This PR introduces a `generateVersionedUrlMatchingFilter` function that can create a filter with two ways of improving query performance:
1. Skip checking type parents (major improvement)
2. Split the type id we're matching on to instead match by its components `baseUrl` and `version`, which are top-level database fields, rather than the id which is a property nested in the type's jsonb schema. This has no discernible impact currently but may be better with more data (it may not if a query using indices on `baseUrl` and `version` perform the same as one using a GIN index on `schema -> $id`, and the number of joins involved has no meaningful impact)

It uses this function where we know we can skip checking parents, i.e. because we are dealing with system types that don't have parents.

The relevant queries are reduced (locally) from `400+ms` to around `10-20ms`

## 🚫 Blocked by

- [ ] Merging #3191 which this is stacked on

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Query performance even when parent types are checked to be improved 

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Run the app locally (backend change, the Vercel deployment will not reflect it)
2. Check response time for relevant queries in the frontend (or in the GraphQL playground), e.g. `hashInstance`, `structuralQueryEntities`

### DIFF
--- a/apps/hash-api/src/graph/knowledge/system-types/hash-instance.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/hash-instance.ts
@@ -1,5 +1,6 @@
 import {
   currentTimeInstantTemporalAxes,
+  generateVersionedUrlMatchingFilter,
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
 import {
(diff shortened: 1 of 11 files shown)


TimDiekmann (approved):

benwerner01 (dismissed):
Thanks Ciaran, changes look great and are working for me locally. Will also make writing these queries in the FE slightly easier.

benwerner01 (approved):

PR #3203, opened 2023-09-19
H-573: Fix frontend hot reloading

CiaranMn:

The frontend's hot reloading, i.e. reflecting changes made to components quickly without needing to refresh the page (aka HMR/hot module replacement, or [Fast Refresh](https://nextjs.org/docs/architecture/fast-refresh) in NextJS terms), was not working for components from `@hashintel/design-system`.

This is because components were re-exported by a file (`main.ts`) which also exported non-React components, meaning NextJS fell back to a full page refresh whenever a design system component was edited. This is annoying and slows down development, both because (1) a full page refresh takes much longer to see the change, and (2) you lose page state, meaning you have to recreate whatever state you were in if the change is state-dependent.

This PR fixes that by splitting out non-component exports from component exports from `@hashintel/design-system`.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Because it is about rearranging exports, `lint:tsc` and `lint:eslint` should flag any breakages.

## ❓ How to test this?

1. Try editing a design system component and see the frontend hot reload rather than fully refresh

#### Before

https://github.com/hashintel/hash/assets/37743469/9ef18353-ef43-45fe-ae23-404861dc1154

#### After

https://github.com/hashintel/hash/assets/37743469/e3f87ea7-0017-4a28-824c-f51673bed1cb


benwerner01 (commented):
Thanks Ciaran for tackling this, it'll definitely be a worthwhile improvement in the long-run!

It's working for me on some components, such as the `OntologyChip` component from your demo, but isn't working on others for me. For example the `TextField` component:

https://github.com/hashintel/hash/assets/42802102/04f3c343-92f8-4f4f-bb33-e90ed7657389

Do you have an idea why this might be? I think this PR is still worth merging as it is, just wanted to flag incase it would be a quick fix.

benwerner01 (dismissed):
Gotcha, thanks for the explanation 👍

benwerner01 (approved):

PR #3204, opened 2023-09-19
H-818: Add Endpoint to toggle entity visibility (public / private)

TimDiekmann:

Entities can be made public and private. This adds two functions to the `hash-api` to toggle this.

## 🚫 Blocked by
- #3189 
- #3179 
- #3190
- #3200

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/authorization/src/zanzibar/api.rs
```diff
@@ -168,7 +168,11 @@ where
         entity: EntityId,
     ) -> Result<Zookie<'static>, ModifyRelationError> {
         Ok(match scope {
-            VisibilityScope::Public => unimplemented!(),
+            VisibilityScope::Public => {
+                self.backend
+                    .create_relations([(entity, EntityRelation::DirectOwner, PublicAccess)])
+                    .await
+            }
```
CiaranMn, line 175: This looks like it's handling adding 'anyone' as an entity owner – is that right, and if so do we need it? Same applies to deleting an owner.

apps/hash-graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -77,6 +81,10 @@ impl RoutedResource for EntityResource {
             "/entities",
             Router::new()
                 .route("/", post(create_entity::<S, A>).put(update_entity::<S, A>))
+                .route(
+                    "/:entity_id/public",
+                    post(make_entity_public::<A>).delete(make_entity_private::<A>),
+                )
```
CiaranMn, line 87: I wonder why we have the update route as `PUT /entities` rather than `PUT /entities/:entity_id` – these routes are arguably inconsistent now because the `entity_id` is in the body in the first but the path in the second. FWIW my expectation would be how you've done it for the new route

CiaranMn (commented):
Thanks @TimDiekmann – a couple of questions

apps/hash-graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: I was wondering the exact same but didn't want to change this as a drive-by.

apps/hash-graph/lib/authorization/src/zanzibar/api.rs
TimDiekmann: Yes, it's just for consistency. Previously if this happened the code would have panicked, now it would fail with an error as the SpiceDB schema does not allow this. Ideally this cannot happen at all and I'm going to change this later but for now it's the easiest way to achieve this.

apps/hash-graph/lib/graph/src/api/rest/entity.rs
CiaranMn: Fair enough. Let's clean it up later then

apps/hash-graph/lib/authorization/src/zanzibar/api.rs
CiaranMn: Okay cool, as long as we change it later

CiaranMn (approved):

PR #3210, opened 2023-09-20
H-816: Add account group ownership and expose account group permission editing

TimDiekmann:
## 🌟 What is the purpose of this PR?

Adds the concept of account group ownership and exposes all functions to the Node API
## 🚫 Blocked by

- #3179 
- #3190
- #3200
- #3204 

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

PR #3211, opened 2023-09-20
H-827: Adjust Node API to use authentication for several cases

TimDiekmann:

With basic entity authorization implemented it's time to adjust a few Node API functions to properly fill the permissions. This includes making users, orgs, and the HASH-instance public and tweak various locations to use more correct actor ids.

This does not enable permissions in `dev` or `prod` by default but with permissions enabled most BE integration tests are passing. The missing parts are the subgraph tests (the snapshots do not know yet about permissions, will be done in H-669) and the instance admin tests (will be done as part of H-817).

## 🚫 Blocked by

- #3179 
- #3190
- #3200
- #3204 
- #3210 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
+++ b/tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
@@ -11,7 +11,11 @@ import {
   updateEntity,
 } from "@apps/hash-api/src/graph/knowledge/primitive/entity";
 import { getLinkEntityRightEntity } from "@apps/hash-api/src/graph/knowledge/primitive/link-entity";
-import { User } from "@apps/hash-api/src/graph/knowledge/system-types/user";
+import { Org } from "@apps/hash-api/src/graph/knowledge/system-types/org";
+import {
+  joinOrg,
+  User,
+} from "@apps/hash-api/src/graph/knowledge/system-types/user";
 import { createDataType } from "@apps/hash-api/src/graph/ontology/primitive/data-type";
 import { createEntityType } from "@apps/hash-api/src/graph/ontology/primitive/entity-type";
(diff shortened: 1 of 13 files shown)


tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
```diff
@@ -222,34 +238,40 @@ describe("Entity CRU", () => {
...
+              equal: [
+                { path: ["ownedById"] },
+                { parameter: testOrg.accountGroupId },
+              ],
+            },
+            {
+              endsWith: [
+                { path: ["type(inheritanceDepth=0)", "baseUrl"] },
+                {
+                  parameter: `/types/entity-type/person/`,
```
CiaranMn, line 256: I haven't seen this shorthand before... does it (1) add the frontend URL or (2) match any type base URL which ends with this string? e.g. would `https://blockprotocol.org/types/entity-type/person/` be matched?

CiaranMn (approved):
Looks good – one non-blocking question

tests/hash-backend-integration/src/tests/graph/knowledge/primitive/entity.test.ts
TimDiekmann: It's treating `baseUrl` as a string, so your example would be matched.

TimDiekmann: That also works with`versionedUrl` btw, so you could ask for `"/types/entity-type/person/v/1"`.

CiaranMn: Understood, thank you

nonparibus (approved):

PR #3212, opened 2023-09-20
H-671: Add a timeout for the type fetcher

TimDiekmann:

To avoid doing [H-670](https://linear.app/hash/issue/H-670/make-the-type-fetcher-a-temporalio-worker) right now, we should try to make the type fetcher more robust to external services being unresponsive (e.g. because bp.org is currently redeploying)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/type-fetcher/src/fetcher_server.rs
+++ b/apps/hash-graph/lib/type-fetcher/src/fetcher_server.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use futures::{stream, StreamExt, TryStreamExt};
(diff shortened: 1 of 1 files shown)


TimDiekmann:
I guess there was not timeout but I also don't know what the implications of that is.
The documentation states
> Enables a request timeout.
>
> The timeout is applied from when the request starts connecting until the response body has finished.

So I guess it otherwise would wait indefinitely? A timeout would not help if the connection is refused and it has no built-in retry logic.

CiaranMn (approved):

PR #3216, opened 2023-09-20
H-677, H-751, H-780: Replacable file entities, file download button, infer image files

CiaranMn:

Makes the following improvements to the file upload feature:
- H-677: allow users to replace the file in an existing entity
  - updates to GraphQL file upload mutations to accommodate specifying an existing entity
  - button to replace a file on an existing file entity's page
  - add preview box for non-image entities, to accommodate the replace button and replacement form
  - update the org settings form to replace files if the user has created an avatar previously, rather than creating new entities
- H-751: add button to download file on the file entity's pages
- H-780: automatically assign `Image` type to uploaded files if they are image files, even if the user uploaded them via the generic `File` type's page

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- There is a better design for the 'uploading' state in the file replacement form, also for implementation on the uploads tab of file entity types, the implementation of which is tracked in H-747

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. Try adding a file, visiting its page and replacing the file

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/883ec3c8-ffff-4bd4-8334-0e3e45153407

https://github.com/hashintel/hash/assets/37743469/860ca05d-2ff2-4ceb-86c1-fced2df7b1d9


apps/hash-api/src/graph/knowledge/system-types/file.ts
```diff
@@ -87,26 +144,24 @@ export const createFileFromUploadRequest: ImpureGraphFunction<
...
-    entityTypeId || types.entityType.file.entityTypeId;
+  const { entityTypeId, existingEntity, mimeType, ownedById } =
+    await generateCommonParameters(ctx, authentication, params, filename);
 
   try {
     const properties: FileProperties = {
       "https://blockprotocol.org/@blockprotocol/types/property-type/description/":
         description ?? undefined,
       "https://blockprotocol.org/@blockprotocol/types/property-type/file-name/":
         filename,
```
benwerner01, line 162: We might want to add these to the `ontology-types.ts` file at some point, so they can be re-used more easily

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx
```diff
@@ -1,50 +1,237 @@
...
+            {isImage ? (
+              <ImageWithCheckedBackground
+                alt={alt}
+                src={url}
+                sx={{ height: previewHeight }}
+              />
+            ) : (
+              <Stack alignItems="center" spacing={2}>
+                <FileIcon sx={{ color: "gray.50", fontSize: 48 }} />
+                <Typography sx={{ color: "gray.70" }}>{fileName}</Typography>
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx:229]:
```suggestion
                <FileIcon
                  sx={{
                    color: ({ palette }) => palette.gray[50],
                    fontSize: 48,
                  }}
                />
                <Typography sx={{ color: ({ palette }) => palette.gray[70] }}>
                  {fileName}
                </Typography>
```

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/inputs/json-input.tsx
```diff
@@ -95,16 +95,18 @@ export const JsonInput = ({ onChange, value }: CellInputProps<any>) => {
...
-                    onChange(JSON.parse(innerValue));
-                  }}
-                >
-                  <FontAwesomeIcon icon={faCheck} />
-                </IconButton>
+                <Box>
+                  <IconButton
+                    disabled={!isValid}
+                    rounded
+                    sx={{ color: "white" }}
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/inputs/json-input.tsx:102]:
```suggestion
                    sx={{ color: ({ palette }) => palette.common.white }}
```

libs/@hashintel/design-system/src/icon-download.tsx
```diff
@@ -0,0 +1,20 @@
+import { SvgIcon, SvgIconProps } from "@mui/material";
+import { FunctionComponent } from "react";
+
+export const DownloadIcon: FunctionComponent<SvgIconProps> = ({
```
benwerner01, line 4: Can we include the variant of this FA icon in the component name? (e.g. `DownloadRegularIcon` if it is `regular`). In the designs we are using both `light` and `regular` icons in various places, so this prevents potential clashes between icon names down the line.

I think this also applies to the other icons introduced in this PR.

libs/@hashintel/design-system/src/components.ts
```diff
@@ -16,12 +16,15 @@ export * from "./icon-barcode";
 export * from "./icon-button";
 export * from "./icon-circle-plus";
 export * from "./icon-close";
+export * from "./icon-download";
+export * from "./icon-file";
```
benwerner01, line 20: Why are we adding these to the design system? I've been sticking new FA HASH app icons into `apps/hash-frontend/src/shared/icons/`, but happy to adapt my behaviour if the design system is a better place for them.

However it would be nice for them to have their own directory to avoid both components and icons living in the `src` directory , and for us to be consistent with how we name them between the HASH app and the design system.

benwerner01 (commented):
Thanks Ciaran, the functionality is working well for me locally. Had a couple of questions and nits which might be worth addressing as part of this PR.

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx:229]: Is there any functional difference to this, or is it for consistency?

apps/hash-api/src/graph/knowledge/system-types/file.ts
CiaranMn: I have already added these to `libs/@local/hash-subgraph/src/main.ts` (maybe not the best place for them, but it's where `linkEntityTypeUrl` already was). In this instance it isn't necessary to use those aliases though, since they are keys on `FileProperties` and therefore we have autocomplete and type safety already.

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx:229]: It [seems not](https://mui.com/system/getting-started/the-sx-prop/#theme-aware-properties), but happy to use this style if we'd rather

libs/@hashintel/design-system/src/components.ts
CiaranMn: Only really in anticipation of us wanting to use them in other libraries, e.g. the type editor, in which case it saves us moving them later (and it makes no real difference otherwise where they are). We can move them to a different directory, yes.

libs/@hashintel/design-system/src/icon-download.tsx
CiaranMn: Done in https://github.com/hashintel/hash/pull/3216/commits/ca72562a79998eb2dce24f7f67b9908d09b55940

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx]: I mainly do it to get type safety when accessing the colors, which prevents the possiblity for typos being un-caught and incase we ever change the palette again (hopefully we don't) we can find where we are accessing the values more easily.

benwerner01 (dismissed):

apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/file-preview-section.tsx]: That's a good reason to, I have updated in https://github.com/hashintel/hash/pull/3216/commits/a71ba3dabbb543c7db831db03b46bf4e3619904a

benwerner01 (approved):

PR #3223, opened 2023-09-21
H-838: Expose function to check if a user can update an entity

TimDiekmann:

To check if a user will be able to update an entity, a function is required to be exposed for that. The first thing where this is used is to determine if the user can edit the instance settings.

## 🚫 Blocked by

- #3200
- #3204 
- #3210 
- #3211 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest/entity.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest/entity.rs
@@ -5,7 +5,12 @@ use std::sync::Arc;
 use authorization::{
     zanzibar::Consistency, AuthorizationApi, AuthorizationApiPool, VisibilityScope,
 };
-use axum::{extract::Path, http::StatusCode, routing::post, Extension, Router};
+use axum::{
+    extract::Path,
+    http::StatusCode,
+    routing::{get, post},
+    Extension, Router,
+};
 use graph_types::{
     knowledge::{
         entity::{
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -81,9 +88,14 @@ impl RoutedResource for EntityResource {
...
-                    "/:entity_id/public",
-                    post(make_entity_public::<A>).delete(make_entity_private::<A>),
+                .nest(
+                    "/:entity_id",
+                    Router::new()
+                        .route(
+                            "/public",
+                            post(make_entity_public::<A>).delete(make_entity_private::<A>),
+                        )
+                        .route("/update", get(can_update_entity::<A>)),
```
CiaranMn, line 98: I wonder if there's a better route we can use for this that makes clear it's a permission check... if we want to have a 'can a user view' check, the equivalent to this would be `GET /entities/:entity_id/view`, which I think is ambiguous.

I don't have any great ideas but something we can think about as we add more routes. Maybe `GET /entities/:entity_id/permission/:permission`?

CiaranMn (approved):

apps/hash-graph/lib/graph/src/api/rest/entity.rs
CiaranMn: Or `GET /entities/:entity_id/authorization/:permission`?

nonparibus (dismissed):

nonparibus (approved):

PR #3229, opened 2023-09-22
H-583, H-665: Improve settings sidebar UX, make type editor alert modals consistent

CiaranMn:

- H-583: Update the settings sidebar
  - so that only one menu at a given depth can be expanded
  - to remove the toggle caret at the top level
  - to make items progressively closer together as you go down the tree
- H-665: Make the alert modals in the type editor closer to the design, and consistent (the 'convert to link style' modal was different to the modals you get when trying to add chosen parents)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Check out locally or view the deployment (frontend changes only)
2. Click around the settings sidebar, try behavior on refreshing at different depths
3. In the type editor, try convert to link type, or adding a link type to a non-link type

#### Settings sidebar
https://github.com/hashintel/hash/assets/37743469/e9a378e0-73db-413e-8f11-f3df272a59e7

#### Modals in type editor

Screenshot 2023-09-22 at 11 35 33

Screenshot 2023-09-22 at 11 27 14


benwerner01 (dismissed):
Thanks Ciaran, the changes look good to me and are working in the deployment as expected.

Something I noticed that wasn't introduced in this PR is that when you refresh on the `https://app.hash.ai/settings/organizations/new` page, you are redirected to the homepage instead of staying on that page which seems like unexpected behaviour. If it's not worth addressing in this PR, maybe we should create a follow-up task?

CiaranMn:
> Something I noticed that wasn't introduced in this PR is that when you refresh on the `https://app.hash.ai/settings/organizations/new` page, you are redirected to the homepage instead of staying on that page which seems like unexpected behaviour. If it's not worth addressing in this PR, maybe we should create a follow-up task?

Thanks @benwerner01 – I've fixed the redirect to accommodate the `/new` path in https://github.com/hashintel/hash/pull/3229/commits/0c05454b946d2211c8b1f7f182b645d8881c2b6a, and also reserved the 'new' shortname since an org called `new` would clash with the current path (if anyone ever wants it we can consider moving it)

benwerner01 (approved):

TimDiekmann:
`yarn install` failed in CI due to `502: Bad Gateway`

PR #3230, opened 2023-09-22
H-832: Update entity/type tables for easier global filtering

CiaranMn:

Improves the 'all entities / all types' tables as follows:
1. By default show records from a user's own web and their organizations' (previously the default 'internal' records shown were only those of the active web)
2. Allow clicking on the external record count to toggle including or excluding external records (previously had to click on the filter to do so)
3. Styling tweaks to bring closer to the design: shorter count chips; text color; hover effect on filter checkboxes

This PR:

The changes in this PR:

The changes in this PR:

#### Old

https://github.com/hashintel/hash/assets/37743469/6e6050d6-02c5-4a13-987b-13dd67077432

#### New

https://github.com/hashintel/hash/assets/37743469/f8c9b075-e966-4c1e-8208-cd4506594852


benwerner01 (dismissed):
Thanks Ciaran, this is a nice improvement. Changes look good to me and are working in the deployment as expected 👍

nonparibus:
@CiaranMn This doesn't quite match the Figma designs (there was an update yesterday, and I wonder if the style portion of this was done beforehand?)

`eye` onhover when in inactive state, `eye-slash` onhover when in active state, and `check` when active. The earth icon remains for the inactive (default) state.

Otherwise looks great. Have we actually changed the functionality to include types from all webs a user has access to, rather than just the current active "browsing as" web? (Pending deprecation of that switcher)

CiaranMn:
> @CiaranMn This doesn't quite match the Figma designs (there was an update yesterday, and I wonder if the style portion of this was done beforehand?)
> 
> `eye` onhover when in inactive state, `eye-slash` onhover when in active state, and `check` when active. The earth icon remains for the inactive (default) state.
> 
> Otherwise looks great. Have we actually changed the functionality to include types from all webs a user has access to, rather than just the current active "browsing as" web? (Pending deprecation of that switcher)

@nonparibus @benwerner01  updated in https://github.com/hashintel/hash/pull/3230/commits/50d000ba2d8cdc346e8e91a871ba0f1d8977ebb5

https://github.com/hashintel/hash/assets/37743469/1d070837-25ba-4b8d-a228-a9188b3e5e0f


Yes, the 'always on' types/entities are now from all webs a user belongs to, not just whatever the currently selected one is.

nonparibus (approved):
This is great, thank you. @CiaranMn If it's easy enough to stop the tooltip wrapping two lines here, let's do it for the sake of minimizing visual shift. Otherwise this can merge.

Capture 2023-09-25 at 09 40 39@2x

Capture 2023-09-25 at 09 40 35@2x

nonparibus (dismissed):

CiaranMn:
> This is great, thank you. @CiaranMn If it's easy enough to stop the tooltip wrapping two lines here, let's do it for the sake of minimizing visual shift. Otherwise this can merge.

@nonparibus Fixed in https://github.com/hashintel/hash/pull/3230/commits/8729943dbddd4ee8cb8dc97eb648967610d3b3a2

nonparibus (approved):
Super, thanks.

PR #3244, opened 2023-09-25
H-206: Allow for insertion of property/link values into pages

benwerner01:

This PR adds an improved version of the mention suggestor, which allows for inserting entities, entity property values, or entity outgoing link values into a page.

## 🔗 Related links

- H-206

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- There are numerous issues with the current implementation of the search input which require addressing (have created H-862 for this):
  - the cursor in the search input is not styled correctly
  - there is missing padding between the `@` and the search text in the search input
  - there is a react error that is thrown occasionally when switching between pages, which I think was introduced as part of the implementation of the search input, which creates nodes in the DOM that aren't controlled by react. It is possible this may go away if we make use of `Decoration.widget` instead of `Decoration.inline` to style the input, as this allows us to use custom react elements to add the styling/functionality of the search input. 

## 🐾 Next steps

- H-862
- H-930
- H-929
- H-928

## 🛡 What tests cover this?

None

## ❓ How to test this?

Because this PR touches `libs/@local/hash-isomorphic-utils/src/prosemirror.ts`, which is used in the BE, it's my understanding that it needs to be tested locally rather than in the deployment.

1. Go to a page
2. Type `@`, which should display a search input and the mention suggester dropdown.
3. To select an entity, click on one or press enter. To select a property or an outgoing link, use the keyboard arrows or click the right chevron to open the entity's sub-menu and select the corresponding value there.
4. Observe the link/property/entity being inserted into the document. When clicking on property/link values, a popup should appear indicating the corresponding link/property type, and source entity.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/bfdc7a2e-9902-4edc-957a-727dfdf9a1f1


libs/@local/hash-graphql-shared/src/graphql/types.ts
```diff
@@ -17,7 +18,18 @@ export type TextToken =
       link?: string;
     }
   | { tokenType: "hardBreak" }
-  | { tokenType: "mention"; mentionType: "user" | "page"; entityId: EntityId };
+  | {
+      tokenType: "mention";
+      mentionType:
+        | "user"
+        | "page"
+        | "entity"
```
benwerner01, line 26: @CiaranMn I don't see the need for the `"user"` and `"page"` mention types, as they require no additional data to render (just the `entityId`). The reason I didn't remove them is because this might break the app in unexpected ways for anyone that has inserted users and pages, as the mention type will be outdated.

Should I create a pre-launch follow-up task for removing these, so we can do it briefly before we wipe the database?

CiaranMn (commented):
Thanks @benwerner01 – I've had a play and it works as expected. I have a few suggestions for UX improvements which @nonparibus will have views on – these could potentially be follow ups if complicated. I appreciate some will be pre-existing behavior but worth improving while we're thinking about this.

These are largely around (a) making it more obvious what each chip represents, and (b) making what happens when you click on them better.

**1. Tooltips on mentions**

At the moment, the only clue as to what each mention chip represents is the URL, which mostly is a UUID. I think a tooltip would help here. The tooltip could either 
(a) describe what it it is, e.g.
- `entity`/`user`/`page`: `[type title] -> [entity label]`
- `property`: `[type title] -> [entity label] -> [property title]`

(b) describe what will happen if you click on it, e.g.
- `entity`/`property`: `View [entity label]'s properties`
- `page`: `View page [page title]`
- `user`: `View [user label]'s profile`
I appreciate there is a design for a fuller tooltip for property values specifically which would be more complicated to implement and involve usage of entity history, but something simple like the above would be a good stopgap.

**2. What happens on click**

For pages/entities I think 'go to page'/'go to entity' is fine.

For users/orgs, I think it should go to their profile (maybe we need to expand/amend `user` to a `profile` mention to include orgs)

For property labels, I think the existing popup (on click) is a bit weird, where you can either click on the property title to go to the _type_ of the entity, or the entity label to go to the entity. I'd get rid of this popup and just make it a link to the entity's page.

Screenshot 2023-09-26 at 19 17 49

**3. Visual distinction**

I think it might be better to visually distinguish at least (a) users/orgs, (b) properties, and (c) entities generally, e.g. by subtle background colours on the chips, and/or icons.

**4. Selector behavior**

Potential improvements:
1. It's impossible to tell if the sub-menus are expanded or not – can we make the arrow rotate down if they are?
2. You can't navigate to the sub-menu titles or 'back to results' by keyboard – how much work is it to make them keyboard navigable? 
3. I would personally expect hitting tab to insert a result – it currently doesn't and just takes focus to some other place (actually a drag handle on a block for me for some reason) – maybe we should make tab insert a result instead, like enter does

nonparibus:
@CiaranMn To be clear, the design actually looks like this:
Capture 2023-09-26 at 15 15 32@2x

The second row (that currently reads 'System Value' in your screenshot) is meant to be a link to the entity whose attribute value has been inserted, while the top row is meant to be a link to the property type, for a 'definition' of the thing that a user is looking at.

Where we `@mention` **entities** directly I am happy to show an icon of their type(s), depending which list they were selected from at the point of `@insertion`.

Where we `@insert` **property values** directly, there will be no icon, and I think it should remain this way for clean referencing of facts/figures/things inline with text without it being disruptive. If a property value has a data type label, these can in the future be shown as seen in the designs:
Capture 2023-09-26 at 15 18 21@2x

Where we `@insert` **link values** directly, I am happy to experiment with both approaches, but I slightly lean towards the latter, so that all 'attributes' of an entity are handled consistently, and it is easy to visually distinguish inserted values from direct mentions.

I am happy to keep the full designed tooltip (envisaged to be onhover) as on-click, and implement our small-sized dark-colored standard tooltip onhover over chips. In the case of attributes these should read `[attribute name] of [entity label]` - e.g. `Market Capitalization of Microsoft` or `Name of Ciaran`.

benwerner01:
> For users/orgs, I think it should go to their profile (maybe we need to expand/amend user to a profile mention to include orgs)

@CiaranMn I've added this in https://github.com/hashintel/hash/pull/3244/commits/58456f33594039318c76f66f67fc3109b48873e1

> I think it might be better to visually distinguish at least (a) users/orgs, (b) properties, and (c) entities generally, e.g. by subtle background colours on the chips, and/or icons.

I've added icons to entity mentions in https://github.com/hashintel/hash/pull/3244/commits/61edacd2e52d0e910173b35c8f6f4472b6b304c8, and left a `@todo` for when we have entity type icons. Here's what it looks like now:
image

I've also used made use of these icons in the dropdown in https://github.com/hashintel/hash/pull/3244/commits/522d1ff34ee1227fab2c470dc940354abf62637a:

image

> I am happy to keep the full designed tooltip (envisaged to be onhover) as on-click, and implement our small-sized dark-colored standard tooltip onhover over chips. In the case of attributes these should read [attribute name] of [entity label] - e.g. Market Capitalization of Microsoft or Name of Ciaran.

@nonparibus I've added a tooltip to link/property values as suggested, and kept the incomplete popover on-click:

https://github.com/hashintel/hash/assets/42802102/3945f3bc-96c5-42a4-a002-287c90ad0816

> It's impossible to tell if the sub-menus are expanded or not – can we make the arrow rotate down if they are?

@CiaranMn I'm not sure I fully understand this point, when a sub-menu is expanded the right chevron of the entity has a background color. Are you referring to something else?
image

> You can't navigate to the sub-menu titles or 'back to results' by keyboard – how much work is it to make them keyboard navigable?

@CiaranMn It should be possible via the right/left arrow keys to navigate to and from a sub-menu, is this not working for you?

https://github.com/hashintel/hash/assets/42802102/8e83a3b8-3ccf-4193-be26-c0691ca5ee4e

> I would personally expect hitting tab to insert a result – it currently doesn't and just takes focus to some other place (actually a drag handle on a block for me for some reason) – maybe we should make tab insert a result instead, like enter does

I've added a listener for "Tab" presses to insert results in https://github.com/hashintel/hash/pull/3244/commits/34d364a479d8c2f46cb7bcd93a0cdfe6b6d6a902, cc @nonparibus for thoughts

nonparibus:
@benwerner01 Tab completion would be nice. 👍

benwerner01:
> Can we also enabling the entry of chips into page titles, if not too much of a PITA given how we handle them with ProseMirror? Ideally we'd allow titles to be 'dynamic' in the same way on-page content can be, when attribute chips are used within.

In theory it's possible, but right now page titles are just strings we store as a property on the page, and we're just using a normal text input instead of prosemirror for this right now. If you want I can create a task for this, don't think it makes sense to do as part of this PR as we'd need to change how we're inputting, storing and rendering page titles.

CiaranMn:
@benwerner01 re. 'sub-menus' – apologies, I mean the titles of the groups within the entity type list, e.g. "Pages", "Users". There is currently no indication as to whether the group is expanded or not. I suggest the caret could point downwards if it is expanded.

I've also noticed that my pages appear repeatedly – maybe this is multiple versions appearing? Or possibly a React state/array key issue.

Also, the group header disappears when clicked.

https://github.com/hashintel/hash/assets/37743469/f1511cf4-8c4f-4331-89d3-5d98cda8676b


Re. keyboard navigation, I was talking about making the group headers part of the keyboard flow, so that you can use the keyboard to navigate to and toggle expansion (and the same for the 'back to results' button at the bottom of the sub-menu). If making these part of the flow is difficult then it could be a follow up. If it's just a matter of adding a tab index to them then suggest we do it.

benwerner01:
> There is currently no indication as to whether the group is expanded or not. I suggest the caret could point downwards if it is expanded.

@CiaranMn that's a good point, it was using the styling from when I had the section titles link to the type's page. Have made the chevron rotate in https://github.com/hashintel/hash/pull/3244/commits/0678bd7412814567987000d6fa60c16e3deed8fe, and also hid the chevrons for sections that can't be expanded because there are no additional entities in https://github.com/hashintel/hash/pull/3244/commits/5c523529230d603dad16e3ade539ba7e8e20d6bd

> I've also noticed that my pages appear repeatedly – maybe this is multiple versions appearing? Or possibly a React state/array key issue.

Thanks for catching this, there were two separate issues:
1. Archived pages weren't being filtered out from the list of entities
2. The entities in all accounts were being included in the results, so pages with the same name across accounts may have appeared as duplicates.

I've refactored the query to only include the entities in the user's workspace, and filtered the archived pages in https://github.com/hashintel/hash/pull/3244/commits/8cbdfeb1f74ea3d56fdea9be73738affb0f10134. We could also include the entities in orgs that the user is a member of, but there would be no way to differentiate between those entities and the user's entities in the current designs (cc @nonparibus).

> Also, the group header disappears when clicked.

Addressed this as part of https://github.com/hashintel/hash/pull/3244/commits/0678bd7412814567987000d6fa60c16e3deed8fe

> I was talking about making the group headers part of the keyboard flow, so that you can use the keyboard to navigate to and toggle expansion (and the same for the 'back to results' button at the bottom of the sub-menu). If making these part of the flow is difficult then it could be a follow up. If it's just a matter of adding a tab index to them then suggest we do it.

Ah understood. We can do this, but it's not a super straightforward change as I need to modify all index calculations in the suggester, so want to confirm with @nonparibus whether this is the behaviour we want. Also happy to address this in follow-up.

apps/hash-frontend/src/blocks/page/mention-view/mention-display.tsx
```diff
@@ -1,156 +1,328 @@
...
+      return entityLabel;
+    } else if (mention.kind === "outgoing-link") {
+      const outgoingLinkAndTargetEntities = getOutgoingLinkAndTargetEntities(
+        entitySubgraph,
+        entityId,
+      ).find(
+        ({ linkEntity: linkEntityRevisions }) =>
+          linkEntityRevisions[0]?.metadata.recordId.entityId ===
+          mention.linkEntityId,
+      );
```
CiaranMn, line 89: If a link is replaced, there will be a new link entity created, rather than the existing one updated. So this approach will only work so long as the link isn't replaced (I'm guessing that the chosen link will not be found in the subgraph once it's archived, or – if it is in the subgraph – it will be out of date).

An approach which would be resilient to 
(a) the link being replaced, and 
(b) the link type being updated 
would be to store the link type's base URL instead, and then find the link from the entity of that type.

The downside to this would be where there are multiple links of the same type. For now we could just take the first one.

apps/hash-frontend/src/blocks/page/mention-view/mention-display.tsx
```diff
@@ -1,156 +1,328 @@
...
+      const targetEntity = outgoingLinkAndTargetEntities?.rightEntity[0];
+
+      const targetEntityLabel = generateEntityLabel(
+        entitySubgraph,
+        targetEntity,
+      );
+
+      return targetEntityLabel;
+    } else {
+      const propertyTypeBaseUrl = extractBaseUrl(mention.propertyTypeId);
```
CiaranMn, line 100: I would store the property type base URL rather than the id, so that the mention does not break if the entity's type is upgraded to change the version of the property type. Looking for the property type in the subgraph would then have to be by base URL rather than versioned URL.

apps/hash-frontend/src/blocks/page/create-suggester/mention-suggester.tsx
```diff
@@ -1,188 +1,632 @@
...
+    StructuralQueryEntitiesQueryVariables
+  >(structuralQueryEntitiesQuery, {
+    variables: {
+      query: {
+        filter: {
+          any: [
+            {
+              equal: [
+                { path: ["ownedById"] },
+                { parameter: authenticatedUser.accountId },
```
CiaranMn, line 140: Re. your question in a comment, I think we should include entities owned by orgs the user is a member of.

apps/hash-frontend/src/blocks/page/create-suggester/mention-suggester.tsx
```diff
@@ -1,188 +1,632 @@
...
+  FunctionComponent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import { useKey } from "rooks";
+
+import { useScrollLock } from "../../../components/grid/utils/override-custom-renderers/use-scroll-lock";
```
CiaranMn, line 35: Can you move this into a shared folder please?

apps/hash-frontend/src/blocks/page/mention-view/mention-display.tsx
```diff
@@ -1,156 +1,328 @@
...
-import { useUsers } from "../../../components/hooks/use-users";
-import { PageIcon } from "../../../components/page-icon";
+import { useGetOwnerForEntity } from "../../../components/hooks/use-get-owner-for-entity";
 import { generateEntityLabel } from "../../../lib/entities";
 import { constructPageRelativeUrl } from "../../../lib/routes";
+import { ArrowUpRightRegularIcon } from "../../../shared/icons/arrow-up-right-regular-icon";
 import { Link } from "../../../shared/ui";
-import { MentionType } from "../create-suggester/mention-suggester";
+import { useEntityIcon } from "../../../shared/use-entity-icon";
+import { Mention } from "../create-suggester/mention-suggester";
```
CiaranMn, line 21: Teechnically this should be in a shared folder (the only folder name that should appear after a `../` is `shared`, although I appreciate we have some legacy buckets like 'components' lying around)

libs/@hashintel/design-system/theme-override.d.ts
```diff
@@ -52,7 +52,7 @@ declare module "@mui/material/styles" {
   interface Palette {
     gray: PaletteValue & { 5: string; 15: string };
     grey: undefined;
-    blue: PaletteValue & { 25: string };
+    blue: PaletteValue & { 25: string; 15: string };
```
CiaranMn, line 55: a nit for my sanity
```suggestion
    blue: PaletteValue & { 15: string; 25: string };
```

libs/@local/hash-graphql-shared/src/graphql/types.ts
CiaranMn: I wouldn't worry about breaking existing mentions... potential reasons to keep distinct types:
1. if the user/page was somehow missing, we'd not know what type of mention was broken
2. we couldn't handle loading state differently, should we want to

Maybe create a task and we can triage it?

CiaranMn (commented):
Thanks @benwerner01, functionality looking good... a few final 🤞 comments on the underlying implementation

nonparibus:
We do want them to be accessible w/ a tab index, yes.

And all entities visible to a user should appear in the insertion menu, not just those in their account. But as discussed in earlier planning, we want to weight a user's own and member web's own entities higher in the returned results list.

apps/hash-frontend/src/blocks/page/mention-view/mention-display.tsx
benwerner01: That's a good point, I've applied this suggestion in https://github.com/hashintel/hash/pull/3244/commits/0ded7492a47c3c492730569cc7da49bfc55d607f

apps/hash-frontend/src/blocks/page/mention-view/mention-display.tsx
benwerner01: Good idea, I've done this in https://github.com/hashintel/hash/pull/3244/commits/63a3c6b4445f9c728bf43610f8239a85aedb91be

benwerner01: Good idea, I've done this in https://github.com/hashintel/hash/pull/3244/commits/63a3c6b4445f9c728bf43610f8239a85aedb91be

benwerner01: Good idea, I've done this in https://github.com/hashintel/hash/pull/3244/commits/63a3c6b4445f9c728bf43610f8239a85aedb91be

nonparibus:
@benwerner01 Further viz tweaks:

1. can we make sure the whitespace to the left of 'Recently used' and named entity types is equalized? At the moment the former appears more indented. Screenshot from the app demonstrating this misalignment:
Capture 2023-09-29 at 16 50 44@2x

2. can we make sure the chevron to the right of entity type names matches the Figma? It should be smaller, and appear vertically centered with the text to its left. Screenshot from Figma showing the ideal state:
Capture 2023-09-29 at 16 52 17@2x

apps/hash-frontend/src/blocks/page/create-suggester/mention-suggester.tsx
benwerner01: I've included org entities in https://github.com/hashintel/hash/pull/3244/commits/64e2aa4b5767f87b8fc08cd35675224d273792d6, have also ensured that the user's entities are displayed before org entities in each section (a suggestion from @nonparibus)

apps/hash-frontend/src/blocks/page/create-suggester/mention-suggester.tsx
benwerner01: Yes, moved in https://github.com/hashintel/hash/pull/3244/commits/2be6c078db5a5f6395a2d865b8e6ac78d05de616

libs/@local/hash-graphql-shared/src/graphql/types.ts
benwerner01: I've created [the triage ticket](https://linear.app/hash/issue/H-902/consider-deprecating-user-and-page-mention-types-as-were-now-using), in this PR I've stopped making use of the `user` and `page` mentions and started using `entity` mentions directly instead - let me know if you think I should change this.

I agree we could handle the loading state differently, but there might be other ways of achieving that with the `entity` mention (by setting `loadingIcon` in the mention's metadata for example)

benwerner01:
> And all entities visible to a user should appear in the insertion menu, not just those in their account. But as discussed in earlier planning, we want to weight a user's own and member web's own entities higher in the returned results list.

@nonparibus I've added the entities from orgs that user is a member of back into the suggester, and introduced a custom sorting function that ensures the user's entities are displayed before any org entities in each section. We will want to rework this though when we are using a search endpoint, rather than filtering client side, as we discussed.

> can we make sure the whitespace to the left of 'Recently used' and named entity types is equalized? At the moment the former appears more indented. Screenshot from the app demonstrating this misalignment:

Thanks for flagging, I saw this and fixed this in an earlier commit it should look like this in the latest version:

image

> can we make sure the chevron to the right of entity type names matches the Figma? It should be smaller, and appear vertically centered with the text to its left.

I've the chevron to have a font-size of `10px` in https://github.com/hashintel/hash/pull/3244/commits/ed53e46e1a370a54e16a41fa1edb6a998ba94628, I noticed that it looks slightly thinner than in the designs but I'm not 100% sure. I'm using the `regular` variant of the FA icon, did you want me to use `solid` here instead? Here's what it looks like now:

image

> can we also tidy up the top-left corner of the dropdown, to match the above Figma? 

Have addressed in https://github.com/hashintel/hash/pull/3244/commits/5128b691bc2d463ffe5000cddc0d90576f103d48, sorry for missing this. Now looks like:

image

nonparibus:
1. "I'm using the regular variant of the FA icon, did you want me to use solid here instead?" @benwerner01 Yes, please. The `solid` variant is the one we're using for this smaller-sized chevron in Figma.

2. Can we slightly increase whitespace between all but the first section header (i.e. 2..n) and thing list element above it? At the moment when that element is hovered over (e.g. as seen here: https://user-images.githubusercontent.com/42802102/271938511-1fe2b601-0de0-4729-85a6-e2042d25731c.png) it appears that the section header is closer to it than the element beneath it, because of the fill background on the element above. You can see correct spacing [here in Figma](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=459%3A21507&mode=design&t=THLgm315jA52JFqq-1).

3. Because I haven't been able to test on the prod instance, can you confirm how multi-line entry field looks, and multi-line chips look?

apps/hash-frontend/src/blocks/page/create-suggester/mention-suggester.tsx
nonparibus: We do want to include **all entities**, not just those a user is a member of. Although we can wait until permissions/authorization are in until we introduce this.

Stuff from a user's own webs (their own, plus those they are a member of) ought to be heavily weighted to generally appear above external ones in the list (exceptions may be made for exact name match things from outside a graph, and/or which are extremely popular and widely used, or which are system types, etc... pending full search algo determination). @CiaranMn @benwerner01

benwerner01:
> "I'm using the regular variant of the FA icon, did you want me to use solid here instead?" @benwerner01 Yes, please. The solid variant is the one we're using for this smaller-sized chevron in Figma.

Addressed in https://github.com/hashintel/hash/pull/3244/commits/6cf7355705c4a0c91429be1605d9cba6bf93ae24

> Can we slightly increase whitespace between all but the first section header (i.e. 2..n) and thing list element above it? At the moment when that element is hovered over (e.g. as seen here: https://user-images.githubusercontent.com/42802102/271938511-1fe2b601-0de0-4729-85a6-e2042d25731c.png) it appears that the section header is closer to it than the element beneath it, because of the fill background on the element above. You can see correct spacing [here in Figma](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=459%3A21507&mode=design&t=THLgm315jA52JFqq-1).

I've increased the spacing slightly in https://github.com/hashintel/hash/pull/3244/commits/f64b80ce6371998085d620733fccb6a184a287d8, to my eyes it wasn't significantly off.

> Because I haven't been able to test on the prod instance, can you confirm how multi-line entry field looks, and multi-line chips look?

It currently looks like this:
image

Looking at the designs I appreciate this doesn't perfectly match them, but this will be quite time consuming to get right as we'd need to do manual calculations somehow for a chip to know when it wraps onto multiple lines, and we'd need to mess with line spacing in prosemirror. Have created [a follow-up ticket](https://linear.app/hash/issue/H-928/improve-styling-of-multi-line-mentions) for now.

CiaranMn (approved):

PR #3245, opened 2023-09-25
H-598: Better error messages when parent types cannot be added

CiaranMn:

Makes the error messages more helpful when users cannot add parent types in the following circumstances:

1. The parent contains a property/link the type also has
Screenshot 2023-09-25 at 20 27 07

2. The parent has a parent which has a property/link the type also has
Screenshot 2023-09-25 at 20 36 19

3. Duplicate link/property types among parents
Screenshot 2023-09-25 at 20 35 19

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Try adding parents which cannot be added as described in the cases above


nonparibus:
@CiaranMn

1. What happens if multiple property types conflict, as opposed to just one, in any of these 3x scenarios?
2. Happy to merge this as-is, provided bullet (1) above is accounted for, but would we be able to do this (and if Y capture as a triage ticket, likely for processing as a nice-to-have future enhancement): re. states (1) and (2) screenshotted in your original description, offer to just remove the conflicting property type declared on the immediate entity type, so the user can proceed with adding the other parent type? Copy might read something like: “New parent type (Charity) declares a property type (Property Type Name) which is already expected by ORIGINAL-TYPE-NAME. Would you like to automatically replace this?” — and then show a “replace Founded Date v3 with Founded Date v4" side-by-side thing.

CiaranMn:
> 1. What happens if multiple property types conflict, as opposed to just one, in any of these 3x scenarios?
> 2. Happy to merge this as-is, provided bullet (1) above is accounted for, but would we be able to do this (and if Y capture as a triage ticket, likely for processing as a nice-to-have future enhancement): re. states (1) and (2) screenshotted in your original description, offer to just remove the conflicting property type declared on the immediate entity type, so the user can proceed with adding the other parent type? Copy might read something like: “New parent type (Charity) declares a property type (Property Type Name) which is already expected by ORIGINAL-TYPE-NAME. Would you like to automatically replace this?” — and then show a “replace Founded Date v3 with Founded Date v4" side-by-side thing.

The user will only see one error at a time, i.e. if they fix one they will then see another when attempting to add a parent.

I've created H-860 to track gathering errors, auto-fixes, and also making the validation more permissive regarding duplicate property/link types. With multiple parents I expect the blanket ban on having any duplicates will be annoying for common property types such as 'description' or 'display name', and we can have a more permissive check that the duplicates do not have conflicting constraints. (@nonparibus)

benwerner01 (approved):
Thanks Ciaran, this looks good to me and is working for me in the deployment as expected.

I'm wondering whether we should use some bold or italic styling around the property/link type names, to make it slightly more immediately obvious what the issue is? Just an idea though, doesn't block merging.

CiaranMn:
> Thanks Ciaran, this looks good to me and is working for me in the deployment as expected.
> 
> I'm wondering whether we should use some bold or italic styling around the property/link type names, to make it slightly more immediately obvious what the issue is? Just an idea though, doesn't block merging.

@benwerner01 this would be ideal, but because some of these errors are thrown from hooks that are used in other places, it doesn't really make sense at the moment for them to return React components. I think we need to rethink validating inheritance more broadly so I would rather not invest more time in it now (more ideas in H-860)

PR #3253, opened 2023-09-26
H-859: Remove duplication of `theme-override.d.ts`

CiaranMn:

We were duplicating `theme-override.d.ts` from the `design-system` library into other packages, but we can avoid the duplication by using a `<reference />` tag to point to the file instead.

This PR is preparation for introducing another package (for the new `hash.ai` site), which might require further modifications to the design system.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `lint:tsc`

## ❓ How to test this?

1. Check it compiles


benwerner01 (approved):
Thanks Ciaran, good idea 👍

PR #3267, opened 2023-09-27
H-74, H-846: Make all property type versions available to type editor. Update media block copy

CiaranMn:

There was a crash in the type editor when clicking 'extend' on an entity type that had a non-latest property type.

This is because we were only getting 'all latest property types' to send to the type editor. There was a hack to merge in the previously-seen type's property types to try and patch over this, but it didn't work consistently, and not if the draft extended type was refreshed (since there was no 'previously seen' type in that case).

I've opted to get _all_ property types for the type editor, rather than getting all property types for all entity types when fetching all entity types, since the former will be adding less data fetching (only adding older versions for property types, which should be a small marginal increase, vs fetching all property types as part of the all entity types requests, where we are currently fetching no property types along with all property types).

Drive-by: H-74, update "Embed Image/Video" to "Insert Image/Video" in media blocks.

## 🔍 What does this change?

- Updates the property type context to fetch all versions of all property types, with a hook to only fetch the latest to maintain the behaviour where only the latest are desired (e.g. the types table).
- Removes the 'merge in previous property types' hack in the type editor page
- Remove a timeout I previously introduced to fix a bug where the react-hook-form state would become malformed. **I'm going slightly crazy with these bugs** because this timeout seemed to be now _causing_ a variation on the bug, and removing it doesn't bring the previous crash back. I don't know if this is something to do with NextJS ugprades, or what.
- **Drive-by:** make the bolding of the link type title consistent with the property type title
- **Drive-by:** don't show the 'upgrade property type' button for inherited property types, since users can't do that.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None / manual.

## ❓ How to test this?

1. Try clicking 'extend type' on an entity type that has a non-latest property type


benwerner01 (approved):
Thanks Ciaran, these changes look good to me and are working in the deployment 👍

PR #3291, opened 2023-09-29
H-886: Bump `max_tokens` for `gpt-4-0613` model

nonparibus:

`gpt-4-0612` has an 8k (not 4k) context window.

## 🔍 What does this change?

This changes `max_tokens` in the `hash-ai-worker-py`'s entity inference workflow.

## 🐾 Next steps

Ideally we'd define the `max_tokens` accepted for each model statically elsewhere (likely alongside pricing information in the internal API) so we don't have to update this in multiple places.

It would still be useful to maintain `max_tokens` inline here separately here, allowing for capping of overall input sizes.


TimDiekmann (approved):
Small correction: this changes the default value, the value is a parameter which is passed to the workflow.

PR #3295, opened 2023-09-30
H-886: Revert "H-886: Bump `max_tokens` for `gpt-4-0613` model"

TimDiekmann:
Reverts hashintel/hash#3291

I'll figure out another way to solve this, but for now it's more important that entity inference is working.

The PR actually broke entity inference with the following status as an example returned:
```json
[
  {
    "code": "UNKNOWN",
    "contents": [],
    "message": "Unable to infer entities: This model's maximum context length is 8192 tokens. However, you requested 8777 tokens (478 in the messages, 107 in the functions, and 8192 in the completion). Please reduce the length of the messages, functions, or completion."
  }
]
```


nonparibus (approved):
Ack :(

TimDiekmann:
> Ack :(

I can probably probably just pass nothing to OpenAI but pydantic threw some weird errors for me. I didn't investigate further because this was not a priority for me.

PR #3314, opened 2023-10-03
H-57, H-894: Adjust GraphQL endpoints to support links and more configuration parameters

TimDiekmann:

The GraphQL endpoints currently do not return the entity id and has a wrong `LinkData` associated (it's not possible to infer the correct entity ID from AI).

With this PR the GraphQL endpoint has all capabilities to infer entities as required for H-57.

## 🔍 What does this change?

Synchronize the parameters/results of GraphQL and the Python worker

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
+++ b/libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
@@ -230,28 +230,25 @@ export const entityTypedef = gql`
       """
       entityTypeIds: [VersionedUrl!]!
       """
-      The model to use for inference. Defaults to "gpt-4-0613".
+      The model to use for inference.
       """
-      model: String
+      model: String!
       """
-      The maximum amount of tokens to generate. If not specified, no limit is applied.
+      The maximum amount of tokens to generate. '0' means that the model's limit will be used.
       """
-      maxTokens: Int
+      maxTokens: Int!
       """
-      Whether to allow empty results. Defaults to true.
+      Whether to allow empty results.
       """
-      allowEmptyResults: Boolean
+      allowEmptyResults: Boolean!
       """
-      The validation to apply to the inferred entities. Defaults to full. Possible values are:
-      - full: All entities must be valid.
-      - partial: Entity properties has to be valid but the "required" condition is omitted
-      - none: No validation is applied.
+      The validation to apply to the inferred entities.
       """
-      validation: EntityValidation
+      validation: EntityValidation!
       """
-      The temperature to use for inference. Defaults to 0.0.
+      The temperature to use for inference.
       """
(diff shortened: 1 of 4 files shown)


libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
```diff
@@ -182,6 +229,29 @@ export const entityTypedef = gql`
...
+      The validation to apply to the inferred entities. Defaults to full. Possible values are:
+      - full: All entities must be valid.
+      - partial: Entity properties has to be valid but the "required" condition is omitted
+      - none: No validation is applied.
+      """
+      validation: EntityValidation
+      """
+      The temperature to use for inference. Defaults to 0.0.
+      """
+      temperature: Float
```
CiaranMn, line 254: We can add the defaults to the schema, and by doing so the GraphQL server will populate the default values if the caller doesn't provide any. One advantage of this is that we can change the defaults of the workflow without changing the defaults here (since the resolver will actually passing explicit values all the time for these properties, rather than nothing)

```suggestion
      model: String = "gpt-4-0613"
      """
      The maximum amount of tokens to generate. If not specified, no limit is applied.
      """
      maxTokens: Int
      """
      Whether to allow empty results. Defaults to true.
      """
      allowEmptyResults: Boolean = true
      """
      The validation to apply to the inferred entities. Defaults to full. Possible values are:
      - full: All entities must be valid.
      - partial: Entity properties has to be valid but the "required" condition is omitted
      - none: No validation is applied.
      """
      validation: EntityValidation = FULL
      """
      The temperature to use for inference. Defaults to 0.0.
      """
      temperature: Float = 0.0
```

CiaranMn (commented):
Looks good, one suggestion on the schema

libs/@local/hash-graphql-shared/src/graphql/type-defs/knowledge/entity.typedef.ts
TimDiekmann: I would probably don't set any defaults in the workflow then, what do you think?

CiaranMn: Sounds good

CiaranMn: Oh, the TypeScript types for the resolver will not understand that these values will be defined, which might be a problem?

TimDiekmann: I made all fields required in bdebdd6b848a8221c7fafc3c12cd58dd898a572d

CiaranMn (approved):

PR #3328, opened 2023-10-04
H-927, H-937: Add test case for full entity inference

TimDiekmann:

This adds a full test case to infer a few entities. This can be improved further but this initial test ensures that the expected functionality is working.

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-py/src/worker/tests/mocks/activity.py
```diff
@@ -69,7 +68,46 @@ async def graph_workflow(
...
+        case _:
+            msg = f"Unknown activity: `{activity}`"
+            raise ValueError(msg)
+
+
+async def graph_workflow(
+    activity: str,
+    arg: Any,  # noqa: ANN401
+    *,
+    task_queue: str = "aipy",
```
indietyp, line 89: would it make sense to restrict this further to Literal?

apps/hash-ai-worker-py/src/worker/tests/mocks/activity.py
```diff
@@ -69,7 +68,46 @@ async def graph_workflow(
     }
 
 
+async def delegate_aipy_queue(
+    activity: str,
```
indietyp, line 72: same here as other comment

indietyp (changes requested):

apps/hash-ai-worker-py/src/worker/tests/mocks/activity.py
TimDiekmann: No, because this is not a Literal but a variable.

indietyp (approved):

PR #3338, opened 2023-10-05
H-866: Implement pinned entity types on User and Org profile pages

benwerner01:

This PR adds the ability to pin entity types to a user's profile page.

In addition, it addressed the outstanding review comments left in https://github.com/hashintel/hash/pull/3319.

Remaining todos:

- [x] pluralise entity type tab titles
- [x] use link icon in edit pinned entity types modal

## 🔗 Related links

- H-866

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

❗️ This PR cannot be tested in the Vercel deployment because it requires backend changes to be applied

1. Go to your user profile page
2. Toggle the tabs, there should be "Profile" and "Pages" initially
3. Add additional pinned entity type tabs by clicking "Modify"

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/c25eefab-757f-485e-9cae-693d3efa6ccc


nonparibus:
@benwerner01 

1. The icons for the "TYPES" listed in the "Modify" modal all appear to be those for entity types, even where link types are listed as the featured thing.
1. What happens when you view a page with 0 matches for a pinned type?
1. Please can we change the 'sort by' dropdowns to read `Recency (newest first)` and `Recency (oldest first)`? Is it trivial to add alphanumerical ascending/descending sort in addition? (`Alphabetized (A-Z)` and `Alphabetized (Z-A)` respectively)

benwerner01:
> The icons for the "TYPES" listed in the "Modify" modal all appear to be those for entity types, even where link types are listed as the featured thing.

Will change the icon for link entity types, have captured this in the PR description

> What happens when you view a page with 0 matches for a pinned type?

@nonparibus currently it looks liked this:
image

Do you want to add some custom copy when there's no entities to display? And maybe a `Create X` button to create an instance of the type?

> Please can we change the 'sort by' dropdowns to read Last edited (newest first) and Last edited (oldest first)? Is it trivial to add alphanumerical ascending/descending sort (Alphabetized (A-Z) and Alphabetized (Z-A) respectively) in addition, along with Creation date (newest first) and Creation date (oldest first)?

I've modified the copy and added alphabetical sorting in https://github.com/hashintel/hash/pull/3338/commits/d40d59a7b42ccea866403cae2fa21c14ebe355fa

> “Add cover image” text is too large in size on profile page (when in ‘edit’ mode on info). Can we drive-by reduce this to correct it?

Thanks for spotting, have reduced in to `12px` in https://github.com/hashintel/hash/pull/3338/commits/ea6b1c9d0266c08094ca29f6789b5725f179d47d

nonparibus:
@benwerner01 

> Do you want to add some custom copy when there's no entities to display? And maybe a Create X button to create an instance of the type?

Yes. Let's also add a loading skeleton while it loads.

Please do the latter while I whip up a design for the former in Figma.

nonparibus:
@benwerner01 I've got a duplicate entity type appearing in the list a few times. This may just be a dev/staging bug, but we should probably make sure logic for deduplicating the list is sound, and ensuring that all 5 selected entity types are different. I don't think we're showing already-listed entity types in the selector list, as I wasn't able to add the entity type an additional time, but we should also make sure this is the case. Maybe removal/re-adding causes duplication? I haven't tried to debug why, so will leave to you.

Capture 2023-10-09 at 11 11 22@2x

apps/hash-frontend/src/pages/[shortname].page.tsx
```diff
@@ -96,6 +101,215 @@ const ProfilePage: NextPageWithLayout = () => {
...
+    () => [
+      {
+        kind: "profile",
+        title: "Profile",
+      },
+      {
+        kind: "profile-pages",
+        title: "Page",
+        pluralTitle: "Pages",
+      },
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page.tsx:114]: Could we just add the Page's entityTypeId into `pinnedEntityTypeBaseUrls` automatically here and let the functions to handle other pinned types deal with fetching and sorting it into groups?

We only need to special case it to stop people removing it, the rest of the functionality should be the same, no?

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
```diff
@@ -0,0 +1,424 @@
...
+import { Org, User } from "../../lib/user-and-org";
+import { useEntityTypesContextRequired } from "../../shared/entity-types-context/hooks/use-entity-types-context-required";
+import { ArrowUpRightRegularIcon } from "../../shared/icons/arrow-up-right-regular-icon";
+import { CustomLinkIcon } from "../../shared/icons/custom-link-icon";
+import { GripDotsVerticalRegularIcon } from "../../shared/icons/grip-dots-vertical-regular-icon";
+import { PlusRegularIcon } from "../../shared/icons/plus-regular";
+import { XMarkRegularIcon } from "../../shared/icons/x-mark-regular-icon";
+import { Button, Link } from "../../shared/ui";
+import { entityTypeIcons } from "../../shared/use-entity-icon";
+import { EntityTypeSelector } from "../[shortname]/entities/[entity-uuid].page/create-entity-page/entity-type-selector";
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:47]: This should be in a shared folder

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
```diff
@@ -0,0 +1,424 @@
...
+          )]: pinnedEntityTypes.map(
+            ({ metadata }) => metadata.recordId.baseUrl,
+          ),
+        },
+      },
+    });
+
+    /** @todo: error handling */
+
+    setLoading(false);
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:155]: If you called `updateEntity` via `useMutation` rather than the BP hook you could use the loading and error return from the ApolloClient hook rather than manually here. Maybe not worth bothering with since we will remove ApolloClient at some point, but probably not for a while and we will replace it with some similar hook when we do

apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx
```diff
@@ -0,0 +1,358 @@
...
+          </Typography>
+          <Typography
+            sx={{
+              color: ({ palette }) => palette.gray[90],
+              fontSize: 11,
+              fontWeight: 500,
+              textTransform: "uppercase",
+            }}
+          >
+            {updatedAtHumanReadable}
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx:114]: Could we have a tooltip here with the full date?

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
```diff
@@ -0,0 +1,424 @@
...
+          paddingTop={2.25}
+        >
+          <Typography
+            variant="smallCaps"
+            sx={{ color: ({ palette }) => palette.gray[70] }}
+          >
+            Types
+          </Typography>
+          <DragDropContext onDragEnd={handleOnDragEnd}>
+            <Droppable droppableId="pinnedEntityTypes">
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:273]: If possible I think it would be better to lock the drag area to this element so that users can't drag the item all around the screen.

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
```diff
@@ -0,0 +1,424 @@
...
+                  ref={providedDroppable.innerRef}
+                  sx={{
+                    marginBottom: 1,
+                    display: "flex",
+                    flexDirection: "column",
+                    rowGap: 0.5,
+                  }}
+                >
+                  {fields.map((field, index) => (
+                    <Draggable
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:286]: I would disable dragging / hide the drag handle if there is only one item

apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const ProfilePageTabs: FunctionComponent<{
+  profile?: User | Org;
+  tabs: ProfilePageTab[];
+  currentTab: ProfilePageTab;
+}> = ({ tabs, currentTab, profile }) => {
+  const currentTabLabel =
+    currentTab.kind === "profile" ? currentTab.title : currentTab.pluralTitle;
+
+  return (
+    <Box sx={{ overflowX: "scroll" }}>
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx:18]: We could made this `auto` so that the scrollbar isn't introduced unless necessary. There will only be layout shift if someone is resizing the window and it appears. We probably also want to style the scrollbar to make it less intrusive.

I also get a vertical scrollbar
Screenshot 2023-10-10 at 12 22 10

apps/hash-frontend/src/pages/[shortname].page/profile-page-header.tsx
```diff
@@ -70,42 +89,86 @@ export const ProfilePageHeader: FunctionComponent<{
...
+            <Box display="flex" justifyContent="space-between">
+              <ProfilePageTabs
+                profile={profile}
+                tabs={tabs}
+                currentTab={currentTab}
               />
-            )}
+              {isEditable ? (
+                <>
+                  <Button
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/profile-page-header.tsx:146]: I would remove the background and border on hover as we don't do this for other similar links... I think the text changing colour is sufficient
Screenshot 2023-10-10 at 12 23 59

CiaranMn (commented):
Thanks Ben, looking good! A few suggestions for code + UX/styling improvements

apps/hash-frontend/src/pages/[shortname].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page.tsx:114]: I can do that, but I'm not really convinced this is worth it. It'd complicate updating `pinnedEntityTypeBaseUrls` in all contexts for something we know we'll always be special casing (when we discussed in standup the other day we concluded we'd want users/orgs to have the non-removable `Pages` tab for the foreseeable future).

It'd also need to be added wherever we create users, so that `pinnedEntityTypeBaseUrls` is populated correctly.

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx]: Addressed in https://github.com/hashintel/hash/pull/3338/commits/579e8cfa9b036947cb67ab568dbb6435b96ee96f

apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx]: Yes, added in https://github.com/hashintel/hash/pull/3338/commits/470d40000fa0417c0a615e4511664fe2cce7830a

Looks like this:
image

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:273]: I spent about 15 minutes trying to get this to work using the `getContainerForClone` prop, and reworking the custom `useDraggableInPortal` hook that I had to add, but couldn't get something working. I think it'd be a good idea, but not sure it's worth investing more time - I can create a follow-up ticket?

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:289]: Good idea, have done so in https://github.com/hashintel/hash/pull/3338/commits/bbb34228bbac17591f552354bbc1b535511e4215

apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/pinned-entity-type-tab-contents.tsx]: Looks good, thank you

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:276]: I wouldn't worry about a follow up, we can change it if it becomes annoying.

Out of interest why the new dnd library? `dnd-kit` handles locking the draggable area / axes with a couple of props.

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:289]: Thank you

apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx]: I've made the overflow `auto` in https://github.com/hashintel/hash/pull/3338/commits/e3f3674a4d2fdd8ee782d0de480840fd923e454d, I can't reproduce the vertical scrollbar. I don't really get why chrome sometimes adds these intrusive scrollbars, and other times uses the native overlaying ones. I'm not sure its worth modifying the styling of this further as I need to rework this once I implement [the updated designs](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?node-id=9717%3A50623&mode=dev) anyway as part of https://linear.app/hash/issue/H-931/display-cover-image-in-user-profile-page

apps/hash-frontend/src/pages/[shortname].page/profile-page-header.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/profile-page-header.tsx:146]: This button is going to be removed once I implement [the updated designs](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?node-id=9717%3A50623&mode=dev) as part of https://linear.app/hash/issue/H-931/display-cover-image-in-user-profile-page, so I think I'll just leave it as is for now.

Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/profile-page-header.tsx:146]: Fair enough!

benwerner01:
> I've got a duplicate entity type appearing in the list a few times. This may just be a dev/staging bug, but we should probably make sure logic for deduplicating the list is sound, and ensuring that all 5 selected entity types are different. I don't think we're showing already-listed entity types in the selector list, as I wasn't able to add the entity type an additional time, but we should also make sure this is the case. Maybe removal/re-adding causes duplication? I haven't tried to debug why, so will leave to you.

@nonparibus I'm not sure how it was possible you got into this state, perhaps it's related to the deployment not having the necessary backend changes, or if you tested the PR before it was ready. Although it should not be possible to select an entity type in the selector that has already been selected, as an additional precaution I've added a filter for duplicates in the update function in https://github.com/hashintel/hash/pull/3338/commits/70dac57bb729013da7e4e858049716e15082302e.

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx]: Good idea, have done so in https://github.com/hashintel/hash/pull/3338/commits/ebe7262785fe3b48e66d9e4966b52014a2393208

apps/hash-frontend/src/pages/[shortname].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page.tsx]: Ah I get it, apologies for misunderstanding I should've thought of that. Agree this would simplify things, have addressed in https://github.com/hashintel/hash/pull/3338/commits/207953360b4752b0ede415934f7f88c48fad71f3

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:282]: To be honest mainly because I had used it before, and I thought `@dnd-kit` didn't support the animations when rearranging but having taken another look at the docs I was wrong about that. It may have been a better choice, would you like me to replace it? Or do it in follow-up?

Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:282]: Have created the ticket: https://linear.app/hash/issue/H-1013/replace-react-beautiful-dnd-with-dnd-kit-in-editpinnedentitytypesmodal

apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/profile-page-tabs.tsx]: We spoke about this and figured out it was because you have overlay scrollbars in Chrome and I have the solid gray ones (that shift the layout). Applying `overflow-y: hidden` to the tabs (somewhat hackily) fixes it

apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/edit-pinned-entity-types-modal.tsx:282]: Thank you

CiaranMn (approved):
:shipit:

PR #3348, opened 2023-10-07
H-965: Move Rust modules to `lib/` crates

TimDiekmann:
## 🌟 What is the purpose of this PR?

Since the CI is able to properly run dependent crates it's finally possible to move out workspace subcrates to `lib/`.

## 🔍 What does this change?

- Moves out the following subcrates:
    - `graph-types` -> `@local/hash-graph-types-rs`
    - `codec` -> `@local/codec-rs` 
    - `temporal-versioning` -> `@local/temporal-versioning-rs`
    - `authorization` -> `@local/hash-authorization-rs` 
    - `hash-graph/test/test-data` -> `@tests/hash-graph-test-data-rs`
    - `hash-graph/test/tests` -> `@tests/hash-graph-integration-rs`
- Simplify a module when querying the SpiceDB server (I didn't mean to commit this in this PR but now it's here and we want that anyway)
-  Fixes a few dependency issues
- Fixes an issue with lint configuration not bein copied when pruning the repository

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing
### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

PR #3349, opened 2023-10-07
H-966: Reverse logic to select clippy lints

TimDiekmann:

Currently, we pick lints which should be warned. Many lints were moved out from clippy::restrictions so it's now easier to opt-out of lints.

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/3348

This PR:

The changes in this PR:

The changes in this PR:


.config/lints.toml
```diff
@@ -1,9 +1,57 @@
...
+    "clippy::tests_outside_test_module",
+    "clippy::unimplemented",
+    "clippy::unreachable",
+
+    # Should be enabled:
+    "clippy::allow_attributes",
+    "clippy::allow_attributes_without_reason",
+    "clippy::allow_attributes_without_reason",
+    "clippy::arithmetic_side_effects",
+    "clippy::as_conversions",
```
indietyp, line 41: imo this is something which would be worth warning about

.config/lints.toml
```diff
@@ -1,9 +1,57 @@
...
+    "clippy::unneeded_field_pattern",
 
-    # The lint is great, but sadly it has too many false positives
-    "clippy::significant_drop_tightening",
+    # Consider enabling:
+    "clippy::absolute_paths",
+    "clippy::default_numeric_fallback",
+    "clippy::impl_trait_in_params",
+    "clippy::min_ident_chars",
+    "clippy::multiple_unsafe_ops_per_block",
```
indietyp, line 27: in favor of enabling

.config/lints.toml
```diff
@@ -1,9 +1,57 @@
...
+    "clippy::separated_literal_suffix", # opposite of `clippy::unseparated_literal_suffix`
+    "clippy::shadow_reuse",
+    "clippy::shadow_same",
+    "clippy::unneeded_field_pattern",
 
-    # The lint is great, but sadly it has too many false positives
-    "clippy::significant_drop_tightening",
+    # Consider enabling:
+    "clippy::absolute_paths",
+    "clippy::default_numeric_fallback",
```
indietyp, line 24: in favor of enabling

indietyp (dismissed):

.config/lints.toml
TimDiekmann: This can get pretty verbose if you do more with numbers but generally, I tend to agree. But this will also require a `_i32` on a simple `+= 1;` assignment which is quite noisy.

indietyp: I thought this would only ever warn when one uses a literal in a case where the default would've been used, e.g. where the type is unclear, but if it's for every literal I agree, it'll be too noisy.

indietyp (approved):

PR #3359, opened 2023-10-09
H-991: Allow a `read/write` permission in addition to the `owner` relation

TimDiekmann:

Only an `owner` and `viewer` is simply not enough.

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/3358

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -670,6 +670,20 @@ export const removeEntityOwner: ImpureGraphFunction<
...
 
+export const addEntityEditor: ImpureGraphFunction<
+  { entityId: EntityId; viewer: AccountId | AccountGroupId },
+  Promise<void>
+> = async ({ graphApi }, { actorId }, params) => {
+  await graphApi.addEntityEditor(actorId, params.entityId, params.viewer);
+};
+
+export const removeEntityEditor: ImpureGraphFunction<
+  { entityId: EntityId; viewer: AccountId | AccountGroupId },
```
CiaranMn, line 681: Should these `viewer` be `editor`? Same for the owner functions above (missed them last time, sorry)

apps/hash-graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -588,6 +600,137 @@ where
...
 }
 
+#[utoipa::path(
+    post,
+    path = "/entities/{entity_id}/editors/{editor}",
+    tag = "Entity",
+    params(
+        ("X-Authenticated-User-Actor-Id" = AccountId, Header, description = "The ID of the actor which is used to authorize the request"),
+        ("entity_id" = EntityId, Path, description = "The Entity to add the editor to"),
+        ("editor" = OwnedById, Path, description = "The editor to add to the entity"),
```
CiaranMn, line 610: Not for this PR: this is one of the places that `OwnedById` as a type is potentially confusing... dunno what to do about it though since any other type would just be an alias for this. Potentially we could rename the type to something else, and just accept that the `ownedById` field would take an `accountIdOrAccountGroupId` (or whatever)

CiaranMn (commented):
I think a field needs renaming (`viewer` -> `editor`/`owner`) but otherwise looks good!

apps/hash-graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: I agree that this is confusing. I didn't want to introduce a new type in this PR solely for this purpose but will do that in one of my follow-ups.

CiaranMn (approved):

PR #3360, opened 2023-10-09
H-908: Add entity inference to browser extension

CiaranMn:

1. Introduce a feature whereby users can (via LLM) have entities suggested for creation from the webpage they are viewing, and then create them in HASH
2. Store extension state in session storage (user, draft notes, entity inference/creation progress) so that it's not lost when closing the popup
3. Show badge on extension when entity inference/creation is pending, errored, successful

## 🔍 What does this change?

- introduce UI for entity inference
- do work in background script for entity inference/creation process, including setting the extension's badge
- store everything worth persisting in session storage, including the existing user and draft quick note, and new feature state
- upgrade `next` to a canary version to get rid of the MUI barrel warnings when compiling the frontend (really spammy and annoying on start up)
- remove some redundant stuff from the webpack config
- address feedback from #3310 

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The production build for the extension has no output. Not sure what's going on there, will return to it when we want to publish it.

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. Run `yarn dev` in the `apps/plugin` folder
2. `chrome://extensions`, dev mode, load unpacked, select folder

### State persistence, error state, success state

https://github.com/hashintel/hash/assets/37743469/dd149636-b324-405a-aa9d-26ad7a9c8d1e

### Bonus

Demo entity inference from HN thread. You can see how running this on a huge thread of interest would be good to pick out companies/people of interest. Skip to `0:40` for results (left the beginning in for context, would love if Kap had a 'speed up this section' feature)

https://github.com/hashintel/hash/assets/37743469/66013698-890f-458c-90f0-697d1a772162


apps/plugin-browser/src/shared/badge.ts
```diff
@@ -0,0 +1,21 @@
+import browser from "webextension-polyfill";
+
+export const clearBadge = () => {
+  void browser.action.setBadgeText({ text: "" });
+};
+
+export const setLoadingBadge = () => {
+  void browser.action.setBadgeText({ text: "💭" });
+  void browser.action.setBadgeBackgroundColor({ color: "#006DC3" });
```
benwerner01, line 9: It should be possible to import this value from the design system package:

```suggestion
  void browser.action.setBadgeBackgroundColor({ color: customColors.blue[80] });
```

apps/plugin-browser/src/shared/badge.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const setErroredBadge = () => {
+  void browser.action.setBadgeText({ text: "❌" });
+  void browser.action.setBadgeBackgroundColor({ color: "black" });
+};
+
+export const setSuccessBadge = (num: number) => {
+  void browser.action.setBadgeBackgroundColor({ color: "#78B040" });
```
benwerner01, line 18:
```suggestion
  void browser.action.setBadgeBackgroundColor({ color: customColors.lime[70] });
```

apps/plugin-browser/src/scripts/background/infer-entities.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const inferEntities = async (message: InferEntitiesRequest) => {
+  void setInSessionStorage("inferenceStatus", { status: "pending" });
+  setLoadingBadge();
+
+  try {
+    const inferredEntities = await inferEntitiesApiCall(
+      message.textInput,
+      message.entityTypeIds,
+    );
```
benwerner01, line 62: I'm not sure if there's a limitation I'm not thinking of, but would it be possible to slice the text here into chunks if it exceeds the maximum supported length (i.e. 7900 character), make multiple calls to `inferEntitiesApiCall` and then combine the results?

apps/plugin-browser/src/pages/popup/popup-contents/action-center/text-field-with-dark-mode.tsx
```diff
@@ -1,24 +1,33 @@
...
+} from "../../../shared/dark-mode-values";
+
 export const TextFieldWithDarkMode = (props: TextFieldProps) => {
   return (
     <TextField
       {...props}
       InputProps={{
-        sx: ({ palette }) => ({
+        sx: () => ({
           "@media (prefers-color-scheme: dark)": {
```
benwerner01, line 17: Maybe not worth addressing, but we could set `palette.mode` to `"dark"` in the theme we pass to the `ThemeProvider`, so we can check the value of this boolean rather than using a media query every time we want to make styling dependent on dark mode. Here's [an example](https://mui.com/material-ui/customization/dark-mode/#system-preference) from their docs.

benwerner01 (approved):
Thanks Ciaran, this is working well for me. Only had a couple of thoughts from the code, none of which block merging.

apps/plugin-browser/src/scripts/background/infer-entities.ts
CiaranMn: We could, although we'd have to do some deduplication of the suggested entities (based on likely name properties) in case they appeared in multiple chunks. I'd rather leave this for follow up improvements once we've used it more.

apps/plugin-browser/src/shared/badge.ts
CiaranMn: Thank you, fixed in https://github.com/hashintel/hash/pull/3360/commits/f0a263efca4552a51e0c6924f635187caaf961e7

apps/plugin-browser/src/shared/badge.ts
CiaranMn: Thank you, fixed in https://github.com/hashintel/hash/commit/f0a263efca4552a51e0c6924f635187caaf961e7

apps/plugin-browser/src/pages/popup/popup-contents/action-center/text-field-with-dark-mode.tsx
CiaranMn: Thank you – I've created H-1029 to do this in a follow up.

PR #3366, opened 2023-10-10
H-697: Migrate from S3 to Cloudflare R2 for file object storage

CiaranMn:

Changes where we store files from S3 to R2.

## 🔍 What does this change?

- Adds environment variables for connecting to R2 (we were previously relying on the AWS IAM role for the API task having authorization)
- Update the file upload request from a presigned POST (not supported in R2) to a PUT
- Validate that the content type and length are as originally requested (this currently simply ensures that the type/length requested by the user for upload, which could be anything, are the same as what they actually upload, but at some point we could reject requests for upload with certain types/sizes, if we wanted)
- Remove Terraform config for S3 bucket

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

We probably want to change how we proxy file downloads to address the following: 
1. Currently, if the storage provider of a HASH Instance is changed, previously uploaded files will not be accessible (since retrieving them uses the current file storage provider configuration) – H-1010
2. It is possible to download files without being authorized to do so, if the file's URL is known (cannot be guessed but once known provides indefinite access to the file) – also H-1010
3. Allow for using a custom domain – H-1011

## 🛡 What tests cover this?

- None really. There's a file test but it doesn't test upload, it mocks it.

## ❓ How to test this?

1. You would need to set the environment variables in `.env.local`, including `FILE_UPLOAD_PROVIDER=AWS_S3`.

### DIFF
--- a/apps/hash-api/src/graph/knowledge/system-types/file.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/file.ts
@@ -12,7 +12,7 @@ import {
   MutationRequestFileUploadArgs,
 } from "../../../graphql/api-types.gen";
 import { AuthenticationContext } from "../../../graphql/context";
-import { PresignedPostUpload } from "../../../storage";
+import { PresignedPutUpload } from "../../../storage";
 import { genId } from "../../../util";
 import { ImpureGraphContext, ImpureGraphFunction } from "../..";
 import {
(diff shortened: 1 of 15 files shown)


benwerner01 (approved):
Thanks Ciaran, the changes look good to me and are working for me locally

CiaranMn:
@TimDiekmann if you could check that `terraform plan` works for you locally and looks good – I'll wait to merge until then. Thank you

PR #3367, opened 2023-10-10
H-867, H-1008: Implement User and Org customizable areas, supporting blocks

benwerner01:

This PR adds an editable bio to the profile page, supporting blocks similar to the page editor.

Drive-by changes:

- [Move "View all pages ->" into the "Page >" section](https://linear.app/hash/issue/H-1008/move-view-all-pages-into-the-page-section)

## 🔗 Related links

- H-867
- H-1008

## 🔍 What does this change?

- A new `BlockCollection` entity type has been introduced, which both the `Page` and the new `ProfileBio` entity types inherit from. `BlockCollection` defines the outgoing `contains` link which links it to blocks.
- The existing `updatePageContents` GQL resolver has been replaced with `updateBlockCollectionContents`, decoupling its functionality from pages.
- The existing `save.ts` script has been modified to decouple its functionality from pages.
- The existing `PageBlock` component has been replaced by a `BlockCollection` component, which removes a variety of page-specific functionality to make it possible to render a collection of blocks outside of the page editor

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- When switching between editing/viewing state in the bio editor, the content briefly disappears. I'm not sure exactly why this is happening, but probably something to do with how prosemirror re-mounts itself when `readonly` is toggled.
- Because saving happens automatically, and there isn't currently a convenient way of triggering it in the parent component of `BlockCollection`, it's possible that if the user presses the "check" icon before the save has been persisted that the latest changes won't be reflected in the UI. The may still have saved, but the `fetchProfileBioContents` function may have been called before the save persisted.

## 🐾 Next steps

- [Store index on block collection contents link in properties instead of link metadata](https://linear.app/hash/issue/H-1030/store-index-on-block-collection-contents-link-in-properties-instead-of)

## 🛡 What tests cover this?

None

## ❓ How to test this?

⚠️ This PR requires backend changes to be applied to `main`, so using the Vercel deployment risks breaking the production environment 

1. Go to the profile page of the logged in user
2. Click the edit icon on the bio to edit the bio. Try inserting blocks, using `@` mentions, etc
3. Go to a page, and ensure the page editor looks and functions exactly like on `main`

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/cd80b6d1-e2be-44ae-90fc-95506ae8c61e


apps/hash-frontend/src/blocks/block-collection/block-view.tsx
```diff
@@ -280,10 +281,12 @@ export class BlockView implements NodeView {
...
               />
-              <CreateBlockCommentButton
-                blockEntityId={blockEntityId}
-                rootNode={this.rootNode}
-              />
+              {this.isCommentingEnabled ? (
+                <CreateBlockCommentButton
+                  blockEntityId={blockEntityId}
+                  rootNode={this.rootNode}
+                />
```
CiaranMn, line 288: This looks like it will enable commenting even if the page is in `readonly` mode... we should preserve the existing behaviour, which is to not show the commenting interface if the page is in `readonly`

apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
```diff
@@ -0,0 +1,371 @@
...
+        (aLinkEntity.linkData?.leftToRightOrder ?? 0) -
+          (bLinkEntity.linkData?.leftToRightOrder ?? 0) ||
+        aLinkEntity.metadata.recordId.entityId.localeCompare(
+          bLinkEntity.metadata.recordId.entityId,
+        ) ||
+        aLinkEntity.metadata.temporalVersioning.decisionTime.start.limit.localeCompare(
+          bLinkEntity.metadata.temporalVersioning.decisionTime.start.limit,
+        )
+      );
+    });
```
Comment (CiaranMn) [inline apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx:63]: This looks duplicative of `getBlockCollectionBlocks` in the API – I guess it's because we use a special-cased `page` query for pages, but not here?

libs/@local/hash-graphql-shared/src/queries/entity.queries.ts
```diff
@@ -0,0 +1,68 @@
+import { gql } from "@apollo/client";
```
CiaranMn, line 1: Can we delete the existing one of these in the frontend and import this one instead, to avoid duplicating it?

CiaranMn (commented):
Looks good, a few small suggestions

apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx:63]: Yes, I didn't think it makes sense to introduce a `BlockCollection` query with the `contents` resolvers in the same way we currently have it for pages, as it's my understanding we want to move away from these custom field resolvers and do more of this work in the FE. This function should probably be moved somewhere else when we no longer make use of field resolvers to query pages, but not sure when we'll get around to refactoring this.

libs/@local/hash-graphql-shared/src/queries/entity.queries.ts
benwerner01: Ah yes, forgot to do this. Have done so in https://github.com/hashintel/hash/pull/3367/commits/f237d7120c807777e5feb4eb05b2cd17022f34fd

apps/hash-frontend/src/blocks/block-collection/block-view.tsx
benwerner01: Thanks for catching, have addressed in https://github.com/hashintel/hash/pull/3367/commits/0772cc63551a4a9da4b55de6fae67a526aa40062.

apps/hash-frontend/src/blocks/block-collection/block-collection.tsx
```diff
@@ -3,31 +3,26 @@ import "prosemirror-view/style/prosemirror.css";
 import { useApolloClient } from "@apollo/client";
```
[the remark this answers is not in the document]
benwerner01: Yes, have done so in https://github.com/hashintel/hash/pull/3367/commits/2b7c23625233e021fec584a1f6ae5fd8d1d14769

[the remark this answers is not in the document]
CiaranMn: Thank you

CiaranMn (approved):

CiaranMn (approved):

PR #3368, opened 2023-10-10
H-938: Add app-wide error boundary with fallback message

CiaranMn:

We have an error boundary around blocks, showing an error message with 'reset' button if a block's code throws an error. But we didn't have one any higher than that in the app tree, meaning that any other crashes would tank the app in production, showing NextJS's default 'A client-side error has occured' screen.

This PR adds an error boundary around the page component with a fallback component, so that the rest of the UI is maintained and the user has a reset button and can see the error id (for identification in Sentry) and error message – see video in demo.

As a drive-by I have also fixed a console error we were gettin for "Invalid Sentry DSN: undefined".

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

I have freestyled the appearance of this component and it could look a lot better.

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. Throw an error somewhere in the code.

## 📹 Demo

Here I throw an error when the mention suggester is used. Note that the modal which appears first only happens in development – in production, only the error fallback component will be shown.

https://github.com/hashintel/hash/assets/37743469/b152a2c0-50a1-446b-9865-c5cba3b200e6


nonparibus (dismissed):
Functional LGTM; thanks!

apps/hash-frontend/src/pages/_app.page/error-fallback.tsx
```diff
@@ -0,0 +1,101 @@
...
+        We've been notified and will investigate — please click the button below
+        to reset the page to its initial state.
+      </Typography>
+      <Box sx={{ textAlign: "center" }}>
+        <Button onClick={resetError} size="small">
+          Reset the page
+        </Button>
+      </Box>
+      <Typography mt={2}>
+        You can also <Link href="/contact">contact us</Link>. If you do, please
```
benwerner01, line 76: We don't have a `/contact` page, is this in anticipation of having moved from `app.hash.ai` to `hash.ai`?

benwerner01 (approved):
Thanks Ciaran, looks good and is working for me locally 👍

apps/hash-frontend/src/pages/_app.page/error-fallback.tsx
CiaranMn: Oops, nope – I've updated the link in https://github.com/hashintel/hash/pull/3368/commits/88dfc66eab15fa87567a80b1d1c0abff46db8396, thank you

benwerner01 (approved):

PR #3373, opened 2023-10-11
H-1028: Disable realtime service by default

CiaranMn:

The `realtime` service reads database changes from a replica slot and adds them to a Redis queue, which can be subscribed to by other services which need to react to changes in the database. Previously, this included the collaborative editing ("collab") service, but currently it is only the service which syncs data changes in HASH back to external integrations.

With the introduction of authorization we need to figure out the permissions for this service. Until we do so, and until we return to developing external integrations, this PR disables `realtime` and the integration syncing function by default (it can be enabled by environment variable).

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/integrations/sync-back-watcher.ts
+++ b/apps/hash-api/src/integrations/sync-back-watcher.ts
@@ -40,38 +40,43 @@ export const createIntegrationSyncBackWatcher = async (
   const queue = new RedisQueueExclusiveConsumer(redisClient);
 
   const processQueueMessage = () => {
-    void queue.pop(queueName, null, async (item: string) => {
-      const message = JSON.parse(item) as Wal2JsonMsg;
+    queue
+      .pop(queueName, null, async (item: string) => {
+        const message = JSON.parse(item) as Wal2JsonMsg;
 
-      const entityEdition = entityEditionRecordFromRealtimeMessage(message);
+        const entityEdition = entityEditionRecordFromRealtimeMessage(message);
 
-      const entity = (
-        await graphApiClient
-          .getEntitiesByQuery(systemUserAccountId, {
-            filter: {
-              equal: [
-                { path: ["editionId"] },
-                { parameter: entityEdition.entityEditionId },
-              ],
-            },
-            graphResolveDepths: zeroedGraphResolveDepths,
-            temporalAxes: fullDecisionTimeAxis,
-          })
-          .then(({ data: subgraph }) =>
-            getRoots(subgraph as Subgraph<EntityRootType>),
-          )
-      )[0];
+        const entity = (
+          await graphApiClient
+            .getEntitiesByQuery(systemUserAccountId, {
+              filter: {
+                equal: [
+                  { path: ["editionId"] },
+                  { parameter: entityEdition.entityEditionId },
+                ],
+              },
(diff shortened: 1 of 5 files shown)


apps/hash-api/src/integrations/sync-back-watcher.ts
```diff
@@ -40,38 +40,43 @@ export const createIntegrationSyncBackWatcher = async (
...
-      sendEntityToRelevantProcessor(entity, graphApiClient);
+        sendEntityToRelevantProcessor(entity, graphApiClient);
 
-      return true;
-    });
+        return true;
+      })
+      .catch((err) => {
+        // eslint-disable-next-line no-console -- caught because this function loses ownership of the queue occasionally in dev
+        console.error(`Could not take message from queue: ${err.message}`);
```
CiaranMn, line 78: Drive-by: the API would occasionally crash in dev if left for a while because it temporarily lost access to the Redis queue for some reason. I'm not sure why this is but am adding this here so that it doesn't tank the API, when realtime sync is enabled.

libs/@local/hash-backend-utils/src/environment.ts
```diff
@@ -33,3 +33,5 @@ export const waitOnResource = async (
 
   logger?.debug(`${resource} is ready`);
 };
+
+export const realtimeSyncEnabled = !!process.env.ENABLE_REALTIME_SYNC;
```
benwerner01, line 37: Maybe let's add `ENABLE_REALTIME_SYNC` to the committed `.env` and set it to `false` so it's easier to re-enable this in development down the line. Then we can modify this check to:

```suggestion
export const realtimeSyncEnabled = process.env.ENABLE_REALTIME_SYNC === "true";
```

apps/hash-api/src/integrations/sync-back-watcher.ts
```diff
@@ -40,38 +40,43 @@ export const createIntegrationSyncBackWatcher = async (
   const queue = new RedisQueueExclusiveConsumer(redisClient);
 
   const processQueueMessage = () => {
-    void queue.pop(queueName, null, async (item: string) => {
-      const message = JSON.parse(item) as Wal2JsonMsg;
+    queue
```
benwerner01, line 43: Strange, I would've expected eslint to complain about `void` missing here. Is it not required when using `.catch`?


```suggestion
    void queue
```

benwerner01 (approved):
Thanks Ciaran, this is working for me locally. Left one small suggestion, but doesn't block merging.

apps/hash-api/src/integrations/sync-back-watcher.ts
CiaranMn: Correct - the requirements are that one of the following list applies:
> Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator

libs/@local/hash-backend-utils/src/environment.ts
CiaranMn: Done in https://github.com/hashintel/hash/pull/3373/commits/8c4346cbc92e67f58ec0fb66fe62179a27327be6

benwerner01 (approved):

PR #3375, opened 2023-10-11
Fix existing test failures debug

CiaranMn:

Debuggin what is going on with the Playwright test failures in #3369


CiaranMn:
This has served its purpose

PR #3376, opened 2023-10-11
H-980: Return all users and relations for a given entity

TimDiekmann:

Provides an interface to read all users and relations for the requested entity.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This is not how I want to implement reading relationships in the first place. This approach is very inflexible and lengthy. In addition, this will not work when introducing more features such as attribute-based permissions (caveats) which are required for more fine-grained permissions. Also, this approach does not scale well and is rather slow. However, the functionality is required for the share menu, so I decided to provide a function that is working.
The approach I wanted to use did not work out fully but can be extended. It's available in [this branch](https://github.com/hashintel/hash/tree/t/h-980-return-all-users-relations-for-a-given-entity-experiment).

### DIFF
--- a/apps/hash-api/src/graph/knowledge/primitive/entity.ts
+++ b/apps/hash-api/src/graph/knowledge/primitive/entity.ts
@@ -1,10 +1,8 @@
 import { VersionedUrl } from "@blockprotocol/type-system";
 import {
-  EntityRelation,
   EntityStructuralQuery,
   Filter,
   GraphResolveDepths,
-  VisibilityScope,
 } from "@local/hash-graph-client";
 import {
   currentTimeInstantTemporalAxes,
@@ -15,6 +13,7 @@ import {
   AccountId,
   BaseUrl,
   Entity,
+  EntityAuthorizationRelationship,
   EntityId,
   EntityMetadata,
   EntityPropertiesObject,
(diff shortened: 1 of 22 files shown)


apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -719,3 +721,11 @@ export const isEntityPublic: ImpureGraphFunction<
   Promise<boolean>
 > = async (ctx, _, params) =>
   canViewEntity(ctx, { actorId: publicUserAccountId }, params);
+
+export const getEntityRelations: ImpureGraphFunction<
```
benwerner01, line 725: I wonder whether it might be better to name this something less generic so that it's more clear that these "relations" are related to authentication? Maybe `getEntityAuthRelations`, or `getEntityAccessRelations`? cc @CiaranMn

CiaranMn: Yes, I would agree with making the name more specific. 

These might be better called `relationships` also... the relation is the target (e.g. a user), the relationship is the connection with a target and relation-type... maybe `getEntityAuthRelationships`? or `getEntityPermissionRelationships`?

apps/hash-graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -459,6 +469,59 @@ impl<'s> ToSchema<'s> for Viewer {
     }
 }
 
+#[derive(Serialize, ToSchema)]
+struct EntityRelationScope {
+    relation: EntityRelation,
+    scope: VisibilityScope,
+}
```
CiaranMn, line 476: Similar to the other comment, I think these are probably better described as `Relationships`... `DirectOwner` is a type of relationship... the user who owns the thing is the relation

apps/hash-api/src/graph/knowledge/primitive/entity.ts
TimDiekmann: The method name is aligned with the terminology used in SpiceDB, where you have "relations" and "permissions". Relations are the thing you can specify, such as "A user is a _viewer_ of an entity", and permissions are derived from that, such as "A user _can view_ an entity". The term "Relationship" is used for the tuple containing the subject, the object, and either permissions _and_ relations, but this function only returns the relations and the subjects. 

Because of this, I wouldn't call it `getEntityPermissionRelationships`. `Relationships` itself could also be confusing I think. Also, I wouldn't use `Auth` because it could be either authentication or authorization.

TimDiekmann: Discussed offline and we're going to use `getEntityAuthorizationRelationships`.

apps/hash-graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: I changed the naming to match the function.

benwerner01 (dismissed):
Thanks Tim, the changes look good to me 👍

CiaranMn (approved):

PR #3377, opened 2023-10-11
H-976: Allow switching pages/entities from public to private

benwerner01:

This PR adds the ability to toggle a page or entity to "Public" and "Private".

## 🔗 Related links

- H-976

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Go to a page or an entity, and click "Share" in the top right
2. Toggle the entity to "Public", and refresh the page to ensure it has persisted correctly
3. Toggle back to "Private" and ensure it has persisted correctly

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/6fc44f49-919e-4b98-aeb0-32c9c657f95c


nonparibus:
@benwerner01 Can we show the icon (e.g. lock/globe/etc) next to the word 'Share' (left-side) so it's obvious without clicking in what the state of the page/entity is? I've updated the designs [in Figma](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=1503%3A124874&mode=design&t=ZABnbcq628qWxpnk-1) (_internal link_) to reflect this.

benwerner01:
> Can we show the icon (e.g. lock/globe/etc) next to the word 'Share' (left-side) so it's obvious without clicking in what the state of the page/entity is? I've updated the designs [in Figma](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=1503%3A124874&mode=design&t=ZABnbcq628qWxpnk-1) (internal link) to reflect this.

Sure, addressed this in https://github.com/hashintel/hash/pull/3377/commits/1f5be19c634526d44cd2408bc9afce00b1840fc9

Now looks like:
image

image

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -355,3 +372,168 @@ export const inferEntitiesResolver: ResolverFn<
 
   return status.contents[0];
 };
+
+const parseGqlOwnerInput = (ownerInput: OwnerInput) => {
```
CiaranMn, line 376: I appreciate you aren't using this yet, but why go through the process of having the frontend need to know if it's an `accountId` or an `accountGroupId`, and then parse it back to a uuid here again, rather than just passing a `subjectId` or however-named uuid straight through?

CiaranMn: One alternative is to make the input have some kind of type enum, which would have the advantage of avoiding the `public: true` input, which looks a bit weird when removing the relationship. Something like:
```graphql
enum SubjectType {
  Public
  Account
  AccountGroup
}

type ViewerInput {
  type: SubjectType!
  subjectId: OwnedById // or some scalar that combines AccountId and AccountGroupId
}

type EditorInput {
  type: SubjectType!
  subjectId: OwnedById!
}
```

benwerner01: I would also prefer using an enum, but to my understanding enums are not supported when defining GQL inputs (see [this explanation](https://github.com/graphql/graphql-wg/blob/main/rfcs/InputUnion.md)).

I can probably unify `accountId` and `accountGroupId` in a `subjectId`, but because unions cannot be expressed I would need to introduce some scalar that represents `AccountId | AccountGroupId` so that the argument is typed correctly. Would you still prefer this?

benwerner01: I stumbled upon `@oneof` whilst looking into this, and it seems to be supported by our GQL setup. This would mean I can convert the types to:

```GraphQL
input OwnerInput @oneOf {
    accountId: AccountId
    accountGroupId: AccountGroupId
  }

  input EditorInput @oneOf {
    accountId: AccountId
    accountGroupId: AccountGroupId
  }

  input ViewerInput @oneOf {
    accountId: AccountId
    accountGroupId: AccountGroupId
    public: Boolean
  }
```

And could get rid of the `parseGqlXInput` methods I made as the generated types will enforce at least one is provided:

image

CiaranMn: > I would also prefer using an enum, but to my understanding enums are not supported when defining GQL inputs (see [this explanation](https://github.com/graphql/graphql-wg/blob/main/rfcs/InputUnion.md)).

You can use an enum as an input type – you cannot use unions of types.

I'm still not clear why the caller needs to know whether they're passing an `accountId` or an `accountGroupId` at all... nothing in the process cares which it is. By making it part of the argument we are forcing callers to know (or to find out) whether a subject is a user or an account.

I didn't know about `@oneof`, that's helpful!

benwerner01: Apologies I misunderstood what you suggested, I've unified the `OwnerInput` and `ViewerInput` into a single `AuthorizationSubjectId` for the inputs in https://github.com/hashintel/hash/pull/3377/commits/cf2a02b592fbc335d69d57145d08bef994bef67e. I still couldn't figure out a better way than to make this a scalar (because of the lack of union support), and also had to re-introduce the `parseGqlAuthorizationViewerInput` method to ensure that either a subject Id is provided when `kind !== "Public"`.

CiaranMn: Looks good, thank you!

CiaranMn (approved):

PR #3380, opened 2023-10-12
H-956: Update authorization checks in GraphQL API, improve typing

CiaranMn:

We have introduced authorization checks in the Graph (Rust) API, and therefore can remove some of the (incomplete) authorization checks that were happening in the GraphQL (Node) API, which acts as a bridge between the frontend and the Graph API (we will be phasing this out at some point as in many cases it's merely forwarding requests back and forth).

This PR:
1. Removes an unused `createUser` mutation and the middlewares that were gating it
2. Defaults users who haven't completed signup to act as a public user, so that we don't accidentally make requests with them as a user where we shouldn't.
3. Deletes other deprecated typedefs (mainly 'linked aggregation' related)
4. Adds typing to resolvers so that:
   - we don't add resolvers that aren't defined in the schema
   - we don't add things to the schema that aren't defined in the resolvers
   - we don't type resolvers incorrectly
5. Fix existing type errors in the resolvers

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- H-990: Update the GraphQL API to return permissions reported by the Graph API, and use those to check 'can user edi

## ❓ How to test this?

1. Check that logging in, signing up still works


apps/hash-api/src/graphql/create-apollo-server.ts
```diff
@@ -64,7 +64,16 @@ export const createApolloServer = ({
...
+        /**
+         * Users who haven't completed signup are only permitted to set a name and shortname.
+         * For all other purposes they are treated as an anonymous user, and so we default them to one here.
+         *
+         * We have middleware which also checks this, but this default ensures that forgetting to apply
+         * the middleware to a new route doesn't give an incomplete user permissions they shouldn't have.
+         */
+        actorId: ctx.req.user?.isAccountSignupComplete
+          ? ctx.req.user.accountId
+          : publicUserAccountId,
```
benwerner01, line 76: I'm not sure we about enforcing this here as we need to remember in the resolvers that the `actorId` might be unreliable if we are dealing with users who aren't signed up. I think I'd prefer just using the middleware check in the BE.

apps/hash-api/src/graphql/resolvers/index.ts
```diff
@@ -188,9 +184,10 @@ export const resolvers = {
     results: linkedAggregationResults,
   },
 
-  // New knowledge field resolvers
   Page: {
+    // @ts-expect-error –– the type requires 'blockChildEntity' inside the return, but we deal with it in a field resolver
     contents: pageContents,
+    // @ts-expect-error –– the type requires 'contents' to be returned here, but we deal with it in a field resolver
     parentPage: parentPageResolver,
   },
```
benwerner01, line 192: This would make the type definition of this object more complicated, but you could try and make use of the existing `UnresolvedBlockGQL` and `UnresolvedPageGql` types to get rid of these TS errors

benwerner01 (approved):
Thanks Ciaran, this is working for me locally. Had two minor thoughts, but neither should block merging.

apps/hash-api/src/graphql/create-apollo-server.ts
CiaranMn: I've updated how we handle this in https://github.com/hashintel/hash/pull/3380/commits/5582e75ac1ba815b11ae6149252646e999de37ae following @TimDiekmann's work in #3408 – this allows us to ensure users don't accidentally get permission to do things that they shouldn't (since we don't create them a web until they complete signup) while maintaining their proper actor id for all requests

apps/hash-api/src/graphql/resolvers/index.ts
CiaranMn: I'm going to leave this for now since I don't think these resolvers need to exist anyway (the frontend could figure it out from a subgraph) and don't want to sink more time into htem

benwerner01 (dismissed):

apps/hash-api/src/graphql/create-apollo-server.ts
CiaranMn: I had to update the tests in https://github.com/hashintel/hash/pull/3380/commits/6d84b5da16437321e36b4ad76f04ce503294c132, because they weren't setting the user's `preferredName` – good example of where the new system is catching things the old one didn't!

benwerner01 (approved):

PR #3383, opened 2023-10-12
H-913: Add non-exhaustive outline of unpublished libraries

TimDiekmann:

To easier see what libraries exist this starts a section on unpublished libraries in the `libs/` folder. I purposefully did that in `libs/` not in `libs/@local` so every library going to be listed in a single file which makes searching for a specific library easier.

### DIFF
--- a/libs/README.md
+++ b/libs/README.md
@@ -42,8 +42,6 @@ Although published to package managers, the following libraries were developed f
 | [@hashintel/design-system]       | TypeScript  | [npm](https://www.npmjs.com/package/@hashintel/design-system)       | To be written | A collection of styleguide-aligned reusable UI primitives for [HASH] and our [hash.ai] website       |
 | [@hashintel/block-design-system] | TypeScript  | [npm](https://www.npmjs.com/package/@hashintel/block-design-system) | To be written | A relatively unopinionated set of reusable UI primitives for use in building [Block Protocol] blocks |
 
-## Internal unpublished libraries
-
 The packages inside of [`@local`](./@local) are libraries used inside this repository which are **not** published to package managers. All of these libraries may be subject to breaking changes. External consumers should be especially careful when using or upgrading these.
 
 The following list is a non-exhaustive list of packages in `@local`:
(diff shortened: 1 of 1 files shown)


libs/README.md
```diff
@@ -42,6 +42,23 @@ Although published to package managers, the following libraries were developed f
 | [@hashintel/design-system]       | TypeScript  | [npm](https://www.npmjs.com/package/@hashintel/design-system)       | To be written | A collection of styleguide-aligned reusable UI primitives for [HASH] and our [hash.ai] website       |
 | [@hashintel/block-design-system] | TypeScript  | [npm](https://www.npmjs.com/package/@hashintel/block-design-system) | To be written | A relatively unopinionated set of reusable UI primitives for use in building [Block Protocol] blocks |
 
+## Internal unpublished libraries
+
```
TimDiekmann, line 46: Maybe we don't need that heading here?
```suggestion
```
Or maybe making it smaller?
```suggestion
# Internal unpublished libraries

```

nonparibus: I'd be happy with...

```suggestion
## Internal libraries

```

...the text beneath makes clear they are unpublished. I think the key thing to highlight anyway is that these are internal, and not intended to be relied on more widely.

libs/README.md
```diff
@@ -42,6 +42,23 @@ Although published to package managers, the following libraries were developed f
...
+The packages inside of [`@local`](./@local) are libraries used inside the repository and are not published to package managers. all of these libraries may be subject to breaking changes. External consumers should be especially careful when using or upgrading these.
```
nonparibus, line 47:
```suggestion
The packages inside of [`@local`](./@local) are libraries used inside this repository which are **not** published to package managers. All of these libraries may be subject to breaking changes. External consumers should be especially careful when using or upgrading these.
```

libs/README.md
```diff
@@ -42,6 +42,23 @@ Although published to package managers, the following libraries were developed f
...
+
+The following list is a non-exhaustive list of packages in `@local`:
+
+| Package                                                         | Language(s) | Docs URL      | Description                                                                                                       |
+| --------------------------------------------------------------- | ----------- | ------------- | ----------------------------------------------------------------------------------------------------------------- |
+| [@local/codec-rs](@local/codec)                                 | Rust        | Not hosted    | Implementation of different `serde` or byte codes used in HASH                                                    |
+| [@local/hash-authorization-rs](@local/hash-authorization)       | Rust        | Not hosted    | Provides the Authorization interface and logic used in the Graph                                                  |
+| [@local/hash-graph-client](@local/hash-graph-client/typescript) | Typescript  | To be written | A generator to create a TypeScript/JavaScript client for the Graph API                                            |
```
nonparibus, line 55:
```suggestion
| [@local/hash-graph-client](@local/hash-graph-client/typescript) | TypeScript  | To be written | A generator to create a TypeScript/JavaScript client for the Graph API                                            |
```

libs/README.md
TimDiekmann: The section above is already called "Internal libraries" 😄

nonparibus (approved):

PR #3392, opened 2023-10-13
H-977, H-978: Allow assigning view/edit access to users and webs of users

benwerner01:

This PR adds the ability to assign edit/view access to specific users and orgs.

## 🔗 Related links

- H-977
- H-978

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Currently modifying the permissions of a page does not mean that the blocks and linked entities in the page inherit the updated permissions (related to H-829)

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Go to a page or entity and click "Share" on the top right
2. Use the UI to invite users or orgs to the entity, and modify the permissions
3. Observe that there is a new "Shared with others" state in the status button once the entity has been shared

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/ef0894fe-5ada-4448-b848-e212529df4dc


apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -489,3 +492,34 @@ export const isEntityPublicResolver: ResolverFn<
     entityId,
   });
 };
+
+export const getEntityAuthorizationRelationshipsResolver: ResolverFn<
+  Promise<EntityAuthorizationRelationship[]>,
```
CiaranMn, line 497: `ResolverFn` already wraps the result in a `Promise` (noticed since I started typing these)
```suggestion
  EntityAuthorizationRelationship[],
```

apps/hash-frontend/src/components/hooks/use-users-with-links.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import {
+  QueryEntitiesQuery,
+  QueryEntitiesQueryVariables,
+} from "../../graphql/api-types.gen";
+import { queryEntitiesQuery } from "../../graphql/queries/knowledge/entity.queries";
+import { constructUser, User } from "../../lib/user-and-org";
+
+/**
+ * Retrieves a specific set of useranizations, with their avatars and members populated
```
CiaranMn, line 20:
```suggestion
 * Retrieves a specific set of users, with their avatars populated
```

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -489,3 +492,34 @@ export const isEntityPublicResolver: ResolverFn<
...
+        : relation === "direct_owner"
+        ? EntityAuthorizationRelation.Owner
+        : EntityAuthorizationRelation.Viewer,
+    subject:
+      subject.type === "accountGroupMembers"
+        ? { accountGroupId: subject.id as AccountGroupId }
+        : subject.type === "account"
+        ? { accountId: subject.id as AccountId }
+        : { public: true },
+  }));
```
CiaranMn, line 524: Still not sure about doing it with `public: true` (looks stranger to me than `type: "public"`, and you can't easily do `subject.public` anyway), nor separately specifying `accountId` and `accountGroupId` (as it makes it more complicated to gather the subject ids elsewhere, and we don't care whether it's a user or an org anywhere)... let's bear this in mind as a potential simplification for future work here, maybe when we add share links, which will have some kind of code associated with the subject we'll need to factor in.

apps/hash-frontend/src/components/hooks/use-users-with-links.ts
```diff
@@ -0,0 +1,83 @@
...
+      constrainsPropertiesOn: { outgoing: 0 },
+      constrainsLinksOn: { outgoing: 0 },
+      constrainsLinkDestinationsOn: { outgoing: 0 },
+      inheritsFrom: { outgoing: 0 },
+      isOfType: { outgoing: 0 },
+      // These depths are chosen to cover the following:
+      // 1. the user's avatar (user -> [hasLeftEntity incoming 1] hasAvatar [hasRightEntity outgoing 1] -> avatar)
+      // 2. the user's user memberships (user <- [hasLeftEntity outgoing 1] userMembership [hasRightEntity incoming 1] <- user)
+      hasLeftEntity: { incoming: 1, outgoing: 1 },
+      hasRightEntity: { incoming: 1, outgoing: 1 },
```
CiaranMn, line 57: We aren't using the user's org memberships so I'd remove fetching it until we need it

apps/hash-frontend/src/pages/shared/top-context-bar/share-dropdown-menu/invite-account-form.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const { orgs } = useOrgsWithLinks({
+    orgAccountGroupIds: minimalOrgs?.map((org) => org.accountGroupId),
+  });
+  const { users } = useUsersWithLinks({
+    userAccountIds: minimalUsers?.map((user) => user.accountId),
+  });
+
+  const options = useMemo(
+    () =>
+      [...(users ?? []), ...(orgs ?? [])].filter(
```
CiaranMn, line 47: You could take `minimalUsers/minimalOrgs` if `orgs`/`users` isn't yet loaded so that the data can load quicker, then avatars can appear when the full ones load... since this is fetching _all_ users and orgs with stuff linked to them, it's going to take a while to load when there is any amount of data in the real app

CiaranMn (commented):
Thanks @benwerner01, the sharing dialog works well. I've said separately that we need to agree who can view / edit permissions, and then hide the dialog appropriately (at the moment people who can't edit permissions can see it, and are given dropdown options which cause an error)

apps/hash-frontend/src/components/hooks/use-users-with-links.ts
nonparibus: useranizations! A find and replace favorite ❤️

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
benwerner01: I did consider having something like:
```GQL
enum AuthorizationSubjectType {
  Account
  AccountGroup
  Public
}

type AuthorizationSubject {
  subjectId: SubjectId
  type: AuthorizationSubjectType!
}

type EntityAuthorizationRelationship {
  objectEntityId: EntityId!
  relation: EntityAuthorizationRelation!
  subject: AuthorizationSubject!
}
```

But the generated TS types in the FE would not be able to let TS infer that `subjectId` is non-nullable when `subject.type === "Account"`, or `subject.type === "AccountGroup"`. Is this something you'd rather we sacrifice than go with the approach I've taken here?

CiaranMn: No, let's not worry about it more for now, we can think about it again when we add more of these objects

apps/hash-frontend/src/pages/shared/top-context-bar/share-dropdown-menu/invite-account-form.tsx
benwerner01: Good idea, have addressed in https://github.com/hashintel/hash/pull/3392/commits/84788ad6de6361189b87f99fa3966b202f2eade3

apps/hash-frontend/src/components/hooks/use-users-with-links.ts
benwerner01: Good point, have addressed in https://github.com/hashintel/hash/pull/3392/commits/750938dadd3a2632adf054bea797b6bdc4129b63

benwerner01: Also noticed we previously didn't make avatars public when they are uploaded, fixed in https://github.com/hashintel/hash/pull/3392/commits/fde36a03b4003b22e64e5f0199bffacb706f73c7

CiaranMn (dismissed):

nonparibus (approved):

PR #3260, opened 2023-09-26
H-480: Scaffold `hashdotai-new` app, upgrade Next and MUI

CiaranMn:
## 🌟 What is the purpose of this PR?

We will be incrementally migrating the site at `https://hash.ai` from a legacy private codebase to this repo.

This PR adds the scaffold for a new NextJS site, at `apps/hashdotai-new`. We will build new pages here and serve them under specific `hash.ai/xxx` routes as they become available. This site uses the new(ish) NextJS App Router approach. 

This PR also takes the opportunity to upgrade Next and MUI everywhere, since we're fairly behind on Next, and there are various fixes to do with the interplay between each and the new App router/Server Components we can benefit from.

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] affected the execution graph, and the `turbo.json`'s have been updated to reflect this

## ⚠️ Known issues

- NextJS has a bunch of build warnings related to MUI, described fully in the new `README.md` in `apps/hashdotai-new`. If these become particularly annoying we can downgrade NextJS to `15.4.x` until they are fixed, but it would be nice to have the DX benefits of `15.5.x` ([various speed increases promised](https://nextjs.org/blog/next-13-5)).

## 🐾 Next steps

- Once this is merged, add a Vercel project for `hashdotai-new`

## 🛡 What tests cover this?

- Some basic ones for `hash-frontend`, none otherwise.

## ❓ How to test this?

1. Check that `hashdotdev` and `hash-frontend` still work (can be done on the deployment)
2. Check that `yarn dev` works locally for the new package

PR #3415, opened 2023-10-17
H-1059, H-1097: Add integration list data, remove `hashdotai-new` folder

CiaranMn:

Two things related to implementation of new pages for https://hash.ai:
1. Remove the `hashdotai-new` folder introduced in #3260, as it is now being prepared in a different repo
2. Add a JSON file containing a list of integrations (mostly dummy data for now), to use in preparing the integrations page

This PR:

The changes in this PR:

The changes in this PR:


apps/hashdotai/integrations/integration-list.json
```diff
@@ -0,0 +1,142 @@
...
+    "connectUrl": "/integrations/linear/connect",
+    "aboutParas": [
+      {
+        "header": "Realtime two-way sync",
+        "copyHtml": "Create, update or delete data in either Linear or HASH while keeping them in sync.",
+        "iconUrl": "https://dummyimage.com/160x160/ddd/000000&text=two-way"
+      },
+      {
+        "header": "Jointly maintained by Linear & HASH",
+        "copyHtml": "<div><strong>This integration is officially supported by both Linear and HASH.</strong></div><div>It is jointly maintained to ensure robust, reliable and realtime operation.</div>",
```
nonparibus, line 15: Per [this Slack comment](https://hashintel.slack.com/archives/C02NW18QGUR/p1697544637942309?thread_ts=1697542510.374379&cid=C02NW18QGUR), we want to define the acceptable `aboutParas` somewhere (outside of any one integration definition).

And we want to auto-insert the name of any given integration where it currently just says "Linear", rather than hard-code as here.

apps/hashdotai/integrations/integration-list.json
```diff
@@ -0,0 +1,142 @@
+[
+  {
+    "name": "Linear",
```
nonparibus, line 3: Appreciate this is a temp solution, but I think it would be easiest to have each integration in its own JSON file. We can defer that for now. I just think it's gonna become v cumbersome to add more than even a few of these in current form. Probably not worth addressing prior to just storing this data in HASH, though.

apps/hashdotai/integrations/integration-list.json
```diff
@@ -0,0 +1,142 @@
...
+    "supportedTypes": [
+      {
+        "title": "Linear Project",
+        "iconUrl": "https://dummyimage.com/160x160/ddd/000000&text=Project",
+        "properties": [
+          {
+            "title": "Logo",
+            "description": "The workspace logo (square)",
+            "iconUrl": "https://dummyimage.com/160x160/ddd/000000&text=Logo",
+            "type": "image",
```
nonparibus, line 30: @CiaranMn Pre-specifying these, versus providing the full type path/URL, and inferring the name/icon from that? Presume this is just a stopgap solution?

nonparibus (dismissed):
A few Qs/comments, non blocking.

apps/hashdotai/integrations/integration-list.json
CiaranMn: Yes, this is a stopgap solution – I imagine this whole thing will actually be composed of multiple types, entities etc that need to be stitched together for the same data

apps/hashdotai/integrations/integration-list.json
CiaranMn: Apologies, I forgot about that. What are the acceptable `aboutParas` to start with? Just something for the existing two, i.e. 'two-way' and 'jointly managed' or something like that?

apps/hashdotai/integrations/integration-list.json
CiaranMn: I would expect that we are not adding many more before moving the data into HASH

apps/hashdotai/integrations/integration-list.json
nonparibus: Sorry missed your reply here. Let's just go with these exact existing two as you suggest here for now. We'll add more as we encounter the need.

apps/hashdotai/integrations/integration-list.json
nonparibus: I think we'll want at least a ~dozen listed for when we switch over to this site (our next post-board milestone), so it depends when we envisage being able to store this info in HASH/what's actually blocking that. Just stable data? Also CDTs (enums)?

CiaranMn: Just stable data, probably. Most of the new ones will be ones to vote on, no? Which are small JSON objects which I think are probably easier to add to a single file than managing multiple, if anything. We can change it later though if not.

apps/hashdotai/integrations/integration-list.json
CiaranMn: Updated in https://github.com/hashintel/hash/pull/3415/commits/2f066b19ce9b48c7799bfcc659181439ebe13f44

nonparibus (approved):

PR #3422, opened 2023-10-18
H-990, H-1067: Use graph permissions in frontend, readonly org settings view

CiaranMn:

This PR updates the Node API and frontend so that the frontend can rely on the permissions from the graph when determining what UI to show users, rather than the placeholder checks implemented prior to proper authorization.

As part of this it introduces a readonly view to the org settings and member management pages, which were previously editable by anyone who could view them.

## 🔍 What does this change?

**Node API**
- Update GraphQL return types:
  - `Subgraph` becomes `SubgraphAndUserPermissions`
  - `Page` gets `userPermissions`
  - `Comment` gets `canUserEdit`
  - ideally we wouldn't treat these latter two differently but we're better off getting rid of the special treatment of them in the GraphQL API entirely rather than trying to do half of it now
- Remove old checks in API resolvers

**Frontend**
- use the new checks everywhere we need to check if something is editable
- Org settings
  - expose `[add|remove]AccountGroupMember` to FE and use when adding/removing members
  - readonly view for membership/settings page 

**Drive-bys**
- The `EntityTypeMismatchError` was wrong in various places, probably due to copy and paste errors

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- All org members can edit the org profile. This will be changed so that only org admins/owners can edit it.
- Sharing a page with another user won't work properly as it only currently shares the main `Page` entity, not the blocks it links to. Thus the page will crash for other users, as the block entities are missing.

## 🛡 What tests cover this?

- A couple of existing Playwright, not much coverage.

## ❓ How to test this?

1. Check that features for a single user still work as expected
2. Check that other users can / can't edit things as appropriate based on sharing settings

## 📹 Demo

**Sharing an entity with a user**

https://github.com/hashintel/hash/assets/37743469/7b81c8cb-ced4-4795-a8cc-ce6e9bc0f1c5

---

**Adding/removing an org member**

https://github.com/hashintel/hash/assets/37743469/68e2bb06-dd1d-458c-b3bb-f23118d38962


apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -717,6 +723,74 @@ export const canUpdateEntity: ImpureGraphFunction<
...
+  Promise<UserPermissions>
+> = async (graphContext, { actorId }, params) => {
+  const { entity } = params;
+
+  const { entityId } = entity.metadata.recordId;
+
+  const isAccountGroup =
+    entity.metadata.entityTypeId === SYSTEM_TYPES.entityType.org.schema.$id;
+
+  const isPublicUser = actorId === publicUserAccountId;
```
TimDiekmann, line 737: I guess this is to avoid making a dedicated request?

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -717,6 +723,74 @@ export const canUpdateEntity: ImpureGraphFunction<
...
+      ? false
+      : await canUpdateEntity(graphContext, { actorId }, { entityId }),
+    isAccountGroup
+      ? isPublicUser
+        ? false
+        : await graphContext.graphApi
+            .canAddGroupMember(actorId, extractEntityUuidFromEntityId(entityId))
+            .then(({ data }) => data.has_permission)
+      : null,
+  ]);
```
TimDiekmann, line 750: To simplify this I like to extend the current reading-interface to allow specifying filters. Things I can filter on the graph:
- entity id
- relation between entity and subject
- subject type (i.e. `account`, `accountGroup` or `public`)
- subject id (i.e. `accountId`, `accountGroupId`)
- subject relation (i.e. `member` for account groups)

If this would help I can do this as one of my next tasks.

libs/@local/hash-graph-client/python/poetry.lock
TimDiekmann, line 1: Diff seems fine to me

package.json
```diff
@@ -18,6 +18,7 @@
     "dev:backend:realtime": "turbo dev --log-order stream --filter '@apps/hash-realtime' --",
     "dev:backend:search-loader": "turbo dev --log-order stream --filter '@apps/hash-search-loader' --",
     "dev:frontend": "turbo dev --log-order stream --filter '@apps/hash-frontend' --",
+    "dev:graph": "RUST_LOG=info yarn workspace @apps/hash-graph dev",
```
TimDiekmann, line 21: Do we want to call it `start` instead? We don't have hot-reloading in Rust.

TimDiekmann:
I have **no** idea why `black` is failing. The changes were done in [`b2c3b5b` (#3419)](https://github.com/hashintel/hash/pull/3419/commits/b2c3b5b8bd7a0f74c6e41f55c9705ca58a28af98) today morning as part of an upgrade of `black` to version `23.10.0`. That version was also installed in the workflow. However, it seems that an old version of black is used when running the lints.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
CiaranMn: While public users can always only view entities, I thought it easier not to bother asking the Graph API anything about them. We can do though if you think safer, it wouldn't be a big deal

TimDiekmann: I think that is reasonable. I don't know how likely it is that this will change at some point, though. We just have to remember that this is encoded here.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
TimDiekmann: Oh, that's true. I think there is an API endpoint on spiceDB to do something like that but I wouldn't look into that now. Sorry for the noise

package.json
CiaranMn: I think it's probably easier to call it `dev` since we're also running `dev` for the frontend and Node API, and it saves having to remember 'oh yeah it's `start` for the graph'. I don't feel strongly about this though.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
CiaranMn: Okay, cool. I think we'll notice because if we ever try and introduce a feature where public users can edit something, it won't work until we change this.

package.json
TimDiekmann: I don't feel strongly either 🤷🏻‍♂️

benwerner01 (dismissed):
Thanks Ciaran, looks good to me and is working for me locally. Had some small thoughts that shouldn't block merging.

TimDiekmann (approved):

PR #3423, opened 2023-10-18
H-1091: Expose batch-modifications of relations

TimDiekmann:

As seen in [H-1070](https://linear.app/hash/issue/H-1070/fix-graph-api-crash-when-attempting-to-remove-entity-owner) it's quite easy to hit a data race. While that Issue may be fixed by retrying (which should be attempted in any case), it's good to mitigate the issue in the first place. This is possible by exposing a batch modification endpoint for relations. SpiceDB offers an endpoint where different write operations (even different types such as create or delete) can be mixed.

## 🔗 Related links

- H-1091

## 🚫 Blocked by

- #3403 
- #3421 

## 🔍 What does this change?

- Exposes a `modifyEntityAuthorizationRelationships` interface
- Replaces old method to change the relationships (does not remove it as the GraphQL resolver is still using these)

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -479,10 +482,19 @@ export const addEntityViewerResolver: ResolverFn<
 > = async (_, { entityId, viewer }, { dataSources, authentication }) => {
   const context = dataSourcesToImpureGraphContext(dataSources);
 
-  await addEntityViewer(context, authentication, {
-    entityId,
-    viewer: parseGqlAuthorizationViewerInput(viewer),
-  });
+  await modifyEntityAuthorizationRelationships(context, authentication, [
+    {
+      operation: "create",
```
CiaranMn, line 487:
```suggestion
      operation: "touch",
```

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -719,7 +723,7 @@ export const getEntityAuthorizationRelationships: ImpureGraphFunction<
       data.map(
         (relationship) =>
           ({
-            object: params.entityId,
+            object: { namespace: "entity", id: params.entityId },
```
CiaranMn, line 726: We discussed and will rename:
1. `namespace` to `kind` (for consistency with other discriminating fields, e.g. in the subgraph)
2. `id` to `objectId` / `subjectId` (avoid unqualified `id`s)

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -659,6 +660,23 @@ export const getLatestEntityRootedSubgraph: ImpureGraphFunction<
...
+    operation: ModifyRelationshipOperation;
+    relationship: EntityAuthorizationRelationship;
+  }[],
+  Promise<void>
+> = async ({ graphApi }, { actorId }, params) => {
+  await graphApi.modifyEntityAuthorizationRelationships(
+    actorId,
+    params.map(({ operation, relationship }) => ({
+      operation,
+      object: relationship.object.id,
```
CiaranMn, line 674: Let's call this `resource`, agreed it is marginally less mental effort, and less vague if it ends up deconstructed and separated from `subject`

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -455,19 +454,23 @@ export const removeEntityEditorResolver: ResolverFn<
...
+    return { namespace: "account", id: viewer as AccountId } as const;
   } else {
     if (!viewer) {
       throw new UserInputError("Viewer Account Group ID must be specified");
     }
-    return viewer;
+    return {
+      namespace: "accountGroup",
+      id: viewer as AccountGroupId,
+      set: "member",
```
CiaranMn, line 472: We agreed to remove `set` as it will always be `member` if it's an account group, and nothing if it isn't, so it isn't adding information (and naming is hard)

We can reintroduce if we get more possible `set` values (e.g. `admin`)

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
TimDiekmann: Thank you, applied in [`7e4aa7f` (#3423)](https://github.com/hashintel/hash/pull/3423/commits/7e4aa7f959fd2357bdfd9489cfa36ca5bae3a2db)

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -659,6 +660,23 @@ export const getLatestEntityRootedSubgraph: ImpureGraphFunction<
...
+      relationSubject: relationship,
```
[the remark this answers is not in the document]
TimDiekmann: Renaming done in [`fee822d` (#3423)](https://github.com/hashintel/hash/pull/3423/commits/fee822d7e3aac0c0f51f74139c3a78afa024a0f5)

libs/@local/hash-authorization/src/schema/entity.rs
```diff
@@ -254,6 +253,116 @@
...
+                    EntityDirectViewerSubject::Account { id } => (EntitySubject::Account(id), None),
+                    EntityDirectViewerSubject::AccountGroup { id, set } => {
+                        (EntitySubject::AccountGroup(id), Some(set))
+                    }
+                    EntityDirectViewerSubject::Public => (EntitySubject::Public, None),
+                },
+            ),
+        };
+        (relation, subject, subject_set)
+    }
```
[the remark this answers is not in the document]
TimDiekmann: Well, that's impressive 😄

apps/hash-api/src/graph/knowledge/primitive/entity.ts
TimDiekmann: - Renamed `namespace` to `kind` in [`d6089df` (#3423)](https://github.com/hashintel/hash/pull/3423/commits/d6089dfd288b8dfc291262968fc6c781864d2dcc)
- Renamed `id` to `objectId` and `subjectId`: [`17f8f18` (#3423)](https://github.com/hashintel/hash/pull/3423/commits/17f8f182f15bf63e0165ec914e5103c20c7dba4c) (I'm not 100% convinced, seems a bit verbose but it could be helpful in some contexts). cc @benwerner01

apps/hash-api/src/graph/knowledge/primitive/entity.ts
TimDiekmann: Renamed `object` to `resource`: [`d836690` (#3423)](https://github.com/hashintel/hash/pull/3423/commits/d836690e8ab320d7c1037ffd89bd25c33df4b9bf)

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -455,19 +454,22 @@ export const removeEntityEditorResolver: ResolverFn<
 const parseGqlAuthorizationViewerInput = ({
```
TimDiekmann, line 454: This function is the only location where `objectId` and `subjectId` are helpful (so far) and this function will be removed as the types will come directly from GraphQL. (just flagging, not feeling strongly but also not 100% convinced)

CiaranMn (dismissed):

CiaranMn (approved):

PR #3432, opened 2023-10-19
H-1014: Add Sentry to browser extension

CiaranMn:

Adds Sentry to the HASH browser extension / plugin.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

1. Cannot currently be used in the background script due to an issue in Sentry (tracked in H-1133)
2. The stack trace when errors come from `browser-polyfill` is only one frame. I have created H-1134 to investigate further.

## ❓ How to test this?

1. You'll need to set a SENTRY_DSN (available from the project in Sentry) in your `.env.local` in the plugin folder
2. Throw an error in the popup window or options screen and check it is sent to Sentry

## 📹 Demo

Screenshot 2023-10-19 at 11 32 15


apps/plugin-browser/package.json
```diff
@@ -32,6 +33,7 @@
     "@local/hash-isomorphic-utils": "0.0.0-private",
     "@local/hash-subgraph": "0.0.0-private",
     "@pmmmwh/react-refresh-webpack-plugin": "0.5.10",
+    "@sentry/webpack-plugin": "1.20.0",
```
benwerner01, line 36: With `1.20.0` I was getting a `sentryWebpackPlugin is not a function` error, I was able to fix this by upgrading to `2.9.0`.

benwerner01 (changes requested):
Thanks Ciaran, was able to get this working with the added environment variables but had to upgrade a package. We probably want to use `2.9.0` instead of `1.20.0`, unless there's a reason we were using the lower version.

benwerner01 (dismissed):
Thanks for finding the fix Ciaran, they're working for me locally 👍

benwerner01 (approved):

PR #3434, opened 2023-10-19
H-540: Add `Quick notes` page with ability to create new notes

benwerner01:

This PR adds a `/notes` page, where users are able to create end edit existing quick notes in their personal web.

## 🔗 Related links

- H-540

## 🔍 What does this change?

See commits.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- There are various [existing issues](https://linear.app/hash/issue/H-1032/improve-block-editing-functionality) with the `BlockCollection` component, which will be apparent when editing quick notes. Some of the most apparent ones:

  - Currently there isn't a way for the parent of the `BlockCollection` component to know when the entity is being saved, or has finished saving. This means that the "Create" button is currently not disabled when the content hasn't finished saving, which risks data-loss if the button is clicked too early. (H-1139)
  - The `BlockCollection` component requires the block collection to have already been created before the component is able to mount. This means that the input for creating a quick note cannot be autofocused on page-load, without creating the quick note. (H-1140)
  - Mounting the `BlockCollection` component is currently slow, which is apparent when rendering many instances on the same page. Therefore when the quick notes page renders many quick notes, it can take a noticeable amount of time for them all to be displayed even when the required data has been fetched. (H-1141)

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Checkout the branch, and run the HASH app locally
2. Click "Quick notes" in the sidebar, or go to `/notes`
3. Test out the functionality on the page (creating quick notes, editing existing ones, archiving, etc)

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/94e02e93-2f55-4485-bba5-6ce6358f4f6a


apps/hash-frontend/src/pages/notes.page/create-quick-note.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const createdQuickNoteEntity = await createBlockCollectionEntity({
+      kind: "quickNote",
+    });
+
+    const { data } = await getEntity({
+      data: {
+        entityId: createdQuickNoteEntity.metadata.recordId.entityId,
+        graphResolveDepths: {
+          hasLeftEntity: { incoming: 2, outgoing: 2 },
+          hasRightEntity: { incoming: 2, outgoing: 2 },
```
CiaranMn, line 87: Can we add a comment expalining what entities it needs these for? When we can be more selectively in what data we fetch it'll help

apps/hash-frontend/src/pages/notes.page/convert-quick-note-to-page-modal.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const { updateEntity } = useBlockProtocolUpdateEntity();
+
+  const { createEntity } = useBlockProtocolCreateEntity(
+    authenticatedUser.accountId as OwnedById,
+  );
+
+  const innerSubmit = handleSubmit(async (data) => {
+    const { parentPage } = data;
+
+    const title = data.title ?? defaultTitle;
```
CiaranMn, line 73: The `defaultTitle` is not being used, because `data.title` defaults to an empty string

apps/hash-frontend/src/pages/notes.page/today-section.tsx
```diff
@@ -0,0 +1,184 @@
...
+                latestQuickNoteEntityWithEmptyContents && quickNotesSubgraph
+                  ? quickNotesSubgraph
+                  : undefined
+              }
+              refetchQuickNotes={refetchQuickNotes}
+              onCreatingQuickNote={setCreatingQuickNote}
+            />
+            <Divider sx={{ borderColor: ({ palette }) => palette.gray[20] }} />
+            <Box display="flex" marginTop={2.25}>
+              <Box display="flex" alignItems="center" gap={1.5}>
```
CiaranMn, line 118: I suggest commenting this out until it does something specific to the note.. at the moment it's a bit weird because they are just links to create a new entity/type with no relation to the note being typed. Mentioned this to @nonparibus who is happy we hide this until it makes sense

apps/hash-frontend/src/pages/shared/block-collection/component-view.tsx
```diff
@@ -92,7 +92,7 @@ export class ComponentView implements NodeView {
     private readonly block: HashBlock,
     private readonly manager: ProsemirrorManager,
     private readonly readonly: boolean,
-    private readonly isInitialPageTitleEmpty: boolean,
+    private readonly autofocus: boolean,
```
CiaranMn, line 95: Appreciate the PR hasn't introduced this problem, but can we make the casing `autoFocus` everywhere, consistent with the React prop? It's only this casing in a few places in our ProseMirror-related code

libs/@local/hash-isomorphic-utils/src/save.ts
```diff
@@ -505,6 +508,17 @@ export const save = async (
...
       variables: { entityId: blockCollectionEntityId, actions },
       mutation: updateBlockCollectionContents,
+      refetchQueries: [
+        {
+          query: getEntityQuery,
+          variables: {
+            entityId: blockCollectionEntityId,
+            ...zeroedGraphResolveDepths,
+            hasLeftEntity: { incoming: 2, outgoing: 2 },
+            hasRightEntity: { incoming: 2, outgoing: 2 },
```
CiaranMn, line 518: Comment why these depths please... if it's important that these variables are the same as in the other place we have them, maybe better to share them rather than relying on keeping them in sync

apps/hash-frontend/src/pages/notes.page/timestamp-column.tsx
```diff
@@ -0,0 +1,57 @@
...
+    {heading ? (
+      <TimestampCollectionHeading>{heading}</TimestampCollectionHeading>
+    ) : null}
+    <TimestampCollectionSubheading>{subheading}</TimestampCollectionSubheading>
+    <Box display="flex" marginRight={-1}>
+      <IconButton disabled={!navigateUp} onClick={navigateUp}>
+        <ChevronUpRegularIcon />
+      </IconButton>
+      <IconButton disabled={!navigateDown} onClick={navigateDown}>
+        <ChevronDownRegularIcon />
```
CiaranMn, line 46: This is enabled for me even without any other day to travel to

apps/hash-frontend/src/pages/notes.page/editable-quick-note.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+  const textBlocks = useMemo(
+    () =>
+      mostRecentContents?.filter(({ rightEntity }) =>
+        isHashTextBlock(rightEntity.componentId),
+      ),
+    [mostRecentContents],
+  );
+
+  const text = useMemo(
```
CiaranMn, line 173: This seems to undercount words.. maybe something to do with the general bugginess of save (not introduced by this PR), since it appears to be counting only the words that get persisted. Possibly blocks are being saved before all words in them are typed and then not updated again

https://github.com/hashintel/hash/assets/37743469/b5d303e5-285e-4a83-bb8e-c39c6232f8c9

CiaranMn (commented):
Thanks @benwerner01, a few suggestions. The main issue with this working well is the existing bugginess of ProseMirror saving which I don't think this PR has anything to do with, although the introduction of word counts makes it even more obvious – see video in a comment

apps/hash-frontend/src/pages/notes.page/editable-quick-note.tsx
benwerner01: You're right that the word count will only be updated once the update has persisted, this is because the parent component relies on a GQL refetch query call to be made by the `save.ts` function to get the updated page contents.

Without a better alternative, I'm not sure how I can get the word count to reflect what the user is currently typing. The calculation of how many words itself though should be robust, the reason the number of words decreases when you clicked "Create" in your recording is because the most recent update did not get persisted (another issue I flagged in the description, as I can't disable the form from submitting before the last save has completed).

apps/hash-frontend/src/pages/notes.page/convert-quick-note-to-page-modal.tsx
benwerner01: Thanks for catching, have addressed in https://github.com/hashintel/hash/pull/3434/commits/0f11122e4bff38d8180c694be32d9414b1ba720e

apps/hash-frontend/src/pages/notes.page/create-quick-note.tsx
benwerner01: Sure, it's to get the blocks and the block text data. Have added in https://github.com/hashintel/hash/pull/3434/commits/297ccde87f2f3666fcdf922b8425b13b80c5e18e

apps/hash-frontend/src/pages/notes.page/today-section.tsx
benwerner01: Alright, have uncommented it for now in https://github.com/hashintel/hash/pull/3434/commits/f576ac5bba2539a3204ac3444beb694bae7f7663

apps/hash-frontend/src/pages/notes.page/convert-quick-note-to-page-modal.tsx
benwerner01: Oh right yeah, have fixed in https://github.com/hashintel/hash/pull/3434/commits/4dcf1557220a87a2371a3bde5eb0392a348e1874

apps/hash-frontend/src/pages/notes.page/timestamp-column.tsx
benwerner01: Fixed in https://github.com/hashintel/hash/pull/3434/commits/b53ffb23c63d1036b08c689daca412d202778704

libs/@local/hash-isomorphic-utils/src/save.ts
CiaranMn: The comment indicates that the required depths are:
```
            hasLeftEntity: { incoming: 2, outgoing: 0 },
            hasRightEntity: { incoming: 0, outgoing: 2 },
```
so at the moment we're overfetching data (things that link _to_ the page rather than being linked from it), unless there's a reason we need such links

libs/@local/eslint-config/legacy-base-eslintrc-to-refactor.cjs
```diff
@@ -109,7 +109,7 @@ module.exports = {
     "default-case": "off",
     "class-methods-use-this": "off",
     "react/no-unescapted-entities": "off",
-    "jsx-a11y/no-autofocus": "off",
+    "jsx-a11y/no-autoFocus": "off",
```
CiaranMn, line 112: The name of the rule is `"jsx-a11y/no-autofocus"`, which is why an eslint error has appeared in the block – we were previously turning the rule off. This change means we are _not_ turning the rule off – we are turning off a rule that doesn't exist. If we revert this change we can remove the eslint disable line added.
```suggestion
    "jsx-a11y/no-autofocus": "off",
```

benwerner01: Ah that makes sense, thanks for catching this. Have addressed in https://github.com/hashintel/hash/pull/3434/commits/d5244c63f3f80232b507261855b0452297223176

CiaranMn (approved):

PR #3442, opened 2023-10-20
H-1138: Use same authorization relationship API for account groups and webs as for entity

TimDiekmann:

The authorization for entities were changed to be more generic. This also simplifies the interface for Web and AccountGroups.

## 🔍 What does this change?

- [Consistently rename namespace to kind](https://github.com/hashintel/hash/commit/356c7728dc3bc3a78176c156e0688ae672c0aaf8)
- [Flatten conversion methods](https://github.com/hashintel/hash/commit/11223a03cfb230c0afe933d89838566b84c4ba2f)
- [Clean up entity relationship definitions](https://github.com/hashintel/hash/commit/a50d57c22cdb4910fedd4647f16e6bee5a37d15a)
- [Make Web authorization API generic](https://github.com/hashintel/hash/commit/0025562881db4b0f820e98e059d1e4402bcb2f33)
- [Make AccountGroup authorization API generic](https://github.com/hashintel/hash/commit/4f4121d3d11a023b4fb1ba99fec30c961eb0940a)

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/lib/graph/src/api/rest/entity.rs
```diff
@@ -658,9 +664,9 @@ where
             StatusCode::INTERNAL_SERVER_ERROR
         })?;
 
-    let subject = match owner_id {
-        OwnerId::Account(id) => EntityDirectOwnerSubject::Account { id },
-        OwnerId::AccountGroupMembers(id) => EntityDirectOwnerSubject::AccountGroup {
+    let owner = match owner_id {
+        WebSubject::Account(id) => EntityDirectOwnerSubject::Account { id },
```
CiaranMn, line 668: why not `WebSubjectId`? and `xId` generally when we are talking about ids and not the whole thing

CiaranMn (approved):
One non-blocking question... I notice in a lot of places in the graph we use terms like `owner`, `viewer`, `subject` when the value is actually just an id. elsewhere we would typically explicitly say `viewerId`, `subjectId` etc

apps/hash-graph/lib/graph/src/api/rest/entity.rs
TimDiekmann: I did not add `Id` to these terms as the ID _is_ the subject in the permission implementation. For `AccountGroup` this is `id` and `set` so this is no longer an ID only.

PR #3444, opened 2023-10-20
H-1145: Fix inviting users/orgs to view a page/entity

benwerner01:

This PR fixes being able to invite users/orgs to view a page or entity.

## 🔗 Related links

- H-1145

## 🔍 What does this change?

- Sets `kind` of the viewer relationship correctly
- Removes all casts from `Entity` to `Entity<XProperties>`, replacing them with type assertions

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

Go to the deployment, try to invite a user or an org to an entity/page. Their name should now display correctly

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/6b141c2e-087e-4481-8b18-7cea3b9e35e1

### DIFF
--- a/apps/hash-frontend/src/lib/user-and-org.ts
+++ b/apps/hash-frontend/src/lib/user-and-org.ts
@@ -5,7 +5,6 @@ import {
   OrgMembershipProperties,
   OrgProperties,
   ProfileBioProperties,
-  ServiceAccountProperties,
   UserProperties,
 } from "@local/hash-isomorphic-utils/system-types/shared";
 import {
@@ -33,7 +32,10 @@ import {
   intervalCompareWithInterval,
   intervalForTimestamp,
 } from "@local/hash-subgraph/stdlib";
-import { LinkEntity } from "@local/hash-subgraph/type-system-patch";
+import {
+  extractBaseUrl,
+  LinkEntity,
+} from "@local/hash-subgraph/type-system-patch";
 
 export const constructMinimalOrg = (params: {
   orgEntity: Entity<OrgProperties>;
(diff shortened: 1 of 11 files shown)


apps/hash-frontend/src/lib/user-and-org.ts
```diff
@@ -148,6 +148,12 @@ export const constructOrg = (params: {
 }): Org => {
   const { subgraph, orgEntity } = params;
 
+  if (orgEntity.metadata.entityTypeId !== types.entityType.org.entityTypeId) {
+    throw new Error(
+      `Entity with type ${orgEntity.metadata.entityTypeId} is not an org entity`,
+    );
+  }
+
```
TimDiekmann, line 156: How this is possible? If this is only possible via a bug it would be good to state that in a comment or remove it again (I guess this is how you found the bug?).

benwerner01: This was possible because we cast from `Entity` to `Entity<UserProperties>` (or `Entity<OrgProperties`) when we pass the entities returned by the structural queries to these functions. In this case it was possible for this bug to occur because the `useOrgWithLinks` and `useUserWithLinks` hooks were querying for entities via their `UUID` only, and did not include checks for the type. I've added these checks in https://github.com/hashintel/hash/pull/3444/commits/1613bf37eee44fad54cab7b1caaeb65daa160371 to make these hooks more robust, but still think having an error here could help catch when an incorrect cast has been made more easily.

TimDiekmann: Great, thank you!

TimDiekmann (approved):
LGTM, I think this will be much more robust, thank you! 🚀

PR #3458, opened 2023-10-23
H-1144: Prepare SpiceDB schema for caveats

TimDiekmann:

With caveats we expect a `general_X` (with `X` being a relation) to grant permissions which may be excluded by `excluded_X`. In addition, there will be a `direct_X`, which always has the specific permissions.

To be forward compatible the relations has to be changed.

## 🚫 Blocked by

- #3442 
- #3443

## 🔍 What does this change?

Renames the following relations:
- `DirectOwner` -> `Owner`
- `Direct*` -> `General*`

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (approved):
We're not sure about these names but since there are many PRs stacked on this let's just merge them and then rename afterwards

PR #3462, opened 2023-10-23
H-1149: Fix block collection mention suggester

benwerner01:

The mention suggester is broken because of an incorrect reference to the subgraph in a query response, and a runtime error caused by a GQL query since the changes in https://github.com/hashintel/hash/pull/3422.

## 🔗 Related links

- H-1149

## 🔍 What does this change?

See commits and comments.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

Try out the mention suggester in the deployment


apps/hash-frontend/src/pages/shared/block-collection/shared/mention-suggester.tsx
```diff
@@ -171,7 +171,7 @@ export const MentionSuggester: FunctionComponent<MentionSuggesterProps> = ({
     fetchPolicy: "cache-and-network",
   });
 
-  const entitiesSubgraph = data?.structuralQueryEntities as
+  const entitiesSubgraph = data?.structuralQueryEntities.subgraph as
     | Subgraph<EntityRootType>
     | undefined;
```
benwerner01, line 176: We should probably get rid of the numerous casts to `Subgraph<EntityRootType>` and make use of some `isEntityRootedSubgraph` type assertion instead, that way we can type the input as `Subgraph` and reference entities in `Subgraph<EntityRootType>` more reliably.

This would've also prevented us from forgetting to update this here.

apps/hash-frontend/src/components/hooks/use-entity-by-id.ts
```diff
@@ -35,6 +35,16 @@ export const useEntityById = ({
...
+        /**
+         * Setting `includePermissions` to `undefined` results in a "Invariant
+         * Violation: Invalid variable referenced in @include directive." runtime
+         * error in Apollo Client.
+         *
+         * @todo: consider making `includePermissions` required in the relevant queries.
+         */
+        ...(typeof includePermissions === "boolean"
+          ? { includePermissions }
+          : {}),
```
benwerner01, line 47: @CiaranMn setting `includePermissions` to `undefined` in the query variables causes a `Invariant Violation: Invalid variable referenced in @include directive.` runtime error in the apollo client.

I'm not sure why this is, but wonder whether we should make the variable non-optional by removing the default value of `false`, so that the generated TS types don't allow us to cause this error accidentally.

apps/hash-frontend/src/components/hooks/use-entity-by-id.ts
```diff
@@ -35,7 +35,16 @@ export const useEntityById = ({
...
```
benwerner01, line 47: @CiaranMn setting `includePermissions` to `undefined` in the query variables causes a `Invariant Violation: Invalid variable referenced in @include directive.` runtime error in the apollo client.

I'm not sure why this is, but wonder whether we should make the variable non-optional by removing the default value of `false`, so that the generated TS types don't allow us to cause this error accidentally.

apps/hash-frontend/src/pages/shared/block-collection/shared/mention-suggester.tsx
benwerner01: It's also possible that there may be more we haven't caught yet, so it might be worth doing a sweep of all casts either way.

nonparibus: Offline resolution: we'll do both of the things suggested in these PR comments.

apps/hash-frontend/src/components/hooks/use-entity-by-id.ts
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3462/commits/2ccdbfaba2a5a765a22b59703b1a3c67585e1d4c

apps/hash-frontend/src/pages/shared/block-collection/shared/mention-suggester.tsx
benwerner01: I've replaced the casts with assertions in https://github.com/hashintel/hash/pull/3462/commits/6c50f15184b22dcb40531bdeebe1e2cc69df536c

CiaranMn (dismissed):

CiaranMn (approved):

PR #3476, opened 2023-10-25
H-1158: Allow using path style for S3 uploads

CiaranMn:

AWS S3-compatible APIs use either 'path style' or 'virtual hosted-style' requests ([docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAPI.html#virtual-hosted-path-style-requests)).

This PR allows specifying that the path style should be used, for systems which do not / will not support the default virtual hosted-style.

I've also added environment variable documentation that was missing.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Add an `.env.local` at the root of the repo if you do not already have one
2. Set `AWS_S3_UPLOADS_FORCE_PATH_STYLE=true` in it
3. `yarn dev`
4.  Try and use a path style-dependent service (e.g. Minio)


indietyp (changes requested):
Thank you! This only works for custom ports (local minio) if one updates:

```diff
-    "@aws-sdk/client-s3": "3.38.0",
-    "@aws-sdk/client-ses": "3.25.0",
-    "@aws-sdk/credential-provider-node": "3.41.0",
-    "@aws-sdk/s3-presigned-post": "3.38.0",
-    "@aws-sdk/s3-request-presigner": "3.38.0",
+    "@aws-sdk/client-s3": "3.435.0",
+    "@aws-sdk/client-ses": "3.435.0",
+    "@aws-sdk/credential-provider-node": "3.435.0",
+    "@aws-sdk/s3-presigned-post": "3.435.0",
+    "@aws-sdk/s3-request-presigner": "3.435.0",
```

in hash-api.

TimDiekmann:
I triggered https://github.com/hashintel/hash/pull/3477

apps/hash-api/README.md
```diff
@@ -7,11 +7,18 @@ The HASH Backend API service is configured using the following environment varia
 - `NODE_ENV`: ("development" or "production") the runtime environment. Controls
   default logging levels and output formatting.
 - `PORT`: the port number the API will listen on.
-- `AWS_REGION`: the AWS region to use for the Simple Email Service (SES) provider.
-- S3 file uploads:
+- `AWS_REGION`: the AWS region to use for the Simple Email Service (SES) provider
```
nonparibus, line 10: @CiaranMn Shall we rename this `AWS_SES_REGION`? It seems a bit weird with `AWS_S3_REGION` to have this just named `AWS_REGION`.

CiaranMn:
> I triggered #3477

Doing it in this PR saves blocking the merge queue for another PR, although only for < 15m

TimDiekmann:
> Doing it in this PR saves blocking the merge queue for another PR, although only for < 15m

They can be tested in parallel 🚀

apps/hash-api/README.md
CiaranMn: It isn't only SES-related, because the S3 config will default to `AWS_REGION` if `AWS_S3_REGION` is undefined

CiaranMn: We could do with clearing up these READMEs because it seems pointless to explain all these variables in both the API readme and the higher-level one

indietyp (approved):

PR #3478, opened 2023-10-25
H-1160, H-1163: Fix org settings form submission, menu alignment

CiaranMn:

Fixes a couple of bugs around org settings:
1. H-1160: Fix left alignment of organization name where it is long
2. H-1163: Don't require touching the website field to enable form submission when creating an org
    - `react-hook-form` requires deconstructing the values of `formState` which should cause a re-render

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Create an org with a long name and check that the left alignment in the settings sidebar is correct
2. Try creating an org and check that the submit button is enabled without having to touch the website field

#### Previously
Screenshot 2023-10-25 at 20 53 09

#### Fixed
Screenshot 2023-10-25 at 20 50 38

### DIFF
--- a/apps/hash-frontend/src/pages/settings/organizations/shared/org-form.tsx
+++ b/apps/hash-frontend/src/pages/settings/organizations/shared/org-form.tsx
@@ -94,17 +94,23 @@ export const OrgForm = ({
 
   const { refetch: refetchUserAndOrgs } = useAuthInfo();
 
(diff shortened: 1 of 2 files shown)


benwerner01 (approved):
Thanks Ciaran, changes are working for me locally 👍

PR #3505, opened 2023-10-30
H-968, H-970: Add entity property validation

TimDiekmann:

When inserting entities into the store we want to validate them first against their entity types. This adds basic entity validation to check the properties of the entities.

## 🔍 What does this change?

- Adds a crate to do entity validation, property validation, and data validation
- This PR does not add the validation to the graph as this PR is big enough already

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Add the validation to the graph on creating/updating entities

## 🛡 What tests cover this?

Tests for all entities and most schemas from `graph-test-data` were


tests/hash-graph-test-data/rust/src/entity/book.json
```diff
@@ -1,5 +1,7 @@
 {
-  "https://blockprotocol.org/@alice/types/property-type/name/": "The Time Machine",
+  "https://blockprotocol.org/@alice/types/property-type/name/": [
+    "The Time Machine"
+  ],
   "https://blockprotocol.org/@alice/types/property-type/blurb/": "brulb",
```
nonparibus, line 5: Sorry, this is not important, but... `brulb`?! 😂

TimDiekmann: I thought that was on purpose and I always left it as it is just because ... why not? 😂

TimDiekmann: I think @nonparibus mainly flagged `"brulb"` instead of `"blurb"` 😄

TimDiekmann: ^ Which is actually `blurb` reversed, so I guess that _is_ on purpose.

libs/@local/hash-validation/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@local/hash-validation-rs",
+  "version": "0.0.0-private",
+  "private": true,
+  "scripts": {
+    "fix:clippy": "just clippy --fix",
+    "lint:clippy": "just clippy",
+    "test:unit": "cargo hack nextest run --feature-powerset --lib && cargo test --all-features --doc"
```
indietyp, line 8: can't wait for us to disable feature powerset again once it explodes...^^

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -0,0 +1,279 @@
+use core::fmt;
+use std::{borrow::Borrow, fmt::Formatter};
```
indietyp, line 2: You use `std` here, but then one above `core`, either `no-std` or just `std`. I think it would otherwise get confusing.

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -0,0 +1,279 @@
...
+use thiserror::Error;
+use type_system::{url::VersionedUrl, DataType, DataTypeReference};
+
+use crate::{
+    error::{Actual, Expected},
+    OntologyTypeProvider, Schema, Validate,
+};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum JsonValueType {
```
indietyp, line 15: would more likely like this to be `JsonSchemaValueType` or something similar, `Integer` is not a valid data type in JSON itself. (or remove integer)

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl<P: Sync> Schema<JsonValue, P> for DataType {
+    type Error = DataValidationError;
+
+    async fn validate_value<'a>(
+        &'a self,
+        value: &'a JsonValue,
+        _provider: &'a P,
+    ) -> Result<(), Report<DataValidationError>> {
+        match self.json_type() {
```
indietyp, line 88: I am not a particular fan of us using a string here, but I know that is more an issue with the `type-system`, than with this library.

libs/@local/hash-validation/src/property_type.rs
```diff
@@ -0,0 +1,702 @@
...
+where
+    P: OntologyTypeProvider<PropertyType> + OntologyTypeProvider<DataType> + Sync,
+{
+    type Error = PropertyValidationError;
+
+    fn validate_value<'a>(
+        &'a self,
+        value: &'a JsonValue,
+        provider: &'a P,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Report<Self::Error>>> + Send + '_>> {
```
indietyp, line 345: the beauty of recursive async functions

indietyp (changes requested):
some minor suggestions, mostly just comments 👍 great stuff.

libs/@local/hash-validation/src/property_type.rs
TimDiekmann: Haha, yes. I just spoke with @nonparibus today about recursive `async` functions in Rust and why this is so special. 😄

libs/@local/hash-validation/src/data_type.rs
TimDiekmann: I agree that this should not be a string. I [added a task](https://linear.app/hash/issue/BP-60/dont-use-string-for-the-json-type-in-datatype-in-type-system) for that.

indietyp (approved):

nonparibus (approved):

PR #3510, opened 2023-10-31
H-968, H-970: Adjust `hash-graph` to use entity validation

TimDiekmann:

#3505 added the `validation` crate. This PR is responsible to use that functionality in the Graph.

## 🚫 Blocked by

- #3505 

## 🔍 What does this change?

- Validate entities on create/update
- Drive-by: Improve errors returned from `create_entity` and `update-entity` endpoints.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/status/rust/src/lib.rs
+++ b/libs/@local/status/rust/src/lib.rs
@@ -28,7 +28,7 @@ where
 
 impl<D> Status<D>
 where
-    D: Send + Sync + Debug + Serialize + for<'de> Deserialize<'de>,
+    D: Send + Sync + Debug + Serialize,
 {
     #[must_use]
     pub fn new(code: StatusCode, message: Option<String>, contents: Vec<D>) -> Self {
--- a/apps/hash-graph/Cargo.lock
+++ b/apps/hash-graph/Cargo.lock
@@ -671,6 +671,7 @@ checksum = "27a72baa257b5e0e2de241967bc5ee8f855d6072351042688621081d66b2a76b"
 dependencies = [
  "anyhow",
  "rustc_version",
+ "serde",
  "tracing-error",
 ]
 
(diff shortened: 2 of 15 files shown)


libs/@local/status/rust/src/lib.rs
```diff
@@ -28,7 +28,7 @@ where
 
 impl<D> Status<D>
 where
-    D: Send + Sync + Debug + Serialize + for<'de> Deserialize<'de>,
```
indietyp, line 31: Why the removal?

indietyp (changes requested):
Excited to see this happen! Do we have any indicators as to the potential performance impact on insertion? (if not also no problem!)

libs/@local/status/rust/src/lib.rs
indietyp: discussed offline

indietyp (dismissed):

TimDiekmann:
> Do we have any indicators as to the potential performance impact on insertion?

Sadly I don't, sorry. it's going to slow down entity insertion (creation/updating) a bit as it has to look-up the entity type and all used property types and data types. As the queries are quite small and it does not require to cross an API boundary (e.g. Rust -> Node) it shouldn't be that bad, though. In any case, this will have impact if you regularly update entities. I have a few optimizations planned around this. I need to test out how well this works but at least three options I like to test:
- Don't lazily request types but all related types at once (could be slightly slower for big entity types, should be faster in most cases. This should have a fall-back to the lazy lookup)
- Only check changed properties. We need the update logic for entities in any case as we cannot (and want not) stick with full updates but need support for partial updates. Partial updates are required for the permission system to prevent updates on certain properties (such as the `User`'s shortname must not be edited, but the preferred name should be allowed to be changed). I will implement this first via a diff on the Graph with the updated entity and the last entity, but we want to pass a partial update to the Graph regardless.
- Common types (such as data types) can be cached so they never have to be fetched.

nonparibus (dismissed):

indietyp (approved):

PR #3515, opened 2023-11-01
H-1198: Fix entities table loading and caching issues

CiaranMn:

Fixes issues related to 'the entities table'. This is the table that appears in the following places:
- When clicking 'View all pages'
- If visiting the `/entities` page (the 'View all pages' page is just this page with a filter applied. We haven't exposed this path via the UI yet)
- When viewing a type's page, clicking the 'entities' tab

The issues fixed are as follows:
- When switching between entity types, the entities tab would continue to show the previous type's entities until the new ones had loaded in
- The way we were fetching entities of a given type was not making use of caching, meaning every visit to an instance of these tables meant waiting for the data to be fetched again, even if it had been visited already during the session. I've updated this so that we do rely on the cache. The cached data will be loaded immediately, with the loading indicator in the tab showing while a refetch is happening in the background. Any additional data will load in once this request has completed.
- While entities were loading, the entities table would show 'No entities of this type visible to you' as an empty state.
- The logic for determining which entities were 'internal' and 'external' to the user was broken (the count was correctly including entities for all webs a user belongs to, but the table rows only counted the 'active workspace' selected as internal)
    - I've fixed the same logic for the types tables 
- Visiting `/entities` and clicking 'external' would cause a crash, because we have a single entity, storing the HASH instance settings, which does not have an org associated.

See demo below.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Clicking on the HASH instance entity in the table will go to a route that will not load, because there is no org. We should figure out what to do about this.

## ❓ How to test this?

1. Recreate the switching between types / tables shown in the demo.
2. Visit `/entities` by typing it in the address bar, and confirm that clicking 'include external' doesn't crash.

### Broken

https://github.com/hashintel/hash/assets/37743469/289b5065-f572-4e08-be0c-6b0eb556e57d

### Fixed

https://github.com/hashintel/hash/assets/37743469/6a21b08f-a5c8-42ba-af12-dac338467fb7

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entities-table.tsx
+++ b/apps/hash-frontend/src/pages/shared/entities-table.tsx
@@ -15,7 +15,6 @@ import { useRouter } from "next/router";
 import {
   FunctionComponent,
   useCallback,
-  useContext,
   useEffect,
   useMemo,
   useState,
@@ -35,6 +34,7 @@ import {
   TableHeader,
   tableHeaderHeight,
 } from "../../shared/table-header";
+import { useAuthenticatedUser } from "./auth-info-context";
 import { renderChipCell } from "./chip-cell";
 import {
   createRenderTextIconCell,
@@ -45,7 +45,6 @@ import {
   useEntitiesTable,
 } from "./entities-table/use-entities-table";
 import { TOP_CONTEXT_BAR_HEIGHT } from "./top-context-bar";
-import { WorkspaceContext } from "./workspace-context";
 
 export const EntitiesTable: FunctionComponent<{
   hideEntityTypeVersionColumn?: boolean;
@@ -53,7 +52,7 @@ export const EntitiesTable: FunctionComponent<{
 }> = ({ hideEntityTypeVersionColumn, hidePropertiesColumns }) => {
   const router = useRouter();
 
-  const { activeWorkspaceOwnedById } = useContext(WorkspaceContext);
+  const { authenticatedUser } = useAuthenticatedUser();
 
   const [filterState, setFilterState] = useState<FilterState>({
     includeGlobal: false,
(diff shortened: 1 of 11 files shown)


apps/hash-frontend/src/pages/shared/entities-table.tsx
```diff
@@ -45,23 +45,37 @@ import {
...
+  } = useEntityTypeEntities();
+
+  const entities = useMemo(
+    /**
+     * If a network request is in process and there is no cached content for the request, return undefined.
+     * There may be stale data in the context related to an earlier request with different variables.
+     */
+    () => (loading && !hadCachedContent ? undefined : lastLoadedEntities),
+    [hadCachedContent, loading, lastLoadedEntities],
+  );
```
benwerner01, line 78: Is this any different than just using the current value for `entities?

```suggestion
  const entities = lastLoadedEntities;
```

I'm wondering because:
- when there is no cached content and it is loading, `lastLoadedEntities` will be `undefined`
- when there is cached content and it is loading, `lastLoadedEntities` will be defined and updated when the latest fetch has resolved
- when there it is not loading and there is cached content, `lasLoadedEntities` will be defined

benwerner01 (commented):
Thanks Ciaran, the changes are working for me but had one thought which might be worth addressing as part of this PR if I've understood correctly

apps/hash-frontend/src/pages/shared/entities-table.tsx
benwerner01: Ah apologies, I think I missed that this wouldn't account for stale data being displayed.

Would it perhaps be simpler to just move this logic into the `useEntityTypeEntitiesContextValue` hook in a way that prevents stale data and only cached data for the correct query variables being returned in the `entities` array? And otherwise to just return `undefined` whilst it's loading? This could help the the hook callers from having to make use of `hadCachedContent`

CiaranMn: There might be contexts in which it is helpful to still have the old data visible while a variable is changed... e.g. if moving to the next page of a table, or changing the sort, it might be slightly less jarring to keep the old rows in place (with some loading indicator) than to wipe the rows while the query resolved. 

This might not happen in practice but I have a slight preference for keeping it how it is and changing it later depending on how else we use it.

benwerner01: Alright, fair enough. Thanks for explaining

benwerner01 (approved):

PR #3521, opened 2023-11-02
H-1179: Allow reading/writing ontology type relationships/permissions

TimDiekmann:

To fully control ontology types this adds three endpoints for all ontology types:
- read relationships
- write relationships
- read permissions

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
+++ b/apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
@@ -4,13 +4,23 @@
 
 use std::{collections::hash_map, sync::Arc};
 
-use authorization::{backend::PermissionAssertion, AuthorizationApiPool};
+use authorization::{
+    backend::{ModifyRelationshipOperation, PermissionAssertion},
+    schema::{
+        EntityTypeGeneralViewerSubject, EntityTypeId, EntityTypeOwnerSubject, EntityTypePermission,
+        EntityTypeRelationAndSubject,
+    },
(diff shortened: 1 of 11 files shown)


apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
```diff
@@ -116,6 +152,168 @@ struct CreateEntityTypeRequest {
...
+                Consistency::FullyConsistent,
+            )
+            .await
+            .map_err(report_to_response)?,
+    ))
+}
+
+#[utoipa::path(
+    get,
+    path = "/entity-types/{entity_type_id}/permissions/{permission}",
```
CiaranMn, line 278: It feels a bit weird to have a whole URL as a path segment of another URL, because we'll end up with things like the below, but I can't think of any actual problems these will introduce
`/entity-types/https://app.hash.ai/@hash/types/entity-type/permissions/v/1/permissions/edit`

CiaranMn (approved):
Thanks again @TimDiekmann!

apps/hash-graph/lib/graph/src/api/rest/entity_type.rs
TimDiekmann: Those URLs are supposed to be encoded. Every client should do this automatically. As these are abstracted away and we don't treat the API as public anyway I don't really see an issue here.

CiaranMn (approved):

PR #3522, opened 2023-11-02
H-1174: remove `tokens` property type in favour of updated `text-content` property type

benwerner01:

This PR removes the `tokens` property type from HASH, replacing it with the updated `textual-content` property type which supports both a string and an array of objects as its values.

Previously the hook service was storing the tokens (an array of objects) in a separate `tokens` property, and attaching it to the callout/heading properties. This is problematic with the introduction of entity validation, which requires all properties on an entity to be defined on the entity's type. Making use of the updated `textual-content` property type, which is defined on the block's schemas, addresses this.

It also makes slight adjustments to the page layout:

- decreases the spacing between the blocks and page title
- aligns comments to the start of blocks, not the top of the page

## 🔗 Related links

- H-1174

## 🔍 What does this change?

- updates the `Callout`, `Paragraph` and `Heading` block to make use of the updated `textual-content` property type, and gracefully handle `string` and an array of objects as values

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None.

## ❓ How to test this?

1. Checkout the branch locally
2. Clear `hash-workspace-user-blocks-v2` in your local storage, so that the latest blocks are fetched
3. Try using the `Callout`, `Paragraph` and `Heading` blocks in a page, and ensure changes persist (indicating there are no validation errors)

## 📹 Demo

image


apps/hash-frontend/src/components/block-loader/block-loader.tsx
```diff
@@ -216,15 +211,16 @@ export const BlockLoader: FunctionComponent<BlockLoaderProps> = ({
...
     /**
-     * Our text blocks ask for a BP `textual-content` property, which is plain `Text`.
-     * Because they use the hook service, we actually generate text as text tokens and persist it under a different property.
-     * The BP spec says that blocks should receive the data they expect even if the hook service is used
-     * – this code makes sure we provide rich text as a plain string for this specific property.
+     * Our text blocks ask for a BP `textual-content` property as a plan `string`ich is plain `Text`.
+     * Because they use the hook service, we actually generate text as an array of text tokens and
+     * persist it under a different property.
+     * The BP spec says that blocks should receive the data they expect even if the hook service is used,
+     * therefore we provides the rich text as a plain string for the `textual-content` property.
```
CiaranMn, line 218:
```suggestion
     * Text fields use a `textual-content` property, with a value of either `Text` (a string) or an opaque `Object`.
     * This is so that the text can be stored as rich text tokens by the embedding application, when the hook service is used,
     * but also still receive by the block as a plain string (a predictable format), complying with the schema in both cases.
     * Here we translate our rich text tokens into a plain string for passing into the block.
```

CiaranMn (commented):
There's a comment that needs updating, otherwise looks good!

apps/hash-frontend/src/components/block-loader/block-loader.tsx
benwerner01: Thanks for spotting, have updated it in https://github.com/hashintel/hash/pull/3522/commits/fc8a7a64ddbc3351b4c04b02c5ea4b3dfb4ff13e

CiaranMn (approved):

PR #3525, opened 2023-11-03
H-61: Fetch all block data for pages in first request

CiaranMn:

For all pages showing blocks (pages, quick notes, profile bios), we were previously fetching each block's data within the block when they were rendered, meaning that loading block data was blocked on the page's data being loaded, and then meant a number of requests, causing a loading effect that was slower and more staggered across blocks than it needs to be.

This PR fixes that by requesting the contents data for all blocks on a page in the same request as the page itself, as well as a few other improvements / fixes listed below.

Note that it **removes an old integration.test.ts file** – this had been broken with all linting disabled for some time and will be simpler to rewrite as new tests when we need it. I noticed this file as it had a reference to the `getPage` query.

## 🔍 What does this change?

- Make block collection loading consistent across pages featuring blocks by:
  - removing the special-cased `getPage` query and use `getEntityQuery` everywhere
  - adding a `BlockCollectionContext` which provides the collection's subgraph and permissions, allowing `BlockLoader` to initialise block data from the collection's subgraph
    - this does mean that blocks will receive more of the graph than making a request rooted at the block. Ideally we would reconstruct the block's subgraph from the page subgraph rather than merely changing the root. This will become a requirement if we sandbox blocks, to avoid blocks receiving more of the graph than they should. H-1206
    - it also means that we are fetching more of the graph than we were previously to resolve a page, because it requires traversal depth of `4` to ensure a graph rooted at the block entity with depth of `2`. An alternative solution which avoids this would be to split the page resolution request into two: one to fetch the page and the root block entities, and another to fetch subgraphs for each block at once (rather than in each block). This would reintroduce the waterfall effect but still avoid staggered loading of blocks. Whether or not this is worth it depends on the difference in loading times between fetching more at the graph in one request, and fetching less of it in two sequential requests. We can see how the approach in this PR performs in real-world usage.
- Removes an unnecessary additional request in `notes.page.tsx`
- Remove `integration.test.ts` (this predates the Graph API)

## Known issues

Text saving is very buggy and the save process needs rewriting – I have confirmed that the following issues exist in `main` as well as this PR, although you have to watch GraphQL requests and the console to see the errors:
1. Deleting paragraphs, especially multiple at a time, can result in a 'Could not find link entity with index x' error (in response to the `updateBlockContents` GraphQL error
2. Paragraphs sometimes don't have all the text typed into them saved
3. Intermittent 'Missing draft entity' or 'Missing saved' entity issues when editing paragraphs in quick succession

I have created H-1209 to take a close look at the save process (which hasn't been touched in some time) to fix it.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None (page Playwright tests are skipped)

## ❓ How to test this?

1. Try adding pages with multiple blocks, reload them
2. Try it on the quick notes page, bio page, regular document pages

#### Old
https://github.com/hashintel/hash/assets/37743469/684a995e-87d3-476f-860e-fa042b7b2df4

#### New

https://github.com/hashintel/hash/assets/37743469/74e3f662-55fc-466d-81c6-a1954e779075

#### Old

https://github.com/hashintel/hash/assets/37743469/f064db9e-13d1-456c-8190-da7956da9054

#### New 

https://github.com/hashintel/hash/assets/37743469/a1bf8a44-9e8f-4462-9977-4e1d8435c31a


apps/hash-frontend/src/pages/notes.page.tsx
```diff
@@ -99,48 +106,6 @@ const NotesPage: NextPageWithLayout = () => {
...
-
-  const [
-    previouslyFetchedQuickNotesAllVersionsData,
-    setPreviouslyFetchedQuickNotesAllVersionsData,
-  ] = useState<StructuralQueryEntitiesQuery>();
-
-  const { data: quickNotesAllVersionsData } = useQuery<
-    StructuralQueryEntitiesQuery,
-    StructuralQueryEntitiesQueryVariables
-  >(structuralQueryEntitiesQuery, {
```
benwerner01, line 124: Apologies Ciaran, whilst testing I realised what the purpose of having this in two queries was: to first determine which quick note entities are archived, then fetch all revisions of those entities. The reason we can't combine these into a single query is because archived quick notes will always be included in the result, as they have prior revisions where archived property will be `undefined` or `false`. This results in archiving quick notes to appear broken in this PR.

We could maybe get around this by checking if the latest revision of the quick note is included in the subgraph, but this would mean we are over-fetching the number of quick notes on every page load. Unless we can somehow specify in the filter that the latest revision of the entity cannot be archived, I think it's probably best to go back to using two queries.

benwerner01 (changes requested):
Thanks Ciaran, this was a great idea and is making a big difference in loading times for me locally 👍

Unfortunately archiving quick notes is broken with the changes made in this PR, as I explain in a review comment - apologies for not having remembered this when you asked me about the purpose for using two queries.

apps/hash-frontend/src/pages/notes.page.tsx
CiaranMn: Ahh, that makes sense. I think we're still better off using one query – although we'll be fetching quick notes that are archived, I expect those will be a minority of all a user's quick notes, and it will be slower to load to fetch the non-archived quick notes and _then_ fetch all the revisions of those, rather than just fetching all the revisions of all quick notes. A quick note can only have more than one revision if it is archived, so they are all likely going to have 1 or 2 revisions to fetch. I think including archived in one request will be quicker than a waterfall of 2 requests.

I've done this in these commits https://github.com/hashintel/hash/pull/3525/files/8bb1feafcf87a69655b856d7368470499fd332e9..d16f5e70891f9203ffb51c730a73a3ab571b43c2

I also noticed a couple of bugs which exist in `main` I fixed:
- Sometimes a `href` somewhere can be an empty string, which causes `isHrefExternal` to crash (probably while something is loading, I didn't look into it, this seems fine)
- A ProseMirror error when loading the Quick Notes page sometimes. I can't recreate this in the remote deployment but I can recreate it by refreshing in `main`

https://github.com/hashintel/hash/assets/37743469/525651cc-0b7f-4cbd-8ab5-18a0a2ea0962

CiaranMn: Forgot to say that I rewrote the reduce to something I find a bit easier to read since I had to update it for an archived filter anyway – `getEntityRevision` will return the latest edition from the subgraph if no timestamp is provided, which seems easier than comparing the timestamps manually.

benwerner01 (dismissed):
Thanks @CiaranMn, the changes look good to me 👍

benwerner01 (dismissed):

benwerner01 (approved):

PR #3526, opened 2023-11-03
H-1205: Introduce relation "levels" for authorization

TimDiekmann:

The naming schema is a generic (but hardcoded) approach to encode different levels of permissions. These levels has to be modelled. By slightly changing the definitions of a `Relation` it's possible to encode the level as a number instead of a string.

## 🔍 What does this change?

- Distinguish between `Permission` and `Relation` in every location
- Slightly adjust the definitions for the `Subject` to use an object-based enum (which will be required for exposing `level` in the follow-up PR)
- Rename SpiceDB relations

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-authorization/Cargo.toml
```diff
@@ -27,4 +28,3 @@ tokio = { version = "1.33.0", default-features = false , features = ["macros", "
 
 [features]
 utoipa = ["dep:utoipa"]
-
```
indietyp, line 30: intended?

indietyp (dismissed):
I really like the RelationshipParts change!

libs/@local/hash-authorization/Cargo.toml
TimDiekmann: Yes, I found the setting to remove trailing blank linkes (but keeping one) again 😄

indietyp (approved):

PR #3527, opened 2023-11-04
H-1137: Clean up migration files

TimDiekmann:

The Postgres migration files grew a lot over time. Currently is a good opportunity to clean the migration files up and do some required renaming.

## 🔍 What does this change?

- Clean up migration files and remove unneeded code
- Remove the SQL formatter - it was not working as expected
- Rename `owners` to `webs` and `owner_id` to `web_id`
- Remove the wrong foreign key constraint from `accounts`/`account_groups` to `webs`
- rename `owned_by_id` to `web_id` in database
- Apply minimal changes to the Rust code base to take the above changes into account

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
am I correct in the assumption that this needs a complete recreation of the database?

TimDiekmann:
> am I correct in the assumption that this needs a complete recreation of the database?

Yes, that is correct. The easiest way is to create a snapshot, remove the tables from the database using a cascade deletion, run the graph migration script, and restore the snapshot again. To help with that I'm going to write a small script.

CiaranMn (approved):

PR #3534, opened 2023-11-06
H-1211: Support reading Entity and `EntityType` from validation

TimDiekmann:

To validate links it's required that entity types and entities can be seen from the validation. This adds the functionality to access these. Also, a function is added to detect if an entity type is inheriting from another type.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-validation/src/lib.rs
+++ b/libs/@local/hash-validation/src/lib.rs
@@ -112,7 +112,6 @@ mod tests {
         error::install_error_stack_hooks, property_type::PropertyValidationError,
     };
 
-    #[expect(clippy::struct_field_names)]
     struct Provider {
         entities: HashMap<EntityId, Entity>,
         entity_types: HashMap<VersionedUrl, EntityType>,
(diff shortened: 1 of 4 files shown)


libs/@local/hash-validation/src/lib.rs
```diff
@@ -130,6 +169,47 @@ mod tests {
...
+                .ok_or_else(|| Report::new(InvalidEntity { id: entity_id }))
+        }
+    }
+
+    impl EntityTypeProvider for Provider {
+        async fn is_parent_of(
+            &self,
+            child: &VersionedUrl,
+            parent: &VersionedUrl,
+        ) -> Result<bool, Report<InvalidPropertyType>> {
```
CiaranMn, line 188: same applies below
```suggestion
        ) -> Result<bool, Report<InvalidEntityType>> {
```

TimDiekmann: Thanks, I fixed this in https://github.com/hashintel/hash/pull/3534/commits/45c2d1a7ec00b3c95010886dbc597ac7d0361bcc

CiaranMn (approved):

PR #3535, opened 2023-11-06
H-1212: Allow prevention to update certain properties

TimDiekmann:

As an easy/fast solution to prevent users from updating types such as EMail and Shortname the permission to do so can globally restricted. While users may want to use these types on their own entity types, this is the quickest solution to achieve that forbids updating these properties.

This PR:

The changes in this PR:

The changes in this PR:


nonparibus:
Wait, so I wouldn't be able to update `email` on a customer record with this patch?

I'm not sure we want to block creating further schemas and dogfooding on this fact... it's a property type we'll be using.

TimDiekmann:
I agree, that these are property types we want to use elsewhere. It just provides a super simple solution which is working and preventing users from updating their mail or shortname. Cell level permissions is one of my next ups but it's not something I can quickly hack together (while this PR only took ~30min)

PR #3536, opened 2023-11-06
H-1156: Update browser extension to use new Quick Note type

benwerner01:

This PR updates the browser extension to create instances of the `Quick Note` system type in HASH.

## 🔗 Related links

- H-1156

## 🔍 What does this change?

- moves GQL queries into dedicated files
- adds codegen so that the GQL query/mutation variables and return types can be typed using TS
- adds the following local dependencies:
  - `@local/hash-subgraph`
  - `@local/hash-isomorphic-utils`
  - `@blockprotocol/type-system`
- upgrades `@graphql-codegen/cli` (added this package via `yarn add`, and thought it might be worth upgrading as we are 3 releases behind the latest one)

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Run the browser plugin locally
2. Create a quick note (you can test out multiple paragraphs)
3. Check in the HASH instance used by the browser plugin whether the quick note has been created

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/9f9d78e7-3715-4681-8ceb-b67385a2d07b


apps/plugin-browser/webpack.config.js
```diff
@@ -143,10 +143,16 @@ const options = {
...
+    fallback: {
+      "process/browser": require.resolve("process/browser"),
+    },
   },
   plugins: [
     isDevelopment && new ReactRefreshWebpackPlugin(),
     new CleanWebpackPlugin({ verbose: false }),
+    new webpack.ProvidePlugin({
+      process: "process/browser",
+    }),
```
benwerner01, line 155: These were required to prevent the imported packages from crashing which were accessing `process.env`.

apps/plugin-browser/src/pages/popup/popup-contents/action-center/quick-note-action.tsx
```diff
@@ -1,33 +1,125 @@
...
-
-const createQuickNote = (text: string) => {
-  return queryApi(createEntityQuery, {
-    entityTypeId:
-      "https://app.hash.ai/@ciaran/types/entity-type/quick-note/v/1",
-    properties: {
-      "https://blockprotocol.org/@blockprotocol/types/property-type/textual-content/":
-        text,
+const quickNoteEntityTypeId =
+  "https://hash.ai/@hash/types/entity-type/quick-note/v/1" as const;
```
benwerner01, line 20: @CiaranMn I ended up hard-coding these in the end because importing the systemTypes would require setting the relevant environment variables at process.env using webpack, which ended up being harder than expected and eventually didn't want to sink more time into.

apps/plugin-browser/src/pages/popup/popup-contents/action-center/quick-note-action.tsx
```diff
@@ -1,33 +1,125 @@
...
+    }),
+    ...paragraphs.map(async (paragraph) => {
+      const [textEntity, blockEntity] = await Promise.all([
+        createEntity({
+          entityTypeId: textEntityTypeId,
+          properties: {
+            [extractBaseUrl(textualContentPropertyTypeId)]: [
+              { tokenType: "text", text: paragraph },
+            ] satisfies TextToken[],
+          },
```
CiaranMn, line 83: Not suggesting you need to do this now, because it relies on regenerating the system types (done in #3563) but with the new fixed-URL system types we can get slightly more type safety by using a type assertion on the entity type we're aiming for, which will warn us if we use an incorrect property for the type, or if the property value is invalid for the property, rather than the current approach which only checks if the value is the value we say it is (not that it matches the property, or that the property matches the entity type).

This relies on the TS compiler detecting insufficient overlap if we use an incorrect property. Unfortunately we can't use `satisfies TextProperties` because the generated types don't have a branded `BaseUrl` as the type of their keys.

```suggestion
          properties: {
            "https://blockprotocol.org/@blockprotocol/types/property-type/textual-content/":
              [{ tokenType: "text", text: paragraph }],
          } as TextProperties,
```

apps/plugin-browser/webpack.config.js
CiaranMn: Ideally we would fix these in the packages if they are supposed to be isomorphic, because `process.env` doesn't exist in the browser. It just happens to in NextJS where we have used them so far. I think we're probably better off catching invalid process access in anything that's supposed to be browser-usable, otherwise we'll encounter this problem in any other browser app that doesn't happen to have `process.env` polyfilled or faked by something.

CiaranMn (approved):
Looks good, two non-blocking comments for future work

apps/plugin-browser/src/pages/popup/popup-contents/action-center/quick-note-action.tsx
CiaranMn: In a follow-up we can avoid hardcoding the property base urls, since we have them typed now, and can write strings for keys that are checked against the type. The entity type id doesn't exist in the generated types, although it would be helpful if it did.

nonparibus:
@benwerner01 If you don't action Ciaran's comments pre-merge/now, let's ticketize in Linear and linkback to the created tickets in the comment thread above.

apps/plugin-browser/src/pages/popup/popup-contents/action-center/quick-note-action.tsx
benwerner01: I've addressed this as part of https://github.com/hashintel/hash/pull/3536/commits/a6b58cc58b4b92e2afed128edae79921e805fc31

nonparibus (dismissed):

nonparibus (approved):

nonparibus (dismissed):

nonparibus:
@benwerner01 Now @CiaranMn is back perhaps he can re-review this (rather than me just provide rote approval).

CiaranMn (approved):

PR #3540, opened 2023-11-07
H-1213: Speed up account page list loading and page interactions. Fix icon/title updating

CiaranMn:

This PR speeds up the loading of the account page list in the sidebar by removing the custom resolver for `pages`, which had a waterfall effect of fetching first the page, then its links, then its parent (if it had one) – replacing it with a single query in the clientside.

While testing this I also noticed a couple more opportunites for improvements:

– In various places we counted refetching queries to update the UI as part of the time the user would be blocked on a loading indicator. Removing these means the user can move on quicker (e.g. to create a new comment reply, or start editing a new page), at the cost of some UI updating after the loading indicator has gone (e.g. new comment appearing, or page appearing in sidebar). I think unblocking further user action feels a bit faster, but we can see how this works in production with more data / time to travel between servers. See demo. This should make the following feel faster:
  - creating a page / subpage
  - creating a comment
  - creating a comment reply
  - editing comment text
  - resolving a comment thread
  - archiving a page

- The way we were fetching pages meant that updating an icon or a title in the page, or the icon in the sidebar, meant that it wasn't properly updated in the page/sidebar/breadcrumbs (see demo). I've fixed this.

## 🚫 Blocked by

- [ ] This will need updating if #3525 merges first, to update `getAccountPageVariables`

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. See demo for things to try. I didn't cover archiving a page in the demo but this has also had the refetch removed.

### Old

https://github.com/hashintel/hash/assets/37743469/65209147-9657-408b-8c78-bb627e72e8cd

### New

https://github.com/hashintel/hash/assets/37743469/fdb400a6-436f-41ee-bad3-2d477f9b7772


nonparibus:
@CiaranMn Solid improvements. One related suggestion/question: would it be a PITA to optimistically update the icon in all places at once? e.g. so this state (below) never appears.
Capture 2023-11-07 at 14 30 18@2x

apps/hash-frontend/src/shared/account-pages-variables.ts
```diff
@@ -0,0 +1,41 @@
...
+          equal: [{ path: ["ownedById"] }, { parameter: ownedById }],
+        },
+        ...(includeArchived ? [] : [notArchivedFilter]),
+      ],
+    },
+    graphResolveDepths: {
+      ...zeroedGraphResolveDepths,
+      isOfType: { outgoing: 1 },
+      hasLeftEntity: { incoming: 1, outgoing: 0 },
+      hasRightEntity: { incoming: 0, outgoing: 1 },
```
benwerner01, line 36: Maybe we can add a comment for what we're fetching here, I assume it's the parents of the page:
```suggestion
      // These depths are chosen to cover the following:
      //  - the page's parent page (page -> [hasLeftEntity incoming 1] parent [hasRightEntity outgoing 1] -> page)
      hasLeftEntity: { incoming: 1, outgoing: 0 },
      hasRightEntity: { incoming: 0, outgoing: 1 },
```

apps/hash-frontend/src/pages/shared/not-archived-filter.ts
```diff
@@ -0,0 +1,29 @@
...
+);
+
+export const notArchivedFilter: Filter = {
+  any: [
+    {
+      equal: [
+        {
+          path: ["properties", archivedBaseUrl],
+        },
+        // @ts-expect-error -- We need to update the type definition of `EntityStructuralQuery` to allow for this
```
benwerner01, line 16: Thanks for adding this! Maybe let's link the task we created the other day here:
```suggestion
        // @ts-expect-error -- We need to update the type definition of `EntityStructuralQuery` to allow for this
        //   @see https://linear.app/hash/issue/H-1207
```

We could also make use of this `notArchivedFilter` in additional places in the codebase, including:
- in `notifications-context.tsx`
- in `notification.tsx`
- I believe also in the BE once you merge in the latest changes from `main`, but that'd require moving this into a shared package

apps/hash-frontend/src/components/hooks/use-account-pages.ts
```diff
@@ -21,17 +42,57 @@ export const useAccountPages = (
...
+      );
+
+      const parentPage = parentLink?.rightEntity[0] ?? null;
+
+      return {
+        ...simplifyProperties(latestPage.properties),
+        metadata: latestPage.metadata,
+        parentPage: parentPage ? { metadata: parentPage.metadata } : null,
+      };
+    });
```
benwerner01, line 94: Maybe I'm missing something, but the query doesn't seem to be returning multiple revisions of pages, so wouldn't it suffice to just do this?
```suggestion
    return getRoots(typedSubgraph).map((latestPage) => {
      const pageOutgoingLinks = getOutgoingLinkAndTargetEntities(
        subgraph,
        latestPage.metadata.recordId.entityId,
      );

      const parentLink = pageOutgoingLinks.find(
        ({ linkEntity }) =>
          linkEntity[0]!.metadata.entityTypeId ===
          systemTypes.linkEntityType.parent.linkEntityTypeId,
      );

      const parentPage = parentLink?.rightEntity[0] ?? null;

      return {
        ...simplifyProperties(latestPage.properties as PageProperties),
        metadata: latestPage.metadata,
        parentPage: parentPage ? { metadata: parentPage.metadata } : null,
      };
    });
```

benwerner01 (commented):
Thanks Ciaran, this is working locally as expected. Have left some thoughts, some of which might be worth addressing as part of this PR

apps/hash-frontend/src/components/hooks/use-account-pages.ts
CiaranMn: Yes, it would – thank you!

apps/hash-frontend/src/pages/shared/not-archived-filter.ts
CiaranMn: Thank you, done in https://github.com/hashintel/hash/pull/3540/commits/84fd8f8b6271d19aad7ef99459d39e8553ac02b3

CiaranMn:
> @CiaranMn Solid improvements. One related suggestion/question: would it be a PITA to optimistically update the icon in all places at once? e.g. so this state (below) never appears. Capture 2023-11-07 at 14 30 18@2x

It's not a big PITA but it's a fair amount of code because we need to introduce a global state for pages which we don't currently have, and I'd rather do it in a separate ticket to get this in today. I've created H-1230 to do so.

benwerner01 (approved):
Thanks Ciaran, looks good to me :)

PR #3541, opened 2023-11-07
H-1126: Don't create Linear integration org if integration not enabled

CiaranMn:

We were creating the Linear system types conditional on a Linear client ID being present, but creating the Linear org unconditionally. We should only create it if the Linear integration is enabled.

A better approach would be either 
(a) to have one check for Linear being enabled, and create the types and org if so, or
(b) check which system types exist, and only create the orgs for the webs they belong to

but this will do for now – we can make it better when we introduce another integration.

## 🔗 Related links

- H-1226

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graph/util.ts
+++ b/apps/hash-api/src/graph/util.ts
@@ -27,6 +27,7 @@ import {
   PropertyTypeWithMetadata,
 } from "@local/hash-subgraph";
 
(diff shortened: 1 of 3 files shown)


TimDiekmann (approved):

benwerner01 (approved):
Thanks Ciaran, the changes look good to me 👍

PR #3542, opened 2023-11-07
H-1187, H-1190, H-1192, H-1193: Rename system types for consistency, add missing descriptions

CiaranMn:

Renames some system types and adds missing description for types.

- H-1187: Rename 'website' property type (a URL string) to 'websiteUrl', to distinguish it from a 'website' entity that might have many properties related to a website
- H-1190: Delete unused properties related to orgs
- Rename 'Org' to 'Organization'
- Add descriptions to types where they were missing

These are about making sure links are described so that they make sense as part of an 'subject predicate object' triple, where the link title is the predicate:
- H-1992: Rename 'block data' (links a block to an entity) to 'has data'
- H-1193: Rename 
  - 'avatar' to 'has avatar'
  - 'parent' to 'has parent'
  - 'author' to 'authored by'
  - 'org membership' to ~'has org membership'~ 'is member of'

This **does not**:
- Update Linear types, which will not be seeded unless the necessary Linear environment variables are set – we can rethink these separately as we need to add many more anyway
- Sort out `Page` and the `contains` link – I will do these in follow up, making two child types of `Page`, `Document` and `Canvas`, which will have new links I intend to call `hasIndexedContent` (integer index in document) and `hasPositionedContent` (x, y, translate) or something like that, but welcome other suggestions

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- H-1194: split Page into `Document` and `Canvas` with different links between them and their blocks

## ❓ How to test this?

1. Check seeding and pages, comments, profile bio still works


nonparibus (dismissed):

TimDiekmann (dismissed):

benwerner01 (approved):
Thanks @CiaranMn, the revised naming looks good to me 👍

nonparibus (approved):

PR #3549, opened 2023-11-08
H-1232: Allow forcing self-hosted behavior, don't create system orgs if self-hosted

CiaranMn:

1. Allow forcing self-hosted behavior via environment variable for local testing / development 
2. Don't create the organization entities that own system types if self-hosted (since they are loaded as external types then)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graph/util.ts
+++ b/apps/hash-api/src/graph/util.ts
@@ -46,11 +46,11 @@ import {
 import { systemAccountId } from "./system-account";
 
 // Whether this is a self-hosted instance, rather than the central HASH hosted instance
-const isSelfHostedInstance = ![
-  "http://localhost:3000",
-  "https://app.hash.ai",
-  "https://hash.ai",
-].includes(frontendUrl);
+const isSelfHostedInstance =
+  process.env.SELF_HOSTED_HASH === "true" ||
+  !["http://localhost:3000", "https://app.hash.ai", "https://hash.ai"].includes(
+    frontendUrl,
+  );
 
 const owningWebs: Record<
   SystemTypeWebShortname,
@@ -128,6 +128,10 @@ const getOrCreateOwningAccountGroupId = async (
 export const ensureAccountGroupOrgsExist = async (params: {
   context: ImpureGraphContext;
 }) => {
+  if (isSelfHostedInstance) {
+    return;
+  }
+
   const { context } = params;
 
   logger.debug("Ensuring account group organization entities exist");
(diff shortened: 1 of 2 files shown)


apps/hash-api/src/graph/util.ts
```diff
@@ -128,6 +128,10 @@ const getOrCreateOwningAccountGroupId = async (
 export const ensureAccountGroupOrgsExist = async (params: {
   context: ImpureGraphContext;
 }) => {
+  if (isSelfHostedInstance) {
+    return;
+  }
+
```
benwerner01, line 134: I think it'd be cleaner to do just not call this function, which would mean moving this check to `ensureSystemGraphIsInitialized`:
```TS
export const ensureSystemGraphIsInitialized = async (params: {
  logger: Logger;
  context: ImpureGraphContext;
}) => {
  await ensureSystemAccountExists(params);

  await ensureSystemTypesExist(params);

  if (isSelfHostedInstance) {
    await ensureAccountGroupOrgsExist(params);
  }

  await ensureSystemEntitiesExists(params);
};

```

benwerner01 (approved):
Thanks Ciaran, the changes make sense to me. Had one suggestion on an improvement, but doesn't block merging.

apps/hash-api/src/graph/util.ts
```diff
@@ -128,6 +128,10 @@ const getOrCreateOwningAccountGroupId = async (
 export const ensureAccountGroupOrgsExist = async (params: {
   context: ImpureGraphContext;
 }) => {
+  if (isSelfHostedInstance) {
+    return;
+  }
```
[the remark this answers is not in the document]
CiaranMn: The `@linear` org is to host system types, which are external if it's self-hosted.

Hosted instance:
- create `@linear` types and org if integration is enabled (saves us seeding the types now since we're going to update them)

Self-hosted instance:
- never create any orgs for system types

apps/hash-api/src/graph/util.ts
CiaranMn: My thinking behind putting it here was that this file contains the other logic for `isSelfHostedInstance` and it's therefore easier to see what is / isn't happening if the instance is self hosted. Ideally we would have a much cleaner process for creating system types which didn't require more than one check

apps/hash-api/src/graph/util.ts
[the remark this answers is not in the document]
TimDiekmann: Alright, I just thought we need the organization regardless if the integration is enabled. Even better that way! 🎉

TimDiekmann (approved):

apps/hash-api/src/graph/util.ts
nonparibus: Captured as H-1233 for follow-up triage.

benwerner01: Understood, I just think it'll be harder to figure out when viewing `ensureSystemGraphIsInitialized` to figure out which of the functions are actually executing. But I don't feel very strongly so happy to leave as is.

PR #3551, opened 2023-11-08
H-325: Improve how we define system types

benwerner01:

This PR changes how we seed the system types in the Graph API. Previously we were initalizing them in a way that required a fresh datastor to ensure all types were seeded correctly, and did not account for migrating system types in an existing datastore.

This PR replaces the system type intitializers with migration files in `apps/hash-api/src/graph/migrate-ontology-types/migrations`, which for ensure the system types used by HASH are created, and ensures the external `blockProtocol` data types have been loaded. We will want to extend these migration files once we want to apply updates to existing system types, or create new ones.

In addition, this PR removes the shared `systemTypes` and `blockProtocolTypes` objects from `libs/@local/hash-isomorphic-utils/src/ontology-types.ts`, as these assume that all system types and block protocol types have a fixed version. These have been replaced with equivalent objects in `libs/@local/hash-isomorphic-utils/src/ontology-type-ids.ts`, which is a generated file containing the latest version of the types. This file can be generated via the `generate-ontology-type-ids` script, which requires the Graph API to be running.

## 🔗 Related links

- H-325

## 🔍 What does this change?

- adds a patch to the `@blockprtocol/graph` `codegen` function to allow for rewriting type IDs so that these are fetched from `localhost:3000` even when they actual type ID starts with `https://hash.ai/...`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- we still have the linear system types defined via the initializers approach, and subsequently still have the `linearTypes` object in `libs/@local/hash-isomorphic-utils/src/ontology-types.ts`. When we redefine how these types are seed in upcoming integrations work we can remove this file and related initializer functions in `apps/hash-api/src/graph/util.ts`.

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Re-create a locally running instance of HASH with a fresh database
2. Run the backend via `yarn dev:backend`, which will run the migration scripts and create the relevant system/blockProtocol types
3. Try out the generate ontology type IDs script via `yarn generate-ontology-type-ids`
4. Run the frontend via `yarn dev:frontend` and ensure the FE works as expected


apps/hash-api/src/generate-ontology-type-ids.ts
```diff
@@ -0,0 +1,285 @@
...
+          all: [
+            {
+              equal: [{ path: ["version"] }, { parameter: "latest" }],
+            },
+            {
+              startsWith: [
+                { path: ["versionedUrl"] },
+                { parameter: "https://blockprotocol.org" },
+              ],
+            },
```
benwerner01, line 128: @CiaranMn because these are external types, my understanding is that they're not in an org which I can query the `ownedById` via. Let me know if there's a better way for obtaining these types

apps/hash-api/src/graph/migrate-ontology-types/util.ts
```diff
@@ -0,0 +1,337 @@
...
+  if (existingEntityType) {
+    return existingEntityType;
+  }
+
+  const entityTypeSchema = generateSystemEntityTypeSchema({
+    ...entityTypeDefinition,
+    entityTypeId,
+  });
+
+  // The type was missing, try and create it
```
benwerner01, line 264: @CiaranMn there's a bunch of code duplication in this file with the `src/graph/util.ts` file, which I'm hoping we can get rid of when we no longer need the initializer logic for the linear system types. If you think it's worth it I can try and share the logic between the `createSystemEntityTypeIfNotExists` and `entityTypeInitializer`, as well as the related functions.

libs/@local/hash-isomorphic-utils/src/ontology-type-ids.ts
```diff
@@ -0,0 +1,380 @@
+export const systemTypes = {
```
CiaranMn, line 1: I think we should:
1. add `baseEntityTypeUrl` to the types to stop us having to `extractBaseUrl` everywhere (probably worth doing for entity types as well as property types)
2. Split this into different `systemEntityTypes`, `systemPropertyTypes` objects so that importers are dealing with smaller objects in memory if they don't need it all, as well as helping for bundling / tree shaking for contexts where bundle size is an issue

apps/hash-api/src/graph/migrate-ontology-types/util.ts
CiaranMn: Let's leave it like this for now and remove it when we update Linear

CiaranMn (commented):
Thanks @benwerner01, big improvement and effort!

A few comments, but general approach looks good

libs/@local/hash-isomorphic-utils/src/ontology-type-ids.ts
benwerner01: I've addressed `1` in https://github.com/hashintel/hash/pull/3551/commits/24ca71b38a286f2fb347db100a443ef1783d1f33, and `2` in https://github.com/hashintel/hash/pull/3551/commits/41b2c48a8fcd14ed1c052ddccb9fefadbfdc7cad

apps/hash-api/src/generate-ontology-type-ids.ts
```diff
@@ -0,0 +1,285 @@
...
+
+import { writeFile } from "node:fs/promises";
+import * as path from "node:path";
+
+import { Logger } from "@local/hash-backend-utils/logger";
+import {
+  currentTimeInstantTemporalAxes,
+  zeroedGraphResolveDepths,
+} from "@local/hash-isomorphic-utils/graph-queries";
+import { EntityTypeWithMetadata } from "@local/hash-subgraph/.";
```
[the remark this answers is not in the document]
benwerner01: Removed as part of https://github.com/hashintel/hash/pull/3551/commits/dde85a32498715ba90cf53ef9fcaf086bc43611f

apps/hash-api/src/graph/migrate-ontology-types/types.ts
```diff
@@ -0,0 +1,16 @@
+import { BaseUrl } from "@local/hash-subgraph/.";
```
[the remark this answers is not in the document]
benwerner01: Removed as part of https://github.com/hashintel/hash/pull/3551/commits/dde85a32498715ba90cf53ef9fcaf086bc43611f

apps/hash-api/src/graph/migrate-ontology-types/util.ts
```diff
@@ -0,0 +1,337 @@
...
+import { extractVersion, VersionedUrl } from "@blockprotocol/type-system";
+import { SchemaKind } from "@local/hash-isomorphic-utils/ontology-types";
+import { slugifyTypeTitle } from "@local/hash-isomorphic-utils/slugify-type-title";
+import {
+  BaseUrl,
+  DataTypeWithMetadata,
+  EntityTypeWithMetadata,
+  OwnedById,
+  PropertyTypeWithMetadata,
+} from "@local/hash-subgraph/.";
```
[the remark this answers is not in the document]
benwerner01: Removed as part of https://github.com/hashintel/hash/pull/3551/commits/dde85a32498715ba90cf53ef9fcaf086bc43611f

apps/hash-api/src/graph/knowledge/system-types/notification.ts
```diff
@@ -63,28 +69,16 @@ export const getNotificationFromEntity: PureGraphFunction<
 /**
  * Create a system notification entity.
  *
- * @param params.title - the title of the notification
- *
  * @see {@link createEntity} for the documentation of the remaining parameters
  */
 export const createNotification: ImpureGraphFunction<
```
CiaranMn, line 74: This function isn't used, might be simpler to remove it to avoid having to think about it in future refactors.

It looks like `getNotificationById` and `getNotificationFromEntity` are also unused. `archiveNotification` is in use but could probably just be `updateEntity`. Maybe not this PR but worth doing at some point

patches/@blockprotocol+graph+0.3.1-canary-20230803163819.patch
```diff
@@ -1,3 +1,43 @@
...
+index 5967d3f..1b7f6ee 100644
+--- a/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs
++++ b/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs
+@@ -709,7 +709,12 @@ const traverseAndCollateSchemas = async (initialContext) => {
+             continue;
+         }
+         initialContext.logDebug(`Fetching ${typeId}...`);
+-        addFetchPromise(fetchTypeAsJson(typeId).then((type) => {
++
++        const rewrittenTypeId = initialContext.parameters.rewriteTypeId
```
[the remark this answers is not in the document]
benwerner01: Sure, added one in https://github.com/hashintel/hash/pull/3551/commits

apps/hash-api/src/graph/knowledge/system-types/notification.ts
benwerner01: Fair point, have removed them in https://github.com/hashintel/hash/pull/3551/commits/102fc812394ae9da4b4878348989f297bc4316dd

PR #3552, opened 2023-11-08
H-392: Use a table instead of a view for entity type inheritance

TimDiekmann:

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
+++ b/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
@@ -27,6 +27,9 @@ pub struct ReindexOperations {
     /// Reindex data types cache
     #[clap(long)]
     pub data_types: bool,
+    /// Reindex entity types cache
+    #[clap(long)]
+    pub entity_types: bool,
 }
 
 pub async fn reindex_cache(args: ReindexCacheArgs) -> Result<(), GraphError> {
@@ -51,11 +54,22 @@ pub async fn reindex_cache(args: ReindexCacheArgs) -> Result<(), GraphError> {
 
     if args.operations.data_types {
         did_something = true;
-        DataTypeStore::reindex_cache(&mut store)
+        DataTypeStore::reindex_data_type_cache(&mut store)
             .await
             .change_context(GraphError)
             .map_err(|report| {
                 tracing::error!(error = ?report, "Failed to reindex data type cache");
+                report
+            })?;
+    }
+
+    if args.operations.entity_types {
+        did_something = true;
+        DataTypeStore::reindex_entity_type_cache(&mut store)
+            .await
+            .change_context(GraphError)
+            .map_err(|report| {
+                tracing::error!(error = ?report, "Failed to reindex entity type cache");
                 report
             })?;
     }
--- a/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
+++ b/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
@@ -84,7 +84,7 @@ pub async fn reindex_cache(args: ReindexCacheArgs) -> Result<(), GraphError> {
             .await
             .change_context(GraphError)
             .map_err(|report| {
-                tracing::error!(error = ?report, "Failed to reindex entity type cache");
+                tracing::error!(error = ?report, "Failed to reindex entities cache");
                 report
             })?;
     }
--- a/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
+++ b/apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
@@ -2,8 +2,8 @@ use authorization::NoAuthorization;
 use clap::Parser;
 use error_stack::{Report, Result, ResultExt, ensure};
 use graph::store::{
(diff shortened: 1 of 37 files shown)


apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -288,14 +288,17 @@ impl<C: AsClient> PostgresStore<C> {
...
-    ) -> Result<raw::EntityType, QueryError> {
+        mut inheritance_depths: HashMap<EntityTypeId, i64>,
+    ) -> Result<(raw::EntityType, HashMap<EntityTypeId, i64>), QueryError> {
         let mut current_type = available_types
             .remove(&entity_type_id)
             .ok_or_else(|| Report::new(QueryError))
             .attach_printable("entity type not available")?;
-        let mut visited_ids = HashSet::from([entity_type_id]);
+        // let mut visited_ids = HashMap::from([(entity_type_id, 0)]);
+        let mut current_depth = 0;
```
indietyp, line 299: Even if this _should_ work, for readabilities sake I'd like to see a "real" queue instead, like:

```rust
struct Item {
	depth: u8,
	id: EntityTypeId
}

let mut queue = Vec::new();

while let Some(item) = queue.pop() {
  /* parent logic */
}
```

while the current solution (ab)uses the merging of parents, this should make it clearer as to what's happening.

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -341,34 +353,84 @@ impl<C: AsClient> PostgresStore<C> {
...
         // We need all types that the provided types inherit from so we can create the closed
         // schemas
         let parent_entity_type_ids = entity_types
             .iter()
-            .flat_map(|(_, (schema, _))| schema.inherits_from().all_of())
+            .flat_map(|(_, entity_type, _)| entity_type.inherits_from().all_of())
             .map(|reference| EntityTypeId::from_url(reference.url()).into_uuid())
             .collect::<Vec<_>>();
 
+        let parent_inheritance_depths: HashMap<EntityTypeId, Vec<(EntityTypeId, i64)>> = self
```
indietyp, line 395: why isn't this a `HashMap` instead of a `Vec`?

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
TimDiekmann: I'm not sure if this would make it necessarily easier to read. The queue would need to be populated after each iteration, so you would not avoid the outer `loop` statement. The queue would just be moved from one location to another location. Currently, it's `for parent in current_type.all_of.elements.clone()` and this would be replaced with a `while` loop.

apps/hash-graph/bins/cli/src/subcommand/reindex_cache.rs
```diff
@@ -60,6 +67,28 @@ pub async fn reindex_cache(args: ReindexCacheArgs) -> Result<(), GraphError> {
...
+            })?;
+    }
+
+    if args.operations.entities {
+        did_something = true;
+        EntityStore::reindex_entity_cache(&mut store)
+            .await
+            .change_context(GraphError)
+            .map_err(|report| {
+                tracing::error!(error = ?report, "Failed to reindex entity type cache");
```
CiaranMn, line 87:
```suggestion
                tracing::error!(error = ?report, "Failed to reindex entities cache");
```

libs/@blockprotocol/type-system/rust/src/schema/closed_resolver.rs
```diff
@@ -0,0 +1,380 @@
...
+                        current_depth,
+                    );
+                }
+            }
+
+            current_depth += 1;
+        }
+
+        if missing_schemas.is_empty() {
+            // We create the resolved metadata for the current data type and update the cache so
```
CiaranMn, line 369:
```suggestion
            // We create the resolved metadata for the current entity type and update the cache so
```

libs/@local/hash-validation/src/lib.rs
```diff
@@ -334,11 +337,48 @@ mod tests {
...
+            serde_json::from_str::<EntityType>(entity_type).expect("failed to parse entity type");
+        let entity_type_uuid = EntityTypeUuid::from_url(&entity_type.id);
+        ontology_type_resolver
+            .add_unresolved_entity_type(entity_type_uuid, Arc::new(entity_type.clone()));
+
+        let resolved_data = ontology_type_resolver
+            .resolve_entity_type_metadata(entity_type_uuid)
+            .expect("entity type not resolved");
+        let closed_entity_type = ClosedEntityType::from_resolve_data(entity_type, &resolved_data)
+            .expect("Could not close church");
```
CiaranMn, line 363: Is this supposed to say 'church'?

CiaranMn (commented):
Looking good Tim, just a few things in comments I spotted

libs/@local/hash-validation/src/lib.rs
TimDiekmann: That's a copy-paste error (we had a `church` and a `building` entity type elsewhere).
```suggestion
            .expect("Could not close entity type");
```

CiaranMn (approved):

CiaranMn (approved):

PR #3564, opened 2023-11-10
H-60, H-1262: Retain text when swapping between blocks. Fix block permissions

CiaranMn:

Fixes a couple of bugs related to blocks:
1. Restore the ability to swap between blocks that accept a text property, and retain the text (H-60)
2. Handle blocks created mid-session properly (fetch their permissions and subgraph, don't rely on it being in the page subgraph)
3. Upgrade the code block's schema to use the new text property type
4. Fix variant properties in header level 4

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

There are still several bugs to do with text editing:
1. Header levels do not set consistently (H-1276)
2. Swapping between header levels via the context menu does not work (H-1276)
3. The code block needs a blur to save, it should save periodically (H-1281)
4. Loading jank for blocks (H-1282)

## ❓ How to test this?

1. Try swapping between blocks that have a `textual-content` property – ideally we would generate this list automatically from the system types
2. Note that swapping **to** the Code block will crash until we have published the new block version, because its schema doesn't accept text tokens.

### DIFF
--- a/apps/hash-frontend/src/components/block-loader/block-loader.tsx
+++ b/apps/hash-frontend/src/components/block-loader/block-loader.tsx
@@ -100,43 +100,47 @@ export const BlockLoader: FunctionComponent<BlockLoaderProps> = ({
    * - fetching the block's subgraph manually
    */
   useEffect(() => {
-    if (blockSubgraph) {
+    if (blockSubgraph || !blockEntityId) {
       return;
     }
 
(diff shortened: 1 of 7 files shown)


benwerner01 (approved):
Thanks Ciaran, working for me locally as described by the description 👍

PR #3572, opened 2023-11-12
H-1241: Allow viewing external types in frontend

CiaranMn:

HASH instances will use entity types which come from other domains, e.g. `https://blockprotocol.org`

This PR adds a route at which these types can be viewed, entities of their type viewed, etc. It reuses the existing type editor interface at a dedicated route for external types.

It also fixes the following bugs:
- file upload via the create menu was not working, as it took you to the type's URL at https://hash.ai, rather than the correct in-instance route
- `useGetOwnerForEntity` was not waiting for users/orgs to load, which could cause a crash
- The wrong icon was being used for canvas pages in the top-right create menu
- When viewing the entity editor, the entity type card wouldn't take you to the exact version of the type the entity was assigned to
- There was a crash if a user added a pinned entity type to their profile page when they had no entiites of that type

## 🔍 What does this change?

- Mainly extracts a new `entity-type-page.tsx` file which has all the stuff from the existing entity type editor route which isn't route specific
- Adds a new `/types/external/entity-type/` route which accepts a Base64 encoded baseURL and optional `/v/n` suffix
- Rewrites external type URLs to the new internal route via a `generateLinkParameters` function

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

It is possible that we may introduce other places where we link to external types and forget to rewrite the URL via `generateLinkParameters`. This might happen if `router.push` is used, for example. It shouldn't happen when using components because the `Link` and `Button` components, which we lint for using instead of alternatives, contain the rewrite.

## ❓ How to test this?

1. Try viewing external types, which can be reached via the all types page, or by clicking through from an internal type which relies on them (e.g. as a link destination, or inheriting from them).

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/238eeb81-5bca-4ce5-9491-4cabbd7e1a73


apps/hash-frontend/src/pages/shared/entity-type-page/type-preview-slide.tsx
```diff
@@ -145,7 +146,10 @@ export const TypePreviewSlide: FunctionComponent<TypePreviewSlideProps> = ({
                     }
                     ontologyChip={
                       <Link
-                        href={remoteEntityType.schema.$id}
+                        href={
+                          generateLinkParameters(remoteEntityType.schema.$id)
+                            .href
+                        }
```
benwerner01, line 152:
```suggestion
                        href={remoteEntityType.schema.$id}
```

apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
```diff
@@ -44,7 +45,7 @@ export const EntityTypeHeader = ({
 
           {!isLatest && (
             <Link
-              href={latestVersionUrl}
+              href={generateLinkParameters(latestVersionUrl).href}
```
benwerner01, line 48:
```suggestion
              href={latestVersionUrl}
```

apps/hash-frontend/src/pages/shared/block-collection/mention-view/mention-display.tsx
```diff
@@ -298,7 +299,7 @@ export const MentionDisplay: FunctionComponent<MentionDisplayProps> = ({
           }}
         >
           <Link
-            href={entityType?.schema.$id ?? "#"}
+            href={generateLinkParameters(entityType?.schema.$id).href || "#"}
```
benwerner01, line 302: Why do we need to do this if the link component handles it for us?

```suggestion
            href={entityType?.schema.$id|| "#"}
```

apps/hash-frontend/src/pages/[shortname].page.tsx
```diff
@@ -193,13 +193,20 @@ const ProfilePage: NextPageWithLayout = () => {
...
-              : [];
+          let entityTypeAndDescendants = entityType ? [entityType] : [];
+          try {
+            entityTypeAndDescendants =
+              entityType && entitiesSubgraph
+                ? getEntityTypeAndDescendantsById(
+                    entitiesSubgraph,
+                    entityType.schema.$id,
+                  )
+                : [];
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname].page.tsx:204]: I think we still want to initialise with the available `entityType` if we can so the title is displayed

```suggestion
            entityTypeAndDescendants =
              entityType && entitiesSubgraph
                ? getEntityTypeAndDescendantsById(
                    entitiesSubgraph,
                    entityType.schema.$id,
                  )
                : entityType
                ? [entityType]
                : [];
```

benwerner01 (commented):
Thanks Ciaran, this is working for me in the deployment. Had a couple of thoughts which might be worth addressing before merging.

apps/hash-frontend/src/pages/[shortname].page.tsx
Comment (benwerner01) [reply apps/hash-frontend/src/pages/[shortname].page.tsx]: Addressed in https://github.com/hashintel/hash/pull/3572/commits/45f0e28ff4ca3c7c80ced64291bd02dbf27eea79

apps/hash-frontend/src/pages/shared/block-collection/mention-view/mention-display.tsx
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3572/commits/b31daf5c29c4daa2fcd9729c030f2852da744120

apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3572/commits/b31daf5c29c4daa2fcd9729c030f2852da744120

apps/hash-frontend/src/pages/shared/entity-type-page/type-preview-slide.tsx
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3572/commits/b31daf5c29c4daa2fcd9729c030f2852da744120

benwerner01 (approved):

PR #3603, opened 2023-11-20
H-1034: Fix keyboard shortcuts firing when they shouldn't

CiaranMn:

To handle keyboard actions involving multiple key presses we were using either `useKeys` from [rooks](https://github.com/imbhargav5/rooks) or an approach adapted from it, both of which rely on manually maintaining which keys have been held by the user. This approach is unreliable because it relies on the `keyup` handler being fired in order to wipe the key map.

A previous attempt to fix this in #2983 involving also wiping the key map on `blur` does not work for some OS-level actions, for example screen capture shortcuts on MacOS (`cmd` + `shift` + `5`).

This PR introduces an approach which relies solely on the `KeyboardEvent`, removing the need for maintaining a manual map. This does mean we cannot support some key combinations, as described in a comment in the diff, but in the absence of any particular need to do so this approach is more reliable.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Run the HASH app
2. Check that shortcuts still work as expected (E.g. cmd + K for command bar, cmd + E for new entity)
3. Press cmd + shift + 5
4. Return to HASH app, press E
5. Confirm that new entity page is not opened

### DIFF
--- a/apps/hash-frontend/src/shared/keyboard-shortcuts-context.tsx
+++ b/apps/hash-frontend/src/shared/keyboard-shortcuts-context.tsx
@@ -62,6 +62,13 @@ export const KeyboardShortcutsContextProvider: FunctionComponent<
   const setKeyboardShortcuts: SetKeyboardShortcutsFunction = useCallback(
     (shortcutsToRegister) => {
       setKeyboardShortcutsState((currentShortcuts) => {
+        /**
+         * This approach means that if Shortcut A's key combination is overridden in some context by Shortcut B,
+         * and then Shortcut B is unset, Shortcut A will no longer be in the list of shortcuts either.
+         * If we ever need to introduce shortcuts with duplicate keys we should change this, and figure out
+         * a way of setting the priority of shortcuts with duplicate keys.
+         * Similarly, {@link unsetKeyboardShortcuts} would need updating to not wipe out all shortcuts for the given keys.
+         */
         const currentShortcutsWithoutDuplicates = currentShortcuts.filter(
           (existingShortcut) =>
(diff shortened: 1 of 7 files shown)


apps/hash-frontend/src/shared/keyboard-shortcuts-context.tsx
```diff
@@ -0,0 +1,173 @@
...
+                currentShortcut.keys.length !== keysToRemove.length ||
+                !currentShortcut.keys.every((key) =>
+                  keysToRemove.includes(key),
+                ),
+            ),
+        ),
+      );
+    },
+    [],
+  );
```
benwerner01, line 98: If I'm understanding correctly, an additional limitation would be that if we have components which override global shortcuts from the command bar for example, then when the component is unmounted and `unsetKeyboardShortcuts` is called the global shortcut won't be restored. But don't think this is worth addressing until we have a use-case for this, but maybe worth keeping in mind or documenting somewhere.

benwerner01 (approved):
Thanks Ciaran, can confirm I can no longer re-produce the issue. Had one thought regarding the replacement, but shouldn't block merging :)

apps/hash-frontend/src/shared/keyboard-shortcuts-context.tsx
CiaranMn: I added a comment in https://github.com/hashintel/hash/pull/3603/commits/e3565b2319f575f2373d940d5f2fb38b6217f121

benwerner01 (approved):

PR #3606, opened 2023-11-21
H-1339: Remove delay on making new blocks editable, fix block re-ordering

CiaranMn:

Fix two bugs:
1. H-1339: A delay in making new text blocks editable was introduced by #3422, which this fixes
2. The logic for detecting if a block needs reassigning was broken for the first and last blocks in a collection

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Try adding a new paragraph in a page and check it can be edited immediately
2. Check that the calls to `updateBlockCollectionContents` when editing a page does not include any move actions, unless a block has actually been moved from its current position


benwerner01 (approved):
Thanks Ciaran, this looks good and is working for me as expected 👍

PR #3610, opened 2023-11-21
H-783: Add entity type icons

benwerner01:

This PR adds editable entity type icons, and displays them in various places in the application.

## 🔗 Related links

- H-783

## 🔍 What does this change?

- adds a `EditEmojiIconButton` component, used to set emoji icons for pages and entity types
- fixes a bug where earlier versions of archived types were being displayed in the sidebar
- fixes a bug where all versions of a type were being displayed in the edit pinned entity type modal

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- display inherited entity type icon when entity type does not have an icon set (H-1367)
- display entity type icons in the types and entities data grid tables (H-1368)

## 🛡 What tests cover this?

None.

## ❓ How to test this?

⚠️ This PR cannot be tested in the deployment, as it requires backend changes.

1. Run HASH locally
2. Create an entity type, and edit its icon in the entity type editor and publish a new version of the type
3. Ensure the type's icon is viewable throughout the application

## 📹 Demo

<details>
  <summary>Entity Type icon in the editor and sidebar</summary>
  image
</details>

<details>
  <summary>Entity Type icons in the entity type selector</summary>
image
</details>

<details>
  <summary>Entity Type icons in the pinned entity type tab on the profile page</summary>
image
</details>

<details>
  <summary>Entity Type icons in the pinned entity type selector</summary>
image
</details>

<details>
  <summary>Entity Type icons in mention suggester</summary>
image
</details>

<details>
  <summary>Entity Type icons in the an entity mention on a page</summary>
image
</details>


apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/entity-type-item.tsx
```diff
@@ -69,6 +69,11 @@ export const EntityTypeItem: FunctionComponent<EntityTypeItemProps> = ({
     router.route === "/[shortname]/types/entity-type/[entity-type-id]" &&
     urlBase === baseUrl;
 
+  const { isLink } = useIsSpecialEntityType({
+    allOf,
+    $id: entityTypeId,
+  });
```
CiaranMn, line 75: I think you can use `isSpecialEntityTypeLookup` here instead, since the entity is already in the db

CiaranMn (changes requested):
Thanks @benwerner01 – in testing this out I couldn't get the entity type icon to save. I think this is at least in part because `getSchemaFromFormData` and `getFormDataFromSchema` functions in `@hashintel/type-editor` need updating to preserve the icon in both directions.

The breadcrumbs also seem to retain the icon in state even when it's not in the API-persisted data, so probably worth a look to see if that needs updating once data is persisting (e.g. if you make some changes and then discard them, does the breadcrumb retain the discarded icon somehow)


https://github.com/hashintel/hash/assets/37743469/670a8b9f-421c-47d4-8ef4-5711c3e93bfb

CiaranMn (commented):
@benwerner01 I rebuilt the backend as you suggested which fixed the bug I reported previously – apologies for this, I forgot there was a db schema change invovled.

I think the one comment I've left previously is worth addressing unless I'm missing something.

Also, we should show the type's icon on the card which is shown when editing an entity:

Screenshot 2023-11-23 at 12 05 00

I mentioned the `getFormDataFromSchema` and `getSchemaFromFormData` functions in the `type-editor` – we shouldn't update these after all because the icon is not part of the schema, but separate metadata, so we can't have it there unless we turn them into some kind of 'getFormDataFromEntityTypeWithMetadata', which I don't suggest we worry about here.

apps/hash-frontend/src/shared/layout/layout-with-sidebar/account-entity-type-list/entity-type-item.tsx
benwerner01: Thanks for flagging, have addressed in https://github.com/hashintel/hash/pull/3610/commits/b0f13528ea01f7dda67ad029ee6eab6f11b4036b

CiaranMn (approved):

PR #3614, opened 2023-11-22
H-1276: Fix swapping between block variants, header levels

CiaranMn:

Fixes a few bugs related to setting variants of blocks. The visible effect of this is that when inserting a header of a specific level, or swapping between headers of specific levels, they will appear at the correct size (rather than defaulting to H1).

The key underlying issue is that when creating a new block with specific properties, or updating the properties on an existing entity via the context menu, what is actually happening is that the ProseMirror code is updating a local entity store which only it uses, which is periodically persisted to the API. The changes in this PR are to ensure that properties set in this way are used in the block loading code, which was previously relying on data from the API. We'd be better off having a single entity store which manages interaction with the API, rather than these piecemeal patches – see H-1351.

The PR also fixes the block swapping menu not closing when switching between the same block, which happens when switching between variants or if the clicks on the exact block that's already selected

## 🔍 What does this change?

- Fixes `block-loader` to 
  - add entities from the draft entity store into the block's subgraph where appropriate
  - accept a `fallbackBlockProperties` prop for newly created block entities, so that properties aren't empty while waiting for the entity to be saved in the API
- Explicitly close the block swapper menu when the same block is selected (it only happened to work when switching to different blocks because a new instance of the block and its context buttons is created when doing so)
- Fix some inconsistencies in the type of entities in the EntityStore vs everywhere else
- Stop non-text properties being lost when updating text in some places in the entity store code

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- We should have a single entity store instead of having to do this patch work – H-1351
- Typing e.g. `/h3` in an empty para inserts the new block below, rather than replacing the empty block. It should replace it instead – H-1266

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1. Try inserting headers at different levels, swapping between them,

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/20b255d1-918d-4d84-9a6e-87d743dbfab3

### DIFF
--- a/apps/hash-frontend/src/pages/shared/block-collection/block-context-menu/block-context-menu-item.tsx
+++ b/apps/hash-frontend/src/pages/shared/block-collection/block-context-menu/block-context-menu-item.tsx
@@ -21,6 +21,7 @@ import {
 import { MenuItem } from "../../../../shared/ui";
 
 type BlockContextMenuItemProps = {
+  closeMenu: () => void;
   itemKey: string;
   onClick?: () => void;
   icon: ReactElement;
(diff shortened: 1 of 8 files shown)


apps/hash-frontend/src/pages/shared/block-collection/block-context-menu/block-context-menu-item.tsx
```diff
@@ -32,88 +33,96 @@ type BlockContextMenuItemProps = {
 export const BlockContextMenuItem = forwardRef<
   HTMLLIElement,
   BlockContextMenuItemProps
->(({ onClick, icon, title, itemKey, subMenu, subMenuWidth }, ref) => {
```
CiaranMn, line 35: Hide whitespace, this is mostly indention changes

apps/hash-frontend/src/pages/shared/block-collection/component-view.tsx
```diff
@@ -199,7 +199,6 @@ export class ComponentView implements NodeView {
               <BlockCollectionContext.Consumer>
                 {(collectionContext) => (
                   <BlockLoader
-                    key={entityId} // reset the component state when the entity changes
```
CiaranMn, line 202: I don't think this is necessary any more and I believe it was contributing to flashes of newly created paragraphs when the `entityId` was received from the API.

libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
```diff
@@ -375,10 +372,34 @@ const entityStoreReducer = (
...
         }
 
+        const now = new Date().toISOString();
+
         draftState.store.draft[action.payload.draftId] = {
           metadata: {
+            archived: false,
             recordId: {
               entityId: action.payload.entityId,
+              editionId: now,
```
benwerner01, line 382: Just for my understanding, is the `editionId` always a timestamp? Or is it just a timestamp in the context of the entity store, and a UUID otherwise?

benwerner01 (approved):
Thanks Ciaran, this is working for me in the deployment and looks good to me 👍

libs/@local/hash-isomorphic-utils/src/entity-store-plugin.ts
CiaranMn: In subgraphs it is always a timestamp, corresponding to the start of the decision time range the edition applies over

PR #3619, opened 2023-11-23
H-1408: Add an `entityCreator` relationship for webs

TimDiekmann:

To allow the system user to create linear entities in a web, a relationship has to be created. As the system user should not be the owner of the web, a new relationship type is added.

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann:
Currently, it's not needed. Will close it as we may come up with another approach. Will re-open if we need this anyway.

PR #3620, opened 2023-11-23
H-1384: Stop full-page refresh after submitting the entity type form

CiaranMn:

In the remote deployment, types are given a `https//hash.ai` origin for their ids, but are hosted on `https://app.hash.ai`, in anticipation of the frontend being moved to `https://hash.ai`. This involves a number of temporary workarounds when users click on links with `https://hash.ai` URLs to actually treat them as internal – introduced in #3485.

One place where a workaround is required was missed – where users are sent to after submitting the type form. This means after submission users were sent to `https://hash.ai`, and redirected back to the same page, resulting in a full refresh (and the time taken to process the redirect). This PR fixes that.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Only possible in production, since the workaround depends on the frontend running on `https://app.hash.ai` – but the logic alone should make sense.


benwerner01 (approved):
Thanks Ciaran

PR #3628, opened 2023-11-25
BP-70: Directly use `(de)serialize` on ontology types

TimDiekmann:

As a quick solution to distinguish between format validation and content validation, the raw module should be made private and the serde implementation should use that internally. This greatly improves the DX around the types.
The amount of code which can be removed because of this speaks for itself.

## 🔍 What does this change?

- Avoid the steady back and forth casting between `raw::` and normal representation

## 🚫 Blocked by

- https://github.com/blockprotocol/blockprotocol/pull/1350

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/type-fetcher/src/fetcher.rs
+++ b/apps/hash-graph/lib/type-fetcher/src/fetcher.rs
@@ -28,7 +28,7 @@ impl fmt::Display for FetcherError {
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum OntologyTypeRepr {
+pub enum FetchedOntologyType {
     DataType(DataType),
     PropertyType(PropertyType),
     EntityType(EntityType),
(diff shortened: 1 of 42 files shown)


apps/hash-graph/lib/type-fetcher/src/fetcher.rs
```diff
@@ -29,9 +29,9 @@ impl fmt::Display for FetcherError {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum OntologyTypeRepr {
```
indietyp, line 31: can probably be renamed, as it is no longer a `repr`, but just a plain `OntologyType`

indietyp (changes requested):
just a minor comment

apps/hash-graph/lib/type-fetcher/src/fetcher.rs
TimDiekmann: `OntologyType` is already the name of a trait so I left this for now.

TimDiekmann: I renamed it to `FetchedOntologyType`. It's only used in the type fetcher and it's better than having `Rear` as part of the name. We can always rename it or refactor it if we come up with another idea 🙂

indietyp (approved):

PR #3632, opened 2023-11-27
H-255: Implement root pagination for structural queries

TimDiekmann:

To query in a more performant way, this adds pagination to the subgraph roots.

## 🔍 What does this change?

Allows two query parameters:
- `limit`: The number of roots to be returned from the graph
- `after`: Omit every entry returned before this cursor
Returns a new header:
- `Link` contains a link to the next page. Will only be returned if `limit` was passed. Please note [the format of the `Link` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link).

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Due to the way how the `"latest"` filter for ontology type is implemented, pagination is not supported if ontology types are filtered by their latest version. See H-1442.

## ❓ How to test this?

- Query a few types or entities with `?limit=X`
- Make sure the `Link` header is returned and at most `X` items
- Use the provided link and query the next page

### DIFF
--- a/apps/hash-graph/Cargo.lock
+++ b/apps/hash-graph/Cargo.lock
@@ -200,9 +200,9 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.4"
+version = "0.21.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9ba43ea6f343b788c8764558649e08df62f86c6ef251fdaeb1ffd010a9ae50a2"
+checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
 
 [[package]]
 name = "basic-toml"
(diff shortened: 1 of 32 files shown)


indietyp:
Have you thought about supporting [`Range`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) or [`Content-Range`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range).

TimDiekmann:
Currently, I only implemented forward cursors. The idea is to implement sorting very soon (probably next week) and with that the implementation for cursors would need adjustments so I only implemented the MVP for pagination for now. After sorting has landed I like to improve the pagination to allow forward and backward cursors. A `Range` header would be straight-forward to add then.

CiaranMn (dismissed):
Great, thanks @TimDiekmann!

PR #3633, opened 2023-11-27
H-1443: Implement caching when validating entities in snapshots

indietyp:

While trying to insert a snapshot from <insert place here>, I found that the restore times were not tolerable. One of the problems encountered seems to be the repeated fetching of the same types thousands of times, leading to massive slowdowns (we're talking about down to multiple hours).

This tries to remedy it by only querying SpiceDB and Postgres once and saving the result in an intermediate cache.

@TimDiekmann, any guidance on how to best write a test case for this?

## 🔍 What does this change?

- Naive caching strategy for types (not entities).

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/validation.rs
+++ b/apps/hash-graph/lib/graph/src/store/validation.rs
@@ -1,4 +1,7 @@
+use std::{collections::HashMap, fmt::Debug, hash::Hash, sync::Arc};
+
 use authorization::{
+    backend::PermissionAssertion,
     schema::{
         DataTypeId, DataTypePermission, EntityPermission, EntityTypeId, EntityTypePermission,
         PropertyTypeId, PropertyTypePermission,
@@ -13,6 +16,7 @@ use graph_types::{
     knowledge::entity::{Entity, EntityId},
     ontology::{DataTypeWithMetadata, EntityTypeWithMetadata, PropertyTypeWithMetadata},
 };
+use tokio::sync::RwLock;
 use tokio_postgres::GenericClient;
 use type_system::{url::VersionedUrl, DataType, EntityType, PropertyType};
 use validation::{EntityProvider, EntityTypeProvider, OntologyTypeProvider};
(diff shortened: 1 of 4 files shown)


apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -52,33 +137,61 @@ where
...
+        if let Err(error) = self.authorize_data_type(data_type_id).await {
+            self.cache.data_types.deny(data_type_id).await;
+            return Err(error);
+        }
+
+        let type_ = self
+            .store
             .read_one(
                 &Filter::<S::Record>::for_versioned_url(type_id),
                 Some(&QueryTemporalAxesUnresolved::default().resolve()),
```
TimDiekmann, line 177: Noticed an error here: this must not be `default`. Could you drive-by this to the following snippet:
```rust
QueryTemporalAxesUnresolved::DecisionTime {
    pinned: PinnedTemporalAxisUnresolved::new(None),
    variable: VariableTemporalAxisUnresolved::new(None, None),
}
```

Also applies to the other two implementations

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -52,33 +137,61 @@ where
...
+        if let Some(cached) = self.cache.data_types.get(&data_type_id).await {
+            return cached;
+        }
+
+        if let Err(error) = self.authorize_data_type(data_type_id).await {
+            self.cache.data_types.deny(data_type_id).await;
+            return Err(error);
+        }
+
+        let type_ = self
```
TimDiekmann, line 173: Let's avoid the weird underscore
```suggestion
        let schema = self
```

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -24,26 +28,107 @@ use crate::{
...
+
+#[derive(Debug)]
 pub struct StoreProvider<'a, S, A> {
     pub store: &'a S,
+    pub cache: StoreCache,
     pub authorization: Option<(&'a A, AccountId, Consistency<'static>)>,
 }
 
-impl<S, A> OntologyTypeProvider<DataType> for StoreProvider<'_, S, A>
+impl<'a, S, A> StoreProvider<'a, S, A>
```
TimDiekmann, line 128:
```suggestion
impl<S, A> StoreProvider<'_, S, A>
```

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -52,33 +137,61 @@ where
...
+            .map(|data_type| data_type.schema)?;
+
+        let type_ = self.cache.data_types.grant(data_type_id, type_).await;
+
+        Ok(type_)
     }
 }
 
-impl<S, A> OntologyTypeProvider<PropertyType> for StoreProvider<'_, S, A>
+impl<'a, S, A> StoreProvider<'a, S, A>
```
TimDiekmann, line 188:
```suggestion
impl<S, A> StoreProvider<'_, S, A>
```

TimDiekmann (changes requested):
Looks good, thank you!

 I have a few minor suggestions.

> any guidance on how to best write a test case for this?

The backend integration tests will pick up this code. That should be enough for now.

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -24,26 +28,115 @@ use crate::{
...
+}
+
+impl<K, V> CacheHashMap<K, V>
+where
+    K: Debug + Eq + Hash + Send + Sync,
+    V: Send + Sync,
+{
+    async fn get(&self, key: &K) -> Option<Result<Arc<V>, Report<QueryError>>> {
+        let guard = self.inner.read().await;
+        let access = guard.get(key)?.clone();
```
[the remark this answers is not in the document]
TimDiekmann: Also, could you please prefer `Arc::clone` instead of `arc.clone()`?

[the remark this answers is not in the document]
indietyp: I had that before, but clippy complained that I held the read guard longer than necessary. (as one can see, I usually very much prefer `Arc::clone(&t)` over `.clone()`

apps/hash-graph/lib/graph/src/store/validation.rs
indietyp: fixed in [204910b](https://github.com/hashintel/hash/pull/3633/commits/204910b84646c5601c7c81c274faba86df980b30)

apps/hash-graph/lib/graph/src/store/validation.rs
indietyp: [204910b](https://github.com/hashintel/hash/pull/3633/commits/204910b84646c5601c7c81c274faba86df980b30)

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -24,26 +28,107 @@ use crate::{
...
+
+impl StoreCache {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            data_types: CacheHashMap::new(),
+            property_types: CacheHashMap::new(),
+            entity_types: CacheHashMap::new(),
+        }
+    }
```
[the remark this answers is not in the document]
indietyp: [204910b](https://github.com/hashintel/hash/pull/3633/commits/204910b84646c5601c7c81c274faba86df980b30)

apps/hash-graph/lib/graph/src/store/validation.rs
indietyp: [204910b](https://github.com/hashintel/hash/pull/3633/commits/204910b84646c5601c7c81c274faba86df980b30)

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -87,30 +200,62 @@ where
...
+
+        let type_ = self
+            .store
             .read_one(
                 &Filter::<S::Record>::for_versioned_url(type_id),
                 Some(&QueryTemporalAxesUnresolved::default().resolve()),
             )
             .await
-            .map(|data_type| data_type.schema)
+            .map(|data_type| data_type.schema)?;
```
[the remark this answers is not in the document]
indietyp: fixed in [204910b](https://github.com/hashintel/hash/pull/3633/commits/204910b84646c5601c7c81c274faba86df980b30)

apps/hash-graph/lib/graph/src/store/validation.rs
TimDiekmann: Great, thank you!

apps/hash-graph/lib/graph/src/store/validation.rs
```diff
@@ -24,26 +28,97 @@ use crate::{
     },
 };
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Clone)]
+enum Access<T> {
+    Granted(T),
+    Denied,
+    Malformed,
```
TimDiekmann, line 35: For what do we need these variants? Just as a small optimization in a multithreading context?

indietyp:
Some stats for those interested, this brought down the restore in [name here] from 2h+ (after that, we stopped due to fear of some infinite loop) to 10 minutes, of course not as good as previously before entity validation, but already a big improvement.

apps/hash-graph/lib/graph/src/store/validation.rs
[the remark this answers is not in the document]
indietyp: there is: https://rust-lang.github.io/rust-clippy/master/index.html#/clone_on_ref_ptr

apps/hash-graph/lib/graph/src/store/validation.rs
indietyp: > I mean if you have a `Denied` or `Malformed` you'll fail anyway, right?

Exactly, which is why we just "replay" that failure, should it happen.

apps/hash-graph/lib/graph/src/store/validation.rs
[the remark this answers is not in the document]
indietyp: How about the change in [4c5a32e](https://github.com/hashintel/hash/pull/3633/commits/4c5a32eccc782653f3ab213297b21bea89b6955b)?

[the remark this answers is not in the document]
TimDiekmann: Whatever uses `Arc::clone` and satisfies slippy 😅

TimDiekmann (approved):

apps/hash-graph/lib/graph/src/store/validation.rs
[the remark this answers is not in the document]
TimDiekmann: Well, apparently clippy knows better 😂

TimDiekmann (approved):

PR #3642, opened 2023-11-28
H-1457: Fix SpiceDB failing when inserting more than 1000 relationships

TimDiekmann:

SpiceDB is configured to fail if more than 1000 relationships are being updated. This is a sensible default and we should chunk the requests instead.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
+++ b/apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
@@ -455,6 +455,14 @@ impl<C: AsClient> EntityTypeStore for PostgresStore<C> {
                 closed_schema,
             } = insertion;
 
+            relationships.push((
+                EntityTypeId::from_url(schema.id()),
+                EntityTypeRelationAndSubject::Viewer {
+                    subject: EntityTypeViewerSubject::Public,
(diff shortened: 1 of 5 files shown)


indietyp (approved):

apps/hash-graph/lib/graph/src/store/postgres/ontology/entity_type.rs
```diff
@@ -455,6 +455,14 @@ impl<C: AsClient> EntityTypeStore for PostgresStore<C> {
                 closed_schema,
             } = insertion;
 
+            relationships.push((
```
indietyp, line 458: Any reason why you moved this around?

apps/hash-graph/lib/graph/src/store/postgres/ontology/property_type.rs
```diff
@@ -290,6 +290,14 @@ impl<C: AsClient> PropertyTypeStore for PostgresStore<C> {
                     .change_context(InsertionError)?;
             }
 
+            relationships.push((
```
indietyp, line 293: same here

indietyp (changes requested):

indietyp (approved):

PR #3650, opened 2023-11-29
H-1450: Restrict non-draft links to non-draft entities

TimDiekmann:

If a link is not a draft it must not link to any draft entity.

## 🚫 Blocked by

- #3640 
- #3646 
- #3649

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-validation/src/entity_type.rs
+++ b/libs/@local/hash-validation/src/entity_type.rs
@@ -43,6 +43,8 @@ pub enum EntityValidationError {
     InvalidLinkTypeId { link_type: VersionedUrl },
     #[error("The link target `{target_type}` is not allowed")]
     InvalidLinkTargetId { target_type: VersionedUrl },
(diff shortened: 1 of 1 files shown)


CiaranMn (approved):
We spoke about what happens when you have required links (`minItems > 0`), e.g. Entity A must have a link to Entity B, summary follows.

If you have draft Entity A with a draft link to Entity B, you need to be able to un-draft both Entity A and the link at the same time to make them both valid (un-draft Entity A would not be valid without an un-draft link to Entity B, and un-draft link would not be valid with draft Entity on its left).

We can merge this but before we add validation on min links we need the ability to un-draft multiple entities/links at once – as well as in other situations, such as the ability to 'replace' a link on an entity with min: 1/max: 1 in a single transaction which archives the existing link and creates the new one.

TimDiekmann:
Thank you for writing down our conclusion!

PR #3652, opened 2023-11-29
H-1465: Add graph visualization to `/entities` page

benwerner01:

This PR adds a graph visualization to the `/entities` page.

## 🔗 Related links

- H-1465

## 🔍 What does this change?

- Moves the `EChart` component out of the `Chart` block and into the shared design-system package. Technically we should therefore republish the block, but as in follow up we are adding the `Graph` chart variant we can republish the block then.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Sometimes when switching between filters the graph hard resets. I'm not sure which of the properties is causing this, and why it only happens sometimes.

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Go to the `/entities` page to see the graph with all entities
2. Go to the `/entities?entityTypeIdOrBaseUrl=https://hash.ai/@hash/types/entity-type/user/` to view all `User` entities and their linked entities, or try out a different type

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/d2f2795b-04d9-4769-92ad-44030c1b04b6

### DIFF
--- a/apps/hash-frontend/src/shared/entity-type-entities-context/use-entity-type-entities-context-value.tsx
+++ b/apps/hash-frontend/src/shared/entity-type-entities-context/use-entity-type-entities-context-value.tsx
@@ -1,6 +1,5 @@
 import { useQuery } from "@apollo/client";
 import {
-  BaseUrl,
   EntityType,
   PropertyType,
   VersionedUrl,
@@ -9,7 +8,7 @@ import {
   mapGqlSubgraphFieldsFragmentToSubgraph,
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
-import { EntityRootType } from "@local/hash-subgraph";
+import { BaseUrl, EntityRootType } from "@local/hash-subgraph";
 import {
   getEntityTypeAndParentsById,
   getPropertyTypeById,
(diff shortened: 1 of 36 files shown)


apps/hash-frontend/src/shared/entity-type-entities-context/use-entity-type-entities-context-value.tsx
```diff
@@ -64,6 +63,8 @@ export const useEntityTypeEntitiesContextValue = (params: {
       constrainsValuesOn: { outgoing: 255 },
       inheritsFrom: { outgoing: 255 },
       isOfType: { outgoing: 1 },
+      hasLeftEntity: { outgoing: 1, incoming: 1 },
+      hasRightEntity: { outgoing: 1, incoming: 1 },
```
CiaranMn, line 67: This is going to significantly slow loading in places where we fetch a lot of entities and don't currently care about linked entities (e.g. entities tab on the Entity Type page, the table on the `/entities` route) – I think we should make this an argument can be passed as needed.

libs/@hashintel/design-system/src/entities-graph-chart.tsx
```diff
@@ -0,0 +1,220 @@
...
+  Entity,
+  EntityId,
+  EntityRootType,
+  Subgraph,
+} from "@local/hash-subgraph";
+import { getEntities, getEntityTypeById } from "@local/hash-subgraph/stdlib";
+import {
+  extractBaseUrl,
+  LinkEntity,
+} from "@local/hash-subgraph/type-system-patch";
```
CiaranMn, line 15: You should be able to use equivalents in `@blockprotocol/graph` (to replace `@local/hash-subgraph`) and `@blockprotocol/type-system/slim` (to replace `@local/hash-subgraph/type-system-patch`) to avoid importing from local packages in a published one

benwerner01: I've moved the component into the `@hashintel/blocks-design-system` package, and removed the local dependencies which I could. Some remain in the `generateEntityLabel` implementation because of the `simplifyProperties` usage.

CiaranMn (approved):

PR #3657, opened 2023-11-30
H-1468: Adjust structural queries to allow excluding drafts

TimDiekmann:

In most cases, draft entities should not be included in the result set. For that it's not enough to filter them out in the roots, but also entirely in the returned subgraph.

## 🚫 Blocked by

- #3656 

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-ts/src/activities/infer-entities.ts
```diff
@@ -89,6 +89,9 @@ export const inferEntities = async ({
           inheritsFrom: { outgoing: 255 },
         },
         temporalAxes: currentTimeInstantTemporalAxes,
+        // @todo Create new inferred entities as drafts instead of actual entities
+        //   see https://linear.app/hash/issue/H-1374
+        includeDrafts: false,
```
CiaranMn, line 94: This query doesn't request any entities (it's for types) so I don't think we need this comment
```suggestion
        includeDrafts: false,
```

apps/hash-ai-worker-ts/src/activities/infer-entities/create-entities.ts
```diff
@@ -202,6 +204,8 @@ export const createEntities = async ({
                 ownedById,
                 owner: ownedById,
                 properties,
+                // @todo Create new inferred entities as drafts instead of actual entities
+                //   see https://linear.app/hash/issue/H-1374
```
CiaranMn, line 208: I am dealing with this in my current branch, we can leave this out
```suggestion
```

CiaranMn (approved):

CiaranMn (approved):

PR #3658, opened 2023-11-30
H-1271: Create "Graph Chart" variant of the chart block

benwerner01:

This PR adds a "Graph Chart" variant to the chart block, re-using the existing implementation of the `EntitiesGraphChart` component.

## 🔗 Related links

- H-1271

## 🔍 What does this change?

- makes the nodes in the graph visualization `blue[70]`, and the label text border `blue[90]`
- make the config menu collapsible (H-1407)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The data that can be loaded into the Graph Chart block is limited to what can be expressed in a single query to the BP query language (e.g. it won't be possible to query entities with different types when constraining a property value as there are no nested conditionals). To address this, we plan on updating the query editor to construct structural queries instead of BP queries, and replace the BP query language with structural query language in the blocks loaded into HASH. (H-1492)
- Clicking the nodes in the graph doesn't do anything, as blocks cannot currently inform the embedding application that an entity has been "clicked" (H-1493)

## 🛡 What tests cover this?

None

## ❓ How to test this?

1. Checkout the branch locally, run the HASH app, and run the chart block in dev mode
2. Load the graph chart block in a page, and add a structural query via the data selector modal
3. View the visualisation in the graph block, and try to set the incoming/outgoing link depths in the configuration

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/58032970-c984-41ba-bcef-36a9cc5c5ded


benwerner01:
> I don't think Graph Chart makes sense, even if this is a variant of the Chart block created using ECharts. I would probably call it Network Diagram.

@nonparibus sure, I've renamed it to `Network Diagram` in https://github.com/hashintel/hash/pull/3658/commits/fcc6ec8acef11c448550e407a57edb7c4c0099e5.

> We'd want this to be discoverable when a user types /graph as well. Do we have any support for specifying aliases/alternative names? Graph Network is the other search term I feel this should show up for (as equivalent to a direct match).

I'm not aware of a way of doing this currently, I'd guess we would want to specify this in the `variants.json` file in a `keywords` array or something like that to assist filtering in the inline block suggester?

CiaranMn (approved):

CiaranMn (approved):

PR #3663, opened 2023-12-01
H-1488: Implement bulk-checking of permissions

TimDiekmann:

When reading `N` entities at once we currently do `N` SpiceDB roundtrips. This should be reduced to ideally a single one.

## 🔍 What does this change?

- Bump SpiceDB to `1.28.0`
- Use the experimental bulk permission check endpoint

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-authorization/src/api.rs
```diff
@@ -125,26 +125,7 @@ pub trait AuthorizationApi {
...
-                } = self
-                    .check_entity_permission(actor, permission, entity, consistency)
-                    .await?;
-                result.insert(entity, has_permission);
-                zookie = checked_at;
-            }
-            Ok((result, zookie))
-        }
-    }
+    ) -> impl Future<Output = Result<(HashMap<EntityUuid, bool>, Zookie<'static>), CheckError>> + Send;
```
indietyp, line 128: The API only mentions `EntityId` instead `EntityUuid`, except for this change. Why?

libs/@local/hash-authorization/src/backend/mod.rs
```diff
@@ -382,7 +437,29 @@ impl CheckResponse {
     }
 }
 
-/// Error returned from [`ZanzibarBackend::check`].
+/// Return value for [`ZanzibarBackend::check_permissions`].
+#[derive(Debug)]
+#[must_use]
+pub struct BulkCheckResponse<I> {
+    /// If the subject has the specified permission or relation to an [`Resource`].
+    pub permission_iterator: I,
```
indietyp, line 445: nit: I dislike how this is suffixed with `_iterator`, in that case, it should just `I` should be constrained, that would maybe be a good idea anyway, considering `BulkCheckItem`. (or call it `permissions`)

non-blocking.

libs/@local/hash-authorization/src/zanzibar/api.rs
```diff
@@ -152,10 +154,51 @@ where
...
+            .backend
+            .check_permissions(
+                entities
+                    .into_iter()
+                    .map(move |entity| (entity.entity_uuid, permission, actor)),
+                consistency,
+            )
+            .await?;
+        let mut status = Ok::<(), Report<RpcError>>(());
+        let permission_map = response
```
indietyp, line 178: nit: dislike the use of `_map` to indicate that it is indeed a map.

indietyp (changes requested):
1 clarification
2 nits (non-blocking)

libs/@local/hash-authorization/src/api.rs
TimDiekmann: SpiceDB only stores the UUIDs, so there is no way to get the owned-by-id back.

PR #3665, opened 2023-12-02
H-1499: Create cargo workspace for HASH

TimDiekmann:

To be able to properly use graph libraries outside of the HASH repository the `hash-graph` crates should be split into smaller, reusable parts (as we already have for e.g. validation, types, or authorization). By just moving them to `libs/@local`, the DX gets bad due to missing commands to run e.g. `cargo clippy` on all packages at once. 
@indietyp  and I agreed that we want a workspace definition.It's possible to add a workspace to `libs/@local` only:
* libraries such as `error-stack` should not be part of that workspace (this could be avoided by not adding it to members)
* lints may be different between crates (this could be avoided by fine-tuning lints via `RUSTFLAGS`, but the general idea is to use the `[lint]` section in `Cargo.toml`. It might be possible to override lints on a per-crate basis but this has to be discovered).
* Different crates may have different dependencies (they could be different, though, they just cannot be specified in the workspace)
* We can always change this later
 
There are a few issues by putting it into `libs/@local` instead of the root:
* If it’s possible to override lints it would be preferable to remove the script-based .config approach as setting RUSTFLAGS has weird side effects (such as not being picked up if the environment variable RUSTFLAGS is set)
* We still have four different HASH-related crates/workspaces so commands still has to be executed multiple times. Also locally there will be four target folders:
    * `apps/hash-graph` (virtual workspace with several packages)
    * `libs/@local` (virtual workspace)
    * `tests/hash-graph-test-data` (crate)
    * `tests/hash-graph-integration` (crate)
* Not every workspace may share the same dependencies, but that’s also not required.

Putting it into the roots seems to be the better option so we're starting off with that.

This PR:

The changes in this PR:

The changes in this PR:


.github/actions/prune-repository/action.yml
```diff
@@ -48,6 +57,7 @@ runs:
     - name: Restore `out` directory
       shell: bash
       run: |
+        ls -lah out/
```
indietyp, line 60: is this still relevant?

.github/actions/prune-repository/action.yml
```diff
@@ -21,16 +21,25 @@ runs:
...
+            echo -e "[package]\nname = \"$(yq '.package.name' -p toml -oy $line)\"" > "out/$DIRECTORY/Cargo.toml"
+          fi
+        done < <(find $(yq '.workspace.members' -p toml -o tsv Cargo.toml | tr '*' ' ') -maxdepth 2 -name Cargo.toml)
+
     - name: Copy required files
       shell: bash
       run: |
-        cp -R yarn.lock patches out/
+        cp -R yarn.lock patches Cargo.toml Cargo.lock rust-toolchain.toml out/
+        cat "out/Cargo.toml"
```
indietyp, line 38: is this still relevant?

.github/workflows/lint.yml
```diff
@@ -75,12 +75,17 @@ jobs:
             | jq '[.tasks[] | select(.task == "codegen" and .command != "<NONEXISTENT>")] != []' || echo 'false')
           echo "codegen=$CODEGEN" >> $GITHUB_OUTPUT
 
-          HAS_RUST=$(test -f "${{ matrix.directory }}/rust-toolchain.toml" && echo 'true' || echo 'false') 
+          HAS_RUST=$([[ -f "${{ matrix.directory }}/Cargo.toml" || ${{ matrix.directory }} = "apps/hash-graph" ]] && echo 'true' || echo 'false')
```
indietyp, line 78: nothing for right now as it is definitely overengineering (although quite elegant IMO) but in the future do we maybe want to introduce a `.hash` toml file for these kind of exceptions?

.github/workflows/test.yml
```diff
@@ -252,17 +257,21 @@ jobs:
...
         id: tests
         run: |
-          HAS_RUST=$(test -f "${{ matrix.directory }}/rust-toolchain.toml" && echo 'true' || echo 'false')
-
           TEST_TASKS=$(turbo run test:integration --dry-run=json --filter "${{ matrix.package }}" | jq -c '.tasks[]')
           REQUIRES_GRAPH=$(echo "$TEST_TASKS" | jq -s '[.[] | select(.package == "@apps/hash-graph" and .task == "build")] != []')
           REQUIRES_AI_WORKER_TS=$(echo "$TEST_TASKS" | jq -s '[.[] | select(.package == "@apps/hash-ai-worker-ts" and .task == "build")] != []')
           REQUIRES_INTEGRATION_WORKER=$(echo "$TEST_TASKS" | jq -s '[.[] | select(.package == "@apps/hash-worker-integration" and .task == "build")] != []')
 
+          HAS_RUST=$([[ -f "${{ matrix.directory }}/Cargo.toml" || ${{ matrix.directory }} = "apps/hash-graph" ]] && echo 'true' || echo 'false')
```
indietyp, line 265: I still think that maybe in the future we should move these into dedicated bash scripts, that way we don't need to duplicate so much work.

indietyp (changes requested):

.github/workflows/lint.yml
TimDiekmann: We don't want to keep that exception. It's because it was a workspace previously and all commands using this assumption, still.

.github/workflows/test.yml
TimDiekmann: Yes, absolutely, I just did not get to it, yet 😉

.github/actions/prune-repository/action.yml
TimDiekmann:
```suggestion
```

indietyp (approved):
🎉

PR #3683, opened 2023-12-06
H-1491: Version consistency for types which refer to themselves

CiaranMn:

An entity type may have a link where the expected type linked to is itself.

E.g . a `Company` may be a `Subsidiary of` `Company`, or `Company` may have `Invested In` `Company`.

Previously, it was impossible to have the versions of `Company` involved consistent via the UI, because there was no way of setting a link destination to be 'the next version of this thing', and creating a new version of the type would increment its version, meaning the version of the link destination was one step behind.

This PR keeps them consistent automatically by amending the schema to make the version of any 'self' link destinations the same as the type's next version.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Create a type which links to itself
2. Check that the version of the type and the link destination are the same

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entity-type-page.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity-type-page.tsx
@@ -1,4 +1,7 @@
-import { extractVersion } from "@blockprotocol/type-system";
+import {
+  EntityTypeReference,
+  extractVersion,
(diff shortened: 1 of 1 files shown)


benwerner01 (approved):
Thanks Ciaran, the changes look good and are working for me in the deployment 👍

PR #3685, opened 2023-12-07
H-1515: Add `/drafts` page

benwerner01:

This PR adds a draft entities page at `/drafts`, which lists the current draft entities for available for review for a user.

## 🔗 Related links

- H-1515

## 🔍 What does this change?

- makes fixes in the Node API to allow for archiving and updating draft entities (required for ignoring/accepting draft entities via the UI)

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Open entity in slide-over instead of redirecting to the entity editor (H-1544)
- Improve filters sidebar: add "Select All" filters, nested "Another user" nested checkbox (H-1543)
- Add source and destination entity names in the archive draft link warning modal (H-1541)

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch locally, as it requires backend fixes to function
2. Create draft entities (I did this locally by modifying the `createEntity` primitive function in the Graph API, but you could also use the browser extension)
3. Go to the `/drafts` page, and inspect the draft entities. You can "Ignore" or "Accept" them.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/ed1d4af5-6737-4587-b079-1c168deb18f4

### DIFF
--- a/apps/hash-frontend/src/components/block-loader/block-loader.tsx
+++ b/apps/hash-frontend/src/components/block-loader/block-loader.tsx
@@ -325,6 +325,14 @@ export const BlockLoader: FunctionComponent<BlockLoaderProps> = ({
     possiblyStaleSubgraph,
   ]);
 
+  /**
+   * If we are able to derive the `blockSubgraph` without the value from the context,
+   * set it in the context so it becomes available to other consumers of the context.
+   */
+  if (!possiblyStaleSubgraph && blockSubgraph) {
+    setBlockSubgraph(blockSubgraph);
+  }
+
   const lastFetchedBlockEntityId = useRef(blockEntityId);
 
   /**
--- a/apps/hash-frontend/src/pages/shared/block-collection/block-context-menu/block-select-data-modal.tsx
+++ b/apps/hash-frontend/src/pages/shared/block-collection/block-context-menu/block-select-data-modal.tsx
@@ -7,13 +7,13 @@ import {
 } from "@local/hash-isomorphic-utils/ontology-type-ids";
 import { simplifyProperties } from "@local/hash-isomorphic-utils/simplify-properties";
 import { QueryProperties } from "@local/hash-isomorphic-utils/system-types/blockprotocol/query";
-import { Entity, OwnedById } from "@local/hash-subgraph";
+import { Entity, EntityId, OwnedById } from "@local/hash-subgraph";
 import {
   getOutgoingLinkAndTargetEntities,
   getRoots,
 } from "@local/hash-subgraph/stdlib";
 import { Box, Typography } from "@mui/material";
-import { FunctionComponent, useCallback, useMemo } from "react";
+import { FunctionComponent, useCallback, useMemo, useState } from "react";
 
 import { useFetchBlockSubgraph } from "../../../../blocks/use-fetch-block-subgraph";
 import { useBlockProtocolCreateEntity } from "../../../../components/hooks/block-protocol-functions/knowledge/use-block-protocol-create-entity";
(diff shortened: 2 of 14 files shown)


apps/hash-frontend/src/shared/icons/calendar-day-light-icon.tsx
```diff
@@ -0,0 +1,16 @@
+import { SvgIcon, SvgIconProps } from "@mui/material";
+import { FunctionComponent } from "react";
+
+export const CalendarDayLightIcon: FunctionComponent<SvgIconProps> = (
+  props,
+) => {
+  return (
+    <SvgIcon {...props} height="16" width="14" viewBox="0 0 448 512">
+      <path
+        opacity="1"
```
CiaranMn, line 10: Is this required on these icons?

apps/hash-frontend/src/pages/drafts.page/draft-entities.tsx
```diff
@@ -0,0 +1,269 @@
...
+      return undefined;
+    }
+
+    return draftEntities.map((entity) => {
+      const firstRevision = getFirstRevision(
+        draftEntityHistoriesSubgraph,
+        entity.metadata.recordId.entityId,
+      );
+
+      const creator = users.find(
```
CiaranMn, line 123: This is to help me find where I need to update when we have non-user creators
```suggestion
      // @todo update to account for machine users – H-1514
      const creator = users.find(
```

apps/hash-frontend/src/pages/drafts.page/draft-entity/draft-entity-viewers.tsx
```diff
@@ -0,0 +1,114 @@
...
+          (viewer, index, all) =>
+            all.findIndex(({ accountId }) => accountId === viewer.accountId) ===
+            index,
+        );
+
+        return (
+          <>
+            Visible to <strong>{viewers.length} people</strong>
+          </>
+        );
```
CiaranMn, line 97: This is a design question for @nonparibus for follow-up, but I wonder if it's better to say 'viewable to members of X' somehow here, rather than just counting the members and leaving it to the user to figure out that this represents the membership count of the org

CiaranMn (commented):
Thanks @benwerner01, great work! Just a couple of small comments

apps/hash-frontend/src/pages/drafts.page/draft-entity/draft-entity-viewers.tsx
nonparibus: I'm not sure the number here will always refer to all members of the org. It may be a subset who can view entities of a certain type, etc. Ideally on hover over the number we would show which users have access in a tooltip.

apps/hash-frontend/src/shared/icons/calendar-day-light-icon.tsx
benwerner01: I don't think so, FA now attaches this for some reason to the SVGs when downloading. Have removed them in https://github.com/hashintel/hash/pull/3685/commits/953519aba56392a8ab28e29048980fa88023166c

apps/hash-frontend/src/pages/drafts.page/draft-entity/draft-entity-viewers.tsx
benwerner01: I've created [this follow-up ticket](https://linear.app/hash/issue/H-1550/display-tooltip-over-visible-to-x-people-to-indicate-whether-those-are) to track this, may require some designs if there are lots of users, or we want to include information about whether this has been shared with an org or specific users.

CiaranMn (approved):

CiaranMn (approved):

PR #3696, opened 2023-12-08
H-1487: Overhaul permissions to allow web-permission inheritance

TimDiekmann:

The current permission system is not flexible enough and we need a way to assign permission on a web-basis and optionally inherit that to the records inside of the web. Further, the `owner` relation is used in many locations and means different things.

## 🚫 Blocked by

- A data migration has to be written

## 🔍 What does this change?

- Implement a `setting` relationship to allow setting boolean flags
- Overhaul the permission schema to take the requirements listed above into accounts
- revisits almost _every_ creation of an entity in the Node API to set the new permission

This PR:

The changes in this PR:

The changes in this PR:


benwerner01 (approved):
Thanks Tim, the TS changes look good to me 👍

CiaranMn (approved):

PR #3709, opened 2023-12-11
H-589: Style cleanup around our slideovers

benwerner01:

This PR makes the following improvements to the type slideover on the entity type page:
- refactors the slideover to be a stack (opening a slideover within a slideover doesn't replace the existing one, but instead opens "on top" of it)
- adds "back" and "close" buttons to the slideover
- fixes the icons in the link destination entity type chip to be more consistent with the designs
- doesn't open the slideover when the selected type is already being displayed on the page (i.e. a circular reference is clicked), and instead scrolls the header into view

In addition this PR fixes a bug in the `block-loader` which was preventing the `blockSubgraph` from being fetched when the `blockSubgraph` is derived from the `blockCollectionSubgraph`. This fixes the entity editor modal that is opened when clicking "edit block" in the block dropdown.

## 🔗 Related links

- H-589

## 🔍 What does this change?

See commits/description

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

Click on a referenced entity type in the entity type editor to test out the slideover.

Click on "edit block" in the block dropdown to test that the entity editor modal works.

## 📹 Demo

Updated slideover:

https://github.com/hashintel/hash/assets/42802102/defb8952-2b66-4aa4-bf7d-ab803468e008

Updated chip icons and styling:

https://github.com/hashintel/hash/assets/42802102/efda02c1-9ce8-4e02-8d84-835c2880a77f

Clicking on a circular reference scrolls instead of opening a slideover:

https://github.com/hashintel/hash/assets/42802102/330563e4-5dea-45c7-b0b8-aef26c7670f4


nonparibus:
@benwerner01 Would it be trivial to add a 'forward' button here, as well? I don't think it, the back button, or the close button _need_ labels, although it's fine to have them (and maybe even nice) while none of the buttons are side-by-side. Introducing the forward button next to the back button will break this, though, and we should just remove the word labels if we introduce it.

nonparibus:
@benwerner01 Nice. Looks great. Thank you very much!

CiaranMn (approved):

PR #3711, opened 2023-12-11
H-1591: Provide a `--skip-validation` flag to snapshot restoration

TimDiekmann:

Snapshot restoration can be a very time consuming task currently if every entity is validated. This is especially the case if the database is on a remote server. As the validation could happen on a separate step, this provides the functionality to bypass validation

### DIFF
--- a/apps/hash-graph/libs/graph/src/snapshot/entity/batch.rs
+++ b/apps/hash-graph/libs/graph/src/snapshot/entity/batch.rs
@@ -172,7 +172,10 @@ impl<C: AsClient> WriteBatch<C> for EntityRowBatch {
         Ok(())
     }
 
-    async fn commit(postgres_client: &PostgresStore<C>) -> Result<(), InsertionError> {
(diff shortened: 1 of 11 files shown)


apps/hash-graph/libs/graph/src/snapshot/entity/batch.rs
```diff
@@ -223,46 +226,48 @@ impl<C: AsClient> WriteBatch<C> for EntityRowBatch {
...
-        )
-        .await
-        .change_context(InsertionError)?;
-
-        let schemas = postgres_client
-            .read_closed_schemas(&Filter::All(Vec::new()), None)
-            .await
-            .change_context(InsertionError)?
-            .try_collect::<HashMap<_, _>>()
+        if validation {
```
TimDiekmann, line 229: This was the only change here. If you hide whitespace changes it's more readable.

CiaranMn (approved):

PR #3714, opened 2023-12-12
H-1363, H-1411, H-1456, H-1462, H-1514: Add machine users, narrow permissions

CiaranMn:

This PR introduces a `Machine` actor in the system in addition to `User`s. Both `User` and `Machine` inherit from `Actor` (H-1514).

As part of this it:
- Creates a machine actor per web, which has the permission to grant global machine actors access to the web
- H-1411/H-1462: Adds the `Linear Integration` machine actor to webs when the integration is set up. Does the same for the `HASH AI` machine actor when the AI Inference feature is used.
- H-1363: Adds notifications for entities created/updated by the `HASH AI` actor
- H-1456: Restricts the instantiation of some entity types (`User`, `Notification`-related, the `HASH Instance` entity) to machine actors

## 🚫 Blocked by

- [x] Get migration working
- [x] Testing of the AI Inference feature
- [x] Finish handling new actors in frontend notifications / drafts page

This PR:

The changes in this PR:

The changes in this PR:


patches/@blockprotocol+graph+0.3.1-canary-20231207113055.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs b/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs
+index 8618e27..050e7ca 100644
+--- a/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs
++++ b/node_modules/@blockprotocol/graph/dist/cjs/codegen.cjs
+@@ -4,7 +4,7 @@ var path = require('node:path');
+ var slim = require('@blockprotocol/type-system/slim');
+ var jsonSchemaToTypescript = require('json-schema-to-typescript');
+ var fs = require('node:fs/promises');
+-var fetch = require('node-fetch');
++// PATCH: rely on native fetch
```
benwerner01, line 10: Just so I'm aware, which change in this PR required this? Maybe worth mentioning in the comment

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/001-create-hash-system-types.migration.ts
```diff
@@ -806,6 +832,7 @@ const migrate: MigrationFunction = async ({
       },
       webShortname: "hash",
       migrationState,
+      instantiator: null,
```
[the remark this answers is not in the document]
benwerner01: Realised later on in the review we are giving the web's machine user permission to instantiate the entity at runtime.

apps/hash-frontend/src/shared/use-actors.ts
```diff
@@ -0,0 +1,99 @@
...
+import { structuralQueryEntitiesQuery } from "../graphql/queries/knowledge/entity.queries";
+import { MinimalUser } from "../lib/user-and-org";
+
+export type MinimalActor =
+  | MinimalUser
+  | {
+      accountId: AccountId;
+      kind: "machine";
+      preferredName: string;
+    };
```
benwerner01, line 31: Can we make this `displayName` to match the property type? Or are you trying to avoid typing headaches that might be caused by that?
```suggestion
  | {
      accountId: AccountId;
      kind: "machine";
      displayName: string;
    };
```

libs/@local/hash-backend-utils/src/machine-actors.ts
```diff
@@ -0,0 +1,297 @@
...
+  const metadata = await context.graphApi
+    .createEntity(machineAccountId, {
+      draft: false,
+      entityTypeId: systemEntityTypes.machine.entityTypeId,
+      ownedById,
+      properties: {
+        [blockProtocolPropertyTypes.displayName.propertyTypeBaseUrl]:
+          displayName,
+        [systemPropertyTypes.machineIdentifier.propertyTypeBaseUrl]: identifier,
+      },
```
benwerner01, line 173: Nit: if I recall correctly we wanted to start hardcoding the base URLs directly? Or did is this no longer the case. Appreciate this won't be as type-safe because we can't use `satisfies` as we discussed previously.
```suggestion
      properties: {
        "https://blockprotocol.org/@blockprotocol/types/property-type/display-name/":
          displayName,
        "https://hash.ai/@hash/types/property-type/machine-identifier/":
          identifier,
      } as MachineProperties,
```

apps/hash-frontend/src/shared/notifications-context.tsx
```diff
@@ -403,8 +419,58 @@ export const NotificationsContextProvider: FunctionComponent<
...
+              isLinkAndRightEntityWithLinkType(
+                systemLinkEntityTypes.occurredInEntity.linkEntityTypeId,
+              ),
+            );
+
+            if (!occurredInEntityLink) {
+              throw new Error(
+                `Graph change notification "${entityId}" is missing required links`,
+              );
+            }
```
benwerner01, line 436: Mentioned in slack, but I think this is going to cause a crash at runtime when a draft entity is discarded in the drafts page, as it will also archive all incoming links which it will be able to do as we don't have link validation preventing required links from being removed if I recall correctly.

benwerner01: We could either handle the link being missing gracefully here, or archive the notification when the draft is discarded so that it isn't being rendered here.

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/001-create-hash-system-types.migration.ts
```diff
@@ -665,6 +688,7 @@ const migrate: MigrationFunction = async ({
     },
     webShortname: "hash",
     migrationState,
+    instantiator: anyUserInstantiator,
```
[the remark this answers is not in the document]
CiaranMn: Yes, exactly – we (or rather @TimDiekmann) will be adjusting the snapshots when we restore them after deploying this, to correct the instantiators

apps/hash-api/src/graph/ensure-system-graph-is-initialized/system-webs-and-entities.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * Mop up step: create web machine actors for existing webs – bots with permissions to add other bots to each existing web,
+   * and to create notifications that aren't tied to specific integrations (e.g. related to comments and @mentions).
+   *
+   * This step is only required to transition existing instances in Dec 2023, and can be deleted once they have been migrated.
+   */
+  const users = await getEntitiesByType(context, authentication, {
+    entityTypeId: "https://hash.ai/@hash/types/entity-type/user/v/1", // @todo this may need to change depending on migration strategy
+  });
```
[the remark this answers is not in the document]
CiaranMn: Yes - I started off with it in the migration file, but on a fresh instance we don't have the `hash` machine entity available until this point, because it can't be created until after the last migration file has run. In theory we could create it in the migration file but it started to feel like a lot of moving things around that would be long-term less satisfactory to avoid this short term hack.

The `hash` machine entity has to exist because we use it as part of giving other machines permissions, and since all that code is currently in `backend-utils` I didn't want to import the `systemAccountId` from `@apps/hash-api` to retrieve the actor, but instead by searching for the entity. The `createMachineActor` code could probably live separately in the API, but it again felt like a compromise which would would be worse in the long term.

apps/hash-frontend/src/shared/notifications-context.tsx
CiaranMn: I will do this in follow-up

patches/@blockprotocol+graph+0.3.1-canary-20231207113055.patch
CiaranMn: Done in https://github.com/hashintel/hash/pull/3714/commits/3f5cae97576f20eb68b67a2fc3f950f742d13e32

libs/@local/hash-backend-utils/src/machine-actors.ts
CiaranMn: Done in https://github.com/hashintel/hash/pull/3714/commits/53ddf2e361b24932e24b2e9709dff4622c71ef3b

apps/hash-frontend/src/shared/use-actors.ts
CiaranMn: Done in https://github.com/hashintel/hash/pull/3714/commits/53ddf2e361b24932e24b2e9709dff4622c71ef3b

benwerner01 (dismissed):

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/util.ts
```diff
@@ -642,21 +675,331 @@ export const createSystemEntityTypeIfNotExists: ImpureGraphFunction<
...
+    ).catch((createError) => {
       // logger.warn(`Failed to create entity type: ${entityTypeSchema.$id}`);
       throw createError;
     });
 
     return createdEntityType;
   }
 };
+
+export const getExistingHashSystemEntityTypeId = ({
```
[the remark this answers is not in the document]
CiaranMn: Done in https://github.com/hashintel/hash/pull/3714/commits/4478b757dc0031994b4ff78ea14b7182ca6bcc08

TimDiekmann (approved):

benwerner01 (dismissed):

benwerner01 (approved):

TimDiekmann (approved):

PR #3724, opened 2023-12-13
H-1021, H-1627, H-1639: Default types in the browser plugin, inference improvements

CiaranMn:

Makes the following improvements to the browser plugin:
- H-1021: Add default types for websites in production (types aren't available locally)
- H-1627: Add a 'remove row' button on the automatic inference config
- H-1639: Fix a black bar appearing at the bottom of the Automated tab mode
- Prevent automatic inference on a page within 30 seconds of a previous automatic inference finishing on the same page

And to the AI Inference Temporal worker:
- Get the underlying entity validation errors back to the LLM, and provide it with the schema to match (not clear how it receives the function definition but it doesn't seem able to match the function definition schema to an entity type on repeat tries)
- If the LLM provided types which weren't requested, don't immediately return to retry but bank any valid creations/updates it has also provided
- Consistently don't prettify logs in production
- Provide a record of successful creations/updates from previous iterations to the `createEntities` function in case a retried creation of a link refers to entities from previous iterations
- Add a couple more name properties to look for when entity matching

**Drive-by:** add garbage collection logging to the Node API to track down memory issue in prod – will be noisy in logs, should remove this after confirming if the Node API does or does not have a leak.

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-ts/src/activities/infer-entities/create-entities.ts
```diff
@@ -81,7 +92,13 @@ export const createEntities = async ({
...
           const { properties = {} } = proposedEntity;
 
-          const nameProperties = ["name", "display-name", "preferred-name"];
+          const nameProperties = [
+            "name",
+            "display-name",
+            "legal-name",
+            "preferred-name",
+            "profile-url",
+          ];
```
benwerner01, line 101: Think I asked previously, wondering why we don't include `title` in this list
```suggestion
          const nameProperties = [
            "name",
            "display-name",
            "legal-name",
            "preferred-name",
            "profile-url",
            "title",
          ];
```

benwerner01 (dismissed):

apps/hash-ai-worker-ts/src/activities/infer-entities/create-entities.ts
CiaranMn: My worry with `title` is that it's also used for things that are less likely to be unique, for example a job title, or Mr/Mrs

apps/plugin-browser/src/pages/popup/popup-contents/action-center.tsx
```diff
@@ -71,7 +72,8 @@ export const ActionCenter = ({
       enabled: false,
       model: "gpt-4-turbo",
       ownedById: user.webOwnedById,
-      rules: [],
+      rules:
+        FRONTEND_ORIGIN === "https://app.hash.ai" ? defaultProductionRules : [],
```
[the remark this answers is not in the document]
CiaranMn: Yes (we should figure out how to deal with future changes once people have it installed)

benwerner01 (approved):

PR #3731, opened 2023-12-14
H-1483: Support embedding querying for entities

TimDiekmann:

To allow similarity search for entities, this adds the minimum amount of functionality to upset embeddings and querying it.

## 🚫 Blocked by

- #3726 

## 🔍 What does this change?

- Add the `pgvector` extension
- Add embedding tables to database
- Allow insert/update embeddings for an entity
    - Each entity may have a global embedding for the whole entity and one embedding per property.
- Allow querying by embeddings using either the L2 distance, cosine distance, or inner product

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Currently, it's not possible to remove embeddings again

## 🛡 What tests cover this?

- A few smaller HTTP tests were added, these tests use hard-coded embeddings to (a) avoid the need to generate the embeddings on the fly and (b) ensure that the result is deterministic


apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1086,6 +1088,70 @@ impl<C: AsClient> EntityStore for PostgresStore<C> {
...
+        actor_id: AccountId,
+        authorization_api: &mut A,
+        embeddings: impl IntoIterator<Item = EntityEmbedding<'_>> + Send,
+    ) -> Result<(), UpdateError> {
+        #[derive(Debug, ToSql)]
+        #[postgres(name = "entity_embeddings")]
+        pub struct EntityEmbeddingsRow<'a> {
+            web_id: OwnedById,
+            entity_uuid: EntityUuid,
+            property: Option<String>,
```
CiaranMn, line 1103: Maybe we should call this `property_base_url`? It isn't clear what it is in a few places

TimDiekmann: I'm not sure about it as we also have e.g. `label_property` and also the type suggests that this is a `BaseUrl` (in this specific case it's a string because it's a local definition which matches the Postgres table and as `BaseUrl` does not implement the `ToSql` trait, so we cannot use that here. I expect this to change once we move the type-system crate). In the database, `property` references the `base_urls` tables.

CiaranMn (approved):

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Talked about this offline, we're not worrying about this now as `label_property` should then also be renamed to `label_property_base_url`. If this becoming confusing we're going to change it.

PR #3735, opened 2023-12-14
H-1648: Move AI inference endpoint to websocket

CiaranMn:

AI inference requests can often take a long time, which in production falls foul of Cloudflare's 100 second maximum wait for a response.

This PR moves inference to a websocket to address this.

I wasn't able to get the websocket handshake request to include the user's cookie for the domain (the previous `fetch` request did so from the same script to the same domain) – this may be a restriction regarding cross-domain insecure cookies (as ours are locally). So I'm getting and sending it manually, which is not ideal – we should see if there's some way to have the cookie sent without needing the permission to read it.

I also added something to ensure that property keys always have a trailing slash, because in testing I noticed the AI forgets these still sometimes.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Run the API, run the plugin, check it works


benwerner01 (approved):
Thanks, this is working for me locally

PR #3736, opened 2023-12-14
H-1668: Change global allocator to `mimalloc`

TimDiekmann:

We encountered a memory leak in our application. After endless debugging and false positives one solution which seems to work is switching out the global allocator. I'm not going into the details why this could solve this, but the TL;DR is that different allocators handle certain scenarios differently.

## 🔍 What does this change?

- Make `mimalloc` the global allocator


indietyp (approved):
let's hope this improves things 🤞

PR #3742, opened 2023-12-15
H-1691: Show types in sidebar which user has entities of in their web

CiaranMn:

Currently, if a user creates entities using types outside their web, it's hard to actually navigate to a list of entities of those types in the UI, because the sidebar only shows types in the user's web.

This PR makes it so that it will also show types which the user has any entities of in their web, excluding some system types. 

Note that this will not show types of entities in _organization_ webs the user is a member of (nor does the existing sidebar show types which belong to other webs the user is a member of). We could do this, although it'll involving fetching more data (all entities in all orgs the user is a member of). We should probably add 'pin type' logic soon, to allow users to customise the sidebar.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Assuming you have some entities of types from outside your web, visit the preview deployment


apps/hash-frontend/src/pages/shared/hidden-types.ts
```diff
@@ -0,0 +1,21 @@
+import { VersionedUrl } from "@blockprotocol/type-system/slim";
+import { systemEntityTypes } from "@local/hash-isomorphic-utils/ontology-type-ids";
+
+/**
+ * Entity types the user may have entities of that should be excluded from view in the UI
```
nonparibus, line 5: This excludes link entity types, right?

nonparibus (approved):

PR #3749, opened 2023-12-16
H-1708: Fix SpiceDB migration not working in prod

TimDiekmann:

Last time when we updated SpiceDB the migration failed due to the wrong connection URI. This was because the migration connection URI was overwritten by the default URI.


nonparibus (approved):
CC @CiaranMn

PR #3753, opened 2023-12-18
H-1472: Support loading entities into table block

benwerner01:

This PR allows for loading entities into the table block via the "Select data" modal.

## 🔗 Related links

- H-1472

## 🔍 What does this change?

- updates the block to use `https://blockprotocol.org/@hash/types/entity-type/table/v/1` and related production types (previously it was using types created in a staging environment)
- adds the "Has Query" link to the block's type, so that the select data modal is displayed in the block's dropdown
- uses the property type's title to name columns, rather than using the slug from its base UrL
- cleaned up some misleading variable names

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch, and load the table block into a locally running instance of HASH
2. Try using the table block in its default mode, where data/columns are manually added
3. Try using the "Select data" modal via the block dropdown to select entities to be rendered in the table block. This should work for more entities of more than one type.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/8ebeaf1b-05ef-415d-8683-5895798b8319

### DIFF
--- a/blocks/table/src/components/table/table-with-query.tsx
+++ b/blocks/table/src/components/table/table-with-query.tsx
@@ -1,12 +1,16 @@
 import {
+  BaseUrl,
   Entity,
   EntityRootType,
+  extractBaseUrl,
+  extractVersion,
   GraphBlockHandler,
   JsonValue,
   MultiFilter,
+  PropertyType,
   Subgraph,
 } from "@blockprotocol/graph";
-import { getRoots } from "@blockprotocol/graph/stdlib";
+import { getPropertyTypes, getRoots } from "@blockprotocol/graph/stdlib";
 import {
   DataEditorProps,
   DataEditorRef,
@@ -50,6 +54,8 @@ export const TableWithQuery = ({
   } = blockEntity;
 
   const [loading, setLoading] = useState(true);
+
+  const [subgraph, setSubgraph] = useState<Subgraph<EntityRootType>>();
   const [entities, setEntities] = useState<Entity[]>([]);
 
   useEffect(() => {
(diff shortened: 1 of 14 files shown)


blocks/table/src/components/table/table-with-query.tsx
```diff
@@ -72,26 +80,75 @@ export const TableWithQuery = ({
         throw new Error(res.errors?.[0]?.message ?? "Unknown error");
       }
 
-      const roots = getRoots(res.data.results);
-      setLoading(false);
+      /** @todo: fix the return type */
+      const fetchedSubgraph = res.data as unknown as Subgraph<EntityRootType>;
```
CiaranMn, line 84: Can we fix it in `knowledge-shim.ts`? The problem is that the patch there has been defined to have a type which is not consistent with the type defined in the BP. 

It should be
```ts
export type QueryEntitiesMessageCallback = MessageCallback<
  QueryEntitiesRequest,
  null,
  MessageReturn<{
    results: Subgraph<EntityRootType>;
    operation: QueryOperationInput;
  }>,
  ReadOrModifyResourceError
>;
```

then we can update `useBlockProtocolQueryEntities` to return the new shape, and update the couple other places we use the hook to take `data.results` instead of just `data`. Saves this tripping us up in the future

benwerner01: Ah thank you for explaining why this was happening, have addressed in https://github.com/hashintel/hash/pull/3753/commits/f66360e9ce3d002be737693c741e10aa41b2087c

CiaranMn (approved):

PR #3756, opened 2023-12-18
H-1735: Validate more custom data types

TimDiekmann:

Add support for some more data type validations:
- `minimum`
- `maximum`
- `exclusiveMinimum`
- `exclusiveMaximum`
- `multipleOf`
- `format = "uri"`
- `format = "email"`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -655,6 +655,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
 
 [[package]]
+name = "email_address"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2153bd83ebc09db15bcbdc3e2194d901804952e3dc96967e1cd3b0c5c32d112"
+
+[[package]]
 name = "encoding_rs"
 version = "0.8.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3376,6 +3382,7 @@ dependencies = [
 name = "validation"
 version = "0.0.0"
 dependencies = [
+ "email_address",
  "error-stack",
  "graph-test-data",
  "graph-types",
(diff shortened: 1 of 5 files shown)


nonparibus:
URI inc. URN, or just URLs? I was caught off-guard by the PR desc.

CiaranMn (approved):

TimDiekmann:
> URI inc. URN, or just URLs? I was caught off-guard by the PR desc.

JSON Schema only allows `uri`: https://json-schema.org/draft/2020-12/json-schema-validation#name-uri-template

nonparibus:
@TimDiekmann So `urn:isbn:9788184894677` would validate as correct for `URL`, even though it is a `URN`?

I think we want to capture `URN`s and `URL`s as separate **Data Types** (both of which would, in the future, exist within a single `URI` **Data Type Group**... assuming my mental map of how all these things inter-relate is correct).

TimDiekmann:
Yes, it would. I also just tested your provided `URN` and it's passing. I agree that we probably want to support more than `URI` (especially as `URI` covers _a lot_), but this is supposed to be part of this PR as this would require to allow custom regexes. Properly validating URLs/URIs/URNs is an extremely ambitious goal.

nonparibus:
Full validation is a lot of work. But basic checks can be done without it being "extremely ambitious", and must be done for us to move on to phase 2 CDTs.

My concern right now is that we want to call the data type, in the frontend, "URL" -- not URI. However, at present here, we accept URNs, not just URLs... so we must call it URI.

We can do that, and ship the thing we have now as "URI" in the user-facing app, because that is what it is. But while "good enough" for what we want right now, it is also not quite what we originally specced out here... and we should track the divergence. @TimDiekmann @CiaranMn

TimDiekmann:
If we would add basic checks to check URLs ourselves we would match _way_ more than the check which are provided by the Rust crate. In addition I'm sure we would miss a lot of valid URLs and our validation would forbid them which results in super bad UX. What is the reason we cannot just use "URL" in the frontend? Every URL will be valid. But we cannot guarantee, that everything you throw at it will actually be a valid URL. Unless we find a URL validator which guarantees that we cannot provide that functionality.

PR #3759, opened 2023-12-18
H-1544, H-1592, H-1748: Open draft entities in the slideover instead of the entity editor

benwerner01:

This PR opens the entity editor in a slide-over when a draft entity is clicked in the `/drafts` page, replacing the existing inline collapsible properties table.

As a drive-by, this PR also:

- fixes an issue in the entity editor where it displays the button for upgrading an entity's type in readonly mode, resulting in a crash when the button is clicked. This was also an issue in the type editor. (H-1592)
- sets the notification and draft queries which are polled at intervals to use `"network-only"` mode, so that previous query results aren't persisted in memory
- fixes the draft entity type filter to only display the latest version of the types in use on the `/drafts` page
- filter external types from blockprotocol.org from the entities section in the sidebar, except for the `Address` entity type.
- change default sort order on the `/drafts` page to be newest first (H-1748)
- move expensive queries out of the global notifications and draft entities contexts

## 🔗 Related links

- H-1544

## 🔍 What does this change?

See description/commits.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Click on a draft entity in the `/drafts` page, which should open the entity editor in a slide-over
2. Try editing the entity, and click "Save changes" to update the entity

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/f2f40398-915d-41b7-bfb7-a4529eb4fd8d


apps/hash-frontend/src/pages/drafts.page/draft-entity.tsx
```diff
@@ -40,14 +56,51 @@ export const DraftEntity: FunctionComponent<{
...
+   * @todo: figure out what the underlying issue is causing the `EntitySelector`
+   * component to re-mount when the subgraph changes.
+   */
+  if (
+    previouslyEvaluatedEntity.current.metadata.recordId.editionId !==
+    entity.metadata.recordId.editionId
+  ) {
+    previouslyEvaluatedEntity.current = entity;
+    setEntityRootedSubgraph(generateEntityRootedSubgraph(entity, subgraph));
+  }
```
benwerner01, line 86: @CiaranMn I spent a significant amount of time trying to figure out the root cause of this, as it should be addressed in the entity editor instead of here. This also unfortunately doesn't fully resolve the issue, as when a property is updated in the entity editor there is are updates to the `entitySubgraph` which cause the entity selector to re-mount. This is obvious when updating a property, and in quick succession trying to add a link:

https://github.com/hashintel/hash/assets/42802102/2341dcf6-427f-4656-9a2c-d636667c99de

I'm quite confused by what's causing the re-mount as the text fields in the property editor are not affected by this, and it's not happening on the dedicated entity editor page. This means it must be something specific to how the subgraph is passed to the entity editor in the `EditEntityModal`, I just haven't been able to think of what that might be.

nonparibus:
n.b. this _may_ also fix H-1734 (client-side memory usage issue), but monitoring to follow... along with an audit to ensure there are no other missed instances/contributors.

apps/hash-frontend/src/pages/drafts.page/draft-entity.tsx
CiaranMn: My theory (might be wrong) is that it could be because you use `entity` as a dependency to the intended approach, which causes the subgraph to be recalculate whenever the object's identity changes. Using `entityId` as the dependency instead might fix it
```ts
   * const entityRootedSubgraph = useMemo<Subgraph<EntityRootType>>(
   *   () => generateEntityRootedSubgraph(entity, subgraph),
   *   [subgraph, entity],
   * );
 ```

benwerner01: I've tried this and it does not fix the issue. Separately I've tested selecting files in the entity selector to verify the subgraph changes still enable previewing/linking to those entities:

image

CiaranMn: Cool, thanks Ben – we talked about this and I made a mistake in my view above, because what you're doing isn't the same as making `entityId` a dependency of the original approach. It's excluding `subgraph` as a dependency of recalculating subgraph. But if things still seem to work let's merge it for now and figure out the issue with the selector in follow-up

CiaranMn (approved):

PR #3761, opened 2023-12-18
H-1633, H-1707: Group by location in autoinference config, persist user tab selection

CiaranMn:

Browser extension improvements:
- H-1633: Allow toggling between "Group by type" and "Group by location" in autoinference config
- H-1707: Persist user tab selection
- Handle a delay in the page title updating after DOM content is loaded during client-side navigation (add a delay to leave time for the client-side JS to update the page title before taking page content)
- Fix dark mode for `Switch`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Because browser extension API access is asynchronous, there is a flash when opening the popup before the tab is known and can be displayed. The height of the tab is unknown ahead of time. I experimented a bit and ended up with just not showing anything as the least offensive transition.
- There is a flash when switching between Location and Type because for some reason the Autocomplete does not render the selected values straight away (if there are many, this means the dropdown and therefore row height increases) 

## 🛡 What tests cover this?

- Manual

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/55085310-a712-49c9-8c0d-044d0e219103


apps/plugin-browser/src/pages/popup/popup-contents/action-center/switch-with-dark-mode.tsx
```diff
@@ -0,0 +1,23 @@
+import { Switch, switchClasses, SwitchProps } from "@mui/material";
+
+export const SwitchWithDarkMode = ({ sx, ...props }: SwitchProps) => {
+  return (
+    <Switch
+      {...props}
+      sx={[
```
benwerner01, line 7: Why are we not making changes like this in the theme itself? Do we not for-see this being re-usable?

apps/plugin-browser/src/pages/popup/popup-contents/action-center/automated/select-scope/rows-by-location.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+      return (a.restrictToDomain ?? "") > (b.restrictToDomain ?? "") ? 1 : -1;
+    });
+
+    return sortedRules;
+  }, [draftRule, rules]);
+
+  return (
+    <>
+      {rulesByLocation.map((rule) => (
```
benwerner01, line 328: If we can, might be nice to sort by when a row was added to prevent re-orderings whilst making changes:

https://github.com/hashintel/hash/assets/42802102/c66480c8-8b61-4709-86f6-4ac493f5c3b0

benwerner01 (approved):
Thanks Ciaran, the changes look good and are working for me locally. Had a suggestion, but doesn't block merging.

apps/plugin-browser/src/pages/popup/popup-contents/action-center/switch-with-dark-mode.tsx
CiaranMn: Basically because I don't want to spend time translating it to the theme yet when it's only used here (I'm slight apprehensive about how much overriding stuff is going on in the theme and breaking something else).

I agree we should move it in there when we need dark mode elsewhere

apps/plugin-browser/src/pages/popup/popup-contents/action-center/automated/select-scope/rows-by-location.tsx
CiaranMn: I agree this would be nice – I've added H-1786 to come back to it (either today if I get time, or in Jan if I don't)

PR #3768, opened 2023-12-19
H-1766: Support more validation constrains defined in JSON Schema

TimDiekmann:

Adds support for
- `enum`
- `minLength`
- `maxLength`
- `pattern`

### DIFF
--- a/libs/@local/hash-validation/src/data_type.rs
+++ b/libs/@local/hash-validation/src/data_type.rs
@@ -333,6 +333,7 @@ fn check_string_additional_property<'a>(
 impl<P: Sync> Schema<JsonValue, P> for DataType {
     type Error = DataValidationError;
 
+    #[expect(clippy::too_many_lines)]
     async fn validate_value<'a>(
         &'a self,
         value: &'a JsonValue,
@@ -354,7 +355,6 @@ impl<P: Sync> Schema<JsonValue, P> for DataType {
                     expected: JsonSchemaValueType::Boolean,
                 }
             ),
-
             "number" => {
                 #[expect(clippy::float_arithmetic)]
                 check_numeric_additional_property(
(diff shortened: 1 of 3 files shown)


libs/@local/hash-validation/src/data_type.rs
```diff
@@ -58,31 +59,65 @@ pub enum DataTypeConstraint {
...
         expected: JsonValue,
     },
+    #[error(
+        "the provided value is not longer than the minimum length, got `{actual}`, expected \
+         `{expected}`"
+    )]
+    MinLength { actual: String, expected: usize },
+    #[error(
+        "the provided value is not shorter than the maximum length, got `{actual}`, expected \
+         `{expected}`"
```
CiaranMn, line 114: 'not longer' implies an exclusive minimum
```suggestion
        "the provided value is shorter than the minimum length, got `{actual}`, expected \
         a string of at least length `{expected}`"
    )]
    MinLength { actual: String, expected: usize },
    #[error(
        "the provided value is longer than the maximum length, got `{actual}`, expected \
         a string of at most length `{expected}`"
```

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -573,4 +695,63 @@ mod tests {
...
+    async fn simple_mail() {
+        let mail_type = serde_json::to_string(&json!({
+            "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
+            "kind": "dataType",
+            "$id": "https://localhost:4000/@alice/types/data-type/simple-email/v/1",
+            "title": "E-Mail",
+            "type": "string",
+            "pattern": "^[^@]+@[^@]+$",
+        }))
+        .expect("failed to serialize simple mail type");
```
CiaranMn, line 709: I think it's confusing that we use an email to test `pattern` when an email schema should be expressed using `format: "email"` (and we don't test `format: "email"`) – can we pick some other regexp to test `pattern`?

TimDiekmann: I changed it to `"Zip Code (US)"`, which is likely a good use for the pattern-property.

CiaranMn (approved):

CiaranMn (approved):

PR #3789, opened 2023-12-22
H-1788, H-1791: Add more common custom data type validations

TimDiekmann:

This adds CDT validation (and support) for:
- `ipv4`
- `ipv6`
- `hostname` (Either an IP or domain)
- `uuid`
- `regex`
- `time`

Also, I removed the `chrono` dependency and added a simple regex for date parsing. In contrast to `chrono` or `time` this does only validate dates which conforms to both, ISO-8601 and RFC 3336.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Various tests and test cases were added/adjusted

### DIFF
--- a/libs/@local/hash-validation/src/data_type.rs
+++ b/libs/@local/hash-validation/src/data_type.rs
@@ -1,5 +1,8 @@
 use core::{borrow::Borrow, fmt};
-use std::str::FromStr;
+use std::{
+    net::{Ipv4Addr, Ipv6Addr},
+    str::FromStr,
+};
 
 use chrono::{DateTime, NaiveDate};
 use email_address::EmailAddress;
@@ -10,6 +13,7 @@ use serde_json::Value as JsonValue;
 use thiserror::Error;
 use type_system::{url::VersionedUrl, DataType, DataTypeReference};
 use url::Url;
+use uuid::Uuid;
 
 use crate::{
     error::{Actual, Expected},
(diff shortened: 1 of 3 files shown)


libs/@local/hash-validation/src/data_type.rs
```diff
@@ -245,6 +250,17 @@ where
 }
 
 fn check_format(value: &str, format: &str) -> Result<(), Report<DataValidationError>> {
+    // Only the simplest date format are supported in all three, RFC-3331, ISO-8601 and HTML
+    const DATE_REGEX_STRING: &str = r"(?P<Y>\d{4})-(?P<M>\d{2})-(?P<D>\d{2})";
+    const DATE_REGEX: OnceCell<Regex> = OnceCell::new();
+
+    // Only the simplest date format are supported in all three, RFC-3331, ISO-8601 and HTML
```
TimDiekmann, line 257:
```suggestion
    // Only the simplest time format are supported in all three, RFC-3331, ISO-8601 and HTML
```

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -757,6 +933,182 @@ mod tests {
...
+        .expect_err("validation succeeded");
+
+        _ = validate_data(
+            json!("localhost:3000"),
+            &hostname_type,
+            ValidationProfile::Full,
+        )
+        .await
+        .expect_err("validation succeeded");
+
```
TimDiekmann, line 1081:
```suggestion
```

libs/@local/hash-validation/src/data_type.rs
```diff
@@ -786,377 +1134,375 @@ mod tests {
...
+            "2023-12-22_17:48:15.083212Z",      // %Y-%M-%D_%h:%m:%s.%uZ
+            "2023-12-22_17:48:15.083z",         // %Y-%M-%D_%h:%m:%.3sz
+            "2023-12-22_17:48:15.083212z",      // %Y-%M-%D_%h:%m:%s.%uz
+            "2023-12-22T18",                    // %Y-%M-%DT%h
+            "2023-12-22T18,8",                  // %Y-%M-%DT%,1h
+            "2023-12-22T18.8",                  // %Y-%M-%DT%.1h
+            "2023-12-22T18:48",                 // %Y-%M-%DT%h:%m
+            "2023-12-22T18:48,2",               // %Y-%M-%DT%h:%,1m
+            "2023-12-22T18:48.2",               // %Y-%M-%DT%h:%.1m
+            "2023-12-22T18:48:15",              // %Y-%M-%DT%h:%m:%s
```
CiaranMn, line 1185: This is the format that the HTML `datetime-local` input outputs (the only native browser datetime picker), so we need to decide what to do to make it into a string with timezone, otherwise the value generated by the input will not validate.

Two options:
1. Whatever datetime the user picked, we store it as UTC (no conversion), i.e. just add a `Z`
    - pros: simple. closest to making no assumptions about what the user intended (although still involves an assumption that they _didn't_ want their timezone preserved)
    - cons: we don't know what timezone the user was in when they picked the datetime
2. We calculate the UTC offset from whatever timezone is set in their browser/device and store that
    - pros: we preserve the user's timezone (at least according to the browser)
    - cons: maybe they didn't want this

In both scenarios we could turn a 'with timezone' back into 'without timezone' if we wanted.

We will need to either 
1. make a decision about this and update the frontend to implement it before merging the PR, or
2. allow timestamps without timezones

We should also update the description of the `DateTime` data type to reflect the decision.

cc @nonparibus as it's probably worth discussing

TimDiekmann: I would probably prefer to add two data types, one with timezone, one without timezone. Mainly, because none of the three solution is satisfying.

CiaranMn: We discucssed and agreed to do (2) for now, but to add an explicit 'DateTime without TZ' later

CiaranMn: I will do (2) in the next ~day, we can merge this for now since we won't be attempting to input datetimes via the frontend in the meantime

CiaranMn (approved):

libs/@local/hash-validation/src/data_type.rs
TimDiekmann: To avoid confusion in the future do we want to explicitly mention in the description that the date time is with a timezone in the `Date Time` data type"? Something like
> A date and time with time zone in the format of RFC 3336.

CiaranMn: Yes, I think we should do that. Can probably update the seed script and then amend in db directly to avoid a migration (as no-one else has run the data type migration, AFAIK)

TimDiekmann: Sounds good!

PR #3815, opened 2024-01-03
H-1800: Fix Playwright setup, existing tests

CiaranMn:

This PR fixes Playwright tests timing out in CI – this was because:
- `test:system` depended on `build` in the Turbo configuration
- meaning that all build commands were run before the Playwright tests
- which caused the frontend (which had already been built and run in a previous step) to be rebuilt
- in a way which meant it loaded but some of the JavaScript file URLs were wrong
- recreating this locally requires bypassing the Turbo cache

The PR fixes the above by removing `build` as a dependency of running `test:system`. The API and frontend must already be built and running for tests to work, nothing needs building again when the actual test command is run. 

It also:
- removes `build` as a dependency of the `dev` command from the root `turbo.json` (individual apps can add dependencies of `build` if their `dev` command depends on it)
- fixes and unskips some existing tests
- adds a test for the signup flow

This PR:

The changes in this PR:

The changes in this PR:

## Next Steps

1. Make the Playwright tests required to merge, and monitor any remaining flakiness
2. Write more tests

## 🛡 What tests cover this?

- The Playwright tests.

## ❓ How to test this?

1. Check that the test run GitHub action passes.

## 📹 Demo

Screenshot 2024-01-04 at 16 30 17

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -464,7 +464,7 @@ jobs:
         run: |
           turbo run build --filter "@apps/hash-frontend"
           yarn workspace @apps/hash-frontend start 2>&1 | tee var/logs/frontend.log &
-          yarn wait-on --timeout 30000 http://0.0.0.0:3000
+          yarn wait-on --timeout 60000 http://0.0.0.0:3000
 
       - name: Show disk usage
         run: df -h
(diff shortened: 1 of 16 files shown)


benwerner01 (approved):
Nice one 🎉

PR #3822, opened 2024-01-04
H-1598: Fix handling of links to externally-hosted types

CiaranMn:

Fixes a couple of bugs related to links to external types, i.e. types which originate from instances of HASH other than the one running:
1. Respect `SELF_HOSTED_HASH` environment variable in the frontend, if it is set (rather than always treating a `localhost:3000` origin as being HASH prime). This allows testing of self-hosted instance frontends locally
2. Only use the special external type page for entity types (since there is no dedicated page for property or data types, whether external or not)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Set `SELF_HOSTED_HASH=true` in `.env.local` and run the app (don't need to reseed)
2. Visit `/types`
3. Click on a HASH entity type, check it loads in the external route
4. Click on a HASH property type, check that the JSON is loaded rather than the external route

### DIFF
--- a/apps/hash-api/src/graph/ontology/primitive/util.ts
+++ b/apps/hash-api/src/graph/ontology/primitive/util.ts
@@ -20,6 +20,7 @@ export const isExternalTypeId = (typeId: VersionedUrl) =>
   !typeId.startsWith(frontendUrl) &&
   // To be removed in H-1172: Temporary provision to serve types with a https://hash.ai URL from https://app.hash.ai
   !(
+    process.env.NEXT_PUBLIC_SELF_HOSTED_HASH !== "true" &&
     ["https://app.hash.ai", "http://localhost:3000"].includes(frontendUrl) &&
     new URL(typeId).hostname === "hash.ai"
   );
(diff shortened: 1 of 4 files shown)


nonparibus:
@CiaranMn Do we have any kind of test for this? If not can we bundle one with this PR?

CiaranMn:
> @CiaranMn Do we have any kind of test for this? If not can we bundle one with this PR?

We don't – I've created [a follow-up task](https://linear.app/hash/issue/H-1807/add-test-for-behavior-of-external-types) because it'll require figuring out how to actually get to click on a link to an external type via Playwright. The easiest UI route is via the 'view all types' page, but because the table is canvas we need to figure out the pixel coordinate of the relevant type to click on. I'm not sure how we will do that off-hand in Playwright so suggest deferring as not the most best cost-benefit test we could write now if we want to spend more time on writing tests.

benwerner01 (commented):
Thanks Ciaran, this is working locally for me for the HASH entity/property types.

I noticed one other thing, which might be worth addressing as part of this PR. When clicking on BP property types, we probably want to redirect to the blockprotocol.org. Currently it sticks to trying to find something at the frontend domain of the current instance, resulting in an error:
image

benwerner01 (approved):

PR #3830, opened 2024-01-05
H-1808: Fix mention bugs, editor crash on property object arrays, and page content reloading

CiaranMn:

The PR started off to fix the `@mention` suggester crashing, and fixed a few other things along the way:
1. Handle missing target entities of links in the mention suggester. 
    - the cause of the original crash – the mention suggester doesn't look for draft entities, and so it picks up 'live' notification entities which point to draft entities, which aren't available
2. Save the `linkEntityTypeBaseUrl` when inserting a mention of the target of a link (this information was not being saved before, meaning the link type wasn't available when returning to the page)
3. Handle type inheritance in the mention display
4. Hide some system types from the mention suggester

Things I fixed in passing because I noticed them when trying to test this:
1. The entity editor was crashing when viewing entities with an array of property objects (e.g. Service Feature entities) – there's no UI implemented for arrays of objects so I'm just taking the first for now (to be addressed as part of H-1785)
2. The page fetching query wasn't set to `cache-and-network`, meaning it didn't update properly when navigating away and back to a page

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None currently – there is still an issue with the page-related tests which I need to fix in a follow-up.

## ❓ How to test this?

1. To test handling of missing right entities the easiest thing to do is to use the plugin to create a draft (or create one in the API and link to it) and temporarily stop filtering out the notification entities.
2. Other mention types can be tested by just trying to insert them and checking they appear correctly across a refresh.

### DIFF
--- a/libs/@local/hash-isomorphic-utils/src/text.ts
+++ b/libs/@local/hash-isomorphic-utils/src/text.ts
@@ -71,7 +71,7 @@ export const textBlockNodeToTextTokens = (node: ComponentNode): TextToken[] => {
           mentionType: child.attrs.mentionType,
           entityId: child.attrs.entityId,
           propertyTypeBaseUrl: child.attrs.propertyTypeBaseUrl,
-          linkEntityTypeBaseUrl: child.attrs.linkEntityTypeId,
+          linkEntityTypeBaseUrl: child.attrs.linkEntityTypeBaseUrl,
         });
         break;
       }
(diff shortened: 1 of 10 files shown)


libs/@local/hash-isomorphic-utils/src/text.ts
```diff
@@ -71,7 +71,7 @@ export const textBlockNodeToTextTokens = (node: ComponentNode): TextToken[] => {
           mentionType: child.attrs.mentionType,
           entityId: child.attrs.entityId,
           propertyTypeBaseUrl: child.attrs.propertyTypeBaseUrl,
-          linkEntityTypeBaseUrl: child.attrs.linkEntityTypeId,
+          linkEntityTypeBaseUrl: child.attrs.linkEntityTypeBaseUrl,
```
benwerner01, line 74: Thanks for catching this

benwerner01 (approved):
Thanks Ciaran, this is working for me locally 👍

PR #3838, opened 2024-01-08
H-1730: Add "load entities/just start typing" buttons to table block + other table block polish

benwerner01:

This PR adds a welcome modal to the table block, which allows the user to choose between using a query or manually populating the table. It also makes some other improvements to the block.

## 🔗 Related links

- H-1730

## 🔍 What does this change?

- upgrades the `@blockprotocol/graph` package, so that the new `requestLinkedData` method becomes available in blocks
- adds a welcome modal to the table block which allows the user to pick between populating the table with a query and populating it manually
- adjust the background/border colours of the glide-grid component in the table block to be more consistent with Figma
- removes the `base.module.scss` file and some other SCSS styling, moving towards making use of MUI theme styling to be more consistent with other blocks/the HASH App
- modifies the freeform table to be initialized with 5 rows, and enforces a minimum of 1 row

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- There is no way to go between a free-form table and a table that is populated with a query from within the table block. Adding a linked query to a free-form table which has been modified will not display the query results. We want to unify the free-form table and the entities tables, so that there is no switching between them required.

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

1. Checkout the branch locally, and run the HASH App and table block in dev mode
2. Load the table block, and click "Load existing entities" to open the query editor and load entities into the table
3. Load a new table block, and click "just start typing" and try out populating the table manually

## 📹 Demo

Populating a table with a query:

https://github.com/hashintel/hash/assets/42802102/ad283914-fb56-49b9-874e-d506d5896e06

Manually populating a table:

https://github.com/hashintel/hash/assets/42802102/529bf3b6-9fed-4cc9-9658-c954f3c9f336


nonparibus:
Thanks. As discussed, still lots to do, but getting there. I'll drop follow-ups in a ticket.

nonparibus:
@benwerner01 Can we make the icons go `Blue 70` onhover as well, though, in addition to the titles?

benwerner01:
> Can we make the icons go Blue 70 onhover as well, though, in addition to the titles?

@nonparibus yes, addressed in https://github.com/hashintel/hash/pull/3838/commits/cbf20a0e6f43c46df6cfecdd6dfd9b47e006b904


https://github.com/hashintel/hash/assets/42802102/09476154-32ee-44e6-a587-0f410142a294

blocks/table/src/components/table/table.tsx
```diff
@@ -153,41 +170,104 @@ export const Table = ({ blockEntity, updateEntity, readonly }: TableProps) => {
...
+  const handleCellsEdited = useCallback<
+    NonNullable<DataEditorProps["onCellsEdited"]>
+  >(
+    (newValues) => {
+      const mergedLocalRows =
+        numberOfRows > localRows.length
+          ? [
+              ...localRows,
+              ...Array<Object>(numberOfRows - localRows.length).fill(
+                {} as Object,
```
CiaranMn, line 189: This seems to pass linting and saves the assertion.

The `Object` that `localRows` is an array of is actually a type named `Object` exported from `@blockprotocol/graph` (I think this is undesirable as it clashes with a JS primitive, and created H-1170 to rename it). It is an alias for `{}`.

I think the type would be better as a JsonObject, e.g. something with string keys and unknown json-permitted values. That requires changing the BP package though and would have knock-on consequences.
```suggestion
              ...Array<{}>(numberOfRows - localRows.length).fill({}),
```

blocks/table/src/components/table/table.tsx
```diff
@@ -153,41 +170,104 @@ export const Table = ({ blockEntity, updateEntity, readonly }: TableProps) => {
...
+        return {
+          kind: GridCellKind.Text,
+          displayData: "",
+          data: "",
+          allowOverlay: false,
+        };
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- todo fix this
+      const value = ((localRows[rowIndex] as any)?.[key] ?? "") as JsonValue;
```
CiaranMn, line 242: One alternative, although `value` ends up inferred as `""` (because no keys exist on `{}`) 🤷 
```suggestion
      const row = localRows[rowIndex];
      const value = row && key in row ? row[key as keyof typeof row] : "";
```

blocks/table/src/app.tsx
```diff
@@ -61,19 +62,58 @@ export const App: BlockComponent<BlockEntity> = ({
...
+    return (
+      (!tableLocalColumn || tableLocalColumn.length === 0) &&
+      (!tableLocalRow || tableLocalRow.length === 0)
+    );
+  }, [blockEntity]);
+
+  const [isUsingLocalTable, setIsUsingLocalTable] =
+    useState(!isLocalTableEmpty);
+
+  if (!isUsingLocalTable && !isLocalTableEmpty) {
```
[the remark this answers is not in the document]
benwerner01: I was trying to account for the table block being updated elsewhere, or receiving updated properties after being mounted. Have left a comment clarifying in https://github.com/hashintel/hash/pull/3838/commits/51d2a765c95655c3c8ffad7843d4a3557504c258

blocks/table/src/components/table/table.tsx
benwerner01: Thanks for the diff, have applied it in https://github.com/hashintel/hash/pull/3838/commits/dc8dbd2bef2f5f1d45633e272cfd5af446c6bcf2. I don't like that we're using `Object` here rather than properly defining nested properties, but didn't want to tackle this now as we likely will be changing the table block's definition further once we've agreed on what changes we're making in next week's sprint. Probably will get rid of this then.

blocks/table/src/app.tsx
```diff
@@ -78,6 +78,11 @@ export const App: BlockComponent<BlockEntity> = ({
   const [isUsingLocalTable, setIsUsingLocalTable] =
     useState(!isLocalTableEmpty);
 
+  /**
+   * The table block entity may have been populated with data elsewhere
+   * after the component is first mounted, in which we know the table
```
CiaranMn, line 83:
```suggestion
   * after the component is first mounted, in which case we know the table
```

CiaranMn (approved):

PR #3850, opened 2024-01-09
H-1722: Support inferring greater numbers of entities from pages

CiaranMn:

One of the limiting factors of the AI entity inference process is the completion token limit – currently 4,096 tokens for all models we use
- This means that we cannot ask the AI to generate many (e.g. 50) entities at once, because it doesn't have space to. 
- With the function call/tools feature we are using, the model will provide incomplete JSON if it hits the completion token limit. It doesn't seem possible to get it to just 'finish off the previous JSON you started', probably due to how it's implemented internally (we might be able to do this if we did our own version of 'provide json' via messages).

The improvement this PR introduces is to split the inference process into two steps:
1. Ask the model to generate a list of entities it can infer from the page, without any properties – just a brief summary which is sufficient for it to uniquely identify the entity in the text.
2. Ask it to provide details of 10 entities from this list at a time.

## 🔍 What does this change?

- Introduce a new `infer-entity-summaries.ts` file which handles the first step.
- Extract some common logic from the original inference function to share with this new file.
- Increases the timeout of the inference workflow to 1 hour to give more overhead for many iterations (a test page involving 60 entities took around 16 minutes)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

There are further improvements that could be made here (H-1847):
1. We don't limit the first step, generating entity summaries. These too may fall foul of the completion token limit somewhere above 100 entities. We could update this to ask the model to suggest 100 entities at a time, passing back the previously inferred 100 each time so it doesn't repeat any

2. We apply a blanket '10 entities per iteration' rule to the second step, providing full entity details.
  - For smaller entities with only a few short properties this is over-padding (with completion coming in it around 2k tokens).
  - For larger entities with many properties it may run into the completion token limit
  - We could save time/tokens for smaller entities, and make it less likely we hit the completion token limit for larger entities, by dynamically adjusting the page size based on the types of entities being requested, how many properties they have and the type of value.

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1. Rebuild the backend, or stop the existing AI worker in your Docker and run it locally 
2. Try inferring from a page with many entities, watch the AI worker logs (or just come back to it in a while and read the log output, which should be in `infer-entities/logs`


benwerner01 (approved):
Nice work, this looks great and is working for me locally 👍

PR #3854, opened 2024-01-10
H-1833: Add timezone to browser-input times, add custom date/time icons

CiaranMn:

We use native browser date/time pickers, which do not add the user's timezone ([rationale](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#setting_timezones)). Our first date/time datatypes _do_ require a timezone, because they follow RFC 3339. Therefore, the output of browser date/time pickers will not validate against our custom data types. Nor will our custom data type values be valid values for the browser inputs.

This PR fixes that by:
1. Converting the output of the browser inputs to add the user's offset from UTC (according to their device)
2. Converting the value to _remove_ the timezone when passing back into the browser inputs

It also:
3. Adds custom icons for date/time data types
4. Adds a `Time` data type
5. Updates the description of `DateTime` and `Date` data types to refer to RFC 3339

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1. Add some local entity types that use date/datetime/time property types and check they work.
2. If changing the system time you will need to reload the browser for the new offset to take effect

## 📹 Demo

At +10 hours from UTC

https://github.com/hashintel/hash/assets/37743469/8110c8da-7056-4d71-b22d-ec86c819fcad


libs/@hashintel/design-system/src/fa-icons/fa-calendar-clock-regular.tsx
```diff
@@ -0,0 +1,9 @@
+import { IconDefinition } from "@fortawesome/free-solid-svg-icons";
```
benwerner01, line 1: At some point it'd be nice to be more consistent with how we add FA icons, appreciate this is approach makes sense here because of the implementation in `data-types-options-context.tsx`

benwerner01 (approved):
Thanks Ciaran, looks good and is working for me locally 👍

libs/@hashintel/design-system/src/fa-icons/fa-calendar-clock-regular.tsx
CiaranMn: Definitely agree

benwerner01 (dismissed):

benwerner01 (approved):

PR #3855, opened 2024-01-10
H-1840: Only update embedding if the entity version is newer or equal

TimDiekmann:

When an entity is updated twice at the same time it's not possible to determine which entity edition will be the "latest" entity (`transaction_time && now() AND decision_time && now()`). This is a general problem with temporal versioning as one update will end up as the "latest" entity and one as an entity with an empty temporal range. It's not possible to determine which will do it. (Note: an empty range cannot be queried from the graph as it does not have any overlap with anything)

Entity embeddings are created from an external service and it's not deterministic, which entity embedding will be finished first. Both entity updates will trigger an update-embedding-workflow. The workflow eventually will create a request to the graph to update the embedding and includes the start times of temporal axes (only the start time is immutable for an entity, the end time is only a current view). The graph will not be able to determine which of the two embeddings which were created belonging to the latest entity as both has the same start time.

I  see two possible solutions to this, neither is good or easy (it's conceptually a hard problem to solve):

1. Instead of storing the timestamps when the entity was created, the full entity is passed alongside the embedding creation (or at least the changeset of the update, but this is even more complicated). When inserting the embedding it's checked if the "latest" entity matches the entity which was used for the embedding. This is a quite slow operation and it's possible that an entity embedding will not be updated for a while if the entity was updated shortly after the embedding generation was kicked of as the entity used for the embedding is already outdated again. To mitigate this it might be possible to still store the timestamps on the embedding and only do a full comparison if in doubt.

2. An external service such as collab is responsible to collect entity updates and make sure that only one update for an entity happens at the same time.

Collab is a great concept but it's 

- hard to implement (especially if collab should be scalable. It's easily possible to run multiple graphs at the same time on the same database and to enable this on collab it's even more complicated.) and 

- requires all services to go through that service (you would never be able to directly communicate with the graph without collab in between) 

So I see the first option as more viable. 

The current update logic introduced in this PR stores both, the decision start time and transaction start time, next to the embedding and if an embedding for a given entity ID is generated and the timestamps of that entity edition which was used to generate the embeddings are ***equal to or later than*** the timestamps of the current embedding it will be updated.

## 🔍 What does this change?

- Alter database to store embedding update times
- Use the time to determine if an embedding is inserted into the database

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-ts/src/activities/graph.ts
```diff
@@ -35,11 +36,15 @@ export const createGraphActivities = ({
...
     };
     embeddings: EntityEmbedding[];
+    updatedAtTransactionTime: Timestamp;
+    updatedAtDecisionTime: Timestamp;
   }): Promise<void> {
     await graphApiClient
       .updateEntityEmbeddings(params.authentication.actorId, {
         embeddings: params.embeddings,
-        reset: true,
+        reset: false,
```
CiaranMn, line 45: In what circumstance would I want to pass `reset: true`? Is it something we actually want users to be able to do?

TimDiekmann: Actually, that was by accident (for some local testing) and this should be `reset: true`. If you would remove a property from an entity the embedding would not be removed. As we neither have partial updates, nor we detect changes to the previous entity, all embeddings for an entity has to be generated anyway.

```suggestion
        reset: true,
```

CiaranMn: Got it, thank you. In what case would I pass `result: false` then?

TimDiekmann: When it's possible to detect that no property was removed, it's not needed to reset the properties. Possibly we can be even smarter when we have the ability to detect changes. We can also remove it for the time being.

CiaranMn: It's fine to leave, since we will only be calling this in one place.

CiaranMn (approved):

PR #3861, opened 2024-01-11
H-1866: Create a Temporal client in Rust

TimDiekmann:
## 🌟 What is the purpose of this PR?

To invoke workflows from Rust it's required to connect to Temporal. Temporal does not offer a simple OpenAPI interface but their SDK is written in Rust so it's possible to use the temporal-client crate for this purpose.

## 🚫 Blocked by

- #3855

## 🔍 What does this change?

- Create a library `temporal-client` which is used to connect to the Temporal.io server
- Add parameters to the Graph service and configure it so it can connect to Temporal.io if the host is specified

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] affected the execution graph, and the dependencies were adjusted

## 🐾 Next steps

- Run the `update_entity_embeddings` workflow when an entity is created or updated

PR #3863, opened 2024-01-11
H-1871: Extend workflow to support multiple entities or a filter

TimDiekmann:

For initial seeding and if an embedding creation failed we need a way to manually trigger entity embedding creation. Ideally, this is done via a structural query.

## 🚫 Blocked by

- #3855
- #3861 
- #3862

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

When using a filter to return all entities, this is still filtered by the actor's permission to see the entities so it's not possible to update all embeddings in one go.

## ❓ How to test this?

With a few simple arguments it's possible to update all entities without embeddings:
```json
[{"authentication":{"actorId":"40bbc079-0d81-4def-bd5d-5fae94edca47"},"filter":{"equal":[{"path":["embedding"]},null]}}]
```

A full call would look like this:
```bash
temporal workflow execute --no-pager --task-queue ai --type updateEntityEmbeddings --namespace HASH -i "$(jq -n '{filter: { "all": [] }, authentication: { actorId: "40bbc079-0d81-4def-bd5d-5fae94edca47"} }')"
```

### DIFF
--- a/apps/hash-ai-worker-ts/src/workflows.ts
+++ b/apps/hash-ai-worker-ts/src/workflows.ts
@@ -1,8 +1,4 @@
-import {
-  EntityStructuralQuery,
-  Filter,
-  FilterExpression,
-} from "@local/hash-graph-client";
+import type { Filter } from "@local/hash-graph-client";
 import type { InferEntitiesCallerParams } from "@local/hash-isomorphic-utils/ai-inference-types";
 import type { AccountId, Entity } from "@local/hash-subgraph";
 import { proxyActivities } from "@temporalio/workflow";
--- a/apps/hash-ai-worker-ts/src/workflows.ts
+++ b/apps/hash-ai-worker-ts/src/workflows.ts
@@ -1,3 +1,8 @@
+import {
+  EntityStructuralQuery,
+  Filter,
+  FilterExpression,
+} from "@local/hash-graph-client";
 import type { InferEntitiesCallerParams } from "@local/hash-isomorphic-utils/ai-inference-types";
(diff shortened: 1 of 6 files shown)


apps/hash-ai-worker-ts/src/workflows.ts
```diff
@@ -24,69 +25,124 @@ const graphActivities = proxyActivities<
...
+
+  const usage: CreateEmbeddingResponse.Usage = {
+    prompt_tokens: 0,
+    total_tokens: 0,
+  };
+
+  for (const entity of entities) {
+    // TODO: The subgraph library does not have the required methods to do this client side so for simplicity we're
+    //       just making another request here. We should add the required methods to the library and do this client
+    //       side.
```
CiaranMn, line 95: I think you could do it with these two methods from the subgraph library:
```ts
const propertyTypeIds = getPropertyTypesReferencedByEntityType(
  subgraph, 
  entity.metadata.entityTypeId
);

const propertyTypes = propertyTypeIds.map(
  propertyTypeId => getPropertyTypeById(subgraph, propertyTypeId)
);
```

TimDiekmann: In theory it should work, yes, but I don't think this will take inheritance in account

TimDiekmann: (To be clear: I did not know about the first function, but I know it was implemented before we implemented inheritance)

CiaranMn (approved):

PR #3864, opened 2024-01-11
H-1651: Persist user plugin settings in database

CiaranMn:

The user's settings in the browser plugin are currently stored in local storage. If the plugin is loaded without a user, local storage is wiped (for privacy). This means that a user's settings are lost if they log out, switch browsers, clear their browser, switch devices, or if the API is temporarily unreachable.

This PR persists a user's plugin settings as an entity, so that they are preserved and reloaded on all browsers/devices/across log-outs.

It also refactors the order in which various system entities are created in the migration files, so that it's possible in migration scripts to use e.g. machine web actors to update existing entities to new type versions. Previously this was not possible because various system entities were created after all migrations ran, meaning that functions that depended on those entities couldn't be used in the migration.

This PR:

- [ ] modifies the plugin, which will need publishing once it's approved

The changes in this PR:

- [x] adds developer-facing docs to the migrations folder

The changes in this PR:

## 🛡 What tests cover this?

- None (I will be adding tests for the plugin soon)

## ❓ How to test this?

1. Load the plugin and set some configuration for manual / automatic inference, write a quick note (without saving)
2. Check that these settings are preserved, e.g.
   - kill the API and open the popup so that local storage is wiped, then restart the API and check the settings reload
   - open the plugin in a different browser or on a different device


apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/util.ts
```diff
@@ -1128,3 +1130,136 @@ export const getEntitiesByType: ImpureGraphFunction<
...
+              { parameter: systemEntityTypes.organization.entityTypeBaseUrl },
+            ],
+          },
+        ],
+      },
+      graphResolveDepths: zeroedGraphResolveDepths,
+      includeDrafts: true,
+      temporalAxes: currentTimeInstantTemporalAxes,
+    },
+  }).then((subgraph) => getRoots(subgraph));
```
benwerner01, line 1172: Why are these two different sequential queries? We could either call these simultaneously, or just call a single query with an `any` condition getting all entities that are users or orgs, and then filter the result for the users and orgs.

apps/plugin-browser/src/shared/storage.ts
```diff
@@ -124,6 +155,64 @@ export const setInLocalStorage = async (
...
+    }
+
+    const currentAutomaticConfig = await getFromLocalStorage(
+      "automaticInferenceConfig",
+    );
+    const currentManualConfig = await getFromLocalStorage(
+      "manualInferenceConfig",
+    );
+    const currentPopupTab = await getFromLocalStorage("popupTab");
+    const currentDraftNote = await getFromLocalStorage("draftQuickNote");
```
benwerner01, line 176: I'm wondering whether this could be problematic because the async calls are happening outside of the debounced function. If I'm understanding right, it would be possible for a user to make changes in quick succession, and for an older change to result in the most recent call to the debounce function, which would make it possible for stale data to be persisted in the db. Granted this is probably quite unlikely, but if it's not a time consuming change maybe it's worth adding these async calls and the creation of the `newProperties` object to the body of `debouncedEntityUpdate`?

CiaranMn: I think whether or not this is technically possible depends on how `browser.storage.get` works, and whether it's possible for the calls to return in a different order than they entered the call stack. I've moved the calls inside the function inhttps://github.com/hashintel/hash/pull/3864/commits/53d5a29660ce76311b08bbd70b33cb7faf8ad929 in any case as it doesn't hurt and will avoid the problem you describe, if it's possible

apps/hash-api/src/graph/ensure-system-graph-is-initialized/system-webs-and-entities.ts
```diff
@@ -137,20 +129,94 @@ export const getOrCreateOwningAccountGroupId = async (
   };
 };
 
+export const createSystemWebEntities = async ({
```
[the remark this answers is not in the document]
CiaranMn: Renamed in https://github.com/hashintel/hash/pull/3864/commits/dc8d62788ac237dd29ea82183216ded041614664

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/util.ts
CiaranMn: I didn't really think about it much – I don't think we need to optimize the migration scripts until running them becomes a big part of start-up time (there are lots of property types we create sequentially where we could in parallel, for example). I've updated these to run in parallel in https://github.com/hashintel/hash/pull/3864/commits/d7121856688a3b25ca25f7646f3ff2c25ade48cd

benwerner01 (approved):

benwerner01 (approved):

PR #3891, opened 2024-01-17
H-1908: Return archived types when JSON is requested

CiaranMn:

When a type's URL is requested outside a browser contet (when there's no `text/html` in the `accept` header), we return JSON instead.

The logic for doing so was excluding archived types. Type archival is meant to remove types from being discovered, but not from being used if they are already being relied on. 

This PR fixes the 'return types as JSON' logic to return an archived type if its URL is requested.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1. Create a property type and archive it
2. `curl [its-id]` and check the JSON is returned
3. in `main` this will not work

### DIFF
--- a/apps/hash-api/package.json
+++ b/apps/hash-api/package.json
@@ -6,7 +6,7 @@
   "license": "AGPL-3.0",
   "scripts": {
     "codegen": "rimraf './src/**/*.gen.*'; graphql-codegen --config codegen.config.ts",
-    "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=2048 tsx watch ./src/index.ts",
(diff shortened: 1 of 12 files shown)


apps/hash-api/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "AGPL-3.0",
   "scripts": {
     "codegen": "rimraf './src/**/*.gen.*'; graphql-codegen --config codegen.config.ts",
-    "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=2048 tsx watch ./src/index.ts",
+    "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=2048 tsx watch --clear-screen=false ./src/index.ts",
```
TimDiekmann, line 9: Thank you ❤️

PR #3897, opened 2024-01-17
H-1894, H-1895: Associate Node and Rust Sentry spans and improve Context information

TimDiekmann:

To make Sentry more useful (or useful at all) the spans between the projects should be connected. Also, the current information stored on Rust events could have been improved. The most important fix this PR is providing that (hopefully) all GraphQL requests get a unique trace ID so they are actually distinguishable in Sentry.

## 🔍 What does this change?

- Rewrite Apollo server setup to use new Sentry API
- Enable integrations for Sentry and Express and use the tracing plugin to create a sentry transaction per GraphQL request
- Add a `capture_report` function to capture an `error_stack::Report`. Currently, this only captures it as a message but also tries to read the surrounding lines in the file.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

I had several issues around the trace IDs. Especially some traces bleeds into other traces when multiple requests were issued. The current approach and combination of plugins/setup seems to work pretty stable but this is something we need to improve. For more information see H-1916.

## 📹 Demo

image
image
image
image

### DIFF
--- a/apps/hash-api/src/graphql/create-apollo-server.ts
+++ b/apps/hash-api/src/graphql/create-apollo-server.ts
@@ -48,6 +48,7 @@ export const createApolloServer = ({
     resolvers,
     inheritResolversFromInterfaces: true,
   });
+
   const getDataSources = () => {
     const sources: GraphQLContext["dataSources"] = {
       graphApi,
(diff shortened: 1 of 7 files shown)


apps/hash-api/src/graphql/create-apollo-server.ts
```diff
@@ -79,55 +80,55 @@ export const createApolloServer = ({
...
+
+          const startTimestamp = performance.now();
 
           return {
-            async didEncounterErrors(errorContext) {
-              const user: Express.Request["user"] = errorContext.context.user;
+            didResolveOperation: async (didResolveOperationCtx) => {
+              const operationName = didResolveOperationCtx.operationName;
+              if (operationName) {
+                // transaction.setName(operationName, "route");
```
CiaranMn, line 95: Is this mean to be here?

apps/hash-api/src/graphql/create-apollo-server.ts
```diff
@@ -171,6 +172,8 @@ export const createApolloServer = ({
                   );
                 }
               }
+
+              // span?.finish();
```
CiaranMn, line 176: Is this meant to be here?

apps/hash-api/src/graphql/create-apollo-server.ts
TimDiekmann: Thanks, that was an old fragment which is not required anymore: bae93498d9c2e458cee6acfe80d71ad0e7415a7b

apps/hash-api/src/graphql/create-apollo-server.ts
TimDiekmann: Thanks, that was an old fragment which is not required anymore: bae93498d9c2e458cee6acfe80d71ad0e7415a7b

CiaranMn (approved):

PR #3922, opened 2024-01-22
H-1626: Enable cancelling inference jobs from the browser plugin

CiaranMn:

Adds a button to inference jobs on the browser plugin allowing the user to cancel them early, receiving whatever entity creations/updates (if any) had been processed up to that point.

## 🚫 Blocked by

- [x] ~I haven't been able to throw a `CancellationFailure` from the workflow _and_ get partial results back to the client. I have asked on the Temporal Slack as to how best to deal with getting partial results back from a cancelled workflow (see comments in diff).~ I have [updated the PR](https://github.com/hashintel/hash/pull/3922/commits/f94e7109252d33e42fa4ff5699760783bdc7cdc6) to allow for querying for partial results from the client. I suspect the `details` not appearing in the error thrown to the client is a bug, and ideally we would be able to capture them so they're attached to the Temporal job, but this isn't important until we are using the Temporal event history for something.

## 🔍 What does this change?

- Adds a button to pending inference requests allowing them to be cancelled (see demo)
  - jobs which are cancelled are marked by red 'Cancelled' text in the inference details
  - they otherwise appear as completed jobs, i.e. a gray dash if they resulted in no changes, and a green tick if they resulted in some changes. We could alternatively use a different icon for 'Cancelled', but it would mean the user would not know if it resulted in any graph changes or not.
- Handle cancellations in the AI inference websocket, and the Temporal activity/workflow
  - a lot of the diff is moving the existing 'handle AI inference request' into its own file, and adding a new file for handling a cancellation request 

This PR:

- [x] modifies the plugin, which I'll publish after this PR is merged as I have no other immediately pending tasks updating it

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1. Stop the Temporal worker in Docker and run it locally
2. `yarn dev` in the plugin folder
3. Try starting and cancelling a job, both immediately, and then once the AI worker has had a chance to do something.

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/c72e4e31-f988-4231-b6a0-12c898aca7e8

### DIFF
--- a/apps/hash-ai-worker-ts/package.json
+++ b/apps/hash-ai-worker-ts/package.json
@@ -41,6 +41,7 @@
     "@local/hash-subgraph": "0.0.0-private",
     "@local/status": "0.0.0-private",
     "@temporalio/activity": "1.8.1",
+    "@temporalio/common": "1.8.1",
     "@temporalio/worker": "1.8.1",
     "@temporalio/workflow": "1.8.1",
     "dedent": "0.7.0",
--- a/apps/hash-ai-worker-ts/src/activities.ts
+++ b/apps/hash-ai-worker-ts/src/activities.ts
@@ -21,29 +21,25 @@ export const createAiActivities = ({
   graphApiClient,
 }: {
   graphApiClient: GraphApi;
-}) => {
-  return {
-    async inferEntitiesActivity(
-      params: InferEntitiesCallerParams,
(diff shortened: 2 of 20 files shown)


benwerner01 (approved):
Thanks Ciaran, this looks good and is working for me 👍

PR #3923, opened 2024-01-22
H-1564: Add "Create new" button to `/entities` page(s)

benwerner01:

This PR adds a "Create new entity" button to the `/entities` page, which allows you to create a new:
- entity of the type currently viewed on the `/entities` page
- entity of an unknown type if there is no type currently specified on the `/entities` page
- page if the type being viewed is the "Document" type

## 🔗 Related links

- H-1564
## 🔍 What does this change?

- fixes the "Create new page" button on the view all pages page, by updating the base URL in the redirect from the outdated to the new slug

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

1. Go to the `/entities` page and create a new entity of an unknown type
2. Go to the `/entities` page for a specific type (via a sidebar), and create an entity of that type

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/6ebd3b24-3cf5-43b0-b2de-70db1b61b439

### DIFF
--- a/apps/hash-frontend/package.json
+++ b/apps/hash-frontend/package.json
@@ -73,6 +73,7 @@
     "next-seo": "6.1.0",
     "nextjs-progressbar": "0.0.16",
     "notistack": "2.0.5",
+    "pluralize": "8.0.0",
     "prismjs": "1.29.0",
     "prosemirror-collab": "1.3.0",
     "prosemirror-commands": "1.3.1",
--- a/apps/hash-frontend/package.json
+++ b/apps/hash-frontend/package.json
@@ -73,7 +73,6 @@
     "next-seo": "6.1.0",
     "nextjs-progressbar": "0.0.16",
     "notistack": "2.0.5",
-    "pluralize": "8.0.0",
     "prismjs": "1.29.0",
     "prosemirror-collab": "1.3.0",
     "prosemirror-commands": "1.3.1",
--- a/apps/hash-frontend/src/pages/entities.page.tsx
+++ b/apps/hash-frontend/src/pages/entities.page.tsx
@@ -21,6 +21,8 @@ import { useHashInstance } from "../components/hooks/use-hash-instance";
 import { EntityTypeEntitiesContext } from "../shared/entity-type-entities-context";
 import { useEntityTypeEntitiesContextValue } from "../shared/entity-type-entities-context/use-entity-type-entities-context-value";
 import { useLatestEntityTypesOptional } from "../shared/entity-types-context/hooks";
+import { useEntityTypesContextRequired } from "../shared/entity-types-context/hooks/use-entity-types-context-required";
+import { generateLinkParameters } from "../shared/generate-link-parameters";
 import { AsteriskLightIcon } from "../shared/icons/asterisk-light-icon";
 import { CanvasNewIcon } from "../shared/icons/canvas-new-icon";
 import { FileCirclePlusRegularIcon } from "../shared/icons/file-circle-plus-regular-icon";
@@ -63,6 +65,8 @@ const EntitiesPage: NextPageWithLayout = () => {
   const { activeWorkspaceOwnedById, activeWorkspace } = useActiveWorkspace();
 
   const { hashInstance } = useHashInstance();
+
+  const { isSpecialEntityTypeLookup } = useEntityTypesContextRequired();
 
   const { entityTypeId, entityTypeBaseUrl } = useMemo(() => {
     if (router.isReady) {
@@ -109,6 +113,14 @@ const EntitiesPage: NextPageWithLayout = () => {
     entityTypeId,
   });
 
+  const isFileEntityType = useMemo(
+    () =>
+      entityType
+        ? isSpecialEntityTypeLookup?.[entityType.schema.$id]?.isFile
+        : false,
+    [isSpecialEntityTypeLookup, entityType],
+  );
+
   const isViewAllPagesPage =
     entityType && entityType.schema.$id === systemEntityTypes.page.entityTypeId;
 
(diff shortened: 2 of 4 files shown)


nonparibus:
> page if the type being viewed is the "Document" type

@benwerner01 You mean `Document` if viewing `Document`, and `Canvas` if viewing `Canvas`? And shows both `Document` and `Canvas` buttons if viewing `Page` (the parent type from which they inherit?

benwerner01:
@nonparibus I've made the changes we discussed in standup, now there are special cased "Create new document" and "Create new canvas" buttons when viewing all pages:

https://github.com/hashintel/hash/assets/42802102/14cbadd7-cb56-4697-bd79-83367b11421f

nonparibus:
@benwerner01 Upon reflection, if not too much work, can we also please set up `/files` so a user's files can be easily viewed at this path, and change the button there to read "Add new file(s)" _instead of_ "Create new file" (as I think it would otherwise default to)? This link should point to `https://app.hash.ai/@hash/types/entity-type/file/v/1?tab=upload` - the same location as the "Add > Upload file" link in the header nav does. Thanks.

benwerner01:
@nonparibus I've addressed this in https://github.com/hashintel/hash/pull/3923/commits/0f94ab63d7936d1cdaa67a0de0b9c59f08bf2cb8, so for any type that is an sub-type of `File` (or the `File` type itself) it redirects to the upload tab of the entity type editor instead of the create entity page:


https://github.com/hashintel/hash/assets/42802102/c540b107-adcc-439a-b675-02b1e73ec55f

benwerner01:
@nonparibus I've addressed this in https://github.com/hashintel/hash/pull/3923/commits/0f94ab63d7936d1cdaa67a0de0b9c59f08bf2cb8, so for any type that is an sub-type of `File` (or the `File` type itself) it redirects to the upload tab of the entity type editor instead of the create entity page:


https://github.com/hashintel/hash/assets/42802102/c540b107-adcc-439a-b675-02b1e73ec55f

benwerner01:
@nonparibus I've addressed this in https://github.com/hashintel/hash/pull/3923/commits/0f94ab63d7936d1cdaa67a0de0b9c59f08bf2cb8, so for any type that is an sub-type of `File` (or the `File` type itself) it redirects to the upload tab of the entity type editor instead of the create entity page:


https://github.com/hashintel/hash/assets/42802102/c540b107-adcc-439a-b675-02b1e73ec55f

apps/hash-frontend/src/pages/entities.page.tsx
```diff
@@ -157,30 +224,41 @@ const EntitiesPage: NextPageWithLayout = () => {
...
-                  [`.${buttonClasses.endIcon}`]: {
-                    color: ({ palette }) => palette.blue[70],
-                  },
-                }}
-              >
-                Create a page
-              </Button>
-            ) : null}
+            <Fade in={displayCreateEntityButton}>
+              {isViewAllPagesPage ? (
```
CiaranMn, line 228: If someone visits this page filtered to `Document` or `Canvas` they will get a create button which takes them to the regular entity editor – I think we might be better off making `CreateButtons` into a proper component which then renders different stuff depending on the entityTypeId filtered to, e.g.

1. if `Document`, show document button
2. If `Canvas`, show canvas button
3. if `Page`, show both
4. show default, with special copy for files

We can then hide all the special casing inside the `CreateButton` component (including checking if pages are enabled), since it isn't needed outside of there

benwerner01: That's a fair point, I've addressed this in https://github.com/hashintel/hash/pull/3923/commits/32e86e7d9a0250b2253daa78fe86084cdb9ba323. I also spotted that the `Canvas` page was being pluralized weirdly to `Canvass`, so added the `pluralize` package in https://github.com/hashintel/hash/pull/3923/commits/f567d5bc21dfc5388dddaea18ae0126ae690880c

apps/hash-frontend/package.json
```diff
@@ -73,6 +73,7 @@
     "next-seo": "6.1.0",
     "nextjs-progressbar": "0.0.16",
     "notistack": "2.0.5",
+    "pluralize": "8.0.0",
```
CiaranMn, line 76: Can you import `pluralize` from `isomorphic-utils` instead? We add some rules to fix some incorrect pluralizations it generates:
```ts
import { pluralize } from "@local/hash-isomorphic-utils/pluralize";
```

apps/hash-frontend/src/pages/entities.page.tsx
```diff
@@ -88,16 +206,24 @@ const EntitiesPage: NextPageWithLayout = () => {
...
 
   const { entities, loading } = entityTypeEntitiesValue;
 
+  const displayCreateEntityButton = useMemo(() => {
+    if (isViewAllPagesPage) {
+      return hashInstance?.properties.pagesAreEnabled;
+    }
+
+    return true;
+  }, [isViewAllPagesPage, hashInstance]);
```
CiaranMn, line 225: Sorry for not spotting this sooner, but we don't want to hide this button any more as far as I know – even if pages are disabled, we want the button for all the other entity types, including file. `pagesAreEnabled` should only affect the display of the page-related buttons.

benwerner01: I'm not sure I understand, the reason I kept the logic in the parent component rather than just return `null` in the `CreateButtons` component is so that the buttons fade in if the pages page is visited and `hashInstance` is `undefined` on the first render, and on the second render `pagesAreEnabled` is evaluated to be `true`.

> even if pages are disabled, we want the button for all the other entity types, including file

Unless I'm missing something, I think that should be the case with the logic above.

CiaranMn: Sorry, I misread the logic, ignore this comment

CiaranMn (approved):

PR #3927, opened 2024-01-23
H-1937: Overhaul Read trait to allow returning Cursor

TimDiekmann:

The current `Read` trait always returns `Self::Record`. This has the downsides that it's not possible to read data which is stored alongside the return value such as the distance or the cursor. Initially, this is required for root sorting but we will also require to dynamically choose the returned values for query methods, e.g. only returning the type of an entity instead of all information.

Generally, this will be required or at least very useful for:

- Root sorting        
- Selective return values
- Provenance

## 🔍 What does this change?

- Rewrite the `Read` trait:
    - Split into `Read` and `ReadPaginated` (as otherwise the generics would be empty)
    - Define the conversion of cursor and records on the records (and cursors) itself, not on the trait
    - This implies, that the `Read` trait is now only defined on two structs: `PostgresStore` and `FetchingStore`
    - This also imply a rather big diff as things were moved around. The underlying logic did not change much (besides the `cursor` is now exposed)
- Prepare interfaces to allow root sorting (H-1440)
- First step to use the builder pattern instead of passing all arguments (H-1466)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The `Read` trait is split into two parts, this is suboptimal and may be revisited at some later point

## 🐾 Next steps

Implement root sorting (H-1440)

## 🛡 What tests cover this?

- This is an internal refactoring and every test which is using the Graph should go through these methods
- New tests are expected to be added later this week or early next week

### DIFF
--- a/.env
+++ b/.env
@@ -49,7 +49,7 @@ HASH_TEMPORAL_VISIBILITY_PG_DATABASE=temporal_visibility
 HASH_GRAPH_PG_USER=graph
 HASH_GRAPH_PG_PASSWORD=graph
 HASH_GRAPH_PG_DATABASE=graph
-HASH_GRAPH_LOG_LEVEL=trace,h2=info,tokio_util=debug,tower=info,tonic=debug,hyper=info,tokio_postgres=debug,rustls=info,tarpc=warn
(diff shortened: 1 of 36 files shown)


CiaranMn (approved):
Excellent work @TimDiekmann! 🥳

PR #3929, opened 2024-01-25
H-1945: Extract and store text from uploaded DOC(X), PPT(X), & PDF files, upon upload

benwerner01:

This PR adds the ability to upload DOCX, PPTX and PDF files to HASH, where their text contents are parsed and added as a `textualContent` property to the corresponding file entity.

It does this by triggering a `parseTextFromFile` temporal workflow in the "after update entity hook".

## 🔗 Related links

- H-1945

## 🔍 What does this change?

- fixes a bug where previously when uploading a non-image file, the type of the file entity would always become the `Image` type
- enforces a maximum file size for file uploads
- adds the `temporalClient` to the impure graph context passed to the Node API primitive methods

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- add support for parsing the contents of Spreadsheets (i.e. `XLSX` files) (H-2002)

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch locally (it requires backend changes to take effect)
2. Try uploading a PDF, DOCX or PPTX file
3. View the created entity, and check whether the textual contents property has been assigned the correct value

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/5f49b694-252a-46ad-ba6e-f4e93508fdff

### DIFF
--- a/apps/hash-api/src/storage/index.ts
+++ b/apps/hash-api/src/storage/index.ts
@@ -1,7 +1,7 @@
 import { extractBaseUrl } from "@blockprotocol/type-system";
 import { apiOrigin } from "@local/hash-isomorphic-utils/environment";
 import {
-  currentTimeInstantTemporalAxes,
+  fullDecisionTimeAxis,
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
 import {
@@ -112,7 +112,7 @@ const getFileEntity = async (
   const { entityId, key, includeDrafts = false } = params;
   const [ownedById, entityUuid] = splitEntityId(entityId);
 
-  const [fileEntity, ...unexpectedEntities] = await graphApi
+  const fileEntityRevisions = await graphApi
     .getEntitiesByQuery(actorId, {
       query: {
         filter: {
@@ -139,7 +139,7 @@ const getFileEntity = async (
           ],
         },
         graphResolveDepths: zeroedGraphResolveDepths,
-        temporalAxes: currentTimeInstantTemporalAxes,
+        temporalAxes: fullDecisionTimeAxis,
         includeDrafts,
       },
     })
@@ -151,13 +151,27 @@ const getFileEntity = async (
       return getRoots(subgraph);
     });
 
-  if (unexpectedEntities.length > 0) {
-    throw new Error(
-      `Critical: More than one file entity with entityId ${entityId} and key ${key}.`,
-    );
-  }
+  const latestFileEntityRevision = fileEntityRevisions.reduce<
+    Entity | undefined
+  >((previousLatestRevision, currentRevision) => {
+    if (!previousLatestRevision) {
+      return currentRevision;
+    }
 
-  return fileEntity;
+    const currentCreatedAt = new Date(
+      currentRevision.metadata.temporalVersioning.decisionTime.start.limit,
+    );
+
+    const previousLatestRevisionCreatedAt = new Date(
+      previousLatestRevision.metadata.temporalVersioning.decisionTime.start.limit,
+    );
+
+    return previousLatestRevisionCreatedAt < currentCreatedAt
+      ? currentRevision
+      : previousLatestRevision;
+  }, fileEntityRevisions[0]);
+
+  return latestFileEntityRevision;
 };
 
 /**


apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/009-create-office-file-types.migration.ts
```diff
@@ -0,0 +1,168 @@
...
+        description: "A Microsoft Excel spreadsheet.",
+      },
+      webShortname: "hash",
+      migrationState,
+      instantiator: anyUserInstantiator,
+    },
+  );
+
+  return migrationState;
+};
```
benwerner01, line 166: @nonparibus should I omit the "Spreadsheet" and "XLSX Spreadsheet" definitions in this PR, given we are deferring defining its properties/links?

benwerner01, line 166: @nonparibus @CiaranMn should I omit the "Spreadsheet" and "XLSX Spreadsheet" definitions in this PR, given we are deferring defining its properties/links?

nonparibus: Let's keep them. :)

apps/hash-api/src/graphql/resolvers/knowledge/file/request-file-upload.ts
```diff
@@ -7,7 +8,11 @@ import {
   ResolverFn,
 } from "../../../api-types.gen";
 import { LoggedInGraphQLContext } from "../../../context";
-import { dataSourcesToImpureGraphContext } from "../../util";
+import { graphQLContextToImpureGraphContext } from "../../util";
+
+const maximumFileSizeInMegaBytes = 100;
+
+const maximumFileSizeInBytes = maximumFileSizeInMegaBytes * 1024 * 1024;
```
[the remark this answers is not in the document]
nonparibus: Let's comment inline why we're doing this, so that we recall the reasoning (and don't just remove the cap when we need to support large videos, etc in the future -- without creating a different process for handling those larger files).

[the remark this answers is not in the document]
benwerner01: I've added a comment in https://github.com/hashintel/hash/pull/3929/commits/8c4f10b91d1350a8a398df0e5f94c4361d3c094e, I'm realising the limitation I recall technically only applies to the files which are parsed in temporal. If we feel strongly about not having a limit, then we could also just trigger the temporal workflow for files larger than X, or just enforce the upload size for the file type's which would trigger a temporal workflow.

Or was there another reason we wanted to add a maximum file size I'm forgetting about?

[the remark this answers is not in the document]
nonparibus: Not that I'm aware of. @CiaranMn?

nonparibus:
BE integration linting is failing here.

apps/hash-api/src/graph/knowledge/primitive/entity/after-update-entity-hooks/file-document-after-update-entity-hook-callback.ts
```diff
@@ -0,0 +1,92 @@
...
+type FileEntityToParseProperties =
+  | DOCXDocumentProperties
+  | PDFDocumentProperties
+  | PPTXPresentationProperties;
+
+export const parseTextFromFileAfterUpdateEntityHookCallback: UpdateEntityHookCallback =
+  async ({ entity, updatedProperties, context, authentication }) => {
+    const { temporalClient } = context;
+
+    const docxFileEntity =
```
CiaranMn, line 29: Should this be called `docx` when it might be two other types?

apps/hash-api/src/graph/knowledge/system-types/file.ts
```diff
@@ -63,12 +84,39 @@ const generateCommonParameters = async (
...
+
+    const entityTypeIdByMimeType = getEntityTypeIdForMimeType(mimeType);
+
+    let entityTypeId: VersionedUrl;
+
+    if (specifiedEntityTypeId) {
+      /**
+       * If there is a mime entity type ID and it is not the same
+       * as the specified entity type ID, override it. Otherwise,
+       * use the specified entity type ID.
```
CiaranMn, line 97: I thought we would allow users to specify their own `entityTypeId` if they wanted, and we would only override it if it was a `hash` system type

apps/hash-api/src/graph/context-types.ts
```diff
@@ -3,11 +3,13 @@ import { DataSource } from "apollo-datasource";
 
 import { AuthenticationContext } from "../graphql/authentication-context";
 import { UploadableStorageProvider } from "../storage/storage-provider";
+import { TemporalClient } from "../temporal";
 
 export type GraphApi = GraphApiClient & DataSource;
 
 export type ImpureGraphContext<WithUpload extends boolean = false> = {
   graphApi: GraphApi;
+  temporalClient: TemporalClient | null;
```
CiaranMn, line 12: If this was omittable then you wouldn't need to change it everywhere... alternatively could take an approach like the `WithUpload` generic whereby the function that needs `temporalClient` is responsible for declaring that it needs it, with limits the required changes to only the functions that need it and their callers. If it's simple to revert the commit that made all the changes and do one of these things it would save future hassle

apps/hash-api/src/graphql/resolvers/knowledge/file/request-file-upload.ts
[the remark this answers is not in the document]
CiaranMn: Given that the entire text content is saved as a property on the entity, it is also a memory issue for the database and anywhere we pass the entity around, depending on how big the strings are derived from these files. Really we don't want entities with 10MB properties in them, let alone 100MB or anything greater. This won't scale (imagine querying and viewing a list of documents that all have huge strings in them) and we need to store the text content some other way before any real usage.

apps/hash-api/src/graph/knowledge/primitive/entity/after-update-entity-hooks.ts
```diff
@@ -1,205 +1,24 @@
...
   },
   {
     entityTypeId: systemEntityTypes.user.entityTypeId,
-    callback: userAfterUpdateHookCallback,
+    callback: userAfterUpdateEntityHookCallback,
   },
+  ...entityTypesToParseTextFrom.map((entityTypeId) => ({
+    entityTypeId,
+    callback: parseTextFromFileAfterUpdateEntityHookCallback,
+  })),
```
CiaranMn, line 23: I should've thought of this sooner, but how this does actually get reliably triggered? The user uploads the file after the entity is created, and the file is not mutated after the upload is complete.

CiaranMn (commented):
Thanks @benwerner01 – a few comments. I think there are two issues we need to figure out:
1. [Now] – actually triggering the text parsing once the file upload is complete. The current flow is that the user requests to upload a file, at which point the entity is created, and then the user uploads the file from the client. The backend is not notified that the file upload is complete and the entity is not mutated again unless the user changes something in it. We need to either (hack) mutate the entity after the upload is complete to trigger the hook, or have some other way of knowing that the file upload is complete so that this process can be triggered. H-684 tracks the creation of a file upload service to handle this better.

2. [Soonish] Don't store the text content as a string in the database for large files. It's going to cause problems in several places if the user wants to retrieve several entities that all have huge strings in them.

apps/hash-external-services/docker-compose.temporal.yml
```diff
@@ -140,7 +140,7 @@ services:
       HASH_GRAPH_API_PORT: "${HASH_GRAPH_API_PORT}"
     tmpfs:
       - /tmp
-    read_only: true
+    read_only: false
```
benwerner01, line 143: @TimDiekmann @CiaranMn just to double check, is there something we have to add to terraform to allow the container to be written to? Or is that already setup? (It can no longer be readonly as the `officeparser` package writes files to the disk temporarily, even if you pass it a `Buffer`)

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/009-create-office-file-types.migration.ts
```diff
@@ -0,0 +1,168 @@
...
+  createSystemEntityTypeIfNotExists,
+} from "../util";
+
+const migrate: MigrationFunction = async ({
+  context,
+  authentication,
+  migrationState,
+}) => {
+  /**
+   * Step 1. Create the `Text Document` entity type and its child entity types.
```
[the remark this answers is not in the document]
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3929/commits/a5373f1c5011e92739e7f11566917749966bca36

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/009-create-office-file-types.migration.ts
```diff
@@ -0,0 +1,168 @@
...
+        description: "A Microsoft Word document.",
+      },
+      webShortname: "hash",
+      migrationState,
+      instantiator: anyUserInstantiator,
+    },
+  );
+
+  /**
+   * Step 2. Create the `Slideshow` entity type and its child entity types.
```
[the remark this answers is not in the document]
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3929/commits/a5373f1c5011e92739e7f11566917749966bca36

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/009-create-office-file-types.migration.ts
```diff
@@ -0,0 +1,168 @@
...
+        description: "A Microsoft PowerPoint presentation.",
+      },
+      webShortname: "hash",
+      migrationState,
+      instantiator: anyUserInstantiator,
+    },
+  );
+
+  /**
+   * Step 3. Create the `Spreadsheet` entity type and its child entity types.
```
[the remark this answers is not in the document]
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3929/commits/a5373f1c5011e92739e7f11566917749966bca36

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/009-create-office-file-types.migration.ts
benwerner01: Yeah I have a similar feeling, have removed it in https://github.com/hashintel/hash/pull/3929/commits/9e6a642e828eb6d55f4a87e8d604c470949e35d6

apps/hash-api/src/graph/knowledge/primitive/entity/after-update-entity-hooks/file-document-after-update-entity-hook-callback.ts
benwerner01: Addressed in https://github.com/hashintel/hash/pull/3929/commits/94af95d3854c8caba8847b1363a76af55f46de5e

apps/hash-api/src/graph/knowledge/system-types/file.ts
benwerner01: We can do this, but this means we have to add the logic in the FE for detecting the file type (so that the correct `entityTypeId` is passed here) which I thought we didn't want. If we do, I can move it there.

apps/hash-api/src/graph/knowledge/primitive/entity/after-update-entity-hooks.ts
benwerner01: I've addressed this via a `uploadCompletedAt` property, as discussed internally. This involved several commits, I recommend looking at the diff since your review to see the changes.

apps/hash-api/src/graph/knowledge/system-types/file.ts
CiaranMn: Detecting it where / why? If they're coming from a HASH system type then we detect the most appropriate specific type here. If they're passing a non-HASH `entityTypeId` then let them use it, no detection is necessary

tests/hash-backend-integration/src/tests/util.ts
```diff
@@ -1,3 +1,6 @@
+/** @todo: figure out why omitting this import causes a TSC error */
+import "@apps/hash-api/src/express";
```
benwerner01, line 2: @CiaranMn I don't understand why this was caused in this PR, but the `@tests/hash-backend-integration` TSC check was failing because it didn't pick up on the overridden `Express` definition. Explicitly importing it was the only way I could think of to resolve this.

benwerner01: Thanks, should've thought of that, have addressed in https://github.com/hashintel/hash/pull/3929/commits/f30b1e786919b4970474da8207bec85cfae6349b. I'm going to keep `express.d.ts` because it allows us to get rid of the eslint ignore and it probably should be in a decleration file anyway.

CiaranMn (approved):
Excellent, thanks @benwerner01 !

apps/hash-api/src/graph/knowledge/system-types/file.ts
benwerner01: The entity type upload tab currently specifies an `entityTypeId`. If we want to let it continue specifying the `entityTypeId`, then we either need to override it here based on the mime type, or override it in the frontend based on the mime type. Otherwise when you upload a DOCX file in the generic "file" uploader in the frontend, you're not going to end up with a file entity of type `DOCX`, but a file entity of type `File`.

benwerner01: Sorry I misunderstood what you were saying. I've made it so that only if a HASH entity type is specified it can be overriden by the mime type, in https://github.com/hashintel/hash/pull/3929/commits/613542848ba315a5ca99c29ede34ae55bacc4190

CiaranMn (dismissed):

apps/hash-api/src/storage/index.ts
```diff
@@ -138,7 +138,7 @@ const getFileEntity = async (
         ],
       },
       graphResolveDepths: zeroedGraphResolveDepths,
-      temporalAxes: fullDecisionTimeAxis,
+      temporalAxes: currentTimeInstantTemporalAxes,
```
benwerner01, line 141: @CiaranMn this was throwing an error when viewing file entities, as we are now updating them so expect there to be more than one version in the history. Not sure why we were using `fullDecisionTimeAxis` here in the first place.

CiaranMn: If the file entity is updated by replacing the file, it gets a new key identifying it in storage – the `key` is passed as part of the query, and includes the edition id (a timestamp).

We need `fullDecisionTimeAxis` to be able to retrieve earlier versions of files. The error is happening now because we are changing the file entity without changing the key, which were doing previously.

This can be tested by going to the file entity's page, saving the key, and clicking the button to replace the image. I assume that attempting to visit the URL for the previous edition's key will now crash.

Probably the easiest answer is to remove the check for unexpected entities and take whatever the latest is with the specified key.

benwerner01: Ah thanks for explaining, I've addressed this in https://github.com/hashintel/hash/pull/3929/commits/20f0f51ef474d80409067407a785c90d9be30695

CiaranMn (approved):

PR #3930, opened 2024-01-25
H-1397: Better logging, error handling in the Node API

CiaranMn:

This improves logging and error handling in the API by:
1. Moving the request logging middleware to before the routes it is supposed to apply to, so that it works
2. Similarly moving the Sentry tracing middleware up above the request routes
3. Setting the request context consistently between GraphQL and non-GraphQL requests
4. Clear the scope and breadcrumbs in each request, since otherwise Sentry has breadcrumbs / scope from other requests (this should not be required so there is still something wrong somehow)
5. Add a fallback error handler for non-GraphQL requests
6. Change the Node API's log level in production from `debug` to `info`

It also **adds some temporary debug logs** to try and figure out why file retrieval is not working in production, which I will remove after looking at them. This relates to H-1943.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

1. Scope should not be bleeding across requests in the first place, clearing it might be hiding other issues
2. Non-GraphQL requests get the full headers object added automatically, but GraphQL requests don't get this.

Both of these issues might be solved by using the Sentry ApolloServer integration, once it works for ApolloServers constructed by passing a `schema` argument (it currently does not).

## 🛡 What tests cover this?

- Existing / manual

## ❓ How to test this?

1. With `SENTRY_DSN` set for the Node API, throw an error in both a non-Sentry route (e.g. `/health-check`) and in the GraphQL playground (just pass any bad query)

### DIFF
--- a/apps/hash-api/src/index.ts
+++ b/apps/hash-api/src/index.ts
@@ -5,7 +5,7 @@ import {
   realtimeSyncEnabled,
   waitOnResource,
 } from "@local/hash-backend-utils/environment";
-import express, { raw } from "express";
+import express, { ErrorRequestHandler, raw } from "express";
 
 // eslint-disable-next-line import/order
 import { initSentry } from "./sentry";
@@ -40,6 +40,7 @@ import {
   addKratosAfterRegistrationHandler,
   createAuthMiddleware,
 } from "./auth/create-auth-handlers";
+import { getActorIdFromRequest } from "./auth/get-actor-id";
 import { setupBlockProtocolExternalServiceMethodProxy } from "./block-protocol-external-service-method-proxy";
 import { RedisCache } from "./cache";
 import {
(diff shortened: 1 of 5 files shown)


apps/hash-api/src/index.ts
```diff
@@ -223,6 +242,39 @@ const main = async () => {
...
+
+    // Clear the scope and breadcrumbs – requests seem to bleed into each other otherwise
+    scope.clear();
+    scope.clearBreadcrumbs();
+
+    /**
+     * Sentry automatically populates a 'Headers' object, but for some reason it doesn't do this for GraphQL requests.
+     * This might be something to do with how Sentry hooks into fetch that doesn't play nicely with ApolloServer,
+     * or how we're loading it.
+     */
```
TimDiekmann, line 261: Does it help if you add `new Sentry.Integrations.GraphQL()` to the `initSentry` function?

CiaranMn: Just tried it, didn't seem to do anything

TimDiekmann (approved):

PR #3933, opened 2024-01-25
H-2007: Fix crash in profile page when adding bio

CiaranMn:

One of the linting "fixes" introduced in #2940 caused a crash when adding a bio to a user or org's profile page. I don't know why because the query is not supposed to be run unless `profile.hasBio` is defined, but it crashes because `profile.hasBio` is undefined once a bio is added.

I've taken the opportunity to add a test for the profile page which checks that it can be rendered and the bio edited.

I noticed that Playwright tests are cached regardless of whether files in the apps they are testing change, so I've disabled caching for the `test:system` command. Since if it passes it only takes a couple of minutes, this seems easier than trying to define all the files that might change it as `inputs`.

This PR also adds some more debug logs to file download in an attempt to debug H-1943. I will remove them in the PR that fixes it.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The added ones.

## ❓ How to test this?

1. Check that the tests pass.

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
+++ b/apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
@@ -41,7 +41,8 @@ export const ProfileBio: FunctionComponent<{
     GetEntityQueryVariables
   >(getEntityQuery, {
     variables: {
-      entityId: profile.hasBio!.profileBioEntity.metadata.recordId.entityId,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain -- potential crash otherwise
+      entityId: profile.hasBio?.profileBioEntity.metadata.recordId.entityId!,
       ...blockCollectionContentsGetEntityVariables,
     },
     skip: !profile.hasBio,
@@ -212,6 +213,7 @@ export const ProfileBio: FunctionComponent<{
               marginRight: -2,
               marginTop: -1,
             }}
+            aria-label={isEditing ? "Save Bio" : "Edit Bio"}
           >
             {isEditing ? <CheckRegularIcon /> : <PenRegularIcon />}
           </IconButton>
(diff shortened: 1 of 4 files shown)


apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
```diff
@@ -41,7 +41,8 @@ export const ProfileBio: FunctionComponent<{
     GetEntityQueryVariables
   >(getEntityQuery, {
     variables: {
-      entityId: profile.hasBio!.profileBioEntity.metadata.recordId.entityId,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain -- potential crash otherwise
+      entityId: profile.hasBio?.profileBioEntity.metadata.recordId.entityId!,
       ...blockCollectionContentsGetEntityVariables,
     },
     skip: !profile.hasBio,
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx:48]: I think this is happening because the entire arguments object is evaluated before it is passed to the hook, so even though it has no use for the variables if the query is skipped `profile.hasBio` is still being accessed. Will keep this in mind in future, thanks for catching.

tests/hash-playwright/tests/profile-page.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  await page.keyboard.type("Alice's bio");
+
+  await page.click("[aria-label='Save Bio']");
+
+  await sleep(2_000);
+
+  await page.reload();
+
+  await expect(page.locator("text=@alice")).toBeVisible();
```
benwerner01, line 40: We could also check if the `Save Bio` persisted

benwerner01 (dismissed):
Thanks Ciaran, had a minor thought but nothing that blocks merging 👍

apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx
Comment (CiaranMn) [reply apps/hash-frontend/src/pages/[shortname].page/profile-bio.tsx:48]: Ah of course it does, yes. Thanks!

tests/hash-playwright/tests/profile-page.spec.ts
CiaranMn: Done in https://github.com/hashintel/hash/pull/3933/commits/4889c986a7e85a9099af507313868f37532fab31

benwerner01 (approved):

PR #3939, opened 2024-01-26
H-1471, H-1481: Use vector similarity search to find existing entities during inference process

CiaranMn:

This PR uses the new vector search during the entity inference process, to look for existing entities to update by semantic distance rather than exact string match. 

It also fixes H-1471 by prioritising the entity's label property in addition to other likely discriminating properties (e.g. `name`).

I've also added something somewhat hacky to the Node API's `getEntities` function to convert any string input for `cosineDistance` into an embedding. This makes it easier to test out distance searches via the Node API, and unblocks adding a search bar or other semantic search features.

## Known issues

At the moment I'm calling the function to `createEntityEmbeddings` directly rather than calling the activity, because activities can't call activities. Ideally we would be orchestrating activities from workflows, but the AI inference activity will take a lot of refactoring to be able to split it up to the point where we can call `createEntityEmbeddings` from it (since the point we need the embedding is deep inside the inference iteration).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None / manual.

## ❓ How to test this?

1. You'll need to be rebuild or run the AI Worker locally
2. Create a `Company` or `Person` type with a name
3. Find pages where entities of this type are described slightly differently – e.g. on wiki, companies sometimes have their full legal name, and sometimes a shorter name. Or individuals have their full name, and elsewhere only firstname/surname.
4. Trigger inference on one of these  pages to create the entity, then visit one where they're described differently and check they are updated rather than a new entity created


apps/hash-ai-worker-ts/src/activities/embeddings.ts
```diff
@@ -1,80 +0,0 @@
-import type {
```
CiaranMn, line 1: This has been moved and a function signature changed a little.

apps/hash-ai-worker-ts/src/activities/shared/embeddings.ts
```diff
@@ -0,0 +1,105 @@
...
+  entityProperties: EntityPropertiesObject;
+  propertyTypes: { title: string; $id: VersionedUrl }[];
+}): Promise<{
+  embeddings: {
+    /** If 'property' is absent, this is the combined embedding for all properties */
+    property?: BaseUrl;
+    embedding: number[];
+  }[];
+  usage: Usage;
+}> => {
```
CiaranMn, line 49: This is the only real change in this file – I changed the signature a little to make it easier to call from elsewhere (it doesn't need all the metadata etc it was previously typed as needing)

apps/hash-frontend/src/pages/shared/hidden-types.ts
```diff
@@ -6,6 +6,7 @@ import { systemEntityTypes } from "@local/hash-isomorphic-utils/ontology-type-id
  */
 export const hiddenEntityTypeIds: VersionedUrl[] = [
   systemEntityTypes.block.entityTypeId,
+  systemEntityTypes.browserPluginSettings.entityTypeId,
```
CiaranMn, line 9: Drive-by: hide this from sidebar

apps/hash-graph/package.json
```diff
@@ -17,8 +17,8 @@
     "build:docker:prod": "docker buildx build --build-arg PROFILE=production --tag hash-graph --tag hash-graph:prod --file docker/Dockerfile ../../ --load",
     "build:docker:test": "docker buildx build --build-arg PROFILE=dev --build-arg ENABLE_TEST_SERVER=yes --tag hash-graph --tag hash-graph:test --file docker/Dockerfile ../../ --load",
     "codegen:generate-openapi-specs": "just generate-openapi-specs",
-    "dev": "HASH_GRAPH_LOG_LEVEL=info cargo run --bin hash-graph --all-features -- server",
-    "dev:test-server": "HASH_GRAPH_LOG_LEVEL=info cargo run --bin hash-graph --all-features -- test-server --api-port 4001",
+    "dev": "cargo run --bin hash-graph --all-features -- server",
+    "dev:test-server": "cargo run --bin hash-graph --all-features -- test-server --api-port 4001",
```
CiaranMn, line 21: Drive-by: don't force into `info`. This should be overwritten in `.env.local` if desired.

apps/hash-api/src/graphql/create-apollo-server.ts
```diff
@@ -176,7 +176,10 @@ export const createApolloServer = ({
         },
       },
       ApolloServerPluginLandingPageGraphQLPlayground({
-        settings: { "request.credentials": "include" },
+        settings: {
+          "request.credentials": "include",
+          "schema.polling.enable": false,
```
CiaranMn, line 181: We don't need to poll the schema in the playground and it makes it extremely difficult to see anything going on in the Graph API at higher log levels

apps/hash-ai-worker-ts/src/activities/infer-entities/dereference-entity-type.ts
```diff
@@ -160,8 +165,18 @@ export const dereferenceEntityType = (
     (entityType) => entityType.schema.$id === linkEntityTypeUrl,
   );
 
+  let labelProperty: BaseUrl | undefined;
   const mergedProperties: DereferencedEntityType["properties"] = {};
+
   for (const entityType of entityTypeWithAncestors) {
+    /**
+     * Take the label property from the first entity type in the inheritance chain which has one.
+     * The first item in the array is th entity type itself.
```
benwerner01, line 174:
```suggestion
     * The first item in the array is the entity type itself.
```

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -134,9 +139,46 @@ export const createEntity: ImpureGraphFunction<
 export const getEntities: ImpureGraphFunction<
   {
     query: EntityStructuralQuery;
+    temporalClient?: TemporalClient;
   },
   Promise<Subgraph<EntityRootType>>
-> = async ({ graphApi }, { actorId }, { query }) => {
+> = async ({ graphApi }, { actorId }, { query, temporalClient }) => {
```
benwerner01, line 145: Just want to note in https://github.com/hashintel/hash/pull/3929 the `temporalClient` will become available in the context (i.e. the first parameter here), but since that will likely merge after this PR I'll address this when resolving the conflicts.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -134,9 +139,46 @@ export const createEntity: ImpureGraphFunction<
...
+          (params: CreateEmbeddingsParams) => Promise<CreateEmbeddingsReturn>
+        >("createEmbeddings", {
+          taskQueue: "ai",
+          args: [
+            {
+              input: [stringInputValue],
+            },
+          ],
+          workflowId: genId(),
+        });
```
benwerner01, line 176: Just for my understanding, what's the thinking behind doing this in a temporal worker? My thinking is it makes sense to call an embeddings API directly when generating embeddings as part of a `getEntities` query, so that getting the response isn't blocked on waiting for your workflow to reach the front of the temporal queue.

benwerner01 (approved):
Thanks Ciaran, this is working for me locally and looks good to me. Had a couple of thoughts/nits, but nothing that blocks merging.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
CiaranMn: I agree that this may become a problem, and have created H-2017 to follow up. We should fix it before actually using this (e.g. via a search bar)

apps/hash-ai-worker-ts/src/activities/infer-entities/dereference-entity-type.ts
CiaranMn: https://github.com/hashintel/hash/pull/3939/commits/07d7f6fa4f1dd0c88246f32871553fe8825504ea

apps/hash-api/src/graph/knowledge/primitive/entity.ts
CiaranMn: Thank you

benwerner01 (approved):

PR #3940, opened 2024-01-26
H-1761: Add support for exporting tables of entities to `CSV`

benwerner01:

This PR adds the ability to export the entities table as a CSV file. It also improves how property values are rendered in the entities table (previously an array or objects would've displayed as `[Object]`, rather then a stringified version of that.

## 🔗 Related links

- H-1761

## 🔍 What does this change?

- fixes a styling issue in the table header, where the "Filter" and "Export to CSV" buttons would've been too tall

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

1. Go to an instance of the entities table (on the `/entities` page, or the entities tab on an entity type)
2. Click "Export to CSV"
3. Open the downloaded CSV file, and view the containing data

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/c8b2ffda-84d1-4e64-b3a7-640ae19b76a6


apps/hash-frontend/src/components/grid/grid.tsx
```diff
@@ -469,6 +479,11 @@ export const Grid = <T extends Row & { rowId: string }>({
     [columns, openFilterColumnKey],
   );
 
+  if (currentlyDisplayedRowsRef && sortedAndFilteredRows) {
+    // eslint-disable-next-line no-param-reassign
+    currentlyDisplayedRowsRef.current = sortedAndFilteredRows;
+  }
```
CiaranMn, line 485: This looked a bit out of place to me... my instinct would be to lift the `sortedAndFilteredRows` state out of the component (as `selectedRows` already is, for example). 

Doing this via a ref takes a bit longer to read, because it prompts the question of why it's done like this. One explanation is to prevent the parent component from re-rendering when the sorted and filtered rows change. Another is that moving it up would also involve moving the column filter state etc up and until we need filtering/sorting in more places, it's not worth it. It might be worth adding a comment explaining why it's done like this.

apps/hash-frontend/src/pages/shared/entities-table.tsx
```diff
@@ -414,9 +439,12 @@ export const EntitiesTable: FunctionComponent<{
...
         columnKey: "lastEditedBy",
-        filterItems: lastEditedByUsers.map(({ accountId, preferredName }) => ({
-          id: accountId,
-          label: preferredName ?? "Unknown User",
+        filterItems: lastEditedByActors.map((actor) => ({
+          id: actor.accountId,
+          label:
+            ("displayName" in actor
+              ? actor.displayName
+              : actor.preferredName) ?? "Unknown Actor",
```
CiaranMn, line 447: I know we've spoken about this before but I'm wondering again if it would be simpler just to make it `displayName` for both machines and human actors in the frontend to save all these ternaries. Not suggesting any changes now but interested in your view having done this PR and whether we do it next time we're touching this stuff

apps/hash-frontend/src/shared/table-header.tsx
```diff
@@ -350,6 +334,9 @@ export const TableHeader: FunctionComponent<TableHeaderProps> = ({
             </Box>
           </Box>
         </Box>
+        {generateCsvFile ? (
+          <ExportToCsvButton generateCsvFile={generateCsvFile} />
```
CiaranMn, line 338: Is there a simple default export function we can use that just exports exactly what is rendered in the table? e.g. for use in the type tables

CiaranMn (commented):
Thanks @benwerner01, entities CSV functionality looks good – just a few questions

apps/hash-frontend/src/components/grid/grid.tsx
benwerner01: I agree it's a bit odd, but the reasons you mentioned I wasn't sure how to best approach exposing this without moving the filtering logic out of the `Grid` component, which I'm not too sure about. I've added a comment in https://github.com/hashintel/hash/pull/3940/commits/c5991703dd91bb3be0697f6378adfc9dea9adf77 to make this a bit clearer for now.

apps/hash-frontend/src/pages/shared/entities-table.tsx
benwerner01: I think it'd make sense and agree save a lot of unnecessary hassle, but I'd still keep `preferredName` on `MinimalUser` as it's the property on the entity even though they'll have the same value. We could probably do the same for `MinimalOrg` as well.

apps/hash-frontend/src/shared/table-header.tsx
benwerner01: Not at the moment, but it shouldn't be too hard to extract a base `generateCsvFromTable` method which is used in the `TableHeader` by default, and then via props specify a second `getAdditionalCsvData` method that fetches the additional data for the entities table. Should I add this as part of this PR?

CiaranMn: I think we might as well, since it's a cheap way of getting an export option for any other tables also

CiaranMn: You could also just default to `generateCsvFromTable` if a custom function isn't passed (if `getAdditionalCsvData` becomes a pain)

benwerner01: I've added the export to CSV button to the `TableHeader` component (and therefore the types tables) in https://github.com/hashintel/hash/pull/3940/commits/ec3e8dee8200dea9e89699d11408586030bb4193

CiaranMn (dismissed):

CiaranMn (approved):

PR #3941, opened 2024-01-26
H-1943: Fix(?) Redis connection in production

CiaranMn:

This PR possibly / hopefully fixes access between the Node API and the Redis cluster in production, which is causing uploaded files to be inaccessible in production (because Redis is involved). I'm not convinced Redis needs to be involved in file download, but it's currently also a dependency of integration outward sync, which we will want soon.

I believe the issue is that the AWS ElastiCache cluster is configured to only accept encrypted traffic, but the Redis client in the repo wasn't configured to send it if needed. We'll only know after this is deployed though.

I updated the version of the client we were using, which also necessitated a few changes to the logic around it.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

See if this works.

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. Merge it

### DIFF
--- a/apps/hash-api/src/cache/adapter.ts
+++ b/apps/hash-api/src/cache/adapter.ts
@@ -25,5 +25,5 @@ export interface CacheAdapter extends DataSource {
    * Push one or more values onto the end of a list.
    * @returns the new length of the list.
    * */
-  rpush(key: string, ...values: string[]): Promise<number>;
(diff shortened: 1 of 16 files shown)


TimDiekmann (approved):
Changes look good, let's see if it helps 🚀

PR #3947, opened 2024-01-29
H-1587, H-1949, H-2103: Better type permission checks, hide warning when creating link types

CiaranMn:

Makes the following type permission-related changes:
1. Allow checking a user's permissions on a type via the API
2. On the new entity page, show an error message if a user can't create entities of that type
3. Hide the 'create entity' button on the type page if a user can't create entities of that type
4. On the type page, ese the permissions from the API to check if the type form should be readonly or not, rather than members of the org can edit any type in the org

Smaller drive-bys:
- fixes H-1949 by hiding the "You are extending a type" warning when creating a new link type (all link types extend the Block Protocol link type)
- catch errors when trying to create an entity and show a warning modal
- hide 'Profile Bio' and 'Text' entities from the sidebar entities list

Following further testing:
- Fix query for entities page (H-2103) and add test
- Don't show 'Loading...' for entities table with no results

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

It would be better to hide types that users can't create from the 'select type of entity to create' dropdown, and also when selecting parents of a type, but this is a much bigger change since it means updating how we fetch all entity types to also fetch their permissions. Created H-2027 to address this later.

## 🛡 What tests cover this?

- Type creation is covered by a Playwright test

## ❓ How to test this?

1. Check that you can still types you're supposed to be able to and can't edit types you're not
2. Visit type page for Actor, check 'create entity' button doesn't show
3. Try creating a new entity of type Actor (view /new/entity route)

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/13c5b235-e1d1-4fe3-bd55-3b3339864816


nonparibus:
@CiaranMn Video demo was helpful, thanks! Related to it...

Q1) If I'm creating an entity, and it has a bunch of types, and maybe even some property values... and then I try to add "Actor" (or another restricted entity type) to the entity, what happens?

Q2) Can we just avoid showing these entity types in the list entirely, in the first place?

CiaranMn:
@nonparibus 

> Q1) If I'm creating an entity, and it has a bunch of types, and maybe even some property values... and then I try to add "Actor" (or another restricted entity type) to the entity, what happens?

The `instantiate` permission only currently gates creation of entities directly of that type, not of types that inherit from it. So in your example everything behaves normally.

> Q2) Can we just avoid showing these entity types in the list entirely, in the first place?

We can – I've created H-2027 to address this later as it involves a wider change in how we populate the list of entity types, and it might be worth bundling with pagination / prioritisation of what types the user actually sees, versus just getting everything in the db as we do now.

nonparibus:
@CiaranMn Okay. Re: Q1, we will want to revisit/double-check this behavior is still working as intended when we support multi-type entities.

CiaranMn:
> One thought I had is we might also want to hide the "Create" button on the /entities page if the user cannot instantiate the type?

@benwerner01 I've added this in https://github.com/hashintel/hash/pull/3947/commits/5819ea2882969194ba5b2fc77095cf523fc193f7

In testing that I also noticed and fixed the following:
1. Entities table would show 'Loading...' if there were no entities. Changed to 'No results' in https://github.com/hashintel/hash/pull/3947/commits/95337f2906abef5c84c7a3323171c37653f1b202
2. Query for entities of a specific type on the entities page was incorrect. I've added a test covering it in https://github.com/hashintel/hash/pull/3947/commits/01c83293a43ca46ba12722283659503aba32a0b2

benwerner01 (approved):

PR #3948, opened 2024-01-29
H-1930, H-2036: Fetch `createdAt`/`createdBy` from entity metadata instead of initial revision

benwerner01:

This PR refactors places in the frontend where we previously fetched the first revision entity to obtain the `createdAt` and `createdBy` fields of an entity, to retrieve this from its metadata instead.

As a drive-by it removes a usage of the BP query language in the `useEntityTypeEntities` hook, so that drafts can be included optionally (which may also address H-1956).

## 🔗 Related links

- H-1930

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

Checkout the branch locally, or view the deployment, ensuring the following behave as they did previously:

1. The `/drafts` page "created by" and "created at" chips
2. The `/notes` page "created at" timestamps
3. The `/notifications` page "created at" timestamps
4. The user profile "joined at" timestamp


CiaranMn (commented):
Thanks Ben, nice to see all these extra requests disappearing!

I think the new fields allow us to also:
- remove fetching the page again in `getPageAuthor` (can probably remove this function altogether, its caller already has `blockCollectionEntity` it can get the creator of)
- address a `@todo` in `comment-block` related to the `createdAt`

benwerner01:
@CiaranMn thanks for catching those additional improvements, have made them in https://github.com/hashintel/hash/pull/3948/commits/20ab9cfcd12bcbf927146410fce267a8458da2da and https://github.com/hashintel/hash/pull/3948/commits/f0811a03f2cf36bdc1c0654f61bcaea64e6a218b

CiaranMn (approved):

PR #3974, opened 2024-02-01
H-2012: Allow browsing file thumbnails in a grid (e.g. images, PDFs, DOCs, PPTXs)

benwerner01:

This PR adds a `Grid` view for the `EntitiesTable` when only file entities are being displayed.

## 🔗 Related links

- H-2012

## 🔍 What does this change?

- fixes a console error where a tooltip was being used on a disabled button
- adds a `useEntityHref` hook to make the logic for generating an entity's href more re-usable

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

Go to a page where only files are being rendered in the `EntitiesTable`, for example on the entities tab on `http://localhost:3000/@hash/types/entity-type/file/v/2`. Click on the `Grid` icon button to view the entities in a grid view.

## 📹 Demo

https://github.com/hashintel/hash/assets/42802102/ad16f4ce-e107-4ab8-9b29-d81a74334f97


nonparibus:
@benwerner01 Can we make this new grid view the default when viewing `Entities` of the `File` type, as well as any other type that extends it?

apps/hash-frontend/src/pages/shared/entities-table.tsx
```diff
@@ -531,25 +555,33 @@ export const EntitiesTable: FunctionComponent<{
...
+              ] satisfies EntityTableView[]
+            ).map((viewName) => (
+              <ToggleButton
+                key={viewName}
+                disableRipple
+                value={viewName}
+                aria-label={viewName}
+              >
+                <Tooltip
+                  title={`${viewName[0]?.toUpperCase()}${viewName.slice(1)} view`}
```
CiaranMn, line 576: This would be be simpler / easier to read if the values were already 'Table', 'Grid', 'Graph' – since it's just local state I think the simplicity is worth more than any ickiness of casing a value like that

apps/hash-frontend/src/pages/shared/entities-table/grid-view.tsx
```diff
@@ -0,0 +1,38 @@
...
+        entities.map((entity, index, all) => (
+          <GridViewItem
+            key={entity.metadata.recordId.entityId}
+            entity={entity}
+            numberOfItems={all.length}
+            index={index}
+          />
+        ))
+      ) : (
+        <>skeleton</>
```
CiaranMn, line 34: Spooky

CiaranMn (commented):
Thanks Ben – a couple of small comments

apps/hash-frontend/src/pages/shared/entities-table/grid-view.tsx
benwerner01: Oops 😅 Added the skeleton in https://github.com/hashintel/hash/pull/3974/commits/38077238c1c1b09a25dfe4b066f922794f591891, looks like this:

image

benwerner01:
> Can we make this new grid view the default when viewing Entities of the File type, as well as any other type that extends it?

@nonparibus I've addressed this in https://github.com/hashintel/hash/pull/3974/commits/fc8e03110dbe27480d776616c67c0314440447c4, have made an effort to ensure the correct view is displayed on the first render of the component.

apps/hash-frontend/src/pages/shared/entities-table.tsx
benwerner01: Agreed, have done so in https://github.com/hashintel/hash/pull/3974/commits/ef9e6a40582fae6880456923307149761c554639

CiaranMn (approved):

PR #3984, opened 2024-02-02
H-142: Allow users to see and edit `labelProperty` in type editor

CiaranMn:

Types have an optional `labelProperty`, which indicates the property on them that should be used to label an entity.

This PR allows users to see and edit the `labelProperty` for a given type (including indicator any label property inherited from a parent, if the type doesn't have one specifically set on it).

## 🔍 What does this change?

- Much of the diff is related to providing metadata to the type editor alongside the schema, which wasn't previously happening
- The PR also adds a tooltip on inherited properties specifying which type they were inherited from

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None for setting / viewing label property

## ❓ How to test this?

1. Create or find an existing type, try setting a property as a label property
2. Extend the type, check the inherited label appears and correctly is superseded by any label property set on the type


apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
```diff
@@ -39,17 +39,13 @@ export const EntityTypeHeader = ({
   const [showExtendTypeModal, setShowExtendTypeModal] = useState(false);
 
   const isLatest =
-    !latestVersion || extractVersion(entityType.$id) === latestVersion;
-  const latestVersionUrl = entityType.$id.replace(/\d+$/, `${latestVersion}`);
+    !latestVersion || extractVersion(entityTypeSchema.$id) === latestVersion;
+  const latestVersionUrl = entityTypeSchema.$id.replace(
+    /\d+$/,
+    `${latestVersion}`,
+  );
```
benwerner01, line 46: Nit: this works, but it might be slightly more readable if we were to use a combination of `extractBaseUrl` and `versionedUrlFromComponents` here. There is a type error this would introduce though with the usage of `latestVersion`.
```suggestion
  const latestVersionUrl = versionedUrlFromComponents(
    extractBaseUrl(entityTypeSchema.$id),
    latestVersion ?? 0,
  );
```

benwerner01 (approved):
Thanks Ciaran, looks good and is working well for me. Had one small nit, but spotted nothing that should block merging 👍

apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
CiaranMn: Updated in https://github.com/hashintel/hash/pull/3984/commits/999f11786f44303ff3f57c81264a26f72890ffcd

benwerner01 (approved):

PR #4022, opened 2024-02-09
H-2159: Implement temporal versioning without a Postgres update trigger

TimDiekmann:

The current implementation uses a trigger to properly handle a bi-temporal update. This comes with a few downsides:

- It's not possible to re-use the trigger for setting an end-time (this is the main issue currently)
- It's not easily possible to get the previous edition id
- Maintainability is bad as migration files must not change

## 🔍 What does this change?

- Remove the triggers from Postgres and implement it manually in the Graph logic. 

This PR:

The changes in this PR:

The changes in this PR:


indietyp (approved):
LGTM! Excited to see this change happening.

CiaranMn (approved):

PR #4030, opened 2024-02-12
H-2080: Extend `EntityId` to include a `draft_id`

TimDiekmann:

Instead of storing `draft: boolean` inside of the entity metadata, the `EntityId` should be used to identify the draft. Currently, a single draft for entities is possible. This can be set to "live" by removing the draft ID from the EntityId.

## 🔍 What does this change?

- Adjust tables to store `draft_id` as part of the entity id
- Remove the `draft` boolean flag in the metadata
- Adjust the Graph service and the Node API to take these changes into account

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Allow forking an entity to create a new draft, which then can be used as the new live entity once desired
## 🛡 What tests cover this?

Every test which touches draft will will test the new functionality. Also, a snapshot test were adjusted to include a draft to test the new `EntityId` format.


apps/hash-ai-worker-ts/src/activities/infer-entities/persist-entities/update-entities.ts
```diff
@@ -111,8 +112,12 @@ export const updateEntities = async ({
               ...properties,
             };
 
+            const draft = !!extractDraftIdFromEntityId(
+              existingEntity.metadata.recordId.entityId,
+            );
```
CiaranMn, line 117: Not to solve now, we'll need to think about how the 'find existing entity to update' logic works once we have multiple drafts of an entity, and a live one... whether the AI is proposing updates to the live entity, or to one of the drafts edits to it – and if the latter, how we choose between them (multiple people might each have a draft working copy of the same entity)

cc @nonparibus

CiaranMn (approved):
Excellent stuff, good work @TimDiekmann!

apps/hash-ai-worker-ts/src/activities/infer-entities/persist-entities/update-entities.ts
nonparibus: We want each attribute of an entity to be independently governed, so we can diff two entities, and only overwrite/update changed or different attribute values, in the case of multiple drafts (i.e. basically supporting "merge" of two entities).

PR #4031, opened 2024-02-13
H-2185: Replace `preferredName` property type with `displayName` property type for users

benwerner01:

This PR updates the `User` entity to inherit the `displayName` property from `Actor`, replacing the existing `preferredName` property type. This means that the display name of all actors (machines and users) can be consistently obtained from the same property.

## 🔗 Related links

- H-2185

## 🔍 What does this change?

- fixes several issues that enable updating the `Machine` entity type and udpating its entities in the migration scripts

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

Checkout the branch and ensure the HASH frontend displays the user's display name where it would've displayed a preferred name previously.

## 📹 Demo

image


libs/@local/hash-isomorphic-utils/src/system-types/actor.ts
```diff
@@ -3,15 +3,19 @@
  */
 
 import {
-  Actor,
-  ActorOutgoingLinkAndTarget,
-  ActorOutgoingLinksByLinkEntityTypeId,
-  ActorProperties,
+  ActorV2,
```
CiaranMn, line 6: Any version qualifier in these generated types is an indication that an older version is still referenced by some existing system types – it's worth reviewing the generated types for these and figuring out what is referencing them still. In this case there's still a User v3 in the system types, which is probably why ActorV1 is still around. And User v3 is probably still around because there are other types which link to it – `comment`, `commentNotification`, `linearIntegration` and `mentionNotification` at least

One way of chasing these old references down is to use the GraphQL playground to get all entity types, and then search the old Verisoned URL in the output
```graphql
{
  queryEntityTypes(
    constrainsValuesOn: { outgoing: 0 }
    inheritsFrom: { outgoing: 0 }
    constrainsPropertiesOn: { outgoing: 0 }
    constrainsLinksOn: { outgoing: 0 }
    constrainsLinkDestinationsOn: { outgoing: 0 }
  ) {
    vertices
  }
}
```

libs/@local/hash-isomorphic-utils/src/system-types/comment.ts
```diff
@@ -91,10 +91,10 @@ import {
   HasTextOutgoingLinkAndTarget,
   HasTextOutgoingLinksByLinkEntityTypeId,
   HasTextProperties,
-  Image,
-  ImageOutgoingLinkAndTarget,
-  ImageOutgoingLinksByLinkEntityTypeId,
-  ImageProperties,
+  ImageV1,
```
CiaranMn, line 94: `ImageV1` and `ImageV2` are both present because migration `009` updated `Image` to v2, but didn't update anything that referred to `Image`. So I think we need a clean-up migration that upgrades `Image`'s dependencies.

I think this is worth doing in this PR, because it is going to involve updating `User` (via has avatar / has cover image).

CiaranMn (commented):
Thanks @benwerner01 – mostly looks good, I think we need to clean up some inconsistencies in the system types' references to each other though

libs/@local/hash-isomorphic-utils/src/system-types/actor.ts
benwerner01: Thank you for spotting this, have addressed in https://github.com/hashintel/hash/pull/4031/commits/c0a1b842a20bc3091f01be292b66e08d00e2109c. I didn't realise the reason the version identifier is introduced is when there are more than one version in the system types output, but this makes sense to me know and will watch out for it in future. Thanks!

libs/@local/hash-isomorphic-utils/src/system-types/comment.ts
benwerner01: I've addressed this in https://github.com/hashintel/hash/pull/4031/commits/e3a19b4ca2ecf5ec65f9463915f30cdd82f22976 as a drive by in the same migration file.

This was more work than I anticipated when I spoke to you earlier today, because it required updating the `Organization` entity type to `v2` which caused the `005` migration script to fail because it was trying to retrieve/create instances of the latest version of the `Organization` entity type, which doesn't exist at that point in the migrations. To resolve this I had to modify the `createOrg` method to allow for creating orgs of a previous version, and the `getOrgByShortname` method to get orgs via the base URL instead of the versioned URL.

In addition, because we were using the latest version of the organization type to determine the system account ID, I had to modify this in [/system-account.ts](https://github.com/hashintel/hash/pull/4031/commits/e3a19b4ca2ecf5ec65f9463915f30cdd82f22976#diff-5b0e2ab5f8b247e9bc0b2402dd7751c10f326b2514bc1ae3d0fe6e08ad391383) to explicitly look for the first version instead of the latest version of the organization, so that the system account ID can be determined before `v2` of the `Organization` type is created.

CiaranMn (approved):

PR #4042, opened 2024-02-15
H-1373, H-2085, H-2160: Implement Draft Entity Updates

TimDiekmann:

Until now it was only possible to create draft entities when initially creating a new entity. This PR adds the functionality to allow creating a new entity draft from an existing entity. It's even possible to create multiple drafts from the same "live" entity. The draft(s) can later be promoted to the new "live" version of that entity. The draft will then be archived.

## 🔍 What does this change?

- Allow creating a draft entity from a given entity 
- Allow switching a draft entity "live"

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- A few tests were added to test the creation of drafts and switching the "live" entity

## ❓ How to test this?

When calling the `update_entity` method, the `draft: bool` flag has to be passed (each time). 
- If the passed entity ID belongs to a draft entity and `draft` is `true` a regular update for the draft entity happens
- If the passed entity ID belongs to a non-draft entity and `draft` is `false` a regular update for the entity happens
- If the passed entity ID belongs to a draft entity and `draft` is `false` the previous "live" entity is archived and the draft becomes the new "live" entity.
- If the passed entity ID belongs to a non-draft entity and `draft` is `true` a new draft from the live entity is created


CiaranMn (approved):
🥳

PR #4064, opened 2024-02-20
H-2201: Configurable notification polling, don't poll without user

CiaranMn:

This PR improves and fixes a few things related to polling for new notifications and draft entities:
1. Corrects the query for draft entities which was changed in #4030 
2. Don't query for draft entities if there's no logged-in user
3. Allow configuring or disabling polling via an environment variable – the main motivation is to enable setting it off to cut down in log noise during development

This PR intentionally does not touch the polling intervals for comments and property value chips, both of which only happen when a user is viewing a specific page. This PR is about being able to configure global polling.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Set `NOTIFICATION_POLL_INTERVAL=0` in `.env.local` to disable notification and draft entity polling, or to any other ms value


apps/hash-frontend/next.config.js
```diff
@@ -36,6 +36,9 @@ process.env.NEXT_PUBLIC_SENTRY_DSN = process.env.SENTRY_DSN ?? "";
 process.env.NEXT_PUBLIC_SENTRY_REPLAY_SESSION_SAMPLE_RATE =
   process.env.SENTRY_REPLAY_SESSION_SAMPLE_RATE ?? 1;
 
+process.env.NEXT_PUBLIC_NOTIFICATION_POLL_INTERVAL =
+  process.env.NOTIFICATION_POLL_INTERVAL;
```
benwerner01, line 40: Whilst testing locally, this broke polling for me entirely when the `NOTIFICATION_POLL_INTERVAL` is not set in `.env.local`. It seems that when the value of it is `undefined` here, without a `?? ""` next stringifies undefined to `"undefined"`, which is then passed to `parseInt` resulting in `NaN` instead of the default `10000` polling interval.

```suggestion
process.env.NEXT_PUBLIC_NOTIFICATION_POLL_INTERVAL =
  process.env.NOTIFICATION_POLL_INTERVAL ?? "";
```

benwerner01 (changes requested):
Thanks @CiaranMn, encountered an issue when `NOTIFICATION_POLL_INTERVAL` is not set, but with the fix I suggested the PR is working as expected 👍

benwerner01 (approved):

PR #4070, opened 2024-02-22
H-1124: Upgrade `glide-data-grid`, fix snapping to cell on hover

CiaranMn:

There is currently weird behavior in our usage of `@glideapps/glide-data-grid`, whereby if the table had horizontal overflow, hovering over a cell that was not fully in view would cause the scroll to jump to the cell (see demo).

It turns out that this was because of a hack we had to enable editing a cell by only clicking on it once – we set the cell as selected on hover, so that clicking on it would count as a second click and open the editor. Setting the cell as selected cause the snapping behavior (which makes sense when clicking on a cell, but not hovering).

Recently released v6 of the package has a prop enabling single-click editing, and so this PR removes the previous hack which fixes the snapping behavior.

It also:
- stops declaring our custom icon names via a patch, which was cumbersome and no longer appears necessary with v6. 
- removes some type assertions that are now unnecessary

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- There is some test coverage of grid usage via the entity creation Playwright test, which involves editing a property.

## ❓ How to test this?

1. See video for testing scroll behavior
2. Check also that editing an entity's properties works as expected, and that selecting rows on the 'all entities/types' table works as expected

### Previous behavior

https://github.com/hashintel/hash/assets/37743469/f5b63439-7ef4-492e-84b6-2315618adb39

### Fixed behavior

https://github.com/hashintel/hash/assets/37743469/fab0d91c-993b-41ee-95a5-7af0c99a8e34


benwerner01 (approved):
Thanks @CiaranMn, the fix works as expected for me. Really great we no longer have to update the patch when adding icons 👍

PR #4072, opened 2024-02-22
H-2210: Store provenance data for first non-draft entity date

TimDiekmann:

It's interesting to know if for a given draft a non-draft entity exists and if so at which time that one was created. This adds two new provenance fields (decision time + transaction time) for the time when an entity of a given ID was first created in a non-draft state.

## 🔍 What does this change?

- Add provenance fields and do required changes to the Graph API
- Drive-by: Fix `created-at-*` not calculated correctly in database
- Drive-by: Fix snapshots not being able to be restored when embeddings are part of the snapshot

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- A bunch of assertions were added to the `draft` tests

### DIFF
--- a/apps/hash-graph/libs/graph/src/knowledge/query.rs
+++ b/apps/hash-graph/libs/graph/src/knowledge/query.rs
@@ -160,8 +160,8 @@ pub enum EntityQueryPath<'p> {
     /// # use serde::Deserialize;
     /// # use serde_json::json;
     /// # use graph::knowledge::EntityQueryPath;
-    /// let path = EntityQueryPath::deserialize(json!(["firstNonDraftCreatedAtTransactionTime"]))?;
-    /// assert_eq!(path, EntityQueryPath::FirstNonDraftCreatedAtTransactionTime);
+    /// let path = EntityQueryPath::deserialize(json!(["firstNonDraftCreatedAtDecisionTime"]))?;
+    /// assert_eq!(path, EntityQueryPath::FirstNonDraftCreatedAtDecisionTime);
     /// # Ok::<(), serde_json::Error>(())
     /// ```
     ///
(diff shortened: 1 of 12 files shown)


apps/hash-graph/libs/graph/src/knowledge/query.rs
```diff
@@ -134,6 +134,41 @@ pub enum EntityQueryPath<'p> {
     /// [`EntityMetadata`]: graph_types::knowledge::entity::EntityMetadata
     /// [`EntityTemporalMetadata`]: graph_types::knowledge::entity::EntityTemporalMetadata
     CreatedAtDecisionTime,
+    /// The timestamp of the transaction time when the [`Entity`] was _first undrafted_ in the
+    /// database.
```
CiaranMn, line 138: I think some clarification like this would be worth it here and on the decision time field
```suggestion
    /// The timestamp of the transaction time when the first non-draft edition of the [`Entity`] 
    /// was inserted into the database. This may be the first edition, if created as non-draft.
```

apps/hash-graph/libs/graph/src/knowledge/query.rs
```diff
@@ -134,6 +134,41 @@ pub enum EntityQueryPath<'p> {
...
+    ///
+    /// This does not take into account if the [`Entity`] was updated with an earlier decision
+    /// time.
+    ///
+    /// ```rust
+    /// # use serde::Deserialize;
+    /// # use serde_json::json;
+    /// # use graph::knowledge::EntityQueryPath;
+    /// let path = EntityQueryPath::deserialize(json!(["firstNonDraftCreatedAtTransactionTime"]))?;
+    /// assert_eq!(path, EntityQueryPath::FirstNonDraftCreatedAtTransactionTime);
```
CiaranMn, line 164: This needs updating on `CreatedAtDecisionTime` also
```suggestion
    /// let path = EntityQueryPath::deserialize(json!(["firstNonDraftCreatedAtDecisionTime"]))?;
    /// assert_eq!(path, EntityQueryPath::FirstNonDraftCreatedAtDecisionTime);
```

CiaranMn (commented):
Thanks @TimDiekmann, just a few comments on the docs

CiaranMn (approved):

PR #4073, opened 2024-02-22
H-2225: Fix Hydra URL environment variable naming, improve consistency

CiaranMn:

I wrote the wrong name for the Hydra URL variables passed to the API in prod.

This PR corrects them.

Because I made the mistake by blindly following the pattern for the Kratos URLs passed to the API, which are different to the pattern we use for other environment variables, I have also renamed the Kratos URL variables for consistency.


TimDiekmann (approved):
Annoying error 🙃

PR #4074, opened 2024-02-22
H-1476: Improve rate limiting on auth routes

CiaranMn:

Makes the following improvements to rate limiting:
1. Set the IP correctly (TBC) in production, accounting for the load balancer
2. On login attempts, rate limit based on the user (email) logging in, not the IP, to avoid brute force attacks spread across multiple IPs
3. Show an error message in the frontend if the rate limit is reached on login

This PR:

The changes in this PR:

### Known issues

The 'Too many attempts' message on the login screen appears above the 'wrong credentials' message, rather than replacing it. I didn't think this worth the time to investigate and fix, and am not sure it's even a problem really. If the user successfully logs in they will stop seeing the messages.

The changes in this PR:

## 🐾 Next steps

- Test that the IP is reported correctly in prod

## ❓ How to test this?

1. Try a failed login 5 times


nonparibus:
> On login attempts, rate limit based on the user (email) logging in, not the IP, to avoid brute force attacks spread across multiple IPs

Can we make sure to rate limit on both, if not doing this already? (Couldn't tell from this comment alone, w/o scanning the PR... but having both is desirable)

CiaranMn:
> Can we make sure to rate limit on both, if not doing this already? (Couldn't tell from this comment alone, w/o scanning the PR... but having both is desirable)

Yes, fair point, done in https://github.com/hashintel/hash/pull/4074/commits/aae41352c23ee18191be7fb59f11735af69c64fe

apps/hash-api/src/index.ts
```diff
@@ -294,6 +284,60 @@ const main = async () => {
...
+   *
+   * Although the proxy would ideally come before the body parser, so that we're passing it through untouched,
+   * we check the body in this process in order to rate limit requests based on the user attempting to log in.
+   */
+  app.use(
+    "/auth/*",
+    authRouteRateLimiter,
+    userIdentifierRateLimiter,
+    cors(CORS_CONFIG),
+    (req, res, next) => {
```
benwerner01, line 298: I assume this would mean we are limiting any kratos request to 5 per 20 seconds, and not rate limiting per kratos route. I can't think of a reason this would be an issue however.

Also not sure we should be rate limiting the `/whoami` request, but we are no longer calling this on every page load so I guess it shouldn't be a problem either.

benwerner01 (approved):
Thanks @CiaranMn, this is working for me locally as expected 👍

benwerner01 (approved):

PR #4089, opened 2024-02-26
H-2232: Reduce size of payload in GPT query entities route

CiaranMn:

This PR reduces the payload size of the GPT query entities route, to help keep it under:
1.  The context window available to the model
2. The maximum response size for GPT actions ([discussion](https://community.openai.com/t/gpt-actions-new-responsetoolargeerror-failure-when-handling-api-response/504153))

It's not clear what either of these values are for custom GPTs, but the PR reduces payload size by avoiding repetition of type information across entities, and moving from a JSON representation to plain text (cutting down on the extra chars/tokens that JSON formatting introduces).

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

It is still possible for the response to be too large – we should also paginate the endpoint (H-2254).

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1. See README in `ai/gpt/`

### DIFF
--- a/apps/hash-api/src/ai/gpt/gpt-query-entities.ts
+++ b/apps/hash-api/src/ai/gpt/gpt-query-entities.ts
@@ -1,6 +1,7 @@
-import { typedEntries } from "@local/advanced-types/typed-entries";
+import { typedEntries, typedKeys } from "@local/advanced-types/typed-entries";
 import { currentTimeInstantTemporalAxes } from "@local/hash-isomorphic-utils/graph-queries";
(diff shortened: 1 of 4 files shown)


benwerner01 (approved):

Whilst testing I encountered the rate limit, to be fair I must've logged in/out within a short amount of time because I was having issues getting it to work, but might be worth keeping an eye on this incase it prevents users from completing the OAuth flow in some circumstances (where they try to authenticate twice, because they used the wrong password on a login attempt for example)


image

PR #4091, opened 2024-02-26
H-2034: Restore basic search bar to frontend

CiaranMn:

This PR restores the search bar in HASH. It searches for entities based on their semantic distance, and links results to the entity editor.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues / next steps

- Because search involves calling a Temporal workflow to create the embedding of the search text, there might be unacceptable delay introduced that could be mitigated by calling OpenAI's embedding model directly from the API. To see how this works in practice and potentially change.
- Results are not paginated, which means they might be slow to load with lots of results. H-2258
- Ideally for entities which are contained in a page (e.g. paragraphs), we would link to the containing Page rather than the entity editor. H-2258. 
- The entity editor crashes if you try and load a paragraph block into it, because the entity editor does not handle properties with an expected value of either a single value or an array of values. H-2257

## 🛡 What tests cover this?

- Manual.

## ❓ How to test this?

1. You must have rebuilt the Graph since https://github.com/hashintel/hash/commit/244f242b84842d5cce111a59fbb5a8021ff00bda (any entities unmodified prior to rebuilding will not have embeddings and won't appear in results – you can update them after rebuilding to generate an embedding for them), or run it outside of Docker
2. You must rebuild the AI Temporal worker or run it ouside of Docker, because it was missing a workflow the search query depends on
3. Try searching for something

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/e27dbd4d-c80f-42f6-a1ed-98473713363e

### DIFF
--- a/apps/hash-ai-worker-ts/src/workflows.ts
+++ b/apps/hash-ai-worker-ts/src/workflows.ts
@@ -1,5 +1,7 @@
 import type { EntityQueryCursor, Filter } from "@local/hash-graph-client";
 import type {
+  CreateEmbeddingsParams,
+  CreateEmbeddingsReturn,
   InferEntitiesCallerParams,
   InferEntitiesReturn,
 } from "@local/hash-isomorphic-utils/ai-inference-types";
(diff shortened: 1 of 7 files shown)


benwerner01 (approved):
Thanks Ciaran, the code looks good to me and is working for me locally as expected 👍

PR #4109, opened 2024-02-29
H-2290: Implement `utoipa` schema for type system URLs

TimDiekmann:

`ToSchema` is not implemented for `BaseUrl` and `VersionedUrl`. For simple types this can be avoided by explicitly specifying the value type, however, when nested this is an issue (as required for multi type entities).

## 🔍 What does this change?

- Implement `ToSchema` for URL types
- Use that definition everywhere (removing workaround code)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3744,6 +3744,7 @@ dependencies = [
  "thiserror",
  "tsify",
  "url",
+ "utoipa",
  "wasm-bindgen",
  "wasm-bindgen-test",
 ]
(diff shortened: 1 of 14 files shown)


indietyp (changes requested):
Looks good to me! but that also means we can remove them from the shared models, right? (really like that we removed one of those pesky `SHARED_` hacks)

TimDiekmann:
No, we still reference `BaseUrl` and `VersionedUrl` from other schemas such as `DataType`:
```json
"$id": {
    "$ref": "./shared.json#/definitions/VersionedUrl"
},
```

indietyp (approved):

PR #4112, opened 2024-03-01
H-2302: Add step to install missing required `yarn global dependencies`

viktor-evdokimov:

update documentation for running hash locally

While trying to follow instructions yarn install fails first on missing typescript and then on missing wasm-pack

### DIFF
--- a/apps/hash/README.md
+++ b/apps/hash/README.md
@@ -87,6 +87,11 @@ To run HASH locally, please follow these steps:
    ```sh
    yarn install
    ```
+   
+1. Install yarn global dependencies
+   ```sh
+   yarn global add typescript wasm-pack
+   ```
 
 1. Ensure Docker is running.
    If you are on Windows or macOS, you should see app icon in the system tray or the menu bar.


nonparibus:
`wasm-pack` should be installed automatically in a `postinstall` step. The command being invoked is:

> wasm-pack --version | grep -q '0.12.1' || cargo install 'wasm-pack' --version '0.12.1' --locked --force

We'll take a look at what might be going on here; thanks for opening this PR.

TimDiekmann:
Hi @viktor-evdokimov and thanks for opening this Pull Request.

Could you post the steps you did in order to install HASH? It's required to run `yarn install` from the root first. This should install all relevant packages. `yarn install` will run a post-install step which will check for the current version of `wasm-pack` and install it via `cargo` (as mentioned by @nonparibus above) in the case the version does not match the specified version or if `wasm-pack` is not installed at all. 
If `yarn install` does not install `wasm-pack` for you could you please post the output of the error?

viktor-evdokimov:
I did run `yarn install` and it was failing, until i installed two dependencies, after which it worked.

nonparibus:
> I did run `yarn install` and it was failing, until i installed two dependencies, after which it worked. And i realized i should've add it before yarn install in the doc to fix problem I had, not after

@viktor-evdokimov We've had a few attempts at replicating this without success. Please could you uninstall `wasm-pack`, clone the repo again, run `yarn install` from the root, and paste the output here?

TimDiekmann:
As `wasm-pack` is supposed to be installed through `cargo install` (the package manager for Rust) I wonder if that is missing for you. I noticed, that this is currently not listed as a dependency. Do you have Rust installed?

nonparibus:
Hi @viktor-evdokimov, any luck re: the above? We'll close this PR out in a ~week or so if we don't hear back. Thanks!

TimDiekmann:
Hi @viktor-evdokimov. As we were not able to reproduce the issue and it's likely that the issue happened due to a missing Rust installation we're going to close this PR. Please feel free to re-open it if the issue persists 🙂

PR #4116, opened 2024-03-02
H-2306: Enable `consistent-type-imports` ESLint

indietyp:

This enables the `consistent-type-imports` lint rule, which is automatically fixable, and restricts type imports to be prefixed with `type.` This is an official recommendation by Typescript because these imports are stripped during execution.

This step also makes the next step of moving from CommonJS to ES-Modules easier.

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (approved):
Thanks @indietyp!

PR #4131, opened 2024-03-07
H-2345: Pin turbo version to patch version

TimDiekmann:

Turbo just released `1.12.5` which breaks CI for us.


CiaranMn (approved):

CiaranMn:
@TimDiekmann why does it? Is there an issue?

TimDiekmann:
I think they changes the output of the dry-runs. I'll investigate that further but this way is the easiest way to quickly fix the CI.

PR #4132, opened 2024-03-07
H-2346: Add `/me` router that redirects to user's profile page

CiaranMn:

Adds a `/me` route that will redirect an authenticated user to `/@their-shortname`.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Visit `/me` while logged in

### DIFF
--- a/.github/actions/warm-up-repo/action.yml
+++ b/.github/actions/warm-up-repo/action.yml
@@ -17,7 +17,7 @@ runs:
         # cache: yarn ## Currently disabled because of frequent timeouts
 
     - name: Install WASM tools
-      uses: taiki-e/install-action@b7add58e53e52e624966da65007ce24524f3dcf3 # v2.27.9
+      uses: taiki-e/install-action@dabb9c1ee51c21c545764a0d517f069ff52e6477 # v2.28.1
       with:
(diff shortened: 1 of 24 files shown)


nonparibus:
@CiaranMn If I visit while logged in this works great.

If I visit while logged out I get taken to `/signin` rather than what I expected: `/signin?return_to=%2Fme`

But then when I actually test using the latter, while correctly redirecting to `/me` upon login, that page shows the `Profile not found` message that I think is our default 404 (as I have seen it on other made-up routes, as well).

TL;DR: looks like this works upon normal user navigation, but not after a login redirect. And we should make sure logged-out users visiting the route get redirected appropriately.

nonparibus (changes requested):
^

CiaranMn:
> If I visit while logged out I get taken to /signin rather than what I expected: /signin?return_to=%2Fme
@nonparibus I have added this to all redirects to `/signin` (it wasn't present before)

> But then when I actually test using the latter, while correctly redirecting to /me upon login, that page shows the Profile not found message that I think is our default 404 (as I have seen it on other made-up routes, as well).

This happens to me locally, we can have a look sync – it might be something to do with preview deployments / cookies

PR #4135, opened 2024-03-07
H-2209: Use Rust edition 2024

TimDiekmann:

The new Rust edition 2024 comes with some. neat new features. However, the lifetime capture rules can be quite annoying. 

## 🔍 What does this change?

- Remove `Future` and `IntoFuture` imports
- Fix lifetime captures for `-> impl Trait` syntax

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

~RustRover does not recognize `Future` and `IntoFuture` prelude imports. Until this is fixed we should not merge this PR.~


libs/@local/hash-authorization/src/backend/spicedb/api.rs
```diff
@@ -66,6 +66,9 @@ impl fmt::Display for StreamError {
 
 impl Error for StreamError {}
 
+type StreamReturn<T: DeserializeOwned, B: Serialize + Sync> =
```
indietyp, line 69: too bad that TAITs are currently the only solution to overcapturing :/

indietyp (approved):
excited to toy around with `gen` and `async gen` :P

libs/@local/hash-authorization/src/backend/spicedb/api.rs
TimDiekmann: Yup, agreed. Do you think we should add a todo in the case this can be changed? Or should we just wait and revisit the code base at that point (Use a regex for type aliases and see if it's an `impl trait` type alias?) As we added the feature in this PR it would be easy to add a comment everywhere but it's also likely that this will be missed in future work and will be forgotten in the future. I would probably just leave it without comment.

indietyp: I agree, don't think a comment here is really helpful considering that we're going to use TAITs sooner or later in more places. Once a more concise syntax is finally proposed and implemented we can grep to implement as that will likely find more than just these initial comments.

PR #4140, opened 2024-03-08
H-2205, H-2354: Implement multi-type entities in the graph

TimDiekmann:

Multi-type entities enables entities to have multiple type without creating a new type.

## 🔍 What does this change?

- Change `EntityTypeMetadata` to store a vector of type ids
- Adjust Rust code to support taking multiple types for creating/updating/validating an entity
- Adjust the Node API to rewrite requests/responds to/from the graph to 

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Querying entities with multiple entity types is not possible, yet, because structural queries do not support this.

## 🐾 Next steps

- Allow querying entities with multiple entity types: H-2351

## 🛡 What tests cover this?

- Every test which uses the Graph to deal with entities will use the newly written code


CiaranMn (approved):
🥳

PR #4163, opened 2024-03-13
H-2385: Convert Graph API entity type to Node API type in AI worker

TimDiekmann:

The `updateEntityEmbeddings` workflow may take a filter or a list of entities. Entities passed from the Graph may contain more than a single type id, so the entities has to be mapped, first.

### DIFF
--- a/apps/hash-ai-worker-ts/src/workflows.ts
+++ b/apps/hash-ai-worker-ts/src/workflows.ts
@@ -1,4 +1,5 @@
 import type { EntityQueryCursor, Filter } from "@local/hash-graph-client";
+import type { Entity as GraphApiEntity } from "@local/hash-graph-client/api";
 import type {
   CreateEmbeddingsParams,
   CreateEmbeddingsReturn,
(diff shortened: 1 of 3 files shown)


CiaranMn (approved):

CiaranMn (approved):

apps/hash-ai-worker-ts/src/workflows.ts
```diff
@@ -14,6 +15,7 @@ import type {
   EntityTypeWithMetadata,
   PropertyTypeWithMetadata,
 } from "@local/hash-subgraph";
+import { mapGraphApiEntityToEntity } from "@local/hash-subgraph/src/stdlib/subgraph/roots";
```
CiaranMn, line 18:
```suggestion
import { mapGraphApiEntityToEntity } from "@local/hash-subgraph/stdlib";
```

CiaranMn (dismissed):

CiaranMn (approved):

PR #4165, opened 2024-03-13
H-2298: Encode `graph` database permissions in Terraform

TimDiekmann:

The `entity_drafts` table wasn't accessible from the graph user in production. A `GRANT` call was able to solve this issue:

```sql
GRANT SELECT, UPDATE, INSERT ON ALL TABLES IN SCHEMA public TO graph;
```

We currently don't manage permissions for the graph database in terraform and this could be the issue as no default privileges are set.

### DIFF
--- a/infra/terraform/hash/postgres_roles/postgres_roles.tf
+++ b/infra/terraform/hash/postgres_roles/postgres_roles.tf
@@ -116,6 +116,16 @@ resource "postgresql_database" "kratos" {
   allow_connections = true
 }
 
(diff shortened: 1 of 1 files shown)


CiaranMn (commented):
Can we do the same for all tables that don't already have a `default_privileges` entry please? In case the external software creates new tables as part of some future migration.

TimDiekmann:
Good point, did it in https://github.com/hashintel/hash/pull/4165/commits/8c4921e623d39e47e08ca806b32c6fdf37d8df17

CiaranMn (approved):

PR #4166, opened 2024-03-13
H-2330, H-2337, H-2339: Fix notifications page + write test, relax rate limiting for tests

CiaranMn:

Some notifications were failing to appear on the notifications / inbox page (H-2337), and when viewing a link entity with draft left/right entities they were appearing as 'Hidden' (not discoverable by the user, H-2339).

These bugs are related to expanding our `EntityId` format to potentially include a `draftId` qualifier, so that they are now either `[webUuid]~[entityUuid]` or `[webUuid]~[entityUuid]~[draftUuid]`. 

An entity is keyed in the subgraph vertices by their `EntityId`, but when we _link_ to another entity, we only use its base `EntityId` as a reference, without any draftId. That means that when attempting to find the target of a link among subgraph vertices, we need to check for vertex ids that _start_ with the base entityId, because the subgraph may only contain `EntityId` with a `draftId` suffix. This PR uses a new canary release of `@blockprotocol/graph` that includes that fix, from the `blockprotocol` repo, and also brings in the changes required to the version of `graph` that now exists in this repo (to be consolidated fully later, once multi-type entities are introduced everywhere).

This PR also:
- adds a test to check that notifications will appear on the `/inbox` page
- makes some changes to reduce test flakiness (H-2330):
  - further relax rate limiting for auth routes in test environments
  - introduces a delay before typing in an entity property table (this was flaky, sometimes only some of the text was being captured) 
  - fixes a bug in the entity type creation test where the 'save' button was being hit twice, probably only worked sometimes before depending on timing

## 🔗 Related links

- https://github.com/blockprotocol/blockprotocol/pull/1374 – PR in `blockprotocol` repo that has the corresponding fix to `@blockprotocol/graph`

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The one this PR adds

## ❓ How to test this?

1. You can test that the bug has been fixed by running the browser plugin to create a draft entity, and checking the notif appears on the inbox page


apps/hash-api/package.json
```diff
@@ -95,7 +95,7 @@
     "ws": "8.15.1"
   },
   "devDependencies": {
-    "@blockprotocol/graph": "0.3.3",
+    "@blockprotocol/graph": "0.3.4-canary-20240313191438",
```
CiaranMn, line 98: This version will conflict with other PRs but I'll sort it out before merging – hopefully can get a proper `0.3.4` released in the meantime once the other pending PR in BP merges

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
```diff
@@ -40,6 +40,56 @@ export const getEntities = <Temporal extends boolean>(
...
+   * – check for vertices which are keyed by a qualified version of the provided entityId
+   */
+  const draftEntityIds = Object.keys(subgraph.vertices).filter((id) =>
+    id.startsWith(`${entityId}~`),
+  );
+
+  if (draftEntityIds.length > 0) {
+    /**
+     * Return a combined version of all the draft editions of this entity present in the subgraph.
+     * Return a combined object of all the draft editions of this entity present in the subgraph.
```
benwerner01, line 72: Is this duplicative?

```suggestion
     * Return a combined object of all the draft editions of this entity present in the subgraph.
```

benwerner01 (approved):
Thanks Ciaran, this is working for me locally and in testing also realised it fixes H-2339 which is assigned to me 👍

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
CiaranMn: Yes, copy/paste error – fixed in https://github.com/hashintel/hash/pull/4166/commits/ffdf60640d711be191da268b42ff6b45b7c42c6c

benwerner01 (approved):

PR #4169, opened 2024-03-13
H-2224, H-2239, H-2373: Bug fixes in entities table, redirect after signin

CiaranMn:

A few small bug fixes:
1. The entities table didn't switch back to non-file view when switching to a non-file type from a file type (via the sidebar)
2. The table wasn't sorting on properties, because the sort function tries to do `row[value]` but the properties were under `row.properties[value]`
3. The `return_to` if a user visited a page they needed to log in for wasn't taking account of dynamic paths (e.g. `/@[shortname]`

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/package.json
+++ b/apps/hash-graph/package.json
@@ -22,6 +22,7 @@
     "exe": "tsx",
     "fix:clippy": "just clippy --fix",
     "lint:clippy": "just clippy",
+    "migrate": "cargo run --bin hash-graph --all-features -- migrate --user postgres --password postgres",
     "reset-database": "httpyac send --all tests/reset-database.http",
     "test:integration": "just test-integration",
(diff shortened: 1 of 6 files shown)


nonparibus (approved):

apps/hash-graph/package.json
```diff
@@ -22,6 +22,7 @@
     "exe": "tsx",
     "fix:clippy": "just clippy --fix",
     "lint:clippy": "just clippy",
+    "migrate": "cargo run --bin hash-graph --all-features -- migrate --user postgres --password postgres",
```
benwerner01, line 25: I assume this is an intentional drive-by change

benwerner01 (approved):
Thanks Ciaran, the changes look good and are working for me 👍

PR #4180, opened 2024-03-15
H-2416: switch from `jest` to `vitest`

indietyp:

Switch all runners from `jest` to `vitest`, this has been extracted from #4117, as the integration tests are currently failing and isolating both changes seems to be a good idea anyway.

Instead of just converting those that are needed, this converts all jest invocations with vitest ones.

## 🔗 Related links

- #4117 

This PR:

The changes in this PR:

The changes in this PR:


libs/@blockprotocol/type-system/typescript/test/shared.ts
```diff
@@ -0,0 +1,22 @@
+import fs from "node:fs/promises";
```
CiaranMn, line 1: I think this will fix the lint error – you will probably need to make the same change to the two other places this appears in the repo when those packages are converted to ESM
```suggestion
import * as fs from "node:fs/promises";
```

tests/hash-backend-integration/src/tests/email.test.ts
```diff
@@ -3,6 +3,7 @@ import "./load-test-env";
 import { AwsSesEmailTransporter } from "@apps/hash-api/src/email/transporters";
 import { getAwsRegion } from "@apps/hash-api/src/lib/aws-config";
 import { getRequiredEnv } from "@local/hash-backend-utils/environment";
+import { it } from "vitest";
 
 if (process.env.HASH_DEV_INTEGRATION_EMAIL) {
   const to = process.env.HASH_DEV_INTEGRATION_EMAIL;
```
CiaranMn, line 9: We have this block down below, maybe we can get rid of it now? Or remove reference to 'jest'
```ts
} else {
  // Need at least one test otherwise jest will complain
  it("always passes", () => undefined);
}
```

CiaranMn (commented):
Thanks Bilal, makes sense to me – good to merge once linting is fixed (suggestion below)

libs/@blockprotocol/type-system/typescript/test/shared.ts
indietyp: This sadly didn't help. I am guessing might have something to do with ts node vs. web?

tests/hash-backend-integration/src/tests/email.test.ts
indietyp: `vitest` still needs at least a single test in a file, but there's a proper way to skip a test instead.

see: [ffd08ac](https://github.com/hashintel/hash/pull/4180/commits/ffd08ac15679fc16920132071c6506803ae577d4)

libs/@blockprotocol/type-system/typescript/test/shared.ts
indietyp: adding `@types/node` (which apparently wasn't yet) fixed the problem.

CiaranMn (approved):
:shipit:

PR #4182, opened 2024-03-15
H-2300, H-2343, H-2341, H-2334: move `/drafts` to `/actions`  + other related crash fixes/changes

benwerner01:

This PR:
- moves the `/drafts` page to `/actions` (H-2300)
- stops the creation of a notification when a draft entity is created in the `inferEntities` flow
- adds the `Actions`/`Notifications` sidebar items, as children of the existing `Inbox` sidebar item (H-2300)
- adds a dedicated "Actions" count/link to the header (H-2300)
- adds tooltips to the header notifications/actions counters/links
- hides the "Bulk Action" dropdown when no drafts are selected (H-2341)
- fixes a crash that previously occurred when a new draft entity was added to the drafts page while it was open, related to the subgraph that was being generated for the draft entity modal (H-2334)
- fixes a crash that previously occurred when a draft entity of a new type was added to the drafts page while it was open, related to how the filter state being generated (H-2334)
- when a draft entity of a new type is added to the drafts page, the type is checked in in the filter state as the entity would otherwise not appear

## 🔗 Related links

- H-2300
- H-2343
- H-2341
- H-2343

## 🔍 What does this change?

See description.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Go to the `/actions` page, and observe the refactored sidebar/header
2. Via the browser extension create a draft of a new type whilst the `/actions` page is open
3. Ensure that there is no crash, and the entity is immediately visible (i.e. its type is added and not filtered out in the filter sidebar)

## 📹 Demo

image


apps/hash-ai-worker-ts/src/activities/infer-entities/persist-entities.ts
```diff
@@ -221,12 +221,10 @@ export const persistEntities = async (params: {
     });
   };
 
-  const createInferredEntityNotification = async ({
+  const createInferredEntityUpdateNotification = async ({
```
CiaranMn, line 224: I thought what we wanted to do here was to retain notifications for _non-draft_ actions (since they won't appear in the drafts queue) – i.e. we check if `createAs === "live"`, and send the notification if true. Rather than disabling notifications on creation (entities can be created as non-draft).

benwerner01: Apologies, you're right - have addressed in https://github.com/hashintel/hash/pull/4182/commits/1dcd06dc661426c0821fd1a1aa57f869d8ca6cbb. Now there'll be notifications for draft/non-draft updates, and non-draft creations in the graph

CiaranMn: I don't think we want notifications for draft updates either - they will also be actions (approve or reject the update)

benwerner01: Makes sense, have done so in https://github.com/hashintel/hash/pull/4182/commits/dc641c78a04df8e021c93b368215e53d23934666

CiaranMn (approved):

PR #4183, opened 2024-03-15
H-1455: Implement partial updates for entities

TimDiekmann:

Some fields like draft (or archived) it's desirable to update only that field or only other fields. The current update mechanism does not support that to avoid data races.

Useful for various things, including draft updates ([H-1527](https://linear.app/hash/issue/H-1527/support-draft-updates-to-entities-in-the-graph)), performance, validation, correctness (if have data races).

## 🚫 Blocked by

- #4173 

## 🔍 What does this change?

- Replace the `PUT` method with `PATCH` which makes all parameters (except the Entity Id) optional and allows to partially updates properties.

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (dismissed):
Excellent!

CiaranMn (approved):

PR #4217, opened 2024-03-21
H-2436: Fix entity editor link table rendering, add test

CiaranMn:

The cells in the link table are not rendering.

It looks like this is a change in behavior in Glide Data Grid, which had a major version upgrade in #4070, whereby custom renderers are not being called if `drawCells` is present and returns `false`. Since we are not using `drawCells` for any cell in the link table I have removed it, and added a test to ensure that the cells render.

Note that because the canvas content is essentially an image, the test just checks that there is at least one non-white pixel in the space where cell content should be. It does not check what the cell content is – in order to do that we would need to know exactly where the content was, and exactly what it was, and check that the pixels from one matched the pixels from another. This would be possible by sharing the logic to draw the text in a cell and checking that the results matched, but feels like more trouble than it's worth at the moment. This test is sufficient to catch the bug the PR fixes.

To help the test, and also because it's better UX, I've made it so that we also search an entity type's description when a user types in the entity type selector (not just its title).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The one this adds.

## ❓ How to test this?

1. Check that the link table renders correctly for an entity with at least one potential outward link


benwerner01 (approved):
Thank you Ciaran, the changes make sense to me and are working in the deployment for me 👍

PR #4232, opened 2024-03-24
H-2457: Improved ESLint configuration

indietyp:

This PR adds a new ESLint configuration, which is more encompassing and stricter than the previous one.

Instead of trying to roll our own completely, needing to figure out which plugins to use and where I opted for an alternative method.

Instead of creating one from scratch, one uses an existing configuration and applies the necessary changes.

The big contenders currently around after some research are:
* xo (I have previous experience with this one)
* sheriff
* hardcore

While hardcore looked promising, it does not yet expose the flat config configuration.
The main problem with `xo` is that it is exposed through an alternative binary, making it unsuitable to use in, e.g., WebStorm. It also, by default, lints stylistic choices covered by prettier.

`sheriff` instead fully embraces ESLint flat configs, is highly configurable, and has an excellent webpage detailing all the configurations used.

This PR will be split into multiple, as some changes won't be immediately visible. This PR only changes the configuration, introducing a package `@repo/eslint` (name TBD). A follow-up PR then applies the changes to all packages.

Because styling is always very subjective, we can preview the changes *without* polluting this PR. Some stylistic changes may be contentious and need further discussion (for example: `react/jsx-curly-brace-presence`). I have already adjusted several rules, to improve it to our needs, enabling some additional ones or removing them.

I would also suggest that once in a semi-stable state one targets this PR with their new changes, so that we can test developer experience.

(I have linted my package in here, and found it quite an improvement)

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
closed in favor of https://github.com/hashintel/hash/pull/5847

PR #4267, opened 2024-04-02
H-2418: Add ability to search for types in HASH search bar

CiaranMn:

Also looks for types when the user types in the search bar.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- I noticed a mismatch between library types and HASH values, mentioned in a comment – H-2489.

## ❓ How to test this?

1. Try searching for a type

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/318ac574-2778-45a8-b777-1832c728c0a2

### DIFF
--- a/apps/hash-frontend/src/shared/layout/layout-with-header/search-bar.tsx
+++ b/apps/hash-frontend/src/shared/layout/layout-with-header/search-bar.tsx
@@ -1,11 +1,18 @@
 import { useQuery } from "@apollo/client";
+import type { EntityType } from "@blockprotocol/type-system";
 import { Chip, IconButton } from "@hashintel/design-system";
+import type { Filter } from "@local/hash-graph-client";
 import { generateEntityLabel } from "@local/hash-isomorphic-utils/generate-entity-label";
 import {
   currentTimeInstantTemporalAxes,
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
-import type { Entity, EntityRootType, Subgraph } from "@local/hash-subgraph";
+import type {
+  Entity,
+  EntityRootType,
+  EntityTypeRootType,
+  Subgraph,
+} from "@local/hash-subgraph";
 import {
   extractEntityUuidFromEntityId,
   extractOwnedByIdFromEntityId,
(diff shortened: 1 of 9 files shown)


apps/hash-frontend/src/shared/layout/layout-with-header/search-bar.tsx
```diff
@@ -120,10 +131,31 @@ const EntityResult: FunctionComponent<{
...
+        <Chip
+          color="aqua"
+          label="Entity Type"
+          sx={{ cursor: "pointer", ml: 1 }}
+        />
       </Link>
-      {entityType && (
-        <Chip color="teal" label={entityType.schema.title} sx={{ ml: 1 }} />
-      )}
     </ResultItem>
```
benwerner01, line 159: Some thoughts I had whilst testing this locally:

1. The color of the "Entity Type" copy is different to the color of the copy in the entity chips
2. Only the text itself is clickable, rather than the entire list item. We probably want the `Link` to wrap the list item in `ResultItem`, rather than be nested inside it
3. When hovering over the chip, the cursor changes from being clickable to non-clickable. We probably need to override how the cursor is being set there
4. When clicking a result, the search bar stays open. We probably want it to close, but maybe leave the search copy populated so the user can easily return to the search results


https://github.com/hashintel/hash/assets/42802102/3a3174d5-d6a4-4dbd-8d83-1fd2fd9deee6

benwerner01 (dismissed):
Thanks Ciaran, searching for entity types is working for me locally, and the changes look good to me. Had some thoughts about the UX/styling of the existing search bar, which shouldn't block merging but might be worth addressing as part of this PR.

apps/hash-frontend/src/shared/layout/layout-with-header/search-bar.tsx
CiaranMn: Thank you, addressed in https://github.com/hashintel/hash/pull/4267/commits/ac8114f81bde6b2043f3434c58a922fb8cd94013

Apart from the chip colour - I don't want to mess with them as they're also used elsewhere. Can have a proper design for this later

benwerner01 (approved):

PR #4272, opened 2024-04-03
H-2451: Partially fix entity editor behavior when clicking between cells

CiaranMn:

This fixes some issues with editing entity properties when clicking from a cell being edited into another cell.

We're currently relying on native HTML input validation attributes (e.g. `type: "url"`, `minLength: 4`), triggering validation of the form when clicked outside of it, and preventing closing the form if it's invalid.

However, the click handler is not triggered when another grid cell is clicked. The form is unmounted immediately without validation being run. In order to force validation we need to somehow change `glide-data-grid`'s behavior to stop an open editor being closed when another cell is clicked based on some conditions. H-1834 tracks that.

In the meantime, this is an interim fix for existing issues (see demos):
1. Persists the value if the form has been unmounted. Previously, switching to another cell would lose the value in the first cell (since the validation handler didn't persist the value if the form wasn't present)
2. Cleans up the handler if the form has been unmounted
3. Doesn't add the validation handler if no constraints are present on the value. We can rely on the relevant HTML input forcing the user to input a number or a string at least, so we don't need additional validation unless there are additional constraints.

This does mean that it's possible to have an invalid value in a cell, e.g. input a non-URL and then click to another cell. But we need form-level validation for other things anyway (e.g. missing required properties), which should flag this up – also H-1834.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- There is a test for editing an entity property

## ❓ How to test this?

1. See demos

### Buggy behavior

1. Clicking into another cell loses existing value
2. Once another cell is clicked into, clicking outside the grid does not close the editor, because the click handler was not cleaned up from the previous cell. You must hit enter to save

https://github.com/hashintel/hash/assets/37743469/9e980680-390e-47e5-adfb-40e1b41e1f07

### New behavior

https://github.com/hashintel/hash/assets/37743469/b4c3bde9-71df-4ae6-8018-58cfc3cbfe71

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/single-value-editor.tsx
+++ b/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/single-value-editor.tsx
@@ -153,22 +153,41 @@ export const SingleValueEditor: ValueCellEditorComponent = (props) => {
     );
   }
 
+  /**
+   * Force validation on the text input form.
+   * If the form is valid or if the form has been unmounted, allow <Grid /> to handle clicks events again.
+   */
   const validationHandler = () => {
-    if (!textInputFormRef.current) {
-      return;
+    if (textInputFormRef.current) {
+      textInputFormRef.current.requestSubmit();
+
+      if (!textInputFormRef.current.checkValidity()) {
+        return;
+      }
     }
-    textInputFormRef.current.requestSubmit();
-    if (textInputFormRef.current.checkValidity()) {
-      document.body.classList.remove(GRID_CLICK_IGNORE_CLASS);
-      onFinishedEditing(latestValueCellRef.current);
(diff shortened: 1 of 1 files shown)


apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/single-value-editor.tsx
```diff
@@ -153,22 +153,41 @@ export const SingleValueEditor: ValueCellEditorComponent = (props) => {
     );
   }
 
+  /**
+   * Force validation on the text input form.
+   * If the form is valid or if the form has been unmounted, allow <Grid /> to handle clicks events again.
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/properties-section/property-table/cells/value-cell/single-value-editor.tsx:158]:
```suggestion
   * If the form is valid or if the form has been unmounted, allow <Grid /> to handle click events again.
```

benwerner01 (dismissed):
Thanks Ciaran, the changes look good to me and are working for me 👍

benwerner01 (approved):

PR #4273, opened 2024-04-03
H-2075: Add Sentry to Temporal workers

TimDiekmann:

We want to capture exceptions which are thrown from worker activities and workflows.

## 🔍 What does this change?

- Add sinks to capture messages and exceptions. These sinks allow workflows to invoke I/O functions. The workflow information is used to populate the Sentry context.
- Add interceptors for activities and workflows to catch exceptions
- Enable both, sinks and interceptors, for the temporal workers

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (approved):
Very helpful addition, thank you @TimDiekmann!

PR #4277, opened 2024-04-04
H-2485: Add `getFileFromUrl` Flow action

CiaranMn:

Adds an action to get a file from a URL, store it in S3 and create a file entity associated with it.

## 🔍 What does this change?

- To avoid the hassle of having file uploads go from Temporal via the Node API, this relies on an S3-compatible object storage Temporal can upload to directly. This involves:
  - adding `minio` so we can use S3-compatible storage locally
  - moving some file-related logic from the Node API into a shared location

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- There are some Flow-wide things we need to sort out such as 
  - being able to specify an `ownedById` for entities created during the flow
  - having a shared `logger` among all the flows

## 🛡 What tests cover this?

- None

## ❓ How to test this?

1. Run the Node API
2. Run or rebuild the AI worker

```graphql
mutation {
  startFlow(
    flowTrigger: {
      triggerDefinitionId: "userTrigger"
      outputs: [
        {
          outputName: "url",
          payload: { 
            value: "https://www.simplilearn.com/ice9/free_resources_article_thumb/what_is_image_Processing.jpg"
          }
        }
      ]
    }
    flowDefinition: {
      name: "Save File From Url"
      flowDefinitionId: "saveFileFromUrl"
      trigger: {
        triggerDefinitionId: "userTrigger"
        kind: "trigger"
        outputs: [
          { payloadKind: "Text", name: "url", array: false }
          { payloadKind: "Text", name: "description", array: false }
          { payloadKind: "Text", name: "displayName", array: false }
        ]
      }
      steps: [
        {
          stepId: "1"
          kind: "action"
          actionDefinitionId: "getFileFromUrl"
          inputSources: [
            {
              inputName: "url"
              kind: "step-output"
              sourceStepId: "trigger"
              sourceStepOutputName: "url"
            }
            {
              inputName: "description"
              kind: "step-output"
              sourceStepId: "trigger"
              sourceStepOutputName: "description"
            }
            {
              inputName: "displayName"
              kind: "step-output"
              sourceStepId: "trigger"
              sourceStepOutputName: "displayName"
            }
          ]
        }
      ]
      outputs: [
        {
          stepId: "1"
          stepOutputName: "fileEntity"
          name: "fileEntity"
          payloadKind: "Entity"
          array: false
        }
      ]
    }
  )
}

```


apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
```diff
@@ -112,7 +112,7 @@ export const runFlowWorkflow = async (
   const flow = initializeFlow({
     flowDefinition,
     flowTrigger,
-    flowId: workflowId as EntityUuid,
+    flowId: workflowId as EntityId,
```
benwerner01, line 115: Is this an `EntityId`? In the `startFlow` resolver we're setting this via `generateUuid()` (i.e. without the `ownedById`).

Also this is something we can address in follow-up, but just want to flag the types here are currently incorrect.

libs/@local/hash-backend-utils/src/file-storage/file-properties.ts
benwerner01, line 1: Did you intend to commit this?

benwerner01 (commented):
Thanks Ciaran, the code looks good to me.

I'm having difficulty testing this, because there are TS errors in the AI worker which are preventing me from being able to run `yarn external-services up`, which I'm assuming is required to get the minio container configured correctly.

apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
CiaranMn: I've reverted this

libs/@local/hash-backend-utils/src/file-storage/file-properties.ts
CiaranMn: Nope, thanks! Deleted

benwerner01 (dismissed):

PR #4283, opened 2024-04-06
H-2504: `harpc` wire protocol

indietyp:

This PR implements the wire protocol, the lowest layer protocol, which enables streaming of data and is (on purpose) minimal.

This isn't simply a re-implementation of the experiment wire protocol (previously known as transport protocol) but instead an improvement, notable improvements include:

* `magic` (every message is now prefixed `harpc`), like other binary protocols this allows us to recognize a message
* `service version`: service version is now made up of a `major: u8, minor: u8`
* ~~`encoding, accept`: clients can now request specific encodings, and encodings used are now explicit rather than implicit.~~
* ~~`authorization` is now optional on a request message, and can be sent at the start of a session to configure an authorized user (see the SVG in the docs)~~ `authorization` has been delegated to a built-in service instead
* ~~`request id`: allow multiple requests to use the same session/channel to send multiple requests simultaneously by associating a request id to every request~~
* `varint` removed, `variable integers` (zig zag) encoding were used for numbers previously, this lead to some more complicated code and variable sized package lengths, integers are now fixed width instead.
* `payload` now has a size limit of `64KiB`, comes with native streaming support, and allows us to clamp the size to u16.

What took a bit longer here than anticipated was that the wire protocol now has an extensive test suite of **120** tests to ensure that the encoding <-> decoding is the same and as expected, with property-based fuzzing tests to ensure this even more.

This PR:

The changes in this PR:

The changes in this PR:

~~Need advice as to how to properly configure `turborepo`~~ resolved through #4338


libs/@local/harpc/crates/wire/src/request/authorization.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use error_stack::Result;
+use graph_types::account::AccountId;
+use tokio::io::{AsyncRead, AsyncWrite};
+
+use crate::codec::{DecodePure, Encode};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(test, derive(test_strategy::Arbitrary))]
+pub struct Authorization {
```
TimDiekmann, line 11: We could move this struct to a more sensible place (another crate) and use it more consistently. Nothing we require to do in this PR, though.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// * Magic (5 bytes)
+/// * Protocol Version (1 byte)
+/// * Request Id (2 bytes)
+/// * Flags (1 byte)
+/// * Service Id (2 bytes)
+/// * Service Version (2 bytes)
+/// * Procedure Id (2 bytes)
+/// * Encoding (2 bytes)
+/// * Accept (2 bytes)
+/// * Authorization (16 bytes)
```
TimDiekmann, line 57: Worth flagging that authorization may not fit into 16 bytes in the future. We can assume this for now but I'm not sure if we can keep this constraint. However, it could be possible we will continue using a single UUID but not the account ID but a generated token.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// * Payload Length (2 bytes)
```
TimDiekmann, line 58: 2 bytes for the payload lengths seems to be too few. I would assume _at least_ 4 bytes if not 8 bytes.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+/// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+/// |  Magic  |P|R. |F|S. |S. |P. |E. |A. |      Authorization      |
+/// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+/// |     |P. |                       Payload                       |
+/// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+///
+/// * Magic (5 bytes)
+/// * Protocol Version (1 byte)
+/// * Request Id (2 bytes)
```
TimDiekmann, line 50: It would be good if every request ID is unique. For that, 2 bytes are not enough.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// * Flags (1 byte)
+/// * Service Id (2 bytes)
+/// * Service Version (2 bytes)
+/// * Procedure Id (2 bytes)
+/// * Encoding (2 bytes)
+/// * Accept (2 bytes)
+/// * Authorization (16 bytes)
+/// * Payload Length (2 bytes)
+/// * Payload (27 bytes)
```
TimDiekmann, line 59: If the above suggestions would be applied, the payload would be 11 bytes (16 bytes request id, no encoding, 8 bytes payload length), which leads to the question if the `begin` frame should contain a payload at all. This is also questionable as the payload would be misaligned which will have an impact on performance.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// total 64 bytes
+/// ```
+///
+/// The payload is of variable size and specified by the `Payload Length` field. `Authorization` is
+/// optional and only present if the `ContainsAuthorization` bit in `Flags` is set.
+/// Packets need to set the `BeginOfRequest` bit in the `Flags` field.
+///
+/// # `Frame` Packet
+///
+/// The layout of a `Frame` packet is as follows:
```
TimDiekmann, line 69: 18% of each frame is metadata, for big payloads this seems to be a lot. Also, the payload is not aligned.

libs/@local/harpc/crates/wire/src/request/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+///
+/// * Magic (5 bytes)
+/// * Protocol Version (1 byte)
+/// * Request Id (2 bytes)
+/// * Flags (1 byte)
+/// * Service Id (2 bytes)
+/// * Service Version (2 bytes)
+/// * Procedure Id (2 bytes)
+/// * Encoding (2 bytes)
+/// * Accept (2 bytes)
```
TimDiekmann, line 56: Why do we need to pass the encoding? The server is responsible to encode the data and the client is responsible to decode the data, they should be in agreement of the encoding.

libs/@local/harpc/crates/types/src/service.rs
```diff
@@ -0,0 +1,55 @@
...
+    #[must_use]
+    pub const fn is_reserved(self) -> bool {
+        // 0xFxxx are reserved for internal use
+        self.0 & 0xF000 == 0xF000
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+pub struct ServiceVersion(Version);
```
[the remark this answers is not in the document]
indietyp: This is a remnant of my refactoring during initial development and will merge. I had multiple instances of `Version` planned, which I did not end up needing.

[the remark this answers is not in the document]
indietyp: Will remove.

Cargo.toml
```diff
@@ -16,6 +16,8 @@ members = [
     "libs/@local/temporal-versioning",
     "tests/hash-graph-integration",
     "tests/hash-graph-test-data/rust",
+    "libs/@local/harpc",
+    "libs/@local/harpc/crates/*"
```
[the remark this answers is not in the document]
indietyp: I can also call it `packages`, as I plan to put the library into multiple crates my idea was to have two directories:

`packages/` for js and ts and `crates/` for the rust "core"

libs/@local/harpc/crates/wire/src/codec/types.rs
```diff
@@ -0,0 +1,139 @@
...
+};
+use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
+
+use crate::codec::{DecodePure, Encode};
+
+impl Encode for Version {
+    type Error = io::Error;
+
+    async fn encode(&self, mut write: impl AsyncWrite + Unpin + Send) -> Result<(), Self::Error> {
+        write.write_u8(self.major).await.map_err(Report::from)?;
```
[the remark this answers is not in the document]
indietyp: It isn't, must've missed. Great catch!

libs/@local/harpc/crates/wire/src/codec/decode.rs
```diff
@@ -0,0 +1,173 @@
+use std::io;
+
+use bytes::Bytes;
+use error_stack::{Context, Result};
+use graph_types::account::AccountId;
+use tokio::io::{AsyncRead, AsyncReadExt};
+use uuid::Uuid;
+
+pub trait DecodePure: Send + Sync + Sized {
```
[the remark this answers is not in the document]
indietyp: It doesn't take a context, tbh only reason it is named `Pure` is because I couldn't come up with a better name. Any suggestion is very much appreciated!

Cargo.toml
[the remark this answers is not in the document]
TimDiekmann: in `packages` we would also have a potential Python client I'd assume? I wonder if we just want to flatten it:
- `libs/@local/harpc/wire-protocol/`
- `libs/@local/harpc/.../`
- `libs/@local/harpc/client/rust`
- `libs/@local/harpc/client/node`
- ...

Basically grouping everything under the `harps` directory and keep the rest consistent with the name schema we already have.

libs/@local/harpc/crates/wire/src/codec/decode.rs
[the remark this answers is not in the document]
TimDiekmann: Can't we just use `Decode`? 😄

libs/@local/harpc/crates/wire/src/encoding.rs
```diff
@@ -0,0 +1,147 @@
...
+// but instead a `ItemOrVec` that's encoded, which is a bit more messy.
+#[enumflags2::bitflags]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(test, derive(test_strategy::Arbitrary))]
+#[repr(u16)]
+pub enum Encoding {
+    Raw = 1 << 0,
+    Json = 1 << 1,
+    Cbor = 1 << 2,
+}
```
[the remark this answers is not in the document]
indietyp: This is because of the `accept` header to address a concern that was raised back in January. The attempt here is to simulate the `Content-Type` and `Accept` header, so that you can be sure that the encoding you get is the one you expected.

I am leaning towards removing this entirely and just say: we know what the encoding will be

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: I have been thinking of moving authorization entirely out of the header instead of relying on the new session mechanism. Basically, you call `0xF0`, and `0x00` to authenticate (`0xFx` is reserved for internal communication) to authenticate a session, and then that's that, instead of needing an authentication on every request. This would also make the packet smaller, albeit increase implementation complexity for authentication.

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: This is to limit to 64KiB, if more than `64KiB` are used you simply send another packet (the frame packet) to continue the payload. Basically, if you have 190KiB of data, you'd send:

1. `RequestBegin` (`64KiB`)
2. `RequestFrame` (`64KiB`)
3. `RequestFrame` (`62KiB`)

We *could* ofc also increase this to an `u64` instead of `u16`.

This was inspired by TCP, which also has a `64KiB` packet size limit.

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: The idea is that the request id is unique *in flight* per session and only used to associate a request with a response. This would only pose a problem if we had more than 65.536 in-flight requests in a single session. Like TCP, this is a rolling number, so once 65.536 is reached, we go back down to 0.

The difference is that TCP uses 32bit instead of 16bit, but the difference is that TCP increments this per packet sent, while we only increment per request sent. So I thought this would be sufficient.

Cargo.toml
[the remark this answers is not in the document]
indietyp: Sure we can do that, my plan was to have an overarching `harpc` crate, but I can just rename it to `client-rust`, you are right.

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: Ah, I thought this would be the total payload size, not the one in the current paket 🙂

libs/@local/harpc/crates/wire/src/encoding.rs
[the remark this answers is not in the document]
TimDiekmann: Let's remove it 🚀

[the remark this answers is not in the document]
TimDiekmann: IIRC in January we decided to start with JSON as encoding, but I don't think this has to be specified, this would just be an implementation detail

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: I think it's fine to move authorization out. I'm not sure about authentication but I think it shouldn't be in the header as well as the size may vary between applications (and it's wasted space if not required). We should chose the more performant way, adding a bit of complexity to the client is fine IMO.

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: I wonder if it would be useful to keep a global request ID for provenance. This ID could be logged e.g. in Sentry. Having a built-in functionality to track a request might be helpful, but I don't feel strongly.

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: sounds good, in that case we'll designate a service in the reserved `0xF` range for authentication and authorization during a session 👍 (it also brings down the header size which is always nice)

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: 65.536 sounds like a lot of in-flight requests and the bottleneck would probably somewhere else, my instinct is that we chose 4 bytes anyway.
In any case we would need a free-list (or an "in-use-list") to find the next free ID as soon as it wraps but that's an implementation detail.

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: As I misunderstood that the payload could be dynamic this comment is obsolete

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: Going to remove encoding, see above.

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: Just to understand: Authorization in this context means "The user is allowed to call the endpoint at all", right?

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: We could designate a global request ID on the server for this (which we could return on response), but this request ID is chosen by the client in the request, so it isn't guaranteed to be global. However, we can hand out and designate global IDs from the server, yes.

but I think both that global request id as well as tracing sentry support might be something that's better implemented on the higher abstraction of the payload instead. What do you think? (So that the response/request gain a field for sentry tracing and the response returns a global request id)

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: I see now that me generating the illustration was misleading, the payload can be up to 64KiB in size, for every packet (payload length only denotes the length of the payload, not the whole response) I will reword this.

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: will be removed

libs/@local/harpc/crates/wire/src/request/mod.rs
TimDiekmann: Unless we increment a global number (-> impossible) we cannot guarantee uniqueness, even with UUIDs. I agree that a tracing identifier is better suited on a higher abstraction layer. The wire protocol should contain the absolute minimum.

How the client would know the request ID before a response of the server is returned?

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: Yes, the idea is simple, in more conceptual terms you call:

```
authenticate(accountId: Uuid);
```

At the beginning of the session. Then every request after that is associated with that account (calling `authenticate` multiple times in the same session or not at the start will result in an error).

Endpoints then *authorize* against the account we provided in the session, denying or allowing the request just like we're doing right now. So nothing really changed there, if that makes sense.

TimDiekmann: Currently, we invoke an external service (SpiceDB) and decide on a per-request basis if something is allowed or not. This highly depends on the parameters passed in (A user might update entity A but not entity B). Also, account ID is only used because the graph is not exposed to user-facing code (the frontend cannot file requests) but it's abstracted away. Otherwise it would be trivial to use a more powerful account ID. At some point this needs to be changed so a more opaque ID is used (e.g. a token).

If this is moved to the `0xF000` interface I guess the size of the authentication parameters would be more dynamic (similar to the payload?), so we would not limit ourselves to 16 bytes?

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: > In any case we would need a free-list (or an "in-use-list") to find the next free ID as soon as it wraps but that's an implementation detail.

Not necessarily; we could be like TCP and assume that once we overflow, assume we're in unused territory. going with 4 bytes to future-proof is a good idea. Especially w/ 2^32 possibilities that's a pretty safe assumption and removes a lot of implementation complexity.

The request id is chosen by the client, and is more like a ticket than anything else, it's basically a: `hey this request is known as A`, then the server is done. This allows for multiple requests on the same band, while not relying on the fact that a will be done after b, the server then replies with: `this is the response to A`. This is commonly used in all sorts of protocols, even JSONRPC.

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: Yes, this would still apply, the code for authorization would be still on per RPC basis, but the idea is that the `accountId` is the same (or an opaque token) across all RPC method calls in that particular session (see the SVG).

it boils down to:


```
pub struct Authentication {
	account: AccountId
}

fn authenticate(session: &Session, authorization: Authorization) -> Result<(), AuthenticationError> {
	// ... e.g. pre-fetch all permissions, or simply do:
	session.insert(account);
}

// then in a "normal" RPC method:

fn do_thing(session: &Session, args: Args) -> Result<(), ThingError> {
	let account = session.get_or_err(account).change_context(ThingError)?;

	// ... check that the user is authorized
	if !account_is_authorized(account) {
		return Err(Report::new(Unauthorized(account)));
	}

	Ok(());
}
```


The key is that this is per session, a longer-lived connection used to do multiple requests with a single user. If a user changes, one creates a new session (a new connection).

libs/@local/harpc/crates/wire/src/codec/encode.rs
```diff
@@ -0,0 +1,159 @@
...
+    Io,
+    #[error("buffer exceeds 64KiB")]
+    TooLarge,
+}
+
+impl Encode for Bytes {
+    type Error = BytesEncodeError;
+
+    async fn encode(&self, mut write: impl AsyncWrite + Unpin + Send) -> Result<(), Self::Error> {
+        // write the length in u16 (this is ok because we never send more than 64KiB).
```
TimDiekmann, line 62:
```suggestion
        // write the length in u16 (this is ok because we never send more than 64 KiB).
```

libs/@local/harpc/crates/wire/src/encoding.rs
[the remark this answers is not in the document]
indietyp: removed

libs/@local/harpc/crates/wire/src/request/mod.rs
indietyp: illustration has been updated

Cargo.toml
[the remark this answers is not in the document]
indietyp: crates have been hoisted as discussed

libs/@local/harpc/crates/types/src/service.rs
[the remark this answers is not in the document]
indietyp: has been removed

libs/@local/harpc/crates/types/src/version.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+pub struct Version {
```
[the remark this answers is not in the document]
indietyp: has been implemented

libs/@local/harpc/crates/types/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "harpc-types"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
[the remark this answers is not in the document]
indietyp: no, proptest is needed in `wire-protocol` so needs to be a feature and optional dependency instead

apps/hash-graph/libs/query-builder-derive/Cargo.toml
```diff
@@ -2,8 +2,9 @@ cargo-features = ["edition2024"]
 
 [package]
 name = "query-builder-derive"
-version = "0.1.0"
+version = "0.0.0"
```
TimDiekmann, line 5: Thank you for the drive-by!

apps/hashdotai/glossary/agent-based-modeling.mdx
TimDiekmann, line 1: What did pick up the formatting changes in this folder?

libs/@local/harpc/types/src/version.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub major: u8,
+    pub minor: u8,
+}
+
+impl Version {
+    #[must_use]
+    pub const fn new(major: u8, minor: u8) -> Self {
+        Self { major, minor }
+    }
+}
```
TimDiekmann, line 15: Fields are public, let's not add a `new` method
```suggestion
```

TimDiekmann (commented):
Looks good to me. Small thing I noticed

apps/hashdotai/glossary/agent-based-modeling.mdx
indietyp: I have no idea, nothing should've, and I don't remember even running prettier.

TimDiekmann: Maybe some RustRover "magic"

TimDiekmann: Let's revert it to keep things clean 🙂

libs/@local/harpc/types/src/version.rs
indietyp: removed in [ded6b3c](https://github.com/hashintel/hash/pull/4283/commits/ded6b3c329c577706ba0c28459744661636d4eee)

TimDiekmann (approved):

nonparibus (dismissed):
Legal ✅

apps/hashdotai/glossary/agent-based-modeling.mdx
nonparibus: I don't particularly mind us leaving this in. Let's get `harpc` merged 🎉 Awesome work, @indietyp

TimDiekmann: Already discussed offline that we're leaving it in 🙂

indietyp:
needed to remove `generic_nonzero` feature, because toolchain update was before this PR

nonparibus (dismissed):

nonparibus:
Can we address the failing [Dependency Review](https://github.com/hashintel/hash/actions/runs/8878986546/job/24375718113?pr=4283) check in follow-up?

libs/@local/harpc/wire-protocol/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "fix:clippy": "just clippy --fix",
+    "lint:clippy": "just clippy",
+    "test:unit": "cargo hack nextest run --feature-powerset --all-targets && cargo test --all-features --doc"
+  },
+  "dependencies": {
+    "@rust/harpc-types": "0.0.0-private"
+  },
+  "devDependencies": {
+    "@rust/harpc-types": "0.0.0-private"
+  }
```
TimDiekmann, line 16:
```suggestion
  }
```

TimDiekmann (approved):

nonparibus (approved):

PR #4294, opened 2024-04-09
H-2465: Support confidence values for entities, links, and properties

TimDiekmann:

This PR adds the possibility to specify a confidence value for:
- The whole entity ("How confident is the user this entity is a thing?")
- Each property ("How confident is the user a specific property exists/is correct?")
- (for links): The left and right entity ("How confident is the user the left/right entity is correct?")

## 🔍 What does this change?

- Allow specifying the confidence values for the above when creating and updating an entity
- Property confidence values are encoded as property path (which currently is a JSON Pointer but this will change with H-2503)
- Confidence values can be set for each property and each property nested inside of another property (object or array)
- If a confidence value is updated, each nested property confidence is reset
- Each property confidence needs to correspond to an actual property (They are validated against the entity's properties, not against the schema)
- When patching an entity, the confidence value has to be specified (Patching an entity will not retain the edition confidence value)
- Link confidences has to be specified on construction. This can be lifted in the future when links can be updated. If required, we can allow updating link confidences inside of `PATCH`

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- A bunch of test cases were added to test this functionality

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1014,7 +1014,6 @@ dependencies = [
  "graph-test-data",
  "graph-types",
  "hash-status",
- "json-patch",
  "mime",
  "postgres-types",
  "refinery",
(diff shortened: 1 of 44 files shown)


CiaranMn (approved):
Excellent, let's go! 🚀

PR #4303, opened 2024-04-12
H-2513: Allow validating an entity without checking linkData

TimDiekmann:

In some circumstances it's desired to only check the entity part, not the link part. This adds the support for this.

## 🔍 What does this change?

- Remove the `ValidationProfile`
- Adds a `ValidateEntityComponents` object which determines what components of an entity has to be validated. Currently, three components can be tweaked:
	- `required_properties`: Check if all required properties are present
	- `num_items`: Check if the number in arrays are correct
	- `link_data`: Check links
By default (or when omitting the parameter) everything will be checked.

The previous `ValidateProfile::Draft` corresponds to `{ required_properties: false, num_items: false }`.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/infer-entities/propose-entities.ts
+++ b/apps/hash-ai-worker-ts/src/activities/infer-entities/propose-entities.ts
@@ -371,7 +371,10 @@ export const proposeEntities = async (params: {
                       try {
                         await graphApiClient.validateEntity(validationActorId, {
                           entityTypes: [entityTypeId],
-                          profile: "draft",
+                          components: {
+                            numItems: false,
+                            requiredProperties: false,
+                          },
                           properties: proposedEntityOfType.properties ?? {},
                         });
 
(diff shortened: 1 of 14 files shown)


apps/hash-ai-worker-ts/src/activities/infer-entities/propose-entities.ts
```diff
@@ -371,7 +371,10 @@ export const proposeEntities = async (params: {
                       try {
                         await graphApiClient.validateEntity(validationActorId, {
                           entityTypes: [entityTypeId],
-                          profile: "draft",
+                          components: {
+                            numItems: false,
```
CiaranMn, line 375: This fixes the issue where links always fail validation at this stage (since we don't have the right/left entities saved to the db)
```suggestion
                            linkData: false,
                            numItems: false,
```

CiaranMn (commented):
Thanks @TimDiekmann! I've added the place we need to turn off `linkData` to save doing it in a separate PR

apps/hash-ai-worker-ts/src/activities/infer-entities/propose-entities.ts
TimDiekmann: Great, thank you!

CiaranMn (approved):

PR #4304, opened 2024-04-12
H-2462, H-2464: Allow storing provenance data

TimDiekmann:

The provenance data required is more complex than a few fields. Some are complex objects and/or arrays. This prepares the backend to store arbitrary typed provenance data in the graph.

## 🚫 Blocked by

- #4294 

## 🔍 What does this change?

- Store all provenance data as JSONB
- Add `archived_by_id` provenance data for entities
- Remove old provenance data query paths
- Add a new query path `provenance` and `editionProvenance` with the ability to look into that object (e.g. `["editionProvenance", "createdById"]` (see change in TS for usage)
- Adjust Node API to optionally provide provenance data (empty object for now)

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

The provenance data is empty for now. A follow-up PR will be filed in the next hours to provide a provenance type. This makes reviewing it very easy.


apps/hash-api/src/graph/ontology/primitive/data-type.ts
```diff
@@ -8,6 +8,7 @@ import type {
   ModifyRelationshipOperation,
   OntologyTemporalMetadata,
   UnarchiveDataTypeParams,
+  UserOntologyEditionProvenanceMetadata,
```
CiaranMn, line 11: We don't really use `edition` to refer to ontology versions anywhere else... is it called this so that the structural query path is the same? Fine to leave it consistent if it has benefits in querying. If it was just naming I would probably go for `Version`

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -10,6 +10,7 @@ import type {
   GraphResolveDepths,
   ModifyRelationshipOperation,
   PropertyConfidence,
+  UserEntityEditionProvenanceMetadata,
```
CiaranMn, line 13: I'm not sure about `User` / `user_defined`... in practice a lot of these fields will be provided by other machine processes. I appreciate the distinction between "Things the Graph sets" and "Things it is provided with", but I think `User` suggests too strongly that it is a human user rather than a caller of the Graph API (which might be some other server) that determines it.

Don't have any great alternative names... `Variable`, `External`... or possibly don't have a distinction in the API interface between `user_defined` and the createdX fields, and instead have a struct which excludes the Graph-determined fields for external input? 

We don't have to get this perfect at first since it's all the same thing in the db, but I'm not a fan of `User` here

CiaranMn (commented):
Thanks @TimDiekmann, just a couple of naming questions/comments

apps/hash-api/src/graph/ontology/primitive/data-type.ts
TimDiekmann: I went for `edition` as we had `editionCreatedById` and `editionArchivedById` as query path for ontology previously. Also, `edition` here refers to the temporal axes of ontology types as `editionCreatedById` will also be set after unarchival. Each version have one or more editions, depending if it was archived or not.

apps/hash-api/src/graph/knowledge/primitive/entity.ts
TimDiekmann: I agree that `User` is not great. What about `Provided` instead?

apps/hash-api/src/graph/ontology/primitive/data-type.ts
CiaranMn: Fair enough, makes sense!

apps/hash-api/src/graph/knowledge/primitive/entity.ts
CiaranMn: `Provided` works, let's do it

TimDiekmann: Done in [`9b2ca54` (#4304)](https://github.com/hashintel/hash/pull/4304/commits/9b2ca547e1c4c2c37c2a265343ffb60ba367f888)

CiaranMn (approved):
:shipit:

PR #4323, opened 2024-04-17
H-2519: Add Slack notification if merge group job is failing

TimDiekmann:

It's hard to see if a PR is removed from the merge group due to failing jobs.

## 🔍 What does this change?

Add a step to the workflows which is triggered in a merge group when some of the job failed. This will send a notification to slack.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This does not cover jobs outside of the three workflows (e.g. Vercel builds), nor it will send a notification if a merge conflict happens, but having a notification in the case any of the three common workflows fails is better than not having a notification at all.


.github/workflows/deploy.yml
```diff
@@ -98,3 +98,15 @@ jobs:
       - name: Check sourcemaps
         run: |
           [[ ${{ needs.sourcemaps.result }} =~ success|skipped ]]
+
+      - name: Notify Slack on failure in merge queue
+        uses: rtCamp/action-slack-notify@19c86e40d4bc3be986b18f1459d4549536887065
+        if: ${{ failure() && github.event_name == 'merge_group' }}
+        env:
+          SLACK_LINK_NAMES: true
+          SLACK_MESSAGE: "At least one deployment job failed for a Pull Request in the Merge Queue failed <@U0143NL4GMP> <@U027NPY8Y3X> <@U02NLJY0FGX>"
```
nonparibus, line 107: There's no way to tag the correct person/group on a per-assignee or merge job creator basis? Wondering if this might get spammy (e.g. for overnight Renovate PRs that are NGMI).

CiaranMn (approved):

CiaranMn (approved):

PR #4306, opened 2024-04-12
H-2463, H-2499: Allow passing additional provenance data

TimDiekmann:
## 🌟 What is the purpose of this PR?

Makes it possible to provide custom provenance data for entities

## 🔍 What does this change?

- Populates the `provenance` field on entities, please see the changed files for more details,

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

PR #4327, opened 2024-04-18
H-2503: Use arrays instead of JSON pointer for property paths

TimDiekmann:

In structural queries, we use an array syntax of path elements into an object. To make it consistent and to avoid encoding/decoding JSON pointers, this syntax should be used for all property paths (e.g. property confidences, or PATCH operations)

## 🚫 Blocked by

- #4306 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2188,9 +2188,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.80"
+version = "1.0.81"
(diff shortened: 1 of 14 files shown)


CiaranMn (approved):
Thanks @TimDiekmann, much easier

CiaranMn (dismissed):

CiaranMn (approved):

CiaranMn (approved):

PR #4338, opened 2024-04-21
H-2535: Syncing between `Cargo.toml` and `package.json` files.

indietyp:

While creating the `harpc` wire protocol #4283 I was a bit annoyed about the error-prone nature of defining a `package.json` file and keeping that file up to date with all the dependencies.

So I thought: hey isn't this something one can automate. Introducing: this very PR.

It adds a new script (`sync-turborepo.rs`), which takes the settings from the Rust `Cargo.toml` and writes them to the `package.json`. This is as close as I could get to native Rust support.

To determine package names easier and consistent across `package.json` and `Cargo.toml` boundaries, I introduced the `@rust` namespace in `package.json`. This might be a bit contentious, but it can be changed.

## 🔍 What does this change?

* `package.json` files are being derived from `Cargo.toml`
* use of the `cargo script` native feature instead of the package

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

* To keep the name of crates in line with packages, `@rust` has been introduced as a namespace

## 🐾 Next steps

* Integration into CI
* prettier on changed files

## ❓ How to test this?

See 5eeec0a879d4852aade937cb2333693c615a42f3 for the output.

### DIFF
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -57,8 +57,18 @@ impl<'a> WorkspaceMember<'a> {
         format!("@rust/{}", self.package.name)
     }
 
+    fn package_version(&self) -> String {
+        let version = self.package.version.to_string();
+
+        if self.is_private() {
+            return format!("{version}-private");
+        }
+
+        version
+    }
+
     fn dependency_declaration(&self) -> (String, String) {
-        (self.package_name(), self.package.version.to_string())
+        (self.package_name(), self.package_version())
     }
 
     fn package_dependencies(&self) -> BTreeMap<String, String> {
@@ -86,6 +96,10 @@ impl<'a> WorkspaceMember<'a> {
 
     fn is_ignored(&self) -> bool {
         self.is_blockprotocol()
+    }
+
+    fn is_private(&self) -> bool {
+        self.package.publish.as_ref().map_or(false, |registries| registries.is_empty())
     }
 
     // first find the package.json file in the package
@@ -118,13 +132,12 @@ impl<'a> WorkspaceMember<'a> {
         }
 
         // set the version of the package.json file to the version of the package
-        package_json.version = Some(self.package.version.to_string());
+        package_json.version = Some(self.package_version());
 
         // set the package to private if the package is private
-        let is_private = self.package.publish.as_ref().map_or(false, |registries| registries.is_empty());
         package_json
             .other_fields
-            .insert("private".to_string(), is_private.into());
+            .insert("private".to_string(), self.is_private().into());
 
         // set the name of the package.json
         package_json.name = Some(self.package_name());
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -46,12 +46,12 @@ impl<'a> WorkspaceMember<'a> {
         self.package
             .manifest_path
             .components()
-            .any(|component| component == "@blockprotocol")
+            .any(|component| component.as_str() == "@blockprotocol")
     }
 
     fn package_name(&self) -> String {
         if self.is_blockprotocol() {
-            return format!("@blockprotocol/{}-rust", self.package.name);
+            return format!("@blockprotocol/{}-rs", self.package.name);
         }
 
         format!("@rust/{}", self.package.name)
@@ -112,7 +112,7 @@ impl<'a> WorkspaceMember<'a> {
             PackageJson::default()
         };
 
-        if self.is_ignored(&package_json) {
+        if self.is_ignored() {
             eprintln!("package.json in {} is ignored", path.display());
             return;
         }
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -19,7 +19,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use cargo_metadata::{DependencyKind, Metadata, MetadataCommand, Package, PackageId, Source};
+use cargo_metadata::{DependencyKind, Metadata, MetadataCommand, Package, PackageId};
 use nodejs_package_json::PackageJson;
 
 // take the metadata of the current package (indicated by the first argument)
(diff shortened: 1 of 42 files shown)


TimDiekmann (commented):
It's pretty nice to have the dependencies being automated. I, however, am not sure if we should rename the dependencies from `@local/...` to `@rust/...`, but I don't dislike it. I defer to @CiaranMn for this. Also, I'm unsure about adding new `package.json` for projects which meant to be part of `apps/hash-graph`. The plan is to move out these crates anyway so I don't feel strongly about this and is probably easier to just keep them.

indietyp:
I've mainly added them because it makes it easier to define the dependencies that way. You still need to manually add them to the `package.json`, but dependencies between the packages will be tracked correctly. So it's a trade-off I guess.

CiaranMn (approved):
> I, however, am not sure if we should rename the dependencies from @local/... to @rust/..., but I don't dislike it. I defer to @CiaranMn for this

I don't think we should rename these packages in isolation. The package names currently follow the folder structure, and this would break it, as well as introducing other inconsistencies between these and other packages (some packages named after their language, some not).

I think we should only change them as part of some wider restructuring of how we rename packages, but I'm not sure there's a particular need to take that on at the moment, unless the current format is causing problems that some better structure will solve

CiaranMn (changes requested):
Meant https://github.com/hashintel/hash/pull/4338#pullrequestreview-2014218372 to just be a comment, sorry 😓

indietyp:
> > I, however, am not sure if we should rename the dependencies from @local/... to @rust/..., but I don't dislike it. I defer to @CiaranMn for this
> 
> I don't think we should rename these packages in isolation. The package names currently follow the folder structure, and this would break it, as well as introducing other inconsistencies between these and other packages (some packages named after their language, some not).
> 
> I think we should only change them as part of some wider restructuring of how we rename packages, but I'm not sure there's a particular need to take that on at the moment, unless the current format is causing problems that some better structure will solve

But these packages/crates are already not wholly consistent. We have some packages that do not follow the naming scheme of `-rs` (see `hash-graph`), some have `-rust` as suffix and some have `-rs` as suffix. This introduces greater consistency by saying: the name in a cargo workspace must be unique (this is something we can guarantee), but we cannot guarantee that any name in `@local`, `@libs` or `@apps` is unique (if derived from the rust package name). Therefore, adding the namespace `@rust` introduces an easy way to ensure that the name is unique (it also makes the naming logic quite a bit easier).

We also never actually want to reference Rust packages in, for example, the JS namespace, so you introduce a separation layer. The `package.json` is not there to interact with the yarn workspace, really, but to register them to turborepo.

Those were at least my thoughts. Of course, I can also change the logic to use a uniform `-rs` suffix and `@local`, `@libs`, and `@apps`, although the logic will be more brittle.

.github/scripts/rust/sync-turborepo.rs
```diff
@@ -0,0 +1,396 @@
...
+            .components()
+            .any(|component| component.as_str() == "@blockprotocol")
+    }
+
+    fn is_local(dependencies: &str) -> bool {
+        // the following prefixes are considered local dependencies
+        // @blockprotocol, @rust, @libs, @apps, @repo
+        dependencies.starts_with("@blockprotocol/")
+            || dependencies.starts_with("@rust/")
+            || dependencies.starts_with("@libs/")
```
CiaranMn, line 106:
```suggestion
        // @blockprotocol, @rust, @local, @apps, @repo
        dependencies.starts_with("@blockprotocol/")
            || dependencies.starts_with("@rust/")
            || dependencies.starts_with("@local/")
```

CiaranMn (commented):
> Of course, I can also change the logic to use a uniform -rs suffix and @local, @libs, and @apps, although the logic will be more brittle.

What is it about the logic that would be more brittle?

The current convention (which I appreciate is not written down) is as follows:
- `@apps/` are executable things
- `@local/` are dependencies in this repo of other things in the repo
- `@anything-else/` is a published dependency of other things

`@local/` just makes it easier to see which dependencies in a JS/TS package are internal, and which are coming from npm. I appreciate this it not a consideration for Rust since Cargo doesn't have namespacing like that.

If we change Rust packages to `@rust/`, we lose the 'app' vs 'lib' distinction. I am not totally opposed to this, but I'd like to understand more what it is about a `-rs` or `-rust` suffix that would be worse, and which makes changing the existing pattern worth it. We would also need to decide what to do about `apps/hash-graph/package.json`, which currently is still named `@apps/hash-graph`.

indietyp:
instead of simply relying on the crate and `Cargo.toml` itself, we would need to look at the back and then decide if it is in the path `apps/` or `libs/@local`, which makes it a bit more complicated (which is already done). Any change in directory structure or moving of packages then potentially involves the renaming of crates. While not a problem for the script, other dependents (which shouldn't exist) won't get those.

Current `hash-graph` still remains, but this is also where the `apps` vs. `libs` distinction breaks down, we wanted to move this stuff anyway, but `hash-graph/libs/graph` would in theory be a `@local/hash-graph-rs`, which means it would need special casing. Then the question remains, is `@apps/hash-graph`, while `hash-graph/bins/cli` would need to be named `@apps/hash-graph-rs`, because `@apps/hash-graph` is already taken. Putting all crates under `@rust` in the package name would solve this issue. Another thing to consider here is that while the distinction exists in the fs structure, it doesn't necessarily exist in a cargo workspace, and as rust crates don't interact with any other languages (exception here is the WASM type-system) I don't really see how there's a benefit in the naming using `@local` or `@apps`, if that name won't ever be used anyway (although I appreciate that it creates a more cohesive system of separation nevertheless, even if not used.)

I won't ever address a cargo crate by `@apps/hash-graph`, I will via `hash-graph`, as namespaces like that simply don't exist in Rust. So a distinction on `package.json` level doesn't really help here in any scenario I can think of.

I hope this makes sense.

CiaranMn (approved):
> I hope this makes sense.

Thanks @indietyp, yes it does make sense. Happy to go with the `@rust/` prefix then. Apologies in the delay in responding, I didn't get a notification which I normally rely on

CiaranMn (approved):

PR #4347, opened 2024-04-23
H-2454: Allow viewing and triggering Flows

CiaranMn:

The PR introduces a UI for viewing a Flow definition, which is a variant of the existing UI for viewing an in-progress Flow run, as well as a form for manually triggering a Flow run and providing its inputs, if any.

Recap of terms:
- Flow definition: defines a sequence of actions 
- Flow run: an instance of a Flow definition, i.e. an execution of those sequence of actions, potentially with some inputs that come from user action or via automated trigger

## 🔍 What does this change?

- Adds variants of the existing Flow UI (background, nodes) for viewing a definition rather than an in-progress run
- Extracts a sharable `EntitySelector` component to allow for selecting specific entity/entities as an input for a Flow (e.g. "research more details about this entity")
- expands the existing `EntityTypeSelector` component to enable it to select multiple entity types
- improves the logic around progress log aggregation
- a few bug fixes around viewing in-progress flows, e.g. avoid crashes in the graph view

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

There are still a few outstanding UI issues which are for follow-up in H-2547:
- update the entity table to use Glide data grid instead of an HTML table
- the edge arrowhead (marker) don't change colour based on status
- the final layer of nodes in a swimlane shouldn't have source handles
- missing components from the [design](https://www.figma.com/file/gydVGka9FjNEg9E2STwhi2/HASH?type=design&node-id=11749-72632&mode=design&t=hfDS5VmHV8bjenMe-4), e.g. sidebar for in-progress flows, top breadcrumbs

## 🛡 What tests cover this?

- None yet.

## ❓ How to test this?

1. Click 'Run' button on a flow definition

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/1eaa2f9e-06dc-4601-b028-ddc68bc51c07

Screenshot 2024-04-23 at 12 47 55

Screenshot 2024-04-23 at 15 21 03

### DIFF
--- a/apps/hash-frontend/src/pages/flows.page/flow-definition.tsx
+++ b/apps/hash-frontend/src/pages/flows.page/flow-definition.tsx
@@ -9,7 +9,7 @@ import type {
 } from "@local/hash-isomorphic-utils/flows/types";
 import type { Entity } from "@local/hash-subgraph";
 import { Box, Stack, Typography } from "@mui/material";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import type { Edge } from "reactflow";
 import { MarkerType, ReactFlowProvider } from "reactflow";
 
@@ -26,6 +26,14 @@ import {
   groupStepsByDependencyLayer,
 } from "./flow-definition/sort-graph";
 import { Swimlane } from "./flow-definition/swimlane";
+import { transitionOptions } from "./flow-definition/shared/styles";
+import { Button } from "../../shared/ui/button";
+import { useMutation } from "@apollo/client";
+import {
+  StartFlowMutation,
+  StartFlowMutationVariables,
+} from "../../graphql/api-types.gen";
+import { startFlowMutation } from "../../graphql/queries/knowledge/entity.queries";
 
 const getGraphFromFlowDefinition = (
   flowDefinition: FlowDefinitionType,
@@ -184,6 +192,8 @@ export const FlowDefinition = () => {
     return getGraphFromFlowDefinition(selectedFlow);
   }, [selectedFlow]);
 
+  const [showRunModal, setShowRunModal] = useState(false);
+
   const nodesAndEdgesByGroup = useMemo(() => {
     const graphsByGroup: Record<
       number,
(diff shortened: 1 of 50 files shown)


apps/hash-frontend/src/pages/flows.page/flow-definition/run-flow-modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+      {required ? "*" : ""}
+      <Box>{children}</Box>
+    </Typography>
+  </Box>
+);
+
+type UnsupportedPayloadKind = Exclude<PayloadKind, LocalPayloadKind>;
+
+const unsupportedPayloadKinds: UnsupportedPayloadKind[] = [
+  "EntityType",
```
benwerner01, line 54: We have a type selector in the form already for `VersionedUrl`, is it difficult to add support for `EntityType` in addition to `VersionedUrl`? If not we could add support for it now as well.

apps/hash-frontend/src/pages/flows.page/flow-definition.tsx
```diff
@@ -305,16 +318,66 @@ export const FlowDefinition = () => {
...
+              flowDefinition: selectedFlow,
+              flowTrigger: {
+                outputs,
+                triggerDefinitionId: "userTrigger",
+              },
+            },
+          });
+          setShowRunModal(false);
+        }}
+      />
```
benwerner01, line 373: Doesn't block merging, but it'd be nice if the flow run that has been started comes into view automatically.

benwerner01 (approved):
Thanks Ciaran, the new form is working well for me locally. Something I'm wondering is whether we should as part of this replace the legacy `/ai` page with the `/flows` page and start linking it in the sidebar? I don't think there's a lot of functionality that page is providing the `/flows` page now isn't.

If so, perhaps we should add an additional feature flag to hide it in the sidebar in production for now.

But that was just a thought, and not necessarily something that needs to happen as part of this PR.

apps/hash-frontend/src/pages/flows.page/flow-definition/run-flow-modal.tsx
CiaranMn: No, it wouldn't be. We might be better off just picking one though – maybe `VersionedUrl` since the `EntityType` can be resolved from it if necessary in any action, whereas `EntityType` provides an additional burden on the source to have already resolved it, which it might not need to?

apps/hash-frontend/src/pages/flows.page/flow-definition.tsx
CiaranMn: We can do this, but it'll involve amending the `runFlow` mutation to just return the id rather than wait for execution to complete. Are you happy for me to do this? I didn't know if you were still using it somehow for testing

benwerner01: Ah I see, I am still using the GQL endpoint directly for testing and find it useful to get the outputs there.

Perhaps we can add an optional `returnOutputs` parameter or something like that, to allow the caller to decide?

CiaranMn: I wouldn't bother, I don't think we really need it to return the outputs. In the GraphQL playground you could trigger a flow run and then switch to running `getFlowRuns` to get the output (optionally filtered to just that one's id). But we can leave it how it is for a bit longer

apps/hash-frontend/src/pages/flows.page/flow-definition/run-flow-modal.tsx
CiaranMn: Cool, I'll remove it as a drive-by in the next PR

apps/hash-frontend/src/pages/flows.page/flow-definition.tsx
benwerner01: Alright, fair enough

PR #4356, opened 2024-04-25
H-2552: Enable logging in graph-integration tests

TimDiekmann:

To properly find errors in tests logging should be enabled here.

## 🔍 What does this change?

- Adjusts the tracing configuration to allow using the test writer (unused)
- Initialize `tracing` in integrationt tests

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- I was not able to properly use `console_layer` - for some reason it did not print anything. Ideally, that should be used and also `ErrorLayer` should be enabled. To workaround this, this PR adds a simple configuration to at least print log entries.


indietyp (dismissed):
Looks good to me, will be creating a follow up PR that will use the init function instead as discussed offline.

indietyp (approved):

PR #4346, opened 2024-04-23
H-2496: Add link creation capabilities to the research entities action, including between proposed and existing entities

benwerner01:
## 🌟 What is the purpose of this PR?

This PR adds the ability for the coordinating agent in the research entities action to create proposed links, whether they are between existing entities, proposed entities or a proposed entity and an existing entity.

It adds support for passing existing entities to the `inferEntitiesFromContent` flow action, so that links created from the content can link between proposed and existing entities.

## 🔗 Related links

- H-2496

## 🔍 What does this change?

- adds retry capabilities to the `createInitialPlan` LLM call in the worker agent
- adds required `explanation` arguments to the tools in the coordinator agent, to be more consistent with the implementation in the worker agent
- adds a check step when the coordinator agent calls the `complete` tool. I added this because it would sometimes not create links for types that were provided, so I wanted to warn it that it hasn't proposed entities for types that were specified in the input. There's also another warning if no proposed entities were submitted.
- fixes behaviour in `persistEntity` action to only search for existing draft entities if it is also creating the entity as a draft. Otherwise only look for live entities.

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🐾 Next steps

- We may want to also allow passing previously proposed entities to the `inferEntitiesFromContent` action, so that the worker agent in the `researchEntities` action is able to link between entities inferred from one web page, and entities proposed from another page via the `inferEntitiesFromContent` action.

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

Locally I've been testing this functionality with a simple `Company` type (with `organization-name`, `description` properties and a `has-parent` link to another `Company`).

To test links created by the coordinator agent in the `researchEntities` action, I've been using the following inputs for the `startFlow` GQL mutation (see "Demo" section for an example response). If you want to test linking to existing entities, you will need to manually populate the `existingEntities` (e.g. with an `Alphabet Inc.` company entity). company entity. You may also need to update the type IDs, depending on how you are testing this.

<details>
  <summary>Example `startFlow` inputs for "Research Task" flow</summary>
  
  ```json
{
  "flowTrigger": {
    "triggerDefinitionId": "userTrigger",
    "outputs": [
      {
[...]

PR #4357, opened 2024-04-25
H-2543: Write to Google Sheet Flow action

CiaranMn:

Adds a 'write to Google Sheet' Flow action, which takes one of (a) a CSV-formatted string, (b) entities, or (b) a query that produces entities, and writes it to a sheet.

## 🔍 What does this change?

As well as adding the action:
- moves Google-related code from the integration Temporal worker to the "AI" (now basically 'Flows') worker
- deletes code related to the old 'Google Sheets Integration', which will be replaced with a Flow
- moves some Google-related functions to shared locations since the API OAuth routes and Temporal worker needs them
- also removes 'integration'-related code from the Google Sheets integration settings page, just leaving the 'add an account' (this whole UI needs updating to account for integrations all being subsumed into Flows)
- adds a `Spreadsheet` entity type with a `Google Sheet` child

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None yet.

## ❓ How to test this?

1. See demo. The 'Answer Question' and 'Research Action' flows now have Google Sheet writes at the end

## 📹 Demo

**Skip to ~0.50 once the Flow starts** (I will get some editing software)

https://github.com/hashintel/hash/assets/37743469/74a91ede-e58b-44de-9e75-ec3192034544


apps/hash-ai-worker-ts/src/activities/flow-activities/write-google-sheet-action/get-filter-from-bp-query-entity.ts
```diff
@@ -0,0 +1,37 @@
...
+  graphApiClient: GraphApi;
+  queryEntityId: EntityId;
+}) => {
+  let queryEntity: Entity<QueryProperties> | undefined;
+  try {
+    queryEntity = (await getLatestEntityById({
+      graphApiClient,
+      authentication,
+      entityId: queryEntityId,
+    })) as Entity<QueryProperties>;
```
benwerner01, line 24: Nit: maybe we can include a check for whether the entity actually has the correct type, or as part of the query filter by type ID

apps/hash-ai-worker-ts/src/main.ts
```diff
@@ -76,6 +77,11 @@ async function run() {
     port: parseInt(getRequiredEnv("HASH_GRAPH_API_PORT"), 10),
   });
 
+  const vaultClient = createVaultClient();
+  if (!vaultClient) {
+    throw new Error("Vault client not created");
+  }
```
benwerner01, line 83: I think the vault environment variables still need to be added to `docker-compose.ai.yml` for CI to pass, and so that the woker can be run in a docker container locally

libs/@local/hash-isomorphic-utils/src/generate-entity-label.ts
```diff
@@ -111,7 +111,9 @@ export const generateEntityLabel = (
   const options = [
     "name",
     "preferredName",
+    "preferred-name", // @todo H-2561 account for proposed entities having non-standard property keys now
     "displayName",
+    "display-name", // @todo H-2561 account for proposed entities having non-standard property keys now
```
benwerner01, line 116: Are these todos and the task still relevant since https://github.com/hashintel/hash/pull/4346 merged? The proposed entities logged to the frontend will now be converted to have Base URLs again.

The `ProposedEntity` outputted by actions should have properties defined via base URLs in general, the simplified properties should only be relevant for LLM calls.

benwerner01 (dismissed):
Nice work Ciaran, this is working well for me locally 💯 

Had a few minor comments, and a fix that's probably need to get CI to pass, but apart from that this looks great to me 👍

libs/@local/hash-isomorphic-utils/src/generate-entity-label.ts
CiaranMn: Nope – removed in https://github.com/hashintel/hash/pull/4357/commits/65436ab3ec05284f449a37d6abddc6bf434ed401

apps/hash-ai-worker-ts/src/main.ts
CiaranMn: Added in https://github.com/hashintel/hash/pull/4357/commits/d923735f701c341395fb976994b329abe1be0ac2

apps/hash-ai-worker-ts/src/activities/flow-activities/write-google-sheet-action/get-filter-from-bp-query-entity.ts
CiaranMn: I added a check for whether it has the correct property in https://github.com/hashintel/hash/pull/4357/commits/7e9955151f3f493acd91dca1cd984c6a42ab9e00

I think it's fine to use other entity types as long as they have the query property (although we should replace it with a structural query entity soon)

benwerner01 (approved):

PR #4376, opened 2024-04-30
Bump Rust dependencies

indietyp:

This calls `cargo update --aggressive`, I first thought about just running `cargo update h2` and `cargo update rustls`, but thought it might make sense to update to the latest point releases either way.

An interesting observation is that `cargo update --workspace --aggressive` does something differently than `cargo update --aggressive`, which was unexpected.

As a drive-by (as that was discovered in #4283, `devDependencies` now never include `dependencies`.

## 🔗 Related links

- https://rustsec.org/advisories/RUSTSEC-2024-0336
- https://rustsec.org/advisories/RUSTSEC-2024-0332

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -81,6 +81,8 @@ impl<'a> WorkspaceMember<'a> {
             }
         }
 
+        let dev_dependencies = &dev_dependencies - &dependencies;
+
         Self {
             package,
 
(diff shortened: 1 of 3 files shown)


TimDiekmann (approved):
> An interesting observation is that `cargo update --workspace --aggressive` does something differently than `cargo update --aggressive`, which was unexpected.

That's indeed unexpected and explains why you have different results than I had. Thank you!

PR #4378, opened 2024-04-30
H-2547, H-2557: Flows UI improvements, remove `/ai` page

CiaranMn:

This PR makes several improvements to the Flows UI, including:
- Adding a sidebar for in-progress Flows that summarises the steps and their status
- Adds a topbar and moves the flow/run selector and 'run flow' buttons there
- Removes source handles from nodes where they don't connect to anything
- Makes edge markers (the arrows on the end) the correct colour depending on status

It also:
- removes the `/ai` testing page, which is now redundant as `/flows` can be used to trigger flow runs
- makes the 'collapsed sidebar' part of the element flow, with a white background, rather than a floating button as it currently is:
Screenshot 2024-04-30 at 14 42 51

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None yet.

## ❓ How to test this?

1. See demo

## 📹 Demo
**Note:** in the demo the timer text in the sidebar for an in-progress run is slightly misaligned, which I have since fixed

https://github.com/hashintel/hash/assets/37743469/053690bf-d249-48fd-ba84-848f6726d13a


nonparibus:
@CiaranMn We can intro separate from this PR, if you want, but probably makes sense to add the `Workers` sidebar tab, with the `Goals`, and `Flows` sub-pages?

_We will be calling them **flows** and not **jobs**, after all (following yesterday's chat, consistent with the existing HAI Guide and older plans)._

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const { flowRuns, selectedFlowRun, setSelectedFlowRun } =
+    useFlowRunsContext();
+
+  const runOptions = useMemo(
+    () =>
+      flowRuns.filter(
+        (run) => run.inputs[0].flowDefinition.name === selectedFlow.name,
+      ),
+    [flowRuns, selectedFlow.name],
+  );
```
benwerner01, line 64: Possibly this is similar to the comment I left in a prior PR, but just want to make you aware when I clicked `Run` on a flow where there are no prior runs the list of runs didn't appear for me for a few seconds. Maybe this is something we can address once we actually switch to the active run after the button is pressed, when we change what the `startFlow` resolver returns.

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
```diff
@@ -0,0 +1,161 @@
...
+    sx={{
+      fill: ({ palette }) => palette.gray[50],
+      fontSize: 18,
+      mr: 0.5,
+      ml: 1,
+    }}
+  />
+);
+
+export const Topbar = ({
```
benwerner01, line 47: Doesn't need to be addressed as part of this PR, if there are still more layout changes to come or changes around how the `Deliverable` section is rendered, but just wanted to make you aware that currently the content of the `Deliverable` section can overflow in a way that pushes the `Run` button off screen. This is a recording from my 13" MB Pro:

https://github.com/hashintel/hash/assets/42802102/7a52ae56-8d7c-4da7-b928-a0a24dab0674

benwerner01 (approved):
Thanks Ciaran, this looks great 👍 

Had a few small thoughts whilst testing it out locally, but don't think they block merging.

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
CiaranMn: Thanks for flagging – I've aded to H-2542 which is to update the deliverables section

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
CiaranMn: Yes – I can do this as a drive-by in any upcoming task if you don't need the `startFlow` resolver to return the outputs any more

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  {run.closedAt
+                    ? ` – ${format(new Date(run.closedAt), "yyyy-MM-dd h:mm a")}`
+                    : " – in progress"}
+                </MenuItem>
+              ))}
+            </Select>
+          </>
+        )}
+      </Stack>
+      <Button onClick={handleRunFlowClicked} size="xs" sx={{ px: "14px" }}>
```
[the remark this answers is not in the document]
CiaranMn: The issue is the background gradient on our default button hover styles... I've removed it in https://github.com/hashintel/hash/pull/4378/commits/5196812c7866f3dd462d4d9e447fdaba38843e32 in favour of turning the play icon white, now looks like this

https://github.com/hashintel/hash/assets/37743469/4d491f90-5f06-440c-950b-14e19c82c41d

CiaranMn:
@nonparibus 
> 1. We can separate it from this PR, if you want, but probably makes sense to add the `Workers` sidebar tab, with the `Goals`, and `Flows` sub-pages? _We will be calling them **flows** and not **jobs**, after all (following yesterday's chat, consistent with the existing HAI Guide and older plans)._

I've added H-2608 to add the list of Flows page, and the links in the sidebar. This could potentially just be 'Flows' or 'Workers' until we have the 'Goals' page (at the moment in would be a sub-menu containing one item, unless we have something for 'Activity Log')

> 2. Can we dim the screen behind the "Run" modal when it pops up? I'm sure we must be doing this in other places, so let's be consistent if we can be. And if we're not doing it anywhere... let's start (as shown in Figma).

Yes, done in https://github.com/hashintel/hash/pull/4378/commits/39345b43050d7f5d31222a3888a445708883318e – I removed it when trying to get the Google File Picker to appear over the modal and forgot to restore it

apps/hash-frontend/src/pages/flows.page/flow-definition/topbar.tsx
nonparibus: In addition I think the "deliverable" and "raw output" have actually been mixed up here (i.e. the table of entities is the raw output, and the Google Sheets file is the deliverable).

However, given the changes we plan in H-2542, which combine table/graph/deliverable views under a single "Output" tab, I'm not sure we need to fix this here.

CiaranMn: It's not easy to tell from the video, but the 'Raw Output' is the table of persisted entities, and the 'Deliverable' is the data in the CSV. So they are the correct way round.

This is a somewhat confusing example because those companies are just made up by the LLM, and there's only one persisted entity, the file, so it looks like the 'file deliverable' is in the table of entities and 'some entities' are in the deliverables table (but they're not, they're just made-up data).

nonparibus (approved):

PR #4379, opened 2024-04-30
H-1832: Track usage in the `getLlmResponse` method

benwerner01:
## 🔗 Related links

- ...

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Try X
1. Confirm that Y

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/flow-activities/answer-question-action.ts
+++ b/apps/hash-ai-worker-ts/src/activities/flow-activities/answer-question-action.ts
@@ -11,7 +11,7 @@ import {
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
 import { mapGraphApiSubgraphToSubgraph } from "@local/hash-isomorphic-utils/subgraph-mapping";
-import type { AccountId, EntityRootType } from "@local/hash-subgraph";
+import type { AccountId, EntityId, EntityRootType } from "@local/hash-subgraph";
 import { extractEntityUuidFromEntityId } from "@local/hash-subgraph";
 import type { Status } from "@local/status";
 import { StatusCode } from "@local/status";
@@ -156,6 +156,7 @@ const callModel = async (
   iteration: number,
   userAccountId: AccountId,
   graphApiClient: GraphApi,
+  flowEntityId: EntityId,
 ): Promise<
   Status<{
     outputs: StepOutput[];
@@ -169,6 +170,13 @@ const callModel = async (
     tools: answerTools,
     userAccountId,
     graphApiClient,
+    linkUsageRecordToEntities: [
+      {
+        linkEntityTypeId:
+          "https://hash.ai/@hash/types/entity-type/incurred-in/v/1",
+        entityId: flowEntityId,
+      },
+    ],
   });
 
   logger.debug(`Open AI Response received: ${stringify(llmResponse)}`);
@@ -279,6 +287,7 @@ const callModel = async (
             iteration + 1,
             userAccountId,
             graphApiClient,
+            flowEntityId,
           );
         }
 
@@ -364,6 +373,7 @@ const callModel = async (
       iteration + 1,
       userAccountId,
       graphApiClient,
+      flowEntityId,
     );
   }
 
@@ -384,12 +394,13 @@ const callModel = async (
     iteration + 1,
     userAccountId,
     graphApiClient,
+    flowEntityId,
   );
 };
 
 export const answerQuestionAction: FlowActionActivity<{
   graphApiClient: GraphApi;
-}> = async ({ graphApiClient, inputs, userAuthentication }) => {
+}> = async ({ graphApiClient, inputs, userAuthentication, flowEntityId }) => {
   const {
     context,
     entities: inputEntities,
@@ -529,5 +540,6 @@ export const answerQuestionAction: FlowActionActivity<{
     1,
     userAuthentication.actorId,
     graphApiClient,
+    flowEntityId,
   );
 };
--- a/apps/hash-ai-worker-ts/src/activities/flow-activities/answer-question-action.ts
+++ b/apps/hash-ai-worker-ts/src/activities/flow-activities/answer-question-action.ts
@@ -11,7 +11,7 @@ import {
   zeroedGraphResolveDepths,
 } from "@local/hash-isomorphic-utils/graph-queries";
 import { mapGraphApiSubgraphToSubgraph } from "@local/hash-isomorphic-utils/subgraph-mapping";
-import type { EntityRootType } from "@local/hash-subgraph";
+import type { AccountId, EntityRootType } from "@local/hash-subgraph";
 import { extractEntityUuidFromEntityId } from "@local/hash-subgraph";
 import type { Status } from "@local/status";
 import { StatusCode } from "@local/status";
@@ -154,6 +154,8 @@ const callModel = async (
   context: string | null,
   codeUsed: string | null,
   iteration: number,
+  userAccountId: AccountId,
+  graphApiClient: GraphApi,
 ): Promise<
   Status<{
     outputs: StepOutput[];
@@ -165,6 +167,8 @@ const callModel = async (
     messages: mapOpenAiMessagesToLlmMessages({ messages }),
     temperature: 0,
     tools: answerTools,
+    userAccountId,
+    graphApiClient,
   });
 
   logger.debug(`Open AI Response received: ${stringify(llmResponse)}`);
@@ -273,6 +277,8 @@ const callModel = async (
             context,
             null,
             iteration + 1,
+            userAccountId,
+            graphApiClient,
           );
         }
 
@@ -356,6 +362,8 @@ const callModel = async (
       context,
       codeUsed,
       iteration + 1,
+      userAccountId,
+      graphApiClient,
     );
   }
 
@@ -374,6 +382,8 @@ const callModel = async (
     context,
     codeUsed,
     iteration + 1,
+    userAccountId,
+    graphApiClient,
   );
 };
 
@@ -512,5 +522,12 @@ export const answerQuestionAction: FlowActionActivity<{
     });
   }
 
-  return await callModel(messages, contextToUpload ?? null, null, 1);
+  return await callModel(
+    messages,
+    contextToUpload ?? null,
+    null,
+    1,
+    userAuthentication.actorId,
+    graphApiClient,
+  );
 };
--- a/apps/hash-ai-worker-ts/src/activities/infer-entities.ts
+++ b/apps/hash-ai-worker-ts/src/activities/infer-entities.ts
@@ -162,6 +162,8 @@ const inferEntities = async ({
     temperature,
     entityTypes,
     inferenceState,
+    userAccountId: userAuthenticationInfo.actorId,
+    graphApiClient,
   });
 
   logger.debug(
--- a/apps/hash-ai-worker-ts/src/activities/infer-entities/infer-entity-summaries.ts
+++ b/apps/hash-ai-worker-ts/src/activities/infer-entities/infer-entity-summaries.ts
@@ -1,6 +1,6 @@
 import type { VersionedUrl } from "@blockprotocol/type-system/slim";
 import type { GraphApi } from "@local/hash-graph-client";
-import type { AccountId, Entity } from "@local/hash-subgraph";
+import type { AccountId, Entity, EntityId } from "@local/hash-subgraph";
 import type { Status } from "@local/status";
 import { StatusCode } from "@local/status";
 import dedent from "dedent";
@@ -39,6 +39,7 @@ export const inferEntitySummaries = async (params: {
   existingEntities?: Entity[];
   userAccountId: AccountId;
   graphApiClient: GraphApi;
+  flowEntityId?: EntityId;
 }): Promise<Status<InferenceState>> => {
   const {
     completionPayload,
@@ -48,6 +49,7 @@ export const inferEntitySummaries = async (params: {
     existingEntities,
     userAccountId,
     graphApiClient,
+    flowEntityId,
   } = params;
 
   const { iterationCount, usage: usageFromPreviousIterations } = inferenceState;
@@ -83,6 +85,15 @@ export const inferEntitySummaries = async (params: {
     tools,
     userAccountId,
     graphApiClient,
+    linkUsageRecordToEntities: flowEntityId
+      ? [
+          {
+            linkEntityTypeId:
+              "https://hash.ai/@hash/types/entity-type/incurred-in/v/1",
+            entityId: flowEntityId,
+          },
+        ]
+      : [],
   });
 
   if (llmResponse.status !== "ok") {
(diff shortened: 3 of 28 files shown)


apps/hash-ai-worker-ts/src/activities/infer-entities/infer-entity-summaries.ts
```diff
@@ -76,12 +83,24 @@ export const inferEntitySummaries = async (params: {
...
+      : [],
   });
 
   if (llmResponse.status !== "ok") {
     return {
       code: StatusCode.Internal,
       contents: [],
+      message: `Error getting LLM response: ${"message" in llmResponse ? llmResponse.message : llmResponse.status}`,
     };
   }
```
benwerner01, line 105: @CiaranMn I think we should have a conversation at some point on how we plan to replace the `Status` return from the activities, so that we can more ergonomically return errors that can be returned by the `getLlmResponse` method. Ideally we could just directly return the possible errors, so there's no mapping between error objects required. I've created [this ticket](https://linear.app/hash/issue/H-2604/restructure-error-statuses-returned-by-flow-action-activities) for now

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -666,15 +689,170 @@ const getOpenAiResponse = async (
...
+     * @todo: consider abstracting this in a wrapper method, or via
+     * generic params (via a `logUsage` method).
+     */
+    userAccountId: AccountId;
+    graphApiClient: GraphApi;
+    linkUsageRecordToEntities: {
+      linkEntityTypeId: [
+        typeof systemLinkEntityTypes.incurredIn.linkEntityTypeId,
+        typeof systemLinkEntityTypes.created.linkEntityTypeId,
+        typeof systemLinkEntityTypes.updated.linkEntityTypeId,
```
CiaranMn, line 705: We never pass these as far as I can see. Probably we can just take the `entityId`

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -666,15 +689,170 @@ const getOpenAiResponse = async (
...
+      ][number];
+      entityId: EntityId;
```
CiaranMn, line 707: Need some way of tracking the `webId` to link to this usage, which could be passed in here... I think best to discuss sync

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -666,15 +689,170 @@ const getOpenAiResponse = async (
...
+    aiAssistantAccountId = await getMachineActorId(
+      { graphApi: graphApiClient },
+      { actorId: userAccountId },
+      { identifier: "hash-ai" },
+    );
+  } catch {
+    return {
+      status: "internal-error",
+      message: `Failed to retrieve AI assistant account ID ${userAccountId}`,
+    };
```
CiaranMn, line 748: This doesn't use `getAiAssistantAccountIdActivity`, which is used in `infer-entities-from-content-action` and `infer-entities` and ensures that the AI Assistant has permission on the specified web. 

It's possible that this code will be executed without having gone through either of those files, which means the AI Assistant may not have been granted permission on the web and creating the usage records will fail. 

It needn't be checked in `infer-entities-from-content-action` since there's nothing in there affecting the web, though it should be here.

We could also consider getting rid of the AI Assistant bot altogether and just use the appropriate web bot, which will have permission. The AI Assistant bot was mainly to be able to identify which things came from AI, but we are covering that in the new provenance system. This would save yet another check in this function, which we'll be calling quite a lot and now has a lot of hoops to jump through (check usage, get AI bot, check AI bot permissions)

This is also linked to how we attribute usage (e.g. are we creating usage records in an org web in some circumstances).

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/015-add-flow-types.dev.migration.ts
```diff
@@ -239,6 +249,72 @@ const migrate: MigrationFunction = async ({
...
+   */
+
+  const incurredInLinkEntityType = await createSystemEntityTypeIfNotExists(
+    context,
+    authentication,
+    {
+      entityTypeDefinition: {
+        allOf: [linkEntityTypeUrl],
+        title: "Incurred In",
+        description: "Something that was incurred by something else.",
```
CiaranMn, line 264: Consistency (don't really mind `Incurred By` or `Incurred During` either)
```suggestion
        description: "Something that was incurred in something else.",
```

CiaranMn (commented):
Thanks for tackling this @benwerner01 – a few comments which could be addressed now, and a couple around usage attribution (users/orgs) I think are easier discussed sync

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -666,15 +689,170 @@ const getOpenAiResponse = async (
...
   }
+
+  /**
+   * @todo: figure out why TS is not inferring the type of `remainingLlmParams`
+   * correctly (may not be relevant once usage tracking related params are removed)
+   */
+  const llmParams = remainingLlmParams as unknown as T;
```
CiaranMn, line 755: I would split them out so you don't have to rely on a `...rest`, e.g.
```ts
export const getLlmResponse = async <T extends LlmParams>(
  llmParams: T,
  usageTracking: {
    //etc
  }
)
```
This solves the type problem and maintains a clearer separation between 'stuff for the LLM call' and 'other internal stuff'

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -666,15 +689,170 @@ const getOpenAiResponse = async (
...
+
+    try {
+      usageRecordEntityMetadata = await createUsageRecord(
+        { graphApi: graphApiClient },
+        { actorId: aiAssistantAccountId },
+        {
+          serviceName: isLlmParamsAnthropicLlmParams(params)
+            ? "Anthropic"
+            : "OpenAI",
+          featureName: params.model,
```
CiaranMn, line 780:
```suggestion
          serviceName: isLlmParamsAnthropicLlmParams(llmParams)
            ? "Anthropic"
            : "OpenAI",
          featureName: llmParams.model,
```

CiaranMn (commented):
A couple of additional comments on the `getLlmResponse` interface

apps/hash-ai-worker-ts/src/activities/infer-entities.ts
```diff
@@ -162,6 +162,8 @@ const inferEntities = async ({
     temperature,
```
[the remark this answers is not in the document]
benwerner01: Have removed it in https://github.com/hashintel/hash/pull/4379/commits/15f973efe11e99cf728821e780a9fca244dd6f12, this means we won't be tracking which entities were created/updated by the usage record as we did previously, but suppose that's not a big deal as we'll replace use of this method with a flow soon anyways.

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: Yes, we can add this back once we have a need for using other links. Have simplified this param in https://github.com/hashintel/hash/pull/4379/commits/80b1df03da2aba99b964922add7e3f97bf2dfe79

apps/hash-ai-worker-ts/src/activities/flow-activities/answer-question-action.ts
```diff
@@ -165,6 +168,15 @@ const callModel = async (
...
     tools: answerTools,
+    userAccountId,
+    graphApiClient,
+    linkUsageRecordToEntities: [
+      {
+        linkEntityTypeId:
+          "https://hash.ai/@hash/types/entity-type/incurred-in/v/1",
+        entityId: flowEntityId,
+      },
+    ],
```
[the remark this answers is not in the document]
benwerner01: Since we only have use for a single kind of link right now, I've simplified this to not allow specifying link types in https://github.com/hashintel/hash/pull/4379/commits/80b1df03da2aba99b964922add7e3f97bf2dfe79.

Regarding 2, as it was implemented by me this would not have been type `string` - I tried to derive this from the `UsageRecordOutgoingLinksByLinkEntityTypeId` type, but for some reason `Omit<keyof UsageRecordOutgoingLinksByLinkEntityTypeId, "https://hash.ai/@hash/types/entity-type/records-usage-of/v/1">` was not being enforced by TS correctly (it was interpreted as `string` when compiling). But suppose that won't be relevant until we have a need for multiple links here.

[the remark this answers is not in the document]
CiaranMn: It may have been a typo in your comment and this still doesn't work for some reason, but it should be `Exclude` on a union (`Omit` to omit properties from an object)

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: Good idea, much prefer this - have done so in https://github.com/hashintel/hash/pull/4379/commits/9fd6cc2de1f0da9bee00171b6f06cade48ca0075

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: Addressed as part of https://github.com/hashintel/hash/pull/4379/commits/9fd6cc2de1f0da9bee00171b6f06cade48ca0075

apps/hash-ai-worker-ts/src/activities/infer-entities.ts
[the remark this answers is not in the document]
CiaranMn: Thank you – we should also update the place where we weren't tracking usage from this legacy code to pass the same thing as everywhere else

Screenshot 2024-05-02 at 09 07 54

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: Thanks for catching this, I've made use of the method in https://github.com/hashintel/hash/pull/4379/commits/7641eddd58f8d7d9b5b57203a2d119d469d7d71a. I've set it to grant permissions in the user's web for now, but this might require changing if we allow setting the `webId` that the usage record is created in.

Also added a todo for using the machine actor instead once we have this in the provenance metadata.

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: Following our conversation, I've added the `webId` parameter for the `runFlow` temporal workflow, so that usage records/other entities in the flow are created in that web.

Relevant commits: https://github.com/hashintel/hash/pull/4379/commits/b08d4e839ff8a64e2e2f5cdb8f69a2a42993bfec, https://github.com/hashintel/hash/pull/4379/commits/342dc120f4c05755a91e0456e204412a52ffc25d

apps/hash-ai-worker-ts/src/activities/shared/get-flow-context.ts
```diff
@@ -0,0 +1,108 @@
...
+  flowEntityId: EntityId;
+  graphApiClient: GraphApi;
+  userAuthentication: { actorId: AccountId };
+  webId: OwnedById;
+};
+
+const graphApiClient = createGraphClient(logToConsole, {
+  host: getRequiredEnv("HASH_GRAPH_API_HOST"),
+  port: parseInt(getRequiredEnv("HASH_GRAPH_API_PORT"), 10),
+});
```
CiaranMn, line 50: Couldn't this just be exported from this file? It's not specific to the Flow. I guess it only makes a difference if there are callers which _only_ need this client (in which case there's no need to wait for the event history to be searched to give it to them). If we add a cache there's no problem with keeping it in the same context function

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -734,26 +724,24 @@ export const getLlmResponse = async <T extends LlmParams>(
...
-      { graphApi: graphApiClient },
-      { actorId: userAccountId },
-      { identifier: "hash-ai" },
-    );
-  } catch {
+  /**
+   * @todo: use the machine actor here directly instead, if the the AI machine
+   * actor is not longer needed to identify which actions were done by an AI
+   * as this will be available in provenance metadata.
+   */
```
CiaranMn, line 731: We decided not to do this
```suggestion
```

apps/hash-ai-worker-ts/src/activities/shared/get-flow-context.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * @todo: consider caching the result of this, so that multiple calls
+   * to `getFlowContext` within the same flow do not result in individual
+   * calls to the Temporal API.
+   */
+  const { workflowExecutionStartedEventAttributes } =
+    await getWorkflowExecutionStartedEventAttributesOfWorkflow({
+      workflowId,
+    });
```
CiaranMn, line 72: Given how frequently we will need to call this function I would add a cache now. Could use [cache-manager](https://www.npmjs.com/package/cache-manager). I've used [node-cache](https://www.npmjs.com/package/node-cache) but it's no longer maintained.

apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
```diff
@@ -172,6 +172,18 @@ export const runFlowWorkflow = async (
...
   }
 
+  // Ensure the user has permission to create entities in specified web
+  const userHasPermissionToRunFlowInWeb =
+    await flowActivities.userHasPermissionToRunFlowInWebActivity();
+
+  if (userHasPermissionToRunFlowInWeb.status !== "ok") {
+    return {
+      code: StatusCode.PermissionDenied,
+      message: `User does not have permission to run flow in web ${webId}, because they are permissions: ${userHasPermissionToRunFlowInWeb.missingPermissions.join(`,`)}`,
```
CiaranMn, line 182:
```suggestion
      message: `User does not have permission to run flow in web ${webId}, because they are missing permissions: ${userHasPermissionToRunFlowInWeb.missingPermissions.join(`,`)}`,
```

apps/hash-ai-worker-ts/src/activities/shared/get-flow-context.ts
benwerner01: Fair enough, have added the cache in https://github.com/hashintel/hash/pull/4379/commits/5de27c1ac9faca589e36f74e10cd04aabf5cbf3f. We'll only be storing the parts of the params we are actually returning in `getFlowContext`, and I've set a maximum of 100 items (100 concurrent "flows") and a time to live of 10 minutes (accounting for longer running flows). The data is fairly minimal so I'm not concerned about the cache size, but if we add things down the line those parameters might need to be adjusted.

apps/hash-ai-worker-ts/src/activities/shared/get-flow-context.ts
benwerner01: I've moved it into a standalone file where it can be imported from in https://github.com/hashintel/hash/pull/4379/commits/e8ae052ca1e7da509f204959d88dfb152b137b3f

CiaranMn (approved):

PR #4389, opened 2024-05-01
H-2615: Refactor `harpc/wire-protocol` to make use of `Buf/BufMut` instead of `AsyncRead/AsyncWrite`

indietyp:

This adjusts the `harpc/wire-protocol` to make use of synchronous code, expecting that all data is present, instead of using `AsyncRead/AsyncWrite`, this has been decided internally, as it provides greater guarantees, throughput as well as enables the use of `tokio-util` to convert between streams more easily.

Most of this diff are simply code changes to accommodate the new paradigm.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/codec/src/harpc/wire.rs
+++ b/libs/@local/codec/src/harpc/wire.rs
@@ -39,14 +39,12 @@ where
     type Item = T;
 
     #[expect(clippy::big_endian_bytes, reason = "This is a protocol requirement")]
+    #[expect(clippy::missing_asserts_for_indexing, reason = "false positive")]
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
         if src.len() < 32 {
             // Not enough data to decode a request
             return Ok(None);
         }
-
-        // this will always be true, but will eliminate the need for a bounds check (clippy)
-        assert!(src.len() > 31);
 
         // The length marker is always at bytes 30 and 31
         let length = u16::from_be_bytes([src[30], src[31]]) as usize;
--- a/libs/@local/codec/src/harpc/wire.rs
+++ b/libs/@local/codec/src/harpc/wire.rs
@@ -12,9 +12,9 @@ use tokio_util::{
     codec::{Decoder, Encoder},
 };
 
-pub struct ProtocolEncoder<T>(PhantomData<fn() -> *const T>);
+pub struct ProtocolCodec<T>(PhantomData<fn() -> *const T>);
 
-impl<T> Encoder<T> for ProtocolEncoder<T>
+impl<T> Encoder<T> for ProtocolCodec<T>
 where
     T: Encode,
 {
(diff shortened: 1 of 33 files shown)


libs/@local/codec/src/harpc/wire.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    fn encode(&mut self, item: T, dst: &mut BytesMut) -> Result<(), Self::Error> {
+        let mut buffer = Buffer::new(dst);
+
+        item.encode(&mut buffer)
+            .change_context(io::Error::from(io::ErrorKind::InvalidData))
+    }
+}
+
+pub struct ProtocolDecoder<T>(PhantomData<fn() -> *const T>);
```
TimDiekmann, line 32: The type of `ProtocolEncoder` and `ProtocolDecoder` is the same. Can't we reuse it?
```suggestion
pub struct ProtocolCodec<T>(PhantomData<fn() -> *const T>);
```

libs/@local/harpc/wire-protocol/src/codec/buffer.rs
```diff
@@ -0,0 +1,347 @@
...
+            return Err(Report::new(BufferError::EarlyEndOfStream));
+        }
+
+        let mut bytes = [0; N];
+        self.0.copy_to_slice(&mut bytes);
+
+        Ok(bytes)
+    }
+
+    pub(crate) fn next_discard(&mut self, count: usize) -> Result<(), BufferError> {
```
TimDiekmann, line 119: As we don't return anything I think I would rename it to `discard`

TimDiekmann (approved):
I think this is the right choice to use `sync` over `async` here. Thank you very much for the work 🎉 
I have some non-blocking suggestions.

libs/@local/harpc/wire-protocol/src/codec/buffer.rs
indietyp: addressed in [05d2e73](https://github.com/hashintel/hash/pull/4389/commits/05d2e73b18cbc0207217af53062812175c631af9)

libs/@local/codec/src/harpc/wire.rs
indietyp: addressed in [05d2e73](https://github.com/hashintel/hash/pull/4389/commits/05d2e73b18cbc0207217af53062812175c631af9)

libs/@local/codec/src/harpc/wire.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl<T> Decoder for ProtocolDecoder<T>
+where
+    T: Decode<Context = ()>,
+{
+    type Error = Report<io::Error>;
+    type Item = T;
+
+    #[expect(clippy::big_endian_bytes, reason = "This is a protocol requirement")]
```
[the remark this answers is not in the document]
indietyp: addressed in [05d2e73](https://github.com/hashintel/hash/pull/4389/commits/05d2e73b18cbc0207217af53062812175c631af9)

TimDiekmann (approved):

PR #4398, opened 2024-05-03
H-1178: Expose creating multiple entities in one transaction

TimDiekmann:

This is required to seed bigger benchmarks and will also partially allow enforcing minimum number of links.

## 🚫 Blocked by

- #4388 

## 🔍 What does this change?

- Create a `create_entities` function
- Make `create_entity` fallback to the newly created function
- Remove old `insert_entities_batched_by_type` function

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -457,275 +467,255 @@ where
...
+                entity_type_ids,
+                Consistency::FullyConsistent,
+            )
+            .await
+            .change_context(InsertionError)?;
+        if instantiate_permissions.values().any(Not::not) {
+            return Err(Report::new(InsertionError)
+                .attach(StatusCode::PermissionDenied)
+                .attach_printable(
+                    "The actor does not have permission to instantiate one or more entity types",
```
CiaranMn, line 659: Will the types denied be in the context? If not it'd be helpful to add

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -457,275 +467,255 @@ where
...
-                .assert_permission()
-                .change_context(InsertionError)
-                .attach(StatusCode::PermissionDenied)?;
+                .change_context(InsertionError)?;
+            if create_entity_permissions.values().any(Not::not) {
+                return Err(Report::new(InsertionError)
+                    .attach(StatusCode::PermissionDenied)
+                    .attach_printable(
+                        "The actor does not have permission to create entities for one or more \
+                         web ids",
```
CiaranMn, line 679: Same question as types, will the webs denied be in the error somewhere?

CiaranMn (commented):
Thanks @TimDiekmann! Just a couple of questions on errors

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: No, they're not. Sadly it's not straightforward to add them as we use the UUIDv5 to identify them in SpiceDB, but I can create a map on an error case. As an alternative, we could at least put the UUID in the context so it could be looked up in the database.

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: No, will add them, good flag!

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Fixed in [`bbcdbe0` (#4398)](https://github.com/hashintel/hash/pull/4398/commits/bbcdbe0c5bf1fd11393bd375684d1db8d91d6fe3)

apps/hash-graph/libs/graph/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Fixed in [`bbcdbe0` (#4398)](https://github.com/hashintel/hash/pull/4398/commits/bbcdbe0c5bf1fd11393bd375684d1db8d91d6fe3)

CiaranMn (approved):
Great, thank you @TimDiekmann! :shipit:

PR #4399, opened 2024-05-03
H-2541, H-2578: Enable worker querying of large PDFs

benwerner01:

This PR adds the ability for the infer entities from webpage worker agent to query large PDF files via a `queryPdf` tool, which on a high level functions like this:

1. Downloads the PDF file
2. Chunks the PDF file into embeddings (via the TS [LlamaIndex `PDFReader`](https://ts.llamaindex.ai/fr/api/classes/PDFReader)), and stores the embeddings in an in-memory vector database (cached to the disk)
3. Queries the in-memory vector database for the relevant chunks
4. Uses a "Filter and Rank Text Chunks" agent to pick the relevant chunks based on the query, and order them by relevance

It also adds a `inferEntitiesFromText` tool, which functions similarly to the existing `inferEntitiesFromHtml` tool and fixes some other bugs related to inferring links.

## 🔗 Related links

- H-2541
- H-2578

## 🔍 What does this change?

- returns file entities used in the research action to propose entities as additional proposed entities, so these are persisted in the graph
- updates the persist entity action to upload files when a proposed file is encountered
- sets some provenance metadata on proposed entities in the research action

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Currently only files that are linked to in web pages can be queried in the worker agent. It's possible that files may be encountered as top-level web search results in the coordinator agent. As a temporary measure, I've filtered out PDF files from web search results but we will likely want to expand the capabilities of the coordinator agent or inputs for the worker agent to directly handle files in some way (H-2668)

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

One way to test the functionality, is by providing the following arguments to the "Research Entities" flow in the UI:

- Prompt: Get the subsidiary companies of Marks and Spencers, you must get this information from their annual investors report in 2023
- Entity Types: Company (seeded from the flow test types)

## 📹 Demo

image


apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
```diff
@@ -0,0 +1,394 @@
...
+              the required entities and their properties from the web page in a single tool call.
+            `),
+        },
+        validAt: {
+          type: "string",
+          format: "date-time",
+          description: dedent(`
+            A date-time string in ISO 8601 format, representing when the provided text content is valid at.
+            If you don't know, assume it is the current date and time.
+            The current time is "${new Date().toISOString()}".
```
CiaranMn, line 251: I appreciate this is pre-existing but haven't looked at it again (see other comment) I'm not sure we should have an assumption that content is valid `now`... in many cases the date won't be inferrable from whatever bit of text the agent is looking at, and we will be giving false confidence that it is current

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent.ts
```diff
@@ -832,26 +673,110 @@ export const inferEntitiesFromWebPageWorkerAgent = async (params: {
...
+                        location: {
+                          uri: toolCallInput.fileUrl,
+                          /** @todo */
+                          name: undefined,
+                          description: undefined,
+                        },
+                        loadedAt: accessedRemoteFile?.loadedAt,
+                        /** @todo */
+                        authors: undefined,
+                        firstPublished: validAt,
```
CiaranMn, line 721: The document content may be valid but have been published earlier – this is especially true given that we assume a document is valid today if there is no information available on its validity (which I think we should remove)
```suggestion
                        firstPublished: undefined,
```

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent.ts
```diff
@@ -832,26 +673,110 @@ export const inferEntitiesFromWebPageWorkerAgent = async (params: {
...
+              )?.payload.value as ProposedEntity[]
+            ).map(
+              (proposedEntity) =>
+                ({
+                  ...proposedEntity,
+                  provenance: editionProvenance,
+                  propertyMetadata: Object.keys(proposedEntity.properties).map(
+                    (propertyBaseUrl) => ({
+                      /** @todo: figure out whether this is the path */
+                      path: [propertyBaseUrl],
```
CiaranMn, line 741: Obsolete comment?

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
```diff
@@ -0,0 +1,394 @@
...
+            If you are looking for specific properties of an entity, name them.
+          `),
+        },
+        exampleText: {
+          type: "string",
+          description: dedent(`
+            An example of the data you are looking for in the PDF document.
+            This can be a table, a paragraph, or any other relevant content.
+          `),
+        },
```
CiaranMn, line 198: Out of interest how does this work in practice? My gut reaction is that it might make up some stuff that bears no resemblance to how the document is written, how tables are formatted etc, or which _does_ but corresponds to the 'wrong' section and thus introduces noise in the matching. But I may be wrong and it helps to find the info

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/open-ai-coordinating-agent.ts
```diff
@@ -257,28 +259,41 @@ const parseCoordinatorInputs = async (params: {
...
     graphApiClient,
-    entityTypeIds: entityTypeIds!,
+    entityTypeIds: [
+      ...entityTypeIds!,
+      /**
+       * We need to include the types of the existing entities, to correctly
+       * determine whether the links are satisfiable.
+       *
+       * @todo: address this in the `getDereferencedEntityTypesActivity` function
+       */
```
CiaranMn, line 278: Are you planning to address this in the near-ish future? If not I'd move the `@todo` into the function itself and create a task for it (you may already have done so)

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  let localFilePath;
+  try {
+    localFilePath = await downloadFileToFileSystem(originalUrl);
+  } catch (err) {
+    const message = `Error downloading file from URL: ${(err as Error).message}`;
+
+    // @todo logger – update once #4290 merges
+    // eslint-disable-next-line no-console
+    console.log(message);
```
CiaranMn, line 151: This can use the logger now

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -0,0 +1,273 @@
...
+  const entityTypeId =
+    params.entityTypeId ??
+    getEntityTypeIdForMimeType(mimeType) ??
+    systemEntityTypes.file.entityTypeId;
+
+  const stats = statSync(localFilePath);
+  const fileSizeInBytes = stats.size;
+
+  /** @todo: allow overriding this via an input, as with other actions (hardcoded in Flow definition?) */
+  const ownedById = userAuthentication.actorId;
```
CiaranMn, line 168: This can be taken from context now

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -0,0 +1,273 @@
...
+      "URL",
+    "https://blockprotocol.org/@blockprotocol/types/property-type/original-url/":
+      originalUrl,
+    "https://blockprotocol.org/@blockprotocol/types/property-type/file-size/":
+      fileSizeInBytes,
+  };
+
+  const incompleteFileEntityMetadata = await graphApiClient
+    .createEntity(
+      // @todo which bot should this be?
```
CiaranMn, line 193:
```suggestion
      // @todo H-2618 use AI bot or web bot as appropriate depending on provenance
```

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -0,0 +1,273 @@
...
+  const properties: FileProperties = {
+    ...initialProperties,
+    "https://blockprotocol.org/@blockprotocol/types/property-type/file-url/":
+      formatFileUrl(key),
+    ...fileStorageProperties,
+  };
+
+  const updatedEntityMetadata = await graphApiClient
+    .patchEntity(
+      // @todo which bot should this be?
```
CiaranMn, line 237:
```suggestion
      // @todo H-2618 use AI bot or web bot as appropriate depending on provenance
```

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -0,0 +1,273 @@
...
+      reject(new Error(`Error reading file: ${error.message}`)),
+    );
+    fileStream.pipe(req);
+    fileStream.on("end", () => req.end());
+  }).finally(() => {
+    void unlink(filePath);
+  });
+};
+
+export const getFileEntityFromUrl = async (params: {
```
CiaranMn, line 115: Can we require provenance here now (which may be empty) and pass it where we call it?

apps/hash-ai-worker-ts/src/activities/flow-activities/web-search-action.ts
```diff
@@ -18,6 +18,13 @@ export const webSearchAction: FlowActionActivity = async ({ inputs }) => {
 
   const webPagesUrls = webSearchResults
     .map(({ url }) => url)
+    /**
+     * The coordinator agent using this method does not have the ability to directly
+     * interact with PDFs, so we filter these out for now.
+     *
+     * @todo: account for PDFs being returned in search results in the coordinator agent
```
CiaranMn, line 25: This seems worth addressing / not overlooking and we probably won't touch this function much, can you create a task for it if you haven't already please, and put it in next sprint?

CiaranMn (commented):
Thanks @benwerner01, great additions!

A few comments, mostly tidying up we can do now with the new/recent functionality. One question on `validAt`

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent.ts
benwerner01: Yes, thanks for catching. Removed in https://github.com/hashintel/hash/pull/4399/commits/66d0d8c6a3c71159276273dd5d6049715a58a0d4

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
benwerner01: Thanks, also found two more places this ticket was referenced. Have addressed in https://github.com/hashintel/hash/pull/4399/commits/43df379d6619a694de928d73b370aa487213fc0f

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
benwerner01: Addressed in https://github.com/hashintel/hash/pull/4399/commits/06508102ebde6e372441fc4b12d3c00373352c6e

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
benwerner01: You are correct that the agent makes up things here. It's usually a lit of information that it's looking for a given entity, i.e. something like "Investment Fund X, Y voting rights, Z% ownership," etc).

This was the only way I managed to get the worker agent to compose a query that actually returns the relevant table via the similarity search within 15 chunks, most of the time. My understanding of why this works better than another generic query is because the naming the properties in the query results in a closer embedding between the query and the relevant chunk.

I've attempted to ask the agent to specify the properties it's looking for in the query, but it often times didn't make such an attempt.

If you have other ideas on alternative approaches we could use to getting the worker agent to produce a better query, let me know.

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
benwerner01: What kind of provenance information are you referring to?

apps/hash-ai-worker-ts/src/activities/flow-activities/web-search-action.ts
benwerner01: There's an existing ticket, have linked it in the comment in https://github.com/hashintel/hash/pull/4399/commits/13a747564972dd57710a742a6fa008847139f3b5

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
CiaranMn: Whatever we can populate – the edition-level metadata at least (e.g. the fact that it's a flow, and the flow id)

When we have a description / title inferred from the AI we can also add that. Maybe the webpage where it was found, for a source on the file URL or something?

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/open-ai-coordinating-agent.ts
benwerner01: Yeah I think we should address it, had a quick stab at this but ended up being more complex than expected. I've created [a ticket](https://linear.app/hash/issue/H-2685/in-getdereferencedentitytypes-allow-for-specifying-additional-entity), moved the comments into the method and linked the ticket in https://github.com/hashintel/hash/pull/4399/commits/47d45bc16efe53d3f5cd8f72627cf35f3fc35c72.

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
benwerner01: Fair enough, I think the reason I added this a while ago was to satisfy the "measured at" property on the "Stock Constituent" entity type, where I thought if a date isn't specified on the web page, we can assume the current time is when the data should be assumed as correct. In the context of the FTSE constituents, that would be correct, but appreciate this might cause false confidence in other use-cases.

Should I remove the default here for now, and allow the agent to optionally pass it? In practice this will likely mean it never passes anything for this parameter, but I expect it will need to be reworked when we are passing facts instead of arbitrary text (where each fact may have a different `validAt` timestamp associated with it, depending on the context, or none if we don't know).

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
benwerner01: I've added the origin to the file entity in https://github.com/hashintel/hash/pull/4399/commits/6a0e5ffd6e5c0e9b020179ab32acc59e3e1352f1, but haven't specified the `actorType` because getting a file from a URL isn't really an "AI" - perhaps we need an `"Machine"` value for this field or something like that?

> When we have a description / title inferred from the AI we can also add that. Maybe the webpage where it was found, for a source on the file URL or something?

I suppose we could do this for the edition, since we have a `sources` array there, but wouldn't these fields make more sense as properties on the file entity, rather than metadata in the provenance?

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
CiaranMn: Not currently, no! Fair enough

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/infer-entities-from-web-page-worker-agent/tool-definitions.ts
CiaranMn: We can leave it since you'll be reworking all this anyway, but I think we should generally not default to any assumptions about when data is current if it can't be inferred

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
CiaranMn: Yes, a `@todo` would be fine. Presumably we have it where we propose the file entity though, can we add it there as a source the same way we do for other proposed entiites?

benwerner01: I've added a `@todo`, and accounted for metadata set on the proposed file entity in https://github.com/hashintel/hash/pull/4399/commits/bc016a28eb42b53c282ba5f7c1b0e71c3557bebf

CiaranMn (dismissed):

nonparibus:
@benwerner01 @CiaranMn I've resolved the FOSSA license error that caused this to fail. If conflicts can be resolved we can re-add to the merge queue.

CiaranMn (approved):

PR #4433, opened 2024-05-07
H-2542: Implement activity log and updated Flow outputs designs

CiaranMn:

Implements the designs for the Flow outputs (around [here](https://www.figma.com/design/gydVGka9FjNEg9E2STwhi2/HASH?node-id=11700-5449&t=zbbqVZqGxqcvCM5H-4), internal link):
1. Adds an activity log with granular progress updates, and adds a 'QueriedWeb' progress log
2. Makes the 'outputs' section have toggleable display of entities table, entities graph, and deliverables
3. Makes the entire bottom section (log and outputs) fixed to the bottom of the screen, rather than having to be scrolled down to
4. Updates the tables used for the entity table, and uses it in the activity log – I've opted for a new virtualized table rather than Glide Grid, as having contextual information from anchor links (e.g. see the URL on hover) is important here, and is not supported by a canvas library (we would need to overlay an anchor link exactly over the canvas pixels on hover). I've tested this and performance is fine up to ~300k rows, after which it stops rendering content (I think possibly due to browser limits on element scrollHeight, unsure)
5. A few bug fixes and consistency improvements in backend Flow handling / definitions

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- We need some solution to the outputs section (right half of bottom bit) being too wide for the screen – it's possible to get a horizontal scroll here if needed but needs further thought

## 🛡 What tests cover this?

- None yet

## ❓ How to test this?

1. Run any flow and see the outputs

## 📹 Demo

I haven't yet downloaded video editing software to do exciting smash cuts so you'll have to scrub through this.

**Note that** the sorting on the tables was bugged in this video, I fixed in https://github.com/hashintel/hash/commit/d4b5f13ac7f8c1be46b04539bd7d71d993e7098a

https://github.com/hashintel/hash/assets/37743469/8bfefa18-d90c-49fa-9978-c1b42c0ab25f


apps/hash-frontend/package.json
```diff
@@ -122,7 +122,7 @@
     "@types/jsonpath": "0.2.0",
     "@types/lodash": "4.14.188",
     "@types/papaparse": "5.3.14",
-    "@types/react": "18.2.79",
+    "@types/react": "18.2.68",
```
CiaranMn, line 125: Downgraded due to some issue later ones have introduced, can check back on it again later

benwerner01 (dismissed):
Thanks Ciaran, this looks great and is working for me locally 👍 

Unrelated to this PR, but one thought I had whilst looking at the UI is that it'd be nice to see the trigger outputs somewhere in the UI, is this something that will be added at some point?
image

apps/hash-ai-worker-ts/src/activities/flow-activities/write-google-sheet-action.ts
```diff
@@ -282,11 +282,15 @@ export const writeGoogleSheetAction: FlowActionActivity<{
...
-        })),
+        ...existingSheets
+          .filter(
+            (sheet) => sheet.properties?.sheetId !== placeholderFirstSheetId,
+          )
+          .map((sheet) => ({
+            deleteSheet: {
+              sheetId: sheet.properties?.sheetId,
+            },
+          })),
```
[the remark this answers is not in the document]
CiaranMn: I think so? At least I got the same thing and added this and haven't seen it since, although I'm not sure how the sheets are ending up in a position where they have the placeholder in them

apps/hash-frontend/src/pages/flows.page/flow-definition/shared/virtualized-table.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+const SortButton = <Sort extends VirtualizedTableSort>({
+  columnId,
+  sort,
+  setSort,
+}: { columnId: NonNullable<Sort["field"]> } & TableSortProps<Sort>) => {
+  const isSorted = sort?.field === columnId;
+  const isSortedAscending = isSorted && sort.direction === "asc";
+
+  console.log({ sort });
```
CiaranMn, line 98:
```suggestion
```

CiaranMn:
> Unrelated to this PR, but one thought I had whilst looking at the UI is that it'd be nice to see the trigger outputs somewhere in the UI, is this something that will be added at some point?

Yes, how TBC (cc @nonparibus).

There is a plan to treat any 'Research Goal' trigger input specially (show it in the box in the top left where the flow description currently is), but we also need to cover the web affected, draft, and other arbitrary inputs

PR #4440, opened 2024-05-09
H-2589: Provide separate script to migrate API

TimDiekmann:

The migrations for the Node API are supposed to run in a dedicated service. This provides a script to run the migrations.

## 🔍 What does this change?

- Remove `getRequiredEnv` from isomorphic utils and use backend utils instead 
- Create script to run embeddings

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

This is only one half of the work to build the API image and upload it to AWS. In the second PR, the changes to the infrastructure will be made.


CiaranMn (approved):
Thanks @TimDiekmann!

PR #4442, opened 2024-05-09
H-2608: Add Flows list page, sidebar entries

CiaranMn:

This PR
- adds the `/flows` route, which _will_ list definitions that a user has in their web(s), but currently just lists the hardcoded flows we're working with in the absence of flow definitions being in the db
  - once they _are_ in the db, any time a user wants to use a flow they will copy it their web, so they can vary the triggers etc, apart from the special flows used to handle browser plugin inference
- adds placeholders for `/goals` (AI-directed flows) and `/workers` (all goal / flow runs) pages
- adds links to the above in the main navigation sidebar
- moves the existing 'view a flow definition/run' page to `/@[shortname]/flows/[flow-uuid]`
  - this will later be split out so that viewing a run lives at a different route (although in practice it'll be the same components), to accommodate viewing both 'goal' and 'flow' runs
- updates the latter to take 'which flow is selected' from the url path rather than local state

Drive-bys:
- update the browser plugin tab shown when it opens a window to do research in
- switch to a newly-started flow run automatically when triggered from a definition page  

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

Visit links in the sidebar

## 📹 Demo

https://github.com/hashintel/hash/assets/37743469/58c00f95-35db-4c64-86c7-f310a67f5ac6


apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
```diff
@@ -56,6 +58,46 @@ export const PageSidebar: FunctionComponent = () => {
 
   const { draftEntities } = useDraftEntities();
 
+  /**
+   * @todo H-2664 – remove this
+   */
+  if (frontendUrl !== "http://localhost:3000") {
+    enabledFeatureFlags.workers = false;
+  }
```
benwerner01, line 66: Is this neccessary? The feature flag will be disabled in production (`workers` won't be present for any users in production), and the alice user is automatically seeded with all feature flags enabled so it should be automatically enabled in development when logged in as that user.
```suggestion
```

benwerner01 (dismissed):
Thanks Ciaran, this looks good to me 👍 

One thought I had is that it seems slightly inconsistent to specify routes like `/flows` and to namespace `/@hash/flows/[slug]`, but I can see from the internal slack thread that this was the plan and I don't feel particularly strongly about it.

Perhaps it's something we'll want to think about again when we have users running flows in their own web, and in an orgs web, when we decide whether to show both of those flows in the same place or require the user to switch between the web selector.

apps/hash-frontend/src/pages/_app.page.tsx
```diff
@@ -225,6 +225,8 @@ const featureFlagHiddenPathnames: Record<FeatureFlag, string[]> = {
   documents: [],
   canvases: [],
   notes: ["/notes"],
+  workers: ["/goals", "/flows", "/workers"],
```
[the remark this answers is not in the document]
CiaranMn: It doesn't, no. We don't do it for pages either. We can address both at some point if thought important

apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
CiaranMn: I guess not, no

CiaranMn:
> One thought I had is that it seems slightly inconsistent to specify routes like /flows and to namespace /@hash/flows/[slug], but I can see from the internal slack thread that this was the plan and I don't feel particularly strongly about it.

> Perhaps it's something we'll want to think about again when we have users running flows in their own web, and in an orgs web, when we decide whether to show both of those flows in the same place or require the user to switch between the web selector.

We will remove the web selector at some point – `/flows` will show all flows that are relevant to a user. Which might just be all flows from webs they belong to, unless we start discriminating between which flows in an org web are of interest to which members of that org

benwerner01 (dismissed):

apps/hash-frontend/src/shared/layout/layout-with-sidebar/sidebar.tsx
```diff
@@ -1,5 +1,6 @@
 import { faHome } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon, IconButton } from "@hashintel/design-system";
+import { frontendUrl } from "@local/hash-isomorphic-utils/environment";
```
CiaranMn, line 3:
```suggestion
```

benwerner01 (approved):

PR #4459, opened 2024-05-13
H-2697: Add `harpc` net

indietyp:

This implements the `harpc` service layer in Rust, which uses libp2p to communicate a stream of `Request` and `Response` streams. Because this is such a critical part, extensive care has been taken to test every facet.

This building block is needed to create an axum-like interface in the future. The code has been written with performance in mind and is quite distributed. Each transaction on a connection will spawn two tasks, which handle receiving and transmitting data, back to a connection, which has two tasks to handle these. This might seem like many spawned tasks, but Tokio tasks are very lightweight, and one shouldn't be afraid to spawn thousands or millions, according to the Tokio docs. This clear separation between the different layers also enables a greater separation of concerns and easier testing.

> Tasks in Tokio are very lightweight. Under the hood, they require only a single allocation and 64 bytes of memory. Applications should feel free to spawn thousands, if not millions of tasks.[^1]

Most of the code is tests, to ensure that any condition (I could think of) is tested to ensure everything is rock solid. This includes 138 different tests over the whole codebase.

The current problem is `yamux`, which seems prone to deadlock.

[^1]: https://tokio.rs/tokio/tutorial/spawning

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/service/src/session/server/session_id.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Creates a new `SessionId` with the given `id`.
+    ///
+    /// This method is hidden, as it should only be used sparangly in rare cases where a
+    /// predetermined `SessionId` is acceptable, such as in tests.
+    #[doc(hidden)]
+    #[must_use]
+    pub const fn new_unchecked(id: usize) -> Self {
+        Self(id)
+    }
+}
```
TimDiekmann, line 33: Creating a session ID seems to be quite easy, can't we just use the producer from above? Or is this because the `Producer` use `pub(crate)`?

libs/@local/harpc/service/src/session/server/write.rs
```diff
@@ -0,0 +1,263 @@
...
+
+            tx,
+        }
+    }
+
+    pub(crate) const fn is_error(&self) -> bool {
+        matches!(self.kind, ResponseKind::Err(_))
+    }
+
+    pub(crate) fn with_no_delay(self, no_delay: bool) -> Self {
```
TimDiekmann, line 65: `without_delay`?

libs/@local/harpc/service/src/session/server/write.rs
```diff
@@ -0,0 +1,263 @@
...
+pub(crate) struct ResponseWriter<'a> {
+    id: RequestId,
+    kind: ResponseKind,
+    index: usize,
+
+    buffer: SegmentedBuf<Bytes>,
+
+    tx: &'a mpsc::Sender<Response>,
+
+    no_delay: bool,
```
TimDiekmann, line 29: Could you add a doc-string what this is doing?

TimDiekmann (commented):
Partial review (skipped files with >400 lines).
This looks very very promising! Super excited about this!

libs/@local/harpc/service/src/config.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    pub command_buffer_size: NonZero<usize>,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            ping: ping::Config::new(),
+            swarm: SwarmConfig::default(),
+            command_buffer_size: NonZero::new(16).expect("infallible"),
```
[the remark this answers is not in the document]
TimDiekmann: Also applies to other locations with `expect("infallible")`.

libs/@local/harpc/service/src/session/client/connection.rs
```diff
@@ -0,0 +1,182 @@
...
+    stream: S,
+
+    tx: Arc<HashIndex<RequestId, mpsc::Sender<Response>>>,
+}
+
+impl<S> ConnectionResponseDelegateTask<S>
+where
+    S: Stream<Item = error_stack::Result<Response, io::Error>> + Send,
+{
+    #[allow(clippy::integer_division_remainder_used)]
```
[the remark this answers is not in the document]
indietyp: Sadly this needs to be done **every** time the select! macro is used :/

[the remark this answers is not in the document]
TimDiekmann: Can we make it `expect` instead then?
```suggestion
    #[expect(clippy::integer_division_remainder_used, reason = "Required for `select!`)]
```

libs/@local/harpc/service/src/session/client/connection.rs
```diff
@@ -0,0 +1,182 @@
...
+struct ConnectionRequestDelegateTask<S> {
+    sink: S,
+    rx: mpsc::Receiver<Request>,
+}
+
+impl<S> ConnectionRequestDelegateTask<S>
+where
+    S: Sink<Request> + Send,
+{
+    #[allow(clippy::integer_division_remainder_used)]
```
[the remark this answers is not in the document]
TimDiekmann:
```suggestion
    #[expect(clippy::integer_division_remainder_used, reason = "Required for `select!`)]
```

libs/@local/codec/src/harpc/wire.rs
```diff
@@ -14,6 +14,13 @@ use tokio_util::{
 
 pub struct ProtocolCodec<T>(PhantomData<fn() -> *const T>);
 
+impl<T> ProtocolCodec<T> {
+    #[must_use]
+    pub const fn new() -> Self {
+        Self(PhantomData)
+    }
+}
+
```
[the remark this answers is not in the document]
indietyp: Yes, because `Default` isn't const.

libs/@local/harpc/service/src/transport/ipc.rs
```diff
@@ -0,0 +1,69 @@
...
+pub(crate) struct TransportLayerIpc {
+    tx: mpsc::Sender<Command>,
+}
+
+impl TransportLayerIpc {
+    pub(super) fn new() -> (Self, mpsc::Receiver<Command>) {
+        let (tx, rx) = mpsc::channel(IPC_BUFFER_SIZE);
+
+        (Self { tx }, rx)
+    }
```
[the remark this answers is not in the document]
indietyp: good catch, it's also still using a constant, I did not want to have around. I'll rework it so that it instead takes a sender as argument.

libs/@local/harpc/service/src/session/server/write.rs
indietyp: This is intentional, because it is `no_delay`, I specifically chose this name because of `TCP_NODELAY`, I'll write comments, what the setting exactly means.

libs/@local/harpc/service/src/transport/error.rs
```diff
@@ -0,0 +1,34 @@
+use error_stack::Report;
+use libp2p::StreamProtocol;
+use libp2p_stream as stream;
+
+/// Errors while opening a new stream.
+#[derive(Debug, Clone, PartialEq, thiserror::Error)]
+pub(crate) enum OpenStreamError {
```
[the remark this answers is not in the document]
indietyp: fixed in [f684611](https://github.com/hashintel/hash/pull/4459/commits/f684611f06713f693884c05afffb59f4032d7e3a)

libs/@local/harpc/service/src/session/client/connection.rs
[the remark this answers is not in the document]
indietyp: fixed in [1d7b9e1](https://github.com/hashintel/hash/pull/4459/commits/1d7b9e1dca17231d11361a94e247f5418edba7a2)

libs/@local/harpc/service/src/session/client/connection.rs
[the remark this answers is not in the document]
indietyp: fixed in [1d7b9e1](https://github.com/hashintel/hash/pull/4459/commits/1d7b9e1dca17231d11361a94e247f5418edba7a2)

libs/@local/harpc/service/src/transport/error.rs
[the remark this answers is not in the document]
indietyp: I did many hours ago, not anymore, fixed in [0083c68](https://github.com/hashintel/hash/pull/4459/commits/0083c6882b2659737f0b446ba8cf6e63a988bc9c)

libs/@local/harpc/service/src/session/server/write.rs
indietyp: added docs about the semantics of the option in [16e2fe8](https://github.com/hashintel/hash/pull/4459/commits/16e2fe86ac3df4c076bd8b0ef4fa012c45dbb0b2)

TimDiekmann (commented):
Partial review as GitHub is bugging for me.
Amazing work!

libs/@local/harpc/net/src/session/server/connection/collection.rs
```diff
@@ -0,0 +1,383 @@
...
+            let has_removed = storage
+                .remove_if_async(&id, |state| state.generation == generation)
+                .await;
+
+            // we only need to check if empty, if we have removed something
+            if has_removed && storage.is_empty() {
+                notify.notify_one();
+            }
+        });
+    }
```
TimDiekmann, line 334: Similar to the client code: it feels wild to spawn a thread in `Drop`.

libs/@local/harpc/net/src/session/server/connection/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                    reason = "This simply returns a drop guard, that is carried through the \
+                              transaction lifetime."
+                )]
+                let (permit, request_tx, request_rx) =
+                    match self.transactions.acquire(request_id).await {
+                        Ok((permit, tx, rx)) => (permit, tx, rx),
+                        Err(error) => {
+                            tracing::info!("transaction limit reached, dropping transaction");
+
+                            return self.respond_error(request_id, error, &tx).await;
```
TimDiekmann, line 133: `respond_error` does return `()`, using a `return` here is misleading.
```suggestion
                            self.respond_error(request_id, error, &tx).await;
                            return;
```

libs/@local/harpc/net/src/session/server/connection/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                #[expect(
+                    clippy::significant_drop_in_scrutinee,
+                    reason = "This simply returns a drop guard, that is carried through the \
+                              transaction lifetime."
+                )]
+                let (permit, request_tx, request_rx) =
+                    match self.transactions.acquire(request_id).await {
+                        Ok((permit, tx, rx)) => (permit, tx, rx),
+                        Err(error) => {
+                            tracing::info!("transaction limit reached, dropping transaction");
```
TimDiekmann, line 131: Shouldn't this be a warning?
```suggestion
                            tracing::warn!("transaction limit reached, dropping transaction");
```

libs/@local/harpc/net/src/session/error.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn code(&self) -> ErrorCode {
+        ErrorCode::CONNECTION_CLOSED
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, thiserror::Error)]
+#[error(
+    "The connection is in the graceful shutdown state and no longer accepts any new transactions"
+)]
+pub struct ConnectionGracefulShutdownError;
```
TimDiekmann, line 65: I wonder if we want to rename this to `ConnectionInShutdown` or `ConnectionInShutdownError`. Or do we distinguish between a graceful shutdown and other shutdowns?

libs/@local/harpc/net/src/session/server/connection/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                            // we don't cancel them, because they might still respond to the requests,
+                            // they'll cancel themselves once the handle has finished
+                            self.transactions.shutdown_senders();
+
+                            // drop the sender (as we no longer can use it anyway)
+                            // this will also stop the delegate task, once all transactions have finished
+                            tx.take();
+                        }
+                        Some(Ok(request)) => {
+                            let tx = tx.clone().expect("infallible; sender is only unavailble once the stream is exhausted");
```
TimDiekmann, line 286:
```suggestion
                            let tx = tx.clone().unwrap_or_else(|| unreachable!("sender is only unavailble once the stream is exhausted"));
```

libs/@local/harpc/net/src/session/server/connection/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                        Ok(Err(error)) => {
+                            tracing::warn!(?error, "connection prematurely closed");
+                        },
+                        Err(error) => {
+                            tracing::warn!(?error, "unable to join connection delegate task");
+                        }
+                    }
+
+                    finished.add_permits(1);
+                },
```
TimDiekmann, line 307: I had to look up what `handle` actually is in this context. We could call it `connection_task_handle` or something else to make this more clear.

libs/@local/harpc/net/src/session/server/connection/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                () = notify.notified() => {
```
TimDiekmann, line 308: Similar to above: The comment below helps a lot but it would be good if `notify` would be renamed. Maybe `item_removed_notify`?

libs/@local/harpc/net/src/session/client/connection/collection.rs
```diff
@@ -0,0 +1,148 @@
...
+        let notify = Arc::clone(&self.notify);
+
+        tokio::spawn(async move {
+            let removed = storage.remove_async(&id).await;
+
+            if removed && storage.is_empty() {
+                // Notify the session that all transactions have been completed
+                notify.notify_one();
+            }
+        });
```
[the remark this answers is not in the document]
indietyp: I had that before, but with `SCC,` one limitation is that if an entry is held across an await point and the sync API is used, it can lead to a deadlock. I have observed that myself, so I am spawning a task.


See: https://github.com/wvwwvwwv/scalable-concurrent-containers/issues/143#issuecomment-2099791565

TimDiekmann (commented):
Another partial review but committing the comments now as I'll be away for a few hours. Again, great work and really helpful inline docs.

libs/@local/harpc/net/src/session/server/connection/mod.rs
indietyp: Yes, this is an überbleibsel from a previous version, where `self.respond_error` did actually return something. Thank you!

libs/@local/harpc/net/src/session/client/connection/collection.rs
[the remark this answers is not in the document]
TimDiekmann: I see, thanks!

libs/@local/harpc/net/src/session/server/config.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use tokio::sync::Semaphore;
+
+use crate::macros::non_zero;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ConcurrentConnectionLimit(u32);
+
+impl ConcurrentConnectionLimit {
+    pub fn new(value: u32) -> Option<Self> {
```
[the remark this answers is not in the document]
indietyp: `0` literally means that we won't accept any connections (this actually works) I wanted to avoid a panic at `MAX_PERMITS` which is why I added this (`tokio` `Semaphore` panics if it is too high)

libs/@local/harpc/net/src/session/error.rs
indietyp: Yes, there's a difference; if you use the cancel token from the `Transport` and say `cancel`, it will shut down everything as soon as possible, which means we won't wait for anything.

indietyp: This is a hard shutdown, just to make sure we clean up if we need to

libs/@local/harpc/net/src/session/server/connection/mod.rs
indietyp: renamed in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/server/connection/mod.rs
indietyp: renamed in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/codec.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn encode_stream(
+        &self,
+        items: impl Stream<Item = T> + Send + Sync + 'static,
+    ) -> impl Future<Output = impl Stream<Item = Result<Bytes, Self::Error>> + Send + Sync + 'static>
+    + Send;
+}
+
+pub trait PlainError: Error {
+    fn code(&self) -> ErrorCode;
+}
```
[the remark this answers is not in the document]
indietyp: renamed in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/client/connection/mod.rs
```diff
@@ -0,0 +1,362 @@
...
+        }
+
+        // we need to cancel any remaining transactions, as we won't be able to route to them
+        // anymore
+        let guard = Guard::new();
+        for (_, state) in self.storage.iter(&guard) {
+            state.sender.close();
+        }
+
+        tracing::debug!("response delegate task has been shut down");
```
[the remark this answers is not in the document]
indietyp: added in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/client/connection/mod.rs
```diff
@@ -0,0 +1,362 @@
...
+/// Takes into account the [`RequestId`] to route the response to the correct transaction
+///
+/// [`RequestId`]: harpc_wire_protocol::request::id::RequestId
+struct ConnectionResponseDelegateTask<S> {
+    config: SessionConfig,
+    stream: S,
+
+    storage: TransactionStorage,
+
+    notify: Arc<Notify>,
```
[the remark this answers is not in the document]
indietyp: renamed in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/client/connection/collection.rs
```diff
@@ -0,0 +1,148 @@
...
+                entry.insert_entry(state);
+            }
+            Entry::Occupied(entry) => {
+                // This should never happen, as the permit is unique and should not be shared
+                // between multiple transactions.
+                // This can only happen in the case that we overflow the u32 RequestId and we have a
+                // connection that never terminated
+                // This is **highly** unlikely, in case it happens we cancel the old task and
+                // replace it with the new one.
+                tracing::warn!("Transaction ID collision detected, cancelling old transaction");
```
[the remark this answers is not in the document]
indietyp: added in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/client/connection/collection.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    pub(crate) const fn storage(&self) -> &TransactionStorage {
+        &self.storage
+    }
+
+    pub(crate) const fn notify(&self) -> &Arc<Notify> {
+        &self.notify
+    }
+
+    pub(crate) async fn acquire(&self) -> (Arc<TransactionPermit>, tachyonix::Receiver<Response>) {
```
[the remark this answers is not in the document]
indietyp: de-arced in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/client/connection/collection.rs
```diff
@@ -0,0 +1,148 @@
...
+pub(crate) struct TransactionState {
+    pub(crate) sender: tachyonix::Sender<Response>,
+    pub(crate) cancel: CancellationToken,
+}
+
+impl Cancellable for TransactionState {
+    fn is_cancelled(&self) -> bool {
+        self.cancel.is_cancelled()
+    }
+}
```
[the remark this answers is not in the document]
indietyp: renamed in [4adf694](https://github.com/hashintel/hash/pull/4459/commits/4adf694885abb40ac80f7f817aa2f1ae0776b701)

libs/@local/harpc/net/src/session/server/session_id.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    pub(crate) fn produce(&mut self) -> SessionId {
+        let id = self.next;
+        self.next = self.next.wrapping_add(1);
+
+        SessionId(id)
+    }
+}
+
+// what is in the session is not part of the service, but higher level!
```
TimDiekmann, line 19: What do you mean?

indietyp: Whoops that isn't supposed to be there, that was me brainstorming. I first wanted put the session logic into the transport layer, then decided against it. This was a question to future me that I forgot to remove :P

libs/@local/harpc/net/src/session/server/transaction/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+                }
+
+                break;
+            };
+
+            match bytes {
+                Ok(_) if writer.is_error() => {
+                    // we had an error previously, so just ignore the rest of the stream
+                    continue;
+                }
```
TimDiekmann, line 101: Does the `writer` print the error?

indietyp: Yes, this is just in the case that we have a stream of:

```
Ok, Ok, Err, Ok, Err, Err
```

In that case, we can safely drop the `Ok` between the errors because we will error out anyway. `is_error` means that the current stream we're transmitting is an error stream.

libs/@local/harpc/net/src/transport/error.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn from(err: stream::OpenStreamError) -> Self {
+        Self(err)
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, thiserror::Error)]
+#[error(
+    "The underlying swarm has been stopped or the request could not be fulfilled by the swarm."
+)]
+pub struct TransportError;
```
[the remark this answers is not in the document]
TimDiekmann: ^ This also helps readability on `.changeContext(TransportError::...)`

libs/@local/harpc/net/src/transport/task.rs
```diff
@@ -0,0 +1,318 @@
...
+    ListenOn {
+        address: Multiaddr,
+        tx: SenderListenerId,
+    },
+    ExternalAddresses {
+        tx: oneshot::Sender<Vec<Multiaddr>>,
+    },
+}
+
+pub(crate) struct Task {
```
TimDiekmann, line 54: Probably should rename this to `TransportTask`?

libs/@local/harpc/net/src/session/test.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    let address = {
+        let server_stream = server
+            .listen(address)
+            .await
+            .expect("should be able to listen on TCP");
+
+        // Give the swarm some time to acquire the external address
+        // This is necessary for CI, as otherwise the tests are a bit flaky.
+        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```
TimDiekmann, line 181:
```suggestion
        // Give the swarm some time to acquire the external address
        // This is necessary for CI, as otherwise the tests are a bit flaky.
        // TODO: Implement waiting for server to be ready
        //   see https://linear.app/hash/issue/H-2837/[bilal]-implement-waiting-for-server-to-be-ready
        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```

libs/@local/harpc/net/src/session/test.rs
```diff
@@ -0,0 +1,426 @@
...
+    let address = {
+        let server_stream = server
+            .listen(address)
+            .await
+            .expect("should be able to listen on TCP");
+
+        // Give the swarm some time to acquire the external address
+        // This is necessary for CI, as otherwise the tests are a bit flaky.
+        // TODO: `listen_on` should wait until the transport layer has acquired said address.
+        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```
TimDiekmann, line 343:
```suggestion
        // Give the swarm some time to acquire the external address
        // This is necessary for CI, as otherwise the tests are a bit flaky.
        // TODO: `listen_on` should wait until the transport layer has acquired said address.
        //   see https://linear.app/hash/issue/H-2837/[bilal]-implement-waiting-for-server-to-be-ready
        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```

libs/@local/harpc/net/src/session/server/connection/collection.rs
```diff
@@ -0,0 +1,383 @@
...
+};
+
+#[derive(Debug)]
+struct ConcurrencyPermit {
+    _permit: OwnedSemaphorePermit,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct ConcurrencyLimit {
+    limit: usize,
```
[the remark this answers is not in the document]
indietyp: A value of 0 means that we do not accept any connections; in theory, we could utilize this to drain a node while creating a new one without canceling all the nodes.

libs/@local/harpc/net/src/session/server/connection/collection.rs
indietyp: added explanation in [7a79dba](https://github.com/hashintel/hash/pull/4459/commits/7a79dbae5a36851a976933d9646bcbec4af143d6)

libs/@local/harpc/net/src/session/server/connection/collection.rs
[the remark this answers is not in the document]
indietyp: The limit itself is only used during reporting, to be like: this is the actual (current limit) that has been exceeded.

libs/@local/harpc/net/src/session/server/config.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[must_use]
+    pub const fn as_usize(self) -> usize {
+        self.0 as usize
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SessionConfig {
+    pub event_buffer_size: NonZero<usize>,
+    pub transaction_buffer_size: NonZero<usize>,
```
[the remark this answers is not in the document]
indietyp: docs in [0027fa5](https://github.com/hashintel/hash/pull/4459/commits/0027fa5c98efce59376f18d938518a8b03ca72fa)

libs/@local/harpc/net/src/session/client/config.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct SessionConfig {
+    pub response_delivery_deadline: Duration,
+
+    pub per_connection_request_buffer_size: NonZero<usize>,
+    pub per_connection_transaction_garbage_collect_interval: Duration,
+
+    pub per_transaction_response_buffer_size: NonZero<usize>,
+    pub per_transaction_response_byte_stream_buffer_size: NonZero<usize>,
+
+    pub no_delay: bool,
```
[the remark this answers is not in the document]
indietyp: added in [0027fa5](https://github.com/hashintel/hash/pull/4459/commits/0027fa5c98efce59376f18d938518a8b03ca72fa)

TimDiekmann (changes requested):
I think your prettier is outdated. I updated the branch and when CI is passing I'm happy to approve 🙂

TimDiekmann (approved):
✅  Great! 🚀

vilkinsons (approved):

PR #4467, opened 2024-05-15
H-2692: Infer facts from text before proposing entities

benwerner01:

This PR modifies how entities are proposed in the research action, and stops making use of the `inferEntitiesFromContent` action to propose entities. The process of proposing entities in the worker agent is now:

1. Summarise all relevant entities in the text provided
2. Infer facts from the text, which have a subject, predicate, and singular object
3. For each summarised entity, propose the entity and its outgoing links based on the facts which have the entity as their "subject"

In follow up we aim to use the underlying pieces of this process to no longer propose entities when processing a single piece of text. Instead we will gather all the facts from different sources on the coordinator level, so that entities can be proposed based on information obtained from a variety of sources.

## 🔗 Related links

- H-2692

## 🔍 What does this change?

- adds mocks for the temporal functionality needed to run flow steps methods in the `vitest` testing library

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- This PR breaks the functionality of proposed entities being able to link to existing entities passed to the research action. This will be partially addressed when the remaining work on moving the fact gathering to the coordinator level takes place, as we can incorporate existing entities in the required fact deduplication work (H-2693). Ideally we will also make the fact inference methods aware of existing methods (H-2713).
- We will need to add additional fields to the facts so that provenance information is captured. This is not yet required for this PR, as we can determine the provenance data as we would have previously as all properties are being derived from a single source.

## 🐾 Next steps

- gather facts at the coordinator level from multiple sources, before proposing the entities (H-2693)
- Add ability to specify existingEntities when inferring facts, so that these can be directly linked from new proposed entities (H-2713)

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

Try out the existing flows that make use of the research action. I've used the "Get subsidiary companies of Google" as a prompt and the `Company` flow test type to produce the demoed result.

## 📹 Demo

image


apps/hash-ai-worker-ts/scripts/compare-llm-response.ts
```diff
@@ -144,6 +88,7 @@ export const compareLlmResponses = async () => {
   const llmResponses = await Promise.all(
     models.map((model) => {
       return getLlmResponse(
+        // @ts-expect-error - figure out what's going wrong here
```
CiaranMn, line 91: The `models` field accepts any model name, but the params demands that only OpenAI model names are sent with OpenAI params, etc.

The type of `CompareLlmResponseConfig` would have to be updated somehow to make sure they were in sync, maybe with generics or something. Not sure it's worth sinking a lot of time into, since it's a testing utility.

apps/hash-ai-worker-ts/src/activities/shared/activity-logger.ts
```diff
@@ -23,15 +23,20 @@ const log = (
...
       fs.mkdirSync(logFolderPath);
     }
     const logFilePath = path.join(logFolderPath, `${requestId}.log`);
     fs.appendFileSync(logFilePath, `${logMessage}\n`);
   }
 
-  logToConsole[level](logMessage);
+  if (process.env.NODE_ENV === "test") {
+    // eslint-disable-next-line no-console
+    console.log(logMessage);
```
CiaranMn, line 36: Why not use `logToConsole` here?

CiaranMn (commented):
Good stuff, only a couple of minor comments

apps/hash-ai-worker-ts/src/activities/shared/activity-logger.ts
benwerner01: Because it doesn't actually log anything when I run the test via `vitest`, not entirely sure why haven't spent a lot of time trying to fix this.

CiaranMn: Fair play, don't worry about it then

apps/hash-ai-worker-ts/scripts/compare-llm-response.ts
benwerner01: Yeah, I made an attempt at this but didn't think it was worth sinking a lot of time into. We'll probably revise this method further once we dedicate time on the model evaluation you've mentioned.

PR #4470, opened 2024-05-16
H-2702: Find corresponding tracing data when analyzing benchmarks

TimDiekmann:

The benchmark analyzer needs to be able to associate the generated artifacts from benchmarks with the benchmark results.

## 🚫 Blocked by

- #4464 
- #4469 

## 🔍 What does this change?

- Re-use algorithm from criterion to create artifact locations so we directly can use the `directory_name` of a measurement to read artifacts
- Read the `tracing.folded` content

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/bench/util.rs
+++ b/apps/hash-graph/bench/util.rs
@@ -1,4 +1,4 @@
-use std::{fs, mem::ManuallyDrop, path::Path};
+use std::{cmp, fs, mem::ManuallyDrop, path::Path};
 
 use authorization::{
     schema::{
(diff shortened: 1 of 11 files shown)


apps/hash-graph/bench/util.rs
```diff
@@ -57,12 +57,52 @@ pub fn setup_subscriber(
...
+    fn truncate_to_character_boundary(s: &mut String, max_len: usize) {
+        let mut boundary = cmp::min(max_len, s.len());
+        while !s.is_char_boundary(boundary) {
+            boundary -= 1;
+        }
+        s.truncate(boundary);
+    }
+
+    // Taken from Criterion source code
+    pub fn make_filename_safe(string: &str) -> String {
```
TimDiekmann, line 70: I wonder if we want to move this function to `repo-chores` or another central place instead and don't rely on the internals of `criterion`

CiaranMn: Can do – no strong opinion

CiaranMn (dismissed):

apps/hash-graph/bench/util.rs
TimDiekmann: I encountered an issue in a later PR using this method, did this in H-2705.

CiaranMn (approved):

PR #4479, opened 2024-05-17
H-1323: Add `/workers` page, update flow run routes

CiaranMn:

The PR introduces a `/workers` page which lists all active worker runs. These may be 'goals' which are research-driven flows (the `/goals` page to be introduced in the next PR), or other flows.

It also updates the routes for viewing:
1. Flow definitions to `/@[namespace]/flows/[id]`
2. Worker runs (flows or goals) to `/@[namespace]/workers/[id]`

These routes have essentially the same UI for now so they share a component. I've also updated the context providers so that we poll for all flow runs with fewer details (that doesn't require the event history) on a slower schedule, and the single selected flow run more frequently.

## 🔗 Related links

- [Design](https://www.figma.com/design/gydVGka9FjNEg9E2STwhi2/HASH?node-id=12148-12196&t=fX96eEss2DP5fBxL-4) (internal)

## 🔍 What does this change?

As well as the main changes above:
- Throws errors without a `cause` from `run-workflow-run.ts` because the cause causes Temporal to not fail the workflow properly
- Update the `persistFlow` activity to use the provided `webId` instead of always the user

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None yet.


apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
```diff
@@ -387,13 +375,7 @@ export const runFlowWorkflow = async (
...
     const errorMessage =
       "No steps have satisfied dependencies when initializing the flow.";
-    throw new Error(errorMessage, {
-      cause: {
-        code: StatusCode.FailedPrecondition,
-        message: errorMessage,
-        contents: [{ flow }],
-      },
-    });
+    throw new Error(errorMessage);
```
benwerner01, line 378: Maybe we can discuss this internally rather than in a thread, but these outputs have been useful for me in the past when debugging errors in flows - but it makes sense to remove them if they're leading to the flow not being failed correctly.

I suppose I can use the `getFlowRuns` resolver in the meantime.

CiaranMn: I did this in the next PR, https://github.com/hashintel/hash/pull/4485/commits/b211df26720c762258bb7b1e25f295645d46b3e6

benwerner01 (approved):

apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
benwerner01: Gotcha, thanks!

PR #4503, opened 2024-05-23
H-2758: Add package for graph SDK

TimDiekmann:

We're going to use an SDK to interact with the Graph in the future. This adds a package to `@local` and makes sure the proper development can be started.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graph-sdk/node/package.json
+++ b/libs/@local/hash-graph-sdk/node/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "@local/hash-graph-sdk",
+  "version": "0.0.0-private",
+  "private": true,
+  "description": "SDK for the HASH Graph API.",
+  "license": "AGPL-3.0",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./pkg/src/*.ts",
+      "node": "./pkg/src/*.ts",
+      "default": "./pkg/src/*.ts"
+    }
+  },
+  "scripts": {
+    "fix:eslint": "eslint --fix .",
+    "lint:eslint": "eslint --report-unused-disable-directives .",
+    "lint:tsc": "tsc --noEmit",
+    "test:unit": "vitest --run"
+  },
+  "dependencies": {
+    "@local/hash-graph-client": "0.0.0-private",
+    "@local/hash-subgraph": "0.0.0-private"
+  },
+  "devDependencies": {
+    "@local/eslint-config": "0.0.0-private",
+    "@local/tsconfig": "0.0.0-private",
+    "@vitest/coverage-istanbul": "1.6.0",
+    "eslint": "8.57.0",
+    "typescript": "5.1.6",
+    "vitest": "1.6.0"
+  }
+}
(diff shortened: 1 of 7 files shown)


libs/@local/hash-graph-sdk/node/src/entity.ts
```diff
@@ -0,0 +1,14 @@
+import type { Entity as GraphApiEntity } from "@local/hash-graph-client/api";
+import type { EntityRecordId } from "@local/hash-subgraph";
+
+export class Entity {
+  #entity: GraphApiEntity;
```
TimDiekmann, line 5: Is `#entity` preferred over `private entity`?

libs/@local/hash-graph-sdk/node/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@local/hash-graph-sdk",
+  "version": "0.0.0-private",
+  "private": true,
+  "description": "SDK for the HASH Graph API.",
```
CiaranMn, line 5: It's not really for the API, right? it doesn't communicate with it... a Graph API SDK would be some combination of HaRPC and this. Maybe just 'SDK for the HASH Graph' for now?

libs/@local/hash-graph-sdk/node/src/entity.ts
CiaranMn: Yes, it provides runtime privacy, `private` does not

libs/@local/hash-graph-sdk/node/src/entity.ts
```diff
@@ -0,0 +1,14 @@
...
+
+  constructor(entity: GraphApiEntity) {
+    this.#entity = entity;
+  }
+
+  public get id(): EntityRecordId {
+    return this.#entity.metadata.recordId as EntityRecordId;
+  }
```
CiaranMn, line 13: I appreciate this is just a first function but we never use `recordId` as a whole so I'd make it `entityId` instead (and qualify `id` since it has many of them)
```suggestion
  public get entityId(): EntityRecordId {
    return this.#entity.metadata.recordId.entityId as EntityId;
  }
```

libs/@local/hash-graph-sdk/node/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "license": "AGPL-3.0",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./pkg/src/*.ts",
+      "node": "./pkg/src/*.ts",
+      "default": "./pkg/src/*.ts"
```
CiaranMn, line 12: what do we need the `pkg` folder for?

libs/@local/hash-graph-sdk/node/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
CiaranMn, line 1: This needs to work in the browser too so I'd call it `typescript` instead of `node`

libs/@local/hash-graph-sdk/node/package.json
TimDiekmann:
```suggestion
  "name": "@local/hash-graph-sdk",
  "version": "0.0.0-private",
  "private": true,
  "description": "SDK for the HASH Graph.",
```

libs/@local/hash-graph-sdk/node/package.json
TimDiekmann: That's a copy/paste error 🙂 
```suggestion
      "types": "./src/*.ts",
      "node": "./src/*.ts",
      "default": "./src/*.ts"
```

CiaranMn (approved):

vilkinsons (approved):

PR #4505, opened 2024-05-23
H-2744: Improve fact creation process + other fixes

benwerner01:
## 🔗 Related links

- ...

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Try X
1. Confirm that Y


apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
```diff
@@ -114,7 +114,7 @@ const nonDuplicateEntitySummaries: LocalEntitySummary[] = [
   },
 ];
 
-test(
+test.skip(
```
CiaranMn, line 117: Is this intentionally skipped?

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
```diff
@@ -156,3 +156,167 @@ test(
...
+    name: "gpt-3.5-turbo-instruct",
+    summary:
+      "GPT-3.5 Turbo Instruct is a large language model compatible with legacy Completions endpoints.",
+    entityTypeId:
+      "https://hash.ai/@ftse/types/entity-type/large-language-model/v/1",
+  },
+];
+
+test(
+  "Test deduplicate entities with FTSE350 companies",
```
CiaranMn, line 308: I think this description needs changing

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ts
```diff
@@ -35,7 +35,9 @@ export const deduplicationAgentSystemPrompt = `
   3. the same or very similar name may refer to different entities
   4. the same or very similar summary may refer to different entities
 
-  If in doubt, do not merge entities. If you are confident that two entities are duplicates, merge them.
+  If an entity is a version of another entity, do not report it as a duplicate as they are not referring to the same thing.
```
CiaranMn, line 38: I don't think it's clear here what 'a version of another entity' means (I only know because of the context of the LLM model example). Whether a newer version of something is a duplicate of something else is probably also use case dependent. I've tried to clarify a bit.

```suggestion
  If an entity is a different version of another entity (e.g. a different version of a piece of software, a newer model of a car, or the Toastmaker 3000 versus the Toastmaker 2000), do not report it as a duplicate as they are not referring to the same thing.
```

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/research-entities-action.ai.test.ts
```diff
@@ -135,7 +135,7 @@ test.skip(
   },
 );
 
-test(
+test.skip(
```
CiaranMn, line 138: Intentionally skipped in this file?

apps/hash-ai-worker-ts/src/activities/get-web-page-activity.ts
```diff
@@ -10,6 +10,49 @@ import { requestExternalInput } from "./shared/request-external-input";
...
+    },
+    disallowedTagsMode: "discard",
+  });
+
+  const slicedSanitizedHtml = sanitizedHtml.slice(
+    0,
+    /**
+     * Assume that each token is 4 characters long.
+     *
+     * @todo: use `js-token` to more accurately determine the number of tokens.
```
CiaranMn, line 48: Is this the right library name?

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -164,7 +164,9 @@ const sanitizeInputBeforeValidation = (params: {
   return input;
 };
 
-const ajv = new Ajv();
+const ajv = new Ajv({
+  strict: false,
+});
```
CiaranMn, line 169: What is the exact failure here? We can probably make the configuration more precise to avoid failures we don't want to allow creeping in

If it's to do with us using non-standard JSON Schema keywords, we can [add them](https://ajv.js.org/strict-mode.html#prohibit-ignored-keywords)

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
benwerner01: I'm using these test files to test out the methods, and am skipping/not skipping tests based on which ones I want to run. I appreciate this adds things to the diff unnecessarily depending on the last test I ran whilst developing, if you'd like maybe I can mark all tests as skipped and manually unskip them without committing in development. Might be an extra layer of safety to prevent someone unintentionally running AI tests.

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
benwerner01: Thanks, addressed in https://github.com/hashintel/hash/pull/4505/commits/f456c9916a65fd9b416024fcc792548877ed0e3e

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
benwerner01: I set this because of the `label` key is used in the graphics card data types, and it's not a supported JSON schema keyword. I've manually added the JSON schema keyword and removed `strict: false`, and tested it to see if it works in https://github.com/hashintel/hash/pull/4505/commits/b4cf7d7e422ac3a7eea6c9ea4b6220cd8b9a850f

I've created a [follow-up ticket](https://linear.app/hash/issue/H-2840/filter-out-label-property-from-dereferenced-entity-types) for filtering out the `label` property from the generated JSON schemas when passing them to the LLM. The reason I didn't directly remove it in the `dereferenceEntityType` method, is because I thought the `label` might be useful for the LLM to better understand the data type.

apps/hash-ai-worker-ts/src/activities/get-web-page-activity.ts
benwerner01: Sorry, you're right it should be `js-tiktoken`. Have addressed in https://github.com/hashintel/hash/pull/4505/commits/fb5423e94953d8ebb83ed9a38ae698e961242981

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
CiaranMn: Maybe yes, and some comment explaining what these are for. Can be in follow up. A lot of them just expect the response to be defined which doesn't seem super useful but I may be misunderstanding still

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
CiaranMn: We can just leave it in as additional context, and remove it if it becomes a problem

CiaranMn (approved):

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/deduplicate-entities.ai.test.ts
benwerner01: I've skipped the tests in https://github.com/hashintel/hash/pull/4505/commits/513f37af6044f441c4d2210769bb96c3305f4dc6.

Sorry if this hasn't been clear, but these "tests" aren't defined to be tests that can pass or fail, they should probably be thought of as scripts which I use to easily run methods with different inputs, observe their logs, and read they're outputs (manually in the console). The reason I used a testing framework for this is to make the process of mocking methods such as `getFlowContext` more easily, so that running these doesn't require creating a flow entity, temporal workflow, etc. I've been committing them because they are useful to be able to return back to, and not have to define from scratch when developing a method.

CiaranMn (approved):

PR #4507, opened 2024-05-24
H-2763: Add provenance data to subgraph library

TimDiekmann:

The fields are missing in the subgraph library, this makes them accessible.


CiaranMn (commented):
How about the property provenance?

TimDiekmann:
Also forgot to update the mapping functions 🙃

CiaranMn (approved):

PR #4526, opened 2024-05-28
H-2758: Use an `Entity` class in Typescript

TimDiekmann:

To extend the SDK functionality for entities, the `Entity` typedef needs to be replaced with a class.

## 🔍 What does this change?

- Extends the `Entity` class to get `properties`, `metadata`, and `linkData` to match the current entity definition. Based on this we can extend the functionality.
- Use the SDK `Entity` everywhere
- Remove the old typedef

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
Some people might not prefer classes in TS, as it breaks many of the structural typing aspects of the language and are quite "heavy" (in a sense); another approach that I like to follow is the one that is extensively used in effect:

./models/Entity.ts
```
const TypeId: unique symbol = Symbol.for("@app/hash/models/Entity");
export type TypeId = typeof TypeId;

export interface Entity extends <common protocols here> {
	[TypeId]: TypeId, // this ensures it is never constructed outside of this function

	/// ... other public properties, preferably readonly
}

interface EntityImpl extends Entity {
	// private properties
}

const EntityProto: Omit<EntityImpl, > = {
	[TypeId]: TypeId,
	// see: https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Duration.ts#L142
}

const make = () => {
	const entity = Object.create(EntityProto);
	// assign all the properties we're missing, have omitted from the proto
}

// expose functions that take as value: `entity: Entity`, and any additional arguments
```

This way, you're still able to reap any benefits from TS, such as type guards, very easily.

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -1,14 +1,124 @@
...
+  return (
+    "entityTypeIds" in
+    (entity as GraphApiEntity | EntityData<Properties>).metadata
+  );
+};
+
+export class Entity<
+  Properties extends EntityPropertiesObject = EntityPropertiesObject,
+> {
+  #entity: EntityData<Properties>;
```
benwerner01, line 53: Is there any reason to not make this `private`?

TimDiekmann: Yes, `#entity` also has runtime privacy while `private entity` is only for type checks. See https://github.com/hashintel/hash/pull/4503#discussion_r1612964485

PR #4534, opened 2024-06-02
H-2793: Use method to access property metadata in TS

TimDiekmann:

The internals of property metadata are not specified and should only be accessed via a method.

## 🔍 What does this change?

- Forbid direct access to `metadata.properties` and use `propertyMetadata()` instead
- Make some more `EntityMetadata` and `LinkData` fields internal 

## 🚫 Blocked by

- #4526 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/history-section/get-history-events.ts
+++ b/apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/history-section/get-history-events.ts
@@ -8,7 +8,6 @@ import {
   getEntityTypeById,
   getPropertyTypeForEntity,
 } from "@local/hash-subgraph/stdlib";
-import { isEqual } from "lodash";
 
 import type { EntityDiff } from "../../../../../../graphql/api-types.gen";
 import type { HistoryEvent } from "./shared/types";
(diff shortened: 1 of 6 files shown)


apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/history-section/get-history-events.ts
```diff
@@ -112,12 +111,12 @@ export const getHistoryEvents = (diffs: EntityDiff[], subgraph: Subgraph) => {
...
-       * @todo H-2775 – handle property objects and changes to array contents
-       */
-      (map) => map.path[0] === key,
-    );
+    /**
+     * @todo H-2775 – handle property objects and changes to array contents
+     */
+    const propertyProvenance = firstEntityEdition.propertyMetadata([
+      key,
+    ])?.provenance;
```
Comment (benwerner01) [inline apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/history-section/get-history-events.ts:119]: This was probably not caused by this PR but the prior one, but just to flag I'm encountering an runtime error on the history:
image

It seems that `getEntityRevision` is not actually returning instances of the `Entity` class, but the serialized version. When I log `firstEntityEdition` before the error is thrown I get this:
```JSON
{
    "firstEntityEdition": {
        "properties": {
           ...
        },
        "metadata": {
            ...
        }
    }
}
```

Comment (TimDiekmann) [reply apps/hash-frontend/src/pages/[shortname]/entities/[entity-uuid].page/entity-editor/history-section/get-history-events.ts:119]: Everything should now work again, thanks! 🙂

CiaranMn (approved):

PR #4536, opened 2024-06-02
H-2794: Create functions for Entity CRUDA operations

TimDiekmann:

To allow usage in the workers this defines the common functionality to the Graph API through functions on the `Entity` class.

## 🚫 Blocked by

- #4534 
- #4526 

## 🔍 What does this change?

- Add functions to `create`, `patch`, `update` (similar to `patch` but with a full property object, which allows returning a new instance), `archive`, and `unarchive`
- Use the functions everywhere if possible 

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -13,9 +17,26 @@ import type {
...
+  properties: EntityPropertiesObject;
+  linkData?: LinkData;
+  entityTypeId: VersionedUrl;
+  entityUuid?: EntityUuid;
+  draft?: boolean;
+  confidence?: number;
+  propertyMetadata?: PropertyMetadataMap;
+  provenance?: ProvidedEntityEditionProvenance;
+  relationships: EntityRelationAndSubject[];
+};
```
benwerner01, line 39: I think @CiaranMn already pointed this out, but it'd be nice to re-use the GraphApi type for the create entity request here and only override the types that don't correspond. That way in future if we add additional fields, we don't have to add them here by default. I think it'd become something like this:
```suggestion
export type CreateEntityParameters = {
  ownedById: OwnedById;
  properties: EntityPropertiesObject;
  entityTypeId: VersionedUrl;
  entityUuid?: EntityUuid;
  draft?: boolean;
} & Omit<
  CreateEntityRequest,
  "ownedById" | "properties" | "entityTypeId" | "entityUuid" | "draft"
>;
```

TimDiekmann: Yes, I actually just prepared a commit for this:
```ts
export type CreateEntityParameters = Omit<
  GraphApiCreateEntityRequest,
  "entityTypeIds" | "decisionTime"
> & {
  ownedById: OwnedById;
  properties: EntityPropertiesObject;
  linkData?: LinkData;
  entityTypeId: VersionedUrl;
  entityUuid?: EntityUuid;
  propertyMetadata?: PropertyMetadataMap;
};

export type PatchEntityParameters = Omit<
  GraphApiPatchEntityParams,
  "entityId" | "entityTypeIds" | "decisionTime"
> & {
  entityTypeId?: VersionedUrl;
};
```

CiaranMn (approved):

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/get-file-entity-from-url.ts
```diff
@@ -284,14 +286,8 @@ export const getFileEntityFromUrl = async (params: {
...
-    })
-    .then(
-      (result) =>
-        new Entity<FileProperties>({
-          metadata: result.data,
-          properties,
-        }),
-    );
+    },
+  )) as Entity<FileProperties>;
```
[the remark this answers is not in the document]
TimDiekmann: Good point, created H-2876

PR #4540, opened 2024-06-03
H-2735: Allow coordinator agent to initialize sub-coordinators that conduct research on a per-entity basis

benwerner01:

This PR adds a sub-task agent, which the coordinator agent is encouraged to call with one or more sub-tasks which may help it get more relevant facts about discovered entities before they are proposed and returned by the `researchEntities` action.

## 🔗 Related links

- H-2735

## 🔍 What does this change?

- Reports the progress the root coordinator agent has made to the latest user message, rather than keeping this in the system prompt. In my testing this has lead to the agent making better decisions about what to do next, potentially because it more closely inspects things at the end of the message thread - but I'm not exactly sure. The same has been done in the sub-task coordinating agents, but still needs to be done in the worker agent.
- Simplifies the definition of entity types passed to LLMs, by destructuring the top-level properties and not just passing the raw JSON Schema. Something similar should probably be done for link types, and property values (see next ups)
- Changes how the worker agent reports the initial plan (forces usage of the `updatePlan` tool, rather than asking it to return a text message with a plan could result in a non-recoverable error state)
- Switches the coordinator and worker agent to use claude 3 opus, which in testing has lead to better tool calls, and more updates to the plan via the `updatePlan` tool
- fixes an issue where authenticated pages weren't accessible via the worker agent, due to the HEAD `fetch` request used to check the file type at the URL

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Long running research tasks result in many `createEntity` requests being sent to the graph, as each LLM call result in a new usage record entity. This can cause SpiceDB to open too many connections to the datastore. To mitigate this, the maximum datastore connections SpiceDB opens has been increased to 100, and the maximum postgres connections has been increased to 300 in the developer environment. We will need a more robust solution for this to use flows in production (H-2822)

## 🐾 Next steps

- Devise an LLM format for property values (currently using the JSON Schema definitions) (H-2826)
- Devise an LLM format for link entity types (currently using the JSON Schema definitions) (H-2826)
- Move the progress report to the last user message in the worker agent (H-2827)

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

Run a research task which will likely require subtasks to obtain specific information (e.g. "Find 3 of the best graphics cards for running AI models")

## 📹 Demo

<details>

<summary>Outputs from running a “Find 3 of the best graphics cards for running AI models”, which initiates sub-tasks to fill all properties on 3 graphics cards</summary>

image

image (1)

image (2)

image (3)

</details>


libs/@local/advanced-types/src/omit-value.ts
```diff
@@ -0,0 +1 @@
+export type OmitValue<T, K> = T extends K ? never : T;
```
CiaranMn, line 1: There is an inbuilt TS utility called `Exclude` which does this

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/sub-task-agent.ts
```diff
@@ -0,0 +1,725 @@
...
+  }
+
+  const { message } = llmResponse;
+
+  const toolCalls = getToolCallsFromLlmAssistantMessage({ message });
+
+  return { toolCalls };
+};
+
+export const runSubTaskAgent = async (params: {
```
CiaranMn, line 374: Not suggesting we do so now, but how much scope is there for deduplicating code between this and the top-level agent? A lot of it looks familiar but I haven't compared in detail

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/sub-task-agent.ts
```diff
@@ -0,0 +1,725 @@
...
+  const {
+    relevantEntities,
+    existingFactsAboutRelevantEntities,
+    linkEntityTypes,
+  } = params.input;
+
+  return dedent(`
+    You are a sub-task agent for a research task.
+
+    The user will provide you with:
```
CiaranMn, line 143: Does this agent need to know it is a 'sub-task' agent? It otherwise looks like it's completing a research goal the same as any other, and we only mention that it's a sub-task of some wider goal here.

libs/@local/advanced-types/src/omit-value.ts
benwerner01: Ah good point, have removed it in https://github.com/hashintel/hash/pull/4540/commits/53a8c36dce95140a26d168bbbe01a589087bcef9

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/sub-task-agent.ts
benwerner01: There might be opportunities for further deduplication, by re-using the same method for `createInitialPlan` and `getNextToolCalls` in both agents, yes. It would require some additional work though, as the type definition for `State` and `Input` are slightly different, and the prompts adjusted conditionally based on what is available in those objects.

PR #4542, opened 2024-06-04
H-2823: Add serde implementation for non-string hash-map keys

TimDiekmann:

Non-string keys are not allowed in JSON. This adds a serialization to convert keys to string for a `HashMap`.

This PR:

The changes in this PR:

The changes in this PR:


indietyp (approved):
Great stuff!

libs/@local/codec/src/serde/string_hash_map.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// The default serializer for [`HashMap`].
+///
+/// This is only here to allow the use of `#[serde(with = "string_hash_map")]`.
+///
+/// # Errors
+///
+/// This function can return any error that can be returned by serializing a [`HashMap`].
+#[inline]
+pub fn serialize<K, V, H, S>(map: &HashMap<K, V, H>, serializer: S) -> Result<S::Ok, S::Error>
```
indietyp, line 57: Wouldn't that lead to an asymmetrical encoding? What if the key doesn't coerce to a string? Should display or smth be used here?

indietyp (changes requested):
Sorry some overlooked changes looked at the deserialization code first asking for clarification

libs/@local/codec/src/serde/string_hash_map.rs
TimDiekmann: I'm not sure if we want to add the overhead of converting it to a string. The implementation of `HashMap` is already asymmetrical. It's serializing to a string as key, but does not deserialize from a string to e.g. `usize`.

indietyp: I see yea then that's fine. Makes sense

indietyp (approved):

indietyp (dismissed):

indietyp (approved):

PR #4549, opened 2024-06-05
H-2834: Fix Sentry capturing errors from Temporal workflows

CiaranMn:

When bundling workflow code for a Temporal worker, we need to specify any interceptors we are using when bundling the code.

I noticed this issue when looking at logs to debug a Temporal worker not connecting to the server in production, though I do not think it is related (it is just preventing Sentry from capturing errors which are thrown from workflows – it is definitely capturing errors thrown from activities).

```
[WARN] Ignoring WorkerOptions.interceptors.workflowModules because WorkerOptions.workflowBundle is set.
To use workflow interceptors with a workflowBundle, pass them in the call to bundleWorkflowCode.
```

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann (approved):
As you said it's likely unrelated to the issue we encounter but good to have this fixed, thanks!

PR #4560, opened 2024-06-06
H-2787, H-2866, H-2867: Improve LLM rate limit handling

benwerner01:

This PR adds functionality for gracefully handling rate limit errors thrown by the Anthropic and OpenAI APIs. When a rate limit error is encountered, the request is retried after the rate limit window is expired, up to 5 times.

## 🔗 Related links

- H-2787
- H-2866
- H-2867

## 🔍 What does this change?

- modifies the infer facts from text method to no longer infer facts for an entity sequentially in chunks of 3, and instead just make requests for all entities in parallel
- splits the `getAnthropicResponse` and `getOpenAiResponse` methods into separate files, to reduce the size of the `getLlmResponse` file.
- adds a provider for amazon bedrock, which is used for compatible models when a rate limit/throttling error is encountered with the anthropic API

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Some manual testing to inspect the structure of the response headers, and ensure that the way we determine whether an error is a rate limit error is working, but testing the Anthropic API was difficult as often throttling would be the reason a request doesn't exceed - not a rate limit error.

## ❓ How to test this?

Because of what it requires to reach a rate limit, this is difficult to test. To help inspect that the code makes sense, here are what the rate limit headers look like for each provider (logged from real responses encountered locally, not obtained from an online source):

Anthropic:
```JSON
"headers": {
    "anthropic-ratelimit-requests-limit": "4000",
    "anthropic-ratelimit-requests-remaining": "3997",
    "anthropic-ratelimit-requests-reset": "2024-06-06T15:45:07Z",
    "anthropic-ratelimit-tokens-limit": "400000",
    "anthropic-ratelimit-tokens-remaining": "0",
    "anthropic-ratelimit-tokens-reset": "2024-06-06T15:45:07Z",
    "cf-cache-status": "DYNAMIC",
    "cf-ray": "88f985a24aa1bc02-ZRH",
    "connection": "keep-alive",
    "content-length": "428",
    "content-type": "application/json",
    "date": "Thu, 06 Jun 2024 15:44:08 GMT",
    "request-id": "req_01Cm4i9RuEQreA3kSWge7ojW",
    "retry-after": "59",
    "server": "cloudflare",
    "via": "1.1 google",
    "x-cloud-trace-context": "1d3eddfc8ac5647f6680a4eeef9be6cc",
    "x-should-retry": "true"
  }
```
OpenAI:

TODO


apps/hash-ai-worker-ts/src/activities/shared/get-llm-response/get-anthropic-response.ts
```diff
@@ -87,6 +87,18 @@ const convertAnthropicRateLimitRequestsResetTimestampToMilliseconds = ({
...
+
+const isErrorAnthropicThrottlingError = (error: unknown): boolean =>
+  error instanceof APIError &&
+  !!error.status &&
+  /**
+   * @todo: This is a temporary solution until we have a better way to
+   * determine if the error is caused by throttling.
+   */
+  error.status >= 500 &&
+  error.status < 600;
```
vilkinsons, line 100: Technically two attributable/known errors are being caught up here, aswell:

- `500` - `api_error`: An unexpected error has occurred internal to Anthropic’s systems.
- `529` - `overloaded_error`: Anthropic’s API is temporarily overloaded.

Whereas the `520` we're experiencing appears to be originating with Anthropic's use of Cloudflare, and isn't an error code they [profess to throw](https://docs.anthropic.com/en/api/errors) themselves.

That said, I think these two known errors can be handled in the same way (with a jittered retry).

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response/anthropic-client.ts
```diff
@@ -43,9 +41,9 @@ export const anthropicMessageModelToContextWindow: Record<
   AnthropicMessageModel,
   number
 > = {
-  "claude-instant-1.2": 100_000,
-  "claude-2.0": 100_000,
-  "claude-2.1": 200_000,
+  // "claude-instant-1.2": 100_000,
+  // "claude-2.0": 100_000,
+  // "claude-2.1": 200_000,
```
CiaranMn, line 46: Should this just be deleted?

CiaranMn (dismissed):

CiaranMn (approved):

PR #4563, opened 2024-06-06
H-2754, H-2759: Store property metadata as JSON

TimDiekmann:

We decided to change the representation of property metadata to be stored as JSON

## 🚫 Blocked by

- #4534 
- #4536

## 🔍 What does this change?

- Remove the `entity_property` table
- Store property metadata in `entity_editions` table as JSON
- Adjust interface to store metadata as JSON
- Adjust Node API to use new representation
- Adjust the `patch` behavior to take the new representation into account

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -103,6 +104,44 @@ const isGraphApiEntity = <Properties extends EntityPropertiesObject>(
...
+  ): void => {
+    if ("properties" in element) {
+      for (const [key, value] of typedEntries(element.properties)) {
+        visitElement([...path, key], value);
+      }
+    } else if ("elements" in element) {
+      for (const [index, value] of element.elements.entries()) {
+        visitElement([...path, index], value);
+      }
+    }
```
TimDiekmann, line 130: The general handling is easier if `PropertyMetadataElement` has different properties. This PR uses `properties` for objects, `elements` for arrays, and `value` for leafs. Another way would be always using `value` but using `type: "..."`. It's also possible to completely omit `type` but with a runtime cost as `isArray` would need to be used.

TimDiekmann: I tried the `value` approach without `type` and it's not as bad as expected: [`0e3533f` (#4563)](https://github.com/hashintel/hash/pull/4563/commits/0e3533f422f2e93c20c44fa8f9cd0535af0e556c)

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/propose-entities-from-facts/propose-entity-from-facts.ts
```diff
@@ -137,15 +137,17 @@ const generatePropertyMetadata = (params: {
       );
     }
 
-    propertyMetadata.push({
-      path: [baseUrl],
+    if (!propertyMetadata.value) {
+      propertyMetadata.value = {};
+    }
```
CiaranMn, line 142: [nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)
```suggestion
    propertyMetadata.value ??= {};
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -22,27 +20,30 @@ import type {
...
+  PropertyMetadataObject,
+  PropertyPath,
 } from "@local/hash-graph-types/entity";
 import type {
   CreatedAtDecisionTime,
   CreatedAtTransactionTime,
 } from "@local/hash-graph-types/temporal-versioning";
 import type { OwnedById } from "@local/hash-graph-types/web";
-import isEqual from "lodash/isEqual";
+import { isArray } from "lodash";
```
CiaranMn, line 32: This is just `Array.isArray`
```suggestion
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -103,6 +104,46 @@ const isGraphApiEntity = <Properties extends EntityPropertiesObject>(
...
+    path: PropertyPath;
+    metadata: Required<PropertyMetadataElement>["metadata"];
+  }[] = [];
+
+  const visitElement = (
+    path: PropertyPath,
+    element: PropertyMetadataElement,
+  ): void => {
+    if ("value" in element && element.value) {
+      if (isArray(element.value)) {
```
CiaranMn, line 123:
```suggestion
      if (Array.isArray(element.value)) {
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -103,6 +104,46 @@ const isGraphApiEntity = <Properties extends EntityPropertiesObject>(
   );
 };
 
+export const flattenedPropertyMetadataMap = (
```
CiaranMn, line 107: Suggest renaming this for two reasons:
1. It didn't sound like an action before, but more like the result (`flattened -> flatten`)
2. it's not producing what I would call a `Map`... I think fine to just drop the word
```suggestion
export const flattenPropertyMetadata = (
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -238,10 +280,32 @@ export class Entity<
...
+        }
+      } else if (!isArray(map.value)) {
+        return map.value[key];
+      } else {
+        return undefined;
+      }
+    }, this.#entity.metadata.properties)?.metadata;
+  }
+
+  public flattenedProperties(): {
```
CiaranMn, line 304:
```suggestion
  public flattenedPropertiesMetadata(): {
```

CiaranMn (commented):
Thanks Tim, looks good – just a couple of naming/TS comments

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -238,10 +280,32 @@ export class Entity<
...
-    )?.metadata;
+  public propertyMetadata(
+    path: PropertyPath,
+  ): PropertyMetadataElement["metadata"] {
+    return path.reduce<PropertyMetadataElement | undefined>((map, key) => {
+      if (!map || !("value" in map) || !map.value) {
+        return undefined;
+      }
+      if (typeof key === "number") {
+        if (isArray(map.value)) {
```
CiaranMn, line 291:
```suggestion
        if (Array.isArray(map.value)) {
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -238,10 +280,32 @@ export class Entity<
...
+          return map.value[key];
+        } else {
+          return undefined;
+        }
+      } else if (!isArray(map.value)) {
```
CiaranMn, line 296:
```suggestion
      } else if (!Array.isArray(map.value)) {
```

TimDiekmann:
Thanks, @CiaranMn, I applied your suggestions 🙂

CiaranMn (approved):

PR #4580, opened 2024-06-07
H-967: Generally improve linting in Rust code

TimDiekmann:

A few lints (especially auto fixable ones) can be enabled.

## 🔍 What does this change?

- For a more easy setup on lints, we chose to move all packages into the global workspace
- Lints are now defined in the root `Cargo.toml`
- Remove the obsolete lint-config generation script

This PR:

The changes in this PR:

The changes in this PR:


indietyp (commented):
That's was a lot of files but I am excited to see this! Was on my mind as well but never got around to it so glad that it happened! 🎉

(I am currently doing this in my phone which might be why there are some outdated review comments which I will delete)

indietyp (approved):

PR #4585, opened 2024-06-08
H-2880: Use `derive_where` instead of `derivative`

TimDiekmann:

`derive_where` generates more idiomatic code, is easier to write, and supports more features.


indietyp (approved):
LGTM, excited for the change, really like `derive-where`! 🎉

PR #4616, opened 2024-06-17
H-2920: Remove temporary migration to fix incomplete user accounts

CiaranMn:

This reverts commit e774c67fbc23aa269ca69bddc1d3dda5ba38d056 from #4614


infra/terraform/hash/main.tf
```diff
@@ -298,10 +298,6 @@ module "application" {
   ])
   api_image              = module.api_ecr
   api_migration_env_vars = concat(var.hash_api_migration_env_vars, [
-    {
-      name  = "INCOMPLETE_USER_ACCOUNTS", secret = true,
-      value = sensitive(data.vault_kv_secret_v2.secrets.data["incomplete_user_accounts"])
-    },
   ])
```
TimDiekmann, line 301:
```suggestion
  api_migration_env_vars = var.hash_api_migration_env_vars
```

CiaranMn: This is the pattern we have for other env vars around it and is what it was previously, does it serve no purpose?

TimDiekmann: We probably had it because at some point we had inline variables as well. For empty arrays, it doesn't serve any purpose, no. 🙂

TimDiekmann (approved):

PR #4619, opened 2024-06-18
H-2928: Fix clippy errors not shown in CI

TimDiekmann:

The CI did not treat an error in the linting script as real error so it didn't print any diagnostics


indietyp (dismissed):
LGTM! a bit concerning tho that CI just... passed like that

indietyp (approved):

PR #4629, opened 2024-06-21
H-2947: Return full entity from Graph on creation

TimDiekmann:

We expect to pass the metadata alongside the properties, but for now we will use them in the SDK as two separate objects. The Graph already constructs an entity, so a traversal in the Node API can be avoided.

Further, it is expected that the Graph will be capable of using hooks, which might change some properties. As the SDK does not necessarily know what properties have been changed, it's good to inform it.

## 🚫 Blocked by
- #4589 

## 🔍 What does this change?

- Change the return type of `createEntity` and `createEntities` to return the full entities instead of only the metadata

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hash-graph-sdk/typescript/package.json
+++ b/libs/@local/hash-graph-sdk/typescript/package.json
@@ -25,13 +25,11 @@
     "@blockprotocol/graph": "0.3.4",
     "@local/advanced-types": "0.0.0-private",
     "@local/hash-graph-client": "0.0.0-private",
-    "@local/hash-graph-types": "0.0.0-private",
-    "lodash": "4.17.21"
+    "@local/hash-graph-types": "0.0.0-private"
   },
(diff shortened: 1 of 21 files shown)


libs/@local/hash-graph-sdk/typescript/package.json
```diff
@@ -25,13 +25,11 @@
     "@blockprotocol/graph": "0.3.4",
     "@local/advanced-types": "0.0.0-private",
     "@local/hash-graph-client": "0.0.0-private",
-    "@local/hash-graph-types": "0.0.0-private",
-    "lodash": "4.17.21"
```
CiaranMn, line 29: Nice

CiaranMn (approved):

CiaranMn (approved):

PR #4630, opened 2024-06-21
H-2948: Pass metadata alongside properties from Graph SDK to the Graph

TimDiekmann:

The Graph internally uses the properties merged into the metadata object. This directly passes this struct to the Graph.

## 🚫 Blocked by

- #4589 
- #4629 

## 🔍 What does this change?

- Change Graph interface to require the properties with metadata to be passed
- Implement function to create the above structure from the SDK

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Pass metadata alongside properties to Graph SDK (H-2927)


libs/@local/hash-graph-types/typescript/src/entity.ts
```diff
@@ -83,22 +83,56 @@ export type PropertyMetadataElement =
...
+export interface PropertyObjectWithMetadata {
+  value: Record<BaseUrl, PropertyWithMetadata>;
+  metadata?: PropertyMetadataObject["metadata"];
+}
+
+export interface PropertyArrayWithMetadata {
+  value: PropertyWithMetadata[];
+  metadata?: PropertyMetadataArray["metadata"];
+}
+
```
TimDiekmann, line 135: @CiaranMn do you have any opinions on the type names? I think making them consistent everywhere would be good. We currently also have `EntityPropertiesObject` which is the same as `PropertyObjectWithMetadata` but without metadata, so maybe rename those to `PropertyObject`?

libs/@local/hash-graph-types/typescript/src/entity.ts
```diff
@@ -83,22 +83,56 @@ export type PropertyMetadataElement =
...
 export interface PropertyMetadataValue {
   metadata: Omit<ValueMetadata, "dataTypeId"> & {
     dataTypeId?: VersionedUrl;
   };
 }
 
 export interface PropertyMetadataObject {
-  value?: Record<BaseUrl, PropertyMetadataElement>;
+  value: Record<BaseUrl, PropertyMetadataElement>;
```
TimDiekmann, line 107: I made `value` required to distinguish between the variants. `PropertyMetadataValue` does not have `value`.

libs/@local/hash-graph-types/typescript/src/entity.ts
TimDiekmann: As these are used quite a lot I'm going to change the naming in a follow-up to avoid conflicts. We can discuss the names either here or in a later PR. Changing them in the head branch is easy.

CiaranMn: I think fine to remove `Entity`

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -102,6 +114,169 @@ const isGraphApiEntity = <Properties extends EntityPropertiesObject>(
...
+  if (isValueMetadata(metadata)) {
+    return {
+      value: property,
+      metadata: metadata.metadata,
+    } satisfies PropertyValueWithMetadata;
+  }
+
+  if (isArrayMetadata(metadata)) {
+    throw new Error(
+      `Expected metadata to be a value, but got metadata for property array: ${JSON.stringify(
```
CiaranMn, line 237:
```suggestion
      `Expected metadata to be for a value, but got metadata for property array: ${JSON.stringify(
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -102,6 +114,169 @@ const isGraphApiEntity = <Properties extends EntityPropertiesObject>(
...
+        metadata,
+        null,
+        2,
+      )}`,
+    );
+  } else {
+    throw new Error(
+      `Expected metadata to be a value, but got metadata for property object: ${JSON.stringify(
```
CiaranMn, line 245:
```suggestion
      `Expected metadata to be for a value, but got metadata for property object: ${JSON.stringify(
```

libs/@local/hash-graph-types/typescript/src/entity.ts
CiaranMn: I think the fact that we use `value` and `metadata` at different levels for different things might get confusing and I generally like to avoid `something.metadata.metadata`
```ts
const propertyObject: PropertyObjectWithMetadata = {
  value: { } // the property object
  metadata: {
    value: {} // the metadata for the object values
    metadata: {} // the metadata for the object itself
  }
}

const array: PropertyArrayWithMetadata = {
  value: [] // the array
  metadata: {
    value: [] // the metadata for the array elements
    metadata:  // the metadata for the array itself
  }
}
```
Maybe we can rename them?
```ts
export interface PropertyMetadataArray {
  values: PropertyMetadataElement[];
  array?: ArrayMetadata;
}

export interface PropertyMetadataObject {
  values: Record<BaseUrl, PropertyMetadataElement>;
  object?: ObjectMetadata;
}
```

TimDiekmann: We don't have any nested metadata, e.g.:
```typescript
{
  value: {
    "https://.../top-level-property-1/": { // Top level property with leaf value
      value: "foo", // Leaf value, only present if the properties are embedded into the object
      metadata: { // metadata of top-level-property-1
        dataTypeId: ...
        provenance: ...
        ...
      }
    },
    "https://.../top-level-property-2/": { // Top level property with nested properties
      value: {
        "https://.../nested-in-property-2/": { // Nested property
          value: "bar",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        }
      },
      metadata: { // metadata of top-level-property-2
        provenance: ...
        ...
      }
    },
    "https://.../top-level-property-3/": { // Top level property with array of values
      value: [
        { // first element
          value: "element 1",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        },
        { // second element
          value: "element 2",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        }
      ],
      metadata: { // metadata of top-level-property-3
        provenance: ...
        ...
      }
    },
  },
  metadata: { // metadata for whole property object
    provenance: ...
    ...
  }
}
```
The difference between `PropertyObjectWithMetadata` and `PropertyMetadataObject` is that the leaf values also contains a `value`.

We, however, could instead use 
- `value` for leaf values as defined by data types
- `array` for array properties
- `object` for nested properties, so the above would become:

```typescript
{
  object: {
    "https://.../top-level-property-1/": { // Top level property with leaf value
      value: "foo", // Leaf value, only present if the properties are embedded into the object
      metadata: { // metadata of top-level-property-1
        dataTypeId: ...
        provenance: ...
        ...
      }
    },
    "https://.../top-level-property-2/": { // Top level property with nested properties
      object: {
        "https://.../nested-in-property-2/": { // Nested property
          value: "bar",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        }
      },
      metadata: { // metadata of top-level-property-2
        provenance: ...
        ...
      }
    },
    "https://.../top-level-property-3/": { // Top level property with array of values
      array: [
        { // first element
          value: "element 1",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        },
        { // second element
          value: "element 2",
          metadata: { // metadata of nested-in-property-2
            dataTypeId: ...
            provenance: ...
            ...
          }
        }
      ],
      metadata: { // metadata of top-level-property-3
        provenance: ...
        ...
      }
    },
  },
  metadata: { // metadata for whole property object
    provenance: ...
    ...
  }
}
```

CiaranMn: What is `PropertyMetadataObject["value"]` being used for?

CiaranMn: Can we add comments to these types explaining what they are? This would help

TimDiekmann: It's here to distinguish between `metadata` and inner values. It would be possible to flatten the inner properties but that does not work for arrays, so I chose to use `value` for both. Also, it's easier to provide better errors when deserializing malformed objects and forbidding any non-allowed keys (flattening the `value` basically means using `additionalProperties: true`.

TimDiekmann: > Can we add comments to these types explaining what they are? This would help

Yes, absolutely!

libs/@local/hash-graph-types/typescript/src/entity.ts
TimDiekmann: I consistently renamed everything (Rust and Javascript - but I left BP untouched)

libs/@local/hash-graph-types/typescript/src/entity.ts
CiaranMn: We spoke about this, the confusion came from there being two representations:
1. When the metadata is defined separately (`properties` object just contains values) – this is `PropertyMetadataObject` / `PropertyMetadataArray` – this representation is only the type for input at the moment, e.g. create and update, although in practice we don't yet actually use it
2. When the metadata appears under `metadata` alongside the value under `value` – this is the other types, as discussed above, and is what we pass to the graph (but is not exposed to any consumer)

Let's leave it for now as it's going to change – we can add comments and clarify as we go

CiaranMn (approved):

CiaranMn (approved):

PR #4639, opened 2024-06-25
H-2950: Simplify coordinator progress report and entity submission

CiaranMn:

This PR refactors the coordinator agent to simplify its task and reduce token usage by:
1. Replace the current process whereby it:
- (1) receives facts, (2) decides which facts to turn into entities with a tool call, (3) decides which entities to submit as 'for return' with a tool call, (4) decides when to declare the task complete with a tool call (at which point the previous submitted entities are returned)
with a process whereby it 
- (1) receives entities which have automaticallly been created from the available facts, and (2) declares the job complete with a tool call, specifying which of the available entities to return at the same time

The intention of this is to:
- reduce chance for error by reducing the number of decisions it has to make
- reduce token usage by it not needing it to see the raw facts
- reduce token usage and time by reducing the number of iterations it goes through to conclude a task

2. Give it the history of things done and discovered in a single progress report message, instead of having those represented as a history of tool calls made and responses received – the only tool calls it now sees in the history are those which have just errored.

Drive-bys:
1. Move any Q&A made to the user to the user's first message instead of the system prompt
2. Add retry to calls to the internal API so that the research task doesn't error out if there's a dropped call to it
3. Some tweaks to prompts to account for regressions in fact gathering performance when switching from Opus to Haiku
4. Don't throw an error if the fact gathering request reaches its max retries (which kills the whole task), instead return any valid facts gathered to that point
5. add the task name to the end of the individual LLM response log to make it easier to scan for logs of interest

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None yet

## ❓ How to test this?

1. Run a Flow


apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action.ts
```diff
@@ -146,43 +165,103 @@ const updateStateFromInferredFacts = async (params: {
...
+    ...state.proposedEntities.filter(
+      ({ localEntityId }) =>
+        !newProposedEntities.some(
+          (newProposedEntity) =>
+            newProposedEntity.localEntityId === localEntityId,
+        ),
+    ),
+    ...newProposedEntities,
+  ];
 };
```
benwerner01, line 265: Just want to flag, unless I've missed something I don't think we are filtering out previously proposed entities, which have been provided with new `localEntityId`s after the deduplication process (may have been marked as a non-canonical duplicate of an entity summary which was obtained in the worker agent).

So we may end up with multiple proposals for the same entity, with different `localEntityId`s. We could update the deduplication logic to also prune proposals which have been deduplicated, or replace their `localEntityId` so they will be filtered out here.

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action/coordinating-agent.ts
```diff
@@ -110,18 +111,15 @@ const generateSystemPromptPrefix = (params: {
     Don't start sub-tasks in parallel which duplicate or overlap, or where one will depend on the result of another (do it in sequence).
     For simpler research tasks you might not need sub-tasks.
 
-    Do not waste resources and time trying to produce a table, report or any other form of output that isn't the proposed entities with all their properties.
-    
-    ${questionsAndAnswers ? `You previously asked the user clarifying questions on the research brief provided below, and received the following answers:\n${questionsAndAnswers}` : ""}
-
-    The "complete" tool for completing the research task will only be available once you have submitted
-      proposed entities that satisfy the research prompt.
+    The "complete" tool for completing the research task will only be available once you have gathered facts.
```
benwerner01, line 114: Technically the "complete" tool is only available once entities have been discovered, should this maybe say so instead?
```suggestion
    The "complete" tool for completing the research task will only be available once entities have been discovered.
```

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/infer-facts-from-text/infer-entity-facts-from-text.ts
```diff
@@ -110,6 +110,11 @@ const systemPrompt = dedent(`
 
   These facts will be later used to construct the entities with the properties and links which the user will specify.
   If any information in the text is relevant for constructing the relevant properties or outgoing links, you must include them as facts.
+  
+  Each fact should be in the format <subject> <predicate> <object>, where the subject is the singular subject of the fact.
+  Example:
+  [{ text: "Company X acquired Company Y.", prepositionalPhrases: ["in 2019", "for $10 million"], subjectEntityLocalId: "abc", objectEntityLocalId: "123" }]
+  Don't include facts which start with a subject you can't provide an id for.
```
benwerner01, line 117: I'm sure you're aware, but would be interesting to see if this is any more impactful when inserted as a synthetic example (user and assistant message), where the user message includes some text including this fact about companies X and Y.

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/infer-facts-from-text/infer-entity-facts-from-text.ts
```diff
@@ -278,9 +283,9 @@ export const inferEntityFactsFromText = async (params: {
       ],
     });
   } else if (llmResponse.status !== "ok") {
-    throw new Error(
-      `Failed to get response from LLM: ${stringify(llmResponse)}`,
-    );
+    return {
+      facts: params.retryContext?.previousValidFacts ?? [],
+    };
```
benwerner01, line 288: You'll see when resolving conflicts, but I've addressed this by retrying the request in the agent here rather than returning the existing facts. This is to account for scenarios where the first response from the LLM leads to a schema validation error that it cannot recover from (which is possible with `haiku`), in a way that doesn't result in 0 facts being observed as easily.

benwerner01 (commented):
Thanks Ciaran, based on the code these changes look good and definitely should simplify the decisions/tasks for the coordinator agent. Have flagged a few things that I spotted which may be worth addressing.

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action.ts
CiaranMn: I think this is happening [here](https://github.com/hashintel/hash/blob/07812d8b86d9d3964493860b828dac259d9462f9/apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action.ts#L177), although I may be misunderstanding.

Worth also flagging that I changed the logic there to remove any proposed entities from state which had been marked as a duplicate, rather than overwriting their `localEntityId` (which would mean two different couples of the entity with the same `localEntityId` in state, possibly with different properties) – it is quite possible I've missed/misunderstood something about this though.

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/infer-facts-from-text/infer-entity-facts-from-text.ts
CiaranMn: Yes – I want to experiment more with examples here, as it's one of the places I expect we will need to focus on improvements most.

apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action.ts
benwerner01: > I think this is happening [here](https://github.com/hashintel/hash/blob/07812d8b86d9d3964493860b828dac259d9462f9/apps/hash-ai-worker-ts/src/activities/flow-activities/research-entities-action.ts#L177), although I may be misunderstanding.

Ah yes, this does address the issue I flagged.

> I changed the logic there to remove any proposed entities from state which had been marked as a duplicate, rather than overwriting their localEntityId (which would mean two different couples of the entity with the same localEntityId in state, possibly with different properties)

Okay, that works but overwriting the `localEntityId` should also work, as we then deduplicate those previously proposed entities based on their `localEntityId` here. The only benefit I can think of is that if there is some error encountered re-proposing the entity, we don't lose the prior proposal. But that may not be something worth accounting for.

vilkinsons (approved):

PR #4646, opened 2024-06-27
H-2251, BP-89: Remove temporal / non-temporal split from `@blockprotocol/graph`

TimDiekmann:

Exporting both temporal and non-temporal versions of everything makes the package hard to navigate, the types more complicated, and means we need a more complicated bundling step.

Since we aren't bothered about supporting non-temporal applications for future versions of the library, we should just make everything temporal.

## 🔍 What does this change?

- Bump the used versions of `@blockprotocol/graph` and `@blockprotocol/type-system` to use in-tree version
- Fix `type-system` crate to export correct TS types
- Remove re-export from `@blockprotocol/type-system` in `@blockprotocol/graph` and account for this in downstream packages
- The in-tree version of the type-system does not suffer from BP-89.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- BP-101: The block scripts package depends on the published version of `@blockprotocol/graph`. The copied-over `@blockprotocol/graph` version has some changes which have to be taken into account when running `block-scripts`, e.g. codegen should not have a temporal parameter and always generates the temporal version.
- BP-102: For the temporal variant of links the `linkEntity` and `rightEntity` should be an array
- H-2981: When the type system crate changed turbo does not recognize changes and use the cached version of the generated dist.


package.json
```diff
@@ -37,7 +37,7 @@
     "fix:yarn-deduplicate": "yarn install && yarn-deduplicate --strategy=fewer && yarn install",
     "lint": "npm-run-all --continue-on-error \"lint:*\"",
     "@TODO.2": "Upgrade or remove these blocks and remove the --ignore-package options (also @TODO.1)",
-    "lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-dep=@blockprotocol/graph --ignore-dep=@blockprotocol/type-system --ignore-package=@apps/hashdotdev --ignore-package=@blocks/chart --ignore-package=@blocks/embed --ignore-package=@blocks/person",
+    "lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-dep=@blockprotocol/graph --ignore-package=@apps/hashdotdev --ignore-package=@blocks/embed --ignore-package=@blocks/person",
```
CiaranMn, line 40: The same change needs to be made to `fix:`

CiaranMn (commented):
Whoo! Thanks @TimDiekmann. Just noticed one minor script change

package.json
TimDiekmann: Whoops, thanks for flagging! [`22b41a4` (#4646)](https://github.com/hashintel/hash/pull/4646/commits/22b41a4d6992454bfaac53ceeb7017b06fd9cb0f)

CiaranMn (approved):

PR #4650, opened 2024-06-27
H-2967: Update type `codegen` to generate `XWithMetadata` types

CiaranMn:

Entity property values can have metadata assigned to them (confidence, sources), and leaf data type values also need to have a `dataTypeId` specified – in order to distinguish between different data types of the same underlying primitive, e.g. a `number`, but with a different semantic meaning, where multiple are valid.

When creating or updating entities we now pass the value and metadata together in a single object (see #4632, #4631). 

This PR updates the script which generates TypeScript types from type system schemas to create those objects, so that we have 'value with metadata' types which are constrained to only the supported shape (e.g. nested arrays, objects) and data type ids permitted by those properties.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

If you run `yarn generate-system-types`, it will error when generating the `@linear` namespace, and error in different places if you run different combinations of multiple namespaces sequentially. If you comment them out to only generate one at a time, it works. I don't know why this is because they should not share any state. H-3027.

## 🛡 What tests cover this?

- None.

## ❓ How to test this?

1. Look at the newly generated types and check they are correct.
`libs/@local/hash-isomorphic-utils/src/system-types/shared.ts` has a bunch


libs/@blockprotocol/graph/src/codegen/context/postprocess.ts
```diff
@@ -30,14 +35,66 @@ export type IdentifierSources =
...
+      definingPath: "@local/hash-graph-sdk/entity",
+    },
+  },
+  Confidence: {
+    locallyImportable: true,
+    source: {
+      kind: "external",
+      definingPath: "@local/hash-graph-types/entity",
+    },
+  },
```
CiaranMn, line 62: Ideally we'd have this exported from the HASH graph client. I had to add this because it was the simplest way of having a type we can import and not be duplicated in the generated types (the underlying library we are using to generate them doesn't allow things like `PropertyProvenance["confidence"]`)

libs/@blockprotocol/graph/src/codegen/initialize/metadata/generate-metadata-schema.ts
```diff
@@ -0,0 +1,246 @@
...
+    kind,
+    properties: {
+      metadata: objectMetadataSchema,
+      value: {
+        type: "object",
+        properties: propertiesSchema,
+        required,
+      },
+    },
+    required: ["value"],
```
CiaranMn, line 182: I have made `value` required in all of these although it does raise the question of how you would update metadata without updating `value`. Maybe we don't want to do that.

TimDiekmann: That's fine for now I guess. For `PropertyWithMetadata` it's currently required as well. If `value` is not required, it wouldn't be possible to distinguish between an object and an array as the metadata fully overlap.

libs/@blockprotocol/graph/src/codegen/initialize/metadata/generate-metadata-schema.ts
```diff
@@ -0,0 +1,246 @@
...
+    $id,
+    kind,
+    properties: {
+      metadata: objectMetadataSchema,
+      value: {
+        type: "object",
+        properties: propertiesSchema,
+        required,
+      },
+    },
```
TimDiekmann, line 181: Is this the location you were wondering if we want `value` in the first place? So if we want `value` at the top level?

libs/@local/hash-isomorphic-utils/src/system-types/linear/attachment.ts
```diff
@@ -310,37 +490,65 @@ export type BelongsToIssueProperties1 = LinkProperties;
...
 
+export type ObjectDataTypeWithMetadata = {
+  value: ObjectDataType;
+  metadata: ObjectDataTypeMetadata;
+};
+export type ObjectDataTypeMetadata = {
+  provenance?: PropertyProvenance;
+  confidence?: Confidence;
+  dataTypeId: "https://blockprotocol.org/@blockprotocol/types/data-type/object/v/1";
+};
```
TimDiekmann, line 526: Should this definition live in this file?

libs/@blockprotocol/graph/src/codegen/initialize/metadata/generate-metadata-schema.ts
CiaranMn: But either (a) there is already a `value`, in which case we can tell, or (b) there is no value (maybe confidence in the absence of a property?), in which case there's nothing to distinguish

libs/@blockprotocol/graph/src/codegen/initialize/metadata/generate-metadata-schema.ts
CiaranMn: Yes, because I'm not sure what it means when applied to the whole entity (as compared to a nested object, where it could be taken to mean 'existince of this thing'). But it's not a big deal, fine to be consistent to start

libs/@local/hash-isomorphic-utils/src/system-types/linear/attachment.ts
CiaranMn: How the codegen file allocation is defined is that if a type is used by more than one file, it is moved into a `shared.ts`. Within the scope of the `linear` types, `ObjectDataType` is only used by `attachment`, which is why it stays here.

We can change that logic if we want, e.g. _always_ move data types into a `shared.ts` file, although it might be weird for very specific data types that clearly 'belong' to some entity type.

CiaranMn: We could also update the codegen so that we pass all the different namespace types at once, rather than as separate calls to `codegen`, in which case it's more likely that things get used by multiple entity types.

TimDiekmann: > Within the scope of the linear types, ObjectDataType is only used by attachment, which is why it stays here.

I think it's fine to leave it as it is then

TimDiekmann (approved):
Amazing, thanks @CiaranMn!

PR #4654, opened 2024-06-28
H-162, BP-74, BP-89: Simplify type-system definitions

TimDiekmann:

Various improvements around the type-system crate

## 🔍 What does this change?

- Rename `Array` to `ArraySchema`, `Object` to `ObjectSchema`, and `OneOf` to `OneOfSchema`
- Simplifies usage of ontology types in Rust (making the fields public)
- Remove link ordering - for real this time
- Fix the `oneOf` type
- Omit `oneOf` and `AllOf` if empty
- Simplify `Tsify` definitions as they are not needed anymore

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/util/upgrade-entity-type-dependencies.ts
```diff
@@ -55,18 +55,17 @@ export const upgradeEntityTypeDependencies = ({
...
         ...linkSchema,
-        items: {
-          ...linkSchema.items,
-          oneOf:
-            "oneOf" in linkSchema.items
-              ? linkSchema.items.oneOf.map((reference) =>
+        items:
+          "oneOf" in linkSchema.items
+            ? {
+                oneOf: linkSchema.items.oneOf.map((reference) =>
```
CiaranMn, line 61: I had a look at this and preserving array length for mapping over a tuple might be fixed at some point in the not too distant future, so I suggest we expect an error instead which will flag up when we can remove the assertion. I can't suggest directly over deleted lines but here's the formatted change required:
```ts
                oneOf: linkSchema.items.oneOf.map(
                  (reference) =>
                    replaceEntityTypeReference({
                      reference,
                      upgradedEntityTypeIds,
                    }),
                  // @ts-expect-error –– @see https://github.com/microsoft/TypeScript/issues/29841
                ) satisfies [EntityTypeReference, ...EntityTypeReference[]],
```

CiaranMn (commented):
One minor comment, looks good otherwise!

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/util/upgrade-entity-type-dependencies.ts
TimDiekmann: This also implies an error a bit further down:
```
      // @ts-expect-error –– @see https://github.com/microsoft/TypeScript/issues/29841
      accumulator[linkTypeId] = schemaWithUpdatedDestinations;
```

Do we still want to add the `@ts-expect-error`?

CiaranMn: Hmmm... okay, no, let's not. That leads to less type safety than the assertion you had before (I didn't get the new error until I ran `codegen` again)

TimDiekmann: I added an `atLeastOne` function which we can remove again once the issue is solved. I chose that way because the `allOf` was also wrongly typed and does not match the type the Graph is generating.

CiaranMn: Cool, thank you!

CiaranMn (approved):

PR #4664, opened 2024-07-01
H-3044: Improve ontology type id typings

CiaranMn:

This PR is spun out of work I'm doing on improving entity property updates, and improves the typing of our generated "ontology type ids" objects by correctly branding the property base URLs as `BaseUrl`, removing the need to extract or assert them as being such everywhere.

## 🔍 What does this change?

- Brand base URLs in the generated ids
- Remove `extractBaseUrl` and `as BaseUrl` where we need to use those base urls

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `tsc`

### DIFF
--- a/apps/hash-api/src/generate-ontology-type-ids.ts
+++ b/apps/hash-api/src/generate-ontology-type-ids.ts
@@ -194,7 +194,6 @@ const generateOntologyIds = async () => {
 
   const graphContext: ImpureGraphContext = {
     graphApi,
-    provenance: { actorType: "human", origin: { type: "migration" } },
     temporalClient: null,
   };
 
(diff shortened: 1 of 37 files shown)


apps/hash-api/src/generate-ontology-type-ids.ts
```diff
@@ -7,6 +7,11 @@ import { createGraphClient } from "@local/hash-backend-utils/create-graph-client
 import { getRequiredEnv } from "@local/hash-backend-utils/environment";
 import { Logger } from "@local/hash-backend-utils/logger";
 import { publicUserAccountId } from "@local/hash-backend-utils/public-user-account-id";
+import type {
+  DataType,
+  EntityType,
+  PropertyType,
+} from "@local/hash-graph-client";
```
TimDiekmann, line 14: We can import these from `@blockprotocol/type-system`

CiaranMn: We can, but does it matter? We import them inconsistently across the codebase, is there a difference?

We should enforce a specific import if there's a reason we shouldn't import them from here

CiaranMn: We will make this consistent, I will separately add an eslint rule

TimDiekmann (dismissed):
Thanks, looking good!

I spotted some linting failures, happy to re-approve as soon as they're fixed.

TimDiekmann (approved):

PR #4678, opened 2024-07-05
GEN-68: Capture source errors when converting `Error` to `Report`

### LINKED ISSUE #2127: GEN-68: `.into_report()` doesn't include error source chains
### Related Problem

The problem I have seen is that when calling `.into_report()` on errors, `Report` only includes the top-most error, and does not include any of the sources.

### Proposed Solution

I propose that `.into_report()` go through the source chain and add each source entry as a printable attachment, (or context frame?) to the error.

### Alternatives

_No response_

### Additional context

I've run into this problem using the Rust AWS SDK, where `SdkError` results in generic `"service error"` messages, where in reality, if we were to follow the error source chain, we would get all the information we'd need to diagnose and fix the error.

TimDiekmann:

It's desired that source errors of `Error` are printed as part of a `Report` as well.

## 🔗 Related links

- Closes #2127

## 🔍 What does this change?

- Adds a hidden `__source` method to `Context` to close the gap between `Context` and `Error`. This allows using `Error::source` without specialization.
- Add contexts for sources returned from `Error::source` to the `Report` when creating it

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,7 +289,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.68",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


indietyp (approved):
LGTM, thank you!

Issue #3353, opened 2023-10-08
GEN-115: Way to disable error location and/or backtrace


SUPERCILEX:
### Proposed Solution

It'd be nice to have a cargo feature available to disable location capturing/printing. Example:

```
Error: File or directory not found: "misclassified.png"
├╴at rmz/src/main.rs:52:17
╰╴Use --force to ignore.
```

For end users, the `at rmz/src/main.rs:52:17` is just noise so ideally I'd be able to keep that stuff on in debug mode and feature flag it off in release mode.

Similarly, it'd be nice to completely compile out backtrace support if the `backtrace` feature isn't enabled in the stdlib: https://doc.rust-lang.org/cargo/reference/unstable.html#build-std-features

TimDiekmann:
It's possible to disable the `location` by setting a custom debug hook for `Location`:
```rust
error_stack::Report::install_debug_hook::<Location>(|_, _| {});
```

For more information about the debug hooks please see [the corresponding documentation](https://docs.rs/error-stack/0.4.1/error_stack/struct.Report.html#method.install_debug_hook).

Related discussion:
- #1215
---

Backtraces are only shown if you run your program with backtraces enabled, typically this is done by [setting `RUST_BACKTRACE=1`](https://doc.rust-lang.org/stable/std/backtrace/index.html#environment-variables). No backtraces will be shown if none were captured.

SUPERCILEX:
Thanks for the debug hook solution!

For backtraces, I'm asking for support to be compiled out completely. `RUST_BACKTRACE=1` won't work with `panic_immediate_abort` so having it in the binary is a waste.

TimDiekmann:
Hmm, I haven't thought about `panic_immediate_abort` (and also only used that in very few contexts so far). But as this is typically used in contexts were binary size do in fact matter it's ag good suggestion. Is there an easy way to detect certain features or is there a dedicated compile-test required in `build.rs`?

SUPERCILEX:
I'm not sure unfortunately, though I think the easiest way to go about this is to have a release mode and all modes feature flag (like the log crate) that compiles out support.

PR #4685, opened 2024-07-06
GEN-115: Don't compile backtrace support if feature is disabled

TimDiekmann:

In some contexts, such as with `panic_immediate_abort` set, backtraces should not be compiled into the binary. Currently, the `std` feature determines if backtraces are enabled or not. Making this a dedicated feature flag helps reduce the binary size.

## 🔗 Related links

- Closes https://github.com/hashintel/hash/issues/3353

## 🔍 What does this change?

- Guard `Backtrace` behind the `backtrace` feature

### DIFF
--- a/libs/error-stack/.justfile
+++ b/libs/error-stack/.justfile
@@ -6,17 +6,16 @@ set fallback
 default:
   @just usage
 
-# TODO: add `--ignore-unknown-features` to `cargo hack` and pass `--workspace`
 cargo-hack-groups := '--group-features eyre,hooks --group-features anyhow,serde'
 profile := env_var_or_default('PROFILE', "dev")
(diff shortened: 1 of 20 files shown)


indietyp (dismissed):

indietyp (approved):

tisonkun:
May we release a 0.5.0 for this change?

TimDiekmann:
I need to check for breaking changes and check if the MSRV did not change. Then we can file a new release, yes! 🙂 
I probably get to this next week.

TimDiekmann:
[Part of `error-stack@0.5.0`](https://github.com/hashintel/hash/releases/tag/error-stack%400.5.0)

PR #4688, opened 2024-07-08
BP-33: Distinguish between format validation and content validation

TimDiekmann:

Currently, some format validation is done when deserializing the data, some is done when converting between the raw representation and the abstract representation. The format should be validated when deserializing, the content validation should be done when converting it to the non-raw representation. If the content validation would happen when deserializing, the interface would become hard to deal with as the fields cannot be made pub.

The issue with the current approach is that it's quite hard to deal with the two layers, the generated types are hard to deal with, and adding more validated values to the ontology types is quite hard as well.

## 🔍 What does this change?

- Remove the global `raw` module (only use a local `raw` module to define the serialized form)
- Define validations as a separate step - decoupled from deserialization
- Only allow valid types to be inserted into the database (enforced by `Valid<T>`)

The new validation approach is extendable and can be used to fully validate ontology types and entities, but this is out of scope of this PR.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-isomorphic-utils/src/util.ts
```diff
@@ -14,6 +14,13 @@ export const isUnknownObject = (
 ): x is { [key in PropertyKey]: unknown } =>
   x !== null && typeof x === "object";
 
+/**
+ * Ensures that the array has at least one element. If not it returns `undefined`.
+ */
+export const atLeastOne = <T>(array: T[]): [T, ...T[]] | undefined =>
+  // @ts-expect-error –– @see https://github.com/microsoft/TypeScript/issues/29841
+  array.length > 0 ? (array satisfies [T, ...T[]]) : undefined;
+
```
CiaranMn, line 23: Is this supposed to be here? It also already exists in `type-system`

TimDiekmann: Whoops, that's a rebase error. Fixed in [`f384e79` (#4688)](https://github.com/hashintel/hash/pull/4688/commits/f384e79d94c6a317a81672eb4f443239b584cc5c)

CiaranMn (approved):

PR #4689, opened 2024-07-08
H-2975: Make all data type constraints strongly typed

TimDiekmann:

Currently, we use `additionalProperties: true`, instead, we need strongly typed constraints.

## 🚫 Blocked at

- #4688 

## 🔍 What does this change?

- Make the constraints typed
- Use validation based on that
- fix issues that `exclusiveMaximum` and `exclusiveMinimum` is a boolean, not a number

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/entity_type/raw.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/entity_type/raw.rs
@@ -2,6 +2,7 @@ use alloc::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 
 use serde::{Deserialize, Serialize};
+use serde_json::Value as JsonValue;
 #[cfg(target_arch = "wasm32")]
 use tsify::Tsify;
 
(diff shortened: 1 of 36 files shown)


libs/@blockprotocol/type-system/rust/src/schema/entity_type/raw.rs
```diff
@@ -66,7 +67,7 @@ pub struct EntityType<'a> {
     #[serde(with = "links", default, skip_serializing_if = "HashMap::is_empty")]
     links: Links,
     #[serde(default, skip_serializing_if = "<[_]>::is_empty")]
-    examples: Cow<'a, [HashMap<String, serde_json::Value>]>,
+    examples: Cow<'a, [HashMap<BaseUrl, JsonValue>]>,
```
CiaranMn, line 70: We were going to remove `examples`, right? Is there a ticket for that?

TimDiekmann: Yes: BP-63
I wanted to remove it as part of this PR but some more parts had to be adjusted for that. Also, I need to check if the production DB is using `examples` anywhere. I'm going to do that in a follow-up.

CiaranMn (approved):

PR #4699, opened 2024-07-11
H-3074, H-3076: Accept generics in `Entity.create`, new utility types in codegen

CiaranMn:

Improves the typing of things returned from the Entity SDK by:
1. having `Entity.create` accept a generic which narrows the `properties` and `entityTypeId` of the created entity to a specific type (e.g. `User`, `Document`), and enforces the input `properties` and `entityTypeId` to match
2. the same for `Entity.createMultiple`, which will 
3. generate new utility types in the TS system types to support this
4. have `EntityRootType` in a `Subgraph` also accept this generic, to narrow the type of the roots of the subgraph to a specific entity type (for use when we the subgraph is the result of a query which limits the roots to a given type(s)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- H-3091: `entity.patch` will return `this`, i.e. whatever type the `entity` was instantiated as – but this is not accurate if the `entityTypeId` is changed as part of the input parameters. I have tried a couple of ways of enforcing a new type to be specified if the `entityTypeId` is changed as part of the input, but couldn't get it to work.
- At least for me, IntelliJ is not flagging missing `properties`, but `tsc` is catching it – unsure why, but as long as the compiler catches it it's alright (slightly worse DX than if the editor did, but better than the previous situation which was properties being unchecked)

Screenshot 2024-07-11 at 11 55 22

## 🛡 What tests cover this?

- `tsc`

## 📹 Demo

1. Typed return from `create`
Screenshot 2024-07-11 at 11 51 31

2. Preserved in `patch`
Screenshot 2024-07-11 at 11 51 16

3. `createMultiple` – tuple
Screenshot 2024-07-11 at 11 49 42

4. input type safety on tuple
Screenshot 2024-07-11 at 11 52 29

5. Mixed type array
Screenshot 2024-07-11 at 11 50 51

6. Single type array
Screenshot 2024-07-11 at 11 51 01

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/flow-activities/process-automatic-browsing-settings-action.ts
+++ b/apps/hash-ai-worker-ts/src/activities/flow-activities/process-automatic-browsing-settings-action.ts
@@ -1,3 +1,4 @@
+import type { Entity } from "@local/hash-graph-sdk/entity";
 import {
   getSimplifiedActionInputs,
   type OutputNameForAction,
@@ -49,7 +50,7 @@ export const processAutomaticBrowsingSettingsAction: FlowActionActivity =
     }
 
     const automaticInferenceConfig = (
-      userBrowserPluginSettings as BrowserPluginSettings
+      userBrowserPluginSettings as Entity<BrowserPluginSettings>
     ).properties[
       "https://hash.ai/@hash/types/property-type/automatic-inference-configuration/"
     ];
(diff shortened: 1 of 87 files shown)


TimDiekmann (approved):
Great work, thanks!

I pushed a minor fix to satisfy ESLint and TSC.

PR #4708, opened 2024-07-12
H-3097: Use `error-stack@0.5.0`

TimDiekmann:

We want the source errors of non-reports covered in reports, so we update to the latest `error-stack` the smart guys from HASH just released 🎉 

## 🚫 Blocked by

- #4706


vilkinsons (approved):
Approving in anticipation of https://github.com/hashintel/hash/pull/4706 merging. We should avoid setting automerge until then.

TimDiekmann:
> Approving in anticipation of #4706 merging. We should avoid setting automerge until then.

Would have been wild if merging would have worked before 😄

vilkinsons (approved):

PR #4737, opened 2024-07-17
Impl AsRef for Report

DoumanAsh:

Adds AsRef to access current context more conveniently in generic context

## 🔗 Related links

---

## 🚫 Blocked by

---

This PR:

no

The changes in this PR:

no

The changes in this PR:

## 🛡 What tests cover this?

-

## ❓ How to test this?

why would you test AsRef

### DIFF
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -738,3 +738,10 @@ impl<Context> Extend<Self> for Report<Context> {
         }
     }
 }
+
+impl<C> AsRef<C> for Report<C> {
+    #[inline(always)]
+    fn as_ref(&self) -> &C {
+        self.current_context()
+    }
+}


indietyp:
I agree; I'd like to add that the [documentation of `AsRef`](https://doc.rust-lang.org/beta/core/convert/trait.AsRef.html) states also states:

> Used to do a cheap reference-to-reference conversion.
>
> This trait is similar to [AsMut](https://doc.rust-lang.org/beta/core/convert/trait.AsMut.html) which is used for converting between mutable references. If you need to do a costly conversion it is better to implement [From](https://doc.rust-lang.org/beta/core/convert/trait.From.html) with type &T or write a custom function.

`current_context` in this context is quite heavy because, as Tim mentioned, we're not just doing a pointer-to-pointer conversion; we need to (potentially) follow multiple pointer indirections in the frame tree and then downcast the current context. Something costly. So, we'd be breaking expectations about how `AsRef` implementations are supposed to work.

TimDiekmann:
> we need to (potentially) follow multiple pointer indirections in the frame tree and then downcast the current context.

Yes, there are quite a few indirections:
- We store the frames in `Box<Vec<Frame>>` to reduce the memory footprint for `Ok` values
    - This implies two indirections from `Report` to the first `Frame`
- Frame stores `Box<dyn FrameImpl>`, which is another indirection to get `FrameImpl`
- Calling `downcast_ref` on `Frame` implies a virtual call to get `dyn Any`
- then, `downcast_ref` on `Any` is called

PR #4738, opened 2024-07-17
H-3124: Update CI Slack notifications

vilkinsons:

Updates our Slack notifications in CI to reflect current areas of ownership.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -397,7 +397,7 @@ jobs:
         if: ${{ failure() && github.event_name == 'merge_group' }}
         env:
           SLACK_LINK_NAMES: true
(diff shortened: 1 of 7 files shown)


vilkinsons:
@CiaranMn Some of these I understand but are others not overly noisy?

CiaranMn:
> @CiaranMn Some of these I understand but are others not overly noisy?

Which ones? The general principle is to get notifications of things we wouldn't otherwise realise without manually checking:

1. If an action fails in the merge queue (no comment is added to the PR) – merge queue is supposed to be a 'fire and forget' thing, and if something gets kicked out of the merge queue we don't know until we go and look at why the PR hasn't merged
2. If an action run on merge to `main` fails, we don't know unless we look at the actions

CiaranMn (approved):

PR #4753, opened 2024-07-23
H-3153: Implement HQL CST

indietyp:

This implements the HQL CST (Concrete Syntax Tree) from serde.

I have investigated several ways to parse JSON into the expression tree, these include
* `justjson` parser API, not used as it didn't provide enough flexibility
* `justjson` token stream (used by deer)
* `serde_json`.

I concluded that while the token stream would allow us to implement fault tolerant parsing and what not, it would mean a spike in complexity (as we would need to ensure correct output). Instead, I decided to use the serde data model (which was quite simple) and created a set of deserializers and visitors. This seems to work great and has the advantage that we don't have to test if we are parsing everything correctly etc (like trailing commas and what not) but comes at the cost of us not being able to provide any meaningful location information, tolerant parsing or multiple errors at once. Although this could easily be fixed in a later iteration (tracking in H-3154). For now, this should be good enough.

The previously written specification also specified several possible strings to use as shortcuts (or what symbols would look like). This PR expands on that by properly allowing unions and intersections in types, and implementing the necessary parsing for symbols, signatures, and types using winnow. While in theory writing a handwritten parser might result in something faster, `hcl-edit` and `toml-edit` both successfully use winnow, so it should be good enough for us.

This PR:

The changes in this PR:

The changes in this PR:

## 📹 Demo

look at the tests


vilkinsons:
Happy to give this the license greenlight once code review is in. Please prod me on Slack whenever!

libs/@local/hql/README.md
```diff
@@ -0,0 +1,42 @@
+# HQL - HASH Query Language
+
+This is the compiler, planner and executor for the HASH Query Language (HQL). HQL is a functional programming language that is created using a subset of JSON.
+
+## Language
+
+The language design docs are found in Notion, but in general the CST of the language is very basic, only consisting of three different constructs:
```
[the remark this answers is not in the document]
indietyp: reworded in [33b9536](https://github.com/hashintel/hash/pull/4753/commits/33b9536a41676ab5c78a0b46b86a28ae9263d966)

libs/@local/hql/cst/src/expr.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the input is a malformed expression.
+    pub fn from_value(arena: &'a Arena, value: &serde_json::Value) -> serde_json::Result<Self> {
+        DeserializeSeed::deserialize(ExprVisitor { arena }, value)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
TimDiekmann, line 50: The function names suggest that you're checking if the expression is of a specific type and to check this we can look at the snapshots, however, it would be helpful if we'd have something like `assert!(matches!(result, Ok(Expr::Call(_))));` in addition

libs/@local/hql/cst/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#![feature(allocator_api, box_into_boxed_slice)]
+
+extern crate alloc;
```
TimDiekmann, line 3: Do I see correctly, that we only need `alloc` for the types in `arena::` and we don't use an allocator from here? Great!

libs/@local/hql/cst/src/parse.rs
```diff
@@ -0,0 +1,119 @@
...
+    Parser,
+};
+
+use crate::arena::{self, Arena};
+
+enum VecOrOneValue<'a, T> {
+    Vec(arena::Vec<'a, T>),
+    One(T),
+    None,
+}
```
TimDiekmann, line 17: I don't really mind as this is internal, but I'd prefer `Option<Vec | One>` instead of `Vec | One | None`

libs/@local/hql/cst/src/snapshots/hql_cst__expr__test__fn_empty.snap
TimDiekmann, line 1: Can we hide snapshots in git as we do in `error-stack`?

libs/@local/hql/cst/src/snapshots/hql_cst__expr__test__fn_empty.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: libs/@local/hql/cst/src/expr.rs
+expression: result
```
TimDiekmann, line 3: Any way to prettify the description of snapshots? If that's too much work we can leave it.

libs/@local/hql/cst/src/symbol.rs
```diff
@@ -0,0 +1,292 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        Display::fmt(&self.0, f)
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) enum ParseRestriction {
+    None,
+    RustOnly,
+}
```
TimDiekmann, line 27: Could you add a small doc-string to describe what this means?

libs/@local/hql/cst/src/symbol.rs
```diff
@@ -0,0 +1,292 @@
...
+) -> impl Parser<Input, Symbol, Error>
+where
+    Input: StreamIsPartial
+        + Stream<Token: AsChar + Clone, Slice: AsRef<str>>
+        + Compare<char>
+        + for<'a> Compare<&'a str>,
+    Error: ParserError<Input>,
+{
+    move |input: &mut Input| {
+        dispatch! {peek(any).map(|token: Input::Token| token.as_char());
```
TimDiekmann, line 69: I don't know the `dispatch!` macro but it looks like a match. The first two match arms are identical, is that correct?

libs/@local/hql/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+The `#` symbol is used to denote a special form instructions and types, and is of form: `#<name>(...)`. Currently only `#sig` is supported for signatures.
+
+In the future these might be expanded upon to allow for more convenient definition of constants, or inputs, e.g.:
+
+- `{"const": 1, "type": "Int"}` -> `#value(Int, 1)`
+- `{"const": 1}` -> `#value(1)`
+- `["input", "variable"]` -> `#input(variable)`
+
+(This is possible because `#` is not allowed as the start of a valid identifier.)
```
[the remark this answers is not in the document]
indietyp: Yes correct. I removed the `#<name>(...)` syntax for now anyway. Was putting it into more of: potential extension, should be find that these cases persist.

libs/@local/hql/cst/src/lib.rs
indietyp: That is correct! All allocations go through the arena.

libs/@local/hql/cst/src/symbol.rs
indietyp: They are not, the difference is

```
char if ... && is_xid_start(char)
```

vs

```
'_' if ...
```

`_` is not part of XID_Start (so like Rust) we need to special case it a bit

libs/@local/hql/cst/src/symbol.rs
indietyp: added in [4004392](https://github.com/hashintel/hash/pull/4753/commits/400439251fa93808d90006ef2673b6d0e9feed08)

libs/@local/hql/README.md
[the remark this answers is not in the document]
indietyp: reworded in [af16d26](https://github.com/hashintel/hash/pull/4753/commits/af16d2654287f5c988652299466659e53f662a9c)

libs/@local/hql/cst/src/snapshots/hql_cst__expr__test__fn_empty.snap
indietyp: done in [e1af082](https://github.com/hashintel/hash/pull/4753/commits/e1af08221e3864ed361a1882c4e2ec72931a77ab)

libs/@local/hql/cst/src/snapshots/hql_cst__expr__test__fn_empty.snap
indietyp: changed in [95dea81](https://github.com/hashintel/hash/pull/4753/commits/95dea817262ed76ed132705fb0cffafe2ec456a7)

libs/@local/hql/cst/src/expr.rs
indietyp: done in [f06e40b](https://github.com/hashintel/hash/pull/4753/commits/f06e40bef971164d9e1287bab84a4539996eb6c5)!

libs/@local/hql/cst/src/parse.rs
indietyp: rewritten in [241f0ee](https://github.com/hashintel/hash/pull/4753/commits/241f0ee12b86d2a3331bcefac4883f3b662f6bcc)

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Language
+
+Language design documentation and a specification of the language will be provided in the future.
+
+The CST of the language is very basic, only consisting of three different constructs:
+
+- Function Invocation: A function invocation has multiple possible forms:
+  - `[name, ...args]`
+  - `{"fn": name, "args": args}`
```
TimDiekmann, line 13: Why do we have two forms for calling a function? Isn't one enough?

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
...
+- Constants: Constants are simple values that are not functions, they are defined using:
+  - `{"const": value}`, this will convert the JSON type to the corresponding `hql` type.
```
TimDiekmann, line 15: In which scenario `type` is not known?

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
...
+  - `{"const": value, "type": type}`, this will convert the JSON type to the specified `hql` type (if possible), if not possible this will result in an error.
+- Variables: Variables are references to values that are defined elsewhere, they are defined using:
+  - `"symbol"` (any string is a reference to a variable)
+  - `{"var": name}` (this is a more explicit way of defining a variable)
```
TimDiekmann, line 19: Why do we have two representation for a variable?

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
...
+- Signature: Signature is a special type, which is parsed independently of variables, they have a specific form (specified in the CST crate) and are defined using:
+  - `{"sig": ...}`
+  - `"sig"`
```
TimDiekmann, line 22: What does `"sig"` represent?

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
+# HQL - HASH Query Language
+
+This is the compiler, planner and executor for the HASH Query Language (HQL). HQL is a functional programming language that is created using a subset of JSON.
+
+## Language
+
+Language design documentation and a specification of the language will be provided in the future.
+
+The CST of the language is very basic, only consisting of three different constructs:
```
TimDiekmann, line 9:
```suggestion
The Concrete syntax tree (CST) of the language is very basic, only consisting of three different constructs:
```

TimDiekmann, line 9: I count four constructs + four additional representations 😉

libs/@local/hql/README.md
```diff
@@ -0,0 +1,51 @@
+# HQL - HASH Query Language
+
+This is the compiler, planner and executor for the HASH Query Language (HQL). HQL is a functional programming language that is created using a subset of JSON.
+
+## Language
```
TimDiekmann, line 5: It's super hard to follow what the CST is representing without a grammar being specified but I'm happy to add this in a follow up. This is also the point where we should mostly agree on a grammar.

On the grammar itself I have a few questions:
- Why we'll have special forms in the future at all?
- Why the CST as presented a few lines beblow does not fully align with the grammar as specified in Notion? I would expect the CST to fully align with the EBNF, e.g. a signature would be
	```ts
	{
		type: "signature",
		generics: {
			ident: Ident,
			where: Type[]
		}[],
		arguments: {
			ident: Ident,
			type: Type
		}[],
		return?: Type
	}
	```

TimDiekmann (changes requested):

libs/@local/hql/cst/src/parse.rs
TimDiekmann: Much better 🙂

libs/@local/hql/cst/src/snapshots/hql_cst__expr__test__fn_empty.snap
TimDiekmann: Much better, thanks!

libs/@local/hql/README.md
indietyp: in case of ambiguity.

I wanted to have an explicit version - the object representation and a more "comfortable" representation (the list representation.

Once parsed, they are the same, but this helps us in multiple scenarios:

* cases of ambiguity, this was especially prevalent when `do` was it's own construct, not just a special form function
* machine tools. We're trying to bridge the gap a bit between easy machine representation and a representation that can easily be authored by a human, which reduces nesting.

So to accomodate all these together, we have the S-expr like for `[name, ...args]`, and the object like form `{"fn": name, "args": args}`.

libs/@local/hql/README.md
indietyp: it's in the case of parsing, e.g. you have a constant of `1`, if you say:

* `{"const": 1}`, it will be converted to the equivalent JSON representation (so just `Number`) or it being inferred.
* `{"const": 1, "type": "Int"}`, will take the value in `const`, and try to assign it to `Int`, which would (statically) fail for e.g. `1.1`.

This is mostly an affordance to the user. I reckon that in most cases you don't need to specify the type.

It's very similar to rust, meaning:

```rs
let x: _ = 2;
// vs.
let x: u8 = 2;
```

Here, if we have `_`, if there's no other usage it would default to `i32`, but if we have a function of e.g. `something(a: u8)`, it would automatically constrain `x` to be `u8` instead.

libs/@local/hql/README.md
indietyp: I can also add the syntax to the README itself, right now it is specified above the parse function for the different constructs. I had to change it a little from the one in the Notion document, to resolve potential parse ambiguities.

libs/@local/hql/README.md
indietyp: A signature string that follows the signature EBNF.

libs/@local/hql/README.md
indietyp: same as in all other cases, mostly not needed, but I want to give the user the option to specifically specify what they want to have, in case there is an ambiguity of something e.g. being parsed as a variable, even tho it should be a signature and vice versa.

I do not forsee this happening right now, but I want to keep the option open.

In future this also will be expanded, so you're able to have:

```json
{"var": [path, to, value, to, access]}
```

(the idea is to allow the same in string representation, for easily human use, but for machines, using an array is easier)

so e.g.

```json
"foo.bar?.baz"
{"var": ["foo", "bar?", "baz"]}
```

libs/@local/hql/README.md
TimDiekmann: I cannot follow, why the signature would not be parsed into the CST?

indietyp: no, it will be, it's:

```
"() -> Unit"
```

that is then parsed into the CST as a signature.

indietyp: which is different to e.g. `"abc"`, which would instead refer to a path, which either points to a type or a variable.

libs/@local/hql/README.md
TimDiekmann: In that case let's encode it as `{type: "constant", value: Value, type?: Type }` instead. The CST for the above, btw, does not omit `type`, it's basically `type: "inferred"` or something like that.

libs/@local/hql/README.md
TimDiekmann: The CST represents the actual syntax, I really don't see why we would need two representation for one and the same syntax. Also, I don't like encoding function calls as arrays btw ;)

libs/@local/hql/README.md
indietyp: Yes that's correct, that is why it is represented by an `Option` in the CST itself, I didn't want to make any type assumptions in the CST, if type checking is part of the AST instead.

I also wanted to keep the syntax terse, which is why I have been using `{"const"}` (or `constant`) instead of `{"type": "constant", "value": ...}`, it also opens up the word `type` to be used for the type instead.

TimDiekmann: The `type: "constant"` is probably not needed, yes, but it helps deserializing, but it doens't really matter as we implement the deserializing anyway.

libs/@local/hql/README.md
TimDiekmann: I think this is highly confusing. Can we please only use a single representation for the exact same thing?

libs/@local/hql/README.md
indietyp: > Why we'll have special forms in the future at all?

Which special form? Special form values or special form functions?

Special form functions are needed to special case certain constructs, functional languages like e.g. closure have them as well.

So to say, we need to special case in the AST or during analysis certain constructs to make our lives easier, or to be able to delay execution.

For example:

```
["do", ...exprs]
```

is a special form function, that denotes a new block of expressions, that introduce a new scoped block (for variables and such) and is lazily computed.

Same thing e.g. `ifElse` and such, where an expression is only evaluated it the condition is either true of false.

Special form values on the other hand are just an idea to make authoring CST easier, the bridge the CST has to "wander" is a perilous one, because we need to consolidate two different use-cases, machine authored queries and human authored queries.
Because we foresee that in the short/mid-term this CST representation is going to be used as the interface, instead of something like GQL or SPARQL, we need to make it possible to easily author that content, while still having a way for us to later, easily implement a e.g. GQL transpiler.

This motivates many of the choices done here, where you have an implicit representation, that is easier to author, and an explicit one, that is easier for tools.

e.g. it's a lot more concise (also in indentation to write):

```json
["do",
	["let", "a", {"const": 2}],
	["let", "b", {"const" 3}],
	["let", "c", ["add", "a", "b"]]
	"c"
]
```

than the "full machine representation":

```json
{
  "fn": "do",
  "args": [
    {
      "fn": { "var": "let" },
      "args": [{ "var": "a" }, { "const": 2, "type": "Integer" }]
    },
    {
      "fn": { "var": "let" },
      "args": [{ "var": "b" }, { "const": 3, "type": "Integer" }]
    },
    {
      "fn": { "var": "let" },
      "args": [
        { "var": "c" },
        { "fn": "add", "args": [{ "var": "a" }, { "var": "b" }] }
      ]
    },
    { "var": "c" }
  ]
}
```

while the latter is a lot easier to generate for a computer program, it is pain to author as a human, something we expect to happen.

indietyp: special form values go into the same direction, by enabling us to be even more concise:

```json
["do",
	["let", "a", "#value(2)"],
	["let", "b", "#value(3)"],
	["let", "c", ["add", "a", "b"]]
	"c"
]
```

libs/@local/hql/README.md
TimDiekmann: > for easily human use

I don't think human users will ever write CSTs. Either they use the syntax or some tools to generate the CSTs for them (e.g. parsing UI elements or in scripts). Having multiple way of doing the same things (especially from the get-go) confuses more than it helps.

libs/@local/hql/cst/src/expr.rs
TimDiekmann: You're checking if the result is an expression, but we shoud also check, that the parsed string is the expected expression, so when parsing `foo(bar)` it's expected to be a function call with `bar` as argument.

libs/@local/hql/README.md
indietyp: This also motivates the difference between the object representation, and the more implicit terse representation:

```json
{"sig": "() -> Unit"} // or the full decoded version
```

is easier to author without the fear of ambiguities, than:

```json
"() -> Unit"
```

libs/@local/hql/README.md
indietyp: I agree to disagree here. I've had the pleasure to write quite a bit of queries using the current language AST, and I don't see that going away.

You will need _a way_ to get the data out programmatically as well, even if just exploratory, and authoring a huge tree of nodes hinders that goal.

libs/@local/hql/cst/src/expr.rs
indietyp: but that is what the snapshots are for, no? I used external snapshot testing here (instead of internal one) because the snapshots tend to get _huge_ (just because of the nesting).

TimDiekmann (approved):
@indietyp and I had an offline discussion about the remaining questions on this PR. Generally, we're going to move parsing into the CST into some frontend/markup crate that translates either the expected syntax (i.e. `let x = 123; let bar = foo(x);`) or a more machine-friendly syntax (i.e. the terse representation) into the CST. This will be done in a follow-up as the next PR is almost ready to be reviewed and the merge-conflicts would be nightmarish. The next PR will add spans to allow diagnostics. Most likely moving the parsing will be done as a follow-up. As the two different representation (where the terse-syntax in a JSON-like S-syntax form will be the interesting part for usage) requires different span informations, the CST will be made agnostic of the spans.

vilkinsons (approved):

PR #4754, opened 2024-07-23
H-2773: Don't start sync back watcher if Linear integration disabled

CiaranMn:

There's a loop in the Node API which checks for Redis messages related to changes to Linear-related entities, to write them back out to Linear.

If the Linear integration is not enabled this loop should not start.

## 🔍 What does this change?

- ...

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons (approved):
We'll want some alternative way of handling this when we add Google Sheets live sync, etc but if this does the job for now then great. We could also add a todo, though? @CiaranMn

CiaranMn:
> We'll want some alternative way of handling this when we add Google Sheets live sync, etc but if this does the job for now then great. We could also add a todo, though?

In the interest of just merging this I'll leave the `@todo` out – adding any other integration sync will be impossible without changing it so we won't overlook it

PR #4756, opened 2024-07-24
H-3159: Move non-external apps out of external servicee

TimDiekmann:

We currently have plenty of in-house developed service embedded into the `external-services` docker compose file. This has various flaws and the main motivation to change this now is that we need to depend on GraphQL in a non-system test but external services do not include the Node API.

## 🔍 What does this change?

- Move all non-external services out of the docker compose files
- Add commands to start services from `yarn start` (see `package.json` for more details)
- Change testing in CI to start depending services before integration tests are starting (avoids the need to build the docker containers - the building CI is persisted to make sure the docker containers are properly built)
- Add a `:healthcheck` command for each service which is blocking until the services becomes available
- Move the test-server to `@rust/graph-test-server` to allow running `turbo run start --filter @rust/graph-test-server` (generally, this cleans up a few things as well)
- Unify all migration steps as `start:migrate`/`start:test:migrate`
- Extend the healthcheck commands in the graph to allow passing `--wait --timeout XX`
- Adjust README
- Remove System Test CI (Playwright tests are now considered integration tests, with the above changes this will just work - but about 2.5x faster)

This PR:

The changes in this PR:

The changes in this PR:


package.json
```diff
@@ -19,8 +19,15 @@
     "dev:backend:realtime": "turbo dev --log-order stream --filter '@apps/hash-realtime' --",
     "dev:backend:search-loader": "turbo dev --log-order stream --filter '@apps/hash-search-loader' --",
     "dev:frontend": "turbo dev --log-order stream --filter '@apps/hash-frontend' --",
-    "dev:graph": "yarn workspace @apps/hash-graph dev",
-    "dev:graph:test-server": "yarn workspace @apps/hash-graph dev:test-server",
+    "start": "turbo run start start:healthcheck --filter @apps/hash-graph --filter @rust/type-fetcher --filter @apps/hash-api --filter @apps/hash-ai-worker-ts --filter @apps/hash-integration-worker --filter @apps/hash-realtime --filter @apps/hash-frontend --env-mode=loose --ui tui --continue",
```
CiaranMn, line 22: We're not using `realtime` so I suggest we take the opportunity to remove from here. Can restore it when we do again
```suggestion
    "start": "turbo run start start:healthcheck --filter @apps/hash-graph --filter @rust/type-fetcher --filter @apps/hash-api --filter @apps/hash-ai-worker-ts --filter @apps/hash-integration-worker --filter @apps/hash-frontend --env-mode=loose --ui tui --continue",
```

package.json
```diff
@@ -19,8 +19,15 @@
...
+    "start:graph": "turbo run start start:healthcheck --filter @apps/hash-graph --filter @rust/type-fetcher --env-mode=loose --ui tui --continue",
+    "start:graph:test-server": "turbo run start start:healthcheck --filter @rust/graph-test-server --ui tui --continue",
+    "start:backend": "turbo run start start:healthcheck --filter @apps/hash-api --filter @apps/hash-realtime --env-mode=loose --ui tui --continue",
```
CiaranMn, line 25:
```suggestion
    "start:backend": "turbo run start start:healthcheck --filter @apps/hash-api --env-mode=loose --ui tui --continue",
```

package.json
```diff
@@ -45,7 +52,6 @@
     "lint:prettier": "prettier --check --ignore-unknown .",
     "lint:tsc": "turbo --continue lint:tsc --",
     "lint:yarn-deduplicate": "yarn-deduplicate --fail --list --strategy=fewer",
-    "migrate:graph": "yarn workspace @apps/hash-graph migrate",
```
CiaranMn, line 48: Is this now handled some other way, or do we have to write a longer command to migrate the graph?

TimDiekmann: When starting the graph it will run the migration as a dependency automatically.

package.json
TimDiekmann: Good idea, I removed the search-loader as well so we don't need to build dependencies for an `echo` command.

CiaranMn (dismissed):

vilkinsons (approved):

vilkinsons (approved):

PR #4767, opened 2024-07-29
H-3165: Set up hooks for artillery

TimDiekmann:

While logging in purely through artillery is working, it's very repetitive. Setting up a hook to log in helps here. Having this in place earlier helps judge what's possible in hooks so new scenarios can be created with that in mind. For example we can write custom engines and function invocations, which may be helpful later.

## 🔍 What does this change?

- Write code to login using the kratos client
- Write a small module to provide typing in Artillery. This is not complete but can be gives enough safety for now (if features are missing we can always add them). The `@types/artillery` package is outdated.
- Bundle TS code using `rollup` to use it in artillery
- Configure the typical linting rules

This PR:

The changes in this PR:

The changes in this PR:


tests/hash-backend-performance/rollup.config.js
```diff
@@ -0,0 +1,40 @@
...
+      sourceMap: !production,
+      inlineSources: !production,
+      outputToFilesystem: false,
+    }),
+    nodeResolve(),
+  ],
+  external: ["@ory/client"],
+});
+
+export default [rolls("cjs", "fat")];
```
CiaranMn, line 40: Unless we need to ever change this config, I'd simplify it to get rid of the variables we're passing around that are constant values. It makes it easier to understand later, because we don't have to figure out what we're using the `env` and `fmt` variables to change and why (nothing)

```js
const outDir = "dist/cjs";

export default [{
  input: "src/main.ts",
  output: {
    dir: outDir,
    entryFileNames: "[name].cjs",
    format: "cjs",
    name: "hash-backend-performance",
    sourcemap: !production,
  },
  plugins: [
    commonjs(),
    typescript({
      declaration: true,
      declarationDir: outDir,
      outDir,
      rootDir: "src",
      sourceMap: !production,
      inlineSources: !production,
      outputToFilesystem: false,
    }),
    nodeResolve(),
  ],
  external: ["@ory/client"],
}];
```

tests/hash-backend-performance/scenarios/login.yml
```diff
@@ -1,39 +1,20 @@
...
-                      subgraph {
-                        roots,
-                        vertices
-                      }
-                    }
+            query: |
+              query {
+                me {
+                  subgraph {
+                    roots,
```
CiaranMn, line 13: nit: not needed
```suggestion
                    roots
```

tests/hash-backend-performance/src/session.ts
```diff
@@ -0,0 +1,69 @@
...
+  session?: {
+    token: string;
+    expiresAt: number;
+  };
+};
+
+const setAuthorizationHeader = (request: RequestParams, token: string) => {
+  if (!request.headers) {
+    request.headers = {};
+  }
```
CiaranMn, line 19:
```suggestion
  request.headers ??= {};
```

tests/hash-backend-performance/src/session.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    }),
+  );
+
+  if (!context.vars.flowId) {
+    const loginFLow = await oryKratosClient
+      .createNativeLoginFlow()
+      .then(({ data }) => data);
+
+    context.vars.flowId = loginFLow.id;
```
CiaranMn, line 46:
```suggestion
    const loginFlow = await oryKratosClient
      .createNativeLoginFlow()
      .then(({ data }) => data);

    context.vars.flowId = loginFlow.id;
```

tests/hash-backend-performance/src/session.ts
```diff
@@ -0,0 +1,69 @@
+import { Configuration, FrontendApi } from "@ory/client";
+
+import type { BeforeRequest, RequestParams } from "./types";
+
+const SESSION_ID_REFRESH_INTERVAL = 1000 * 60; // 60 seconds
```
CiaranMn, line 5: Why do we expire the session every 60 seconds? This isn't what happens in practice (sessions are very long-lived, you can get the actual expiry date from `session.expires_at`)

I don't think it would make a huge difference to the scenario but I wouldn't bother to refresh the token unless the session had actually expired

tests/hash-backend-performance/rollup.config.js
TimDiekmann: I simplified the rollup config and made it a TS file instead

tests/hash-backend-performance/src/session.ts
TimDiekmann: I took the `expires_at` from kratos into account. If it's not set I assume there is no expiration date. I think this will become an issue, though, if the expiration date is very close to the current date as further requests will probably not be valid anymore? Anyway, if this becomes an issue (e.g. if we shorten the lifetime) we can still adjust it.
Mainly, this was here to show a usage of the hooks :)

CiaranMn: Cool, thank you!

CiaranMn (approved):

PR #4769, opened 2024-07-30
H-3180: Read user entity after logging in

TimDiekmann:

Reading the owned-by-id from a subgraph is not trivial (and probably not possible using JsonPath).

## 🔍 What does this change?

- Allows hooks to connect to the graph directly
- Reads the user entity after login
- stores the user properties and the owned-by-id in the session
- asserts, that the `me` query returns the correct entity
- changes rollup to emit an ESM module instead of CJS

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/tests/hash-backend-performance/rollup.config.ts
+++ b/tests/hash-backend-performance/rollup.config.ts
@@ -19,6 +19,7 @@ const bundles: RollupOptions[] = [
     plugins: [
       commonjs(),
       typescript({
+        declaration: true,
         outDir: OUT_DIR,
         rootDir: "src",
         sourceMap: !PRODUCTION,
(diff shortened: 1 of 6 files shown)


tests/hash-backend-performance/rollup.config.ts
```diff
@@ -4,15 +4,15 @@ import typescript from "@rollup/plugin-typescript";
...
+const OUT_DIR = "dist/esm";
 
 const bundles: RollupOptions[] = [
   {
     input: "src/main.ts",
     output: {
       dir: OUT_DIR,
-      entryFileNames: "[name].cjs",
-      format: "cjs",
+      entryFileNames: "[name].mjs",
```
CiaranMn, line 14: Strictly speaking I don't think this is required, as `.js` files will automatically be interpreted as type `"module"` due to its presence in `package.json`, but not a problem to leave it as-is
```suggestion
      entryFileNames: "[name].js",
```

CiaranMn (approved):

tests/hash-backend-performance/rollup.config.ts
CiaranMn: Ahh, fair enough!

PR #4784, opened 2024-08-05
H-3186: Improve tracing capabilities when running load tests

TimDiekmann:

We cannot use Artillery's setup for OpenTelemetry because they don't support passing trace IDs. This sets up a workaround to track which request comes from what load test.

## 🚫 Blocked by

- #4773 

## 🔍 What does this change?

- Set up the OTel SDK in Artillery and store scenario and request relevant information on the context
- Add actions to start/end a new scenario/request, this creates spans
- Extend graph-API client creation to allow specifying an injector (used to read the current Span ID)
- Read the trace IDs before doing the requests and add them as `traceparent` to the header

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Artillery's design is not suited for Telemetry tracking at all so this needs to work around this limitation by storing the scenario and request OTel context. Due to this, sometimes the spans are mixed up. I was not able to figure out why and tried several approaches. The best approach probably would have been creating a custom context manager, but this is a complicated task.


CiaranMn (dismissed):

vilkinsons (approved):
Legal 👍

PR #4785, opened 2024-08-06
H-3204: Update `Core Concepts` page content

vilkinsons:

Adds the content that corresponds with this _internal PR_: https://github.com/hashintel/internal-gitstart/pull/111

**Merge this PR first.**

Also:

- expands and improves a few other guide pages
- adds default opt-outs of telemetry enabled by various third-party tools, so that HASH is run more privately by default (internal sister PR: https://github.com/hashintel/internal/pull/159)

### DIFF
--- a/.env
+++ b/.env
@@ -108,6 +108,9 @@ HASH_TELEMETRY_DESTINATION=REPLACE_ME.aws.com
 # Is used for differentiating different apps, can be any value
 HASH_TELEMETRY_APP_ID=hash-app
 
+# Disable external telemetry collected by Vercel
+NEXT_TELEMETRY_DISABLED=1
+
 ###########################################
 ## Aliases for dockerized external services
 ###########################################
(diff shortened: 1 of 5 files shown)


CiaranMn (approved):

.env
```diff
@@ -108,6 +108,13 @@ HASH_TELEMETRY_DESTINATION=REPLACE_ME.aws.com
 # Is used for differentiating different apps, can be any value
 HASH_TELEMETRY_APP_ID=hash-app
 
+# Disable telemetry from third-party dependencies who transmit IP addresses
+NEXT_TELEMETRY_DISABLED=1 # Vercel Next.js
+TURBO_TELEMETRY_DISABLED=1 # Vercel Turborepo
+YARN_ENABLE_TELEMETRY=0 # Yarn
+ARTILLERY_DISABLE_TELEMETRY=true # Artillery
+CHECKPOINT_DISABLE=1 # Terraform and others
+
```
TimDiekmann, line 117: In most cases setting these here has no effect as `.env` is not read as part of it. This may leads to confusing that we specify these variables, however, they appear to not work.

In our applications, we specifically invoke `dotenv-flow`, which loads this and other `.env` files, however, if we just run a command like `turbo` this won't be seen.

Also important to notice that this _will not_ have any impact on the production instance.

vilkinsons: Following up with you separately.

PR #4786, opened 2024-08-06
H-3206: Allow missing data type ID in validation if unambiguous

TimDiekmann:

If an entity can be inserted unambiguously, there isn't strictly the requirement to specify the data type id.

## 🔍 What does this change?

- Change the `oneOf` validation of property types to act like a real `oneOf`, this implies that if a value could be inserted by satisfying two schemas it will fail
- By tightening that constraint we can make sure, that the data is not ambiguous. Normally, we wouldn't care about that and this is also the reason why we're going to replace `oneOf` with `anyOf`, but currently it's the simplest way to make sure that a missing data type ID is still an unambiguous entity.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- In theory, we want to allow `oneOf: [meter, meter]`, but this PR prevents that. With JSON schema this is forbidden anyway so it's technically more correct, but as soon as we make `dataTypeId` actually required we can lift this constraint again and make the `oneOf` an `anyOf`.
- All elements in a `oneOf` needs to be checked, we cannot exit early,

## 🛡 What tests cover this?

- A new test was added

### DIFF
--- a/libs/@local/hash-validation/src/property_type.rs
+++ b/libs/@local/hash-validation/src/property_type.rs
@@ -14,7 +14,7 @@ use type_system::{
 
 use crate::{
     error::{Actual, Expected},
-    OntologyTypeProvider, Schema, Validate, ValidateEntityComponents,
+    OntologyTypeProvider, Schema, ValidateEntityComponents,
 };
 
 macro_rules! extend_report {
@@ -74,6 +74,8 @@ pub enum PropertyValidationError {
         actual: VersionedUrl,
         expected: VersionedUrl,
     },
+    #[error("The property provided is ambiguous")]
+    AmbiguousProperty { actual: PropertyWithMetadata },
 }
 
 impl<P> Schema<PropertyWithMetadata, P> for PropertyType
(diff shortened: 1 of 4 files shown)


libs/@local/hash-validation/src/property_type.rs
```diff
@@ -101,22 +103,6 @@ where
     }
 }
 
-impl<P> Validate<PropertyType, P> for PropertyWithMetadata
```
indietyp, line 104: why the removal of validate? I get it is just redirecting to validate value.

libs/@local/hash-validation/src/property_type.rs
```diff
@@ -211,32 +181,42 @@ where
     }
 }
 
-impl<V, P, S> Schema<V, P> for OneOfSchema<S>
+impl<P> Schema<PropertyWithMetadata, P> for OneOfSchema<PropertyValues>
```
indietyp, line 184: why is this now only working on `PropertyWithMetadata`? I am guessing it's for the fact that `oneOf` will be relaxed only for property types, but not others?

indietyp (changes requested):
minor questions

libs/@local/hash-validation/src/property_type.rs
TimDiekmann: Yes, it's just a redirection which bloats code. Not needed.

libs/@local/hash-validation/src/property_type.rs
TimDiekmann: We only have `oneOf` in property types, no need to support it for all types. This way, we know the concrete error type.

indietyp (approved):

PR #4787, opened 2024-08-06
H-3207: Clean up package.json and sync script

TimDiekmann:

The sync script generates wrong `private` flags for crates which are not set to `publish = false`. Also, it generates empty sections in `package.json`. In addition, a few package.jsons are missing.

### DIFF
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -239,7 +239,21 @@ impl<'a> WorkspaceMember<'a> {
     }
 
     fn is_ignored(&self) -> bool {
-        self.is_blockprotocol()
+        let Some(sync) = self.package.metadata.get("sync") else {
+            return false;
+        };
+
+        let Some(turborepo) = sync.get("turborepo") else {
+            return false;
+        };
+
+        let Some(ignored) = turborepo.get("ignore") else {
+            return false;
+        };
+
+        ignored
+            .as_bool()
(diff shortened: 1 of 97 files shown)


indietyp:
We might want to add `workspace.package.publish = false` into the root workspace as well.

one of the supported keys: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

indietyp:
version is also supported, so we could inherit `version = "0.0.0"`, tho idk if that is a good idea.

TimDiekmann:
I would add neither. Versions should be explicit and publish is just a boolean flag, I don't see a point of inheriting it.

CiaranMn (dismissed):

indietyp:
primarily for `cargo new` (that's how I generate packages), adding `publish` and `version` is the thing I usually forget.

also sorry for the recent publish mistakes D: it's so easy to forget adding the two keys to the manifest.

TimDiekmann:
Oh, the `cargo new` fact is interesting. I think I _never_ used that command 😄

vilkinsons:
We could consider doing this for `cargo new` ^ Happy to provide re-approval when we reach a decision on this.

TimDiekmann:
Already did: [`1667f4c` (#4787)](https://github.com/hashintel/hash/pull/4787/commits/1667f4c1af547d1dc22e19ca13d5967ba5c5121f)

vilkinsons (dismissed):

vilkinsons (approved):

PR #4802, opened 2024-08-07
 H-3208: Enable workspace lints for `harpc-net` and `harpc-tower`

indietyp:

Enables and fixes workspace lints for `harpc-net` and `harpc-tower`.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/harpc/net/src/session/server/test.rs
+++ b/libs/@local/harpc/net/src/session/server/test.rs
@@ -225,8 +225,7 @@ impl EchoService {
 
     fn spawn(self, mut server: ListenStream) {
         tokio::spawn(async move {
-            // you may ask yourself: why not `while let`?
-            //  significant drop in scrutiny clippy warning.
+            // Using `loop` instead of `while let` to avoid significant drop in scrutiny
             loop {
                 let Some(transaction) = server.next().await else {
                     break;
--- a/libs/@local/harpc/net/src/session/server/test.rs
+++ b/libs/@local/harpc/net/src/session/server/test.rs
@@ -225,7 +225,13 @@ impl EchoService {
 
     fn spawn(self, mut server: ListenStream) {
         tokio::spawn(async move {
-            while let Some(transaction) = server.next().await {
+            // you may ask yourself: why not `while let`?
+            //  significant drop in scrutiny clippy warning.
+            loop {
+                let Some(transaction) = server.next().await else {
+                    break;
+                };
+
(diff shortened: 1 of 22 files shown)


libs/@local/harpc/net/src/session/server/test.rs
```diff
@@ -225,7 +225,13 @@ impl EchoService {
 
     fn spawn(self, mut server: ListenStream) {
         tokio::spawn(async move {
-            while let Some(transaction) = server.next().await {
+            // you may ask yourself: why not `while let`?
+            //  significant drop in scrutiny clippy warning.
+            loop {
```
TimDiekmann, line 230:
```suggestion
            // Using `loop` instead of `while let` to avoid significant drop in scrutiny
            loop {
```

libs/@local/harpc/net/src/session/test.rs
```diff
@@ -121,7 +121,12 @@ impl SimpleEchoService {
 
     fn spawn(mut server: ListenStream) {
         tokio::spawn(async move {
-            while let Some(transaction) = server.next().await {
+            // not a while let loop because of "significant scrutinee in drop" clippy lint
```
TimDiekmann, line 124:
```suggestion
            // Using `loop` instead of `while let` to avoid significant drop in scrutiny
```

TimDiekmann (changes requested):

libs/@local/harpc/net/src/session/server/test.rs
indietyp: got too sassy in my explanation comment D:

TimDiekmann: 🤏🏻 😄

libs/@local/harpc/net/src/session/server/test.rs
```diff
@@ -225,7 +225,12 @@ impl EchoService {
 
     fn spawn(self, mut server: ListenStream) {
         tokio::spawn(async move {
-            while let Some(transaction) = server.next().await {
+            // Using `loop` instead of `while let` to avoid significant drop in scrutiny
```
TimDiekmann, line 228:
```suggestion
            // Using `loop` instead of `while let` to avoid significant scrutiny in drop
```

TimDiekmann (approved):

TimDiekmann (approved):

PR #4826, opened 2024-08-09
H-3239: HQL: Make span storage insertion non mutable

indietyp:

Using `orx-concurrent-vec` `SpanStorage` is now able to have `insert(&self)`, this simplifies carrying around the storage to various parts.

To increase the benefits, the internal vector has been wrapped in `Arc`, allowing for the cloning of the storage. This is especially useful to reduce lifetimes around parsing code.

Additional as a drive-by diagnostics now only add color if requested.

The full history of the branch can be seen in `bm/hql/cst-preserve`, which will be deleted once this stack of PRs is resolved.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hql/span/src/storage.rs
+++ b/libs/@local/hql/span/src/storage.rs
@@ -11,7 +11,6 @@ use crate::{tree::SpanNode, Span, SpanId};
 /// which is returned on insertion.
 #[derive(Debug)]
 pub struct SpanStorage<S> {
-    // ConcurrentVec is also used by salsa / nikomatsakis
     inner: Arc<ConcurrentVec<S>>,
 }
 
--- a/libs/@local/hql/span/src/storage.rs
+++ b/libs/@local/hql/span/src/storage.rs
@@ -1,3 +1,7 @@
+use alloc::sync::Arc;
+
+use orx_concurrent_vec::ConcurrentVec;
+
 use crate::{tree::SpanNode, Span, SpanId};
 
 /// A collection of spans within a single source.
@@ -5,14 +9,18 @@ use crate::{tree::SpanNode, Span, SpanId};
 /// The storage is append-only and does not support the removal or modification of spans.
 /// Once inserted spans are to be considered immutable and can be referred to via their `SpanId`,
 /// which is returned on insertion.
+#[derive(Debug)]
 pub struct SpanStorage<S> {
-    inner: Vec<S>,
+    // ConcurrentVec is also used by salsa / nikomatsakis
+    inner: Arc<ConcurrentVec<S>>,
 }
 
 impl<S> SpanStorage<S> {
     #[must_use]
-    pub const fn new() -> Self {
-        Self { inner: Vec::new() }
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(ConcurrentVec::new()),
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


libs/@local/hql/span/src/storage.rs
```diff
@@ -1,18 +1,26 @@
...
 /// A collection of spans within a single source.
 ///
 /// The storage is append-only and does not support the removal or modification of spans.
 /// Once inserted spans are to be considered immutable and can be referred to via their `SpanId`,
 /// which is returned on insertion.
+#[derive(Debug)]
 pub struct SpanStorage<S> {
-    inner: Vec<S>,
+    // ConcurrentVec is also used by salsa / nikomatsakis
+    inner: Arc<ConcurrentVec<S>>,
```
TimDiekmann, line 15: I'm not a fan of making this an `Arc`. Instead, I think we should make this `!Clone` again and wrap it in an `Arc` where we need. `span_storage.clone()` is misleading as this converts the storage into a smart pointer.

TimDiekmann (changes requested):

libs/@local/hql/span/src/storage.rs
indietyp: I've done this primarily for the parsing (#4828), I had it `!Clone` before, but it made handling the storage quite a bit challenging. (just a lot of lifetimes), and once wrapped, you can't unwrap it.

I can `!Clone` it here, but then would need to make the upstream changes.

TimDiekmann: Why you cannot pass `Arc<SpanStorage<S>>` around when parsing?

indietyp: sure you can, but then you can't remove it out of the `Arc` again, so I thought, that it might just make things easier (although not as visible) to put the `Arc` just into the storage itself.

indietyp: because you're likely to use it in an `Arc` anyway

TimDiekmann: For what scenario you need to remove it from the `Arc` again? It has internal mutability and it's possible to get references to the inner type easily. If functions only require a reference to the storage you possibly even safe indirections and it's more visible what the storage actually is.

indietyp: > For what scenario you need to remove it from the Arc again?

exactly, that is why I put the `Arc` into it, I can - of course - also move it out of storage - tho I didn't want to have duplicate typing for it

TimDiekmann: The semantic meaning of `Clone` on a storage is very different. We have lints to _not_ call `smart_pointer.clone()`, but we don't have that for `SpanStorage`. Further, it is not possible to implement a deep clone on `SpanStorage` anymore. 
Can we please move out the `Arc`? It should be possible to simply replace all usages of `SpanStorage` with `Arc<SpanStorage>` (which leads to the question: Do it needs to be `Arc` or is `Rc` enough?).

indietyp: No, I get that, you're right, I just wanted to avoid writing `Arc<SpanStorage<Span>>` all over the place 😅

indietyp: fixed in [e7d3c7a](https://github.com/hashintel/hash/pull/4826/commits/e7d3c7aab1ee1d9f3de2643e466067de9637837b)

TimDiekmann: I absolutely hear you and see your motivation 🙂 
Thank you!

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #4827, opened 2024-08-09
H-3240: HQL: remove parsing from CST and make it location aware

indietyp:

This is in preparation to another PR, and removes any parsing from the CST, as well as makes it location aware.

Any node in the CST now has a `SpanId` attached to it, which is used to track the location of the node in the source.

Specifically this means that `Expr` -> `ExprKind` and so on. It also introduces the `Program`, which is simply a list of expressions to be executed, a `program` is basically a do expression on the top level.

During testing I found that in most cases, having a single expression at the top level was quite cumbersome / unexpected, so `Program` tries to solve this. The concept of `Program` might be removed in a future iteration, as I am still unsure about it.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hql/cst/src/expr/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+use self::{call::Call, constant::Constant, path::Path, signature::Signature};
+use crate::Spanned;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExprKind<'arena, 'source> {
+    Call(Call<'arena, 'source>),
+    Signature(Signature<'arena>),
+    Path(Path<'arena>),
+    Constant(Constant<'arena, 'source>),
+}
```
TimDiekmann, line 17: I find the name misleading, I'd expect `kind` be small (`Copy`), but it's just the inner expression without the span ID. I'm not able to come up with a better name, though, so I won't block the PR on this but maybe you have an idea.

libs/@local/hql/cst/src/value.rs
```diff
@@ -0,0 +1,46 @@
...
+    Array(arena::Vec<'arena, Value<'arena, 'source>>),
+
+    /// Represents a JSON object
+    Object(arena::HashMap<'arena, Cow<'source, str>, Entry<'arena, 'source>>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Value<'arena, 'source> {
+    pub kind: ValueKind<'arena, 'source>,
+    pub span: SpanId,
```
TimDiekmann, line 39: I thought we won't have nested spans in values? For arrays and objects that's a quite big nesting, no? Or is this expected to happen quite rarely?

TimDiekmann (approved):
Looks good, thank you!
Two comments but non-blocking 🎉 🚀

libs/@local/hql/cst/src/expr/mod.rs
indietyp: Yea me too, but I couldn't find a better name for it, so I stuck with that one for now 😅 .

libs/@local/hql/cst/src/value.rs
indietyp: I thought we did? Now that we have our own value, it's quite easy to add, and it's really useful in conversion code, and handling values.

You can see that with the duplicate key diagnostic, that wouldn't have been possible otherwise.

TimDiekmann: Yes, I see it's being useful, no worries. My only concern is that resolving takes longer but that is only in the error path so all good 🙂

PR #4985, opened 2024-09-03
H-3301: Remove unused Rust dependencies

TimDiekmann:

I have not found a way to properly check this automatically, but cargo-machete does a decent jobs after some manual tweaks (it does not support Rust 2024 and has a lot of false positives, however, it detected quite a few unused dependencies)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,7 +379,6 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
- "type-system",
  "utoipa",
  "uuid",
 ]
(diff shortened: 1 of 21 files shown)


TimDiekmann:
@indietyp do you have an idea how this can be checked automatically?
- `cargo udeps` did not detect any unused dependencies
- `cargo machete` does not support Edition 2024 (need to change to `edition = "2021"` and `resolver = "3"`) + a bunch of dependencies being wrong (mainly public dependencies, but also dependencies used in macros)

vilkinsons (dismissed):

indietyp (approved):
Thanks! Still sad that this isn't a built-in functionality in cargo

PR #5006, opened 2024-09-06
H-3291: Implement a pre-processor for entities and infer data type IDs

TimDiekmann:

The graph eventually needs to change entities before they are inserted into the database. Initially, this required for two cases:
- Store the data type ID if it is inferred (a temporary workaround until we always pass it)
- Calculate the canonical value for a value

## 🔍 What does this change?

- Introduce `EntityVisitor` - a mutating visitor which recursively visits all properties
	- The name may change to distinguish between a mutable and an immutable visitor, currently, only the mutable version is implemented
	- It defaults to the corresponding `walk_*` functions, which implement traversal logic and validate the entity shape
- Add an `EntityPreprocessor` which implements `EntityVisitor`
	- Does the same property validation logic as the previous entity validator
	- Fails, if a data type ID is not set and cannot be inferred. If it can be inferred, it's stored on the entity metadata. This implies that all entities will now have data type IDs stored alongside values in the database - a very important step towards data-type based queries
- Remove the old property validation logic (it's covered in `EntityPreprocessor`
- Use the `EntityPreprocessor in
	- entity creation
	- entity patching
	- snapshot restoration

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Error handling currently enforces a single error enum. This needs to be improved
- The `EntityPreprocessor` is defined in the `validation` crate but also does infer data type IDs, this might be confusing and should be split.

## 🛡 What tests cover this?

- The integration test suite checks for the data type ID being set after none were provided


libs/@local/hash-validation/src/test_data_type.rs
TimDiekmann, line 1: Moved from `data_type.rs` and removed non-test code

libs/@local/hash-validation/src/test_property_type.rs
TimDiekmann, line 1: Moved from `property_type.rs` and removed non-test code

vilkinsons (approved):

vilkinsons (approved):

PR #5007, opened 2024-09-06
H-3311: Use Rust edition 2021

TimDiekmann:

We currently use the unstable 2024 edition. We don't have issues with that directly, however, some tools like renovate have issues with unstable features. To prevent issues we can use the old edition.

## 🔍 What does this change?

- Replace `gen fn` with vector-based approach (to be reverted with edition 2024 but it should not have meaningful impact currently)
- Use faster approach to count types (2024 is smarter about lifetimes)
- Add missing imports


indietyp (approved):
sad to see edition 2024 be gone, hopefully we can re-enable it soon!

TimDiekmann:
I hope so, too!

PR #5009, opened 2024-09-07
GEN-51: Run Renovate through GitHub Actions

TimDiekmann:

With renovate being self-hosted it's possible to use a nightly version of Cargo which is required to properly generate artifacts. The hosted instance of renovate does not (and probably never will!) support nightly toolchains.

Further, we are in full control of the invocations and have access to self-hosted configurations.

## 🔍 What does this change?

- Add a workflow to run renovate every 30 min and on manual dispatch
	- The flow will authenticate with Vault and use the stored credentials for the [HASH worker bot](https://github.com/apps/hash-worker)
	- It installs Rust (the toolchain specified in `rust-toolchain.toml` and node)
	- Runs renovate
	- Stores the renovate cache (quite small) as GH artifact. In the future we might want to use S3 for that
- Add an auto-approval bot for dependencies. It will run if (and only if) the target branch is `main`, the actor (not the author of the PR) is the HASH worker, and the branch starts with `deps/`
- Allow renovate to create config-file migration PRs

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The Rust dependency bumps still results in empty commits which is the reason why a lot of PRs are not created properly (this is a dedicated issue), but we won't get artifact-update problems with this PR
- We might need to further restrict the auto-approval constraints, but for now this should be good enough

## 🐾 Next steps

- Use S3 as Cache
- Figure out why some Rust dependencies are not bumped properly

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -11,10 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-    # Further restrict this to
+    # TODO: Further restrict this to
     # - Check the PR author (not only the actor)
     # - Check the event (only commits should trigger auto-approvals)
     # - (optional) check for green CI
+    # see: https://linear.app/hash/issue/H-3313/further-restrict-auto-approval-workflow
     if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
     steps:
       - name: Renovate
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-    if: github.actor == 'Renovate[bot]' && startsWith(github.head_ref, 'deps/')
+    if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
     steps:
       - name: Renovate
         uses: hmarr/auto-approve-action@v4
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -11,6 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
+    # Further restrict this to
+    # - Check the PR author (not only the actor)
+    # - Check the event (only commits should trigger auto-approvals)
+    # - (optional) check for green CI
     if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
     steps:
       - name: Renovate
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -1,0 +1,19 @@
+name: Auto-approve
+
+on:
+  pull_request_target:
+    branches:
+      - main
+
+jobs:
+  auto-approve:
+    name: Dependency bump
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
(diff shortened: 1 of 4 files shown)


.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - main
+
+jobs:
+  auto-approve:
+    name: Dependency bump
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
```
indietyp, line 14: This is a potential vulnerability, any branch in `deps/` will be auto-approved, would guard against that by also checking that the PR author is `hash-worker[bot]`.

indietyp (changes requested):
minor concern, otherwise LGTM

.github/workflows/auto-approve.yml
TimDiekmann: I have noted that already in the known-issues:
> We might need to further restrict the auto-approval constraints, but for now this should be good enough

The worker currently is _only_ used for creating commits/open PRs, so this is not a concern, yet. We also need to restrict the event to precisely say when an approval should happen. Ideally, only if CI is green.
```suggestion
    # Further restrict this to
    # - Check the PR author (not only the actor)
    # - Check the event (only commits should trigger auto-approvals)
    # - (optional) check for green CI
    if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
```

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,23 @@
...
+  auto-approve:
+    name: Dependency bump
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    # Further restrict this to
+    # - Check the PR author (not only the actor)
+    # - Check the event (only commits should trigger auto-approvals)
+    # - (optional) check for green CI
+    if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
```
TimDiekmann, line 18:
```suggestion
    # TODO: Further restrict this to
    # - Check the PR author (not only the actor)
    # - Check the event (only commits should trigger auto-approvals)
    # - (optional) check for green CI
    # see: https://linear.app/hash/issue/H-3313/further-restrict-auto-approval-workflow
    if: github.actor == 'hash-worker[bot]' && startsWith(github.head_ref, 'deps/')
```

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: npm install -g renovate
+
+      - name: Run renovate
+        env:
+          LOG_LEVEL: ${{ inputs.logLevel || 'info' }}
+          RENOVATE_FORCE: ${{ github.event.inputs.overrideSchedule == 'true' && '{"schedule":null}' || '' }}
+          RENOVATE_DRY_RUN: ${{ env.dry_run == 'disabled' && 'null' || inputs.dryRun }}
+          RENOVATE_PLATFORM_COMMIT: enabled
+          RENOVATE_REPOSITORIES: ${{ github.repository }}
+          # TODO: Consider using S3 by setting `RENOVATE_REPOSITORY_CACHE_TYPE`
```
TimDiekmann, line 116:
```suggestion
          # TODO: Consider using S3 by setting `RENOVATE_REPOSITORY_CACHE_TYPE`
          # see: https://linear.app/hash/issue/H-3315/use-s3-to-store-renovate-cache
```

indietyp (approved):

PR #5019, opened 2024-09-08
Revert H-3311: Bump Rust edition back to 2024

TimDiekmann:

I experimented around with renovate, further, and it turned out that the ability to not update the artifacts is the reason why PRs are not created. Also, I was able to create a self-hosted renovate version that is able to use nightly features.

## 🔗 Related links

- #5007

### DIFF
--- a/.github/scripts/rust/sync-turborepo.rs
+++ b/.github/scripts/rust/sync-turborepo.rs
@@ -1,7 +1,9 @@
 #!/usr/bin/env -S cargo +nightly -Zscript
 ---
+cargo-features = ["edition2024"]
+
 [package]
-edition = "2021"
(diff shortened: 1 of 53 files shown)


indietyp (changes requested):
LGTM, quick question: what happened to the async gen fn that was rolled back in the last PR as well? Do we want to keep the non async gen fn version?

TimDiekmann:
@indietyp Good point, I removed the `gen fn` usage for now. If we really need it we can use it and I don't have a strong opinion against it, but if we can avoid it for now it's probably better as some tools/IDEs cannot properly parse them anyway.

indietyp (approved):

PR #5020, opened 2024-09-08
H-3312: Enable some more Rust lints

TimDiekmann:

Enable some code-style lints and enable cargo-lints (I fixed them before already)

## 🔍 What does this change?

- Enable cargo lints (there are currently three:
	- `implicit_features`: Enforced in 2024 (Require `feature = ["dep:dependency"]` in order to enable optional dependencies)
	- `unknown_lints`: warn by default
	- `unused_optional_dependency`: warn by default
- Enable the `let_underscore_untyped` lint:
	- as we don't have a stable API, it's more likely to us to happen
	- I added an assert if possible to get some more confidence
- Enable `min_ident_chars` lint (setting to 1 letter): Single-letter variable names are band. This mainly is used for:
    - `f` for `fmt::Formatter`: We already agreed on using `fmt` instead
    - `f` for functions: `func` is better
    - `k`/`v` for `key`/`value`
    - `v` for `value`
    - `a`/`b` for `lhs`/`rhs`
- Drive-by fix array length in a test case


indietyp (approved):
thank you so much! I wish that one could overwrite the renamed params for certain traits, like serde or debug, instead of straight up ignoring them.

indietyp (approved):

PR #5022, opened 2024-09-08
H-3284, H-3287: Allow sorting Entities by entity type's properties

TimDiekmann:

We want to allow sorting entities by e.g. it's type's title, so we can sort by

- label
- type title
- entity ID

## 🔍 What does this change?

- Add sorting paths for the above
- Add logic to dynamically use `label` property (returns `null` if not present)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/libs/graph/src/store/postgres/query/table.rs
+++ b/apps/hash-graph/libs/graph/src/store/postgres/query/table.rs
@@ -383,6 +383,7 @@ pub enum JsonField<'p> {
     JsonPath(&'p JsonPath<'p>),
     JsonPathParameter(usize),
     StaticText(&'static str),
+    Label { inheritance_depth: Option<u32> },
 }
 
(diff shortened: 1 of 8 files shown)


apps/hash-graph/libs/graph/src/store/postgres/query/table.rs
```diff
@@ -560,30 +562,6 @@ impl DatabaseColumn for OntologyTemporalMetadata {
     }
 }
 
-fn transpile_json_field(
-    path: JsonField<'static>,
-    name: &'static str,
-    table: &impl Transpile,
-    fmt: &mut fmt::Formatter,
-) -> fmt::Result {
```
TimDiekmann, line 568: This was not used

vilkinsons (approved):

vilkinsons (approved):

PR #5047, opened 2024-09-10
GEN-193: `error-stack`: clear API distinction for tree vs. stack contexts

indietyp:

This PR implements #2377, by making a clear API distinction between multiple related errors (`Report<[C]>`) and a single current context (`Report<C>`). 

Having worked with this API distinction while adjusting the tests I found this to be actually quite useful and informative!

Note: `Report::<[C]>::attach` keeps `Report<[C]>`, even tho we push a new frame, meaning current_frames will be 1 again. I initially thought this might turn out to be an issue, but that wasn't the case, as we can guarantee that every branch in the tree has at least a single current context. Therefore we simply descend the tree until we reach said context, branching of in the case where a frame has multiple sources.

This makes the following snippet possible.

```rs
#[test]
fn attach_group() {
    let mut root = create_report()
        .attach(PrintableA)
        .attach(PrintableB(0))
        .expand();
    let nested = create_report().attach(AttachmentA).attach(AttachmentB);

    root.push(nested);

    // bury the first error under a couple of attachments
    let mut root = root
        .attach_printable(PrintableB(0))
        .attach_printable(PrintableB(1))
        .attach_printable(PrintableB(2));

    let shallow = create_report().attach(AttachmentA).attach(AttachmentB);

    root.push(shallow);

    assert_eq!(root.current_contexts().count(), 3);
}
```

Note: this will create quite some churn in the codebase.

## 🔗 Related links

- discussion #2377

This PR:

The changes in this PR:

The changes in this PR:

## 📹 Demo

look at the demo example

### DIFF
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -242,7 +242,7 @@ use crate::{
 /// ```
 #[must_use]
 #[allow(clippy::field_scoped_visibility_modifiers)]
-pub struct Report<C> {
+pub struct Report<C: ?Sized> {
     // The vector is boxed as this implies a memory footprint equal to a single pointer size
     // instead of three pointer sizes. Even for small `Result::Ok` variants, the `Result` would
     // still have at least the size of `Report`, even at the happy path. It's unexpected, that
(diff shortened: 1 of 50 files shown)


libs/error-stack/src/report.rs
```diff
@@ -480,6 +483,87 @@ impl<C> Report<C> {
...
+    /// You should prefer using [`Report::current_frame`] or [`Report::current_frames`] instead of
+    /// this function, as those properly interact with the type state of the `Report`.
+    ///
+    /// # Use Cases
+    ///
+    /// This function is primarily used to implement traits that require access to the frames,
+    /// such as [`Debug`]. It allows for code reuse between `Report<C>` and `Report<[C]>`
+    /// implementations without duplicating logic.
+    #[must_use]
+    pub fn current_frames_unchecked(&self) -> &[Frame] {
```
TimDiekmann, line 563: I think `_unchecked` is misleading here. Do you have a better name? Do we want to make this private?
```suggestion
    fn current_frames_unchecked(&self) -> &[Frame] {
```
or
```suggestion
    pub(crate) fn current_frames_unchecked(&self) -> &[Frame] {
```

libs/error-stack/src/report.rs
```diff
@@ -632,44 +886,52 @@ impl<C> Report<C> {
...
+    /// a.push(b);
+    ///
+    /// let io_error = a.current_contexts();
+    /// assert_eq!(io_error.count(), 2);
     /// ```
-    #[must_use]
-    pub fn current_context(&self) -> &C
+    ///
+    /// [`current_context`]: Self::current_context
+    pub fn current_contexts(&self) -> impl Iterator<Item = &C>
```
TimDiekmann, line 899: Can't wait for `gen fn` 😃

TimDiekmann (commented):
Amazing! 🎉 

Do you want to add a follow-up PR to enable this in this repository by adding a `path` to the `error-stack` version in the top-level `Cargo.toml`?

libs/error-stack/src/report.rs
indietyp: I had the exact same thought as you did, but I wanted to give users the chance to implement traits regardless of if it is a single context, or multiple contexts, so I'd like to keep it public (e.g. see our implementation of `Debug`). I initially called it `leaves` or `entries` but both didn't sound quite right so any suggestion is very welcome!

TimDiekmann: Downstream users can still implement `impl<C> Trait for Report<C>` and `impl<C> Trait for Report<[C]>`, no?

indietyp: Yea, tho it results in some duplicate code, but that should be fine I think. Then I think I'd just do `pub(crate)`. I really don't want to `pub(crate)` the frames field to be able to access it (it would be fine if we had restrictions in place, but alas, we're not here just yet)

TimDiekmann: Yeah, that is fine. Or can you think of an (internal) facade trait which returns the same? (probably not worth it, though)

indietyp:
I'll do a follow up PR that'll enable it in this workspace

TimDiekmann (approved):

Issue #4777, opened 2024-08-03
OptionExt to chain errors


tisonkun:
### Proposed Solution

Things like snafu's OptionExt - https://docs.rs/snafu/latest/snafu/trait.OptionExt.html

```rust
let make_error = || Error::Internal(format!("failed to submit query: {query}"));
optional_result.change_context_lazy(make_error)
```

Currently, we can use `ok_or_else(make_error)?` and there is an auto `Into`. But an OptionExt should provide more ergonomic practice.

TimDiekmann:
Hi @tisonkun and thanks for asking!

Actually, we actively decided _against_ an `OptionExt`. We had this before, but removed it before we initially released `error-stack`. The reason is the [same as in `eyre`](https://docs.rs/eyre/0.6.8/eyre/index.html#context-and-option): An option does not have a context, so `change_context` or `attach` does not really make sense here. It would be possible to create an `OptionExt` which provides `optional.ok_or_report(error)` and `optional.ok_or_report_lazy(make_error)`, but this basically is the same as `optional.ok_or(Report::new(error))` and `optional.ok_or_else(make_error).map_err(Report::new)` and I'm not sure if the overhead of (a) importing `OptionExt` and (b) learning another API just to safe a few characters is really worth it, I tend to avoid this.

Another reason is the semantic meaning of an option. We only want to create reports for errors, but an optional value is not an error, it needs to be made into an error in order to pack it into a report.

indietyp:
I can only echo on @TimDiekmann, especially because of the - aforementioned - semantic meaning `change_context` indicates that we already have a report, we're simply changing the context of said report. That isn't the case with an `Option`, in that case we would would be **creating** a new `Report`, not changing it.

I don't really see the benefit of something like `OptionExt`, sure it would make writing `ok_or_else(|| Report::new(..))` a bit more terse, but the downside is twofold:
* you have yet another trait import
* you have additional cognitive load for people new to the codebase that are maybe unfamiliar with `error-stack`

Everyone known `ok_or_else`, that wouldn't be the case with `ok_or_report` (or something similar).

tisonkun:
Thanks for your answers. All is fair point :D

Closed as won't do.

PR #5095, opened 2024-09-11
Introduce `ThinContext` trait for static `Context`s and zero sized types

mkatychev:

A lot of of the value in reports is the separation of error type from metadata (contexts & attachments).
In my daily use of error-stack, I found the greatest value when the `Context` was a zero sized type, allowing things like seamless conversions of [`Option<T>` into `Report<NotFound>`](https://github.com/knox-networks/bigerror/blob/1e9c47758e4c311c4a79866b0f57f58991861857/bigerror/src/lib.rs#L465-L476) types if `NotFound` was a zero sized type.

This PR upstreams the `ThinContext` trait to allow behaviour that is only possible when a `Context` is a static value at compile time:
https://github.com/hashintel/hash/blob/5aeb3bfa66a56c9c3f7d5ee7468a4fcdd7275b7a/libs/error-stack/src/context.rs#L134-L145

This allows a report to get a value implicitly when calling `Report::change_context` greatly increasing the ergonomics of propagating errors as well as incrementing just the location of when an error could propagate multiple times:
`err.change_context(MyError)` becomes `err.into_ctx()`

## 🔗 Related links

Fork of error-stack with `ThinContext` implemented: https://github.com/knox-networks/bigerror

- ...

## 🚫 Blocked by

This adds a `#[derive(ThinContext)]` macro to the error-stack macro crate, it will need to be published for error-stack to feature this on crates.io

## 🔍 What does this change?

* Feature addtion to `error-stack` to scope behaviour for a subset of contexts

This PR:

yes 

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

The test coverage and documentation are lacking in this branch currently. This PR is intended to test the waters for this kind of feature.

## 🛡 What tests cover this?

* `error-stack/tests/thin_ctx.rs`

-

## ❓ How to test this?

1. run `cargo test` in error-stack

## 📹 Demo
[The test code](https://github.com/knox-networks/bigerror/blob/1e9c47758e4c311c4a79866b0f57f58991861857/bigerror/src/lib.rs#L675-L685):

```rust
#[test]
fn result_into_ctx() {
    fn output() -> Result<usize, Report<MyError>> {
        "NaN"
        .parse::<usize>()
        .map_err(ConversionError::from::<&str, usize>)
        .into_ctx()
        .attach_printable(
            "Report::into_ctx<MyError> will be called twice without creating an extra Frame!",
        )
    }

    assert_err!(output().into_ctx::<MyError>());
}
```

image

### DIFF
--- a/libs/error-stack/macros/Cargo.toml
+++ b/libs/error-stack/macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
-name = "error-stack-derive"
-version = "0.5.0"
+name = "error-stack-macros"
+version = "0.1.0"
 edition = "2021"
 rust-version = "1.63.0"
 license = "MIT OR Apache-2.0"
--- a/libs/error-stack/macros/Cargo.toml
+++ b/libs/error-stack/macros/Cargo.toml
@@ -1,25 +1,21 @@
 [package]
-name = "error-stack-macros"
-version = "0.0.0-reserved"
-authors = { workspace = true }
+name = "error-stack-derive"
+version = "0.5.0"
 edition = "2021"
 rust-version = "1.63.0"
 license = "MIT OR Apache-2.0"
-description = "Macros for the `error-stack` crate"
-documentation = "https://docs.rs/error-stack"
-readme = "../README.md"
-repository = "https://github.com/hashintel/hash/tree/main/libs/error-stack"
+description = "procedural macros for the error-stack library"
+repository = "https://github.com/hashintel/hash/tree/main/libs/error-stack/macros"
 keywords = ["errorstack", "error-handling", "macro", "derive", "no_std"]
 categories = ["rust-patterns", "no-std"]
-publish = false
 
 [lib]
 proc-macro = true
 
 [dependencies]
-
-[dev-dependencies]
-error-stack = { path = "..", default-features = false }
+proc-macro2 = { workspace = true }
+quote = { workspace = true }
+syn = { workspace = true }
 
 [lints]
 workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,7 +239,7 @@ checksum = "7378575ff571966e99a744addeff0bff98b8ada0dedf1956d59e634db95eaac1"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
  "synstructure",
 ]
 
@@ -251,7 +251,7 @@ checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
@@ -324,7 +324,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
@@ -335,7 +335,7 @@ checksum = "a27b8a3a6e1a44fa4c8baf1f653e4172e81486d4941f2237e20dc2d0cf4ddff1"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
@@ -1209,7 +1209,7 @@ dependencies = [
  "heck 0.5.0",
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
@@ -1472,7 +1472,7 @@ checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
@@ -1520,7 +1520,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim 0.11.1",
- "syn 2.0.71",
+ "syn 2.0.77",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


libs/error-stack/macros/Cargo.toml
```diff
@@ -1,25 +1,21 @@
 [package]
-name = "error-stack-macros"
-version = "0.0.0-reserved"
-authors = { workspace = true }
+name = "error-stack-derive"
```
TimDiekmann, line 2: Could you elaborate why you rename a crate owned by us? The crate is [published on `crates.io`](https://crates.io/crates/error-stack-macros) already. You did not even mention this in the PR description!

TimDiekmann (changes requested):
Hi!

I'm not entirely sure if I get the reason behind this implementation. Is the goal to simplify the generation of `Report`s? The example you gave is to create `Report<NotFound>` from an `Option<T>`. This would result in quite a few `OptionExt`-like traits (which we explicitly don't want to implement in `error-stack`: https://github.com/hashintel/hash/issues/4777#issuecomment-2275377679), which makes usage quite hard to follow. Also, I don't see why `into_ctx` needs to be called twice.

libs/error-stack/macros/Cargo.toml
mkatychev: That's an oversight, I'll revert

mkatychev:
> Hi!
> 
> I'm not entirely sure if I get the reason behind this implementation. Is the goal to simplify the generation of `Report`s? The example you gave is to create `Report<NotFound>` from an `Option<T>`. This would result in quite a few `OptionExt`-like traits (which we explicitly don't want to implement in `error-stack`: [#4777 (comment)](https://github.com/hashintel/hash/issues/4777#issuecomment-2275377679)), which makes usage quite hard to follow. Also, I don't see why `into_ctx` needs to be called twice.

Perhaps the `NotFound` was a poor example. I don't think introducing error types into error-stack is a good idea for the reasons you mentioned (the `OptionExt` trait may need some rethinking or not even used at all, they are valid points).

The intent of this PR was to allow ZSTs and const types to have a smaller overhead and to allow adding attachments rather than growing the context depth if you have a lot of functions that produce the same `Report<MyZstError>`s through nested function calls.

Errors whose value is known at compile time have no need to increment context depth so long as we are propagating the same type  so something like `Report<MyZstErr>` -> `Report<MyZstErr>` -> `Report<MyZstErr>` would give a clearer picture with this kind of output:

```
MyZstErr
- location.rs 10:10
- location_2.rs 100:10
- location.rs 20:10
```

Over the current approach:
```
MyZstErr
- location.rs 10:10
  MyZstErr
  - location_2.rs 100:10
  MyZstErr
  - location.rs 20:10
```

mkatychev:
I've updated the PR to reduce scope, address feedback, and elaborate the test to clarify the use case for a `ThinContext`.

indietyp:
Thank you for your PR but I share the sentiment of @TimDiekmann, I don't think it really belongs into error-stack (and looking at the implementation it actually doesn't even require access to error-stack internals). I would even go as far as saying that something like this goes against the philosophy of error-stack. Let me explain:

In error-stack we made sure that everything is explicit and deliberate - every single change that you do to a report - such as changing the context are deliberate and explicit - this removes that explicitness moving from a: this always changes the context to: "this sometimes changes the context under certain circumstances" it also goes against the philosophy in the sense that every context change is deliberate as it delineates a sort of "border" - the types from this part of the code-base return this error.

To quote the documentation:

> The current context should encapsulate how the current code interprets the error. As the error propagates, it will cross boundaries where new information is available, and the previous level of detail is no longer applicable. These boundaries will often occur when crossing between major modules, or when execution crosses between crates.

Your implementation fundamentally changes this allowing for a `change_context(A).change_context(A)`, this is fine if we cross a boundary or it is intentional, but your code makes this boundary conditional (in my personal opinion it's also a code-smell, if you have an error of `Report<C>` and you want to propagate it why call `change_context` at all if the "border" you use doesn't change).

This is it from a philosophical standpoint. From a code standpoint there are two major questions that arise for me:

A) why is a derive macro needed? Don't we already have `thiserror` or `derive-more` (I would kinda understand this if the point was to e.g. take advantage of things that aren't implemented, such as proper support of the `provide` API, but that's another discussion)
B) your code doesn't do what it says.
C) why `ctx` if everywhere else we deliberately use `context`

You intend that `ThinContext` is used only for ZSTs, but that isn't the case, nothing in the trait prohibits (I also don't know _if_ there is even a way to enforce that a trait be only implemented on ZSTs) but instead _any_ type that is constructable in a const context, using that assumption also automatically invalidates the trait implementation. What happens if my current context is `JsonError::InvalidChar` and I want to change it to `JsonError::UnexpectedLBrace`, `JsonError` is constructable in a const context and implements (in this example) `ThinContext`, if I now use the conditional `change_context` my change to `JsonError::UnexpectedLBrace` would simply be ignored and that would be super frustrating to debug, is hidden unintended behaviour.

mkatychev:
> Your implementation fundamentally changes this allowing for a `change_context(A).change_context(A)`, this is fine if we cross a boundary or it is intentional, but your code makes this boundary conditional (in my personal opinion it's also a code-smell, if you have an error of `Report<C>` and you want to propagate it why call `change_context` at all if the "border" you use doesn't change).

While the context does not change the location that it is propagated through does, if a ZST context is propagated through may functions there is often only one `Location` recorded making the `?` operator be the non canonical way to use error-stack. This PR doesn't fix that problem (perhaps this will be easier with try_v2 🤞)

> B) your code doesn't do what it says.

I don't know how to reply to this statement, could you add some context?

> Instead `ThinContext` allows _any_ type that is constructable in a const context

that was the intent of the derive throwing an err on an enum, I don't have a solution to that other than not allowing than to not have enums imlement `ThinContext` in the same way a user would not implement `Default` on an enum.

I appreciate the warm discussion and it makes sense for this code to not live in error-stack, the gap of propagating a context whilst only attaching new location still exists, I realize that one does not _change_ the context but there doesn't seem (in stable) a way to only add a location to a report before propagation, perhaps a `Report::add_location` public function could be used?

Thanks for making a great error library!

mkatychev:
Thanks for the discussion @indietyp @TimDiekmann, appreciate the courteous feedback!

PR #5161, opened 2024-09-17
H-3330: Optimize allocation behavior during report construction

indietyp:

This optimizes the allocation behavior of a report by introducing a backing vector.

This is an experimental change and might not be merged into error-stack!

This PR:

The changes in this PR:

The changes in this PR:


hashdotai (commented):
This PR introduces significant optimizations to error-stack's allocation behavior by implementing a backing vector approach. The changes look well-thought-out and should lead to improved performance when creating large error reports.

The main improvements include:

1. **Reduced allocations**: By using a fragment-based approach with a backing vector, the implementation can avoid frequent reallocations when building up error reports.

2. **Smart memory management**: The use of `RawSlice` and the fragment capacity doubling strategy should lead to more efficient memory usage patterns.

3. **Unsized frames**: Making `Frame` unsized and only accessible through references provides stronger guarantees about how frames are allocated and managed.

While the implementation is generally sound, I have some concerns about the safety of mutable slices and how errors are handled in some parts of the code. Additionally, some of the comments could be clearer to help future maintainers understand the design decisions.

Overall, this is a valuable optimization that should improve performance, but a few details could be refined to make the code more robust and maintainable.

PR #5164, opened 2024-09-17
H-3332: Move out the `AccountStore` trait of the `graph` crate

TimDiekmann:

In preparation of HaRPC and because in many locations we need the Store backend available (e.g. when searching for a conversion) moving out the Store traits helps.

## 🔍 What does this change?

- Add a minimal `store` crate
- Move `AccountStore`
- Refine the returned errors slightly

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/libs/store/src/account/mod.rs
```diff
@@ -41,50 +51,53 @@ pub struct InsertWebIdParams {
...
 /// Describes the API of a store implementation for accounts.
-#[async_trait]
 pub trait AccountStore {
     /// Inserts the specified [`AccountId`] into the database.
     ///
     /// # Errors
     ///
     /// - if insertion failed, e.g. because the [`AccountId`] already exists.
-    async fn insert_account_id(
+    fn insert_account_id(
```
indietyp, line 65: why `Result<.., Report<..>>` not `error_stack::Result`?

indietyp (changes requested):
LGTM, single comment

apps/hash-graph/libs/store/src/account/mod.rs
TimDiekmann: I mentioned this before somewhere: I find `error_stack::Result` quite annoying as it's not visible from looking at the code that we deal with `Report` here.

vilkinsons (approved):

apps/hash-graph/libs/store/src/account/mod.rs
indietyp: I agree, maybe we can come up with a better name to signal that, because I always find `Result<T, Report<C>>` quite cumbersome to write^^

indietyp (approved):

PR #5165, opened 2024-09-17
H-3331: Validate all data type constraints

TimDiekmann:

Currently, we only validate the constraints on the same data type but not the parents.

## 🔍 What does this change?

- Add the validation for the `allOf` field in data types

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/tests/hash-graph-http/tests/ambiguous.http
+++ b/tests/hash-graph-http/tests/ambiguous.http
@@ -60,8 +60,7 @@ X-Authenticated-User-Actor-Id: {{account_id}}
         "$id": "http://localhost:3000/@alice/types/data-type/meter/v/1",
         "title": "Meter",
         "type": "number",
-        "description": "A unit of length",
-        "minimum": 0
+        "description": "A unit of length"
       }
   ],
   "conversions": {},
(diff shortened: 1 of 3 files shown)


tests/hash-graph-http/tests/ambiguous.http
```diff
@@ -60,8 +60,7 @@ X-Authenticated-User-Actor-Id: {{account_id}}
         "$id": "http://localhost:3000/@alice/types/data-type/meter/v/1",
         "title": "Meter",
         "type": "number",
-        "description": "A unit of length",
-        "minimum": 0
```
CiaranMn, line 64: Out of interest, why did these need to be removed?

CiaranMn (approved):
Looks good, one non-blocking question

tests/hash-graph-http/tests/ambiguous.http
TimDiekmann: They should be inherited from the parent anyway. I don't use the constraint in this file, however, I think it's better practice to not repeat every constraint on child data types if not needed.

PR #5166, opened 2024-09-17
H-754, H-3333: Move `Filter` and `Subgraph` to store crate

TimDiekmann:

Strictly, they don't belong to the store but they are very related. This simplifies a lot of things. It's required to move them in order to move the store to a more accessible location.

## 🔍 What does this change?

- Split `QueryRecord` and `SubgraphRecord` (long overdue)
- Move `Filter`, `Subgraph`, and accommodating API to `hash-graph-store`

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/libs/store/src/subgraph/record.rs
```diff
@@ -0,0 +1,85 @@
...
+use temporal_versioning::{ClosedTemporalBound, TemporalTagged, TimeAxis};
+
+use crate::subgraph::identifier::{
+    DataTypeVertexId, EntityTypeVertexId, EntityVertexId, PropertyTypeVertexId, VertexId,
+};
+
+/// A record persisted in the [`store`].
+///
+/// [`store`]: crate::store
+// TODO: Split trait into subgraph part and query part
```
indietyp, line 14: isn't this resolved in this very PR?

indietyp (approved):
minor non-blocking question, otherwise LGTM

apps/hash-graph/libs/store/src/subgraph/record.rs
TimDiekmann: Yes, removed that 😅

indietyp (approved):

TimDiekmann:
Sorry, I've overseen a private intra-doc link 🙃

PR #5172, opened 2024-09-18
H-1982: Allow visualizing types in a graph

CiaranMn:

Introduces a graph visualization for types – specifically, the relationships between entity types (which entity types can link to other entity types).

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- H-3328: Use [sigma.js](https://www.sigmajs.org/) (used in this PR) instead of the existing ECharts solution for visualizing _entities_ in a chart, which should be more performant and allow us to use the same interactivity features (e.g. highlight a subgraph on hover/click).

## 🛡 What tests cover this?

- None yet

## ❓ How to test this?

1. Visit `/types`

## 📹 Demo

https://github.com/user-attachments/assets/49561633-ee35-43f6-8192-ccd668fca2c3


vilkinsons (approved):
🚀

PR #5181, opened 2024-09-20
GEN-196: Move `error-stack-experimental` into `error-stack`

indietyp:

This moves `error-stack-experimental` into the `error-stack` crate, reason for this is that we didn't plan on publishing `error-stack-experimental` anyway, this allows us to use the experimental features, while also giving it to users under a feature flag and makes developing these unstable features easier.

The only problem is that this means that `error-stack` `unstable` features won't be semver compatible anymore, but we judged that that was a good tradeoff to be made.

This PR also replaces the `ResultMultiExt` with a `ReportSink`, sadly due to `must_use` limitations we cannot solely rely on `must_use` to ensure the value is consumed, instead `ReportSink` makes use of runtime validation (which is only active during `debug_assertions`). `ReportSink::new()` will create a sink that will stderr if the sink isn't finished, `Report::new_armed` goes a bit further and will _panic_. This ensures that the value really is used every single time, but is quite disruptive, so is not the default.

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons (approved):
Rationale makes sense. Legal approval-only here. Tim for UAT.

libs/error-stack/src/ext/iter.rs
```diff
@@ -81,10 +81,22 @@ where
...
+///
+/// # Performance Considerations
+///
+/// These methods may have performance implications as they potentially iterate
+/// through the entire collection, even after encountering errors.
+///
+/// # Unstable Feature
+///
+/// This trait is currently available only under the `unstable` feature flag and
+/// does not adhere to semver guarantees. Its API may change in future releases.
```
TimDiekmann, line 97: I don't think we need to mention this everywhere as the documentation is properly generating this:
image

libs/error-stack/src/macros.rs
```diff
@@ -190,13 +190,96 @@ macro_rules! report {
...
+/// }
+///
+/// impl Context for PermissionDenied {}
+///
+/// if !has_permission(&user, &resource) {
+///     bail!(PermissionDenied(user, resource));
+/// }
+/// # Ok(())
+/// ```
+#[cfg(feature = "unstable")]
```
TimDiekmann, line 260: Is there a way to not include this feature flag in the docs? Maybe somethig like `#[cfg(not(docs))]`? Because currently, we have this:
image

libs/error-stack/src/sink.rs
```diff
@@ -0,0 +1,645 @@
...
+/// The `Bomb` type is used to enforce proper usage of `ReportSink` at runtime.
+///
+/// It addresses a limitation of the `#[must_use]` attribute, which becomes ineffective
+/// when methods like `&mut self` are called, marking the value as used prematurely.
+///
+/// By moving this check to runtime, `Bomb` ensures that `ReportSink` is properly
+/// consumed. Depending on its configuration, it will either:
+/// - Panic if the `ReportSink` is dropped without being used (when set to `Panic` mode)
+/// - Emit a warning to stderr (when in `Warn` mode, which is the default)
+/// - Do nothing if properly defused (i.e., when `ReportSink` is correctly used)
```
TimDiekmann, line 17: Do we want to move the documentation to the variants instead?

libs/error-stack/src/sink.rs
```diff
@@ -0,0 +1,645 @@
...
+    /// )));
+    /// ```
+    pub fn add(&mut self, report: impl Into<Report<[C]>>) {
+        let report = report.into();
+
+        match self.report.as_mut() {
+            Some(existing) => existing.append(report),
+            None => self.report = Some(report),
+        }
+    }
```
TimDiekmann, line 164: `add` is a term which is rarely used in Rust. Maybe we want to use `append` or the `Extend` trait? I don't have a strong opinion on that, I just want to mention it 🙂

libs/error-stack/src/sink.rs
```diff
@@ -0,0 +1,645 @@
...
+    /// let mut sink = ReportSink::new();
+    /// # // needed for type inference
+    /// # sink.capture(io::Error::new(io::ErrorKind::Other, "I/O error"));
+    /// // ... add errors ...
+    /// let result = sink.finish_with_value(42);
+    /// # let _result = result;
+    /// ```
+    ///
+    /// [`finish`]: ReportSink::finish
+    pub fn finish_with_value<T>(mut self, ok: T) -> Result<T, Report<[C]>> {
```
TimDiekmann, line 317: Non-blocking: Similar to `bool.then_some` we could use `finish_ok()`.

TimDiekmann (commented):
Mainly a few comments/questions. Looks very good, excited to test this out!

libs/error-stack/src/sink.rs
indietyp: Thought about that as well but didn't rename it because of the follow up PR that applies the changes to the worktree. I was thinking of renaming it to either "push" or "append" (more likely append) to be in line with the existing methods in Report<[C]>.

What di you think about append?

libs/error-stack/src/ext/iter.rs
indietyp: I just wanted to make sure to mention that it is exempt from semver guarantees and why, not necessarily that it's unstable.

libs/error-stack/src/macros.rs
indietyp: I would've liked if there's a way to conditionally disable match arms that would circumvent the problem. Sadly I didn't find an obvious ergonomic way to do so. I'll just disable the unstable macro during generation.

TimDiekmann: I noticed, that the `bail` macro is probably not working as expected:
```rust
error[E0034]: multiple applicable items in scope
  --> libs/error-stack/tests/test_macros.rs:36:55
   |
36 |     let report = capture_error(|| bail!([RootError])).attach_printable(PrintableA(0));
   |                                                       ^^^^^^^^^^^^^^^^ multiple `attach_printable` found
   |
   = note: candidate #1 is defined in an impl for the type `error_stack::Report<C>`
   = note: candidate #2 is defined in an impl for the type `error_stack::Report<[C]>`
```

Probably an issue for a follow-up (-> non-blocking)

libs/error-stack/src/sink.rs
TimDiekmann: As I proposed `append` myself I'm fine with that 🚀 

We can always change it later, no rush 👍🏻

libs/error-stack/src/macros.rs
TimDiekmann: I tried around with it a bit and failed to properly generate a doc here. We might want to remove the current `bail` from the docs and add a `bail` macro with both variants and a notice that the second variant is `unstable`.

TimDiekmann (dismissed):
A few things we like to do in follow-up but nothing blocking this PR from being merged 🚀

libs/error-stack/src/macros.rs
indietyp: I noticed this as well, I wanted to do this in a follow up - because the code for `Report<[C]>` and `Report<C>` is actually the same now (previously the return type was different)

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #5214, opened 2024-09-23
H-3348: Use `tsify-next` instead of `tsify`

TimDiekmann:

We encountered issues with `Tsify`. As that crate is unmaintained, we use a fork that aims to fix issues.


indietyp (approved):

vilkinsons (approved):

TimDiekmann:
The replacement crate results in more errors than before.

PR #5232, opened 2024-09-24
H-2958: Support `anyOf` in type system

TimDiekmann:

To allow a `Value` type we need the type system to support `anyOf` instead of specifying a `type` for data types.

## 🔍 What does this change?

- Support `anyOf` in the type system
- Add a test for `Value`
- Add proper validation logic for `anyOf` (future-proof so in theory any type could use `anyOf`)
- Add missing documentation for validations

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/mod.rs
@@ -225,7 +225,6 @@ mod raw {
             common: ValueSchemaMetadata,
             r#const: [JsonValue; 0],
         },
-        #[serde(skip)]
         AnyOf {
             #[serde(flatten)]
             common: ValueSchemaMetadata,
@@ -695,7 +694,6 @@ mod tests {
     };
 
     #[tokio::test]
-    #[ignore = "AnyOf constraint is not exposed"]
     async fn value() {
         ensure_validation_from_str::<DataType, _>(
             graph_test_data::data_type::VALUE_V1,
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/mod.rs
@@ -4,11 +4,11 @@ mod conversion;
 pub use self::{
     closed::{ClosedDataType, ClosedDataTypeMetadata},
     constraint::{
-        AnyOfConstraints, AnyOfSchema, ArrayConstraints, ArraySchema, ArrayTypeTag,
-        ArrayValidationError, BooleanTypeTag, ConstraintError, NullTypeTag, NumberConstraints,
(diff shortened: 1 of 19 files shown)


libs/@hashintel/type-editor/src/shared/data-types-options-context.tsx
```diff
@@ -199,28 +204,47 @@ const getArrayDataTypeDisplay = (
...
+    return expectedValuesDisplayMap.mixedArray;
+
     // TODO: Remove when `const` is not allowed on arrays
     //   see https://linear.app/hash/issue/H-3368/remove-const-from-array-constraints
   } else if (!isArrayConstraints(dataType)) {
     return expectedValuesDisplayMap.emptyList;
   } else if (!dataType.items) {
     return expectedValuesDisplayMap.mixedArray;
+  } else if ("anyOf" in dataType.items) {
+    return expectedValuesDisplayMap.mixedArray;
```
TimDiekmann, line 244: @CiaranMn Do we want to use the same approach for `anyOf` as for `prefixItems`? because for example this is still a number array:
```json5
{
	anyOf: [
		{
			type: "number",
			multipleOf: 2
		},
		{
			type: "number",
			multipleOf: 3
		},
	]
}
```

CiaranMn: Yes, we can change this, it will help as a marker for future work

TimDiekmann: Alright, I'll apply the changes

CiaranMn: Thank you!

TimDiekmann: Applied the changes in [`eda3a5a` (#5232)](https://github.com/hashintel/hash/pull/5232/commits/eda3a5ae8069c97b55ecb47395921c53718da46f)

TimDiekmann: Well, `anyOf` is not allowed in `items` anymore anyway 😅

libs/@blockprotocol/type-system/rust/src/schema/data_type/mod.rs
```diff
@@ -765,6 +784,48 @@ mod tests {
...
+              ]
+            }),
+            DataTypeValidator,
+            JsonEqualityCheck::Yes,
+        )
+        .await;
+    }
+
+    #[tokio::test]
+    async fn array2() {
```
TimDiekmann, line 811:
```suggestion
    async fn array() {
```

CiaranMn (approved):

PR #5239, opened 2024-09-26
H-3337: Align type-system data-type constraints with Node API (Part II)

TimDiekmann:

The Node API exposes more constraints than the Graph. We precisely want to match the type system and the Node API. This extends the previous PR by only exposing used constraints

## 🔗 Related links

- https://github.com/hashintel/hash/pull/5180

## 🔍 What does this change?

- Only allow constraints used in the Node API
- Add tests to make sure that additional properties are not allowed

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- `compileJsonSchema(type as unknown as JsonSchema, type.title, {` to convert `DataType` to `JsonSchema` (old PR, but worth mentioning)
- Omit `allOf` from `MinimalDataType` (it would not work with OpenAI, I guess for that we want the closed schema?)
- `const` for `type: "array"` can only be `[]` now and will be removed when we do H-3368

## 🐾 Next steps

- https://github.com/hashintel/hash/pull/5232

## 🛡 What tests cover this?

Various tests were added to check constraints

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
@@ -7,7 +7,7 @@ mod number;
 mod object;
 mod string;
 
-use error_stack::{bail, Report};
+use error_stack::{Report, bail};
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
 
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
@@ -24,12 +24,8 @@ use crate::schema::{DataTypeLabel, JsonSchemaValueType};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(target_arch = "wasm32", derive(tsify::Tsify))]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
-pub struct AnyOfSchema {
-    #[cfg_attr(
-        target_arch = "wasm32",
-        tsify(type = "[ValueConstraints, ...ValueConstraints[]]")
-    )]
-    pub any_of: Vec<ValueSchema>,
+pub struct AnyOfConstraint {
+    pub any_of: Vec<ValueConstraints>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
     #[serde(default, skip_serializing_if = "DataTypeLabel::is_empty")]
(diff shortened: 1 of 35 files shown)


libs/@local/hash-isomorphic-utils/src/data-types.ts
```diff
@@ -38,7 +38,7 @@ const createFormattedParts = ({
 
 export const formatDataValue = (
   value: JsonValue,
-  schema: ValueConstraint | null,
+  schema?: DataType | SimpleValueSchema,
```
CiaranMn, line 41: What's the motivation for changing this?

The intention of an explicit `null` is to avoid people just calling `formatDataValue(value)` and not passing the schema by mistake (which is easy to do).

tests/hash-graph-test-data/rust/src/data_type/value.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
+  "kind": "dataType",
+  "$id": "https://blockprotocol.org/@blockprotocol/types/data-type/value/v/1",
+  "title": "Value",
+  "description": "A value that can be stored in a graph",
+  "anyOf": [{ "type": "null" }]
```
CiaranMn, line 7: Why not the full list here? I'd call this 'Null' to avoid confusion if it's just for testing

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
```diff
@@ -6,32 +7,216 @@ mod number;
...
+        StringConstraints, StringFormat, StringFormatError, StringSchema, StringTypeTag,
+        StringValidationError,
+    },
 };
-use crate::schema::JsonSchemaValueType;
+use crate::schema::{JsonSchemaValueType, ValueLabel};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SimpleTypedValueSchema {
```
CiaranMn, line 31: Can we discuss these names? Not sure about them:
- why `SimpleValueSchema` to cover both the `anyOf` and the other stuff? Can't this just be called `ValueSchema`?
- re. `SimpleTypedValueSchema`... isn't all this stuff `Typed`? Maybe this could be `SingleValueSchema` instead, to distinguish between `anyOf`, which is multiple schemas? And then `SingleValueConstraint`

CiaranMn (commented):
Looking good, just a few questions

TimDiekmann:
@CiaranMn and I talked offline and agreed that all comments will be addressed but in a follow-up.

CiaranMn (approved):

PR #5240, opened 2024-09-26
H-2959: Implement a `Value` data type and make all data types inherit from it

TimDiekmann:

Every data type should inherit (directly or indirectly) from `Value`.

## 🚫 Blocked by

- #5232 

## 🔍 What does this change?

- Add a `Value` data type definitions:
	```json
	{
		"$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
		"kind": "dataType",
		"$id": "https://blockprotocol.org/@blockprotocol/types/data-type/value/v/1",
		"title": "Value",
		"description": "A value that can be stored in a graph",
		"anyOf": [
			{ "type": "null" },
			{ "type": "boolean" },
			{ "type": "number" },
			{ "type": "string" },
			{ "type": "array" },
			{ "type": "object" }
		]
	}
	```
- Move cached external types from Node migration to the type fetcher. This is the natural flow for reading external types and the type fetcher can read the cached types to avoid going to well-known types. In comparison to the Node API migration this implies that the types may always be available if the type fetcher is running and there is no custom logic required to mimic fetching external types.
- Change existing primitive types to inherit from `Value`
- Change `empty-list` type to use `items: false` instead of `const: []`

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The production instance needs to be patched to read the new data

## 🐾 Next steps

Enforce a parent in data types (except for `Value`)


apps/hash-graph/libs/type-fetcher/predefined_types/data_types/value_v1.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
+  "kind": "dataType",
+  "$id": "https://blockprotocol.org/@blockprotocol/types/data-type/value/v/1",
+  "title": "Value",
+  "description": "A value that can be stored in a graph",
```
CiaranMn, line 6: The current description is circular. Not sure what a great alternative is, here's one suggestion:
```suggestion
  "description": "A piece of data that can be used to convey information about an attribute, quality or state of something.",
```

tests/hash-graph-http/tests/ambiguous.http
```diff
@@ -130,6 +133,9 @@ X-Authenticated-User-Actor-Id: {{account_id}}
     "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
     "kind": "dataType",
     "$id": "http://localhost:3000/@alice/types/data-type/miles/v/1",
+    "allOf": [{
+        "$ref": "https://blockprotocol.org/@blockprotocol/types/data-type/number/v/1"
```
CiaranMn, line 137: Conceptually this should have `length` as a parent like the others above... if that is going to mess up the tests then maybe we can pick some other example than `miles` so that some future person doesn't wonder why it has `number` and not `length` (e.g. `integer`)

tests/hash-graph-http/tests/friendship.http
```diff
@@ -124,6 +124,7 @@ X-Authenticated-User-Actor-Id: {{account_id}}
     "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/data-type",
     "kind": "dataType",
     "$id": "http://localhost:3000/@alice/types/data-type/meter/v/1",
+    "allOf": [{ "$ref": "https://blockprotocol.org/@blockprotocol/types/data-type/value/v/1" }],
```
CiaranMn, line 127: Similar comment to above (applies below also)... not a big deal as these are tests but it would be nicest if the inheritance structure was consistent across the example data types

apps/hash-graph/libs/type-fetcher/predefined_types/data_types/value_v1.json
TimDiekmann: The description is definitely  better than my (halfway generated) one 🙂 
[`2e7e6e2` (#5240)](https://github.com/hashintel/hash/pull/5240/commits/2e7e6e204ea746842c641048e51122c88496e419)

tests/hash-graph-http/tests/ambiguous.http
TimDiekmann: I agree. I mean we should move those tests to Rust anyway and then we'd use the test data as defined there, however, until we have it it's better to have it consistent. [`2e7e6e2` (#5240)](https://github.com/hashintel/hash/pull/5240/commits/2e7e6e204ea746842c641048e51122c88496e419)

tests/hash-graph-http/tests/friendship.http
TimDiekmann: `length` is not available in this scenario, I just added it 😄 [`2e7e6e2` (#5240)](https://github.com/hashintel/hash/pull/5240/commits/2e7e6e204ea746842c641048e51122c88496e419)

apps/hash-graph/libs/type-fetcher/predefined_types/data_types/value_v1.json
TimDiekmann: I wonder if we want to extend the description and mention that every data type (directly or indirectly) is inherited from this type or if we want to leave that out.

CiaranMn: I would personally leave that out, I think we could cover it in documentation about the type system.

We don't currently talk about the type system or inheritance in the type descriptions themselves, and if we're going to do so we should do so consistently across all the descriptions. At the moment the descriptions are standalone explanations of what the thing in the world is, ignoring their place in the type system.

cc @vilkinsons in case of any view on this

vilkinsons: I'd also leave it out, and place it in the surrounding docs. We can then choose if it's something we pass along to LLMs, rather than it be perpetually bundled as part of the basic description (if they're ever exposed to the 'Value' data type... possibly not even, if we can avoid it).

PR #5241, opened 2024-09-26
H-3372: Use logger to print error-stack warnings

TimDiekmann:

The warnings on the terminal are annoying and cannot easily be silenced. This adds a `tracing` feature to `error-stack` to use `tracing` instead of `eprintln!`. Also, the message alone was kind-of pointless without a location, so on a `Warn`-level it will capture the caller location as well.

Example output:
```
`ReportSink` was dropped without being consumed at /<...>/hash/libs/@local/hash-graph-types/rust/src/knowledge/property/visitor.rs:452:22
```

### DIFF
--- a/apps/hash-graph/libs/api/Cargo.toml
+++ b/apps/hash-graph/libs/api/Cargo.toml
@@ -29,7 +29,7 @@ tower-http = { workspace = true, public = true }
 tracing = { workspace = true, public = true }
 
 # Private workspace dependencies
-error-stack = { workspace = true }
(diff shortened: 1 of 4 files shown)


indietyp (approved):
Thank you so much, that improves the non-lethal bomb quite a bit!

indietyp (approved):

TimDiekmann:
> Thank you so much, that improves the non-lethal bomb quite a bit!

The entity validator has quite some locations where the bomb is not defused, with this we can at least silence the warnings.

PR #5243, opened 2024-09-26
H-3354: Expand data types, introduce parents

CiaranMn:

This PR takes advantage of the recent ability for data types to inherit from other data types by introducing parents to sit above our existing data types.

It also takes the advantage to add some new data types which were obvious gaps given the existing ones (e.g. GigaByte but no other unit of information).

**Note to reviewers**
Please particulary check
1. That you are happy with the inheritance structure proposed
2. That you are happy with the titles and descriptions of data types
3. That the conversions defined are correct 

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- As some of these changes amend existing data types in place, the production database will need to be migrated separately to rewrite them.


apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/016-add-initial-currency-data-types.dev.migration.ts
```diff
@@ -19,9 +43,28 @@ const migrate: MigrationFunction = async ({
...
+  });
+
+  await createSystemDataTypeIfNotExists(context, authentication, {
+    dataTypeDefinition: {
+      allOf: [{ $ref: currencyDataTypes.schema.$id }],
+      title: "EUR",
+      description: "An amount denominated in Euros.",
       type: "number",
+      label: {
+        left: "€",
```
TimDiekmann, line 66: Euros should have the label on the right side

CiaranMn: [Not in English](https://en.wikipedia.org/wiki/Euro_sign#Use) 😅 

I think this is an internationalization issue... 

cc @vilkinsons

vilkinsons: This is indeed a case of the above. Given we're developing the English (US) version of this at the moment and deferring the localizaion/internationalization, this structure is fine for now (same decision as made recently for multi-plurality).

vilkinsons (dismissed):

TimDiekmann (approved):

PR #5283, opened 2024-09-30
H-3383: Implement `harpc` service trait

indietyp:

This implements the `harpc` service trait. This is the machinery a user (or proc-macro) is required to implement. While this may not look like much, it actually took quite a bit longer to implement, due to considerations of how to do encoding of payload and having a working example.

The example `examples/account.rs` shows what an example would look like.

This service trait machinery can be split into separate primary parts:

* user defined trait
	* instead of following a function based design, like `axum`, we use a `tarpc` like approach, where the user defines a trait that he then implements. This trait is split into two distinct roles with the same signature. `Server` and `Client` the code for the client can be easily generated (if wanted) and only the server side needs to be implemented by the user.
* `ServiceDelegate`, the service delegate is there to provide a bridge between the typed trait interface and the tower service. it takes just a handful of arguments and then depending on the required argument decodes them. As shown in the example implementing this per hand is quite straight forward, but it would also be possible to do this via machine generated code quite easily, although it would most likely require some sort of procedural macro.
* `Service` and `Procedure` trait. Those are not _strictly_ needed, but allow us to, at a later stage, verify some additional information, such as: the current version vs the requested version, procedures that are part of it, and most importantly it allows us to have a bidirectional type level link between the payload (procedure) and every service, through the use of marker traits. The `Procedure` trait and implementation are not strictly needed, but included here as they allow us to provide better documentation for the user (e.g. when was this method added, etc.)

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/service/src/delegate.rs
```diff
@@ -0,0 +1,17 @@
+use harpc_tower::{body::Body, request::Request, response::Response};
+use harpc_wire_protocol::response::kind::ResponseKind;
+
+use crate::Service;
+
+pub trait ServiceDelegate<S, C> {
```
TimDiekmann, line 6: Can we add a documentation for this trait please?

libs/@local/harpc/service/examples/account.rs
```diff
@@ -0,0 +1,195 @@
...
+            Self::CreateAccount => ProcedureId::new(0),
+        }
+    }
+}
+
+struct Account;
+
+impl Service for Account {
+    type ProcedureId = AccountProcedureId;
+    type Procedures = HList![CreateAccount];
```
TimDiekmann, line 54: Why we cannot use tuples here?

libs/@local/harpc/service/src/metadata.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/// Represents the removal information for a procedure or service.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Removal {
+    /// The version at which the procedure/service will be removed.
+    pub version: Version,
+    /// The reason for removal.
+    pub reason: Option<&'static str>,
+}
```
TimDiekmann, line 21: I don't really understand this. How I can get information for a removed procedure?

libs/@local/harpc/service/examples/account.rs
```diff
@@ -0,0 +1,195 @@
...
+struct AccountServiceImpl;
+
+impl<S> AccountService<Server<S>> for AccountServiceImpl
+where
+    S: Send + Sync,
+{
+    async fn create_account(
+        &self,
+        _: &S,
+        _: CreateAccount,
```
TimDiekmann, line 118: Can we add names to variables here?

TimDiekmann (changes requested):
Looks good so far, very exciting that we see a service now!

A few comments mainly to clarify a few things. I'm not sure how useful the metadata will be initially because we won't have a versioned API, yet.

libs/@local/harpc/service/src/metadata.rs
indietyp: It's about _when_ something is scheduled for removal, like on v6.0 this will be removed.

libs/@local/harpc/service/examples/account.rs
indietyp: We could, but I found that a `HList` (or similar h-list implementation) is just a lot easier to work with. With tuples you have some unique problems/constraints, like:

* to implement them you need to use a macro, the code is very repetitive
* you have an upper limit (like 16). You can circumvent that limit by using more traits / a proxy trait, but that is just very messy in general and hard to decipher.

That's why I semi recently-ish pivoted to use `HList` from `frunk` for these kind of things, instead of using tuples. It's a lot easier to write and reason about and the `HList!` and `hlist!` macro make it easy to construct these types. (also the crate itself has some cool things like coproduct, but that's beside the point).

I found the code to just be a lot more readable on the implementation side, with a bit of a negative (because now you're compelled to use a type-level macro) but found that the tradeoffs were good enough, if you disagree please let me know and I'll try to use tuples here.

libs/@local/harpc/service/examples/account.rs
indietyp: see: [d3cdbb9](https://github.com/hashintel/hash/pull/5283/commits/d3cdbb9bc463ff11427b80ff6103fb04c00e65e3)

libs/@local/harpc/service/src/delegate.rs
indietyp: see: [350295c](https://github.com/hashintel/hash/pull/5283/commits/350295c35004f4e27abc8ba6114adc8c42659e64)

libs/@local/harpc/service/examples/account.rs
TimDiekmann: We can stick with HList for now so we can make progress here, but generally, I prefer language-constructs (such as tuples). I don't see why construction of `HList` is easier when in a constant scenario such as defining `Procedure`. In addition, the documentation is much easier to read when looking at tuples and if we use a macro to implement a trait on each tuple elements that's fine. Something along these lines:
```rust
// Potential user-defined implementation if required
impl Trait for CreateAccount {...}

impl<T: Trait> Trait for (T,) { ... }
impl<T: Trait, U: Trait> Trait for (T,U) { ... }
```
while the `Trait` bound might not be needed. 

Let's keep it or now, but let's also consider changing it later. Generally, we don't need to create procedures dynamically I guess? If we build it dynamically we could use `frunk` as a helper to construct a tuple.

libs/@local/harpc/service/src/metadata.rs
TimDiekmann: I don't think this is needed. A deprecation warning should be good enough for that. If a function is stable in `v3`, then it's deprecated in `v4`, then it's expected to be removed later.

TimDiekmann (approved):
Happy to keep `Removal` in, but also happy if it would go away 😉 
Let's defer `HList`->`tuple` (discussion and potential switch) to a later point when we have more services.

libs/@local/harpc/service/examples/account.rs
indietyp: Yes. The reason I like this approach more is that it allows for a clean separation of concerns.

It allows us to have two different traits, one for just the list, another one for the trait itself and it doesn't require a macro to implement everything.

I also found that during dynamically building up a list it's a bit nicer.

I also fully get your point! In a perfect would we would have variable sized tuples we could implement traits over.

TimDiekmann: > I also fully get your point! In a perfect would we would have variable sized tuples we could implement traits over.

Variadic generics or at least variadic tuples would be a game changer for many things 🚀

libs/@local/harpc/service/src/metadata.rs
indietyp: you are right, done in [6551827](https://github.com/hashintel/hash/pull/5283/commits/6551827b6ccb2654673dfc6f1b8bf7ee29942222)

TimDiekmann (approved):

vilkinsons (approved):

PR #5287, opened 2024-10-01
H-3382: Ensure only primitive types can inherit from `Value`

TimDiekmann:

Every type should currently inherit from a primitive type, not from `Value` directly. This adds a validation step for this.

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/validation.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/validation.rs
@@ -1,11 +1,12 @@
-use std::collections::HashSet;
+use core::str::FromStr;
+use std::{collections::HashSet, sync::LazyLock};
 
 use serde_json::Value as JsonValue;
 use thiserror::Error;
 
 use crate::{
     Valid, Validator,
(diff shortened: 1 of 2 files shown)


libs/@blockprotocol/type-system/rust/src/schema/data_type/validation.rs
```diff
@@ -22,22 +23,57 @@ pub enum ValidateDataTypeError {
...
+        )
+        .expect("Invalid URL"),
+        VersionedUrl::from_str(
+            "https://blockprotocol.org/@blockprotocol/types/data-type/number/v/1",
+        )
+        .expect("Invalid URL"),
+        VersionedUrl::from_str("https://blockprotocol.org/@blockprotocol/types/data-type/text/v/1")
+            .expect("Invalid URL"),
+        VersionedUrl::from_str(
+            "https://blockprotocol.org/@blockprotocol/types/data-type/empty-list/v/1",
```
CiaranMn, line 54:
```suggestion
            "https://blockprotocol.org/@blockprotocol/types/data-type/list/v/1",
```

CiaranMn (dismissed):

CiaranMn (approved):

PR #5315, opened 2024-10-05
H-721: Remove `async-trait` dependency

TimDiekmann:

It's not needed anymore (except to implement third-party crates for `axum`)

## 🔍 What does this change?

- Remove `#[async_trait]` for all define traits

### DIFF
--- a/apps/hash-graph/libs/graph/src/store/crud.rs
+++ b/apps/hash-graph/libs/graph/src/store/crud.rs
@@ -6,9 +6,8 @@
 //!
 //! [`Store`]: crate::store::Store
 
-use async_trait::async_trait;
 use error_stack::Result;
-use futures::{Stream, TryStreamExt};
+use futures::{Stream, TryFutureExt, TryStreamExt};
 use hash_graph_store::{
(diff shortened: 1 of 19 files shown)


apps/hash-graph/libs/graph/src/store/crud.rs
```diff
@@ -202,36 +201,33 @@ pub trait ReadPaginated<R: QueryRecord, S: Sorting + Sync>: Read<R> {
...
         temporal_axes: Option<&QueryTemporalAxes>,
         include_drafts: bool,
-    ) -> Result<Vec<R>, QueryError> {
+    ) -> impl Future<Output = Result<Vec<R>, QueryError>> + Send {
         self.read(filter, temporal_axes, include_drafts)
-            .await?
-            .try_collect()
-            .await
+            .and_then(TryStreamExt::try_collect)
     }
```
TimDiekmann, line 223: `and_then` works quite nice here, no `await` is needed here 🚀

indietyp (approved):
big fan of the change!

PR #5320, opened 2024-10-06
H-3413: `harpc` implement `tower`-based server

indietyp:

This implement a `harpc` server that's based on `tower`, it's basically a culmination of the efforts over the last few weeks, integrating `harpc-tower` and `harpc-service` into one cohesive thing.

This is mostly all working, but there are currently some things that I need to come back to, specifically around: 

* `SessionStorageTask` - this should all work, but needs to be - ergonomically - hooked into `harpc-net` via the receiver. I have an idea how to do so, but will do that in a separate task instead. (just to keep this focused on the server component)
* `Codec` during this PR I found problems specifically with our current `Codec`, while it works, it's a bit hard to grapple, especially I found it hard to have a `Body<Error = E>` turn into a `Body<Error = !>` (this is possible, but the current design makes this harder than it should be. You'd need to get the first error, then encode that error (which is a future), you get back a bytes object, which you then `poll_frame` instead of the inner. That's quite a handful of things to get right, and I want to streamline the issue.[^1] [^2]

To showcase the changes, account.rs moved from `tower` to `server`.

[^1]: this also opens the question if the encode/decode function even need to be async and if we don't want to use something like serde. I have tried to stay away from using serde specifically here because of object safety and because it allows us down the line to implement codecs using `bincode`, etc. that have a nicer API outside of serde

[^2]: another problem - and why `Codec` was born - is that we're unable to encode/decode errors correctly, especially decoding reports, I tried to circumvent that a bit, but maybe this isn't the right solution. Another thing here is that we kinda want to guarantee that encoding an error doesn't lead to an error, because then we have the problem of: how do we report that encoding an error failed?!

> side note: I am kinda proud how we don't have a lot of boxing here^^

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

* ergonomic integration of `SessionStorageTask` and `harpc-net`
* client integration
* sort out `Codec` - either by revamping it / limiting it or by moving to serde.


Cargo.toml
```diff
@@ -30,7 +31,7 @@ members = [
     "libs/deer/macros",
     "libs/error-stack",
     "libs/error-stack/macros",
-    "libs/sarif",
+    "libs/sarif", "libs/@local/harpc/server",
```
TimDiekmann, line 34: Listed twice
```suggestion
    "libs/sarif",
```

libs/@local/harpc/server/src/delegate.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use harpc_service::delegate::ServiceDelegate;
+use harpc_tower::{body::Body, request::Request, response::Response};
+use tower::Service;
+
+use crate::session::{Session, SessionStorage};
+
+#[derive_where::derive_where(Clone; D: Clone, C: Clone)]
+#[derive_where(Debug; D: Debug, S: Debug + 'static, C: Debug)]
+pub struct ServiceDelegateHandler<D, S, C> {
```
TimDiekmann, line 15: Not a fan of `*Handler` structs, it's not clear what they are handling. Could you add documentation on what this does and why this is needed?

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct Handler<S> {
+    service: ServiceId,
+    version: Version,
+
+    inner: S,
+}
+
+pub trait Route<B, C> {
+    type ResponseBodyError;
+    type Future: Future<Output = Result<BoxedResponse<Self::ResponseBodyError>, !>> + Send;
```
TimDiekmann, line 31: Why do we require the `Future` to have a `!` in the `Err` variant? Can't we just use the `Ok` part instead?

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    fn call(&self, request: Request<B>, codec: C) -> Self::Future
+    where
+        B: Body<Control = !, Error: Send + Sync> + Send + Sync,
+        C: ErrorEncoder + Send + Sync;
+}
+
+// The clone requirement might seem odd here, but is the same as in axum's router implementation.
+// see: https://docs.rs/axum/latest/src/axum/routing/route.rs.html#45
```
TimDiekmann, line 40: It's not helpful to compare with other crates and what they are using. Why the `Clone` is needed?

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,244 @@
...
+        if self.head.service == request.service().id
+            && requirement.compatible(request.service().version)
+        {
+            let service = self.head.inner.clone();
+
+            Either::Left(service.oneshot(request))
+        } else {
+            Either::Right(self.tail.call(request, codec))
+        }
+    }
```
TimDiekmann, line 69: Just so I understand correctly: Is this a list iteration and if we found the service we `oneshot` the request, otherwise we go to the next one and error if we hit `HNil`? Could you add a short comment please?

libs/@local/harpc/server/src/serve.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        tokio::spawn(async move {
+            let Ok(response): Result<BoxedResponse<!>, !> = service.oneshot(request).await;
+            let response = response.into_body();
+
+            let pack = Pack::new(response).map(Ok);
+            if let Err(error) = pack.forward(sink).await {
+                tracing::error!(?error, "failed to send response");
+            }
+        });
```
TimDiekmann, line 44: What is happening to the handles? Shouldn't we at least return a drop-guard so the threads are properly joined?

libs/@local/harpc/server/src/session.rs
TimDiekmann, line 1: Could you add documentation for what a `Session` is used and what role the manager is playing?

TimDiekmann (changes requested):

libs/@local/harpc/server/src/delegate.rs
indietyp: not a fan of the name myself, but a `ServiceDelegateService` just sounded.... weird

libs/@local/harpc/server/src/session.rs
indietyp: see: [1e966ef](https://github.com/hashintel/hash/pull/5320/commits/1e966efcdacabb7210df7f00d2f6e94f25168973)

libs/@local/harpc/server/src/router.rs
indietyp: see: [1e966ef](https://github.com/hashintel/hash/pull/5320/commits/1e966efcdacabb7210df7f00d2f6e94f25168973)

libs/@local/harpc/server/src/router.rs
indietyp: see: [1e966ef](https://github.com/hashintel/hash/pull/5320/commits/1e966efcdacabb7210df7f00d2f6e94f25168973)

libs/@local/harpc/server/src/router.rs
indietyp: see: [1e966ef](https://github.com/hashintel/hash/pull/5320/commits/1e966efcdacabb7210df7f00d2f6e94f25168973)

Cargo.toml
indietyp: see: [1e966ef](https://github.com/hashintel/hash/pull/5320/commits/1e966efcdacabb7210df7f00d2f6e94f25168973)

libs/@local/harpc/server/src/serve.rs
indietyp: see: [e876238](https://github.com/hashintel/hash/pull/5320/commits/e876238c04a3905235eb8c066d4115ee31792859)

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/// Route requests to the appropriate handler based on the request's service and version.
+///
+/// This is essentially a type-level linked list of handlers, it's boxed equivalent is [`Steer`],
+/// but unlike [`Steer`] it doesn't require the same type for all handlers and a separation of both
+/// the meta information (service id and version) and the actual handler. Unlike [`Steer`], it also
+/// doesn't require `&mut self` access, which allows for more granular cloning.
+///
+/// The reason why we essentially require `Clone`/`Copy` for the handlers is that once the route is
```
TimDiekmann, line 39:
```suggestion
/// # Design motivations
///
/// The reason why we essentially require `Clone`/`Copy` for the handlers is that once the route is
```

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,280 @@
...
+/// such as axum. The inner routes are stored in an `Arc<T>`, which is cheap to clone, but means we
+/// need to require `&self` during routing. Once a route was chosen, we simply clone the service
+/// (and oneshot) the service. This keeps the cloned data minimal and allows for multi-threading.
+///
+/// The downside is that we're unable to keep any state in a service delegate that's persisted
+/// across invocations. To signal this, `ServiceDelegate` takes `self` and consumes it (even though
+/// it isn't strictly needed), as well as the use of sessions. To store any information across
+/// calls, one must make use of smart pointers, such as `Arc`.
+///
+/// [`Steer`]: https://docs.rs/tower/latest/tower/steer/struct.Steer.html
```
TimDiekmann, line 59: Can't we use this?
```suggestion
/// [`Steer`]: tower::steer::Steer
```

libs/@local/harpc/server/src/router.rs
TimDiekmann: Much better, thank you!

TimDiekmann (approved):

TimDiekmann (approved):

libs/@local/harpc/server/src/router.rs
```diff
@@ -0,0 +1,282 @@
...
+/// such as axum. The inner routes are stored in an `Arc<T>`, which is cheap to clone, but means we
+/// need to require `&self` during routing. Once a route was chosen, we simply clone the service
+/// (and oneshot) the service. This keeps the cloned data minimal and allows for multi-threading.
+///
+/// The downside is that we're unable to keep any state in a service delegate that's persisted
+/// across invocations. To signal this, `ServiceDelegate` takes `self` and consumes it (even though
+/// it isn't strictly needed), as well as the use of sessions. To store any information across
+/// calls, one must make use of smart pointers, such as `Arc`.
+///
+/// [`Steer`]: tower::steer::Steer
```
TimDiekmann, line 61: Well
image
```suggestion
/// [`Steer`]: https://docs.rs/tower/latest/tower/steer/struct.Steer.html
```

TimDiekmann (approved):

vilkinsons (approved):

PR #5361, opened 2024-10-09
H-3421: Resolve closed data types on creation/updating

TimDiekmann:

Data type can inherit from parent data types. This has some implications and to avoid resolving the full chain of all data types every time we need a way to resolve them when creating/updating data types.
- `description` and `label` can be inherited from the parents as they are optional. The `ClosedDataType` will always inherit both at the same time if none is set on the schema itself.
- Most importantly, the constraints needs to be iherited. Currently, this is "just" a flattened list of `allOf`. In a follow-up this list will be combined into a more concise representation.

## 🔍 What does this change?

- Add logic to resolve closed data types
- This does not yet expose the closed data types to the Node API nor does it cache the closed schema, it just calculates it.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- When calculating the closed schema it's required to sort the constraints. It's unlikely that this will even be a bottleneck, however, this could be optimized. However, we probably still want to do that to improve reading closed data types from the database.

## 🐾 Next steps

- Expose `get_closed_data_types` to Node API
- Create a more concise view of the constraints (hard)
- cache the generated closed schema to the database 

## 🛡 What tests cover this?

A test with a lot of permutations was added to make sure the types are resolved correctly in any order


apps/hash-graph/libs/graph/src/store/postgres/ontology/data_type.rs
```diff
@@ -498,35 +523,26 @@ where
...
-            .map(|(data_type_id, _)| {
-                ontology_type_resolver
+            .map(|(data_type_id, data_type)| {
+                let closed_metadata = ontology_type_resolver
                     .resolve_data_type_metadata(*data_type_id)
-                    .change_context(InsertionError)
+                    .change_context(InsertionError)?;
+                let closed_schema =
+                    ClosedDataType::from_resolve_data((**data_type).clone(), &closed_metadata)
+                        .expect("Could not create closed schema");
```
CiaranMn, line 538: In case this bubbles up when creating a closed entity type schema (unless there's something else which will distinguish it, in which case feel free to ignore). Also applies further below
```suggestion
                        .expect("Could not create closed data type schema");
```

CiaranMn (dismissed):
Looks good, one non-blocking comment – can reapprove when conflicts fixed

apps/hash-graph/libs/graph/src/store/postgres/ontology/data_type.rs
TimDiekmann: That code should not be here at all, it should not panic (`expect` errors), but propagate the error up. The error, however, contains the information we want. We mainly use expect for tests and a few times if we are sure that this will not happen or would lead to unexpected behavior when preconditions are not met (bug at the implementation). Fixed in 6cf493f8cb9a0265c89d23c082d052543215a3f7

CiaranMn (dismissed):

CiaranMn (approved):

PR #5380, opened 2024-10-11
H-3426: Implement folding for data type constraints

TimDiekmann:

When data types are resolved the constraints should be folded and minimized. This is the groundwork and logic to be used when combining different constraints.

## 🔍 What does this change?

- It's not a good idea to implement `combine` on `Constraint` as `Constraint` is generic and `combine` would not use the generic parameter. Instead, I renamed `Constraint` to `ConstraintValidator` and added a new `Constraint` trait. This adds a `combine` method which takes a mutable reference to a constraint and consumes another constraint. If they can be fully combined, `None` is returned, otherwise `Some` with the remaining constraints. If the constraints cannot be combined at all because it would be unsatisfiable it will error.
- Add `fold_intersections` function to construct a vector of all intersections using the `combine` method. This Is used to shrink the `allOf` constraints in the closed data types.

This PR is still in draft as I have another PR in the queue to implement the number combination logic and need some more testing around the behavior and signatures. Probably I'm replacing the mutable reference with an owned value. Methods with mutable references returning a results need to be careful to not change the value if they return an error (side-effects), in most cases this is not desired.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- H-3427: Implement folding for number constraints
- H-3428: Implement folding for string constraints
- H-3429: Implement folding for array constraints
- H-3430: Implement folding for anyOf constraints

## 🛡 What tests cover this?

Tests will be added in the follow-ups

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/array.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/array.rs
@@ -47,13 +47,22 @@ pub enum ArrayItemConstraints {
 }
 
 impl Constraint for ArrayItemConstraints {
-    fn combine(&mut self, other: Self) -> Result<Option<Self>, Report<ResolveClosedDataTypeError>> {
-        Ok(match (self, other) {
-            (Self::Boolean(lhs), Self::Boolean(rhs)) => lhs.combine(rhs)?.map(Self::Boolean),
-            (Self::Number(lhs), Self::Number(rhs)) => lhs.combine(rhs)?.map(Self::Number),
-            (Self::String(lhs), Self::String(rhs)) => lhs.combine(rhs)?.map(Self::String),
+    fn combine(
+        self,
+        other: Self,
+    ) -> Result<(Self, Option<Self>), Report<ResolveClosedDataTypeError>> {
+        match (self, other) {
+            (Self::Boolean(lhs), Self::Boolean(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::Boolean(lhs), rhs.map(Self::Boolean))),
+            (Self::Number(lhs), Self::Number(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::Number(lhs), rhs.map(Self::Number))),
+            (Self::String(lhs), Self::String(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::String(lhs), rhs.map(Self::String))),
             _ => bail!(ResolveClosedDataTypeError::IntersectedDifferentTypes),
-        })
+        }
     }
 }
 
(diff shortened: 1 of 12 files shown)


libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/array.rs
```diff
@@ -43,7 +46,27 @@ pub enum ArrayItemConstraints {
...
+            (Self::Boolean(lhs), Self::Boolean(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::Boolean(lhs), rhs.map(Self::Boolean))),
+            (Self::Number(lhs), Self::Number(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::Number(lhs), rhs.map(Self::Number))),
+            (Self::String(lhs), Self::String(rhs)) => lhs
+                .combine(rhs)
+                .map(|(lhs, rhs)| (Self::String(lhs), rhs.map(Self::String))),
+            _ => bail!(ResolveClosedDataTypeError::IntersectedDifferentTypes),
```
indietyp, line 64: you are using language that's a bit confusing, there is `combine`, `intersect` and `fold`, which one is it?

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/boolean.rs
```diff
@@ -16,7 +19,16 @@ pub enum BooleanTypeTag {
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct BooleanSchema;
```
indietyp, line 20: is `BooleanSchema` supposed to have fields added at a later stage?

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
```diff
@@ -61,7 +76,79 @@ pub enum ValueConstraints {
...
 
-impl Constraint<JsonValue> for ValueConstraints {
+impl ValueConstraints {
+    /// Folds multiple constraints into fewer constraints.
+    ///
+    /// This function attempts to combine as many constraints as possible. If two constraints
+    /// cannot be fully merged, they are kept separate.
+    ///
+    /// The algorithm works as follows:
+    /// - It iterattes over all constraints
```
indietyp, line 86:
```suggestion
    /// - It iterates over all constraints
```

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
```diff
@@ -61,7 +76,79 @@ pub enum ValueConstraints {
...
+    ///
+    /// If two constraints exclude each other, an error is returned.
+    pub fn fold_intersections(
+        schemas: impl IntoIterator<Item = Self>,
+    ) -> Result<Vec<Self>, Report<ResolveClosedDataTypeError>> {
+        schemas
+            .into_iter()
+            .map(Some)
+            .try_fold(Vec::<Self>::new(), |mut folded, mut constraints| {
+                folded = folded
```
indietyp, line 102: non-blocking: I wonder if it would be more readable using a loop instead of iterator combinators.

indietyp (changes requested):
minor question / spelling mistake

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/boolean.rs
TimDiekmann: At least not in the immediate future, no. But it would be possible to add a `const` field to represent something like `{ myFlag: false }` (as a type).

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/array.rs
TimDiekmann: Renamed it to `intersection`

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
TimDiekmann: I'd keep it as it is for now, but I hear you and I may going to rewrite it as for-loop when I have to add more features in this logic anyway.

indietyp (approved):

PR #5381, opened 2024-10-11
H-3427: Implement folding for number constraints

TimDiekmann:

This adds the `combine` functionality for numbers. 

## 🔍 What does this change?

- Implement behavior to combine number constraints

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Due to the mutable nature of `combine` there is quite some duplication as the `lhs` and `rhs` typically cannot be swapped simply. This probably requires some changes to H-3426

## 🛡 What tests cover this?

Various test cases were added. Also, previous tests were adjusted to take the new changes into account.

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
@@ -271,10 +271,13 @@ mod tests {
     use core::fmt::Display;
     use std::collections::HashSet;
 
-    use error_stack::Frame;
-    use serde_json::Value as JsonValue;
+    use error_stack::{Frame, Report};
+    use serde_json::{Value as JsonValue, json};
 
-    use crate::schema::data_type::constraint::{ConstraintValidator, ValueConstraints};
+    use crate::schema::data_type::{
+        closed::ResolveClosedDataTypeError,
+        constraint::{ConstraintValidator, ValueConstraints},
+    };
 
     pub(crate) fn read_schema(schema: &JsonValue) -> ValueConstraints {
         let parsed = serde_json::from_value(schema.clone()).expect("Failed to parse schema");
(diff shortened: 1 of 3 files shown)


libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/number.rs
```diff
@@ -120,16 +120,110 @@ impl Constraint for NumberSchema {
...
+                let passed = r#enum
+                    .iter()
+                    .filter(|&value| constraints.is_valid(value))
+                    .copied()
+                    .collect::<Vec<_>>();
+
+                match passed[..] {
+                    [] => {
+                        // We now properly capture errors to return it to the caller.
+                        let _: Vec<()> = r#enum
```
indietyp, line 153: `()` implements `FromIterator<()>`, therefore we shouldn't even need the vec here.

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
```diff
@@ -313,4 +316,52 @@ mod tests {
 
         assert_eq!(errors, actual_errors);
     }
+
+    pub(crate) fn intersect_schemas(
+        schemas: impl IntoIterator<Item = JsonValue>,
+    ) -> Result<Vec<JsonValue>, Report<ResolveClosedDataTypeError>> {
+        let schemas = schemas
+            .into_iter()
+            .map(|schema| serde_json::from_value(schema).expect("Failed to parse schema"))
```
indietyp, line 325: I see copilot wrote these expect messages :P

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/number.rs
```diff
@@ -549,4 +779,651 @@ mod tests {
...
+                    "type": "number",
+                    "multipleOf": 2.0,
+                }),
+            ],
+            [json!({
+                "type": "number",
+                "const": 10.0,
+            })],
+        );
+    }
```
indietyp, line 1428: non-blocking, but I am guessing schema validation stops `minimum: 7, maximum: 5` from occurring on the same object?

indietyp (changes requested):

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
TimDiekmann: What's wrong with this message?

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/number.rs
TimDiekmann: Oh, good to know, thanks!

libs/@blockprotocol/type-system/rust/src/schema/data_type/constraint/mod.rs
indietyp: shouldn't it be: "should be able to parse schema"?

> We recommend that expect messages are used to describe the reason you expect the Result should be Ok.

from: https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.expect

You expect that the schema is correct, not that it is incorrect.

TimDiekmann: Personally I find this quite readable (ignoring the `untagged` thing 🙄):
> `Failed to parse schema: Error("data did not match any variant of untagged enum ValueConstraints", line: 0, column: 0)`

We use this format pretty scattered in the whole code base. As long as it's not `expect("schema to be valid")` I think we're good. I'm up for consistently change expect messages in the code base by actively searching for `.expect` (or even better use `Result` in tests instead, although this could become verbose). The other related tests in this crate uses the same format so I think we should just leave it untouched in this PR.

TimDiekmann: Actually, I created H-3498 for that: Consistently use expect message in code base

indietyp (approved):

PR #5394, opened 2024-10-13
H-3432: `harpc`: Expose background task

indietyp:

This adds an ergonomic way to spawn and create the background task that's required to operate without leaking memory.

Primarily it makes the use of `tokio::sync::broadcast` an implementation detail, thought that might make sense, as the user shouldn't care what we use under the hood. It's just a stream of events.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7232,6 +7232,7 @@ dependencies = [
  "futures-core",
  "pin-project-lite",
  "tokio",
+ "tokio-util",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


TimDiekmann (approved):
Why is clean-up needed at all, why the session storage cannot be cleaned up after a session was terminated?

indietyp:
this is for session termination, but we have no way to indicate to that a session has been terminated inline. so the net layer has a stream, with events, the only event (currently) is whenever a session is terminated. That is then relayed to this task which removes the session.

Another way this is often solved is through timing (e.g. 60 seconds of inactivity) but I wanted to avoid that approach.

PR #5409, opened 2024-10-16
H-3490: `harpc`: Implement tower client logic

indietyp:

This implements the tower client logic and fixes some things in the tower server implementation to make this possible.

You can run the tests yourself via `cargo run --example account --release -- server` for the server and `cargo run --example account --release` for the client example.

I know that this PR is probably quite a mess but I finishing the client <-> server integration was a bit harder.

Next steps here is to clean up, and remove the `ErrorEncoder`/`ErrorDecoder` as discussed.

This PR:

The changes in this PR:

The changes in this PR:

## 📹 Demo

https://github.com/user-attachments/assets/1474bc55-75aa-4805-bfe8-1eb8cbf7a029


libs/@local/harpc/net/src/session/server/session_id.rs
```diff
@@ -10,7 +10,12 @@ impl SessionIdProducer {
 
     pub(crate) fn produce(&mut self) -> SessionId {
         let id = self.next;
-        self.next = self.next.wrapping_add(1);
+        self.next += 1;
+
+        // never overflows because right before it reaches usize::MAX, it will be reset to 0
```
TimDiekmann, line 15: Could you add a reason why? (Because of the constant below). I think it's good to mention that `MAX` is a reserved special value.

libs/@local/harpc/server/examples/account.rs
```diff
@@ -1,41 +1,60 @@
 #![feature(never_type, impl_trait_in_assoc_type)]
 #![expect(
     clippy::unwrap_used,
-    clippy::empty_enum,
-    clippy::todo,
+    clippy::print_stdout,
+    clippy::use_debug,
     unused_variables,
-    reason = "non-working example code"
+    reason = "example"
```
TimDiekmann, line 7: I really laughed when I saw this change 😅

libs/@local/harpc/server/src/lib.rs
```diff
@@ -7,4 +7,121 @@ pub mod error;
...
+    Debug,
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    derive_more::Display,
+    derive_more::Error,
```
TimDiekmann, line 38: What is the reason to not use thiserror?

TimDiekmann (approved):

libs/@local/harpc/server/src/lib.rs
indietyp: `derive_more` was already a dependency. So I thought instead of pulling in `thiserror` into server, I'd just use that instead.

That raises a larger question - that I will revisit at some point - I've been increasingly dissatisfied with `thiserror` (`generic_member_access` not being a thing, wanting to derive `Display` - just like `Debug` - outside of error) to see where to go next. I personally like that `derive_more` is more of a compositional approach, but it doesn't implement `generic_member_access` (tho I personally feel like it'd be easier to contribute to `derive_more` than `thiserror`, than again - we have the proposal of #5352 with an example macro that might be separate from those two)

TimDiekmann: I absolutely hear you and deriving `Error` for a type manually when `Display` is implemented is quite easy as well (without `generic_member_access` it's a one-liner). The hard thing about `Error` _is_ `Display`. Personally, I would probably tend to use `derive_more` more generally over `thiserror` because of it's separation of concerns as it only derives one trait per statement.

TimDiekmann (approved):

vilkinsons (approved):

PR #5460, opened 2024-10-22
H-3507: Move `icon` and `labelProperty` to entity type schema

TimDiekmann:

This makes the `icon` and `labelProperty` part of the `EntityType` schema instead of the metadata. This enables fetching external types with these properties and bulk-creation of entity types with these is possible as well.

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-ts/src/activities/shared/find-existing-entity.ts
```diff
@@ -124,7 +124,7 @@ export const findExistingEntity = async ({
   /**
    * First find suitable specific properties to match on
    */
-  const labelProperty = entityType.labelProperty;
+  const labelProperty = entityType.labelProperty as BaseUrl | undefined;
```
TimDiekmann, line 127: An explicit cast is required because `BaseUrl` in the type system is an alias to `string`.

CiaranMn (commented):
Looks good, some linting issues that sadly aren't auto-fixable (removing unused imports)... think we could set them to be auto-fixable but it would be annoying for fix on save? Can't remember

TimDiekmann:
> think we could set them to be auto-fixable but it would be annoying for fix on save?

I think this was the motivation yes. However, if there would be a command which includes this autofix it would be a huge helper. Especially as ESLint is pretty slow..

CiaranMn (approved):

PR #5479, opened 2024-10-24
H-3518: Differentiate a `ClosedEntityType` and `ClosedMultiEntityType`

TimDiekmann:

We have three things we're interested for entity types:

- `EntityType`: the plain data type as a schema
- `ClosedEntityType`: An EntityType with all it's allOf merged into a single schema
- `ClosedMultiEntityType`: Entities may have multiple types and we need a representation to merge these as well.

`EntityType` remains unchanged. `ClosedEntityType` is the same as `EntityType` but without `allOf`. `ClosedMultiEntityType` has all `ClosedEntityTypes` stored in an `allOf` but merged `properties`, `links`, and `required` together at the top level.

## 🔍 What does this change?

- Change implementation for `ClosedEntityType` and distinguish between `ClosedMultiEntityType`
- Adjust logic to read parents (previously they were stored in the closed schemas

The new types look like this:
```ts
export interface EntityTypeSchemaMetadata {
    $id: VersionedUrl;
    title: string;
    titlePlural?: string;
    description?: string;
    inverse?: InverseEntityTypeMetadata;
    labelProperty?: BaseUrl;
    icon?: string;
}

export interface EntityConstraints {
    properties: Record<BaseUrl, ValueOrArray<PropertyTypeReference>>;
    required?: [BaseUrl, ...BaseUrl[]];
    links?: Record<VersionedUrl, PropertyValueArray<OneOfSchema<EntityTypeReference> | Record<string, never>>>;
}

export interface EntityType extends EntityTypeSchemaMetadata, EntityConstraints {
    $schema: EntityTypeSchemaTag;
    kind: EntityTypeKindTag;
    type: EntityTypeTag;
    allOf?: [EntityTypeReference, ...EntityTypeReference[]];
}

export interface ClosedEntityType extends EntityTypeSchemaMetadata, EntityConstraints {}

export interface ClosedMultiEntityType extends EntityConstraints {
    allOf: [EntityTypeSchemaMetadata, ...EntityTypeSchemaMetadata[]];
}
```

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The validation when creating/updating entities might be slightly slower as the parent schemas are not cached. This can be optimized but should be done after we have more benchmarks for this very code path in place.

## 🐾 Next steps

Expose querying of `ClosedEntityType` and `ClosedMultiEntityType` (H-3080)

## 🛡 What tests cover this?

The main thing to test in the PR is the usage of the types in validation. An extensive test case was added to cover a quite complex scenario.


libs/@local/hash-validation/src/lib.rs
```diff
@@ -361,18 +366,18 @@ mod tests {
...
                 let resolved_data = ontology_type_resolver
                     .resolve_entity_type_metadata(entity_type_uuid)
                     .expect("entity type not resolved");
-                let entity_type_id = entity_type.id.clone();
-                let closed_entity_type =
-                    ClosedEntityType::from_resolve_data(entity_type, &resolved_data)
-                        .expect("Could not close church");
-                (entity_type_id, closed_entity_type)
+                ClosedEntityType::from_resolve_data(entity_type, &resolved_data)
+                    .expect("Could not close church")
```
CiaranMn, line 380: Test log in the wrong place?
```suggestion
                    .expect("Could not close entity type")
```

CiaranMn (approved):
🚀 

Looks good, just one (existing) log that might need updating

CiaranMn (approved):

PR #5512, opened 2024-10-29
H-3528, H-3529: Set up Pdfium & preprocess PDFs as images

JesusFileto:

This intial PR serves as the first small step in the segmenting, chunking, and embedding package Chonky. Currently, we are setting up the environment to receive a file path to a pdf, and preprocess the pdf into images using `Pdfium-render` that will be used for structural embeddings later on. 

This PR also sets up the `Pdfium-render` to be used for text extraction. 

## 🔗 Related links

_(internal)_ 
[Implementation Doc](https://www.notion.so/HASH-Chunking-and-Embedding-LLD-1273c5e2fd9a80fb9900c8c7c964399e?pvs=4)

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Loading Pdfs given a local file path
- Convert Pdfs to page by page images

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Need for proper implementation and use of `error-stack` for error handling

## 🐾 Next steps

- Extracting textual information from pdfs using `pdfium-render` and enriching schema with document layout information
- Implement `clap` for parsing CLI arguments 

## 🛡 What tests cover this?

- New tests were created

## ❓ How to test this?

- running the command `cargo run tests/docs/test-doc.pdf` will save the pdf images to the `out` direction in the Chonky package


libs/chonky/Cargo.toml
```diff
@@ -17,6 +17,9 @@ edition.workspace = true
 error-stack = { workspace = true, public = true }
 
 # Public third-party dependencies
+pdfium-render = { version = "0.8.25", public = true }
+image = { version = "0.24.9", public = true }
+thiserror = { version = "1.0.65", default-features = false }
```
TimDiekmann, line 22: We typically store this in the top-level `Cargo.toml` and reference them here with `workspace = true`. Typically, we try to disable as many features as possible by default by setting `default-features = false`, this often results in a smaller sets of crates being required.
We already have `thiserror` listed there, so you can just use 
```toml
thiserror = { workspace = true }
```
The default-features is inheritace from the top-level. 

I don't think any of these libraries needs to be a public dependency as none of them is exposed in the API. By default you can just make them not public, clippy will complain if they should be public.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,106 @@
...
+    Arguments,
+}
+
+/// Function to read the pdf
 ///
-/// ```rust
-/// use chonky::add;
+/// # Errors
+/// Will return `Err` if `filename` does not exist or the user does not have
+/// permission to read it.
```
TimDiekmann, line 24: Minor code-style: it's more readable if there is a newline after headings
```suggestion
/// # Errors
///
/// Will return `Err` if `filename` does not exist or the user does not have
/// permission to read it.
```

We consistently have this in the repository.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,106 @@
...
+pub fn load_pdf<'a>(
+    pdfium: &'a Pdfium,
+    file_path: &str,
+) -> Result<PdfDocument<'a>, Report<PdfiumError>> {
+    let pdf = pdfium.load_pdf_from_file(file_path, None)?;
+    Ok(pdf)
```
TimDiekmann, line 30: There is no need to create a temporary file. However, the `Ok` and `?` is required here as it's converting `PdfiumError` to `Report<PdfiumError>`.
I'd go with either of these:
```suggestion
    pdfium.load_pdf_from_file(file_path, None).map_err(Report::new)
```
or 
```suggestion
    Ok(pdfium.load_pdf_from_file(file_path, None)?)
```

Either works, the first one is more explicit (and probably requires `rustfmt` to run 😅) and I probably like this more. Whatever suits you best!

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,106 @@
...
-pub const fn add(left: u64, right: u64) -> u64 {
-    left + right
+/// # Errors
+/// Return an `PdfiumError::ImageError` if there was a problem with the image processing operation,
+/// occurs when image cannot be encoded into specific image format
+pub fn pdf_to_images(pdf: &PdfDocument) -> Result<Vec<DynamicImage>, Report<PdfiumError>> {
+    let mut images: Vec<DynamicImage> = Vec::new();
+
+    for page in pdf.pages().iter() {
+        let resolution_width = 1000; //may adjust resolution depending on need
```
TimDiekmann, line 43: `rustfmt` for some reason sometimes has issues formatting the file when appending the comment. Also, when the comment gets to long it's likely that it's either not really readable or (worse) it's breaking into the next line, which could be read as a comment for the line after. This is clearer (and more consistent with the rest of the code):
```suggestion
        // may adjust resolution depending on the need
        let resolution_width = 1000;
```

libs/chonky/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+use chonky::{ChonkyError, load_pdf, pdf_to_images};
+use error_stack::{Report, ResultExt};
+use pdfium_render::prelude::*;
+
+fn main() -> Result<(), Report<ChonkyError>> {
+    let args: Vec<String> = env::args().collect(); // read file path arguments
+
+    if args.len() < 2 {
+        return Err(Report::new(ChonkyError::Arguments));
+    }
```
TimDiekmann, line 12: `error_stack` has a handy [macro](https://docs.rs/error-stack/latest/error_stack/macro.ensure.html) for this:
```rust
use error_stack::ensure;
```

```suggestion
    ensure!(args.len() > 1, ChonkyError::Arguments);
```

But this is only really usable for simple errors such as this one. The macro is agnostic over the second argument, so it can be either a `Report` or a plain error, the result will always be a `Report` wrapped in an `Err`.

TimDiekmann: cc @indietyp the documentation seems a bit broken, the `stringify` is not supposed to be shown I guess 😄 
image
I'm going to look into that at some point later: H-3536

TimDiekmann (commented):
Thanks @JesusFileto!

I had a look through the PR and so far it really looks good! I have a few minor suggestions, but really nothing critical.

libs/chonky/Cargo.toml
JesusFileto: Ok done, I do have to note that since pdfium and image are being returned by functions that are then used by main they do need to be public (for now), once we have a proper examples page we can wrap these implementations so they are not in the public interface.

libs/chonky/src/lib.rs
JesusFileto: Ok will make note for the future thanks!

libs/chonky/src/lib.rs
JesusFileto: I will most likely stick to the first one also because its clear in its intention.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,106 @@
...
+
+        let test_pdf_string = "tests/docs/test-doc.pdf";
+
+        let pdf = load_pdf(&pdfium, test_pdf_string).unwrap();
+
+        let preprocessed_pdf = pdf_to_images(&pdf).unwrap();
+
+        let num_pages = 38; //number of pages of pdf
+
+        assert_eq!(preprocessed_pdf.len(), num_pages) //length of vector should be number of pages
```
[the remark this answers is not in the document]
JesusFileto: Ok nice to know for future error handling!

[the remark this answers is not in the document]
TimDiekmann: Yes, I should do this more often as well... I find myself scattering assertions in tests but sometimes even a few strings help debugging if something is breaking unexpectedly.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,134 @@
...
+    ReadPdf,
+    #[error("pdfium error")]
+    Pdfium,
+    #[error("write error to system")]
+    Write,
+    #[error("Issues with CLI input")]
+    Arguments,
+}
+
+pub mod pdf_segmentation {
```
TimDiekmann, line 17: It's good practice to group functionality, thus a module makes sense. However, defining inline modules is quite rare (In this case the size of the module is quite small so it doesn't matter). In most cases you just define the module as in `pub mod pdf_segmentation;` and add a `pdf_segmentation.rs` with the content. (We don't need to do this now, maybe in the future it makes sense, but I guess as the crate grows we will probably shake the module tree quite a few times.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,134 @@
...
+            .load_pdf_from_file(file_path, None)
+            .map_err(|err| Report::new(err).change_context(ChonkyError::Pdfium))
+    }
+
+    /// Takes in a pdf document and returns a vector list where each page
+    /// is processed into a raw image that can be later converted to any image format
+    ///
+    /// # Errors
+    /// Return an `PdfiumError::ImageError` if there was a problem with the image processing
+    /// operation, occurs when image cannot be encoded into specific image format
```
TimDiekmann, line 44: It's possible to cross-link documentation ([intra-doc links](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html)). There is a quite easy rule: If it's in the module scope (e.g. as `DynamicImage`, it can be used directly, otherwise you can link it like it would be a markdown link. I.e.
```suggestion
    /// # Errors
    ///
    /// Return an [`PdfiumError::ImageError`] if there was a problem with the image processing
    /// operation, occurs when image cannot be encoded into specific image format
```
or
```suggestion
    /// # Errors
    ///
    /// Return an [`ImageError`] if there was a problem with the image processing
    /// operation, occurs when image cannot be encoded into specific image format
    ///
    /// [`ImageError`]: PdfiumError::ImageError
```

If a link is not valid, the CI will complain.

libs/chonky/src/lib.rs
JesusFileto: Ok I was planning on doing this once the codebase expands slightly more (as in more modules arise), but it would be better practice to start implementing the module file system now instead of later.

Cargo.toml
```diff
@@ -105,6 +105,7 @@ futures-sink = { version = "=0.3.31", default-features = false }
 futures-util = { version = "=0.3.31", default-features = false }
 hashbrown = { version = "=0.15.0", default-features = false, features = ["nightly", "inline-more"] }
 http = { version = "=1.1.0", default-features = false }
+image = { version = "0.24.9", default-features = true }
```
TimDiekmann, line 108: `default-features = true` is the default and doesn't need writing down. Anyway, we want no default features:
```suggestion
image = { version = "0.24.9", default-features = false }
```

For `image`, the [default features](https://docs.rs/crate/image/latest/features) are `rayon` and `default-formats`. It's not documented, but [`rayon`](https://docs.rs/rayon/latest/rayon/) is most likely used to speed up processing by putting it into threads. I propose to test if the feature flag is worth enabling by comparing the time it takes to process images. `default-formats` is probably not needed. I think it's better to explicitly mention what formats are supported.

The features should be enabled in `libs/chonky/Cargo.toml`, so if other libraries in the repository depend on `image` it doesn't require the same feature set, e.g.:
```toml
image = { workspace = true, public = true, features = ["png"] }
```

Cargo.toml
```diff
@@ -116,6 +117,7 @@ libp2p-identity = { version = "=0.2.9", default-features = false }
 libp2p-ping = { version = "=0.45.0", default-features = false }
 libp2p-swarm = { version = "=0.45.1", default-features = false }
 libp2p-yamux = { version = "=0.46.0", default-features = false }
+pdfium-render = { version = "0.8.25", default-features = true }
```
TimDiekmann, line 120: Same as above basically. `pdfium-render`, however, does [document the features](https://docs.rs/pdfium-render/latest/pdfium_render/#crate-features) and here the default features make sense
```suggestion
pdfium-render = { version = "0.8.25" }
```

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,134 @@
...
+    #[error("write error to system")]
+    Write,
+    #[error("Issues with CLI input")]
+    Arguments,
+}
+
+pub mod pdf_segmentation {
+    use error_stack::{Report, ResultExt};
+    use image::DynamicImage;
+    use pdfium_render::prelude::*;
```
TimDiekmann, line 20: We typically try to avoid glob-imports (except in tests, which often imports `super::*`). This pulls in a whole namespace, similar as you would do `std::*` in C++. `pdfium-render` seems a bit weird here because from the documentation it seems that _only_ a prelude is provided (and the prelude is quite big). Let's not use a glob-import here.

libs/chonky/libs/libpdfium.dylib
TimDiekmann, line 1: Do you have an idea, yet, how we can avoid uploading a `dylib` to the repository?

Cargo.toml
indietyp: I've worked with the image crate quite a few times in the past, and `rayon` only enables parallel iterators, about the `features`, it'd probably be worth it to have `0` features, `pdfium` directly pipes in RGBA data into a dynamic image, so doesn't need to decode anything.

libs/chonky/libs/libpdfium.dylib
indietyp: `pdfium` has a `static` feature that can be enabled - I think we should be able to make use of it - it is then embedded in the binary. You then load the library via: `Pdfium::new(Pdfium::bind_to_statically_linked_library()?);` instead.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,134 @@
...
+
+        let pdf = pdf_segmentation::load_pdf(&pdfium, test_pdf_string)
+            .change_context(ChonkyError::Pdfium)?;
+
+        let preprocessed_pdf =
+            pdf_segmentation::pdf_to_images(&pdf).change_context(ChonkyError::Pdfium)?;
+
+        let num_pages: usize = pdf.pages().len().into(); //number of pages of pdf
+
+        match preprocessed_pdf.len() {
```
indietyp, line 128: this isn't really a pressing issue, but it seems to me that this is only testing that `pdfium` converts a document into a number of pages - but we won't know what it'll look like. Maybe it'd make sense to take a smaller PDF (like 2 pages) and then use a kind of snapshot test - where we save the expected image output and the compare them. This way we're able to ensure that conversion actually takes place and that said conversion is consistent.

> as a side-note as it may be of interest - the snapshotting library we usually use, insta, has added experimental support for binary snapshots: https://github.com/mitsuhiko/insta/pull/610, while not really needed here - it might be interesting to see if it'd work here. Libraries like these allow us to more easily update snapshot tests, which is why I tend to use them.

libs/chonky/tests/docs/test-doc.pdf
indietyp, line 1: unsure about the procedure here, but I feel like this approaches a size where it may make sense to use git-lfs, @TimDiekmann what do you think?

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,134 @@
...
+            //may adjust resolution depending on need
+            let resolution_width = 1000;
+
+            // Render the entire page to an image
+            let rendered_page = page
+                .render_with_config(&PdfRenderConfig::new().set_target_width(resolution_width))
+                .change_context(ChonkyError::Pdfium)?; // Renders the page to a PdfBitmap
+
+            // Convert PdfBitmap to DynamicImage
+            let dynamic_image = rendered_page.as_image();
```
indietyp, line 58: `as_image` is a bit deceptive here, as I've worked with `image` before, converting anything into an image object is fallible. Looking at the source code that is also the case here, but it will [simply panic](https://docs.rs/pdfium-render/latest/src/pdfium_render/pdf/bitmap.rs.html#312), as all methods used are public it might make sense to vendor this function in the future to remove the panicking code.

indietyp (commented):
Hi! 👋  I am a contributor to HASH working on a couple of things and got curious when I saw this PR, so I thought I'd leave some comments. Don't feel any pressure to implement any of these, just hoping you'll find these helpful in some way! 😊

libs/chonky/tests/docs/test-doc.pdf
TimDiekmann: For small files, it's probably fine to just use plain git, but I thought about `git-lfs` as well.
I defer the decision to @CiaranMn and @vilkinsons.

libs/chonky/src/lib.rs
TimDiekmann: I think vendoring in this function makes sense until they return `Result<DynamicImage, PdfiumError>`, I agree. Good spot, thank you!

libs/chonky/tests/docs/test-doc.pdf
vilkinsons: I think ideally we’d try to avoid use of git-lfs here, and use smaller-sized/more compressed PDFs as test files.

libs/chonky/libs/libpdfium.dylib
TimDiekmann: How would that static library be built? We could look into build it in `build.rs` but that might be hard to achieve. (although it's probably fine to only support Linux (x86 and ARM) and macOS (Intel and ARM) for now).
Another way would be to require the library to be installed in the system as dylib and provide a linker path. 
Personally, I would probably use a dynamic library for now (but not uploading it to the repository) and try to builda static library in a follow-up PR. We require static linking in order to make a minimal Docker image, though, otherwise, we couldn't use `FROM scratch`.

libs/chonky/tests/docs/test-doc.pdf
CiaranMn: Once we are dealing with large files and they are only to be used in tests I would consider putting them somewhere else to be downloaded only when running the tests.

Adding them to `git-lfs` means that the latest version of the files have to be downloaded by anyone using the repo. If they are big enough to warrant putting them in `git-lfs`, that is going to make downloading the repo much slower everywhere.

We are also charged for storage _and_ bandwidth, including in GitHub actions. This can quickly add up if every CI run has to download a repo with huge files in it.

libs/chonky/libs/libpdfium.dylib
indietyp: Looking at the `build.rs` you see you set a env var. Might actually be worth a shot. (I tried `static` on `aarch64` and it worked without a problem)

see:
* https://github.com/ajrcarey/pdfium-render/blob/bc1195b23a777325afea96a5fc895e9f3db381c6/build.rs#L175-L198
* https://github.com/ajrcarey/pdfium-render?tab=readme-ov-file#static-linking

That way we'd be able to skip the dylib and keep the image minimal.

TimDiekmann: If it's that easy we should try it, yes :)

JesusFileto: Yes have been looking into linking but believe I may have some issues with my Rust configuration for mac, as I seem to be getting linker errors when using the env var `PDFIUM_DYNAMIC_LIB_PATH` with `aarch64`. 

Also found a handy script, [here](https://github.com/ajrcarey/pdfium-render/issues/53) that I am toying with to have static builds locally for the user that we can link to.

libs/chonky/src/lib.rs
JesusFileto: Yes, I'll look into some of the examples they have and try to implement this now especially if the library will be used in the rest of the package.

libs/chonky/tests/docs/test-doc.pdf
JesusFileto: Ok I will make sure to use smaller 1-2 page pdfs, and compress them. I decided to use one slightly larger pdf to get a sense of performance and scale during real-world use.

CiaranMn: We are going to have to use larger PDFs to meaningfully test it. So my view would be that we put anything that's _really big_ in an S3 bucket or similar and hardcode a URL. Yes, they're not version controlled, but they're not going to change and they can be publicly accessible and inspectible. 

But there's no harm in putting a few MB PDFs in the repo. So let's pick somewhere sensible between 1-2 pages and 100MB (where `git-lfs` kicks in). The [Alphabet annual report](https://abc.xyz/assets/52/88/5de1d06943cebc569ee3aa3a6ded/goog023-alphabet-2023-annual-report-web-1.pdf) is 2MB and 100 pages, for example.

CiaranMn: One of the advantages of `git-lfs` in managing large files is that it's not going to download all of the files in the history by default, and you can prune older versions if you're created many of them (e.g. you have a 50MB file and you change a few bytes repeatedly). Neither applies in this case because these aren't files we're going to change. We put one 2MB PDF in the repo and it's going to stay as a 2MB contribution to the entire history because it never changes. 

So until we want to put anything that's much bigger in I would just pick whatever selection of PDFs we think we need to meaningfully test it, and don't try and optimize for size. All of which to say: forget about size, choose a few PDFs that seem like they will act as good tests of the functionality. If we notice they're adding up to some problem we can offload them somewhere else (but not `git-lfs` given the cost implications).

libs/chonky/libs/libpdfium.dylib
indietyp: To fix this it's a two step process on MacOS. You can download a pre-compiled static library from [here](https://github.com/paulocoutinhox/pdfium-lib), pre-compiled dynamic libraries are available from [here](https://github.com/bblanchon/pdfium-binaries).

You then need to create a `build.rs` in the `chonky` library _not the src_ repository. `build.rs` files are run during compilation and can be used to modify the complication process, like doing code generation. In our case we need to instruct `chonky` to link against two additional libraries required to make this work on macos: `CoreGraphics` and `libc++`. (the `cc` library links against `libc++` automatically, but `pdfium-render` sadly doesn't make use of it).

(The `C++` linking shenanigans can also be done via another [crate](https://docs.rs/link-cplusplus/latest/link_cplusplus/)

This all boils down to the following `build.rs`:

```rs
fn main() {
    #[cfg(target_os = "macos")]
    link_macos();

    #[cfg(target_os = "linux")]
    link_linux();
}

#[cfg(target_os = "macos")]
fn link_macos() {
    // `pdfium` requires linking to the `CoreGraphics` and `libc++` frameworks on macOS
    println!("cargo:rustc-link-lib=framework=CoreGraphics");
    // `pdfium` also has the `libc++` feature, which can be used instead, but this makes it more
    // explicit that we need both.
    println!("cargo:rustc-link-lib=c++");
}

#[cfg(target_os = "linux")]
fn link_linux() {
    // `pdfium` also has a `libstdc++` feature, which can be used instead, but this makes it more
    // explicit that we need a different one depending on architecture.
    println!("cargo:rustc-link-lib=stdc++");
}
```

> Side Note: while looking at the `build.rs` for `pdfium-render` I found that `println!("cargo:rerun-if-env-changed=PDFIUM_STATIC_LIB_PATH");` is missing from it.

libs/chonky/src/lib.rs
```diff
@@ -164,15 +240,33 @@ mod tests {
...
+        // now check if the image contents are the same using insta snapshots
+        // start by converting images to binary
+        let preprocessed_pdf: Vec<Vec<u8>> = preprocessed_pdf
+            .into_iter()
+            .map(image::DynamicImage::into_bytes)
+            .collect();
+
+        // we only really need to check the first three pages
+        for (index, page_bytes) in preprocessed_pdf.into_iter().enumerate().take(3) {
+            assert_binary_snapshot!(format!("page_{}.BIN", index + 1).as_str(), page_bytes);
```
indietyp, line 267: I wonder if we can make use of encoding images as `bmp` here, as that process is lossless it should be fine, but it'd allow us to actually look at the images during review, which could be neat 😊

The change _should_ be relatively straight forward and I think could look like this (rough sketch) - encoding via the `image` crate isn't always super straightforward, so thought this little snippet may help^^

```rs
let mut buffer = Vec::new();
let encoder = image::codecs::bmp::BmpEncoder::new(&mut buffer);

page
    .write_with_encoder(encoder)
    .expect("image should be able to be encoded into a bitmap");

assert_binary_snapshot!(&format!("page_{}.bmp", index + 1), buffer);
```

libs/chonky/src/main.rs
TimDiekmann: ^ Fixed as part of #5533

libs/chonky/src/lib.rs
JesusFileto: Seeing the files does make comparing much easier! Just implemented this on my last commit.

vilkinsons:
Hey @JesusFileto, this is looking good now - thank you! Would it be possible to resolve conflicts on this branch? Then we can get this merged 🥳

TimDiekmann:
> @TimDiekmann Any action required in this PR on failing CI?

Yes, a few:
- `clippy` is currently failing (all seems to be auto-fixable using `--fix` - `rustfmt` afterwards is not unlikely to be required)
- The linking is not working, yet (and the `dylib` is still present)
- We need to figure out how the `Build` steps can be done to make it pass for forks, but that should be easy. I'll push a fix for this once the rest of the CI is green (`cache-to` has to be disabled I guess)

Also, a few comments from @indietyp are not resolved.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,270 @@
...
+        // since we assume dimensions must be positive this is not as issue
+        let width = bitmap
+            .width()
+            .try_into()
+            .map_err(|_foo| ChonkyError::Pdfium)?;
+
+        let height = bitmap
+            .height()
+            .try_into()
+            .map_err(|_foo| ChonkyError::Pdfium)?;
```
TimDiekmann, line 170: Personally, I try to use `(try_)from` over `(try_)into` because it explicitly tells the reader what the expected type is. This means here it would be `i32::try_from(bitmap.width())`.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,270 @@
...
+        Ok(bitmap)
+    }
+
+    ///A vendored function from pdfium-render's `as_image` function that returns a result instead
+    /// of panicking
+    ///
+    /// Errors#
+    ///
+    /// [`PdfiumError::ImageError`] when the image had an error being processed
+    fn as_image(bitmap: &PdfBitmap) -> Result<DynamicImage, ChonkyError> {
```
TimDiekmann, line 157: To capture lines on errors it probably make sense to return a `Report` here:
```suggestion
    fn as_image(bitmap: &PdfBitmap) -> Result<DynamicImage, Report<ChonkyError>> {
```

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,270 @@
...
+        let bytes = bitmap.as_rgba_bytes();
+
+        // clippy complains if we directly cast into u32 from i32 because of sign loss
+        // since we assume dimensions must be positive this is not as issue
+        let width = bitmap
+            .width()
+            .try_into()
+            .map_err(|_foo| ChonkyError::Pdfium)?;
```
TimDiekmann, line 165: And with the `Err` variant of `Result` being a `Report` you can directly call `change_context`:
```suggestion
        let width = i32::try_from(bitmap.width())
            .change_context(ChonkyError::Pdfium)?;
```
This way, the error message from [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html) is captured as well.

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,270 @@
...
+
+        let height = bitmap
+            .height()
+            .try_into()
+            .map_err(|_foo| ChonkyError::Pdfium)?;
+
+        Ok(match bitmap.format().map_err(|_foo| ChonkyError::Pdfium)? {
+            #[expect(deprecated)]
```
TimDiekmann, line 173: Could you add why we use deprecated functionality here? You can do this in the `reason` attribute:
```rust
#[expect(deprecated, reason = "...")]
```

libs/chonky/src/lib.rs
```diff
@@ -1,26 +1,270 @@
...
+
+        //number of pages of pdf
+        let num_pages: usize = pdf.pages().len().into();
+
+        let preprocessed_pdf =
+            pdf_segmentation::pdf_to_images(&pdf).change_context(ChonkyError::Pdfium)?;
+
+        //start by checking if proper amount of images are converted
+        match preprocessed_pdf.len() {
+            x if x == num_pages => (),
```
TimDiekmann, line 244: As this is the only match arm you could replace the `match` with `if x != num_pages { return Err(...) }`

libs/chonky/src/lib.rs
JesusFileto: Originally included this because some of the bitmap formats were marked for deprecation, but instead I will reformat it to return errors if using these formats

libs/chonky/libs/libpdfium.dylib
JesusFileto: In attempting to use the documentation provided [here](https://github.com/ajrcarey/pdfium-render/tree/master?tab=readme-ov-file#static-linking), using the environment variable `PDFIUM_DYNAMIC_LIB_PATH` performs proper symbolic linking to the path of `.dylib`. But panics during runtime giving `Library not loaded: ./libpdfium.dylib`, seeming there is no testing or other documentation in the codebase for this function, it may not be fully implemented/tested. I will log an issue for this for next steps. 

From this point there are two options:
- Remove the static package and use environment variables in the `lib.rs` file to bind to a hardcoded directory we require the user to include. This would remove the possibility of static binding for now.
- Wait for the issue to be answered on github to not modify the codebase and default to static linking (since this is our goal)

I think the best option is to removing the dynamic library now to merge these changes, since most of the work for creating a static package have not been implemented.

Since

JesusFileto: Looking more into the static linking, I found that the user would need to set the environment variable in the command line since `pdfium-render` uses `std::env` to read the file path, and environment variables in dependencies can only be set in the command line.

libs/chonky/src/snapshots/chonky__tests__page_1.snap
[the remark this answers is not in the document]
TimDiekmann: I added the `.gitattributes` file but did not move the snapshots as the configuration is easier this way.

TimDiekmann (approved):

PR #5527, opened 2024-11-01
H-3225: Allow querying of closed data schemas

TimDiekmann:

Closed data types are stored in the database but the graph currently cannot query them.
## 🔍 What does this change?

- After quite a bit of back and forth (thanks to @indietyp for the discussion about RTNs!) I moved out basic lookup of data types into a trait which we can improve later. (Future improvements covers building a caching layer around the trait or providing a proper in-memory store)
- This currently uses the `trait-variant` crate for simplicity (the issues we encountered don't apply in this PR).
- Implement (fairly simple) querying of closed data type schemas

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-graph-types/rust/src/ontology/data_type/lookup.rs
```diff
@@ -0,0 +1,46 @@
...
+use type_system::{
+    schema::{ClosedDataType, ConversionExpression, DataTypeReference, DataTypeUuid},
+    url::BaseUrl,
+};
+
+use crate::ontology::DataTypeWithMetadata;
+
+#[trait_variant::make(Send)]
+pub trait DataTypeLookup {
+    type DataTypeWithMetadata: Borrow<DataTypeWithMetadata> + Send;
```
indietyp, line 13: Are you using `Borrow` here for the additional guarantees, or would `AsRef` be sufficient here?

indietyp (changes requested):
one question, otherwise LGTM!

libs/@local/hash-graph-types/rust/src/ontology/data_type/lookup.rs
TimDiekmann: `AsRef` would probably be sufficient. But as `Borrow` has stronger guarantees (at least semantically) I chose that. Do you think I should change it?

TimDiekmann: Oh, and we use `Borrow` in other locations as well 😅

indietyp: I just wanted to make sure that you intended that those stronger guarantees are upheld^^

indietyp (approved):

indietyp (approved):

PR #5533, opened 2024-11-02
GEN-202: Deprecate `error_stack::Result` and `error_stack::Context`

TimDiekmann:
### `Context`

`core::error::Error` is stabilized and we don't need an extra trait which is doing the same thing.

### `Result`

`error_stack::Result<T, C>` is an alias to `core::result::Report<T, Report<C>>` and sometimes saves one word when typing. However, as a Rust developer one is _very_ used to `Result` and if `Result` behaves differently or has implications, this is not good. In comparison to other libraries such as `anyhow`, `error_stack` requires the context to be set as the `Err` variant, it replaces the `Result` instead of extending it. `error-stack` prefers an explicit API interface and the `Result` alias does not help.
In addition, it's almost not possible to copy code from `error-stack` as in regular cases the target still use `core::result::Result` or a mixture.

We're going to require `as _` syntax for traits in the very near future. To avoid touching the code twice and keep this PR clean, I pinned `error-stack` in the workspace to the current commit on main. 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/error-stack/src/context.rs
+++ b/libs/error-stack/src/context.rs
@@ -17,7 +17,7 @@ use crate::Report;
 /// ```rust
 /// use std::{fmt, fs, io};
 ///
-/// use error_stack::{Context, Result, ResultExt, Report};
+/// use error_stack::{Context, ResultExt, Report};
 ///
 /// # type Config = ();
 /// #[derive(Debug)]
@@ -38,12 +38,12 @@ use crate::Report;
 /// // `Context` manually.
 /// impl Context for ConfigError {}
 ///
-/// pub fn read_file(path: &str) -> Result<String, io::Error> {
+/// pub fn read_file(path: &str) -> Result<String, Report<io::Error>> {
 ///     // Creates a `Report` from `io::Error`, the current context is `io::Error`
 ///     fs::read_to_string(path).map_err(Report::from)
 /// }
 ///
-/// pub fn parse_config(path: &str) -> Result<Config, ConfigError> {
+/// pub fn parse_config(path: &str) -> Result<Config, Report<ConfigError>> {
 ///     // The return type of `parse_config` requires another context. By calling `change_context`
 ///     // the context may be changed.
 ///     read_file(path).change_context(ConfigError::ParseError)?;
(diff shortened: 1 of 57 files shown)


libs/error-stack/src/context.rs
```diff
@@ -56,6 +58,7 @@ use crate::Report;
 /// # assert!(err.contains::<io::Error>());
 /// # assert!(err.contains::<ConfigError>());
 /// ```
+#[deprecated(note = "Use `core::error::Error` instead", since = "0.6.0")]
```
indietyp, line 61: should we say instead `Use core::error::Error + Send + Sync + 'static` instead? Because that's the newly required trait bound.

libs/error-stack/src/fmt/mod.rs
```diff
@@ -754,7 +756,7 @@ fn collect<'a>(root: &'a Frame, prefix: &'a [&Frame]) -> (Vec<&'a Frame>, &'a [F
     (stack, next)
 }
 
-/// Partition the tree, this looks for the first `Context`,
+/// Partition the tree, this looks for the first `ontext`,
```
indietyp, line 759:
```suggestion
/// Partition the tree, this looks for the first context frame,
```

libs/error-stack/src/macros.rs
```diff
@@ -354,14 +359,14 @@ macro_rules! bail {
 /// impl fmt::Display for PermissionDenied {
 ///     # #[allow(unused_variables)]
 ///     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-///         const _: &str = stringify! {
+///         # const _: &str = stringify! {
```
indietyp, line 362: nice drive-by!

indietyp (changes requested):
Thank you! Just some minor comments. 🎉

libs/error-stack/src/context.rs
TimDiekmann: In most cases the user just implements `Context`. Only if you want to be generic over the error you need those bounds. I think this is good enough.

indietyp (approved):

PR #5543, opened 2024-11-02
H-3549: Enable `unused_trait_names` lint

TimDiekmann:

A great lint which is now auto-fixable. Also, Rust Analyzer supports importing `as _`.

## 🔍 What does this change?

- Use `as _` for traits where only the methods are required.


tests/hash-graph-integration/postgres/sorting.rs
```diff
@@ -78,7 +78,7 @@ async fn test_root_sorting<A: AuthorizationApi>(
                 },
                 sorting: EntityQuerySorting {
                     paths: sorting_paths.clone(),
-                    cursor: cursor.take(),
+                    cursor: Option::take(&mut cursor),
```
indietyp, line 81: sneaky. :P

indietyp (dismissed):

indietyp (approved):

PR #5578, opened 2024-11-04
H-3543: Add interface to read closed multi-entity type schemas

TimDiekmann:

We are able to query closed multi-entity types from the graph but it would be more helpful if it could be passed when querying entities.

## 🔍 What does this change?

- Add an `includeClosedMultiEntityTypes = false` to the parameters when requesting entities or an entity subgraph
- Create `getClosedMultiEntityTypesFromResponse` in the `hash-graph-sdk` which takes the response object and the entity type ids. That should give the same type as a query to the graph with the same entity type ids.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The graph implementation might not be the fastest, but until we benchmark this code path there is no point in optimizing this.

## 🛡 What tests cover this?

Tests were added to compare the results of the returned query against the schema returned from the graph when querying the types independently.


libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -463,6 +470,32 @@ export const flattenPropertyMetadata = (
...
+  const [firstEntityTypeId, ...restEntityTypesIds] = entityTypesIds.sort();
+  let currentClosedMultiEntityTypeMap =
+    response.closedMultiEntityTypes[firstEntityTypeId];
+
+  for (const id of restEntityTypesIds) {
+    if (!currentClosedMultiEntityTypeMap?.inner) {
+      return;
+    }
+    currentClosedMultiEntityTypeMap = currentClosedMultiEntityTypeMap.inner[id];
+  }
```
CiaranMn, line 490: Would like to understand what is going on here, I would have thought a closed multi-entity type would be keyed under some combination of all its typeIds but here we have items keyed by a single typeId and are returning whatever the item in the map with the last entityTypeId in the list is (unless one is missing, in which case we return nothing). Why is it like this, and what happens if there are multiple multi-type entities which have the same entityTypeId amongst their types?

CiaranMn: Oh, it's a nested map which is nested to however many levels there are numbers of types for the entity... is this what we want long-term, or is it just temporary? I thought we'd just sort and concatenate the typeIds or something. Is there a benefit to doing it this way?

libs/@local/hash-graph-sdk/typescript/src/entity.ts
```diff
@@ -463,6 +470,32 @@ export const flattenPropertyMetadata = (
...
 
+export const getClosedMultiEntityTypesFromResponse = (
+  response: GetEntitySubgraphResponse | GetEntitiesResponse,
+  entityTypesIds: [VersionedUrl, ...VersionedUrl[]],
+): ClosedMultiEntityType | undefined => {
+  if (!response.closedMultiEntityTypes) {
+    return;
+  }
+
+  const [firstEntityTypeId, ...restEntityTypesIds] = entityTypesIds.sort();
```
CiaranMn, line 481: I'm not sure it really matters here, but to note that `sort` mutates the original array, we can use this newish alternative to avoid mutating the input array
```suggestion
  const [firstEntityTypeId, ...restEntityTypesIds] = entityTypesIds.toSorted();
```

libs/@local/hash-graph-sdk/typescript/src/entity.ts
TimDiekmann: I don't know if we would keep it longer term. This is the reason why I have "hidden" the implementation behind the `response` object. I'm currently writing a few comments to explain this more. Also, I try to rewrite it to use `reduce` instead.
The reason why I used this is because it's easier to build and does not need to reallocate strings when concatenating strings. It can be iteratively be built in the graph as it's layered.

libs/@local/hash-graph-sdk/typescript/src/entity.ts
TimDiekmann: This is one of the reasons why I like Rust, if there is a side effect it's impossible to have an existing reference to `entityTypeIds`, so it's possible to just sort the array and profit from it later again if we call `sort()` again (I assume `O(n)` for sorted arrays).
I changed it to `toSorted` but had to add `!` to `firstEntityTypeId` as it does not retain the exact type of the array.

PR #5597, opened 2024-11-07
H-3557: Introduce `Subsystem` and `SubsystemIdentifier`

indietyp:

This introduces the `Subsystem` and `SubsystemIdentifier`, by renaming `harpc_service::Service` to `Subsystem`. Additionally the `harpc_server` has been made aware of the `SubsystemIdentifier` and will now only allow routes that have the same subsystem.

The change of term has also been propagated to the underlying layers.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/libs/api/src/rpc/mod.rs
+++ b/apps/hash-graph/libs/api/src/rpc/mod.rs
@@ -3,6 +3,9 @@ pub mod auth;
 pub mod echo;
 mod session;
 
+use harpc_service::SubsystemIdentifier;
+use harpc_types::subsystem::SubsystemId;
+
 mod role {
     use harpc_client::connection::Connection;
     use harpc_server::session::Session;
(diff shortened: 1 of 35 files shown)


apps/hash-graph/libs/api/src/rpc/mod.rs
```diff
@@ -13,3 +16,32 @@ mod role {
...
+    }
+
+    fn into_id(self) -> SubsystemId {
+        match self {
+            Self::Echo => SubsystemId::new(0x00),
+            Self::Authentication => SubsystemId::new(0x01),
+            Self::Account => SubsystemId::new(0x02),
+        }
+    }
+}
```
TimDiekmann, line 47: Do we want to use `From` instead?

TimDiekmann (dismissed):
Thank you!

One non-blocking question.

apps/hash-graph/libs/api/src/rpc/mod.rs
indietyp: We'd need to use `TryFrom`, but there isn't really an error here. Simply that it doesn't match. Otherwise we could say: `SubsystemIdentifier: Into<SubsystemId>`. I wanted to keep this separate tho to make it clear that it's a specific operation with intent. In the future (once it lands as a non-incomplete feature) I'd like to make these actually const as a requirement.

TimDiekmann: Makes sense!

TimDiekmann (approved):

PR #5600, opened 2024-11-07
H-3564: Use consistent library structure for Rust crates

TimDiekmann:

Currently it's a wild mixture of `apps/*/lib` and `libs/@local` and the crate names varies between no prefix, `graph-` prefix and `hash-graph` prefix. As we're going to use the `hash` namespace, all crates should use hash- as prefix in the meantime. This has the advantage of them being grouped together in imports, it's easier to see where the packages are coming from, and what part of the project they belong to. 

## 🔍 What does this change?

- Remove `query-builder-derive` (unused and not expected to be used due to HQL)
- Prefix HASH-related crates with `hash-`
- Prefix Graph-related crates with `graph-` (in addition to `hash-`, so the prefix becomes `hash-graph-`)
- **Move graph-libraries out of `apps/` (into `libs/@local/graph`)**
- Move graph-binary to `apps/hash-graph`

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/graph/graph/Cargo.toml
TimDiekmann, line 1: This crate will be split into other crates as part of `H-1000`

indietyp (approved):

vilkinsons (approved):

tests/hash-graph-test-data/rust/LICENSE.md
```diff
@@ -1,607 +0,0 @@
-GNU Affero General Public License
```
vilkinsons, line 1: @TimDiekmann What's the rationale behind the deletion of this file? We should probably just move it up a level, if thought to be inappropriate here, in keeping with licensing primary app components under AGPLv3.

TimDiekmann: It was not deleted, GitHub just didn't show moved files correctly. The removed file is from query-builder-derive which is fully removed now.
https://github.com/hashintel/hash/blob/t/h-3564-use-consistent-library-structure-for-rust-crates/tests/graph/test-data/rust/LICENSE.md

CiaranMn (approved):

PR #5608, opened 2024-11-08
H-3565: Remove `Role` and use associated type in subsystem traits

indietyp:

`Role` was a trait that was used in an earlier iteration of `harpc` and is redundant.

> In case we want a separation (as that still might be useful, we could add a: `ExecutionContext` associated type, said type would then indicate if it is either: `Server`, `Client` or `Implementation`.

I moved the `Session` (now called `ExecutionScope`) into an associated type. Both full blown generics and associated types have their pros and cons, which - for this case - i have listed below. In the future we might want to look back at this if we find any case that is truly unergonomic (the biggest I see here is taking specific clients as arguments, as they now require 2 generics).

Having the `ExecutionScope` as an associated type has some benefits:

1) _very minor_/stylistic: it allows us to have a more expressive name for the type, usually in Rust, generic parameters are just short one letter names instead
2) Typing gets a whole lot easier
3) it's more readable (a lot less `<` nesting)

But it also comes with some negatives that I found, which is why existing traits, like [Service](https://docs.rs/tower/latest/tower/trait.Service.html) tend to have a `Request` as a generic parameter instead:

1) Generic types that we require need to move to the type that implements them, in our case this can be seen in each client implementation, which now requires 2 marker generics
2) types that are completely generic over the execution scope need to have a generic (e.g. they don't care what the scope is). **this does not apply to wrapping services, they can just do `type ExecutionScope = T::ExecutionScope`**

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/libs/api/src/rpc/account.rs
```diff
@@ -576,30 +581,51 @@ where
...
-pub struct AccountClient;
+#[derive_where::derive_where(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct AccountClient<S, C> {
+    _session: PhantomData<fn() -> *const S>,
+    _codec: PhantomData<fn() -> *const C>,
+}
+
+impl<S, C> AccountClient<S, C> {
+    #[must_use]
+    pub fn new() -> Self {
```
TimDiekmann, line 592:
```suggestion
    pub const fn new() -> Self {
```

apps/hash-graph/libs/api/src/rpc/account.rs
```diff
@@ -623,28 +649,27 @@ where
...
         account_group_id: AccountGroupId,
         account_id: AccountId,
     ) -> Result<(), Report<AccountError>> {
-        invoke_call_discrete(session, meta::AccountProcedureId::AddAccountGroupMember, [
-            (account_group_id, account_id),
-        ])
+        invoke_call_discrete(scope, meta::AccountProcedureId::AddAccountGroupMember, [(
+            account_group_id,
+            account_id,
+        )])
```
TimDiekmann, line 659: Sometimes `rustfmt` is weird. A shorter name results in more lines 🙃

apps/hash-graph/libs/api/src/rpc/account.rs
```diff
@@ -623,28 +649,27 @@ where
...
         account_id: AccountId,
     ) -> Result<(), Report<AccountError>> {
-        invoke_call_discrete(
-            session,
-            meta::AccountProcedureId::RemoveAccountGroupMember,
-            [(account_group_id, account_id)],
-        )
+        invoke_call_discrete(scope, meta::AccountProcedureId::RemoveAccountGroupMember, [
+            (account_group_id, account_id),
+        ])
```
TimDiekmann, line 672: And sometimes not 😂

apps/hash-graph/libs/api/src/rpc/echo.rs
```diff
@@ -137,28 +141,33 @@ where
...
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
-pub struct EchoClient;
+#[derive_where::derive_where(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct EchoClient<S, C> {
+    _service: PhantomData<fn() -> *const S>,
+    _codec: PhantomData<fn() -> *const C>,
+}
```
TimDiekmann, line 156: We probably want to add a `new` and `default` method as well.

libs/@local/harpc/server/examples/account.rs
```diff
@@ -115,146 +120,78 @@ impl Procedure for CreateAccount {
...
-    #[error("unable to encode request")]
-    Encode,
-    #[error("unable to decode response")]
-    Decode,
-    #[error("expected at least a single response")]
-    ExpectedResponse,
-}
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, derive_more::Display, derive_more::Error)]
+#[display("unable to fullfil account request")]
+pub struct AccountError;
```
TimDiekmann, line 125: Let's do that for any error we touch:
```suggestion
#[must_use]
pub struct AccountError;
```

TimDiekmann (commented):
Big fan of this, thanks for doing! Let's slightly tweak it, resolve the conflicts (only a few files have been moved) and merge this! 🚀

libs/@local/harpc/server/examples/account.rs
TimDiekmann: Also, let's try to use `derive_more` more often (as you did in this PR)

indietyp: see [575d32b](https://github.com/hashintel/hash/pull/5608/commits/575d32b4973e2b0570429e8a7880f917590c633e)

apps/hash-graph/libs/api/src/rpc/auth.rs
```diff
@@ -151,19 +153,40 @@ where
...
-pub struct AuthenticationClient;
+pub struct AuthenticationClient<S, C> {
+    _service: PhantomData<fn() -> *const S>,
+    _codec: PhantomData<fn() -> *const C>,
+}
 
-impl<Svc, C> AuthenticationSystem<role::Client<Svc, C>> for AuthenticationClient
+impl<S, C> AuthenticationClient<S, C> {
+    #[must_use]
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
indietyp: see: [575d32b](https://github.com/hashintel/hash/pull/5608/commits/575d32b4973e2b0570429e8a7880f917590c633e)

apps/hash-graph/libs/api/src/rpc/account.rs
indietyp: see: [575d32b](https://github.com/hashintel/hash/pull/5608/commits/575d32b4973e2b0570429e8a7880f917590c633e)

TimDiekmann (approved):

PR #5609, opened 2024-11-08
H-1000: Split the Graph into multiple lib modules

TimDiekmann:

Splitting the graph into multiple crates has some major advantages:
- Required for H-387: Unify interfaces for different graph endpoints (e.g. API, validation, testing, etc.)
- Compilation is faster, which is a big DX when developing Rust
- Crates can be checked/tested standalone (which has a huge impact on DX as seen with the validation crate)
- If dedicated features are implemented in dedicated crates it's easier to understand the crates, especially for new devs
- Avoid cyclic dependencies in @apps/hash-graph - a package which serves as a dump for graph-related things but is not really used directly.

Many of these expected crates were already defined but not properly used, yet, for example the `store` crate only holds partial definitions and the `DataTypeStore`. This PR moves the definitions from the `graph` crate and distributes it to the other crates (luckily, I did a lot of these as a drive-by previously)

## 🔍 What does this change?

- Move `migrations` module to `store`
- Remove the unused `ReadParameter` struct
- Move `PropertyTypeStore`, `EntityTypeStore`, and `EntityStore` to `store`
- Move `StorePool` to `store`
- Move `Read` traits to `store` and allow implementing `Read` from other crates
- Move type-fetcher related implementations and definitions to `type-fetcher`
- Rename `graph` to `postgres-store` (only the Postgres and snapshot implementation remained)

Most notable is that `validation` now depends on `store`, not the other way around.

Before:
image

After:
image

Generally, this moves all `store` related declarations further up

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons (changes requested):
@TimDiekmann As part of this PR, please can we add descriptive READMEs that explain lib/app contents? e.g. to:

- [`libs/@blockprotocol/type-system/rust`](https://github.com/hashintel/hash/tree/t/h-1000-split-the-graph-into-multiple-lib-modules/libs/%40blockprotocol/type-system/rust) - or moving it up a level to the `type-system` directory and combining it with the existing one found in the TypeScript peer folder
- [libs/@local/graph`](https://github.com/hashintel/hash/tree/t/h-1000-split-the-graph-into-multiple-lib-modules/libs/%40local/graph) and/or all of its child directories (I think the former is fine)

I also think we should add them for other library directories not touched here, like [`libs/@local/codec`](https://github.com/hashintel/hash/tree/main/libs/%40local/codec) and [`libs/@local/harpc`](https://github.com/hashintel/hash/tree/main/libs/%40local/harpc) which I note also has a `codec` subdirectory (cc @indietyp if you have thoughts).

This documentation describing the architecture and individual package boundaries/goals/intentions will be invaluable in helping us publicly document the boundaries between things, making it easier for future OSS contributors, and prove useful when we provide the repository as context to AI workers in the future.

TimDiekmann:
> As part of this PR, please can we add descriptive READMEs that explain lib/app contents?

Sure! But I only roughly mentioned in the READMEs what a crate is doing as otherwise they'll fall out of sync too quickly. Once, we host the dev-docs it should be easy to just explore these. For now I added an instruction on how to generate the docs in the `graph` folder (probably the most important part of all of it). 

I also experimented around with generating a diagram of the dependencies. Ideally I want to include that in the in-doc docs but I didn't manage to automatically generate them, yet. I'll probably take a look at this later again, but that's too much for this PR. 
I achieved this so far but that involved quite a bit of manual work/AI work:
image

For `HaRPC` I created H-3566 and for `hash-codec` I created H-3567.

vilkinsons (approved):

indietyp (approved):

PR #5610, opened 2024-11-08
Expose RPC server in binary

indietyp:

It finally happened, the RPC server is now exposed. This has been a long time coming and a huge milestone. Excited to try it out in different scenarios!

> Most of the code here is actually a boxed router implementation that I didn't end up needing^^ and relaxation of `Body` bounds

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-graph/Cargo.toml
```diff
@@ -48,6 +48,9 @@ tokio = { workspace = true }
 tokio-postgres = { workspace = true }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
+harpc-server.workspace = true
+multiaddr.workspace = true
```
TimDiekmann, line 52:
```suggestion
harpc-server = { workspace = true }
multiaddr = { workspace = true }
```

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -67,10 +105,18 @@ pub struct ServerArgs {
...
 
-    /// The address the REST client is listening at.
+    /// The address the REST server is listening at.
     #[clap(flatten)]
     pub api_address: ApiAddress,
 
+    /// Enable the experimental RPC server.
+    #[clap(long, default_value_t = false, env = "HASH_GRAPH_RPC_ENABLED")]
+    pub rpc_enabled: bool,
+
```
TimDiekmann, line 115: Alternatively we could disable it if no address is set

libs/@local/harpc/server/src/router.rs
```diff
@@ -167,6 +163,34 @@ pub struct Router<R> {
...
+        reason = "false positive, the router may still be Send + Sync, we just don't enforce it \
+                  on boxing"
+    )]
+    #[must_use]
+    pub fn boxed(self) -> BoxedRouter
+    where
+        R: Route<BoxReqBody, Future: Send + 'static, ResponseBody: Send + 'static> + 'static,
+    {
+        let routes = Arc::try_unwrap(self.routes).unwrap_or_else(|_routes| {
+            panic!("`Router::boxed()` should be called before any requests have been serviced")
```
TimDiekmann, line 183: Isn't this just an `.expect`?

libs/@local/graph/api/src/rpc/mod.rs
```diff
@@ -34,3 +56,47 @@ impl SubsystemIdentifier for GraphSubsystemId {
...
+        .with_builder(|builder| {
+            builder
+                .layer(HandleReportLayer::new())
+                .layer(HandleBodyReportLayer::new())
+        })
+        .register(AuthenticationDelegate::new(AuthenticationServer))
+        .register(AccountDelegate::new(AccountServer {
+            store_pool: dependencies.store,
+            authorization_api_pool: dependencies.authorization_api,
+            temporal_client: dependencies.temporal_client.map(Arc::new),
```
TimDiekmann, line 93: I didn't know that the account system requires a temporal client, however, it also doesn't hurt to have it. Especially because we can use this code for other systems as well.

TimDiekmann (commented):
Great, we are very close!

apps/hash-graph/Cargo.toml
indietyp: see: [03a1ceb](https://github.com/hashintel/hash/pull/5610/commits/03a1cebc399918262c2176fe80e7af9ed8d89e59) (i forgot to push that one last night^^)

apps/hash-graph/src/subcommand/server.rs
indietyp: I thought about that as well, but I wanted to give the opportunity to have a default for the port and address as well (also I was unsure how clap would handle a `clap(flatten)` on an `Option<RpcAddress>` - not that I tried). I feel like this way you can still have a default which is nice.

I am not opposed to make it optional tho.

libs/@local/graph/api/src/rpc/mod.rs
indietyp: The temporal client is required for `store_pool.acquire`, see: https://github.com/hashintel/hash/blob/03a1cebc399918262c2176fe80e7af9ed8d89e59/libs/@local/graph/api/src/rest/account.rs#L176-L177 and https://github.com/hashintel/hash/blob/03a1cebc399918262c2176fe80e7af9ed8d89e59/libs/@local/graph/api/src/rpc/account.rs#L236-L237

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -59,6 +65,38 @@ impl TryFrom<ApiAddress> for SocketAddr {
...
 }
 
+#[derive(Debug, Clone, Parser)]
+pub struct RpcAddress {
+    /// The host the RPC client is listening at.
+    #[clap(long, default_value = "127.0.0.1", env = "HASH_GRAPH_RPC_HOST")]
+    pub rpc_host: String,
+
+    /// The port the RPC client is listening at.
+    #[clap(long, default_value_t = 25489, env = "HASH_GRAPH_RPC_PORT")]
```
indietyp, line 75: This port has been expertly chosen by a random number generator. Thought that would be better than putting it adjacent to 4001 or something similar, as that range is usually quite contended with dev servers and such.

libs/@local/harpc/server/src/router.rs
TimDiekmann: That's true.

libs/@local/graph/api/src/rpc/mod.rs
TimDiekmann: Ah, yes, not very proud of this part, though :D (let's keep it as it is, that has nothing to do with this PR).

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: We can probably use `4002`

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: I hear your argument about defaults. This also again implies the issue of unsetting the address if it's set in `.env`. I'm fine with leaving it. At some point we need to overhaul the environment variables anyway.

apps/hash-graph/src/subcommand/server.rs
indietyp: 🤝 

[5289323](https://github.com/hashintel/hash/pull/5610/commits/52893235939b74642d9064b2f5b8ac30e6c82898)

TimDiekmann (approved):

PR #5645, opened 2024-11-14
H-3590: Allow all CI to pass when opening PRs from a fork

TimDiekmann:

When opening PRs from a fork some CI cannot pass because they require authentication. To avoid this issue, this PR disables these steps (they will still be checked when running the merge-queue)

## 🔍 What does this change?

- Avoid uploading the cache for Build steps
- drive-by: Don't upload the Build results as artifact anymore, it's not used
- Use `checkout` action for Benchmark CI instead of manual commands
- Don't post the Benchmark-result as comment (the action runs from a fork so we cannot use `GH_TOKEN` here) - this can possibly be avoided by adding a `pull_request_target` run but that requires another CI workflow as we must not run the benchmarks using `pull_request_target`.
- Don't upload source maps to Sentry
- Don't attempt to plan a Terraform run

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    if: github.repository == 'hashintel/hash'
+    if: github.repository == github.repository
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    if: github.repository == github.repository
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    if: github.repository == 'hashintel/hash'
+    if: github.repository == github.repository
```
CiaranMn, line 12: Why this change? It will always be true, releasing to npm isn't going to work on forks, and it'll only run on merge to `main` in any case.

TimDiekmann: Whops, that was a search/replace error (I used `'dashintel/hash/'` previously but replaced it with `GitHub.repository`.
```suggestion
    if: github.repository == 'hashintel/hash'
```

CiaranMn (approved):

PR #5670, opened 2024-11-19
H-3594: Add initial implementation for new migration strategy

TimDiekmann:

The new migration strategy needs a way to use Rust code. As an approach I chose to provide a simple `Migration` trait which has to be implemented for a migration file:
```rust
pub trait Migration {
    type Context;

    async fn up(self, context: &mut Self::Context) -> Result<(), Report<Self::Error>>;
    async fn down(self, context: &mut Self::Context) -> Result<(), Report<Self::Error>>;
}
```

The migration files are parsed, and the information about the files are extracted. All migrations are traversed based on the target migration number. A plan is created based on the already applied migrations and the target. The applied migrations are read from the database (which acts as the store for the state)

This PR does not add migrations, it just provides the interface to write them. The migration logic will be implemented in a later PR. This PR, however, adds a very small binary on top of the library for easier testing (see "How to test this?" below)

## 🔍 What does this change?

- It was quite a big challenge to use the `Migration` trait, because it's not `dyn` compatible. I experimented around with `#[async_trait]` but decided against it, mainly because it introduces boilerplate elsewhere. Instead, a very simple type-list is used inspired by `frunk`. With this, a simple `MIgrationList` trait is implemented which can be traversed back and forth.
- `MigrationList::traverse` calls back to the `MigrationRunner` which then will run the migration.
- The `MigrationPlanBuilder` allows to build a plan. This requires information such as:
	- The full migration list available
	- the state store (where the current migrations are stored)
	- The context which is used for the migrations (actually, a `ContextProvider` is passed which can return different contexts as different migrations may require different context. E.g. some only need the Postgres client, other may need a SpiceDB client)
	- (optional) a target migration number (
- The plan built from be builder can be `execute`d and will call the `traverse()` function on the `MigrationList`. Depending on the target and the latest migration it will use `up` or `down` migrations

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

All sub tickets of H-3580

## 🛡 What tests cover this?

None, yet. Tests will be added later

## ❓ How to test this?

Run 
```sh
cargo run --package hash-graph-migrations --bin cli --features cli -- run
```

and optionally provide a `--target` (append `--help` to the above command for more info)


libs/@local/graph/migrations/src/plan.rs
```diff
@@ -0,0 +1,258 @@
+use alloc::collections::BTreeMap;
+use core::ops::{Bound, RangeBounds as _};
+
+use error_stack::{Report, ResultExt as _, bail};
+
+use crate::{Migration, MigrationInfo, MigrationList, StateStore, list::MigrationError};
+pub trait MigrationRunner: Sized {
```
indietyp, line 7:
```suggestion

pub trait MigrationRunner {
```

indietyp (changes requested):
small change, otherwise LGTM 🎉

indietyp (approved):

PR #5674, opened 2024-11-20
H-3613: Implement proc-macro to read migration files

TimDiekmann:

Currently, we manually have to adjust the list of migration files. This is error-prone and repetitive. In addition. adding more functionality (such as providing the list of all migrations) becomes quite hard. Instead, a proc-macro is desired.

## 🔍 What does this change?

- Remove the `embedded` module
- Create proc-macro crate and implement it
	- Find all migration-modules in a directory
	- Calculates the SHA-256 for the whole module
	- Creates a function as the previous `embedded` module

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Tests are not written, yet. This will be done in a follow-up PR.


libs/@local/graph/migrations-macros/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+            let entry = entry?;
+            if !entry.file_type().is_dir()  {
+                buffer.push_str(&fs::read_to_string(entry.path())?);
+            }
+            Ok(buffer)
+        })
+}
+
+const BASE_REGEX: &str = r"^v(\d+)__(\w+)";
+static DIRECTORY_REGEX: OnceLock<Regex> = OnceLock::new();
```
indietyp, line 47: isn't a [`LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html) more appropriate here?

libs/@local/graph/migrations-macros/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+    let included_dir = input.location.to_string_lossy();
+    let mut migration_files = BTreeMap::new();
+    for migration_file in find_migration_files(&input.location) {
+        let migration_file = migration_file?;
+        match migration_files.entry(migration_file.number) {
+            Entry::Vacant(entry) => {
+                entry.insert(migration_file);
+            }
+            Entry::Occupied(entry) => {
+                return Err(ReadMigrationFileError::DuplicateMigration(
```
indietyp, line 257: have you thought about maybe using something like https://lib.rs/crates/proc-macro-error2 for better error locations?

libs/@local/graph/migrations/graph-migrations/v001__initial_setup.rs
indietyp, line 1: out of curiosity: how does this work with RA? is RA still able to provide diagnostics for these files? (I'd imagine that's the case)

libs/@local/graph/migrations-macros/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+#[derive(Debug, Clone)]
+struct MigrationFile {
+    module_path: PathBuf,
+    name: String,
+    number: u32,
+    size: usize,
+    digest: [u8; 32],
+}
+
+fn read_directory_content(path: impl AsRef<Path>) -> io::Result<String> {
```
indietyp, line 32: in this case, merely renaming a file wouldn't trigger a change in digest.

You also don't need to have a single buffer that you return. You can also just:

```rs
fn hash_directory_contents(path: impl AsRef<Path>, digest: &mut impl Digest) -> io::Result<String> {
	for path in ... {
		digest.update(entry.file_name());
		digest.update(&...);
	}
}

// in the caller you can then just call `digest.finalize`
```

I also don't know if `fs::read_to_string` is correct here, that means it'd require any dependent file to be valid utf-8, so binary blobs couldn't be used.

indietyp (changes requested):
exciting change, looks good to me, just some minor comments.

libs/@local/graph/migrations-macros/src/lib.rs
TimDiekmann: That's true, thanks!

libs/@local/graph/migrations-macros/src/lib.rs
TimDiekmann: I didn't know about that crate. But changing the errors now is a bigger piece of work. I'm going to use it when I add more features such as storing a list of all migration infos which is required to detect missing migrations.

libs/@local/graph/migrations/graph-migrations/v001__initial_setup.rs
TimDiekmann: With a small fix it's now working. RA has issue with nested modules where the inner module uses `#[path = "..."]` so I removed the nesting.

libs/@local/graph/migrations-macros/src/lib.rs
TimDiekmann: Did as drive-by in [`17d3483` (#5674)](https://github.com/hashintel/hash/pull/5674/commits/17d3483303950307d995d5daf13a09793f013b89)

libs/@local/graph/migrations-macros/src/lib.rs
TimDiekmann: Good point, I updated the digest generation.

PR #5687, opened 2024-11-21
H-3618: Add GitHub action step to authenticate with Google Cloud, for `chonky` tests

CiaranMn:

Adds a step prior to `chonky` tests which authenticates to Google Cloud Platform.

See [GitHub action](https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account) and [GCP](https://cloud.google.com/iam/docs/workload-identity-federation-with-deployment-pipelines#impersonation) docs.

The service account being impersonated is limited to Vertex AI access.

Also passes the `HUGGING_FACE_TOKEN` to tests.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. See if the `chonky` tests run successfully.


.github/workflows/test.yml
```diff
@@ -293,6 +316,15 @@ jobs:
           rm -rf $temp_dir
           echo "PDFIUM_DYNAMIC_LIB_PATH=$(pwd)/${{ matrix.directory }}/libs/" >> $GITHUB_ENV
 
+      # Sets GOOGLE_APPLICATION_CREDENTIALS in the environment, to be consumed by gcloud or client libraries
+      - name: Generate Google Cloud credential configuration
+        if: matrix.package == '@rust/chonky'
```
JesusFileto, line 321: From the documentation for `403` error codes, there seems to be priority in checking glcoud authentication before checking the valid URL, which implies that for the most part the rest of the workflow is functioning as intended. (link to documentation [here](https://cloud.google.com/vision-ai/docs/error-messages))

PR #5742, opened 2024-11-26
H-3684: Upgrade to Yarn 4

indietyp:

Currently the HASH workspace uses yarn 1 as the JavaScript package manager. yarn 1 is quite old (having been replaced by yarn 2 in 2020, nearly 5 years ago) and slow.

Changing to yarn 4 will hopefully bring us to something modern, while reducing the required effort to switch to a minimal.

yarn 4 can operate in 3 different modes
* pnp (the default)
* pnpm
* node_modules

PnP (also known as Plug'n'Play) uses zip archives that can be traverse, while the fastest in installation time, editors need to support it. Currently only VSCode is officially supported with built-in support in the WebStorm IDE (which can vary quite a bit from personal experience)

pnpm mirrors the style of [pnpm](https://pnpm.io/), where instead every package has a list of symlinks for their dependencies. This has the upside that installation is way faster, but also means that packages need to properly specify their dependencies. Some dependencies do not set their dependencies correctly, so [package extensions](https://yarnpkg.com/configuration/yarnrc#packageExtensions) need to be used. pnpm mode usually is the best compromise between speed, reproducibility and editor support.

The HASH monorepo currently experiences a problem, where installation during pnpm mode results in an alloc failure and can therefore not be used. This issue is known, see: https://github.com/yarnpkg/berry/issues/3580, https://github.com/yarnpkg/berry/issues/6550, https://github.com/yarnpkg/berry/issues/3972.

node_modules is the slowest of the 3 options and mirrors what yarn@1 and npm are doing, a node_modules folder where dependencies are shared.

This PR uses the node_modules approach, with the option of investigating the issue underlying the ZipFS failure in the future, hopefully being able to switch to pnpm mode.

This PR:

The changes in this PR:

The changes in this PR:


.github/actions/prune-repository/action.yml
```diff
@@ -8,17 +8,31 @@ inputs:
...
       run: |
-        while IFS= read -r line; do
-          if [[ -n "$line" ]]; then
-            turbo prune --scope="$line"
-          fi
-        done <<< "${{ inputs.scope }}"
+        # Convert newline-separated input to space-separated arguments
+        SCOPES=$(echo "${{ inputs.scope }}" | tr '\n' ' ')
+
+        # TODO: remove once deer is in labs
```
TimDiekmann, line 23:
```suggestion
        # TODO: remove once deer is in labs
        #   see https://linear.app/hash/issue/H-3692/move-deer-to-labs
```

TimDiekmann: It's easier to detect TODOs if they have an issue link 🙂

TimDiekmann (approved):
Amazing! 🎉

CiaranMn (approved):

PR #5755, opened 2024-11-28
H-3696: Create follow-up Yarn 4 documentation

indietyp:

#5742 has set up Yarn 4, but while doing so several opportunities to enhance the documentation were found.

These are:
* setup corepack
* author and remove patches

This PR adds these to the general developer documentation, as well as enables corepack for renovate as a drive-by.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash/README.md
+++ b/apps/hash/README.md
@@ -432,9 +432,9 @@ This will automatically update the patch file with your changes. Do not forget t
 
 ### Removing a patch
 
-Locate any `patch:` protocol entries in any workspace `package.json` and remove them. The entry will look somewhat similar to: `patch:@changesets/assemble-release-plan@npm%3A5.2.4#~/.yarn/patches/@changesets-assemble-release-plan-npm-5.2.4-2920e4dc4c.patch`, to remove the patch simply extract out the package (everything after the `patch:` and before `#`) and url-decode it and extract the version from it, so for the example it would be `npm:5.2.4`. You should **not** completely remove the line from the `package.json`.
+Locate any `patch:` protocol entries in any workspace `package.json` and remove them. The entry will look somewhat similar to: `patch:@changesets/assemble-release-plan@npm%3A5.2.4#~/.yarn/patches/@changesets-assemble-release-plan-npm-5.2.4-2920e4dc4c.patch`, to remove the patch simply extract out the package (everything after the `patch:` and before `#`) and url-decode it and extract the version from it, so for the example it would be `5.2.4`. You should **not** completely remove the line from the `package.json`.
 
-In case the patch has been applied in the resolutions field you should also check if the resolution is made redundant. This is the case if the left side is the same as the right, e.g. `"react@npm:18.2.0": "18.2.0"` is redundant, same as `"react@npm:18.2.0": "npm:18.2.0"`, or `"react@npm:18.2.0": "npm:react@18.2.0"`, but `react: "npm:react@18.2.0"` is **not** redundant.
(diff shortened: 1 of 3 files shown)


apps/hash/README.md
```diff
@@ -378,6 +386,68 @@ New packages which are to be built as JavaScript, whether as an app or dependenc
...
+yarn patch-commit -s /private/var/folders/lk/j93xz9pd7nqgd5_2wlyxmbh00000gp/T/xfs-d772c076/user
+```
+
+This will automatically update the patch file with your changes. Do not forget to run `yarn install` for the patch to be installed and applied to the lockfile.
+
+### Removing a patch
+
+Locate any `patch:` protocol entries in any workspace `package.json` and remove them. The entry will look somewhat similar to: `patch:@changesets/assemble-release-plan@npm%3A5.2.4#~/.yarn/patches/@changesets-assemble-release-plan-npm-5.2.4-2920e4dc4c.patch`, to remove the patch simply extract out the package (everything after the `patch:` and before `#`) and url-decode it and extract the version from it, so for the example it would be `npm:5.2.4`. You should **not** completely remove the line from the `package.json`.
+
+In case the patch has been applied in the resolutions field you should also check if the resolution is made redundant. This is the case if the left side is the same as the right, e.g. `"react@npm:18.2.0": "18.2.0"` is redundant, same as `"react@npm:18.2.0": "npm:18.2.0"`, or `"react@npm:18.2.0": "npm:react@18.2.0"`, but `react: "npm:react@18.2.0"` is **not** redundant.
```
CiaranMn, line 437:
```suggestion
In case the patch has been applied in the resolutions field you should also check if the resolution is made redundant. This is the case if the left side is the same as the right, e.g. `"react@npm:18.2.0": "18.2.0"` is redundant, same as `"react@npm:18.2.0": "npm:18.2.0"`, or `"react@npm:18.2.0": "npm:react@18.2.0"`, but `"react": "npm:react@18.2.0"` is **not** redundant.
```

apps/hash/README.md
```diff
@@ -378,6 +386,68 @@ New packages which are to be built as JavaScript, whether as an app or dependenc
...
+
+```sh
+yarn patch-commit -s /private/var/folders/lk/j93xz9pd7nqgd5_2wlyxmbh00000gp/T/xfs-d772c076/user
+```
+
+This will automatically update the patch file with your changes. Do not forget to run `yarn install` for the patch to be installed and applied to the lockfile.
+
+### Removing a patch
+
+Locate any `patch:` protocol entries in any workspace `package.json` and remove them. The entry will look somewhat similar to: `patch:@changesets/assemble-release-plan@npm%3A5.2.4#~/.yarn/patches/@changesets-assemble-release-plan-npm-5.2.4-2920e4dc4c.patch`, to remove the patch simply extract out the package (everything after the `patch:` and before `#`) and url-decode it and extract the version from it, so for the example it would be `npm:5.2.4`. You should **not** completely remove the line from the `package.json`.
```
CiaranMn, line 435: Do we need the `npm:`? Isn't it the case that we'd only need to specify it if we wanted the package to come from a different registry than the one we use by default, which is `npm`?

I know `yarn` is adding it automatically for patches but when we want to revert to a 'normal' version or resolution, we don't need to specify `npm` and it'll look a bit weird to have some deps specifying it and some not (when not managed by `yarn`).

```suggestion
Locate any `patch:` protocol entries in any workspace `package.json` and remove them. The entry will look somewhat similar to: `patch:@changesets/assemble-release-plan@npm%3A5.2.4#~/.yarn/patches/@changesets-assemble-release-plan-npm-5.2.4-2920e4dc4c.patch`, to remove the patch simply extract out the package (everything after the `patch:` and before `#`) and url-decode it and extract the version from it, so for the example it would be `5.2.4`. You should **not** completely remove the line from the `package.json`.
```

indietyp: not strictly needed no, might be best to remove it, you're right

CiaranMn (approved):

PR #5761, opened 2024-12-01
H-3707, H-3708: Implement libp2p frontend

indietyp:

This implements the libp2p frontend, that is used to connect and interact with the underlying libp2p implementation.

This was initially split up into multiple tasks, but they have been merged into one, as keeping them separate was a bit more of a challenge than it should have been.

Care has been taken that the effect code and non effect code (like logging) integrate properly.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/client/typescript/src/net/internal/networkLogger.ts
```diff
@@ -0,0 +1,247 @@
...
+} from "effect";
+import type { ReadonlyRecord } from "effect/Record";
+import { CID } from "multiformats";
+import { base32 } from "multiformats/bases/base32";
+import { base58btc } from "multiformats/bases/base58";
+import { base64 } from "multiformats/bases/base64";
+
+interface Instruction {
+  type: string; // single alpha character
+}
```
CiaranMn, line 25: Not sure if either of these are important if this interface is never going to be exposed outside the file, but:

1. Could move the comment so that the fact that it's a single alpha character is available in editor hints at the point of use
```suggestion
interface Instruction {
  /** single alpha character */
  type: string;
}
```

2. The set of possible values is small enough to type directly, but might be more annoying to use than is worth it
```suggestion
type AlphaCharacter = "a" | "b" | "c" | "d" | "e" | "f"; // etc

interface Instruction {
  /** single alpha character */
  type: AlphaCharacter | Uppercase<AlphaCharacter>;
}
```

CiaranMn (approved):
Cool :) 

One comment I think probably doesn't make any practical difference, just writing it because I thought it, can be ignored

libs/@local/harpc/client/typescript/src/net/internal/networkLogger.ts
indietyp: I thought about doing (2) but then stopped as writing out 26 characters seemed a bit repetitive for an internal only thing. Although probably worth the typing affordances it gives. Considering that this is partially exposed to the user (if they want to extend it / modify it) I added more strict typing to ensure that everything is well-formed, see: [09344bd](https://github.com/hashintel/hash/pull/5761/commits/09344bd4140777eb9f448eb43e8991f27a27c56c)

CiaranMn (approved):

PR #5769, opened 2024-12-02
H-3689: Replace `check-dependency-version-consistency` with yarn constraints

indietyp:

Replaces `check-dependency-version-consistency` with yarn constraints, a new feature in Yarn 4, this allows us to unify some of the linting packages we previously relied on.

Additionally the `enforceDevDependenciesAreProperlyDeclared` constraint has been rewritten to be more robust.

> constraints are now also auto-applied on commit

As a drive-by `cross-env` has been removed as it is no longer needed in Yarn 4 due to the introduction of a POSIX compliant OS independent shell implementation.

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (approved):
Much simpler, thank you

PR #5784, opened 2024-12-03
H-3733: Experiment with lefthook as a husky/lint-staged replacement

indietyp:

husky and lint-staged are both dependency heavy and there have been challenges in trying to make external tools work (tools that cannot be installed with yarn).

The goal of this is to try to see if lefthook can be a minimally invasive alternative that allows for sqlfluff (and other packages) to be used.

> Personally I've also seen instances where husky/lint-staged just doesn't work because because e.g. it picked up the wrong node version from my system (I use mise to configure those) which has been very annoying in the past.

**husky doesn't clean up correctly**, while it works fine without doing it, keeping your hook location to the one of husky will mean you have a permanent stage file that shouldn't be committed, unset the git change that husky applied via: `git config --unset core.hooksPath`

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.config/husky/pre-commit
+++ b/.config/husky/pre-commit
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-yarn lint-staged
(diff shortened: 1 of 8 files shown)


TimDiekmann (approved):
Works like a charm, many thanks!!

PR #5820, opened 2024-12-06
H-3371, H-3722: Overhaul return behavior for entity validation

TimDiekmann:

To allow proper handling of validation errors, this PR completely overhauls how validation errors are returned from the graph. Most notably is that the validation function itself won't error anymore but returned a well-defined structure.

## 🔍 What does this change?

Add a validation structure which looks similar to this:
```ts
export interface EntityValidationReport {
  properties?: PropertyValidationReport;
  link?: LinkValidationReport;
  metadata?: MetadataValidationReport;
}

export interface LinkValidationReport {
  leftEntity?: LinkedEntityError | null;
  rightEntity?: LinkedEntityError | null;
  linkData?: LinkDataStateError | null;
  linkType?: LinkError | null;
  targetType?: LinkTargetError | null;
}

export interface MetadataValidationReport {
  entityTypes?: EntityTypesError | null;
  properties?: any | null;
}

export type LinkedEntityError =
  | { type: "entityRetrieval"; error: Report }
  | { type: "entityTypeRetrieval"; error: Report }
  | { type: "resolveClosedEntityType"; error: Report };

export type LinkError = {
  type: "unexpectedEntityType";
  data: {
    actual: Array<string>;
    expected: Array<string>;
  };
};

export type LinkTargetError = {
  type: "unexpectedEntityType";
  data: {
    actual: Array<string>;
    expected: Array<string>;
  };
};

export type LinkDataStateError =
  | { type: "missing"; error: Report }
  | { type: "unexpected"; error: Report };

export type EntityTypesError =
  | { type: "empty"; error: Report }
  | { type: "entityTypeRetrieval"; error: Report }
  | { type: "resolveClosedEntityType"; error: Report };

export interface PropertyValidationReport {
  error?: Report | null;
}
```

The `properties` field in the top level is not adjusted yet. For now it just contains a `Report` (basically all the errors which could happen in property validation). This will be adjusted in a follow-up PR as the structure of the preprocessor needs to change quite a bit for that.
`metadata.properties` should include metadata validation for properties which currently does nothing. The OpenAPI generator uses any for this (please don’t ask me why), it should always be empty but was kept to follow the code paths in Rust.

The Graph API endpoint itself are changed so they return `{ [key: string]: EntityValidationReport }` while `key` is the position of the entity in the parameter (0-based index, but JSON does not support numbers as keys). Any parameter with a failed validation will be reported.
This means:
- `createEntities` and `updateEntity` will return the result as previously, but if a validation error happens, the above structure is passed alongside the reports in the response
- `validateEntities` will always return the map. An empty map means no validation errors.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

Overhaul the property validation reporting as well


.github/workflows/lint.yml
```diff
@@ -11,6 +11,7 @@ env:
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
   TURBO_TEAM: hashintel
   TURBO_REMOTE_ONLY: true
+  TURBO_FORCE: true
```
TimDiekmann, line 14:
```suggestion
```

CiaranMn (approved):

PR #5827, opened 2024-12-06
H-2976: Improve validation for ontology types

TimDiekmann:

Some ontology types don't make sense, e.g. having a minimum higher than the maximum or an `allOf: [number, string]`. An error should be emitted in this case.

Entity/property types should be validated: e.g. required only specifies actual properties.

The references don't need to be validated up-front as they will be checked when inserted into the graph anyway.

## 🔍 What does this change?

By now, most of the schemas are either validated when created or when creating the closed variant of it.
This PR adds a validation for array property types to ensure that `minItems` must be less than or equal to `maxItems`.
Further, it allows `required` to point to a property which is specified in a parent entity type (by removing this constraint on the `EntityType` but keep it in `ClosedEntityType`).

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/schema/array/validation.rs
+++ b/libs/@blockprotocol/type-system/rust/src/schema/array/validation.rs
@@ -5,8 +5,8 @@ use crate::{Valid, Validator, schema::PropertyValueArray};
 #[derive(Debug, Error)]
 pub enum ArraySchemaValidationError {
     #[error(
-        "Unsatisfiable item number constraints, expected minimum of {min} and maximum of {max} \
-         items"
+        "Unsatisfiable item number constraints, expected minimum amount of items ({min}) to be \
+         less than or equal to the maximum amount of items ({max})"
     )]
     IncompatibleItemNumberConstraints { min: usize, max: usize },
 }
@@ -21,7 +21,11 @@ impl<T: Sync> Validator<PropertyValueArray<T>> for ArraySchemaValidator {
         value: &'v PropertyValueArray<T>,
     ) -> Result<&'v Valid<PropertyValueArray<T>>, Self::Error> {
         if let Some((min, max)) = value.min_items.zip(value.max_items) {
-            return Err(ArraySchemaValidationError::IncompatibleItemNumberConstraints { min, max });
+            if min > max {
+                return Err(
+                    ArraySchemaValidationError::IncompatibleItemNumberConstraints { min, max },
+                );
+            }
         }
 
         Ok(Valid::new_ref_unchecked(value))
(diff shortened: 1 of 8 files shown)


libs/@blockprotocol/type-system/rust/src/schema/array/validation.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct ArraySchemaValidator;
+
+impl<T: Sync> Validator<PropertyValueArray<T>> for ArraySchemaValidator {
+    type Error = ArraySchemaValidationError;
+
+    fn validate_ref<'v>(
+        &self,
+        value: &'v PropertyValueArray<T>,
+    ) -> Result<&'v Valid<PropertyValueArray<T>>, Self::Error> {
+        if let Some((min, max)) = value.min_items.zip(value.max_items) {
```
CiaranMn, line 23: I'm not understanding how this achieves the description in the PR:

> This PR adds a validation for array property types to ensure that minItems must be less than or equal to maxItems.

Does it happen elsewhere / am I missing something? 

And do we already do this for min/max in an entity type's links schema?

TimDiekmann: This validates the array property ***types***. If a property type expects an array with a `minItems` and `maxItems` value it would not be possible to instantiate a property for it if `minItems > `maxItems`.

e.g. this property type would now fail:
```json
{
  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/property-type",
  "kind": "propertyType",
  "$id": "https://blockprotocol.org/@alice/types/property-type/contrived-property/v/1",
  "title": "Contrived Property",
  "description": "A contrived property for testing purposes.",
  "oneOf": [
    {
      "type": "array",
      "items": {
        "oneOf": [
          {
            "$ref": "https://blockprotocol.org/@blockprotocol/types/data-type/number/v/1"
          }
        ]
      },
      "minItems": 4
      "maxItems": 2
    }
  ]
}
```

TimDiekmann: > And do we already do this for min/max in an entity type's links schema?

Good flag, no, we don't

TimDiekmann: I added the proper check, improved the error message, and added link count validation as well: [`7565585` (#5827)](https://github.com/hashintel/hash/pull/5827/commits/75655855d2ff37f1b8c9de732fc58eb5a17601db)

CiaranMn: What about if `max < min`?

TimDiekmann: `max < min` is the same as `min > max` I think 🙂

CiaranMn: Oh yeah :D

CiaranMn (approved):

PR #5847, opened 2024-12-10
H-2457: Improved ESLint configuration

indietyp:

ESLint 9 introduces flat configs which breaks the existing configuration format, in light of this we decided to move to a more encompassing ruleset in the longterm.

Previous attempts have been made to apply a stricter ruleset, but it was found that due to the sheer size of some projects it just isn't feasible to transition every project in one go, instead the approach is as follows:

1. We transition to ESLint 9 and use flat configs
2. During this transition to ESLint 9 we implement a new (stricter) ruleset. The ruleset won't be applied directly, but instead gradually over all projects. The existing legacy ruleset is used in its stead for now.
3. Key existing libraries/apps (HaRPC client, `hash-isomorphic-utils`, and others) could be moved over to newer configs.
4. New projects use the new ruleset instead of the deprecated one. 

For stricter rulesets there are quite some prebuilt ones, these include:

* `xo`
* `sheriff`
* `@antfu/eslint-config`

The main problem with `xo` is that it is exposed through an alternative binary, making it unsuitable to use in, e.g., WebStorm. It also, by default, lints stylistic choices covered by prettier.

`@antfu/eslint-config` is just eslint, but like `xo` also pretty heavily features stylistic rules.

`sheriff` instead fully embraces ESLint flat configs, is highly configurable, and has an excellent webpage detailing all the configurations used.

To see a comprehensive comparison of different configs, see: [https://www.eslint-config-sheriff.dev/docs/prior-art](https://www.eslint-config-sheriff.dev/docs/prior-art)

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/client/typescript/src/codec/Decoder.ts
```diff
@@ -42,6 +42,7 @@ export interface Decoder<E = DecodingError, R = never>
   };
 }
 
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
CiaranMn, line 45: Maybe not worth doing but could set a rule in this package:

```
{
    rules: {
      "@typescript-eslint/no-empty-object-type": ["error", { allowInterfaces: "with-single-extends" }]
    },
}
```

CiaranMn (approved):
🚀

libs/@local/harpc/client/typescript/src/codec/Decoder.ts
indietyp: didn't know it had that option, good shout!

added in [ee2cac6](https://github.com/hashintel/hash/pull/5847/commits/ee2cac6e84c5058c9f8a7f14d2f14827cc914b14)

CiaranMn (dismissed):

CiaranMn (approved):

PR #5850, opened 2024-12-10
H-3755: Improve validation structure for entity properties

TimDiekmann:

H-3722 overhauled the validation results but kept the Report of the entity preprocessor. Those results also have to be adjusted to be more structured.
This PR adjusts the validation for properties so they return a structured validation report as well.

## 🔍 What does this change?

- Change the pre-processor to return a validation structure instead of plain `Report`s
- Redefine the types in TS so they can be properly used.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/graph/types/typescript/src/validation.ts
```diff
@@ -0,0 +1,235 @@
...
+      // The property was found at the entity but is not expected in the schema
+      type: "unexpected";
+    }
+  | {
+      // It was not possible for the graph to read the property type
+      type: "retrieval";
+      error: Report;
+    }
+  | {
+      // The property was found at the entity but the type is not the expected type from Athe schema
```
CiaranMn, line 51:
```suggestion
      // The property was found at the entity but the property type is not the expected type from the schema
```

libs/@local/graph/types/typescript/src/validation.ts
```diff
@@ -0,0 +1,235 @@
...
+
+export type PropertyValidationReport = Subtype<
+  PropertyValidationReportGraphApi,
+  | {
+      // The property value validation failed
+      type: "value";
+      /*
+       * Validation for each constraint in the `oneOf` field. The validation is assumed to pass if exactly one of the
+       * constraints passes. In this case, this field will be omitted. Whenever this field is present it can be assumed
+       * that the validation failed.
```
CiaranMn, line 82: Exactly one, or at least one? We want the latter (and therefore want to rename `oneOf` to `anyOf` at some point)

libs/@local/graph/types/typescript/src/validation.ts
```diff
@@ -0,0 +1,235 @@
...
+      type: "valueValidation";
+      data: ValueValidationReport;
+    }
+>;
+
+export type ValueValidationReport = Omit<
+  ValueValidationReportGraphApi,
+  "actual" | "desired" | "abstract" | "incompatible"
+> & {
+  // The actual value has a validation error
```
CiaranMn, line 139:
```suggestion
  // The value could not be validated against the provided data type
```

libs/@local/graph/types/typescript/src/validation.ts
```diff
@@ -0,0 +1,235 @@
...
+  actual?: ValueValidationError;
+  // The value could not be validated against the data type specified in the schema
+  desired?: ValueValidationError;
```
CiaranMn, line 142: I slightly prefer `provided` over `actual` because I think it makes it clear that it's what the caller has provided – rather than 'actual' which could mean 'actually provided' or 'actually in the schema'.

Similar for 'desired' – is it desired by the caller or the schema? Although feel less strongly about this one.

```suggestion
  provided?: ValueValidationError;
  // The value could not be validated against the data type specified in the schema
  target?: ValueValidationError;
```

libs/@local/graph/types/typescript/src/validation.ts
```diff
@@ -0,0 +1,235 @@
...
+  // The actual schema is abstract
+  abstract?: VersionedUrl;
+  // The actual schema is incompatible with the desired schema, i.e. the actual schema is not a subtype of the desired schema
```
CiaranMn, line 145:
```suggestion
  // The provided DataType is abstract
  abstract?: VersionedUrl;
  // The provided DataType is incompatible with the desired DataType, i.e. the actual DataType is not a subtype of the target DataType
```

CiaranMn (commented):
Thanks @TimDiekmann, just a couple of comments. A suggestion on naming we can discuss if needed

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: Currently it's exactly one. We can change this as part of this PR if you like.

CiaranMn: Let's change it now so it doesn't trip us up later

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: [`8fa0baa` (#5850)](https://github.com/hashintel/hash/pull/5850/commits/8fa0baa5fa34141623ff8e99f5733f3ad388bf91)

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: [`8fa0baa` (#5850)](https://github.com/hashintel/hash/pull/5850/commits/8fa0baa5fa34141623ff8e99f5733f3ad388bf91)

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: [`8fa0baa` (#5850)](https://github.com/hashintel/hash/pull/5850/commits/8fa0baa5fa34141623ff8e99f5733f3ad388bf91)

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: [`8fa0baa` (#5850)](https://github.com/hashintel/hash/pull/5850/commits/8fa0baa5fa34141623ff8e99f5733f3ad388bf91)

libs/@local/graph/types/typescript/src/validation.ts
TimDiekmann: https://github.com/hashintel/hash/pull/5850/files/8fa0baa5fa34141623ff8e99f5733f3ad388bf91..59f72104d3cc1f879c009218225a1cdd28adc8f6

CiaranMn (approved):

CiaranMn (approved):

TimDiekmann:
Waiting for https://github.com/hashintel/hash/pull/5857 to avoid TF configurations to be overwritten in prod (in touches immutable configs)

PR #5856, opened 2024-12-11
H-3710: Support DNS for HaRPC client and align environment variables

TimDiekmann:

To allow HaRPC in AWS we need to allow DNS lookups in the client. This is needed to connect the different services as the IP of the graph could change with a redeployment and the DNS is stable. The situation is used to align the environment variables for the RPC and HTTP endpoints for the graph.

## 🔍 What does this change?

- Use `dns` for the address for the HaRPC client
- Enable HaRPC in dev-mode locally
- Set an appropriated logging level
- Align environment variables for the Graph
- Prepare HaRPC on AWS (but keep it disabled)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

I did a lot of manual testing but I need the graph to be deployed in production to test this. As we use x86 in AWS I currently don't have a way to build those images locally and upload them without changing the AWS config.
There is a small chance that some environment variable was missed, but without a staging environment we cannot avoid this risk.


.env
```diff
@@ -70,16 +70,18 @@ HASH_TEMPORAL_VISIBILITY_PG_DATABASE=temporal_visibility
...
 HASH_GRAPH_TYPE_FETCHER_PORT=4455
 
 HASH_GRAPH_REALTIME_PG_USER=realtime
 HASH_GRAPH_REALTIME_PG_PASSWORD=realtime
 
-HASH_GRAPH_API_HOST=127.0.0.1
-HASH_GRAPH_API_PORT=4000
+HASH_GRAPH_HTTP_HOST=127.0.0.1
+HASH_GRAPH_HTTP_PORT=4000
+HASH_GRAPH_RPC_HOST=127.0.0.1
```
indietyp, line 83: are we sure we want to use the same variable for both client and host? considering that the client might have different host to the server (I expect the server to always bind to an IP, but the client to connect over DNS)

apps/hash-api/src/index.ts
```diff
@@ -560,10 +560,10 @@ const main = async () => {
       }).pipe(
         Effect.provide(
           RpcClient.connectLayer(
-            Transport.multiaddr(`/ip4/${rpcHost}/tcp/${rpcPort}`),
+            Transport.multiaddr(`/dns/${rpcHost}/tcp/${rpcPort}`),
```
indietyp, line 563: thank you for going down the dns rabbit hole 😅 I kinda missed enabling that, but glad it was straightforward - out of curiosity - in typescript does a `/dns/127.0.0.1/tcp/4002` still work?

indietyp (approved):
Thank you! 🎉  Minor remaining non-blocking questions

.env
TimDiekmann: This is how we handle other environment variables currently as well. In production you typically specify the host per instance anyway, so you can pass different environment variable to different services. E.g. in Docker you'd use the container name of the server for the client and set the server manually to `127.0.0.1`. In AWS you'd sets the host to `0.0.0.0` and (in our case) use the DNS name of the server for the client.

apps/hash-api/src/index.ts
TimDiekmann: Yes, I did not expected this actually but it just works. I was about to check if `rpcHost` is an IP or something else, but this is straight forward.

TimDiekmann: Also, to be 100% sure I just tested it again locally

PR #5864, opened 2024-12-11
H-3686, H-3700, H-3775: Introduce new data type selector

CiaranMn:

This PR updates the component for selecting a data type to account for data type inheritance and `abstract` types (which are semantic groupings but not assignable to values – e.g .`Length`).

It uses it in:
1. The type editor, when choosing the expected value(s) of a property type
2. In the entity editor, when choosing what data type to assign to a value (where it is ambiguous, either because there are multiple expected types defined, or because a single expected type has children)

See demo.

In passing the PR also:
- H-3686: Fixes a performance issue with dynamically rendered components using MUI's `TextField`
- H-3775: Fixes locking scroll while the editor is open in the entity editor (and other `Grid`-based components)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Deferred to follow-up (H-3779):
- show parents on cards in data type flat view (when search text is entered)
- keyboard navigation

## 🛡 What tests cover this?

- Basic editing test on entity editor and type editor, nothing specific on the new picker.

## ❓ How to test this?

1. Not possible on the preview deployment due to API changes


vilkinsons (dismissed):
Thanks! Will have a play in prod once fully deployed.

vilkinsons (approved):

PR #5876, opened 2024-12-12
H-3788: Do not force FQDN on DNS lookup

indietyp:

We're currently forcing an FQDN DNS lookup, this was previously needed, but prohibits the use of search domains.

This is usually wanted, as otherwise, `machine` might get coerced into `machine.aws.com` on the local system, but it seems like that the AWS system relies on search domains.

## 🔗 Related links

* [Fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name)
* [Search domain](https://en.wikipedia.org/wiki/Search_domain)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/harpc/client/typescript/src/net/internal/transport.ts
+++ b/libs/@local/harpc/client/typescript/src/net/internal/transport.ts
@@ -98,38 +98,34 @@ const resolveDnsMultiaddrSegment = (code: number, value?: string) =>
       return [[code, value] as const];
     }
 
-    let fqdn = value;
+    let hostname = value;
 
     let family: 0 | 4 | 6;
     if (code === DNS4_PROTOCOL.code) {
       family = 4;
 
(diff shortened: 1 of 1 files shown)


TimDiekmann (dismissed):

TimDiekmann:
Did an in-depth review. Took a while but I carefully visited all lines! 🚀

libs/@local/harpc/client/typescript/src/net/internal/transport.ts
```diff
@@ -98,38 +98,34 @@ const resolveDnsMultiaddrSegment = (code: number, value?: string) =>
       return [[code, value] as const];
     }
 
-    let fqdn = value;
+    let hostname = value;
```
TimDiekmann, line 101:
```suggestion
    const hostname = value;
```

TimDiekmann (approved):

PR #5894, opened 2024-12-13
H-3792: Switch to `dns.resolve` in HaRPC

indietyp:

It seems like that `dns.resolve` wasn't the issue, instead it seems that the issue was that we were forcing FQDNs. This is currently unproven and needs to be deployed on AWS for verification. 

This also moves the dns code into it's own module to make it easier to understand (and `transport.ts` was getting quite cramped)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/harpc/client/typescript/src/net/internal/dns.ts
+++ b/libs/@local/harpc/client/typescript/src/net/internal/dns.ts
@@ -1,0 +1,136 @@
+import dns from "node:dns/promises";
+import { isIPv4, isIPv6 } from "node:net";
+
+import { Array, Cause, Data, Duration, Effect, Function } from "effect";
+import type { NonEmptyReadonlyArray } from "effect/Array";
+
+/** @internal */
+export class DnsError extends Data.TaggedError("DnsError")<{
+  cause: unknown;
+}> {
+  get message() {
+    return "Underlying DNS resolver experienced an error";
+  }
+}
+
+/** @internal */
+export type RecordType = "A" | "AAAA";
+
+/** @internal */
+export interface Ipv4AddressRecord {
+  type: "A";
+
+  address: string;
+  timeToLive: Duration.Duration;
+}
+
+/** @internal */
+export interface Ipv6AddressRecord {
+  type: "AAAA";
+
+  address: string;
+  timeToLive: Duration.Duration;
+}
+
+/** @internal */
+export type DnsRecord = Ipv4AddressRecord | Ipv6AddressRecord;
+
+const resolveA = (hostname: string) =>
+  Effect.tryPromise({
+    try: () => dns.resolve4(hostname, { ttl: true }),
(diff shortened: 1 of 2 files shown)


libs/@local/harpc/client/typescript/src/net/internal/dns.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (query.records.includes("A")) {
+      resolvers.push(resolveA(hostname));
+
+      if (isIPv4(hostname)) {
+        return [
+          {
+            type: "A",
+
+            address: hostname,
+            timeToLive: Duration.infinity,
```
TimDiekmann, line 91: This is simple but genius 😃

libs/@local/harpc/client/typescript/src/net/internal/transport.ts
```diff
@@ -100,46 +89,30 @@ const resolveDnsMultiaddrSegment = (code: number, value?: string) =>
...
-        return [[IPV6_PROTOCOL.code, hostname] as const];
-      }
+    if (code === DNS_PROTOCOL.code || code === DNS6_PROTOCOL.code) {
+      types.push("AAAA");
     }
 
-    const responses = yield* Effect.tryPromise({
-      try: () => dns.lookup(hostname, { all: true, family }),
-      catch: (cause) => new DnsError({ cause }),
+    const records = yield* Dns.resolve(hostname, {
```
TimDiekmann, line 102: I guess we're going to add `dns.resolveSoa` or use `dns.resolveAny` in a follow up?

TimDiekmann (approved):

libs/@local/harpc/client/typescript/src/net/internal/transport.ts
indietyp: if any it's going to be `dns.resolveAny`, I didn't use `dns.resolve` because you can't get the TTL that way, you can with `resolve4`/`resolve6`. 

(I actually forgot about `dns.resolveAny`, I will add it if this works! 🤞)

TimDiekmann: Yup, let's see if it's working in the first place. If `resolveAny` does not work we can explore `resolve4` or `resolve6` to get the TTL. (Or always try `resolve6` first and if that fails use `resolve4` as fallback)

indietyp: We already do that 😄 

https://github.com/hashintel/hash/blob/9576fa6e5e76b241e163cf088456986cc47844be/libs/%40local/harpc/client/typescript/src/net/internal/dns.ts#L38-L70

TimDiekmann: We should test at some point what is faster, two dns calls as we have currently, or a single `resolveAny` which potentially unused data.

PR #5955, opened 2024-12-18
H-3777: Replace `markdownlint-cli` with `markdownlint-cli2`

indietyp:

`markdownlint-cli2` is from the creator of `markdownlint` itself and runs in parallel, has a lot more options and most importantly does not need the use of a separate `markdownlintignore` file.

This PR:

The changes in this PR:

The changes in this PR:


CiaranMn (dismissed):
Nice to lose an `ignore` file, thank you!

CiaranMn (approved):

PR #6031, opened 2024-12-30
H-3762, H-3763, H-3757: Experiment with biome

indietyp:

Biome is a suite of TypeScript tools that are written in Rust (originally known as Rome), it allows us to lint, format, and analyse a whole suite of problems.

The hope is that biome enables developers to have a faster workflow with minimal intrusions.

Some stats: `fix:prettier` previously took 20 seconds(!), `fix:biome` now takes 293ms, that's a **70x** speed increase.

We also no longer need a `.prettierignore` file, meaning we're just down to `.dockerignore` and `.gitignore`.

## 🔗 Related links

- [Biome](https://biomejs.dev/)

## 🔍 What does this change?

- initialize biome
- replace prettier with biome
- replace prettier package.json plugin with `sort-package-json` package
- adjust lefthook (drive-by)
- adjust CI

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
CI is failing because `yarn install` modifies the `package.json` files and formats them, even though they're formatted by `prettier`.

package.json
```diff
@@ -40,17 +40,19 @@
     "fix:eslint": "turbo --continue fix:eslint --",
     "fix:taplo": "taplo fmt",
     "fix:markdownlint": "markdownlint-cli2 --fix",
-    "fix:prettier": "prettier --write  --ignore-unknown .",
+    "fix:biome": "biome format --write",
```
vilkinsons, line 43: I wonder if it's worth us keeping `yarn fix:prettier` around as an alias for `yarn fix:biome`, for some period of time, as a convenience function (in case of muscle memory) while we still use the command in our other repos (both `blockprotocol` and internal).

yarn.config.cjs
```diff
@@ -8,6 +8,7 @@ const { defineConfig } = require(`@yarnpkg/types`);
 
 const enforcedDevDependencies = {
   prettier: { commands: ["prettier"], ident: "prettier" },
```
vilkinsons, line 10: Do we want to keep this and other retained references to `prettier` in this `yarn.config.cjs` file, or are they still required?

indietyp: these are still here for blocks, I tried to keep those as untouched as possible, we can remove it though!

package.json
indietyp: done in [08c9169](https://github.com/hashintel/hash/pull/6031/commits/08c9169eea68c8af7087fb1d3f842a420b87d952)

vilkinsons: Thanks!

yarn.config.cjs
vilkinsons: Okay, this makes sense - let's keep as is. Thank you!

package.json
TimDiekmann: Alternative: `fix:format` and `lint:format` - it does not really matter how it's formatted as long as it's done somehow.

CiaranMn: Maybe we can add running `fix:package-json` to this alias?

i.e. so it runs fixes for both `biome` and `package-json`, maintaining the existing behavior of `fix:prettier` – otherwise you can push something to CI and get a failure on package json formatting even having run `fix:prettier`. 

I'd be fine with renaming it to `fix:format` (and `lint:format`).

CiaranMn (commented):
One comment – otherwise looks good, great speed up!

package.json
indietyp: I've left out directly combining `biome` and `package-json`, as the `package-json` formatter is a lot slower, instead I added it to the pre-commit hooks, in the hope that that'll be sufficient instead, but I can also add a `fix:format` command that executes both.

CiaranMn: Fair enough, let's see how it works on pre-commit. If it becomes annoying in pre-commit (I'm not sure how slow it is) then we can move it into a `fix:format` command.

CiaranMn (approved):

PR #6102, opened 2025-01-09
H-3162: Look for matching entities in database when inserting entities from Flow

CiaranMn:

When a Flow infers entities from the web/documents, we want to check if there is a matching entity in the database to create an update to that entity rather than a duplicate entity.

This PR introduces a process to find semantically similar entities – or, for link entities, links of the same type with the same source/target – and then asks an LLM to:
1. judge whether any of the potential matches are 'the same thing'
2. provide new combined versions of properties which are suitable for merging (e.g. merge descriptions from the old and new value)
3. Merges the sources for a given property if they (a) are identical across, or (b) are a merged version of, the old and new values

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- H-3883: This does not check that the addition of new links will violate max link constraints set on an entity type's schema
- H-3900: Source merging is done for primitive values only – there is no checking if values in property objects or arrays are identical 

## 🛡 What tests cover this?

- Not a CI or functional test, but there's a manually-triggered optimization test with a few test cases to help optimize AI performance.

## ❓ How to test this?

1. Not possible via preview deployment.


.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
     "javascriptreact",
     "typescript",
     "typescriptreact"
-  ]
+  ],
+  "vitest.disableWorkspaceWarning": true
```
vilkinsons, line 8: Would be great to address this as some point. Previously flagged [here](https://hashintel.slack.com/archives/C03TWBSRT8B/p1718695843240629) (_internal link_).

vilkinsons (approved):

apps/hash-api/src/graph/ensure-system-graph-is-initialized/migrate-ontology-types/migrations/007-create-api-usage-tracking.migration.ts
```diff
@@ -358,24 +340,6 @@ const migrate: MigrationFunction = async ({
       inputUnitCost: 0.00000025,
       outputUnitCost: 0.00000125,
     },
```
vilkinsons, line 342: I don't think this pricing info is correct @CiaranMn per https://www.anthropic.com/pricing#anthropic-api

CiaranMn: Haiku is 0.25 per m input tokens, 0.00000025 *1m is 0.25 so it looks right to me

PR #6106, opened 2025-01-10
H-3879: Add protocol dropdown to early access form URL input

CiaranMn:

We want to move property types which have URLs as their values to using the `URL` data type. This means values for them must pass URL validation, e.g. include a scheme at the start (`https://`) rather than just be e.g. `example.com`

But we also want users to be able to just type `example.com` into some inputs, e.g. in the early access form.

This PR adds a protocol dropdown to the early access form's website URL input so that we can pass a full URL to the database but users don't have to type it out properly themselves.

It also allows for having a path in the URL (e.g. `example.com/my-page`) which the early access form website URL input was previously not allowing.

This PR:

The changes in this PR:

The changes in this PR:

## 📹 Demo

https://github.com/user-attachments/assets/12c790d7-dfff-4606-b046-fe176f1570b8

### DIFF
--- a/apps/hash-frontend/src/pages/index.page/waitlisted/early-access-modal.tsx
+++ b/apps/hash-frontend/src/pages/index.page/waitlisted/early-access-modal.tsx
@@ -9,6 +9,7 @@ import { Button } from "../../../shared/ui/button";
 import { MenuItem } from "../../../shared/ui/menu-item";
 import { Modal } from "../../../shared/ui/modal";
 import { useAuthenticatedUser } from "../../shared/auth-info-context";
+import { UrlInput } from "../../shared/url-input";
 
 type FormFieldMetadata = {
   label: string;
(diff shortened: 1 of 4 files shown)


vilkinsons (changes requested):
@CiaranMn Thanks! Can we please remove the whitespace/gap between the protocol dropdown input and the adjoining text input, so that are one contiguous component?

CiaranMn:
@vilkinsons like this?


https://github.com/user-attachments/assets/21afd42b-cb04-4550-a3be-be25e0b4b9ec

vilkinsons (dismissed):

vilkinsons (approved):

vilkinsons:
Separate thread ongoing in Slack re: unsupported protocols here, and how we handle them across _website_ URL inputs in HASH.

PR #6163, opened 2025-01-15
H-3919: Return error instead of `panic` in `SelectCompiler`

TimDiekmann:

In some very special cases the `SelectCompiler` panics. To avoid crashing the graph if this happens, this is changed to errors instead.

## 🔍 What does this change?

- Expose a `SelectComilerError` instead of panics
- Drive-by: Allow `limit` in more locations
- Drive-by: Make sure the compiler correctly errors when a cursor is passed

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-graph/src/subcommand/server.rs
+++ b/apps/hash-graph/src/subcommand/server.rs
@@ -294,7 +294,7 @@ pub async fn server(args: ServerArgs) -> Result<(), Report<GraphError>> {
     let mut zanzibar_client = ZanzibarClient::new(spicedb_client);
     zanzibar_client.seed().await.change_context(GraphError)?;
 
-    let temporal_client_fn = async |host: Option<String>, port: u16| {
+    let temporal_client_fn = |host: Option<String>, port: u16| async move {
         if let Some(host) = host {
             TemporalClientConfig::new(
                 Url::from_str(&format!("{host}:{port}")).change_context(GraphError)?,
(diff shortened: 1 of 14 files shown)


apps/hash-graph/src/subcommand/server.rs
```diff
@@ -294,7 +294,7 @@ pub async fn server(args: ServerArgs) -> Result<(), Report<GraphError>> {
     let mut zanzibar_client = ZanzibarClient::new(spicedb_client);
     zanzibar_client.seed().await.change_context(GraphError)?;
 
-    let temporal_client_fn = async |host: Option<String>, port: u16| {
+    let temporal_client_fn = |host: Option<String>, port: u16| async move {
```
TimDiekmann, line 297: Drive-by: async closures are unstable, we can use the stable way

libs/@local/graph/api/src/rest/entity.rs
```diff
@@ -661,7 +661,7 @@ impl<'q, 's, 'p: 'q> From<GetEntitiesRequest<'q, 's, 'p>> for GetEntitiesParams<
 #[tracing::instrument(
     level = "info",
     skip_all,
-    fields(actor=%actor_id, %request)
+    fields(actor=%actor_id)
```
TimDiekmann, line 664: Drive-by: This would include the embedding in the log, which is quite large.

CiaranMn (approved):

CiaranMn (approved):

PR #6254, opened 2025-01-25
H-3943: Use correct DNS lookup in HaRPC to allow connections in AWS

indietyp:

This fixes several issues small issues to make harpc work on AWS

1. use lookup instead of resolve, this is slower but at least resolves the `/etc/hosts`
2. add every address associated with a peer during dialing to the cache, meaning repeated cache misses won't happen
3. wrap `libp2p/Connection` using `PeerConnection`, to clean up the logs
4. fix an issue where we would take the whole buffer, leaking `yamux` payload while trying to decode responses 
5. remove `logReverse`, as it is no longer needed and would take seconds to complete when using an address from `/etc/hosts`

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/client/typescript/src/wire-protocol/stream/ResponseFromBytesStream.ts
```diff
@@ -84,6 +84,20 @@ export const make = <E, R>(
 
   return pipe(
     stream,
+    Stream.tap((chunk) => {
```
indietyp, line 87: i am in the process of authoring a PR in effect that'd allow us to do something similar to the [`enabled!` ](https://docs.rs/tracing/latest/tracing/macro.enabled.html)macro in tracing (gate this somewhat expensive computation - and others - to only when enabled)

infra/terraform/hash/hash_application/api.tf
```diff
@@ -76,7 +76,7 @@ locals {
     mountPoints = []
     volumesFrom = []
     dependsOn = [
-      { condition = "SUCCESS", containerName = local.api_migration_container_def.name },
+      # { condition = "SUCCESS", containerName = local.api_migration_container_def.name },
```
TimDiekmann, line 79: This was for testing only
```suggestion
      { condition = "SUCCESS", containerName = local.api_migration_container_def.name },
```

apps/hash-api/src/index.ts
```diff
@@ -586,7 +586,7 @@ const main = async () => {
       }).pipe(
         Effect.provide(
           RpcClient.connectLayer(
-            Transport.multiaddr(`/dns/${rpcHost}/tcp/${rpcPort}`),
+            Transport.multiaddr(`/dns4/${rpcHost}/tcp/${rpcPort}`),
```
TimDiekmann, line 589: What would be required to allow `dns`? shouldn't it be enough to enable listening on `IPv4` on the graph? Is it worth trying out to use `/dns/` and see if it now properly connects to the IPv4 under the assumption that everything else is working now?

libs/@local/harpc/client/typescript/src/net/Connection.ts
```diff
@@ -227,7 +227,13 @@ export const makeUnchecked = (
       ),
       Stream.mapConcat((list) => list),
       // cast needed as uint8arraylist doesn't support Uint8Array<ArrayBuffer> yet
```
TimDiekmann, line 229: I guess this would also be the old behavior then?
```suggestion
```

libs/@local/harpc/client/typescript/src/net/internal/dns.ts
```diff
@@ -177,8 +149,57 @@ export const resolve = (
...
+
+    yield* Effect.fork(logEnvironment(hostname));
+
+    // partition into A and AAAA records
+    const [excluded, satisfying] = Array.partition(
+      records,
+      (record) => record.family === 4,
+    );
+
+    // we cannot determine the TTL of lookup records, therefore we set it to infinity
```
TimDiekmann, line 175: Is this still planned in H-3793? If so, can we add a `TODO` or otherwise cancel the task?

libs/@local/harpc/client/typescript/src/wire-protocol/stream/ResponseFromBytesStream.ts
TimDiekmann: For the time being we keep this code or do you want to wait?

TimDiekmann (commented):
Looks good to me, thank you very much!
I've tested this in AWS and it works as expected 🎉 

I have a few comments, but nothing major.

libs/@local/harpc/client/typescript/src/net/Connection.ts
indietyp: This is still the case, the generic for `Uint8Array` is relatively new and uint8arraylist doesn't make use of it yet, therefore we need to still cast. The current type stipulates that also a `SharedArrayBuffer` could be used, which isn't correct.

libs/@local/harpc/client/typescript/src/net/internal/dns.ts
indietyp: this is still planned, but the TTL on a lookup via `getaddrinfo` doesn't reveal a TTL. For that we'd need [H-3785](https://linear.app/hash/issue/H-3785/create-typescripteffect-dns-package). I can add a comment explaining this.

libs/@local/harpc/client/typescript/src/wire-protocol/stream/ResponseFromBytesStream.ts
indietyp: I'd keep it, there's a bit of performance penalty, but it isn't too large.

libs/@local/harpc/client/typescript/src/net/internal/dns.ts
indietyp: added in 527c77f38

apps/hash-api/src/index.ts
TimDiekmann: I'd say we enable IPv6 in the Graph then rather than restricting it to IPv4

PR #6263, opened 2025-01-27
H-3938, H-3945, H-3946: Custom 'not found' component, editor bug fixes

CiaranMn:

Fixes a bug frontend things:
1. Use a custom 'not found/not accessible' error component. Make sure it show it when visiting an entity the user can't find (and in other appropriate places)
2. In the array editor in the entity editor, show the data type selector if the data type is abstract
3. In the type editor, prevent users creating arrays with no expected types, or property objects with no properties
4. In the type editor, fix a bug where the dropdown became impossible to open after deleting multiple expected types (impossible without unfocusing and refocusing it anyway)

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Testable via preview deployment. Visit a 404 URL or an entity you can't visit
2. For editing fixes, try behavior which should no longer be possible

## 📹 Demo

Screenshot 2025-01-27 at 15 05 37


vilkinsons (approved):
Works like a charm.

For some reason if I navigate to a non-existant page like `https://hash-nu2x21o4b.stage.hash.ai/2340234` (latest staging URL base used here), I see the "user profile page" loading skeleton flash up before the error message appears, though.

We should retain the skeleton while actually loading user pages, but lose it in other cases, to avoid unnecessary jank. Do we want to fix here or in drive-by?

vilkinsons:
@CiaranMn I'll re-review once conflicts resolved

vilkinsons (approved):

PR #6277, opened 2025-01-28
H-3976: Replace `Buffer` in favour of `MutableBuffer`

indietyp:

Right now the `Buffer` is immutable in the sense that we make use of `SubscriptionRef` for atomic effects, should a buffer be used across forks/fibers. This has quite the overhead, especially for something so central, instead the `Effect` code should be replaced with `Either` code.

Profiling has shown that this is a 400-600% speedup, primarily because we no longer synchronize the index and have therefore a lot lower of an overhead.

> in Rust terms this basically moved this from `async` to `sync` code.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/harpc/client/typescript/src/types/ResponseKind.ts
```diff
@@ -126,31 +127,28 @@ export type ResponseKind = Ok | Err;
...
-      // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      if (isOk(kind)) {
-        return yield* Buffer.putU16(buffer, 0);
-      }
-
-      return yield* ErrorCode.encode(buffer, kind.code);
-    }),
+export const encode = implEncode((buffer, kind: ResponseKind) =>
+  // eslint-disable-next-line @typescript-eslint/no-use-before-define
+  isOk(kind)
```
CiaranMn, line 132: What's the motivation for having the eslint disable here (and above on `OkProto`) rather than moving `isResponseKind`, `isOk`, `isErr` above `OkProto`? I appreciate we do this in other files also, just curious, it isn't causing any practical issue.

CiaranMn (approved):
Looks good, one non-blocking question

libs/@local/harpc/client/typescript/src/types/ResponseKind.ts
indietyp: I have the files structured into sections, putting the function at the top was confusing to me when I did it first, it's always:

```
TypeId

Interface

Proto

Functions
	make
	(encode if available)
	(decode if available)
	... other functions
	isX function
	arbitrary
```

I have yet to enforce that order - I don't know how to^^ (but most files should follow it tho)

CiaranMn (approved):

PR #6302, opened 2025-01-30
H-3145: Show sources on property hover in entity editor; bug fixes

CiaranMn:

This PR:
- Shows the sources for a given property value (if any) on hovering over the value in the entity editor
- Fixes some behavior issues with grid table tooltips in general (make them disappear when leaving the tooltip, not when leaving the cell)
- Fixes a couple of bugs in the document inference action
- Fixes a bug when navigating client-side to a type's page

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. If you have sources for properties, testable in preview deployment

## 📹 Demo

https://github.com/user-attachments/assets/f5b2ab26-fd38-46f2-a8a6-11fc350e2aeb

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/flow-activities/infer-metadata-from-document-action.ts
+++ b/apps/hash-ai-worker-ts/src/activities/flow-activities/infer-metadata-from-document-action.ts
@@ -27,10 +27,7 @@ import {
   blockProtocolPropertyTypes,
   systemPropertyTypes,
 } from "@local/hash-isomorphic-utils/ontology-type-ids";
-import type {
-  File,
-  TitlePropertyValue,
-} from "@local/hash-isomorphic-utils/system-types/shared";
+import type { File } from "@local/hash-isomorphic-utils/system-types/shared";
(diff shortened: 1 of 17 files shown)


vilkinsons (approved):
Nice, thank you! Do we need Chonky for page-level (as opposed to doc-level) context citations?

CiaranMn:
> Nice, thank you! Do we need Chonky for page-level (as opposed to doc-level) context citations?

There are two possible ways of getting page-level citations:
1. Ask the LLM to tell us what page they found it on (accuracy needs testing)
2. Only pass in one page at a time, then we know what page it came from (needs Chonky or similar)

vilkinsons:
I'm sure you feel the same, but I don't particularly like the idea of (1) introducing something else subject to error into the loop (and then having to have tests for it, it sometimes fail, etc.) - just seems like working around the issue.

(2) Like you say basically works in that we'd be splitting docs into their semantic components (and creating vector embeddings for each, arranging the 'chunks' hierarchically). We'd then be able to infer claims from chunks and know which one they came from (and where it is situated within the wider document), providing this feedback to users.

PR #6303, opened 2025-01-31
H-3947: Implement higher precision when convert data types in the graph

TimDiekmann:

As a stop-gap solution to do math for any numbers, we're going to

- Use our own JSON type (instead of `serde_json::Value`)
- Use a decimal library to do conversions in the graph
- Store only `f64`

## 🔍 What does this change?

- Introduce a new `Value` type which replaces `serde_json::Value` to have more control over it
- Implement a `Real` type which represents a decimal number
- Use the `Real` type in the newly added `Value`

`Real` behavior:
- Creation from integers is lossless
- Conversion from `f32` and `f64` fails if it's `NaN`
- Converting to primitives may round and truncate
- Can be converted to/from `FLOAT4` and `FLOAT8` in Postgres
- It can be deserialized from number primitives and strings

`Value` behavior:
- Parses and outputs `Number` from/as `f64`
- This means, outside of the graph (API and Postgres) only `f64` precision is retained, but internally, a higher precision is used

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons:
How many number-related data types are we exposing to users as HASH "Data Types" in the ontology/entities UI? (Integer, 64 bit float, 128 bit float, etc.)

TimDiekmann:
> How many number-related data types are we exposing to users as HASH "Data Types" in the ontology/entities UI? (Integer, 64 bit float, 128 bit float, etc.)

Currently, we only expose `number` in TypeScript, which corresponds to a 64 bit floating point (IEEE 754). For embeddings, we use 32 bit. See H-3988 for more thoughts on this topic.
The frontend may restrict `number` further depending on constraints such as whole numbers, min, max, etc.

CiaranMn (dismissed):
Looks good, thanks @TimDiekmann!

TimDiekmann:
Thanks! I need a re-approval due to merge conflicts due to changes in `rustfmt` in the toolchain bump merged earlier today.

CiaranMn (dismissed):

CiaranMn (approved):

PR #6314, opened 2025-02-02
H-3998: Adjust global `turbo` dependencies

TimDiekmann:

The current global dependencies are quite outdated and many files are not included so we manually do it by parsing the // package (the root package) in CI.

To avoid that, the global dependencies should be overhauled.

## 🔍 What does this change?

See the changed files

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -101,15 +101,26 @@
     }
   },
   "globalDependencies": [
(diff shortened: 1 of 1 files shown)


indietyp (approved):
Thank you!

PR #6317, opened 2025-02-02
H-3999: Use `turbo query` to determine changed packages in CI

TimDiekmann:

Currently, we use `turbo run ... dry-run. Instead, we can use `turbo query`.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,11 +36,8 @@ jobs:
       - name: Determine changed packages
         id: packages
         run: |
-          FILTER=$(turbo run lint --dry-run=json --filter '...[HEAD^]' | jq -e '.packages | contains(["//"])' > /dev/null && echo '' || echo '--filter ...[HEAD^]')
-          PACKAGES=$(sh -c "turbo run lint --dry-run=json $FILTER" \
-            | jq '.tasks[]' \
-            | jq 'select(.task == "lint")' \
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -36,11 +36,8 @@ jobs:
       - name: Determine changed packages
         id: packages
         run: |
-          FILTER=$(turbo run lint --dry-run=json --filter '...[HEAD^]' | jq -e '.packages | contains(["//"])' > /dev/null && echo '' || echo '--filter ...[HEAD^]')
-          PACKAGES=$(sh -c "turbo run lint --dry-run=json $FILTER" \
-            | jq '.tasks[]' \
-            | jq 'select(.task == "lint")' \
-            | jq --compact-output --slurp '{ package: [.[].package] | unique, include: [( .[] | {package: .package, directory: .directory })] | unique }')
+          PACKAGES=$(turbo query 'query { affectedPackages { items { name path } } }' \
```
indietyp, line 39: shouldn't this filter for packages that have the lint task?

indietyp (changes requested):
minor comment

.github/workflows/lint.yml
TimDiekmann: There is no `lint` task per package. I have no idea why it worked before, probably because we have a top-level `lint` task. It just assumes that we want to lint every package which has been changed. This is also why I had to exclude `//` in `jq`.

vilkinsons (approved):

indietyp (approved):

PR #6331, opened 2025-02-03
H-4003: Use `uuid.NIL` for public account ID

TimDiekmann:

Semgrep complains about hardcoding an account ID. As this is a well-known constant, we can just use that instead.

### DIFF
--- a/libs/@local/hash-backend-utils/package.json
+++ b/libs/@local/hash-backend-utils/package.json
@@ -52,6 +52,7 @@
     "logform": "2.7.0",
     "redis": "4.7.0",
     "slonik": "24.2.0",
+    "uuid": "11.0.5",
     "wait-on": "8.0.2",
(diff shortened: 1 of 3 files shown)


CiaranMn (approved):

vilkinsons:
@TimDiekmann @CiaranMn We use this in [a lot more places](https://github.com/search?q=repo%3Ahashintel%2Fhash+00000000-0000-0000-0000-000000000000&type=code)... should we change them all?

TimDiekmann:
> @TimDiekmann @CiaranMn We use this in [a lot more places](https://github.com/search?q=repo%3Ahashintel%2Fhash+00000000-0000-0000-0000-000000000000&type=code)... should we change them all?

Changed them when possible: https://github.com/hashintel/hash/pull/6334

PR #6335, opened 2025-02-03
H-4006: SCM base `main` not acceptable

TimDiekmann:

The SCM base `main` is not useable in PRs. Instead, we assign it manually.


indietyp (approved):
Thank you, hopefully this fixes the issue!

PR #6337, opened 2025-02-03
H-3990: Fix duplicated entities in querying if entity has multiple types

TimDiekmann:

When sorting entities by `label` or `typeTitle` and the entity has multiple entity types, the entity is returned twice

## 🔍 What does this change?

- The Query compiler should be adjusted to allow CTEs for filtering and select `DISTINCT ON` the actual `SELECT`, but this was out of scope for this bug
- This PR adds a few views which selects the first/last label/title (so 4 views in total, while 2 additional views are used to avoid duplications)
- These views are used for sorting paths

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- We want to make ontology type IDs ordered in inheritance and entities (H-3997) but this is a bigger topic which requires bigger adjustments. This is a working stop-gap solution to avoid the bug in the first place.


libs/@local/graph/migrations/graph-migrations/v009__entities/down.sql
```diff
@@ -7,3 +7,11 @@ DROP TABLE entity_temporal_metadata;
 DROP TABLE entity_editions;
 DROP TABLE entity_drafts;
 DROP TABLE entity_ids;
+
+DROP VIEW IF EXISTS first_entity_titles;
+DROP VIEW IF EXISTS last_entity_titles;
+DROP VIEW IF EXISTS entity_title_agg;
```
CiaranMn, line 13: Might be better to call this `first_entity_type_title` (and insert `type` everywhere else appropriate) as it correponds to the `type_title`. Entity's don't have titles, their types do.

e.g. also
> Corresponds to the first title of the [`Entity`].

->

> Corresponds to the title of the [`Entity`]'s first `EntityType`.

TimDiekmann: I had that first but found that confusing as well (as it's the title of the entity's types). We could do `first_entity_type_title_for_entity_edition` but this is quite verbose. What do you think?

CiaranMn: `first_type_title_for_entity`?

TimDiekmann: It think that's good!

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
```diff
@@ -716,6 +724,62 @@ impl DatabaseColumn for EntityTypes {
...
+            Self::Label => "label_property",
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum EntityTitles {
+    EditionId,
+    Title,
+}
```
CiaranMn, line 759:
```suggestion
pub enum TypeTitleForEntity {
    EditionId,
    TypeTitle,
}
```

libs/@local/graph/store/src/entity/query.rs
```diff
@@ -459,6 +459,30 @@ pub enum EntityQueryPath<'p> {
     /// # Ok::<(), serde_json::Error>(())
     /// ```
     Embedding,
+    /// Corresponds to the first title of the [`Entity`].
```
CiaranMn, line 462:
```suggestion
    /// Corresponds to the title of the [`Entity`]'s first `EntityType`.
```
etc

CiaranMn (approved):

PR #6339, opened 2025-02-03
H-4008: Fix base branch detection for CI

TimDiekmann:

`turbo` is not able to properly find the commit SHA. Instead, we use `HEAD^`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,9 +41,9 @@ jobs:
       - name: Determine changed packages
         id: packages
         run: |
-          UNIT_TEST_QUERY='query { affectedPackages(base: "[HEAD^]", filter: {has: {field: TASK_NAME, value: "test:unit"}}) { items { name path } } }'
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -91,7 +90,6 @@ jobs:
 
           echo "unit-tests=$UNIT_TEST_PACKAGES" | tee -a $GITHUB_OUTPUT
           echo "integration-tests=$INTEGRATION_TEST_PACKAGES" | tee -a $GITHUB_OUTPUT
-          echo "system-tests=$SYSTEM_TEST_PACKAGES" | tee -a $GITHUB_OUTPUT
```
indietyp, line 94: Why was this removed?

indietyp (changes requested):

.github/workflows/test.yml
TimDiekmann: It's not used

indietyp (approved):

PR #6350, opened 2025-02-04
H-4012: Satisfy required precondition of DNS library

indietyp:

Create the library skeletons for both the N-API library (@local/effect-dns-hickory) and core library (@local/effect-dns-core).

> Tagging both Tim and Ciaran as this is both Rust and Typescript.

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons:
@indietyp Please tag me when test are passing for legal

libs/@local/effect-dns/hickory/.cargo/config.toml
TimDiekmann, line 1: When compiling from the root, does cargo find this file?

indietyp: > Cargo allows local configuration for a particular package as well as global configuration. It looks for configuration files in the current directory and all parent directories. \[…\] If a key is specified in multiple config files, the values will get merged together.

from: [https://doc.rust-lang.org/cargo/reference/config.html](https://doc.rust-lang.org/cargo/reference/config.html)

indietyp: upon further inspection that isn't the case, but I wasn't sure how to add the configuration for only a package in the root workspace.

> At present, when being invoked from a workspace, Cargo does not read config files from crates within the workspace.

TimDiekmann: > Cargo allows local configuration for a particular package as well as global configuration. 

Yes, "local configuration" here refers to the repository root and "global configuration" refers to the home directory I think.

> but I wasn't sure how to add the configuration for only a package in the root workspace.

I don't think this is possible

indietyp: I'd keep it for now like this, if we run into problems we know where to look 🧐

libs/@local/effect-dns/hickory/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "rimraf dist && yarn build:release && tsc --build tsconfig.build.json",
+    "build:debug": "napi build dist --platform --js=impl.js --dts=impl.d.ts",
+    "build:release": "napi build dist --platform --release --js=impl.js --dts=impl.d.ts",
+    "fix:eslint": "eslint --fix .",
+    "lint:eslint": "eslint --report-unused-disable-directives .",
+    "lint:tsc": "tsc --noEmit",
+    "regenerate-packages": "napi create-npm-dir --target=.",
```
TimDiekmann, line 17: I don't find a documentation for this command online. What does it? When it is invoked?

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+napi        = { version = "2.16.15", default-features = false, features = ["napi9", "tokio_rt"] }
+napi-derive = "2.16.13"
+
+[build-dependencies]
+napi-build = "2.1.4"
+
+[package.metadata.sync.turborepo]
+ignore = true
```
TimDiekmann, line 17: Why we need to ignore it?

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+edition = "2021"
+name    = "local_effect-dns-hickory"
+version = "0.0.0"
```
TimDiekmann, line 4:
```suggestion
name              = "local_effect-dns-hickory"
authors.workspace = true
version           = "0.0.0"
edition.workspace = true
license           = "MIT OR Apache-2.0"
publish           = false
```

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
```
TimDiekmann, line 9:
```suggestion
[dependencies]
# Public workspace dependencies

# Public third-party dependencies

# Private workspace dependencies

# Private third-party dependencies
```

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+napi        = { version = "2.16.15", default-features = false, features = ["napi9", "tokio_rt"] }
+napi-derive = "2.16.13"
```
TimDiekmann, line 11:
```suggestion
napi        = { workspace = true, features = ["napi9", "tokio_rt"] }
napi-derive = { workspace = true }
```

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[build-dependencies]
+napi-build = "2.1.4"
```
TimDiekmann, line 14:
```suggestion
napi-build = { workspace = true }
```

libs/@local/effect-dns/hickory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
```
TimDiekmann, line 15:
```suggestion

[lints]
workspace = true

```

TimDiekmann (changes requested):

libs/@local/effect-dns/hickory/Cargo.toml
indietyp: Because the `package.json` isn't auto-generated. Otherwise the name would be rewritten as `@rust/...` and dependencies would be lost.

libs/@local/effect-dns/hickory/package.json
indietyp: This is in preparation for if we want to publish. Not needed when in-tree, but when publishing we need to create a package for _every single target_, which is then installed depending on the architecture. `create-npm-dir` creates the packages and directory structure required for publishing.

TimDiekmann: Makes sense, thanks for explaining!

TimDiekmann (dismissed):

CiaranMn (approved):

vilkinsons (approved):

vilkinsons (approved):

TimDiekmann (approved):

PR #6411, opened 2025-02-11
H-4047: Fix HaRPC tests not running in CI

TimDiekmann:

Some scripts were missing from `package.json`, this fixes that.


indietyp (approved):
Thank you, can't believe I missed those 🤦

PR #6413, opened 2025-02-11
H-4048: Don't suggest ignoring a warning in vitest extension

TimDiekmann:

Instead of suggesting to ignore a warning the warning should be fixed. This removes the setting from the `.vscode/settings.json`. Also, it adds a few other recommended settings.


vilkinsons (approved):
...but we should actually fix the issue. It's lingered a long time.

PR #6416, opened 2025-02-11
H-4023, H-4024, H-4059: Allow updating multiple ontology types at once

TimDiekmann:

We can create multiple types at once but not update them. This adds bulk operations to update ontology types.


CiaranMn (approved):
Tested and working, thanks!

CiaranMn (approved):

PR #6432, opened 2025-02-13
H-4064: Restrict `User` and `Organization` instantiators

CiaranMn:

This PR does a few pieces of cleanup around permissions:
1. Restricts anyone from creating `User` and `Organization` entities outside of the dedicated routes (creating the entities elsewhere doesn't grant any permissions, but if someone does it we will have broken User/Orgs lying around)
2. Don't bother to maintain instantiator permission on `Machine` for anyone (just grant and remove it as needed)
3. Fix migration scripts to account for arbitrary needs to grant/remove permissions to add new types to entities

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Migration scripts and some BE integration tests involve creating orgs

## ❓ How to test this?

1. Check that migrations still run as expected
2. Check that user can create an org via UI (workspace switcher -> create org)
3. Check that user cannot create an organisation or user entity via e.g. `createEntity`


TimDiekmann (dismissed):
Looks good. One ESLint error about an unused import.

TimDiekmann (approved):
Great, thank you!

PR #6566, opened 2025-02-27
H-4103: Write test for common Policies

TimDiekmann:

We require a set of policies to drive the Graph in general. This PR adds common policies and missing features associated with them.

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/6518

## 🔍 What does this change?

- Add `IsAllTypes` and `IsAnyType` to restrict policies to multiple (at least one or all) types in an entity
- Allow filtering entity types (A combination of `All`, `Any`, `Not`, `IsBaseUrl` and `IsVersion`
- Add a range of policies for testing
- Run different sets of request against the policies

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/authorization/src/policies/mod.rs
+++ b/libs/@local/graph/authorization/src/policies/mod.rs
@@ -24,7 +24,8 @@ use self::{
 };
 pub use self::{
     context::{Context, ContextBuilder, ContextError},
-    validation::{PolicyValidationError, Validator},
+    set::{PolicyEvaluationError, PolicySet, PolicySetInsertionError},
+    validation::{PolicyValidationError, PolicyValidator},
 };
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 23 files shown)


libs/@local/graph/authorization/src/policies/mod.rs
```diff
@@ -135,6 +142,23 @@ impl Policy {
...
+        //     } => {
+        //         let resource_filter =
+        //             if arg1.expr_kind() == ast::ExprKind::Lit(ast::Literal::Bool(true)) {
+        //                 Some(arg2)
+        //             } else {
+        //                 None
+        //             };
+        //     }
+        //     _ => return Err(InvalidPolicy::InvalidSyntax.into()),
+        // };
```
TimDiekmann, line 160: Initial attempt to translate a filter from Cedar to our syntax. As we don't need it (yet) I commented out.

CiaranMn (dismissed):

TimDiekmann:
Rebased onto `main`.

CiaranMn (approved):

PR #6644, opened 2025-03-07
H-4183: Enforce correct ordered list numbering in markdown files

TimDiekmann:

# Use proper markdown enumerations

Currently, we are forced to write enumerations like this:

```
1. ...
1. ...
1. ...
```

Instead, it would be better to write them as

```
1. ...
2. ...
3. ...
```

The idea was to make it easier to add new items, but it’s bad for readability and also with AI it’s trivial to update. Probably even linters can easily do that. Actually, I would prefer that every line needs a change if something was inserted before the end.

## 🔍 What does this change?

- Enforce correct ordered list numbering in markdown files
- Update markdownlint commands to use mise exec


apps/hash/README.md
```diff
@@ -139,13 +139,15 @@ See the [respective section in the parent README](../README.md#hash) for descrip
...
-1. Log in
+9. Log in
 
    There are three users seeded automatically for development. Their passwords are all `password`.
 
    - `alice@example.com`, `bob@example.com` – regular users
    - `admin@example.com` – an admin
 
+10. Create an account
+
```
CiaranMn, line 150:
```suggestion
```

PR #6648, opened 2025-03-07
H-4176: Add Model Context Protocol (MCP) servers for Linear, Notion

CiaranMn:

This PR adds [MCP](https://modelcontextprotocol.io/introduction) servers to help in AI-assisted development for:
1. Linear: retrieve issues in current sprint; get single issue and comments.
2. Notion: search pages by title; retrieve page content.

Instructions on adding each to Cursor and Claude Code are included.

It also adds config to automatically load the FileSystem MCP in Cursor, and instructions on how to add the GitHub MCP to Cursor.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Try using the MCP servers in Cursor


apps/mcp/linear/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "@apps/linear-mcp",
+  "version": "0.1.0",
+  "private": true,
+  "description": "MCP server for basic Linear operations",
+  "license": "AGPL-3.0",
+  "type": "module",
+  "bin": "./dist/index.js",
+  "scripts": {
+    "build": "rimraf dist && tsc --build && shx chmod +x dist/*.js",
+    "dev": "chokidar 'dist/main.js' -c 'chmod +x dist/main.js' & tsc --noCheck --watch & yarn inspector & shx chmod +x dist/*.js",
```
TimDiekmann, line 11: Is it intended that all these commands should run in the background (`&`) instead of sequentially (`&&`)?

TimDiekmann (approved):

vilkinsons (approved):

apps/mcp/linear/README.md
```diff
@@ -0,0 +1,63 @@
...
+### Claude Code
+
+**Note**: the Claude config is currently saved for the **directory you ran it from**.
+
+If you move to a different directory, the MCP server will not be there and will need to be readded.
+
+See [this issue](https://github.com/anthropics/claude-code/issues/374).
+
+1. Set `MCP_LINEAR_API_KEY=YOUR_INTEGRATION_SECRET` in the root `.env.local`
+2. `claude mcp add linear yarn workspace @apps/linear-mcp start`
```
TimDiekmann, line 41: I used a slightly different approach btw: I used `Claude mcp add`, added the command, and specified the environment here. That avoids having the MCP key in `.env.local`.

CiaranMn: That works too. The advantage for me in having it in `.env.local` is if I run Claude from a different directory for some reason, I don't have to get the key again and add it (since the config is directory-scoped). I don't think it makes a big difference though. If Cursor worked with loading env from the repo it would be better as you'd only need to write it to `env.local` once and then forget about it rather than having to repeatedly get it.

TimDiekmann (approved):

vilkinsons (approved):

apps/mcp/linear/package.json
CiaranMn: Yes – the `tsc --watch` is recompiling the TS files to JS when the source changes, and `chokidar` is setting `main.js` to be executable when it's recreated. So these two need to keep watching to reactively change files. Then the inspector is the UI client for debugging

vilkinsons:
@CiaranMn Originally [booted](https://github.com/hashintel/hash/actions/runs/13752720314/job/38455714946) from the MQ for lint + deployment failures.

TimDiekmann (dismissed):

vilkinsons (approved):

vilkinsons (approved):

PR #6654, opened 2025-03-09
H-3568: Create diagrams for dependencies for in-code documentation

TimDiekmann:

When looking at crate documentation it's not immediately visible what crates are depending on this crate or what crates this crate depends on. This initiative creates mermaid diagrams and inserts them into the Rustdoc-generated output.

## 🔍 What does this change?

- Add a `dependency-diagram` subcommand to `repo-chores`
- Update all workspace crates to generate a dependency graph

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Displaying the mermaid in the docs requires a script
- To include the mermaid into the docs we use the `simple-mermaid` crate. It's required to be a normal dependency, not a dev-dependency due to how rustdoc works. The crate only imports a script and includes the mermaid, so it's really simple. We could do that on our own if needed.

## 🛡 What tests cover this?

- We have a CI step to verify this

## ❓ How to test this?

Run `cargo doc --workspace --no-deps --all-features --open` to view the generated documentation.

## 📹 Demo

```text
Generate a dependency diagram in Mermaid format

Usage: repo-chores-cli dependency-diagram [OPTIONS]

Options:
  -o, --output <OUTPUT>
          Output file path for the mermaid diagram
          
          When specified, writes the diagram to this file path. When not specified, outputs to stdout.

  -i, --include <INCLUDE>
          Include only crates matching these patterns
          
          Supports glob patterns like "*-graph*" to filter which crates are included. When multiple patterns are
          provided, crates matching any pattern are included.

  -e, --exclude <EXCLUDE>
          Crates to exclude from the diagram
          
          Supports glob patterns like "*-test*" to filter which crates are excluded. When multiple patterns are
          provided, crates matching any pattern are excluded.

  -r, --root <ROOT>
          The root crate to highlight with a thicker border (if any)
          
          When specified, this crate will be highlighted in the diagram with thicker borders.

      --root-deps-only
          Show only dependencies of the root crate
          
          When enabled, the diagram will only include the root crate and its dependencies (direct and transitive).
          Requires the `--root` option.

      --root-deps-and-dependents
          Show both dependencies and dependents of the root crate
          
          When enabled, the diagram will include crates that depend on the root crate as well as the root crate's
          dependencies. Requires the `--root` option.

      --link-mode <LINK_MODE>
          Link generation mode
          
          Controls how documentation links are generated in the diagram.

          Possible values:
          - all:       Create documentation links for all crates
          - non-roots: Create documentation links for all crates except the root
          - none:      Don't create documentation links for any crates

  -n, --no-dedup-transitive
          Do not deduplicate transitive dependencies
          
          By default, the diagram deduplicates transitive dependencies for clarity. Enable this flag to show all
          dependencies even if they're transitive.

  -t, --include-dev-deps
          Include dev dependencies (used for tests and examples)
          
          When enabled, includes dev dependencies in the diagram with dotted arrow style.

  -b, --include-build-deps
          Include build dependencies (used for build scripts)
          
          When enabled, includes build dependencies in the diagram with dashed arrow style.

  -h, --help
          Print help (see a summary with '-h')
```

Example for `hash-graph` as root. In the crate documentation the crates are clickable so you can navigate between crates quicker.

```mermaid
graph TD
    %% Configure the diagram
    linkStyle default stroke-width:1.5px
    classDef default stroke-width:1px
    classDef root stroke-width:3px
    classDef dev stroke-width:1px
    classDef build stroke-width:1px
    %% Legend
    %% --> : Normal dependency
    %% -.-> : Dev dependency
    %% ---> : Build dependency
    0[hash-graph]
    class 0 root
    1[type-system]
    2[hash-tracing]
    3[hash-temporal-client]
    4[hash-graph-types]
    5[hash-graph-type-fetcher]
    6[hash-graph-test-server]
    class 6 dev
    7[hash-graph-store]
    8[hash-graph-postgres-store]
    9[hash-graph-authorization]
    10[hash-graph-api]
    11[hash-codec]
    12[harpc-server]
    13[harpc-codec]
    14[error-stack]
    15[hash-graph-test-data]
    class 15 dev
    16[hash-graph-temporal-versioning]
    17[hash-status]
    18[hash-graph-type-defs]
    19[hash-graph-validation]
    20[harpc-types]
    21[harpc-tower]
    22[harpc-system]
    23[harpc-client]
    24[harpc-wire-protocol]
    25[harpc-net]
    0 --> 2
    0 --> 6
    1 -.-> 15
    1 --> 11
    2 --> 14
    3 --> 4
    4 --> 1
    4 --> 16
    5 --> 7
    6 --> 10
    7 --> 3
    7 --> 9
    8 --> 17
    8 --> 19
    9 --> 4
    10 --> 5
    10 --> 18
    10 --> 8
    10 --> 12
    11 --> 24
    12 -.-> 4
    12 -.-> 23
    13 --> 20
    13 --> 14
    16 --> 11
    18 --> 17
    19 --> 7
    21 -.-> 25
    21 --> 25
    22 --> 21
    23 --> 22
    24 -.-> 20
    24 --> 20
    24 --> 14
    25 -.-> 13
    25 --> 11
    25 --> 13
```


apps/hash-graph/src/main.rs
```diff
@@ -1,3 +1,7 @@
+//! # Hash Graph
```
vilkinsons, line 1:
```suggestion
//! # HASH Graph
```

libs/@local/hql/span/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+//! # HQL Span
```
vilkinsons, line 1:
```suggestion
//! # HashQL Span
```

libs/@local/repo-chores/rust/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+//! # Hash Repo Chores
```
vilkinsons, line 1:
```suggestion
//! # HASH Repo Chores
```

vilkinsons (approved):

vilkinsons:
> To include the mermaid into the docs we use the simple-mermaid crate. It's required to be a normal dependency, not a dev-dependency due to how rustdoc works. The crate only imports a script and includes the mermaid, so it's really simple. We could do that on our own if needed.

@TimDiekmann Captured this as [H-4195](https://linear.app/hash/issue/H-4195).

CiaranMn (approved):

PR #6728, opened 2025-03-19
H-4241: Don't allow coercing single GraphQL inputs into arrays

CiaranMn:

The [GraphQL spec](https://spec.graphql.org/draft/#sel-FAHjBJFCAACE_Gh7d) allows clients to pass a single value where  a list of values is expected, and to have this coerced into a list of size 1. 

Because of this, by default the codegen tool we use to generate TS types from GraphQL schemas produces types where single values are acceptable for arrays. This meant we didn't notice an issue converting from a query accepting `VersionedUrl[]` moving to `VersionedUrl[][]` – passing a single `VersionedUrl[]` was not producing a type error. The effect of this is that adding multiple types to an entity caused a crash. 

The PR fixes that by disabling the array input coercion behavior, and also moves common GraphQL codegen configs into a shared place so that it's easier to modify it for all places we are using the codegen. There's more that could be done to reduce duplication across configs but the abstraction is probably more effort than it's worth at this point (because it would need to take various arguments which vary across codegen config locations).

As a drive-by I also stopped the app making requests to the API for things that aren't available for a local draft entity, which wasn't causing any crashes or breaking anything, but was polluting error logs with bad requests.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing tests cover entity creation and updating but not yet multiple types.

## ❓ How to test this?

1. Check that creating a new entity and adding a couple of types to it works (you don't need to save it to the db, in prod it is crashing at the point of adding a second type to a local draft entity)


TimDiekmann (dismissed):
Oh, that explains a lot, thank you. Except a few linting errors due to #6718 it looks good 🙂

TimDiekmann (approved):

PR #6738, opened 2025-03-21
H-4250: Use a `IntoReport` trait to determine conversion to a `Report`

TimDiekmann:

Currently, when using errors in traits, it's annoying that you cannot properly specify errors in associated types.
```rust
// Requires `Result<_, Report<Self::Error>>` as return to allow `.change_context()` 
```

Also, `Result<_, Report<!>>` is not working as well as it could work.

This PR (re-)adds `IntoReport` to allow it in associated types, such as:
```rust
// Allows `impl Error`, `Report<impl Error>`, and `!` as type
type Error: IntoReport;
```

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

A trait is added to the interface but the trait is very simple and outweighs the drawbacks of a new trait.

## 🛡 What tests cover this?

The `ResultExt` trait was changed, so pretty much every test is using the new trait indirectly

### DIFF
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -948,7 +948,7 @@ impl<C> Extend<Self> for Report<[C]> {
     }
 }
 
-/// Provides methods to work with errors in the context of error-stack's reporting system.
+/// Provides unified way to convert an error-like structure to a [`Report`].
 ///
 /// This trait allows both [`Report<C>`] instances and regular error types to be converted into a
 /// [`Report`]. It is automatically implemented for any type that can be converted into a [`Report`]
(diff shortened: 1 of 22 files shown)


libs/error-stack/src/report.rs
```diff
@@ -947,3 +947,56 @@ impl<C> Extend<Self> for Report<[C]> {
         }
     }
 }
+
+/// Provides methods to work with errors in the context of error-stack's reporting system.
```
indietyp, line 951: I don't know if I like this heading. It promises more than the trait is supposed to do

indietyp (changes requested):
minor documentation comment otherwise looking good. Thank you!

libs/error-stack/src/report.rs
TimDiekmann:
```suggestion
/// Provides unified way to convert an error-like structure to a [`Report`].
```

indietyp (approved):

PR #6761, opened 2025-03-23
H-4253: Refactor HashQL Diagnostic Categories

indietyp:

The previous system made use of a quite complicated system to declare diagnostic categories. This PR changes the system to an enum and trait based approach instead.

Additionally the `rob` module has been removed and `Severity` is now always boxed.

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann (approved):
Great to see that change, thank you!

PR #6785, opened 2025-03-25
H-4015: Fix behavior of `any` in structural queries

TimDiekmann:

Filtering optional fields using the `any` filter currently has issues as the implementation effectively removes rows if there couldn't be a join.

## 🔍 What does this change?

- Chained `OUTER JOIN` statements will now use `LEFT OUTER JOIN` to retain the optionality
- merge-constraints are added to the JOIN expression itself
- Drive-by: Use `pretty-assertions` in Postgres-unit tests

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/graph/postgres-store/src/store/postgres/query/compile.rs
```diff
@@ -1001,6 +1015,12 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
                     },
                 );
 
+                if is_outer_join_chain {
+                    join_expression.join = JoinType::LeftOuter;
+                } else if join_expression.join != JoinType::Inner {
+                    is_outer_join_chain = true;
+                }
```
TimDiekmann, line 1022: @hashdotai Are we covered with this logic? It implies that we going to change any subsequent join to be a `LEFT OUTER JOIN`, even if it would be another JOIN as defined in the relations.

TimDiekmann: Well, it commented on this code section twice separately, I assume yes 😅

PR #6799, opened 2025-03-27
H-4296: Remove docs reference to `Empty List` data type as a primitive

vilkinsons:

Removes references to `Empty List` as a primitive data type from the docs.

### DIFF
--- a/apps/hashdotai/guide/04_types/04_data-types.mdx
+++ b/apps/hashdotai/guide/04_types/04_data-types.mdx
@@ -28,7 +28,6 @@ The following primtive data types are also supported in HASH:
 - `Null`: A placeholder value representing 'nothing'
 - `Object`: A plain JSON object with no predefined structure
 
-Historically, an `Empty List` data type was also supported, ensuring that all possible values representable within JavaScript Object Notation (JSON) were also representable in HASH by default, but this was later removed as a generally-available "primitive".
 
 </Toggle>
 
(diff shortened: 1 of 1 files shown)


vilkinsons:
@CiaranMn Should we just lose line 31 entirely, do you think?

apps/hashdotai/guide/04_types/04_data-types.mdx
```diff
@@ -19,15 +19,16 @@ HASH contains a series of primitive data types, which are widely used, and from
...
+<Toggle title="Advanced primitive data types">
 
-The following primtive data types are also supported in HASH, ensuring that all values representable within JavaScript Object Notation (JSON) are also representable in HASH:
+The following primtive data types are also supported in HASH:
 
 - `Null`: A placeholder value representing 'nothing'
 - `Object`: A plain JSON object with no predefined structure
-- `Empty List`: An empty List
+
+Historically, an `Empty List` data type was also supported, ensuring that all possible values representable within JavaScript Object Notation (JSON) were also representable in HASH by default, but this was later removed as a generally-available "primitive".
```
vilkinsons, line 31:
```suggestion
Historically, an `Empty List` data type was also provided, ensuring that all possible values representable within JavaScript Object Notation (JSON) were also provided for by primitive data types in HASH (by default). This was later removed, and to replicate its functionality (if required) you should use an array without any elements.
```

vilkinsons: @TimDiekmann Is this accurate/could it be better explained?

TimDiekmann: I would probably just drop that line TBH

CiaranMn: Agreed

vilkinsons:
```suggestion
```

PR #6828, opened 2025-03-31
H-4252: Adjust J-Expr to new syntax

indietyp:

During the GB we have slightly adjusted the syntax to be less complicated. These changes should be upstreamed to J-Expr.
As capabilities have expanded significantly and some changes were more intrusive than others I've decided to overhauled the parser (or parts of the parser) to be more up to the language specification we outlined.

Additionally the parser and lexer are now fully tested and boast an exhaustive test suite.
I also took this as an opportunity to bring the error messages and the errors that we report up to a new level. Specifically we've moved away from more generic to more specific errors to help guide the user to a solution.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/syntax-jexpr/src/lexer/snapshots/hashql_syntax_jexpr__lexer__test__emoji_family.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: libs/@local/hashql/syntax-jexpr/src/lexer/mod.rs
+description: Family emoji with zero-width joiners
+expression: 👨‍👩‍👧‍👦
```
TimDiekmann, line 4: I appreciate you're testing multiple emojis, but isn't one enough? 😄

libs/@local/hashql/syntax-jexpr/src/parser/array/snapshots/hashql_syntax_jexpr__parser__array__visit__tests__unclosed_array.snap
```diff
@@ -0,0 +1,14 @@
...
+---
+[31m[lexer::lexer::unexpected-eof] Error:[0m Lexer
+   ╭─[ <unknown>:1:6 ]
+   │
+ 1 │ [1, 2
+   │      │ 
+   │      ╰─ Unexpected end of file
+   │ 
+   │ Help: Make sure all expressions, brackets, and string literals are properly closed
+───╯
```
TimDiekmann, line 14: I guess this is the best we can do here? (Absolutely fine with an EOF error)

TimDiekmann (approved):
I have a few minor questions, but it looks great, very good test coverage!

libs/@local/hashql/syntax-jexpr/src/lexer/snapshots/hashql_syntax_jexpr__lexer__test__trailing_comma.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: libs/@local/hashql/syntax-jexpr/src/lexer/mod.rs
+description: Array with trailing comma
+expression: "[1, 2, 3,]"
```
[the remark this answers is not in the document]
indietyp: We don't support this, we only lex it. The parser is currently 100% compliant with the JSON spec. In the future we can opt to demote some diagnostics (like single - or even multiple - trailing commas) to a diagnostic or just swallow them. But I left that as an exercise for later because it's easier to write a correct parser first, and then make some affordances to the user than the other way around.

libs/@local/hashql/syntax-jexpr/src/lexer/snapshots/hashql_syntax_jexpr__lexer__test__emoji_family.snap
indietyp: Sadly not. Emoji are a mind field, and depending on the type of multi-byte emoji and how they're joined the representation in e.g. ariadne or during parsing is different. It's a world full of fun and adventure.

libs/@local/hashql/syntax-jexpr/src/parser/array/snapshots/hashql_syntax_jexpr__parser__array__visit__tests__unclosed_array.snap
indietyp: I'll defer this to another PR (as this is already a very chonky) one, but actually no! There are ways to - very easily - make the diagnostic better here.

I wanted to do this anyway, but basically the idea is that we add a new method `advance_require` or `advance_expect` which takes a `SyntaxKindSet` of allowed tokens. (the normal `advance` would basically just be `advance_require` with `SyntaxKindSet::ALL`. If that token is not present we simply error out, even if EOF. That way the error message would change from: `Unexpected end of file` to `Unexpected end of file, expected ]`. Which is a lot friendlier.

PR #6863, opened 2025-04-04
H-4316: Implement action and policy handling

TimDiekmann:

We can already store principals in the database. This PR allows storing actions and full policies as well.
In addition, it changes a few minor things around action constraints.

## 🔍 What does this change?

- Implement database layout for actions and policies
- Change `ActionConstraints` to be only a vector of actions
- Implement querying for policies basted on an actor

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Expose policy handling to Node API
- Migrate existing database to use new tables
- Use new policies to check permissions (Migrate away from SpiceDB)

## 🛡 What tests cover this?

- An extensive test suite for policy and action handling has been added


libs/@local/graph/migrations/graph-migrations/v003__policies/up.sql
TimDiekmann, line 1: @hashdotai please check for consistency and redundancy. Do the constraints make sense? Do we use it consistently in `postgres-store`?

libs/@local/graph/postgres-store/src/permissions/error.rs
TimDiekmann, line 1: @hashdotai Are all errors consistently used/used at all? To all functions mention the used errors?

libs/@local/graph/postgres-store/src/permissions/mod.rs
TimDiekmann, line 1: @hashdotai carefully review these changes, Check for consistency, proper error handling, accurate documentation, etc.

libs/@local/graph/postgres-store/src/permissions/mod.rs
```diff
@@ -862,108 +925,448 @@ impl<C: AsClient, A: AuthorizationApi> PostgresStore<C, A> {
...
+    /// # Errors
+    ///
+    /// - [`PrincipalNotFound`] if the actor with the given ID doesn't exist
+    /// - [`StoreError`] if a database error occurs
+    ///
+    /// [`PrincipalNotFound`]: PrincipalError::PrincipalNotFound
+    /// [`StoreError`]: PrincipalError::StoreError
+    // TODO: We should probably allow filtering based on the action as well here
+    #[expect(clippy::too_many_lines)]
+    pub async fn get_policies_for_actor(
```
TimDiekmann, line 1219: @hashdotai this is the hearth of this module. Carefully review the function and make sure the documentation is accurate, all code paths are covered (which is hard to measure with Codecov as this largely comes from an SQL statement), and it's properly tested. (Note, that the mentioned TODO on filtering based on an action is a potential future optimization and is intentionally not considered in this PR)

libs/@local/graph/postgres-store/src/permissions/error.rs
TimDiekmann: What do you mean by error chains? We use `change_context` to make sure the error root is retained.

libs/@local/graph/postgres-store/src/permissions/mod.rs
TimDiekmann: Putting it into a constant is good. Where would you store the prepared statement? Create a ticket for both.

libs/@local/graph/authorization/src/policies/action/mod.rs
```diff
@@ -1,28 +1,81 @@
-#![expect(
-    clippy::empty_enum,
-    reason = "serde::Deserialize does not use the never-type"
-)]
-
 use alloc::sync::Arc;
-use core::fmt;
+#[cfg(feature = "postgres")]
+use core::error::Error;
+use core::{fmt, str::FromStr};
```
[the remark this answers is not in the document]
TimDiekmann: `FromStr` uses `Error`, yes, but we don't need the type to be imported for that.

libs/@local/graph/postgres-store/src/permissions/mod.rs
TimDiekmann: @hashdotai Could you elaborate on that?

> The handling of policy constraints is currently simple (always None)

Do you mean the tests or in this implementation? What do you mean by "policy constraints" in the first place, we have three constraints, principals, actions, and resources.

libs/@local/graph/postgres-store/src/permissions/mod.rs
TimDiekmann: I created H-4333 for using prepared statements. Just moving out the Query to the constant does not help here because it's only used once and the prepared parameters are important.

libs/@local/graph/authorization/src/policies/action/mod.rs
```diff
@@ -1,28 +1,79 @@
...
+    ViewEntity,
+
     Update,
+
     Instantiate,
 }
 
-impl CedarEntityId for ActionId {
+#[cfg(feature = "postgres")]
+impl<'a> postgres_types::FromSql<'a> for ActionName {
```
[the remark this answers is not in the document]
TimDiekmann: Semi-intentional. I marked `postgres_types` as public optional dependency.

libs/@local/graph/postgres-store/src/permissions/mod.rs
TimDiekmann: Ah, got it. Yes, that's a placeholder for potential future constraints, but I think I'll just remove that in a follow-up.

CiaranMn (approved):

PR #6918, opened 2025-04-14
H-4386: HashQL: Type System Cleanup

indietyp:

Smaller tasks that need to be done:

1. UnificationContext should be renamed and re-purposed as Environment

2. Remove the `Error` variant

The variant is superfluous and means that type errors are "spreading" during unification, which is undesired.

Instead we should make use of the `diagnostics` in `Environment`, taking them into account we're able to determine if execution failed.

3. Remove the coercion of `Unknown`

Unknown right now functions basically as a weaker `Infer`, this is undesired. Given: `let x: ? = 3`, narrowing `x` to be `Integer` instead would seem to be correct here on first glance, but has the problem that any subsequent unification will see `let x: Integer = 3`, which means that e.g. setting any number here, would result in a type error. This is undesired.

For example:

```typescript
let x: ? = 3 in
let y = x in
x + y
```

in the current system this unifies to:

```
let x: Integer = 3 in
let y: Integer = x in
Integer + Integer
```

This seems correct, but let's take a look at a more complex example:

```typescript
// Type ID 1 is Unknown
// Type ID 2 is Number
// Type ID 3 is String

// A function that works with an 'any' (Unknown) type
let process = fn<T>(value: T): T -> value in

// Usage in two different contexts
let x = process(42) in        // T = Number
let y = process("hello") in   // T = String
```

This would *fail*, because the first `process(42)` narrows `T` from `Unknown` to `Number`, meaning that the second invocation would fail.

Note that this does not affect `Infer`, the following will still compile:

```typescript
let x: _ = 3 in
let y = x in
x + y
```

into:

```typescript
let x: Integer = 3 in
let y: Integer = x in
Integer + Integer
```

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/type/environment.rs
```diff
@@ -0,0 +1,370 @@
...
+        &mut self,
+        variance: Variance,
+        closure: impl FnOnce(&mut Self) -> T,
+    ) -> T {
+        let old_variance = self.variance;
+
+        // Apply variance composition rules
+        self.variance = match (old_variance, variance) {
+            // When going from covariant to contravariant context or vice versa, flip to
+            // contravariant
```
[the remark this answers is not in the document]
indietyp: that comment does not exist.

libs/@local/hashql/core/src/type/mod.rs
```diff
@@ -399,157 +347,131 @@ fn unify_type_covariant(context: &mut UnificationContext, lhs: TypeId, rhs: Type
...
-        (TypeKind::Unknown, rhs) => {
+        (TypeKind::Unknown, _) => {
             // In covariant context: Unknown (lhs) is a supertype of any type (rhs).
             // This is always valid - any value can be used where Unknown is expected.
-            // Even in a strictly variance-aware system, updating Unknown to a more precise type
-            // is beneficial for type inference and error reporting.
-            // The compatibility check succeeds due to variance, but we also update for precision
-            let rhs = rhs.clone();
-            context.update_kind(lhs_id, rhs);
+            // We do not narrow the type because that would potentially create a false-positive type
```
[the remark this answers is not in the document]
indietyp: this is just for testing. So this is fine. We do this because we also want to catch non-fatal diagnostics.

libs/@local/hashql/core/src/type/environment.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        unify_type_impl(self, lhs, rhs);
+
+        self.boundary.exit(lhs, rhs);
+    }
+
+    pub(crate) fn structurally_equivalent(&self, lhs: TypeId, rhs: TypeId) -> bool {
+        let mut environment = StructuralEquivalenceEnvironment::new(self.environment);
+
+        environment.structurally_equivalent(lhs, rhs)
```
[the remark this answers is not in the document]
indietyp: We have exactly that? The environment for both is different as the concerns are different.

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,508 @@
...
+    /// # Panics
+    ///
+    /// If the item's ID is not found in the arena.
+    pub fn update(&mut self, item: T)
+    where
+        T: HasId,
+    {
+        let id = item.id();
+
+        let inserted = self.items.set_mut(id.as_usize(), item);
```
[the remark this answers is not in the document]
indietyp: The panic is documented and intended.

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,508 @@
...
+impl<T> Index<T::Id> for TransactionalArena<T>
+where
+    T: HasId,
+{
+    type Output = T;
+
+    fn index(&self, index: T::Id) -> &Self::Output {
+        self.items
+            .get(index.as_usize())
+            .expect("Item with id {id} not found")
```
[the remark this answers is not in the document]
indietyp: fixed in [278027c](https://github.com/hashintel/hash/pull/6918/commits/278027ce0101712d0ed072afa9bf2c2fa4c33bda)

Cargo.toml
TimDiekmann, line 1: Why changing this?

libs/@local/hashql/core/Cargo.toml
TimDiekmann, line 1: Please use workspace to set dependencies

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,510 @@
...
+        self.items.push_back_mut(item);
+        id
+    }
+
+    /// Updates an item in the arena.
+    ///
+    /// # Panics
+    ///
+    /// If the item's ID is not found in the arena.
+    pub fn update(&mut self, item: T)
```
TimDiekmann, line 72: You can call `update_with` from here instead

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    /// Updates an item in the arena using the provided closure.
+    ///
+    /// # Panics
+    ///
+    /// If the item with the given ID is not found in the arena.
+    pub fn update_with(&mut self, id: T::Id, closure: impl FnOnce(&mut T))
+    where
+        T: HasId + Clone,
```
TimDiekmann, line 89: Why we need `Clone`?

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,510 @@
...
+        self.items.push_back_mut(item(id));
+        id
+    }
+
+    /// Adds an existing item to the arena.
+    ///
+    /// # Returns
+    ///
+    /// The ID assigned to the newly added item.
+    pub fn push(&mut self, item: T) -> T::Id
```
TimDiekmann, line 57: You can use `push_with` instead

libs/@local/hashql/core/src/arena/transaction.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    /// Retrieves a mutable reference to an item from the arena by its ID.
+    ///
+    /// # Returns
+    ///
+    /// `Some(&mut T)` if the item exists, or `None` if no item with the given ID was found.
+    pub fn get_mut(&mut self, id: T::Id) -> Option<&mut T>
+    where
+        T: HasId + Clone,
```
TimDiekmann, line 117: Why is cloning required?

libs/@local/hashql/core/src/type/environment.rs
```diff
@@ -0,0 +1,370 @@
...
+    arena::transaction::{Checkpoint, TransactionalArena},
+    span::SpanId,
+};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default)]
+pub enum Variance {
+    #[default]
+    Covariant, // Same direction
+    Contravariant, // Opposite direction
+    Invariant,     // Exact match
```
TimDiekmann, line 21: The comments should be doc-comments or at least be placed like one

libs/@local/hashql/core/src/type/environment.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// - Two structs with the same field names and types are structurally equivalent
+    /// - Two closures with the same parameter types and return type are structurally equivalent
+    /// - Two generic types are structurally equivalent if their parameters and constraints match
+    ///
+    /// This function handles recursive types by using a recursion guard to prevent infinite
+    /// recursion.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the types are structurally equivalent, `false` otherwise
```
TimDiekmann, line 235: A quite redundant comment I think

libs/@local/hashql/core/src/type/environment.rs
```diff
@@ -0,0 +1,370 @@
...
+    pub fn structurally_equivalent(&mut self, lhs: TypeId, rhs: TypeId) -> bool {
+        let lhs_type = &self.environment.arena[lhs];
+        let rhs_type = &self.environment.arena[rhs];
```
TimDiekmann, line 238: These panic if the type id does not exist. Either we should handle the case or document it

Cargo.toml
indietyp: seems like taplo did some formatting here. did not intend to change anything. reverted in [caf172b](https://github.com/hashintel/hash/pull/6918/commits/caf172b64e4c0c480e5dd40f90ec33f89fa63a7b)

libs/@local/hashql/core/src/type/environment.rs
indietyp: documented in [a4a7829](https://github.com/hashintel/hash/pull/6918/commits/a4a7829ca46e5f4a762eb74fe98f9b706c2e2780)

libs/@local/hashql/core/src/type/environment.rs
indietyp: removed in [a4a7829](https://github.com/hashintel/hash/pull/6918/commits/a4a7829ca46e5f4a762eb74fe98f9b706c2e2780)

libs/@local/hashql/core/src/type/environment.rs
indietyp: Considering that we assume that `TypeId` is always in the environment (otherwise it'll fail later on) this is intended. We assume throughout the `hashql-core/types` that every type exists in the arena, (as every type needs to be allocated) considering one can only update or add but not remove this is a pretty safe bet - except in cases where one uses the TypeId of a different arena - which would either way be a tremendous bug and something we can't really recover from. Additionally there's no real way to have a structurally equivalent type if the type isn't present.

I have added panic docs in [a4a7829](https://github.com/hashintel/hash/pull/6918/commits/a4a7829ca46e5f4a762eb74fe98f9b706c2e2780)

libs/@local/hashql/core/src/arena/transaction.rs
indietyp: Requirement of `rpds`, this is using structural sharing for the checkpointing system, meaning that to be able to change the type, in case the inner Vector is currently shared with e.g a checkpoint, a copy (well clone in this case) of that chunk needs to be made first to allow for mutable access.

libs/@local/hashql/core/src/arena/transaction.rs
indietyp: explained in https://github.com/hashintel/hash/pull/6918#discussion_r2043837651

libs/@local/hashql/core/src/arena/transaction.rs
indietyp: removed `push` in [b466c4b](https://github.com/hashintel/hash/pull/6918/commits/b466c4b109aae5c83fdd623df7d1cfe297208a34), as it is prone to breaking an important invariant of the arena - `T::id` should return the assigned id, with `push` you can just push in random types where the id isn't the same.

I've also added an additional guard in [c828ccf](https://github.com/hashintel/hash/pull/6918/commits/c828ccfb057b40daa4793a9594c9481878f592a9)

libs/@local/hashql/core/Cargo.toml
indietyp: fixed in [b40ca01](https://github.com/hashintel/hash/pull/6918/commits/b40ca01d67d771036b1d91198345afffdbdecff3)

TimDiekmann (approved):

PR #6956, opened 2025-04-20
H-4383, H-4385, H-4395, H-4396: Rework Type System

indietyp:

Completely rework the type system to be based on a lattice approach, instead of only being unification. When implementing H-4383 I found that the current system is insufficient. So after a lot of tears, sweat and hard word I've re-implemented the system with proper underpinnings in type theory creating a type lattice (type-🥬).

This resolved many of the existing tickets in one fell swoop. *a lot* of the code is just testing, making sure that everything is up to par.

This implements much of the required infrastructure to complete:

* https://linear.app/hash/issue/H-4394/hashql-genericparameter-should-lazily-resolve
* https://linear.app/hash/issue/H-4387/hashql-infer-variable-should-lazily-resolve
* https://linear.app/hash/issue/H-4407/hashql-type-constraint-gathering
* https://linear.app/hash/issue/H-4384/hashql-type-instantiation

Which will be (hopefully) all be smaller in scope.

This implements much of the logic that is present in languages like TypeScript, from equivalence over intersections and unions to full support for recursive types to subtyping relationships and proper variance controls.

> One of the central tests are the `assert_lattice_laws`, it's an automated test suite that takes 3 unrelated types and checks if all of the lattice equivalence relationships are upheld. Every single of the types passes it.

> The tests are so plentiful, as many of the behaviour that's tested is nuanced, but we have a coverage percentage of >95%

This PR:

The changes in this PR:

The changes in this PR:


TimDiekmann (approved):
Thanks a lot, looks good to me. As mentioned in the offline call, there might be a few things which can be improved, but these are only minor (such as a few smallish naming suggestions). Generally, this looks very solid!

PR #6970, opened 2025-04-23
H-4346: Bump tool and recursive Rust dependencies

TimDiekmann:

Renovate currently does not cover all dependency updates. To update them manually, the easiest way is to upgrade via
```sh
mise upgrade --bump
cargo update  --ignore-rust-version
```


indietyp (approved):
Thank you!

PR #7027, opened 2025-04-29
H-4438: Implement initial codegen capabilities

TimDiekmann:

We need to generate types based on our Rust types. The current capabilities are quite limited, this is why we use `oxc` to generate TypeScript based on the `specta::Type` derive macro. Long term, we want to move to Facet, however, the facet crate is quite new and does not have the feature set, yet. For this reason, we don't use the structures from Specta but write our own. This also reflects the features we support which would be harder to track with externally managed types.

🔗 Related links

- https://github.com/oxc-project/oxc/pull/10676

## 🔍 What does this change?

- Adds a new `hash-codegen` crate to generate TypeScript types from Rust types
- Implements type definition conversion from `specta::Type` to our own type system
- Adds support for TypeScript code generation using the `oxc` crate
- Handles various enum serialization strategies (externally tagged, internally tagged, adjacently tagged, untagged)
- Adds comprehensive test suite with snapshot tests for different enum types
- (temporarily) Removes the dependency on `facet` and `facet-core` in favor of `Specta`

Pre-Merge Checklist 🚀

## 🚢 Has this modified a publishable library?

This PR:

The changes in this PR:

## 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

## ⚠️ Known issues

- The implementation is focused on enum types for now, with limited support for other complex types
- The TypeScript generator currently does not support all TypeScript features
- Some advanced features like struct flattening, structs, or generics are not yet implemented
- The `oxc` crate currently does not allow the usage of `allocator-api2` with the `nightly` feature enabled. A PR to fix this is created, thanks to their team on the quick responses! https://github.com/oxc-project/oxc/pull/10676

## 🐾 Next steps

- H-4458: Implement field flattening in Codegen

## 🛡 What tests cover this?

- Snapshot tests for different enum types (externally tagged, internally tagged, adjacently tagged, untagged)
- Tests verify both the type definition conversion and the TypeScript output

❓ How to test this?

  1. Checkout the branch
  2. Run cargo test --package hash-codegen to verify the test suite passes
  3. Examine the generated TypeScript outputs in the snapshot files


libs/@local/codegen/src/typescript.rs
```diff
@@ -0,0 +1,284 @@
...
+                            Some(
+                                self.ast
+                                    .alloc_ts_type_annotation(SPAN, self.visit_type(&field.r#type)),
+                            ),
+                        ),
+                    );
+                }
+            }
+            Fields::Unnamed { fields: _ } => {
+                unimplemented!("Internally tagged tuple-variant `{}`", variant.name);
```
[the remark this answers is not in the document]
TimDiekmann: Unnamed internally tagged enums don't make sense.

Cargo.toml
```diff
@@ -375,3 +376,7 @@ unneeded_field_pattern           = "allow" # Actually, the opposite would be a g
 [workspace.metadata.cargo-shear]
 # We use the `futures` crate but the dependencies are required for public dependencies
 ignored = ["futures-core", "futures-io", "futures-sink"]
+
+[patch.crates-io]
+oxc    = { git = "https://github.com/TimDiekmann/oxc", rev = "333f583e" }
```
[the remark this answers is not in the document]
TimDiekmann: Create a ticket for this.

libs/@local/codegen/src/lib.rs
```diff
@@ -4,12 +4,50 @@
...
+pub struct TypeCollection {
+    types: HashMap<Cow<'static, str>, (SpectaID, TypeDefinition)>,
+    collection: specta::TypeCollection,
+    // collection: HashMap<Cow<'static, str>, Definition>,
+}
+
+impl TypeCollection {
+    pub fn register<T: specta::NamedType>(&mut self) {
+        self.collection.register_mut::<T>();
+        let data_type = self.collection.get(T::ID).unwrap_or_else(|| unreachable!());
```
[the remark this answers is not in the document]
TimDiekmann: This is unreachable, it's literally added a line above

libs/@local/codegen/src/typescript.rs
```diff
@@ -0,0 +1,284 @@
...
+            .build(
+                &self.ast.program(
+                    SPAN,
+                    ast::SourceType::d_ts(),
+                    "test",
+                    self.ast.vec(),
+                    None,
+                    self.ast.vec(),
+                    self.ast.vec1(
+                        self.visit_type_definition(name, &self.collection.types[name].1)
```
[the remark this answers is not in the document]
TimDiekmann: This might be helpful in the future. The relevant ticket is H-4460: Export generated type definitions to a file

libs/@local/codegen/src/typescript.rs
```diff
@@ -0,0 +1,284 @@
...
+        let mut types = self.ast.vec();
+
+        for variant in &enum_type.variants {
+            types.push(self.visit_enum_variant(variant, &enum_type.tagging));
+        }
+
+        self.ast.ts_type_union_type(SPAN, types)
+    }
+
+    fn visit_type(&self, r#type: &Type) -> ast::TSType<'a> {
```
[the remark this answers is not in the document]
TimDiekmann: Yes, you're right. We should split the single match statements into multiple... not.

libs/@local/codegen/tests/codegen/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    let settings = TypeScriptGeneratorSettings::default();
+    let generator = TypeScriptGenerator::new(&settings, collection);
+
+    let mut tests = Vec::new();
+    for name in &names {
+        for target in &targets {
+            let test_name = format!("{name}::{target:?}");
+            let generated = generator.generate(name);
+            tests.push(Trial::test(test_name.clone(), move || {
+                assert_snapshot!(test_name, generated);
```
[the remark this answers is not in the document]
TimDiekmann: It's a snapshot test. This means, a snapshot is created. That snapshot counts as test.

libs/@local/codegen/src/definitions/fields.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, Clone)]
+pub enum Fields {
+    Named {
+        fields: Vec<(Cow<'static, str>, Field)>,
+    },
+    Unnamed {
+        fields: Vec<Field>,
+    },
+    Unit,
+}
```
indietyp, line 39: non-blocking, it might make sense to instead replicate how rustc handles and names this:

https://github.com/rust-lang/rust/blob/efcbb94dccbe0fcc1076f031f88332db4a6a2a52/compiler/rustc_hir/src/hir.rs#L3877-L3896

https://github.com/rust-lang/rust/blob/efcbb94dccbe0fcc1076f031f88332db4a6a2a52/compiler/rustc_hir/src/hir.rs#L3979-L3993

https://github.com/rust-lang/rust/blob/efcbb94dccbe0fcc1076f031f88332db4a6a2a52/compiler/rustc_hir/src/hir.rs#L4294-L4297

libs/@local/codegen/src/definitions/primitive.rs
```diff
@@ -0,0 +1,32 @@
+use specta::datatype;
+
+#[derive(Debug, Clone)]
+pub enum Primitive {
+    String,
+    Number,
```
indietyp, line 6: I appreciate that for now the type here is the same, but would it make sense to (in the future) to refine this to e.g. `Integer` and `Number` (with signed and width) in particular as validators like zod, arktype or effect allow to enforce this from the typescript side and we're loosing quite some information here.

Additionally anything over the width of 53 (for integers) should become a BigInt, as anything above that turns into a lossy float.

libs/@local/codegen/src/definitions/type.rs
```diff
@@ -0,0 +1,60 @@
+use alloc::borrow::Cow;
+
+use specta::datatype;
+
+use super::{Enum, Primitive};
+
+#[derive(Debug, Clone)]
+pub enum Type {
+    Reference(Cow<'static, str>),
```
indietyp, line 9: can we guarantee that the name is unique here? What about cases of:

```
struct X { .. }

fn foo() {
	struct X { .. }
}
```

I believe that these will have the same name, so there'd be some ambiguity here in resolving the type.

I think it'd make sense to instead rely on our own `TypeId` that's just a monotonically increasing counter and maps to a SID/Facet ID.

libs/@local/codegen/src/typescript.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    #[must_use]
+    pub fn generate(&self, name: &str) -> String {
+        Codegen::new()
+            .build(
+                &self.ast.program(
+                    SPAN,
+                    ast::SourceType::d_ts(),
+                    "test",
```
indietyp, line 47: nit: we're not in a test environment here, so maybe a name like: `unnamed` or similar, might be more appropriate, or even a user supplied name

libs/@local/codegen/src/typescript.rs
```diff
@@ -0,0 +1,284 @@
...
+                    self.ast.vec1(
+                        self.visit_type_definition(name, &self.collection.types[name].1)
+                            .into(),
+                    ),
+                ),
+            )
+            .code
+    }
+
+    fn visit_type_definition(
```
indietyp, line 60: Do you think it'd make sense (in the future) to move this into a dedicated visitor, a lá https://github.com/hashintel/hash/blob/c041062cda2d075110b434c1db8801c4d6092b0f/libs/%40local/hashql/ast/src/visit.rs#L126 instead?

libs/@local/codegen/tests/codegen/snapshots/codegen__AdjacentlyTaggedEnum::Typescript.snap
```diff
@@ -0,0 +1,30 @@
...
+	content: []
+} | {
+	type: "SingleUnnamed"
+	content: number
+} | {
+	type: "DoubleUnnamed"
+	content: [boolean, string]
+} | {
+	type: "EmptyNamed"
+	content: {}
```
indietyp, line 18: This might be incorrect actually, this might need to be `Record<string, never>` instead, `{}` admits *any* record in TypeScript. See: https://github.com/microsoft/TypeScript/wiki/FAQ#--does-not-refer-to-objects-with-no-properties

libs/@local/hashql/core/Cargo.toml
```diff
@@ -14,7 +14,7 @@ hashql-diagnostics = { workspace = true, public = true }
 anstyle            = { workspace = true, public = true }
 ecow               = { workspace = true, public = true }
 foldhash           = { workspace = true, public = true }
-hashbrown          = { workspace = true, features = ["nightly"], public = true }
```
indietyp, line 17: Why the removal of `nightly` here?

indietyp (changes requested):
Some minor comments - nothing blocking. One major flaw (see snapshot), otherwise looks good to me. Thank you!

libs/@local/codegen/src/typescript.rs
TimDiekmann: That's over-engineering. It does not server any purpose to move it into a trait and add all kinds of different return types just to create a trait. It would create a trait just to have. trait which would be implemented exactly once. We can always add one later.
The goal is to get it working and improve over time, not come up with a pattern which is future proof long-term. We currently don't have any concrete plans to add more visitor implementations in the near future. A lot more things need to be considered before.

libs/@local/hashql/core/Cargo.toml
TimDiekmann: I moved the flag to the global `Cargo.toml`. If we would use `hashql` with another crate using `allocator-api2` indirectly we would get a compile error if the other crate does not take `allocator_api` into account. This globally enforces this.

libs/@local/codegen/src/definitions/type.rs
TimDiekmann: Yes, that's going to be done in a follow up. The corresponding ticket is H-4486. For now I will just wrap `SID` and this can be replaced later by the facet ID.

libs/@local/codegen/src/definitions/primitive.rs
TimDiekmann: This can be changed if we actually construct these types in TypeScript. But for now, `number` is the only type we have in TS. See H-3988

libs/@local/codegen/src/definitions/fields.rs
TimDiekmann: I think renaming `Field` makes sense, but I would incorporate this in a follow-up for simplicity (as this does not really hurt to have now).

libs/@local/codegen/tests/codegen/snapshots/codegen__AdjacentlyTaggedEnum::Typescript.snap
TimDiekmann: It's not incorrect, just not ideal. `{}` is a valid type for an empty map. I agree that we could add an edge case here, but I'd prefer to do this in a follow up: H-4487. It's not an important thing because it does not happen regularly. But to use it, the type _must_ be specified with `deny_unknown_fields`, otherwise, `Record<string, never>` would be wrong

libs/@local/codegen/src/typescript.rs
TimDiekmann:
```suggestion
                    "unnamed",
```

indietyp (approved):

indietyp (approved):
now with dep-graph™

libs/@local/codegen/src/typescript.rs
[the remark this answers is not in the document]
TimDiekmann: Thanks for the suggestion, but I’m going to leave this as a single match. Each arm here is a one-line delegation or a trivial AST call—there’s no multi-step logic hiding in any of them. Pulling them all into separate functions would scatter the core dispatch in seven different places, making it harder to see at a glance which Type variants we support.

If we hit a genuinely complex arm (for example one that builds up multiple fields or invokes several helper methods), I’m happy to extract it. But until then, this match is both exhaustive—and immediately clear—exactly where it needs to be.

PR #7127, opened 2025-05-10
H-4542: Move policy routes to permissions module and clean up actor resource

TimDiekmann:

With newly added policy and principal handling, the REST API requires some cleanup as they got merged previously in existing resources.

## 🚫 Blocked by

- #7121 

## 🔍 What does this change?

- Move principal and policy related endpoints to other resources. No logic changes involved except renaming.

This PR:

The changes in this PR:

The changes in this PR:


godofcodeeyes:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

libs/@local/graph/api/src/rest/permissions.rs
```diff
@@ -0,0 +1,237 @@
...
+
+impl PermissionResource {
+    /// Create routes for interacting with actors.
+    pub(crate) fn routes<S, A>() -> Router
+    where
+        S: StorePool + Send + Sync + 'static,
+        A: AuthorizationApiPool + Send + Sync + 'static,
+        for<'p, 'a> S::Store<'p, A::Api<'a>>: PolicyStore,
+    {
+        // TODO: The URL format here is preliminary and will have to change.
```
indietyp, line 49: nb: should this have a linear issue so that we don't forget?

libs/@local/graph/api/src/rest/principal.rs
```diff
@@ -0,0 +1,668 @@
...
+
+impl PrincipalResource {
+    /// Create routes for interacting with actors.
+    pub(crate) fn routes<S, A>() -> Router
+    where
+        S: StorePool + Send + Sync + 'static,
+        A: AuthorizationApiPool + Send + Sync + 'static,
+        for<'p, 'a> S::Store<'p, A::Api<'a>>: PrincipalStore,
+    {
+        // TODO: The URL format here is preliminary and will have to change.
```
indietyp, line 92: same here

indietyp (approved):
1 minor non-blocking note

libs/@local/graph/api/src/rest/permissions.rs
TimDiekmann: It's unlikely that we would do that before HaRPC lands, so I wouldn't worry about it.

PR #7145, opened 2025-05-11
H-4536: HashQL: Implement HIR Visitor traits and infrastructor

indietyp:

Implements two complementary traits for working with the HIR:

* `Visitor` - For immutable traversal of the HIR tree, useful for analysis, linting, and type-checking
* `Fold` - For transforming the HIR tree while preserving node identity where possible

### Visitor

The `Visitor` trait enables immutable traversal of the HIR tree, similar to rustc's [`intravisit`](https://github.com/rust-lang/rust/blob/dbab4e152bd9eed57ccfc096ef6060dd9af45a99/compiler/rustc_hir/src/intravisit.rs#L177) module. It's designed for:

* Linting and static analysis
* Type checking
* Collecting information from the HIR without modifying it

### Fold

The `Fold` trait supports tree transformations by creating new nodes as needed, inspired by [swc_visit](https://docs.rs/swc_visit/latest/swc_visit/). Key features:

* Uses copy-on-write semantics to minimize allocations
* Preserves sharing between unmodified subtrees
* Particularly useful for optimizations like dead-code elimination, alias propagation, and inlining

### Beef

This PR introduces a `Beef` type, which is optimized for cases where modifications rarely occur:

* Benchmarked against alternative implementations for our use case
* Provides efficient copy-on-write semantics, but unlike `Cow` does so on an per-element basis
* Minimizes memory allocations during tree transformations

### Interned HIR

HIR nodes are interned to automatically deduplicate identical subtrees, which:

1. Reduces memory usage
2. Speeds up equality comparisons
3. Allows efficient structural sharing during transformations

#### Example: Inlining Let Literals

Given this HIR structure:

```mermaid
flowchart TD
    a["let x"]
    a -->|value| e["2"]
    a -->|body| b["\+"]
    b -->|left| c["x"]
    b -->|right| d["x"]
```

After transformation with our `Fold` implementation:

```mermaid
flowchart TD
    e["2"]
    b["\+"]
    b -->|left| e
    b -->|right| e
```

The transformation creates a new tree with shared references to the original literal node, without modifying any existing nodes, and automatically deduplicating the duplicate reference.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/hir/src/fold/beef.rs
+++ b/libs/@local/hashql/hir/src/fold/beef.rs
@@ -1,0 +1,566 @@
+//! This module provides a copy-on-write slice implementation optimized for interned data.
+//!
+//! The primary type is [`Beef`], which allows efficient element-wise modification
+//! of interned slices by only copying the underlying data when a modification
+//! actually occurs.
+//!
+//! The name "Beef" is a wordplay - it's a "slice" of a "`CoW`" (Copy-on-Write).
+
+use core::{hash::Hash, ops::Try};
+
+use hashql_core::{
+    collection::SmallVec,
+    intern::{InternSet, Interned},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum BeefData<'heap, T> {
+    Owned(SmallVec<T>),
+    Interned(Interned<'heap, [T]>),
+}
+
+/// A Copy-on-Write slice optimized for element-wise operations on interned data.
+///
+/// `Beef` (a play on `CoW` - Copy on Write - as in a "slice of `CoW`") provides efficient
+/// element-wise transformation capabilities while minimizing copying. It only
+/// transitions from a shared interned state to an owned state when an actual
+/// modification occurs.
+///
+/// This is particularly useful in scenarios in which no modifications occur, allowing for zero-copy
+/// operations. For comparison, given a `SmallVec<T>` as comparison, `Beef` is consistently about
+/// 25x faster, whenever no modifications occur, with a negligible overhead in case of modification
+/// (5-10%). The overhead is only present due to the fact that we have branching logic to transition
+/// between the stages, which is negligible in most cases.
+///
+/// This is particularly useful for transformation passes that often don't modify
+/// every element, allowing zero-copy operations when no changes are needed.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Beef<'heap, T>(BeefData<'heap, T>);
+
+impl<'heap, T> Beef<'heap, T>
+where
+    T: Copy + Eq + Hash,
+{
+    /// Creates a new [`Beef`] instance from an interned slice.
+    ///
+    /// This is a zero-cost operation that simply wraps the interned slice in the [`Beef`] container
+    /// without any copying.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use hashql_core::intern::{InternSet, Interned};
+    /// # use hashql_core::heap::Heap;
+    /// # use hashql_hir::fold::beef::Beef;
+    /// # let heap = Heap::new();
+    /// # let interner = InternSet::new(&heap);
+    /// let interned = interner.intern_slice(&[1, 2, 3]);
+    /// let beef = Beef::new(interned);
+    /// ```
+    #[must_use]
+    pub const fn new(slice: Interned<'heap, [T]>) -> Self {
+        Self(BeefData::Interned(slice))
+    }
+
+    /// Maps each element in the slice using the provided closure.
+    ///
+    /// This method applies a transformation function to each element. The key
+    /// optimization is that it only transitions to an owned state when an actual
+    /// modification occurs (when the mapped value differs from the original).
+    ///
+    /// The implementation is optimized to:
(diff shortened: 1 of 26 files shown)


libs/@local/hashql/hir/src/fold/beef.rs
```diff
@@ -0,0 +1,588 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum BeefData<'heap, T> {
+    Owned(SmallVec<T>),
+    Interned(Interned<'heap, [T]>),
+}
+
+/// A Copy-on-Write slice optimized for element-wise operations on interned data.
+///
+/// [`Beef`] (a play on [`Cow`] - Copy on Write - as in a "slice of [`Cow`]") provides efficient
```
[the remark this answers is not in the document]
indietyp: ?

libs/@local/hashql/hir/src/fold/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+        &mut self,
+        signature: ClosureSignature<'heap>,
+    ) -> Self::Output<ClosureSignature<'heap>> {
+        walk_closure_signature(self, signature)
+    }
+
+    fn fold_graph(&mut self, graph: Graph<'heap>) -> Self::Output<Graph<'heap>> {
+        walk_graph(self, graph)
+    }
+}
```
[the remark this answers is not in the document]
indietyp: ?

libs/@local/hashql/hir/src/fold/beef.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// # use hashql_hir::fold::beef::Beef;
+    /// # let heap = Heap::new();
+    /// # let interner = InternSet::new(&heap);
+    /// let interned = interner.intern_slice(&[1, 2, 3]);
+    /// let beef = Beef::new(interned);
+    /// ```
+    #[must_use]
+    pub const fn new(slice: Interned<'heap, [T]>) -> Self {
+        Self(BeefData::Interned(slice))
+    }
```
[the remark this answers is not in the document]
indietyp: No, that'd remove any point of beef

libs/@local/hashql/hir/src/fold/beef.rs
```diff
@@ -0,0 +1,588 @@
...
+/// about 25x faster, whenever no modifications occur, with a negligible overhead in case of
+/// modification (5-10%). The overhead is only present due to the fact that we have branching logic
+/// to transition between the stages, which is negligible in most cases.
+///
+/// This is particularly useful for transformation passes that often don't modify
+/// every element, allowing zero-copy operations when no changes are needed.
+///
+/// [`Cow`]: std::borrow::Cow
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Beef<'heap, T>(BeefData<'heap, T>);
```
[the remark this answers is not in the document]
indietyp: done in [ba655db](https://github.com/hashintel/hash/pull/7145/commits/ba655db01f87bae382b48c58c21af34bc8f2c1f8)

libs/@local/hashql/hir/src/fold/beef.rs
```diff
@@ -0,0 +1,604 @@
+//! This module provides a copy-on-write slice implementation optimized for interned data.
+//!
+//! The primary type is [`Beef`] (a play on words, as it is a slice of [`Cow`]), which allows
+//! efficient element-wise modification of interned slices by only copying the underlying data when
+//! a modification actually occurs.
```
TimDiekmann, line 5: What is the advantage of `Beef<T>` over `Cow<[T]>` or is it basically that but with other underlying data structures? If so, doesn't it make sense to use a custom `T` newtype and implement `ToOwned` on it?

libs/@local/hashql/hir/src/fold/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+pub fn walk_ident<'heap, T: Fold<'heap> + ?Sized>(
```
TimDiekmann, line 292: Is `Fold` dyn-safe?

TimDiekmann (commented):
Couple of questions, but generally looks good.

libs/@local/hashql/hir/src/fold/beef.rs
indietyp: Beef works on CoW on an element basis instead of value basis. We only know if a value changed after it has been computed. So with a Cow we'd either need to have an additional function (eg will change), either implement the functionality we have rn in beef or always use into_owned.

libs/@local/hashql/hir/src/fold/mod.rs
indietyp: Fold is not dyn safe (and doesn't need to be - at least for the foreseeable future ) because of the GAT.

TimDiekmann (approved):

libs/@local/hashql/hir/src/fold/mod.rs
indietyp: In theory we don't but I copy-pasted the visitor code so that was left over. As you've pointed out it doesn't really hurt and makes migration in the future easier I guess^^.

PR #7162, opened 2025-05-16
H-4524, H-4450: HashQL: Introduce `Generic` type variant

indietyp:

The generic type variant is required to support generic arguments in arbitrary positions, like:

```
type Foo<T> = Bar<T, String>
```

as we're otherwise unable to assign e.g. constraints or bind `T` without a body. This is super straight forward, it's basically `Apply` and `Opaque` combined. It is essentially completely transparent, and only has a `GenericArguments` attached.

This is also required for unions, in which a generic is present in either branch:

```
type Foo<T> = Bar<T> | Baz<T>
```

In the current implementation the generic arguments are unique for `Bar` and `Baz`, which makes them completely independent, e.g. turning into:

```
type Foo<T> = Bar<T?1> | Baz<T?2>
```

This is *in theory* fine, but makes it so that (A) generic instantiation through substitutions is harder - we now need to track all permutations, instead of a single one and (B) it makes inference less helpful as both arms cannot influence each other.

Additionally this cleans up the implementation a bit - because `Generic` is now a separate type it allows us to centralize some of the logic, allow for easier simplifications. Like `Apply`, `Generic` will be simplified out in most cases after it has been applied.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/type/kind/generic/apply.rs
```diff
@@ -285,6 +286,14 @@ impl<'heap> Lattice<'heap> for Apply<'heap> {
         env.is_subtype_of(self.kind.base, supertype.kind.base)
     }
 
+    fn is_equivalent(
+        self: Type<'heap, Self>,
+        other: Type<'heap, Self>,
+        env: &mut AnalysisEnvironment<'_, 'heap>,
+    ) -> bool {
+        env.is_equivalent(self.kind.base, other.kind.base)
```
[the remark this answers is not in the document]
indietyp: Equivalence is only based on the base type, not the substitutions that are present. These are only used during inference / at a later stage. This has already been the case in the previous implementation. There's an argument to be made that these could be taken into account, but this would then potentially clash in case of simplification - we can only simplify the type out if we ensure the invariant that the `Generic` type is the same as the underlying wrapped type. This is not the case if we take the substitutions into account for equivalence, same with `is_subtype_of`. Considering that these are just taken into account at inference this should be fine.

libs/@local/hashql/core/src/type/environment/mod.rs
```diff
@@ -111,13 +111,39 @@ impl<'heap> Environment<'heap> {
...
+        arguments.sort_unstable();
         // Unlike `intern_struct_fields`, where we error out on duplicates, we simply remove them
         // here, as any duplicate means they're the same argument and therefore not necessarily an
         // error.
-        let (dedupe, _) = arguments.partition_dedup_by_key(|argument| argument.id);
+        let (dedupe, _) = arguments.partition_dedup();
+
+        // If there are any `None` constraints, when `Some` constraints with the same id are
+        // present, remove them. This is relatively easy, as we know there's only ever a *single*
+        // `None` value out there, which is immediately followed by a `Some` value of the same id.
```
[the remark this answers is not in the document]
indietyp: There are already tests for this in Generic

libs/@local/hashql/core/src/type/kind/generic/mod.rs
```diff
@@ -23,8 +36,8 @@ newtype!(
 
 newtype_producer!(pub struct GenericArgumentIdProducer(GenericArgumentId));
 
-// The name is stored in the environment, to allow for `!Drop`
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+// PartialEq, Eq, PartialOrd, Ord, Hash are derived, with `name` ignored
```
TimDiekmann, line 39: I see why you do this, however, I'm not particularly a fan of this. To me, equality in 99% of the case means structural equality. Leaving out a field for comparison may result in really unexpected behavior. 
Do you see a chance in this being changed?
For equality, we could implement an `equiv(&self, &Self) -> bool` or something. For `Hash`, we might want to implement it on another type only consisting of `id` and `constraint` and use that as Hash key.

indietyp: I have had the same thought. Primarily considered these because of the `partition_dedupe` and `sort_unstable` calls when creating the types. Maybe it'd make sense to create a new struct, call it: `AnonymousGenericArgument` and have a function like `as_anonymous` to create it. Considering that we're only shuffling around ~12 bytes (4 bytes `GenericArgumentId`, 4 bytes `TypeId`, 4 bytes overhead because of `Option`) this should be fine.

What do you think?

TimDiekmann: Yes, let's go for that

indietyp: done in [b48290b](https://github.com/hashintel/hash/pull/7162/commits/b48290b51764a50aef9bbf15b9544fd2212063e8)

TimDiekmann: Thank you!

PR #7167, opened 2025-05-16
H-4578: Rework Rust crate coverage

indietyp:

Currently, most of the crates don't actually upload coverage information to codecov, this PR changes it by creating a mise task that can be used by any crate to do proper coverage, as well as centralizing configuration.

> There are several crates I've chosen not to convert yet, notably I needed to rollback converting both: `hash-graph-authorization` and `hash-graph-postgres-store`. This is left for a follow-up PR as it requires some more plumbing. Tracked in https://linear.app/hash/issue/H-4579/instrument-hash-graph-authorization-and-hash-graph-postgres-store

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.config/mise/tasks/test/unit.sh
+++ b/.config/mise/tasks/test/unit.sh
@@ -62,5 +62,5 @@ if [[ $POWERSET == "true" || ${TEST_POWERSET:-false} == 'true' || ${TEST_POWERSE
     exit 0
 fi
 
-cargo nextest run -p "$CRATE" $ARGUMENTS
+cargo nextest run -p "$CRATE" --all-targets $ARGUMENTS
 cargo test --all-features --doc -p "$CRATE"
--- a/.config/mise/tasks/test/unit.sh
+++ b/.config/mise/tasks/test/unit.sh
@@ -23,6 +23,10 @@ CRATE=${PACKAGE#*@rust/}
 
 # Run coverage instead of unit tests if `TEST_COVERAGE` is set to `true` or `1`
 if [[ $COVERAGE == "true" || ${TEST_COVERAGE:-false} == 'true' || ${TEST_COVERAGE:-false} == '1' ]]; then
+    # Exclude any crates that are not the current crate, this is required due to a limitation of llvm-cov, which doesn't allow
+    # for testing the current crate. Any dependency (that is part of the workspace) will be covered as well, tanking coverage.
+    # This is obviously not what's intended in 99% of the cases, as we want to test the coverage of the current crate only, and not it's coverage in any dependencies.
+    # See https://github.com/taiki-e/cargo-llvm-cov/issues/361 for more information on this issue.
     EXCLUSIONS=$(
         cargo metadata --format-version=1 --no-deps \
             | jq -r --arg crate "$CRATE" '.packages[] | select(.name != $crate) | .manifest_path | rtrimstr("/Cargo.toml") | gsub("/"; "\\/"; "g")' \
--- a/.config/mise/tasks/test/unit.sh
+++ b/.config/mise/tasks/test/unit.sh
@@ -7,13 +7,13 @@ set -euo pipefail
 #USAGE arg "<package>" help="The package to run unit tests for"
 #USAGE arg "[arguments]..." double_dash="required" default="" help="Additional arguments to pass to the test runner"
 
-PACKAGE=$usage_package;
+PACKAGE=$usage_package
 COVERAGE=$usage_coverage
 POWERSET=$usage_powerset
 ARGUMENTS=$usage_arguments
 
 # Check if the package argument starts with `@rust/` if that isn't the case exit out
-if [[ "$PACKAGE" != "@rust/"* ]]; then
+if [[ $PACKAGE != "@rust/"* ]]; then
     echo "Error: Only rust crates are supported for now"
     exit 1
 fi
@@ -22,13 +22,15 @@ fi
 CRATE=${usage_package#*@rust/}
 
 # Run coverage instead of unit tests if `TEST_COVERAGE` is set to `true` or `1`
-if [[ "$COVERAGE" = "true" || "${TEST_COVERAGE:-false}" = 'true' || "${TEST_COVERAGE:-false}" = '1' ]]; then
-    EXCLUSIONS=$(cargo metadata --format-version=1 --no-deps | \
-        jq -r --arg crate "$CRATE" '.packages[] | select(.name != $crate) | .manifest_path | rtrimstr("/Cargo.toml") | gsub("/"; "\\/"; "g")' | \
-        paste -sd "|" -)
+if [[ $COVERAGE == "true" || ${TEST_COVERAGE:-false} == 'true' || ${TEST_COVERAGE:-false} == '1' ]]; then
+    EXCLUSIONS=$(
+        cargo metadata --format-version=1 --no-deps \
+            | jq -r --arg crate "$CRATE" '.packages[] | select(.name != $crate) | .manifest_path | rtrimstr("/Cargo.toml") | gsub("/"; "\\/"; "g")' \
+            | paste -sd "|" -
+    )
 
     # under CI we use LCOV
-    if [[ "${CI:-0}" = "1" ]]; then
+    if [[ ${CI:-0} == "1" ]]; then
         RENDER="--lcov --output-path lcov.info"
     else
         RENDER="--html --open"
@@ -42,7 +44,7 @@ if [[ "$COVERAGE" = "true" || "${TEST_COVERAGE:-false}" = 'true' || "${TEST_COVE
 fi
 
 # Run unit tests with powerset if `TEST_POWERSET` is set to `true` or `1`
-if [[ "$POWERSET" = "true" || "${TEST_POWERSET:-false}" = 'true' || "${TEST_POWERSET:-false}" = '1' ]]; then
+if [[ $POWERSET == "true" || ${TEST_POWERSET:-false} == 'true' || ${TEST_POWERSET:-false} == '1' ]]; then
     cargo hack --optional-deps --feature-powerset nextest run -p $CRATE $ARGUMENTS
     cargo test --all-features --doc -p $CRATE
     exit 0
@@ -51,7 +53,7 @@ fi
 LOGFILE=$(mktemp)
 trap 'rm -f "$LOGFILE"' EXIT
 
-cargo test --all-features --doc -p $CRATE >"$LOGFILE" 2>&1  &
+cargo test --all-features --doc -p $CRATE >"$LOGFILE" 2>&1 &
 DOC_PID=$!
 
 cargo nextest run -p $CRATE $ARGUMENTS
--- a/.config/mise/tasks/test/unit.sh
+++ b/.config/mise/tasks/test/unit.sh
@@ -19,7 +19,7 @@ if [[ $PACKAGE != "@rust/"* ]]; then
 fi
 
 # Remove the package namespace from the package to get the crate name
-CRATE=${usage_package#*@rust/}
+CRATE=${PACKAGE#*@rust/}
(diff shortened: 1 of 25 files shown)


.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+    exit 1
+fi
+
+# Remove the package namespace from the package to get the crate name
+CRATE=${usage_package#*@rust/}
+
+# Run coverage instead of unit tests if `TEST_COVERAGE` is set to `true` or `1`
+if [[ $COVERAGE == "true" || ${TEST_COVERAGE:-false} == 'true' || ${TEST_COVERAGE:-false} == '1' ]]; then
+    EXCLUSIONS=$(
+        cargo metadata --format-version=1 --no-deps \
```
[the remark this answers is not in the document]
indietyp: jq is provided via mise

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+    # under CI we use LCOV
+    if [[ ${CI:-0} == "1" || ${CI:-0} == "true" ]]; then
+        RENDER="--lcov --output-path lcov.info"
+    else
+        RENDER="--html --open"
+    fi
+
+    cargo llvm-cov clean --workspace
+    cargo llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-report nextest --all-features --all-targets --cargo-profile coverage $ARGUMENTS
+    cargo llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-clean $RENDER --doctests test --all-features --profile coverage --doc
```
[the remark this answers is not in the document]
indietyp: This is intended.

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+
+    # under CI we use LCOV
+    if [[ ${CI:-0} == "1" || ${CI:-0} == "true" ]]; then
+        RENDER="--lcov --output-path lcov.info"
+    else
+        RENDER="--html --open"
+    fi
+
+    cargo llvm-cov clean --workspace
+    cargo llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-report nextest --all-features --all-targets --cargo-profile coverage $ARGUMENTS
```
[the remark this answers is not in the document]
indietyp: This is intended

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+    cargo llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-clean $RENDER --doctests test --all-features --profile coverage --doc
+
+    exit 0
+fi
+
+# Run unit tests with powerset if `TEST_POWERSET` is set to `true` or `1`
+if [[ $POWERSET == "true" || ${TEST_POWERSET:-false} == 'true' || ${TEST_POWERSET:-false} == '1' ]]; then
+    cargo hack --optional-deps --feature-powerset nextest run -p $CRATE $ARGUMENTS
```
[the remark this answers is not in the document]
indietyp: This is intended

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+    exit 0
+fi
+
+LOGFILE=$(mktemp)
+trap 'rm -f "$LOGFILE"' EXIT
+
+cargo test --all-features --doc -p $CRATE >"$LOGFILE" 2>&1 &
+DOC_PID=$!
+
+cargo nextest run -p $CRATE $ARGUMENTS
```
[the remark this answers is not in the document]
indietyp: this is intended

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,70 @@
...
+ARGUMENTS=$usage_arguments
+
+# Check if the package argument starts with `@rust/` if that isn't the case exit out
+if [[ $PACKAGE != "@rust/"* ]]; then
+    echo "Error: Only rust crates are supported for now"
+    exit 1
+fi
+
+# Remove the package namespace from the package to get the crate name
+CRATE=${usage_package#*@rust/}
```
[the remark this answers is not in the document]
indietyp: done in [7fa0c3b](https://github.com/hashintel/hash/pull/7167/commits/7fa0c3bc07b568d7aab777d19ce05e685428f70e)

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,78 @@
...
+if [[ $POWERSET == "true" || ${TEST_POWERSET:-false} == 'true' || ${TEST_POWERSET:-false} == '1' ]]; then
+    cargo hack --optional-deps --feature-powerset nextest run -p "$CRATE" $ARGUMENTS
+    cargo test --all-features --doc -p "$CRATE"
+    exit 0
+fi
+
+LOGFILE=$(mktemp)
+trap 'rm -f "$LOGFILE"' EXIT
+
+cargo test --all-features --doc -p "$CRATE" >"$LOGFILE" 2>&1 &
```
TimDiekmann, line 64: I appreciate why you do that, I guess because doc-tests are used to be slow as they compile per test. But this is likely conflicting with the other test invocation, so the majority of time they'd block each other and since recently, doc tests are only slow if they fail. If they pass, they are compiled in a single binary, which greatly increases execution time.

Let's stick to the simple approach and just call doc-tests after normal tests

TimDiekmann (changes requested):

.config/mise/tasks/test/unit.sh
indietyp: compile-fail doctests (which hashql-core has) are still quite slow because of it (which is what initially motivated this approach) but you're right in that it doesn't make sense to overcomplicate the script because of it. done in [a670d59](https://github.com/hashintel/hash/pull/7167/commits/a670d59ce22415b57105182c601a51929b109c07)

TimDiekmann: I actually did not t hink of compile-fail doc-tests. Maybe, it's possible to split these later into two profiles or so, but nothing we need to worry about now I guess. When running low-level tests such as `hashql-core` we need to test downstream packages anyway, which takes way longer to run (such as integration tests). Thank you!

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "Error: Only rust crates are supported for now"
+    exit 1
+fi
+
+# Remove the package namespace from the package to get the crate name
+CRATE=${PACKAGE#*@rust/}
+
+# Run coverage instead of unit tests if `TEST_COVERAGE` is set to `true` or `1`
+if [[ $COVERAGE == "true" || ${TEST_COVERAGE:-false} == 'true' || ${TEST_COVERAGE:-false} == '1' ]]; then
+    EXCLUSIONS=$(
```
TimDiekmann, line 26: Could you add a very brief comment what you exclude? I think it's basically every non-related crate?

TimDiekmann (dismissed):
One remaining non-blocking comment. Thanks for looking at this!

TimDiekmann (approved):

.config/mise/tasks/test/unit.sh
indietyp: I've added an extensive comment in [257738c](https://github.com/hashintel/hash/pull/7167/commits/257738c33b551896800d7d27888f8856325d8b15) I hope this is sufficient.

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,66 @@
...
+    cargo llvm-cov clean --workspace
+    cargo $CARGO_COMMAND llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-report nextest $NEXTEST_ARGS --all-targets --cargo-profile coverage $ARGUMENTS
+    cargo llvm-cov --ignore-filename-regex "$EXCLUSIONS" -p "$CRATE" --branch --no-clean $REPORT_FLAGS --doctests test --all-features --profile coverage --doc
+
+    exit 0
+fi
+
+# Run unit tests with powerset if `TEST_POWERSET` is set to `true` or `1`
+if [[ $POWERSET == "true" || ${TEST_POWERSET:-false} == 'true' || ${TEST_POWERSET:-false} == '1' ]]; then
+    cargo hack --optional-deps --feature-powerset nextest run -p "$CRATE" $ARGUMENTS
```
TimDiekmann, line 60: Do we need `--all-targets` here as well?

.config/mise/tasks/test/unit.sh
```diff
@@ -0,0 +1,66 @@
...
+    cargo test --all-features --doc -p "$CRATE"
+    exit 0
+fi
```
TimDiekmann, line 63: We could put the doc-test outside the if and put the non-powerset call in the `else`. Avoids a duplication and an `exit 0`

TimDiekmann (changes requested):

.config/mise/tasks/test/unit.sh
indietyp: I wanted to keep the script semi-straightforward (well as straightforward as possible) in it's iteration. I plan on centralizing some of the functionality in the follow-up.

.config/mise/tasks/test/unit.sh
indietyp: You're right. Although `--all-targets` is the default, done in [f77c751](https://github.com/hashintel/hash/pull/7167/commits/f77c75120c8236dd57dff96dd2bd55a47921562a)

.config/mise/tasks/test/unit.sh
TimDiekmann: Works for me!

TimDiekmann (approved):

.config/mise/tasks/test/unit.sh
TimDiekmann: While I understand the reasoning, it's not entirely true. We have some integration tests moved out to the `/tests/graph/` folder which are supposed to test quite a few functionality. However, we probably want to move that at some point to a better location and leave that folder for more e2e tests.

indietyp: I've contemplated that as well, but thought that it'd be a mistake to add them for primarily two reasons:

1. purpose - integration tests have the express purpose of checking that the system works end-to-end and *integrates*, they might generate coverage, but they shouldn't be the main contributor to coverage. Isolated unit tests (that are usually a lot more narrow in scope) are the more intentional tool to doing so. Integration tests test "Hey can I actually run this end-to-end, does it work?" while tests in the crate itself have the express purpose of testing just said crate and not if it integrates directly (That's more a subjective and software development philosophy reason, so I am always happy to hear other ideas surrounding this, and ofc that's not a perfect system, like for example how do you generate or test that axum endpoints actually behave correctly)
2. it's just hard to coordinate - the coverage reports for the downstream crates need to be somehow merged together, to be able to generate a complete view we would need to generate a coverage report for every single downstream crate as well, and then use that - that's quite messy, and is basically replicating turbo. I think some(?) of this is handled by codecov, but I am not sure. I know for a fact that we couldn't just rely on turbo for this, because the `--doctests` flag is evil. The `--doctests` llvm-cov run needs to generate the report, if it isn't the one generating the report, any previous results, or it's results will be discarded - I don't know why - but that's just the case (I had a `llvm-cov report` stage before, with a `--no-report` on the doctests as well, and that was... fun to figure out)

TimDiekmann: I completely agree on both points. In particular because e2e test may change over time by adding more complex features which might remove previously combinations of features.

PR #7189, opened 2025-05-19
H-4579: Cleanup `test:unit` command

indietyp:

Enable `test:unit` for `hash-graph-authorization` and `hash-graph-postgres-store`, as these need special nextest flags they currently can't make use of the mise `test:unit` task. Change this by adding the required options to the task.

Additionally cleans up several aspects of the initial script, centralizing most of the handling and enabling new use-cases.

This PR:

The changes in this PR:

The changes in this PR:


libs/error-stack/package.json
```diff
@@ -8,7 +8,8 @@
     "fix:clippy": "just clippy --fix",
     "lint:clippy": "just clippy",
     "test:miri": "just miri",
-    "test:unit": "RUST_BACKTRACE=1 mise run test:unit @rust/error-stack"
+    "test:unit": "RUST_BACKTRACE=1 mise run test:unit @rust/error-stack --test-strategy=extremes",
+    "update-snapshots": "INSTA_FORCE_PASS=1 INSTA_UPDATE=new UPDATE_EXPECT=1 yarn test:unit"
```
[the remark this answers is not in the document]
indietyp: that's intended, so that we don't duplicate the test arguments...

.config/mise/tasks/test/unit.sh
```diff
@@ -21,6 +23,37 @@ fi
...
+        HACK_ARGUMENTS+=("--feature-powerset" "--depth=1" "--group-features" "$FEATURES")
+        ;;
+    "powerset")
+        HACK_ARGUMENTS+=("--optional-deps" "--feature-powerset")
+        ;;
+    *)
+        echo "Error: Invalid strategy '$STRATEGY'"
+        exit 1
+        ;;
+esac
```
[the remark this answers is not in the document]
indietyp: thanks bud, done in [373f376](https://github.com/hashintel/hash/pull/7189/commits/373f3769c1a275cc5ce66317f236fd9d8095dff3)

libs/error-stack/.justfile
TimDiekmann, line 1: Will this be an issue for the `clippy` commands?

TimDiekmann (approved):
Thank you very much for this!

Is there a way to avoid this CI output? If not it's not a real issue, just a nicety
image

libs/error-stack/.justfile
indietyp: I don't think we're actually not relying on `lint:clippy` in CI, the only difference would be that when executing `just lint:clippy` locally it'll use the powerset without the groupings. I can re-add the file though if desired!

indietyp:
re: output - sadly not I need to give it all of the mappings, there's no way around it :/

libs/error-stack/.justfile
indietyp: reverted deletion for clippy in [02f2a9b](https://github.com/hashintel/hash/pull/7189/commits/02f2a9b6ddc04c519f28ac449727bcce011d0466)

TimDiekmann (approved):

PR #7196, opened 2025-05-19
H-4588: Allow policies to be named

TimDiekmann:

We need to update certain policies and matching the pattern is hard, in particular if the policies get more complex.
This adds an optional `name` field to policies. The `name` needs to be unique per principal, so there can only be a single `default-web-creator` for the system-actor principal (but there could be a `default-web-creator` for the public).


CiaranMn (dismissed):

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -911,16 +915,17 @@ where
                 LEFT JOIN policy_action ON policy.id = policy_action.policy_id
                 WHERE policy.id = $1
                 GROUP BY
-                    policy.id, policy.effect, policy.principal_id, policy.principal_type,
-                    policy.actor_type, policy.resource_constraint
+                    policy.id, policy.name, policy.effect, policy.principal_id,
+                    policy.principal_type, policy.actor_type, policy.resource_constraint
```
TimDiekmann, line 919:
```suggestion
                    policy.name, policy.effect, policy.principal_id, policy.principal_type,
                    policy.actor_type, policy.resource_constraint
```

CiaranMn (approved):

PR #7211, opened 2025-05-21
H-4534: HashQL: Reify AST into HIR

indietyp:

Reify the AST into the HIR, additionally it also adds a pretty printing module (so that the compiletests can run) and hoists the previous pretty-print infrastructure from the type module in core into the core so that other modules can make use of it.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/ast/src/lowering/mod.rs
```diff
@@ -98,9 +100,11 @@ pub fn lower<'heap>(
...
 
     (
-        named_types,
-        anonymous_types,
-        closure_signatures,
+        ExtractedTypes {
+            locals: named_types,
+            anonymous: anonymous_types,
+            signatures: closure_signatures,
+        },
```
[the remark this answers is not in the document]
indietyp: what are you talking about?

libs/@local/hashql/core/src/lib.rs
```diff
@@ -13,7 +13,8 @@
     iter_map_windows,
     cold_path,
     type_alias_impl_trait,
-    associated_type_defaults
+    associated_type_defaults,
+    default_field_values
```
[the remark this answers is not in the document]
indietyp: it literally is...

libs/@local/hashql/core/src/type/kind/generic/snapshots/hashql_core__r#type__kind__generic__apply__tests__instantiate_partial.snap
TimDiekmann, line 1: Most snapshot diffs now look better, but in this case I prefer the older one 😅 (non-blocking)

libs/@local/hashql/hir/tests/ui/reify/call-multiple-arguments.jsonc
TimDiekmann, line 1: Do we actually support `string + string`? If not, do we want to use number literals here?

TimDiekmann (dismissed):

libs/@local/hashql/hir/tests/ui/reify/call-multiple-arguments.jsonc
indietyp: We do not, I just needed _something_ here to test multiple args. (The LLM initially tried to use `::std::concat`, which... we don't have, so I substituted it with just `+`), either way I don't really see us supporting variable arguments either way, so this was just supposed to be a validating test case. I can change the name tho!

libs/@local/hashql/core/src/type/kind/generic/snapshots/hashql_core__r#type__kind__generic__apply__tests__instantiate_partial.snap
indietyp: I'll continue to iterate on the output in future PRs, as I am not 100% happy with the output myself (although I feel like we're getting closer to something suitable)  The primary goal of this change was to make everything a bit more consistent and pretty.

It all comes down to setting the correct `line` and `softline` "breakpoints" and groupings, which is a bit tricky to get right, but I didn't want to get an improvement before perfection go to waste :P

TimDiekmann: Don't get me wrong, this is great already! Formatting source code is super annoying.

libs/@local/hashql/hir/tests/ui/reify/call-multiple-arguments.jsonc
TimDiekmann: I don't really mind either way. For the test purpose it's fine to keep it as well.

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #7215, opened 2025-05-21
H-4597: Implement conversion between Partially evaluated conditions and filters

TimDiekmann:

The partially evaluated resource conditions needs to be converted to a Query filter. This implements the conversion.

### DIFF
--- a/libs/@local/graph/store/src/filter/mod.rs
+++ b/libs/@local/graph/store/src/filter/mod.rs
@@ -7,7 +7,10 @@ use std::collections::HashMap;
 
 use derive_where::derive_where;
 use error_stack::{Report, ResultExt as _, bail};
-use hash_graph_authorization::policies::{PartialResourceId, PolicyExpressionTree};
+use hash_graph_authorization::policies::{
+    PartialResourceId, PolicyExpressionTree,
+    evaluation::{PermissionCondition, ResourceAttribute},
+};
 use hash_graph_types::ontology::DataTypeLookup;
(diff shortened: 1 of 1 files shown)


libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -606,6 +609,141 @@ impl TryFrom<PolicyExpressionTree> for Filter<'_, EntityTypeWithMetadata> {
...
+impl TryFrom<PermissionCondition> for Filter<'_, Entity> {
+    type Error = Report<InvalidPermissionCondition>;
+
+    fn try_from(expression: PermissionCondition) -> Result<Self, Self::Error> {
+        match expression {
+            PermissionCondition::All(expressions) => Ok(Self::All(
+                expressions
+                    .into_iter()
+                    .map(Self::try_from)
+                    .collect::<Result<_, _>>()?,
```
indietyp, line 627: non-blocking - but out of curiosity - if you use `try_collect` instead of `collect` - would you still need to use turbofish syntax or could we completely forgo adding the type?

indietyp (approved):

libs/@local/graph/store/src/filter/mod.rs
TimDiekmann: The fish can be omitted with `try_collect`, but I did not use `try_collect` at all, yet, as it's unstable. (Also don't feel against using it, but we should do it in more locations then by searching for `collect::<Result` I guess.

PR #7216, opened 2025-05-21
H-4598: Allow multiple filters in a single request in the Postgres query builder

TimDiekmann:

When policies are partially evaluated, we want to add that as a dedicated filters.

## 🔍 What does this change?

- Changes the signatures of the `read`-functions to take a slice instead of a simple reference (an iterator would also work, but that makes the signature even more complex, so I chose to use a simple slice here for now)
- Put every call in a slice instead and remove the filter if it's an `all([])`-filter

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/crud.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/crud.rs
@@ -74,14 +74,14 @@ where
     type ReadPaginatedStream =
         impl Stream<Item = Result<Self::QueryResult, Report<QueryError>>> + Send + Sync;
 
-    #[tracing::instrument(level = "info", skip(self, filter, sorting))]
+    #[tracing::instrument(level = "info", skip(self, filters, sorting))]
     #[expect(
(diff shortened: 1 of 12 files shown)


libs/@local/graph/postgres-store/src/store/postgres/crud.rs
```diff
@@ -74,14 +74,14 @@ where
...
-    #[tracing::instrument(level = "info", skip(self, filter, sorting))]
+    #[tracing::instrument(level = "info", skip(self, filters, sorting))]
     #[expect(
         clippy::type_complexity,
         reason = "The complexity comes from the `instrument` macro"
     )]
     async fn read_paginated(
         &self,
-        filter: &Filter<'_, R>,
+        filters: &[Filter<'_, R>],
```
indietyp, line 84: couldn't this be relaxed even further to `impl IntoIterator<Item = Filter<'_, R>`?

indietyp (approved):

PR #7219, opened 2025-05-22
H-4615: Implement initial set of view permissions

TimDiekmann:

This PRs adds several policies to the seeding logic in the graph:

### `createWeb` permission

* `"system-machine-create-web"`: The system actor can create webs

###  `instantiate` permission

* `"authenticated-instantiate"`: Every authenticated actor (users, machines, AIs) can instantiate every entity type, except:
  * hash-instance
  * actor
  * organization

* `"machine-instantiate"`: Every machine can instantiate:
  * actors
  * organization

### `view` permission

* `"public-view-entity"`: Every user (authenticated or not) can see entities of these types
  * hash-instance
  * user
  * organization

* `"authenticated-view-entity"`: Every authenticated actor (users, machines, AIs) can see entities of these types:
  * machine
  * service-feature
  * is-member-of

* `"system-machine-view-entity"`: The system machine actor can see entities of these types
  * prospective-user
  * sync-linear-data-with

* `"default-web-view-entity"`: Every administrator and member of a web can see entities of these types in the corresponding web, except entities of these types:
  * user-secret
  * usage-record

* `"hash-admins-view-entity"`: Every administrator and member of the `hash-instance` team can see entities of these types:
  * incurred-in
  * usage-record
  * prospective-user

## 🔍 What does this change?

- Move policy seeding from Node to the Graph
- Add the above policies
- drive-by: Fix `ARGUMENTS` in `mise test:unit` on bash versions <4

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

The `instantiate` and `createWeb` permissions are tested as before. The `view` permissions are not yet tested.


libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -0,0 +1,327 @@
...
+        }),
+        actions: vec![ActionName::CreateWeb],
+        resource: None,
+    })
+}
+
+pub(crate) fn system_actor_policies(
+    system_machine_actor: ActorId,
+) -> impl Iterator<Item = PolicyCreationParams> {
+    iter::once(system_actor_create_web_policy(system_machine_actor))
```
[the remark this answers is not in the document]
TimDiekmann: You're right, fixed in https://github.com/hashintel/hash/pull/7219/commits/6632070898c1bfa426763e79fff63d791c2fff0b

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
```diff
@@ -1753,40 +1753,79 @@ where
...
+                    .collect::<HashSet<_>>(),
+            );
+        }
+
+        let types_and_parents = entity_type_id_set
+            .iter()
+            .flat_map(|(base, parents)| iter::once(base.clone()).chain(parents.clone()))
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .collect::<Vec<_>>();
```
indietyp, line 1786: > **Big Disclaimer**: This is just curiosity and is not a request to change any of this. This approach is fine and this isn't a path that would likely benefit from such a micro-optimization.

This is totally non-blocking, but I got curious - for our specific use-case - what are the runtime characteristics of this? So I thought might as well test it!
The results are in (and not what I expected) - a dedup over `contains` is the 2nd fastest in cases where <50 elements, with `vec_sort_dedup` being consistently the fastest. I've attached the benchmark plots. The code I've used to benchmark is available here: https://gist.github.com/indietyp/8ac357ea9907a85cd5cf0e91cd70e42d (to make it fair everything is wrapped in a `CluelessIter` which removes the size-hint and therefore removes any specialization from Vec, BTreeSet and HashSet, just like it'd be the case in this code snippet.

I was surprised that `BTreeSet` was faster, than `HashSet` but that's only with the default hasher! With `foldhash` the `HashSet` is indeed faster.

screenshot 2025-05-23T10-19-08
screenshot 2025-05-23T10-15-41
screenshot 2025-05-23T10-31-11

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
```diff
@@ -1753,40 +1753,79 @@ where
...
+                                context: RequestContext::default(),
+                            },
+                            &policy_context,
+                        )
+                        .change_context(QueryError)
+                        .map(|authorized| match authorized {
+                            Authorized::Always => true,
+                            Authorized::Never => false,
+                            Authorized::Partial(partial) => {
+                                tracing::error!(
```
indietyp, line 1816: shouldn't this error out instead of failing silently?

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -0,0 +1,333 @@
...
+    base_url: BaseUrl,
+    max_version: u32,
+) -> impl Iterator<Item = EntityResourceFilter> {
+    let mut base_url = Some(base_url);
+
+    (1..=max_version).filter_map(move |version| {
+        Some(EntityResourceFilter::IsOfType {
+            entity_type: VersionedUrl {
+                base_url: if version < max_version {
+                    base_url.as_ref()?.clone()
```
indietyp, line 49: I think a `unwrap_or(|| unreachable!())` would be more intentional here than a `filter_map`. The value not being present would be a logic/programming error in this specific function and therefore you'd leak such an implementation error to the consumer. I mean the `None` case should never happen, no?

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -0,0 +1,333 @@
...
+            actor: system_machine_actor,
+        }),
+        actions: vec![ActionName::CreateWeb],
+        resource: None,
+    }
+}
+
+fn system_actor_view_entity_policies(
+    system_machine_actor: ActorId,
+) -> impl Iterator<Item = PolicyCreationParams> {
```
indietyp, line 75: why is this an iterator, but the above is not?

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
TimDiekmann: This is actually something which should never happen because we define an ID above. I could make it an `unreachable!` instead

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
TimDiekmann: I expect to add more things here in the future. It's also only called in an iterator context, so I can simply chain them.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
TimDiekmann: Previously, I actually had `(1..)` instead of `(1..=max_version)` but clippy complained about an infinite loop 😂 
Fixed in 45c4f6c370fb2b803f81d32cdd9e8d5f1f7e49e9

TimDiekmann: And c872515ffa556117e8ee92563c39ca853942e7ff

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
TimDiekmann: 757ac38b05dd520a21f360a684fb686abb3b2f7a

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
TimDiekmann: Interesting indeed. Would you be willing to compare `VersionedUrl`s instead `Uuid`s as well? (I didn't care much about the `clone()` calls here because this will be optimized in the future anyway)

indietyp (approved):

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -0,0 +1,332 @@
...
+        resource: Some(ResourceConstraint::Entity(EntityResourceConstraint::Web {
+            web_id: role.web_id,
+            filter: EntityResourceFilter::Not {
+                filter: Box::new(EntityResourceFilter::Any {
+                    filters: create_version_filters(
+                        base_url!("https://hash.ai/@h/types/entity-type/user-secret/"),
+                        1,
+                    )
+                    .chain(create_version_filters(
+                        base_url!("https://hash.ai/@h/types/entity-type/usage-record/"),
```
CiaranMn, line 237: Not super important as we'll make these not entities at some point, but web _admins_ should be able to view these

TimDiekmann: I actually just implemented this in a follow-up 😄

indietyp (approved):

PR #7254, opened 2025-05-26
H-4625: Implement entity-based view-policies

TimDiekmann:

[H-4615](https://linear.app/hash/issue/H-4615/implement-initial-set-of-view-permissions) implements global policies, such as web-based or for the instance admins. There are a few entities, which requires special view permissions. For details, see [H-4442](https://linear.app/hash/issue/H-4442/write-down-exceptions-to-default-authorization-relationships).

## 🔍 What does this change?

- Add view-policies for entity-based permissions for:
	- prospective user
	- user secret
	- service usage
	- Synced linear entities
- Add global policies for Google and Linear integrations
- Incorporate machine-only permissions into authorized-actor policy
- Tweak some parameters to enforce correct actor ID typing

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hash-backend-utils/src/service-usage.ts
```diff
@@ -283,17 +287,27 @@ export const createUsageRecord = async (
...
   for (const additionalViewer of additionalViewers ?? []) {
     entityRelationships.push({
       relation: "viewer",
       subject: {
         kind: "account",
-        subjectId: additionalViewer,
+        subjectId: additionalViewer.id,
       },
     });
+    viewPrincipals.push({ type: "actor", ...additionalViewer });
```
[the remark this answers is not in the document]
TimDiekmann: While technically true, this would be covered by `tsc`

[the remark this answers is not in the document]
TimDiekmann: @CiaranMn what would you prefer?

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
```diff
@@ -177,12 +182,16 @@ export const createUserSecret = async <
     );
   }
 
+  const userSecretEntityUuid = generateUuid() as EntityUuid;
+  const usesUuserSecretEntityUuid = generateUuid() as EntityUuid;
```
CiaranMn:
```suggestion
  const usesUserSecretEntityUuid = generateUuid() as EntityUuid;
```

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
```diff
@@ -194,6 +203,7 @@ export const createUserSecret = async <
     authentication,
     {
       webId: userAccountId as WebId,
+      entityUuid: usesUuserSecretEntityUuid,
```
CiaranMn:
```suggestion
      entityUuid: usesUserSecretEntityUuid,
```

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
```diff
@@ -204,5 +214,35 @@ export const createUserSecret = async <
...
+    {
+      type: "actor",
+      actorType: "user",
+      id: userAccountId,
+    },
+    {
+      type: "actor",
+      actorType: "machine",
+      id: managingBotAccountId,
+    },
```
CiaranMn, line 227: To make a generic policy we don't need to keep repeating for each secret, I think we need:
1. A condition that if a user's accountId matches the `webId` of the secret, they can see it
2. To somehow be able to determine the `managingBotAccountId`... currently they're not creating the secret, so we'd need to change it so that they are creating it (and therefore we can create a similar rule for the `created_by_id`)

apps/hash-api/src/graphql/resolvers/knowledge/user/submit-early-access-form.ts
```diff
@@ -128,6 +137,23 @@ export const submitEarlyAccessFormResolver: ResolverFn<
...
+      actorType: "user",
+      id: user.accountId,
+    },
+    effect: "permit",
+    actions: ["viewEntity"],
+    resource: {
+      type: "entity",
+      id: extractEntityUuidFromEntityId(entity.entityId),
+    },
+  });
```
CiaranMn, line 155: This one we can handle similarly to the secret: allow any `accountId` matching the `webId` of an entity to see a Prospective User.

apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -96,7 +98,25 @@ const createHashEntity = async (params: {
...
+    principal: {
+      type: "actor",
+      actorType: "machine",
+      id: params.authentication.actorId,
+    },
+    actions: ["viewEntity"],
+    resource: {
+      type: "entity",
+      id: entityUuid,
+    },
```
CiaranMn, line 116: This one I think the Graph needs to know about Linear types and the Linear bot's id, to be able to say it can do whatever with Linear types in webs.

apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -96,7 +98,25 @@ const createHashEntity = async (params: {
...
+  const entityUuid = extractEntityUuidFromEntityId(entity.entityId);
+  await createPolicy(graphApiClient, params.authentication, {
+    name: `linear-synced-administer-entity-${entityUuid}`,
+    effect: "permit",
+    principal: {
+      type: "actor",
+      actorType: "machine",
+      id: params.authentication.actorId,
+    },
+    actions: ["viewEntity"],
```
CiaranMn, line 112: I believe this should be whatever the `administrator` equivalent is, as that's what we're adding on top of the web defaults above.

apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -132,10 +152,30 @@ const createHashEntity = async (params: {
...
+    const linkEntityUuid = extractEntityUuidFromEntityId(linkEntity.entityId);
+    await createPolicy(graphApiClient, params.authentication, {
+      name: `linear-synced-administer-entity-${linkEntityUuid}`,
+      effect: "permit",
+      principal: {
+        type: "actor",
+        actorType: "machine",
+        id: params.authentication.actorId,
+      },
+      actions: ["viewEntity"],
```
CiaranMn, line 168: I believe this should be whatever the `administrator` equivalent is, as that's what we're adding on top of the web defaults above.

libs/@local/hash-backend-utils/src/service-usage.ts
```diff
@@ -258,7 +262,7 @@ export const createUsageRecord = async (
       relation: "administrator",
       subject: {
         kind: "accountGroup",
-        subjectId: hashInstanceAdminGroupId,
+        subjectId: hashInstanceAdminGroup.id,
         subjectSet: "member",
       },
```
CiaranMn, line 267: How / where is this preserved?

libs/@local/hash-backend-utils/src/service-usage.ts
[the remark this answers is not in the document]
CiaranMn: The Graph API will reject additional fields so it's something we should be mindful of if with objects that are passed in from elsewhere that might end up with additional fields.

Since this is just an `ActorId` that is unlikely to end up with more stuff in it I think it's fine here.

CiaranMn (commented):
Thanks Tim, a few comments / questions

libs/@local/hash-backend-utils/src/service-usage.ts
TimDiekmann: The instance admin group is known to the graph, so it's encoded there:
https://github.com/hashintel/hash/blob/314207d48a47e514957e35a32b778a183bdda91b/libs/%40local/graph/postgres-store/src/store/postgres/seed_policies.rs#L267-L296

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
TimDiekmann: Captured the additional conditions in H-4699: Support custom principal conditions on policies
Captured the creator in H-4700: Change creator of user-secret entity

The `managingBotAccountId` is the linear system bot, right? Because that one is known to the graph as it's created through `getOrCreateSystemMachine("linear")`

apps/hash-api/src/graphql/resolvers/knowledge/user/submit-early-access-form.ts
TimDiekmann: Captured the additional conditions in H-4699: Support custom principal conditions on policies

libs/@local/hash-backend-utils/src/service-usage.ts
CiaranMn: Ah okay, I think `recordsUsageOf` needs adding to that list then

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
CiaranMn: It could be the Linear bot, it could also be the Google bot. Basically any current and future bot that might need to access a user secret to use some external API. So I think it has to be somehow tied to a field related to the entity

apps/hash-integration-worker/src/linear-activities.ts
TimDiekmann: Yup, that's currently captured under H-4442 (at the very bottom):
image

I tried to somehow capture that to some policies more actions needs to be added in the future.

apps/hash-integration-worker/src/linear-activities.ts
TimDiekmann: The linear bot is known, shall we just grant view-permissions to all entities from `@linear`? The types could be copied over (annoying but it works).

CiaranMn: The Linear bot entity itself is in the same web (`@linear`) as the types, so if you could check that somehow it would save maintaining the list of types (which can change). Would need the `webId` of the bot and the types though

TimDiekmann: If the types are external, the types don't have a `WebId`. This I think will be the case for self-hosted instances?

apps/hash-integration-worker/src/linear-activities.ts
TimDiekmann: Same as [below](https://github.com/hashintel/hash/pull/7254#discussion_r2121623877)

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
TimDiekmann: I wonder if it makes sense to introduce some kind of tags here. But this could over-complicate it as well. We can keep it the way we have it for now as it's easy to remove the policies again later.

apps/hash-integration-worker/src/linear-activities.ts
CiaranMn: Ah that's true, yes. I think maintaining a list of copied types will be quite fragile so we need to consider a different approach. Maybe the type creator/importer? Since it will also be the Linear bot. We probably should also protect system namespaces somehow so other people can't load them as external, or similar

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
CiaranMn: Sounds good

apps/hash-integration-worker/src/linear-activities.ts
TimDiekmann: Yes, that should work. That should be the `createdById`. I guess in both cases that's `@linear`, right?

libs/@local/hash-backend-utils/src/service-usage.ts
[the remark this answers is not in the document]
TimDiekmann: I just tried to manually specify the fields but as the `id` has to match the `actorType`, `tsc` fails without casting to `ActorId` again.

libs/@local/hash-backend-utils/src/service-usage.ts
TimDiekmann: Applied in [`d20332e` (#7254)](https://github.com/hashintel/hash/pull/7254/commits/d20332ee55af5dff462e3a26ffb80ad136b5c340)

apps/hash-integration-worker/src/linear-activities.ts
CiaranMn: Yes, it would be

CiaranMn (approved):

PR #7256, opened 2025-05-26
`error-stack`: AnyReport: easier error propagation for arbitrary errors when no specialised user error needed, bridging the gap with anyhow/eyre error handling patterns

zakstucke:

`error-stack` is a great library, but is somewhat painful to use in comparison to something like `anyhow`, when no specific error is needed in a given context, one has to add `.change_context(AnyErr)?` everywhere from both other `Report<E>` types, and base errors implementing `core::error::Error`.

This PR adds a `struct AnyReport(Report<AnyErr>);` type, which: 
- `impl<E: Into<Report<E>>> From<E> for AnyReport`
- `impl<C: ?Sized> From<Report<C>> for AnyReport`
- `impl IntoReport for AnyReport`
- implements `Deref` and `DerefMut` to `Report<AnyErr>`, allowing access to underlying methods
- reimplements remaining `fn (self, ..)` `Report` methods to provide parity with any other `Report<E>` type
- Exposes the `AnyErr` type and the `AnyReport` from the root of the `error-stack` library.

Imagine this type exactly like a `Report<AnyErr>`, except it can be `?` propagated from `E: Error` / `Report`, without needing `.change_context(AnyErr)`, making the base case of not needing targeted error handling much more ergonomic with `error-stack`.

This `AnyReport` type cannot be implemented in user-land, due to `upstream crates may add a new impl of trait` limitations, nor can these `From<>` impls be added directly to `Report<AnyErr>` due to trait conflicts.

This PR:

The changes in this PR:

  - Happy to add docs once the overall PR's intention is confirmed as wanted 

The changes in this PR:

## 🛡 What tests cover this?

Added some new basic tests confirming `?` propagation.

### DIFF
--- a/libs/error-stack/src/any.rs
+++ b/libs/error-stack/src/any.rs
@@ -1,36 +1,23 @@
 #![expect(deprecated, reason = "We use `Context` to maintain compatibility")]
 use core::{
-    error::Error,
     fmt::{self, Debug, Display},
     ops::{Deref, DerefMut},
+    panic::Location,
 };
 
-use crate::{IntoReport, Report};
+use crate::{Context, IntoReport, Report};
 
-/// The error type used inside an [`AnyReport`]
-/// and the concrete error type used when converted to a `Report<E>`.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub struct AnyError;
-
-impl Display for AnyError {
-    fn fmt(&self, formatter: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(formatter, "AnyError")
-    }
-}
-
-impl Error for AnyError {}
-
-/// A special concrete wrapping type acting as `Report<AnyError>`
+/// A special concrete wrapping type acting as `Report<dyn Error + Send + Sync>`
 /// which can be used as a catch-all error type when no custom error type is needed.
 ///
 /// This type is benefical because it implements `From<E: Into<Report<E>>`,
-/// meaning any error implementing [`std::error::Error`], or is a `Report<E>` can be
+/// meaning any error implementing [`core::error::Error`], or is a `Report<E>` can be
 /// propagated automatically with `?` without requiring `.change_context(AnyError)`.
 ///
-/// This type implements all the same methods the underlying `Report<AnyError>` implements.
+/// This type implements all the same methods the underlying `Report<dyn Error + Send + Sync>`
+/// implements.
 #[must_use]
-pub struct AnyReport(Report<AnyError>);
+pub struct AnyReport(Report<dyn Context + Send + Sync + 'static>);
 
 impl Display for AnyReport {
     fn fmt(&self, formatter: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
@@ -45,7 +32,7 @@ impl Debug for AnyReport {
 }
 
 impl IntoReport for AnyReport {
-    type Context = AnyError;
+    type Context = dyn Context + Send + Sync + 'static;
 
     #[track_caller]
     fn into_report(self) -> Report<Self::Context> {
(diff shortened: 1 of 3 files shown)


TimDiekmann (commented):
Hi @zakstucke and thanks for opening this PR.

I’m trying to understand the motivation for `AnyReport` versus just using `Box<dyn Error>`. Could you clarify a concrete scenario where `AnyReport` provides benefits that `Box<dyn Error>` can’t address?

For comparison, this pattern works without additional wrappers:
```rust
fn x() -> Result<(), Report<Box<dyn Error>>> { /* … */ }
fn main() -> Result<(), Box<dyn Error>> {
    x()?;
    Ok(())
}
```

It propagates both concrete and generic errors seamlessly.

Does this approach cover your use cases adequately, or is there a specific gap that `AnyReport` fills?

Thanks for your insight!

zakstucke:
Hi @TimDiekmann,

The usecase isn't for top-level error handling, but arbitrary catch-all handling at some specififc layer of an application, maybe later mapping to an e.g. `MyError::InternalError`.

This quick example shows `base()`+`outer()` using `AnyReport`, and the `2` variants trying to use `Box<dyn Error>`, this specific example fails at `.change_context()`.

```rust
        use core::error::Error;

        use error_stack::{AnyReport, Report};

        #[derive(Debug)]
        struct OtherErr;

        impl std::fmt::Display for OtherErr {
            fn fmt(&self, formatter: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
                write!(formatter, "OtherErr")
            }
        }

        impl core::error::Error for OtherErr {}

        fn base() -> Result<(), AnyReport> {
            let e_result = Err(std::io::Error::from(std::io::ErrorKind::NotFound));
            e_result?;
            let report_result = Err(OtherErr.into_report());
            report_result?;
            Ok(())
        }

        fn outer() -> Result<(), Report<OtherErr>> {
            base().change_context(OtherErr)?;
            Ok(())
        }

        fn base2() -> Result<(), Box<dyn Error>> {
            let e_result = Err(std::io::Error::from(std::io::ErrorKind::NotFound));
            e_result?;
            let report_result = Err(OtherErr.into_report());
            report_result?;
            Ok(())
        }

        fn outer2() -> Result<(), Report<OtherErr>> {
            base2().change_context(OtherErr)?; // <-- ERROR: the method `change_context` exists for enum...
            Ok(())
        }
```

On top of this example, reasons for a distinct type even if we could bring parity to `Box<dyn Error>`:
- `Box<dyn Error>` is less performant
- We want `AnyReport` to act as if it was a normal `Report<AnyErr>`, it only exists as a wrapper to satisfy trait conflict rules.
- It's a little less confusing to a newcomer, I feel more idiomatic, having `AnyReport` as the catch-all than peppering `Box<dyn Error>` over a codebase

Thanks for your time

indietyp:
👋 Hi! 

From my side, I’m fundamentally opposed to adding a `Report<dyn Error>` or `AnyReport`. Even if it were technically possible, it doesn’t align at all with the core idea behind `error-stack`. The whole point of `error-stack` is explicitly defining transitions between contexts. Introducing an untyped dynamic error variant in the middle of a pipeline breaks this core principle and dilutes what makes error-stack valuable.

Also, dynamic error handling (`Box<dyn Error>` etc.) usually reduces errors to something closer to exceptions: vague and unclear. Rust’s strength is precisely its explicit and meaningful error handling, and dynamic approaches sacrifice exactly this benefit.

I think the real frustration here is understandable - changing error contexts can be verbose (I've experienced this myself a couple of times^^). But a fully dynamic error feels like the wrong solution to this specific problem. I propose that we could instead explore a trait based approach (name still TBD):

```rust
trait TransitionContext<C> {
    fn make() -> Self;
}

impl<T, C> From<T> for Report<C>
where
    T: TransitionContext<C>,
{
	...
}
```

This would let us do simple context transitions (`func(..)?`) in most cases, while still keeping things explicit through the additional trait implementation.

(The context parameter `C` here helps us control trait implementations, limiting it to specific contexts or changing the returned value depending on the context.)

This would initially introduce some additional boilerplate (as we'd need to implement an extra trait), but that could be mitigated with a derive macro.

zakstucke:
Thanks for both your comments @TimDiekmann & @indietyp .

I should preface this by saying I absolutely love `error-stack` and its ideology, I've been daily driving it for a couple of years now. 

> It sounds like you want a catch-all error type you can use mid-pipeline. But error-stack deliberately prevents that: we want real, domain-specific errors at each layer. For top-level convenience you can still convert to Box<dyn Error>, and if you really need an “any” report you can do:

```rust
Report::<AnyErr>::change_context(AnyErr)
```

I should make clear this is the only real reason I'm asking for an `AnyReport`, quite literally just to replace the above with `?` when converting from arbitrary `Report<E>` and raw `E`'s that implement `Error`. A `Report<dyn Error>` would also suffice if `?` could work in the same way and still capture the location of the initial `?` propagation.

The PR boils down to: 
**an anyhow-esq type is useful in some cases, and having one that can interop seamlessly with the rest of error-stack would improve ergonomics substantially over a user-defined `AnyError` type, or using something like anyhow directly in the general context of a project using ES.**

I've been quite okay with the explicitness of `.change_context(AnyErr)` up to this point, but I recently introduced ES to a new project, and the biggest pushback from some of the team has been surrounding the unnecessary pain of similar things to `.change_context(AnyErr)?` when there is no extra error information to capture and just a bunch of `.change_context(AnyErr)?` over a function, using just `?` would be much preferred.

I understand these cases don't fit with ES's ideology, but they do exist in practice, having to leave the ES ecosystem for these catch-all cases and "scratch" work seems unfortunate.

I understand if this is still rejected, I do get the dislike. If that's the case I think I'll publish some sort of fork that does include something like this. For me and I think others, it would be a very useful pattern.

zakstucke:
I managed to remove the internal `AnyError` and replace it with `Report<dyn Context + Send + Sync + 'static>`, which also lead to exploring #7274

zakstucke:
https://github.com/hashintel/hash/pull/7274 has been extended with type extraction and a report default generic, it now supersedes this PR.

TimDiekmann:
Superseded by #7274.

PR #7258, opened 2025-05-26
H-4640: Don't seed the graph system policies from Node

TimDiekmann:

The system policies are now seeded when the graph starts, there is no need to also do that from Node.

### DIFF
--- a/apps/hash-api/src/ensure-system-graph-is-initialized.ts
+++ b/apps/hash-api/src/ensure-system-graph-is-initialized.ts
@@ -15,9 +15,13 @@ const context: ImpureGraphContext<false, true> = {
   },
   graphApi: createGraphClient(logger, {
     host: getRequiredEnv("HASH_GRAPH_HTTP_HOST"),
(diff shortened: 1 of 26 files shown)


CiaranMn (dismissed):

apps/hash-api/src/ensure-system-graph-is-initialized.ts
```diff
@@ -15,9 +15,13 @@ const context: ImpureGraphContext<false, true> = {
   },
   graphApi: createGraphClient(logger, {
     host: getRequiredEnv("HASH_GRAPH_HTTP_HOST"),
-    port: parseInt(getRequiredEnv("HASH_GRAPH_HTTP_PORT"), 10),
+    port: Number.parseInt(getRequiredEnv("HASH_GRAPH_HTTP_PORT"), 10),
```
TimDiekmann, line 18: Probably some autofix?

CiaranMn: Yes, it's fine

PR #7264, opened 2025-05-26
H-4650: Initial implementation of `Subscript` inference

indietyp:

This allows for the discharge and inference of constructs such as:

```
let x: Dict[String, String | Number] = Dict(a = "12", b = 13) in
let y = x["a"] // String | Number | Null
```

There are a bunch of follow-up PRs that finish the implementation across the different types. (`Subscript` is the last operation that's required before we can do type checking in the HIR)

Right now we have it so that the result is `| Null`, but that can be changed very easily.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/type/kind/intrinsic.rs
```diff
@@ -361,6 +394,31 @@ impl<'heap> Lattice<'heap> for DictType {
...
+        // Check if `index` and `key` are concrete, if not collect the constraints between them
+        if !env.is_concrete(index) || !env.is_concrete(self.kind.key) {
+            infer.in_invariant(|infer| {
+                infer.collect_constraints(index, self.kind.key);
+            });
+
+            return Subscript::Pending;
+        }
+
+        // Dict keys are invariant, and therefore the index must be equivalent to the key
```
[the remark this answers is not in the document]
indietyp: Yes, because you cannot check for equivalence if they're not concrete. That'll give you false positives. Additionally we need to discharge additional constraints if they're not concrete.

libs/@local/hashql/core/src/type/inference/solver/test.rs
```diff
@@ -1579,3 +1595,334 @@ fn projection_unify_variables_equal_is_not_equivalent() {
...
+            [
+                primitive!(env, PrimitiveType::String),
+                primitive!(env, PrimitiveType::Null)
+            ]
+        )]
+    );
+}
+
+#[test]
+fn early_subscript() {
```
TimDiekmann, line 1686: Shouldn't this be `late_subscript` and the other one `early_subscript`?

TimDiekmann (approved):

libs/@local/hashql/core/src/type/inference/solver/test.rs
indietyp: The name pertains to the fact that the inference variable has been added to the constraint set *before* we add the subscript. So it's early in the sense that we already know the type at this point, even if we weren't using fix-point evaluation.

TimDiekmann (approved):

Issue #4295, opened 2024-04-10
`ResultExt` doesn't cover `Report<EyreReport>` for some reason


0x009922:
### Describe the bug

For a reason I haven't yet understood, `ResultExt` trait doesn't work on `error_stack::Result<EyreReport>` (e.g. `change_context` is not found), although `EyreReport` seems to satisfy the `Context` trait.

However, it is possible to call `Report::change_context` on it without a problem.

### To reproduce

```Rust
use error_stack::{IntoReportCompat, ResultExt};

fn eyre_result() -> eyre::Result<()> {
    Err(eyre::eyre!("Something happened"))
}

#[derive(Debug, thiserror::Error)]
#[error("my error occurred")]
struct MyError;

fn main() -> error_stack::Result<(), MyError> {
    // doesn't compile
    eyre_result()
        .into_report()
        .change_context(MyError)?;

    // fine
    eyre_result()
        .into_report()
        .map_err(|report| report.change_context(MyError))?;


    Ok(())
}
```

### Expected behavior

Expected `.into_report().change_context(...)` to work

### Rust compiler

1.76.0 (07dca489a 2024-02-04)

### Host

aarch64-apple-darwin

### Target

aarch64-apple-darwin

### Version

0.4.1

### Features

default, eyre

### Additional context

Compiler error:

```
error[E0599]: no method named `change_context` found for enum `Result` in the current scope
   --> src/main.rs:15:10
    |
13  | /     eyre_result()
14  | |         .into_report()
15  | |         .change_context(MyError)?;
    | |         -^^^^^^^^^^^^^^ method not found in `Result<(), Report<Report>>`
    | |_________|
    | 
    |
note: the method `change_context` exists on the type `error_stack::Report<ErrReport>`
   --> [redacted]/error-stack-0.4.1/src/report.rs:462:5
    |
462 | /     pub fn change_context<T>(mut self, context: T) -> Report<T>
463 | |     where
464 | |         T: Context,
    | |___________________^
help: use the `?` operator to extract the `error_stack::Report<ErrReport>` value, propagating a `Result::Err` value to the caller
    |
14  |         .into_report()?
    |                       +


```

TimDiekmann:
Hi @0x009922 and thanks for opening the issue.
I will look into this, it feels wrong that the `map_err` is working but the trait extension on `Result` does not.

TimDiekmann:
I investigated this further:
[...]

PR #7274, opened 2025-05-29
AnyReport alternative, Report<dyn Context> that can be defaulted to with error_stack::Result<T>

zakstucke:
Alternative to #7256 that adds `?` support to `Report<dyn Context + Send + Sync + 'static>` instead of an `AnyReport` wrapper.

This is even more powerful because the `error_stack::Result` type alias can be updated to: 
```rust
pub type Result<T, C = dyn Context + Send + Sync + 'static> = core::result::Result<T, Report<C>>;
```

Which makes switching between specific contexts and a catch all completely painless. `Result<T, E>` can simply remove the `E` to make it catch-all.

This is at the cost of a breaking change, the new `From` impl breaks `Report::from` type inference, meaning some `Report::from` need changing to `Report::new`. #7256 is a breaking change.

### DIFF
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -471,6 +471,27 @@ impl<C> Report<C> {
     }
 }
 
+impl<E: Into<Report<E>>> From<E> for Report<dyn Context + Send + Sync + 'static> {
+    #[track_caller]
+    fn from(value: E) -> Self {
+        value.into().into()
+    }
+}
+
+impl<C> From<Report<C>> for Report<dyn Context + Send + Sync + 'static> {
+    #[track_caller]
+    fn from(value: Report<C>) -> Self {
+        Report {
+            frames: value.frames,
+            _context: PhantomData,
+        }
+        .attach_printable(format!(
+            "into Report<dyn Context> at {}",
+            Location::caller()
+        ))
+    }
+}
+
 impl<C: ?Sized> Report<C> {
     /// Retrieves the current frames of the `Report`, regardless of its current type state.
     ///
--- a/libs/error-stack/tests/test_conversion.rs
+++ b/libs/error-stack/tests/test_conversion.rs
@@ -87,8 +87,8 @@ fn error() {
 }
 
 #[test]
-fn into_report() {
-    let report = io_error().map_err(Report::from).expect_err("not an error");
+fn mapped_report() {
+    let report = io_error().map_err(Report::new).expect_err("not an error");
     assert!(report.contains::<io::Error>());
     assert_eq!(report.current_context().kind(), io::ErrorKind::Other);
 }
--- a/libs/error-stack/src/result.rs
+++ b/libs/error-stack/src/result.rs
@@ -45,7 +45,7 @@ use crate::{Context, IntoReport, Report};
     note = "Use `core::result::Result<T, Report<C>>` instead",
     since = "0.6.0"
 )]
-pub type Result<T, C> = core::result::Result<T, Report<C>>;
+pub type Result<T, C = dyn Context + Send + Sync + 'static> = core::result::Result<T, Report<C>>;
 
 /// Extension trait for [`Result`][core::result::Result] to provide context information on
 /// [`Report`]s.
--- a/libs/error-stack/tests/test_provision.rs
+++ b/libs/error-stack/tests/test_provision.rs
@@ -45,6 +45,6 @@ fn request_context() {
 
 #[test]
 fn context_provision() {
-    let report = Report::from(ContextA(10));
+    let report = Report::new(ContextA(10));
     assert_eq!(report.request_ref::<u32>().count(), 1);
 }


zakstucke:
I have now updated the `Report` struct definition to:
```rust
struct Report<C: ?Sized = dyn Context + Send + Sync + 'static>
```
this makes an "untyped report" a first class citizen and means `Report` and `Report<C>` can be used interchangeably, an untyped `Report` acts like `AnyReport` does in #7256.

The concept of an untyped `Report` feeds nicely into a new feature, owned type extraction:
```rust
impl<C> Report<C> {
    pub fn pop_current_context(self) -> (Report, C)
        where
            C: Send + Sync + 'static

    pub fn into_current_context(self) -> C
        where
            C: Send + Sync + 'static
}

impl<C: ?Sized> Report<C> {
    pub fn downcast<T: Send + Sync + 'static>(self) -> Result<T, Self>

    pub fn downcast_take<T: Send + Sync + 'static>(self) -> Result<(Report, T), Self>
}
```

This addresses the second painpoint of ES, some error types don't implement `Clone`, or cloning is very expensive, despite needing to own them.

This PR now implements these methods, making sure to not affect the printable representation of a report despite extracting types, by eagerly formatting on extraction.

libs/error-stack/src/report.rs
```diff
@@ -452,6 +452,89 @@ impl<C> Report<C> {
...
+    ///     ...
+    ///     # };
+    ///     # fs::read_to_string(path.as_ref()).map_err(Report::from)
+    /// }
+    ///
+    /// let report = read_file("test.txt").unwrap_err();
+    /// let (_report, io_error) = report.pop_current_context();
+    /// assert_eq!(io_error.kind(), io::ErrorKind::NotFound);
+    /// ```
+    pub fn pop_current_context(self) -> (Report, C)
```
TimDiekmann: This is actually an interesting use-case, however, I'd probably revert the return statements.

TimDiekmann (commented):
As mentioned by @vilkinsons, we're going to discuss this -- most likely on Tuesday (I'm currently on vacation). I like `Report<dyn Context>` much more than the artificial error type. But a few comments regardless:

On `Result<T, C = Report<dyn Context>>` and the deprecation: We deprecated `Result` for a quite simple reason: `Result<T, E>` behaves differently as soon as you import `error_stack::Result`. This was quite annoying and -- as you mentioned -- this can be done downstream quite easily. If it would be defined like this
```rust
type Result<T, C = dyn Context> = Result<T, Report<C>>
```
it would still behave differently from `std::result::Result`, so I would not touch `Result`.

On `dyn Context` in general: `Context` is deprecated in favor of `core::error::Error`, which was stabilized a while back. This means, `Context` is not needed at all and we should not build anything on top of it but simply use `dyn Error` here.

On defaulting `C` in `Report<C>` to `dyn Context`: If we would add `dyn Context` as possible target, I don't like us to add a default to it as this encourages omitting `C` in most cases. The user should explicitly opt-in to the generic error. `Report<dyn Context>` (or dyn Error) should only be used if there really is not another way and a named `C` should be used per default.

libs/error-stack/src/report.rs
zakstucke: I agree! Updated

zakstucke:
> ...behaves differently as soon as you import error_stack::Result. This was quite annoying...

I agree, I actually don't use this, but a renamed `ResultStack` version instead that does the same thing, the alias is useful, it just shouldn't overwrite `core::result::Result`. Anyhow, it's scheduled for removal and can be done in userland, I've reverted the change.

> Context is not needed at all and we should not build anything on top of it but simply use dyn Error here.

Agreed, updated!

> I don't like us to add a default to it as this encourages omitting C in most cases. The user should explicitly opt-in to the generic error.

I disagree with this, but understand the stance. I think having a "split"/"untyped" `Report` as a concept is powerful and useful as shown in the `C` extraction methods, I also feel ES can explain in docs and style how it's intended to be used, without introducing hurdles for users that do want to use ES in more flexible ways. Finally like with the alias, it makes migrations easier and more gradual.
Given a user can re-type alias `Report` in userland, I'd be fine with removing this default if it unblocked this PR.

libs/error-stack/src/report.rs
```diff
@@ -243,7 +243,7 @@ use crate::{
 /// ```
 #[must_use]
 #[expect(clippy::field_scoped_visibility_modifiers)]
-pub struct Report<C: ?Sized> {
+pub struct Report<C: ?Sized = dyn Error + Send + Sync + 'static> {
```
TimDiekmann: We don’t want to encourage people to use `dyn Error` as the common case. People should explicitly specify, if a `Report` is untyped. 
```suggestion
pub struct Report<C: ?Sized>
```
Also, we should add a documentation in which scenarios `Report<dyn Error>` is discouraged and when it’s useful:
- in return statements in an `Err` variant is discouraged, as a typed error is more representative. `Report<dyn Error>` does not add any contextual information to the `Report` which is a core principal of `error-stack`
- in argument position (e.g. for error handling) might be useful if the handling code does not need to know the current context. For example if it just searches for a `Suggestion` in the `Report` it doesn’t need to know `C`

The last bullet brings me to another point: We might need to make sure, that `current_context` is not callable on `Report<dyn Error>`. I haven’t looked at this, yet, but it might cause issues. Maybe it also just works.

libs/error-stack/src/report.rs
```diff
@@ -471,6 +554,24 @@ impl<C> Report<C> {
     }
 }
 
+impl<E: Into<Report<E>>> From<E> for Report<dyn Error + Send + Sync + 'static> {
```
TimDiekmann: `dyn Error` is only a marker type, it does not really serve any purpose. In theory, it could be anything, such as `()`. `dyn Error` is descriptive and `Report<C>` always implements `Send + Sync` as it only can be constructed with such a type. It should be enough to just use
```suggestion
impl<E: Into<Report<E>>> From<E> for Report<dyn Error> {
```
This also aligns better with not setting a default for `C` as it’s less to type.

libs/error-stack/src/report.rs
```diff
@@ -471,6 +554,24 @@ impl<C> Report<C> {
...
+    #[track_caller]
+    fn from(value: E) -> Self {
+        value.into().into()
+    }
+}
+
+impl<C> From<Report<C>> for Report<dyn Error + Send + Sync + 'static> {
```
TimDiekmann:
```suggestion
impl<C> From<Report<C>> for Report<dyn Error> {
```

libs/error-stack/src/report.rs
```diff
@@ -471,6 +554,24 @@ impl<C> Report<C> {
...
+    #[track_caller]
+    fn from(value: Report<C>) -> Self {
+        Report {
+            frames: value.frames,
+            _context: PhantomData,
+        }
+        .attach_printable(format!("into untyped Report at {}", Location::caller()))
```
TimDiekmann: It’s a good idea to add some attachment to cover the location, but I don’t know if we want to do that using a `String` or just append `Location::caller`. @indietyp thoughts?

libs/error-stack/src/report.rs
```diff
@@ -452,6 +452,89 @@ impl<C> Report<C> {
...
+    ///     ...
+    ///     # };
+    ///     # fs::read_to_string(path.as_ref()).map_err(Report::from)
+    /// }
+    ///
+    /// let report = read_file("test.txt").unwrap_err();
+    /// let (io_error, _report) = report.pop_current_context();
+    /// assert_eq!(io_error.kind(), io::ErrorKind::NotFound);
+    /// ```
+    pub fn pop_current_context(self) -> (C, Report)
```
TimDiekmann: I see the motivation behind this function, but I’m having my issues with that.
The encoded behavior is weird. `pop` typically removes something from a stack, but here it’s replaced with a dummy value leaving the `Report` in another behavior state. We might want to think about another function, such as `replace_current_context` to achieve this, but `pop` should really remove a context. However, removing a context comes with various other issues:
- What happens if we hit a multi-error after popping a context? I don’t see us supporting `Report<[dyn Context]>` (at least not initially). `[dyn Context]` first of all is not valid Rust, as slices must not have unsized types as values. Also, it effectively says, that the multi-context could be of a different type, which really is against the concept of multi-contexts.
- What happens if we hit the last context? We would need to return an `Option<Report>` instead

It would be good if we would not add such a functionality as drive-by.

libs/error-stack/src/report.rs
```diff
@@ -452,6 +452,89 @@ impl<C> Report<C> {
...
+    ///     # };
+    ///     # fs::read_to_string(path.as_ref()).map_err(Report::from)
+    /// }
+    ///
+    /// let report = read_file("test.txt").unwrap_err();
+    /// let io_error = report.into_current_context();
+    /// assert_eq!(io_error.kind(), io::ErrorKind::NotFound);
+    /// ```
+    #[must_use]
+    pub fn into_current_context(self) -> C
```
TimDiekmann: This discards everything from the current context the `Report`. This includes other contexts and all attachments. Ideally, this would somehow return the rest of the `Report` somehow as well (similar to what I would expect `pop_current_context` to behave), but sometimes the sole context is everything which is required. Do you have a concrete use-case where you cannot use `current_context()` and need `into_current_context`?

It would be good, if we would not add this functionality as a drive-by.

libs/error-stack/src/frame/mod.rs
```diff
@@ -90,6 +92,72 @@ impl Frame {
         self.frame.as_any_mut().downcast_mut()
     }
 
+    /// `replace_with_printable`: if true, `Debug` and `Display` implementations will be eagerly
+    ///     evaluated and stored in place of the taken type,
+    ///     can be set to false when the `Report` will no longer be used to avoid the cost of
+    ///     formatting
+    #[must_use]
+    pub(crate) fn downcast_take<T: Send + Sync + 'static>(
```
TimDiekmann: As mentioned above, we should not add this as part of this PR. This and any other related work should be removed from this PR to make this PR focused on `Report<dyn Error>`.

Also, the behavior to replace the context with another opaque context feels like an anti pattern. If we want such a functionality, the new context should be part of the signature, and this could be defined on any `Report` (e.g. `report.replace_context(MyContext)`. Let's remove it for now.

TimDiekmann (changes requested):
Hey @zakstucke and thank you for your patience.

We agreed that we’re going to add `Report<dyn Error>`. However, we see a few issues in the current implementation. Mainly, this boils down to not adding the consuming operations (`pop_current_context` and `into_current_context`), at least not yet.
I’d propose we close #7256, and focus this PR on only adding the code for `Report<dyn Error>`. As an alternative, I’d be happy to implement it on my own PR (either later this day or sometimes this week), that’s up to you.
For the future if we want to add more current-context related API, we should park that under the `unstable` feature flag.

Another thing we might want to explore is if we could solve #4295 with `Report<dyn Error>`.

libs/error-stack/src/report.rs
zakstucke: > We don’t want to encourage people to use dyn Error as the common case. People should explicitly specify, if a Report is untyped.

I've mentioned above why I disagree on the default/no default but I won't fight it given we can type alias in userland :).

> We might need to make sure, that current_context is not callable on Report<dyn Error>.

It is not callable, as that impl block requires `C: Sized`.

libs/error-stack/src/report.rs
zakstucke: > In theory, it could be anything, such as ().

It couldn't be () as that's `Sized` and you'd lose the ability to specialise methods that don't apply to `dyn Error`, like was the case with `current_context, pop_current_context, into_current_context`. Further, using `dyn Error` is what allows the `From<Report<C>> for Report<dyn Error>` to work.

You're right on the `Send + Sync + 'static`, cool! Updated.

libs/error-stack/src/report.rs
zakstucke: Updated!

libs/error-stack/src/report.rs
zakstucke: It would be good to specify what's happened here, given no default now I guess that would be `"into Report<dyn Error> at {}"` instead?

libs/error-stack/src/report.rs
zakstucke: > pop typically removes something from a stack, but here it’s replaced with a dummy value leaving the Report in another behavior state.

As a user of the crate, this behaves identically to a pop. By replacing the type internally we don't need to worry about the boxed slices all over the internal frame layout, it has no real affect on the user, and maintains the original formatting of the report. I suppose the one "leakage" is frame iteration, which I guess we could update frame iteration to just skip these.

> What happens if we hit the last context? We would need to return an Option<Report> instead

This is impossible, `pop_current_context()` is implemented for `Report<C>` but `C` must be `Sized`, but not for `Report<dyn Error>`, after calling `pop_current_context()` you are returned an untyped report, to extract any further types you must use `downcast/downcast_take`. This is why this API fits so nicely together.

libs/error-stack/src/report.rs
zakstucke: > Ideally, this would somehow return the rest of the Report somehow as well (similar to what I would expect pop_current_context to behave)

This is exactly what `pop_current_context()` does from a user's perspective, if you don't want the top level context, you can use `downcast_take()`. 

>  Do you have a concrete use-case where you cannot use current_context() and need into_current_context?

Yes, some external error types do not implement `Clone`, but we need to extract the type to handle it, after logging the report. **This has been the most painful part of a migration to ES as there is no workaround currently.**, likewise a lack of `downcast/downcast_take`.

This exists on top of `pop_current_context()`, because it's slightly more efficient - it doesn't need to eagerly extract fmt strings like the pop variant does.

libs/error-stack/src/frame/mod.rs
zakstucke: This would be really unfortunate. It's currently a real api/usage hole not being able to get owned types out of a `Report`.

> the new context should be part of the signature, and this could be defined on any Report (e.g. report.replace_context(MyContext)

The goal of this is not to replace the error or change the printable representation in any way, but to extract a non cloneable type / a type that is expensive to clone from anywhere in the stack.

I've separated out that work into a new PR #7324, to not block this one.

zakstucke:
Thanks! I've split out the extract methods into #7324 to not block this one, I've explained the usecase there and in other review comments. 

> Another thing we might want to explore is if we could solve https://github.com/hashintel/hash/issues/4295 with Report<dyn Error>.

I've just tested the reproducer for that issue, it seems like it already works with the current version of error stack.

Compiles fine after updating the [reproducer repo](https://github.com/0x009922/error-stack-eyre-compat-issue)'s Cargo.toml they provided with:

```toml
[patch.crates-io]
error-stack.git = "https://github.com/hashintel/hash"

[dependencies]
error-stack = { version = "0.5", features = ["eyre"] }
```

libs/error-stack/src/report.rs
zakstucke: > Also, we should add a documentation in which scenarios Report<dyn Error> is discouraged and when it’s useful

Maybe it's useful to mention, `Report<dyn Error>` was already allowed before this PR, this PR now just makes it directly propagatable with `?` from any `E: Error` or `Report<C> where C: Sized`, maybe further documentation is now outside the scope of this PR?

libs/error-stack/src/report.rs
indietyp: I would recommend away from a `attach_printable`, either have a `Location::caller` directly, or a struct that provides a `Display` implementation that provides the location. Just having a printable here removes _a lot_ of valuable information and introspectability.

libs/error-stack/src/report.rs
indietyp: couldn't you just `downcast_mut`, then `mem::take`? This wouldn't violate the structure and you could just replace it with a dummy variant.

We cannot just _simply_ replace the `Context` with an attachment, it _completely_ changes the semantics of a report *and* violates many of the internal variants. Each attachment is attached to a context, by replacing it with another context you just "joink" the rug and attach attachments to an unrelated context that might not even find these useful. What about the types it provides via the provide api?

1. Every context *must have* a single context as the first layer, both formatting and internals rely on it. It's a context stacked on top of other contexts with attachments sandwiched in between
2. As @TimDiekmann eluded to there are some questions about `Report<[C]>` types, how do they fit into the equation - when I replace a report, does that mean that one arm now doesn't have a report? What about the introduced asymmetry? Which arm is chosen?

Also remember: we're talking about the unhappy path - do we really need to micro-optimize for the sake of it *without any measured benchmarking*, if so - _how can we implement this without sacrificing **correctness**_?

libs/error-stack/src/frame/mod.rs
indietyp: > It's currently a real api/usage hole

Could you elaborate on this, maybe even provide code samples - if available (or even pseudocode or a concrete example) - at least I've used the current API for _many_ different projects and haven't found need for this supposed API hole, so would we very interested in hearing the use case! 😄

libs/error-stack/tests/test_conversion.rs
```diff
@@ -88,7 +88,7 @@ fn error() {
 
 #[test]
 fn into_report() {
-    let report = io_error().map_err(Report::from).expect_err("not an error");
+    let report: Report<io::Error> = io_error().map_err(Report::from).expect_err("not an error");
```
TimDiekmann: This change is very concerning and after running the full CI (which uses pretty much all features of error-stack), this is a quite big breaking change and a big QoL regression: https://github.com/hashintel/hash/actions/runs/15422334171/job/43400766905?pr=7274

libs/error-stack/src/report.rs
zakstucke: Replied over in #7324 :)

libs/error-stack/tests/test_conversion.rs
zakstucke: This specific update was just done like this to maintain the under-test semantics of that specific test, otherwise I would've just replaced `Report::from` with `Report::new`. 

The other CI failures mainly seem to be about needing `impl Into<Report<C>>` -> `impl IntoReport<Context = C>` which I think is the current recommended way of defining these anyway. I hope at least! I'll try updating them all.

libs/error-stack/src/report.rs
zakstucke: Makes sense! Will update

libs/error-stack/src/frame/mod.rs
zakstucke: Replied in https://github.com/hashintel/hash/pull/7324

libs/error-stack/tests/test_conversion.rs
zakstucke: @TimDiekmann lets see how CI likes it now 🙏 , each of the changes I made repo wide made sense to me and improved things or stayed the same don't think a blocker.

libs/error-stack/src/report.rs
```diff
@@ -471,6 +471,27 @@ impl<C> Report<C> {
     }
 }
 
+impl<E: Into<Report<E>>> From<E> for Report<dyn Error> {
+    #[track_caller]
+    fn from(value: E) -> Self {
+        value.into().into()
+    }
+}
+
```
TimDiekmann: This is actually the issue why the change is largely breaking. Generally, a double `.into()` is a sign that this will easily result in compile errors if no type is specified. 
```suggestion
```

With that, we can revert the other changes.

libs/error-stack/src/report.rs
TimDiekmann: Yes, it was allowed but it was not possible to construct it (unless you use `unsafe` transmutation). I'd like to have some documentation around `dyn Error` before merging this.

TimDiekmann (changes requested):
Ii think we can make the implementation less breaking. To require specifying the `Report` type is suboptimal. 

Also, I like to have a bit more test coverage.
The `size` test can be expanded with this
```diff
     #[test]
     fn size() {
         assert_eq!(mem::size_of::<Report<()>>(), mem::size_of::<*const ()>());
+        assert_eq!(
+            mem::size_of::<Report<dyn Error>>(),
+            mem::size_of::<*const ()>()
+        );
     }
```

and some invariants should be checked:
```diff
   const fn report() {
         assert_send::<Report<()>>();
         assert_sync::<Report<()>>();
         assert_static::<Report<()>>();
+
+        assert_send::<Report<dyn Error>>();
+        assert_sync::<Report<dyn Error>>();
+        assert_static::<Report<dyn Error>>();
    }
```

Two notable things:
We can provide a `current_context` method for `Report<dyn Error>` (but I would probably prefer to add that in a later PR):
```rust
impl Report<dyn Error> {
    #[must_use]
    pub fn current_context(&self) -> &dyn Context {
        self.frames()
            .find_map(|frame| match frame.kind() {
                FrameKind::Context(context) => Some(context),
                FrameKind::Attachment(_) => None,
            })
            .unwrap_or_else(|| {
                // Panics if there isn't an attached context which matches `T`. As it's not possible
                // to create a `Report` without a valid context and this method can
                // only be called when `T` is a valid context, it's guaranteed that
                // the context is available.
                unreachable!(
                "Report does not contain a context. This is considered a bug and should be \
                reported to https://github.com/hashintel/hash/issues/new/choose"
            );
            })
    }
}
```

And your assumption about when exactly what signature is used is not fully correct.
We have three different states of `Report`:
1. `Report<C>`
2. `Report<dyn Error>`
3. `Report<[C]>`

To share some implementations, these signatures can be used

|                                     | `Report<C>` | `Report<dyn Error>` | `Report<[C]>` |
|:-----------------------------------:|:-----------:|:-------------------:|:-------------:|
|         `impl<C> Report<C>`         |      ✅      |          ❌          |       ❌       |
| `impl Report<dyn Error>` |      ❌      |          ✅          |       ❌       |
| `impl<C> Report<[C]>` |      ❌      |          ❌          |       ✅       |
| `impl<C: ?Sized + Error> Report<C>` |      ✅      |          ✅          |       ❌       |
|     `impl<C: ?Sized> Report<C>`     |      ✅      |          ✅          |       ✅       |

This might be useful when implementing functionality on the different states.

libs/@local/graph/api/src/rest/data_type.rs
```diff
@@ -438,9 +438,7 @@ where
             actor_id,
             // Manually deserialize the query from a JSON value to allow borrowed deserialization
             // and better error reporting.
-            GetDataTypesParams::deserialize(&request)
-                .map_err(Report::from)
-                .map_err(report_to_response)?,
+            GetDataTypesParams::deserialize(&request).map_err(report_to_response)?,
```
TimDiekmann: We can keep this drive-by regardless of the above comment. It was from a time where we did not have `IntoReport`.

libs/error-stack/src/report.rs
TimDiekmann: I don't know if we want to revert the other changes, but it would be possible. As an alternative implementation we could also use `impl<E: Context> From<E>`.

TimDiekmann: Regardless of the implementation, we should not use the `.into()` below but construct it ourselves. Otherwise, we'd have the attachment that it got converted to `Report<dyn Error>`.

libs/error-stack/src/report.rs
```diff
@@ -471,6 +471,27 @@ impl<C> Report<C> {
...
+    #[track_caller]
+    fn from(value: Report<C>) -> Self {
+        Report {
+            frames: value.frames,
+            _context: PhantomData,
+        }
+        .attach_printable(alloc::format!(
+            "into Report<dyn Error> at {}",
+            Location::caller()
+        ))
```
TimDiekmann: You mentioned you updated this, but it's still a string. Either, we want to have a dedicated attachment struct which wraps the location or we simply add the `Location`. The latter is easier to modify when outputting, which is why I would prefer that route (people frequently asked how to hide the location, and this is a potential issue as this message is not easy to hide if it's a string or a private struct).

libs/error-stack/src/report.rs
```diff
@@ -892,6 +913,12 @@ impl<C> From<Report<C>> for Report<[C]> {
     }
 }
 
+impl<C: Context> From<C> for Report<[C]> {
+    fn from(ctx: C) -> Self {
+        Report::new(ctx).into()
+    }
+}
+
```
TimDiekmann: cc @indietyp thoughts? I think it's fine but probably should be generalized using `into_report()` using
```rust
impl<E: IntoReport<Context: Sized>> From<E> for Report<[E::Context]> {
    fn from(error: E) -> Self {
        Self {
            frames: error.into_report().frames,
            _context: PhantomData,
        }
    }
}
```

but only if we keep `impl<E: Into<Report<E>>> From<E> for Report<dyn Error>` (or however we'd implement it)

libs/error-stack/src/report.rs
TimDiekmann: Also, the changelog needs updating

libs/@local/repo-chores/rust/src/benches/analyze/tracing.rs
```diff
@@ -41,7 +41,7 @@ impl FoldedStacks {
     pub fn create_flame_graph(
         &self,
         mut options: flamegraph::Options,
-    ) -> Result<FlameGraph, Report<impl Error + Send + Sync + 'static>> {
+    ) -> Result<FlameGraph, Report<dyn Error>> {
```
indietyp: I am against this change, it _fundamentally_ weakens the guarantees of what we're trying to achieve here. The error is still concrete and imho should stay concrete, even if opaque. Moving it to `dyn` doesn't solve the problem, just makes it weaker for no reason.

libs/error-stack/src/report.rs
zakstucke: This generalisation doesn't quite work: 
```
conflicting implementations of trait `std::convert::From<report::Report<_>>` for type `report::Report<[_]>`
conflicting implementation for `report::Report<[_]>`
```

libs/error-stack/src/report.rs
zakstucke: > Yes, it was allowed but it was not possible to construct it (unless you use unsafe transmutation). I'd like to have some documentation around dyn Error before merging this.

Got it, cool! I've added a changelog entry, do you want to take a stab at documentation? I think you have a much better feel for how you'd like it presented.

libs/error-stack/src/report.rs
zakstucke: > Makes sense! Will update

Sorry, just hadn't got to this one yet! Update**d**

libs/@local/repo-chores/rust/src/benches/analyze/tracing.rs
zakstucke: Updated :)

libs/error-stack/src/report.rs
zakstucke: Fixed the `.into().into()`.

We could change to `impl<E: Context> From<E>`, it doesn't affect the type inference tweak as far as I can tell.

> This is actually the issue why the change is largely breaking. Generally, a double .into() is a sign that this will easily result in compile errors if no type is specified.

Unfortunately it's the specific part I opened the PR for, if you look at the actual changes required across your repo, and the additional benefit of allowing `impl<C: Context> From<C> for Report<[C]>`, I hope you'll be fine accepting this.

libs/error-stack/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ All notable changes to `error-stack` will be documented in this file.
 - `Extend` is no longer implemented by `Report<C>`, instead it is implemented on `Report<[C]>`, either use `From` or `Report::expand` to convert between `Report<C>` into `Report<[C]>`. ([#5047](https://github.com/hashintel/hash/pull/5047))
 - `extend_one` has been renamed to `push` and is only implemented on `Report<[C]>`. ([#5047](https://github.com/hashintel/hash/pull/5047))
 - `bail!(report,)` has been removed, one must now use `bail!(report)`. This is in preparation for the unstable `bail!` macro that allows to construct `Report<[C]>`. ([#5047](https://github.com/hashintel/hash/pull/5047))
+- `From<C> for Report<[C]>` and support for constructing `Report<dyn Error>` with `From<E: IntoReport<E>>` and `From<E: Report<C>>` alter type inferrence in some cases. Each instance will require one of: replacing `Report::from` with `Report::new`, replacing `Into<Report<C>>` with `IntoReport<Context = C>`, or specifying a concrete `C`. This makes multi-error reports easier to instanstiate and allows flexible usage of dynamic errors inside a report. ([#7274](https://github.com/hashintel/hash/pull/7274))
```
TimDiekmann: Fundamental parts are missing:
- We did introduce `Report<dyn Error>`. At the same time, it should be said, it's discouraged to be used as return type and is mainly interesting for locations where we don't need the context type, e.g. in error-handling. It's expected to be seen more in argument position in general
- I said above, that we're not going to add the additional `From` implementation involving `into().into()` as it breaks inference. It's not expected, that `Report::from` or `Into<Report<C>>` needs to change.
- I don't see, why we added a `From` implementation for multi-errors. Even if we would, this PR has a fundamentally other focus.

libs/error-stack/src/report.rs
TimDiekmann: You can add the docs on `Report` itself. Just give it a go and we can iterate on it.

libs/error-stack/src/report.rs
TimDiekmann: I'm not so much about the `.into().into()` but on the trait signature itself. It doesn't matter how the trait is implemented, it's where it applies. If you implement a trait, which is something along
```rust
impl<E: Into<X>> From<E> for Y
```
it's generally a bad code design as it will always break inference. Conceptually it says everything which can be converted to `X` can also be converted to `Y`. Such a transitive conversion is almost never a good idea.

I thought the motivation for the pull request was to allow an intermediate report?

libs/error-stack/src/report.rs
TimDiekmann: This would require to change some other PRs as well. We should not add this conversion as part of this PR, it's highly unrelated.

libs/@local/repo-chores/rust/src/benches/analyze/tracing.rs
TimDiekmann: I really don't see why you updated this to `io::Error`. We purposefully used `impl Error` here as `io::Error` might change. It just happened to be `io::Error`, but downstream crates should not rely on that. 
- It's not `Report<io::Error>` because we don't want to specify `io::Error`
- It's not `Report<dyn Error>` as it's always a single concrete type

TimDiekmann (changes requested):

libs/error-stack/src/report.rs
zakstucke: > I thought the motivation for the pull request was to allow an intermediate report?

The title of the PR:
> Alternative to https://github.com/hashintel/hash/pull/7256 that adds ? support to Report<dyn Context + Send + Sync + 'static> instead of an AnyReport wrapper.

Would you be open to having this impl behind a feature flag? The tradeoff here i would make always, but understand completely why you wouldn't. It really was the crux of the PR for me, motivation explained back in #7256. 

The PR would still allow a user to reinvent a wrapping report type that has this dyn Error inside but would be much nicer if i could enable a feature and not reinvent the wheel on top

libs/@local/repo-chores/rust/src/benches/analyze/tracing.rs
zakstucke: Type inference change from the `From<E>` impl prevents `impl Error`, will revert once the from impl is removed / feature flagged

libs/error-stack/src/report.rs
TimDiekmann: I have reservations about introducing this `From<E>` for `Report<dyn Error>` implementation. While it offers a degree of ergonomic convenience, it also introduces several concerns:

- **Type Erasure:** This conversion inherently erases the concrete error type `E`, which can hinder precise error handling and diagnostics. We don't want to encourage people to use `Report<dyn Error>` in return position, and this implementation suggests doing that.
- **Ambiguity in `Report::from`:** With this implementation, `Report::from` becomes less predictable. Depending on the context, it might produce either `Report<E>` or `Report<dyn Error>`, leading to potential confusion and maintenance challenges.
- **Feature Flag Complications:** Introducing this under a feature flag doesn’t resolve the underlying issues. Feature flags in Rust are additive, and enabling this in one crate can inadvertently affect others in the dependency graph, leading to unexpected type inference failures and coherence conflicts.

Given these considerations, I am against merging this implementation. Instead, we should favor explicit conversions to `Report<dyn Error>` where necessary, ensuring clarity and preserving type information, e.g. by using `E.into_report()?`.

zakstucke: This is where it gets tricky for me, it's one of the fundamental thing I'm looking for, we've come to the conclusion we need to use ES more flexibly like this, across triple digit crates in various stages of ES usage and are happy with the tradeoff being made re type erasure. 

I think I'll have to publish a synced fork then that makes this alternate choice. I've just tried to implement wrappers in userland to avoid this (which I thought I could in the previous comment now `Report<dyn Error>` can be created), but it still doesn't seem feasible. The behaviour cannot be achieved without either an `AnyReport` wrapper inside the ES crate (https://github.com/hashintel/hash/pull/7256), or this `From<E> for Report<dyn Error>` impl here.

I'll clean up this PR to the way you want it now.

libs/@local/graph/api/src/rest/data_type.rs
zakstucke: This was only possible because of `impl<C: Context> From<C> for Report<[C]>`, reverted.

libs/@local/repo-chores/rust/src/benches/analyze/tracing.rs
zakstucke: Reverted, btw, the `from_file()` method just above it returns `Result<Self, Report<io::Error>>`, I guess you want to change that to `Result<Self, Report<impl Error + Send + Sync + 'static>>` to fit the reason you have it on `create_flame_graph`

libs/error-stack/CHANGELOG.md
zakstucke: Updated

TimDiekmann:
As requested changes were not applied (i.e. documentation changes) and we also don't encourage using `Report<dyn Error>`, I'm going ahead and close this.

PR #7321, opened 2025-06-03
H-4704: HashQL: Move type tests into separate files

indietyp:

Many unit tests for the type system (which are required) take up significant space in the file itself, to the point where more than ~50% are just tests. That makes navigation harder than necessary and skews the perceived complexity and coverage.

You should also take a look at: https://github.com/hashintel/hash/blob/ab9a056f97f12213a1eab6fcaf28c8a5cea13114/libs/@local/hashql/core/src/type/kind/tests/closure.rs#L34-L105

It outlines the idea behind https://linear.app/hash/issue/H-4547/hashql-compress-type-test-suite, I've decided to not continue this in this PR, due to the sheer volume, but the idea is to move away from the individual closures to the type builder in conjunction with rstest cases to make the overall code required for tests smaller and the ability to add additional tests easier.

This PR:

The changes in this PR:

The changes in this PR:


Cargo.toml
```diff
@@ -73,7 +73,6 @@ hash-codec.path                     = "libs/@local/codec"
 hash-codegen.path                   = "libs/@local/codegen"
 hash-graph-api.path                 = "libs/@local/graph/api"
 hash-graph-authorization.path       = "libs/@local/graph/authorization"
-hash-graph-migrations.path          = "libs/@local/graph/migrations"
```
indietyp: istg cargo add, why do you modify the main cargo.toml >.<

libs/@local/hashql/core/Cargo.toml
```diff
@@ -44,5 +44,5 @@ serde = ["dep:serde", "text-size/serde"]
 workspace = true
 
 [dev-dependencies]
-insta     = { workspace = true }
-test-case = "3.3.1"
+insta  = { workspace = true }
+rstest = "0.25.0"
```
TimDiekmann:
```suggestion
rstest = { workspace = true }
```

PR #7322, opened 2025-06-03
H-4624: Implement versioning for policies

TimDiekmann:

We want to track the history of policies.

## 🔍 What does this change?

- Alter the tables related to policies to include temporal versioning
- Adjust queries to use the versioning
- Introduce an `archive_policy_by_id`
- Change the default mode in the Node API to archive instead of delete a policy (this will be the only option in the future as deletion is only used for temporary policies, which will be moved to the graph instead)
- Name a few more policies. They should be deleted afterwards but in case something goes wrong it's always safe to remove them.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graph/knowledge/system-types/hash-instance.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/hash-instance.ts
@@ -129,7 +129,7 @@ export const createHashInstance: ImpureGraphFunction<
     return getHashInstanceFromEntity({ entity });
   } finally {
     await deletePolicyById(ctx.graphApi, authentication, instantiationPolicy, {
-      permanent: false,
+      permanent: true,
     });
   }
 };


libs/@local/graph/api/src/rest/permissions.rs
```diff
@@ -54,11 +55,16 @@ impl PermissionResource {
...
-                        .put(update_policy_by_id::<S, A>)
-                        .delete(delete_policy_by_id::<S, A>),
+                    Router::new()
+                        .route(
+                            "/",
+                            get(get_policy_by_id::<S, A>)
+                                .put(update_policy_by_id::<S, A>)
+                                .delete(archive_policy_by_id::<S, A>),
+                        )
+                        .route("/delete", delete(delete_policy_by_id::<S, A>)),
```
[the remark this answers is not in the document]
TimDiekmann: The `delete` route will be removed in the future, so this is fine.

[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer I understand your concern about potential confusion. The decision to use DELETE for both operations follows REST principles where DELETE is the appropriate verb for removal operations, but with different semantics based on the path.

However, as @TimDiekmann noted, this is a transitional design - the permanent deletion endpoint (`/policies/:id/delete`) will be removed in the future, leaving only the archive operation. This makes sense if archiving is becoming the standard operation and permanent deletion will be phased out.

In the meantime, the endpoint separation through different paths (`:id` vs `:id/delete`) should provide sufficient distinction for API consumers. The OpenAPI documentation also clearly distinguishes between these operations with different `operationId` values (`archive_policy_by_id` vs `delete_policy_by_id`).

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -960,19 +974,22 @@ where
...
                 SELECT
-                    policy.name,
-                    policy.effect,
-                    policy.principal_id,
-                    policy.principal_type,
-                    policy.actor_type,
-                    policy.resource_constraint,
+                    policy_edition.name,
+                    policy_edition.effect,
+                    policy_edition.principal_id,
```
hashdotai, line 979: Since `policy_edition` now has a `transaction_time` column, consider adding a check in the `get_policy_by_id` function to verify that the policy exists but might be archived. This would provide a better error message than just 'PolicyNotFound' when someone tries to access an archived policy.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1264,69 +1286,122 @@ where
...
+                        )
+                        SELECT
+                            id,
+                            name,
+                            tstzrange(now(), NULL, '[)'),
+                            {effect},
+                            principal_id,
+                            principal_type,
+                            actor_type,
+                            {resource_constraint}
```
hashdotai, line 1397: You're using `resource_constraint` in the SQL query here but referencing `resources_constraint` in the Rust code. One of them should be changed to match the other for consistency.

libs/@local/graph/sdk/typescript/src/policy.ts
```diff
@@ -98,13 +98,17 @@ export const updatePolicyById = (
     .then(({ data: policy }) => policy as Policy);
 
 /**
- * Permanently removes the policy specified by it's ID.
+ * Removes the policy specified by it's ID.
```
hashdotai: This comment should be updated to reflect that it archives policies rather than deleting them.

libs/@local/graph/migrations/graph-migrations/v003__policies/down.sql
```diff
@@ -1,12 +1,8 @@
--- Drop policy-action junction table
 DROP TABLE policy_action;
-
--- Drop policy table
```
hashdotai, line 4: The down migration should drop tables in the reverse order of creation to avoid foreign key constraint issues. Make sure to drop `policy_action` before `policy_edition`, and `policy_edition` before `policy`.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1264,69 +1286,122 @@ where
...
         }
 
+        if !actions_to_remove.is_empty() {
+            let actions_to_remove = actions_to_remove.iter().collect::<Vec<_>>();
+            transaction
+                .as_client()
+                .execute(
+                    "
+                        UPDATE policy_action
+                        SET transaction_time = tstzrange(lower(transaction_time), now(), '[)')
```
hashdotai, line 1324: Consider adding a helper function or utility to close the time range for both `policy_edition` and `policy_action` tables, since this pattern is repeated in multiple places in the code. This would reduce duplication and make future maintenance easier.

libs/@local/graph/sdk/typescript/src/policy.ts
TimDiekmann: Archival is not directly visible to the user. For the user, this feels like a removal. I purpusfilly said "Removes", not "Deletes".

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1178,37 +1192,44 @@ where
...
+                    policy_edition.effect,
+                    policy_edition.principal_id,
+                    policy_edition.principal_type,
+                    policy_edition.actor_type,
+                    policy_edition.resource_constraint,
                     array_remove(array_agg(policy_action.action_name), NULL)
-                FROM policy
-                LEFT JOIN policy_action ON policy.id = policy_action.policy_id
+                FROM policy_edition
+                LEFT JOIN policy_action ON policy_edition.id = policy_action.policy_id
```
indietyp: don't we need a where clause on the transaction time as well here?

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: Actually not a bad idea. Captured in H-4707: Return a dedicated error if a user tries to access an archived policy.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: Yes, thank you!
```suggestion
                LEFT JOIN policy_action
                       ON policy_edition.transaction_time @> now()
                      AND policy_edition.id = policy_action.policy_id
```

libs/@local/graph/migrations/graph-migrations/v003__policies/down.sql
TimDiekmann: That's how it's implemented...

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: The overhead of adding the function (note, that's it Postgres code) would not outweigh the saved effort. Also, we're going to move to `clorinde` for simple queries (simple queries = non-dynamic queries).

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: This is what we have an AI reviewer for 😅 
19a1fe2a1dd826f2665577b5d89cc667d73d289f

apps/hash-api/src/graph/knowledge/system-types/hash-instance.ts
```diff
@@ -127,6 +128,8 @@ export const createHashInstance: ImpureGraphFunction<
 
     return getHashInstanceFromEntity({ entity });
   } finally {
-    await deletePolicyById(ctx.graphApi, authentication, instantiationPolicy);
+    await deletePolicyById(ctx.graphApi, authentication, instantiationPolicy, {
+      permanent: false,
```
CiaranMn: Why is deletion not permanent, if the policy is `tmp`?

TimDiekmann: Whoops, fixed in 8ef4ee8b144ecc21a30a562b4b81c51fa0ece3d3, thank you!
(I had an `archive = true` flag previously and changed it to `permanent = false`, so this is how the issue happened)

PR #7323, opened 2025-06-03
H-4705: Initial setup for clorinde

TimDiekmann:

`clorinde` generates Rust bindings for existing SQL queries. This greatly improve DX and robustness.

## 🔍 What does this change?

- Set up codegen for `clorinde`
- Add `hash-graph-postgres-queries` crates which re-exports the queries from `clorinde`
- Add test to automatically generate `clorinde` and checks for diffs
	- As part of this, I enhanced the `sync-turborepo` script to support passing a list of packages to generate
- default-exclude `clorinde` from being mentioned in the dependency doc

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

The code is generated from a test, which will automatically check the diff


libs/@local/graph/postgres-queries/clorinde.toml
```diff
@@ -0,0 +1,12 @@
+use-workspace-deps = "../../../../Cargo.toml"
```
hashdotai, line 2: It's a good practice to add a comment at the top of this file explaining what this configuration file is for and how it's used by clorinde. This would make it easier for future contributors to understand the purpose of this file.

libs/@local/graph/postgres-queries/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@rust/hash-graph-postgres-queries",
+  "version": "0.0.0-private",
+  "private": true,
+  "license": "AGPL-3",
+  "scripts": {
+    "doc:dependency-diagram": "cargo run -p hash-repo-chores -- dependency-diagram --output docs/dependency-diagram.mmd --root hash-graph-postgres-queries --root-deps-and-dependents --link-mode non-roots --include-dev-deps --include-build-deps --logging-console-level info",
+    "fix:clippy": "just clippy --fix",
+    "lint:clippy": "just clippy",
+    "test:integration": "clorinde live \"postgresql://${HASH_GRAPH_PG_USER:-graph}:${HASH_GRAPH_PG_PASSWORD:-graph}@${HASH_GRAPH_PG_HOST:-localhost}:${HASH_GRAPH_PG_PORT:-5432}/${HASH_GRAPH_PG_DATABASE:-graph}\" && printf '\n[lints.rust]\ndead_code = \"allow\"\nexported_private_dependencies = \"allow\"\nunused_imports = \"allow\"' >> clorinde/Cargo.toml && just sync-turborepo @rust/clorinde && git diff --exit-code clorinde"
```
hashdotai: This test command is quite dense and might be difficult to understand at a glance. Consider breaking it into separate scripts or adding comments to explain what each part does. For example, you could split it into `generate-clorinde`, `append-lints`, and `verify-no-changes`.

.github/scripts/rust/sync-turborepo.rs
```diff
@@ -465,7 +470,12 @@ fn main() {
         workspace: &workspace,
     };
 
+    let args_set = iter::from_fn(|| args.next()).collect::<HashSet<_>>();
```
hashdotai: Consider using a more descriptive variable name than `args_set`. Something like `requested_packages` would better communicate the intent of this collection.

libs/@local/graph/postgres-store/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tokio-postgres    = { workspace = true, public = true }
 # Private workspace dependencies
 error-stack                    = { workspace = true, features = ["std", "serde", "unstable"] }
 hash-codec                     = { workspace = true, features = ["numeric", "postgres"] }
+hash-graph-postgres-queries    = { workspace = true }
```
hashdotai, line 27: Make sure this new dependency is properly documented somewhere (perhaps in a comment) to explain why the postgres-store crate now depends on postgres-queries.

libs/@local/graph/postgres-queries/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+//! # HASH Graph Postgres queries
+//!
+//! ## Workspace dependencies
+#![cfg_attr(doc, doc = simple_mermaid::mermaid!("../docs/dependency-diagram.mmd"))]
+
+pub use ::clorinde::queries::*;
```
hashdotai, line 6: Consider adding a brief comment explaining what types of queries are being re-exported here, and how they're intended to be used.

libs/@local/graph/postgres-queries/clorinde.toml
TimDiekmann: We're not going to add comments to each configuration file how it's defined. That's what their documentation is for.

libs/@local/graph/postgres-queries/package.json
TimDiekmann: I agree that this is not ideal, however, these commands should always be executed together, so I'll keep it.

.github/scripts/rust/sync-turborepo.rs
TimDiekmann: [`5e850ac` (#7323)](https://github.com/hashintel/hash/pull/7323/commits/5e850ac563fd95b9f8d365c5edfa7799b82d0e56)

libs/@local/graph/postgres-queries/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+# This file was generated with `clorinde`. Do not modify.
```
indietyp: is this statement still true?

TimDiekmann: Nope 😅 
```suggestion
```

libs/@local/graph/postgres-queries/package.json
indietyp: (for completeness sake - @hashdotai we wouldn't put them into separate test commands, but rather a dedicated mise task instead - but I agree with Tim that this isn't required here _yet_)

indietyp (dismissed):
lgtm 🎉

libs/@local/graph/postgres-queries/package.json
TimDiekmann: Is it possible to have "local" mise tasks? Tasks which only apply to a single folder?

indietyp: Yes, you just create a `.config/mise.toml` file in the directory. You can then figure out the loading with `mise cfg` - if needed.

> These files recurse upwards, so if you have a ~/src/work/myproj/mise.toml file, what is defined there will override anything set in ~/src/work/mise.toml or ~/.config/mise.toml. The config contents are merged together.

from: https://mise.jdx.dev/configuration.html

TimDiekmann: Oh, great to know! This indeed will help defining tasks.

vilkinsons (approved):

libs/@local/graph/postgres-queries/package.json
TimDiekmann: Just tested this and it works as well. Will add this in a follow-up:
```bash
#!/usr/bin/env bash

set -euo pipefail

clorinde live "postgresql://${HASH_GRAPH_PG_USER:-graph}:${HASH_GRAPH_PG_PASSWORD:-graph}@${HASH_GRAPH_PG_HOST:-localhost}:${HASH_GRAPH_PG_PORT:-5432}/${HASH_GRAPH_PG_DATABASE:-graph}"

cat >> clorinde/Cargo.toml <<- EOF

[lints.rust]
dead_code                     = "allow"
exported_private_dependencies = "allow"
unused_imports                = "allow"
EOF

just sync-turborepo @rust/clorinde

git diff --exit-code clorinde
```

vilkinsons (approved):

PR #7325, opened 2025-06-03
H-4675: Optionally authenticate to Vault via AWS IAM

CiaranMn:

When setting up a Vault client, if there is no `ROOT_TOKEN` in the environment, attempt to log in via AWS IAM instead.

This assumes the Vault server is appropriately configured with the `aws` auth backend enabled and a Vault role associated with it that is bound to an ARN pattern that matches the ARN of the role running the AWS task (e.g. the one running the Node API, or the AI worker).

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- See if it works.

## ❓ How to test this?

1. Merge it


libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -1,6 +1,127 @@
...
+
+  /**
+   * Sign the request to AWS. This does not involve a network call – it creates a hash using AWS's signing algorithm.
+   * The access ID of the credentials are transmitted as plain text, so that when Vault forwards the request to AWS,
+   * AWS can look up the secret key associated with the access ID and verify the request using the same algorithm.
+   */
+  let signed: IHttpRequest;
+  try {
+    signed = await signer.sign(unsigned);
+  } catch {
```
hashdotai, line 65: This catch block should include the error parameter and log it to provide more context when debugging. Without catching the error, it's hard to diagnose what went wrong with AWS credential signing.

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -1,6 +1,127 @@
...
+
+    return data.auth;
+  } catch (error) {
+    if (error instanceof AxiosError) {
+      throw new Error(`Failed to renew Vault token: ${error.message}`);
+    }
+
+    throw new Error(`Failed to renew Vault token: ${stringifyError(error)}`);
+  }
+};
```
hashdotai, line 124: This function is missing JSDoc comments to explain what it does and the parameters it takes, which would be helpful since it's handling token renewal logic.

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -41,14 +205,55 @@ export class VaultClient {
...
   }
 
+  private async ensureToken() {
+    if (typeof this.#token === "string") {
+      return;
+    }
+
+    const now = Date.now();
+
+    if (now < this.#token.leaseDurationMs - 60_000) {
```
hashdotai, line 216: Add a buffer window for token renewal. Currently, we renew if there's less than 60 seconds left, but it would be safer to use a larger window (e.g., 5 minutes) to ensure there's plenty of time for renewal to complete, especially if there are network delays.

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -73,23 +278,55 @@ export class VaultClient {
...
+  logger: Logger;
+}) => {
+  if (!process.env.HASH_VAULT_HOST || !process.env.HASH_VAULT_PORT) {
+    logger.info(
+      "No HASH_VAULT_HOST or HASH_VAULT_PORT provided, skipping Vault client creation",
+    );
+    return undefined;
+  }
+
+  if (!process.env.HASH_VAULT_ROOT_TOKEN) {
```
hashdotai, line 318: The log level should be more descriptive here. Consider changing this from 'info' to 'debug' since this is more of a detailed operational message rather than something significant to the application state.

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -41,14 +205,55 @@ export class VaultClient {
...
+        const renewedToken = await renewToken(
+          this.#vaultAddr,
+          this.#token.clientToken,
+        );
+
+        this.#token.leaseDurationMs = now + renewedToken.lease_duration * 1_000;
+
+        this.#token.renewable = renewedToken.renewable;
+        return;
+      } catch {
```
hashdotai: This catch block should include the error parameter and log it (even if just at debug level) to provide more context for debugging failures in token renewal.

apps/hash-api/src/index.ts
```diff
@@ -243,7 +243,11 @@ const main = async () => {
 
   const temporalClient = await createTemporalClient(logger);
 
-  const vaultClient = createVaultClient();
+  const vaultClient = await createVaultClient({ logger });
+
+  if (!vaultClient && !isSelfHostedInstance) {
```
hashdotai, line 248: The condition `!vaultClient && !isSelfHostedInstance` might not capture all relevant cases. Consider separating the concerns: first check if the instance is self-hosted, and only then check if the Vault client is required.

apps/hash-api/src/integrations/sync-back-watcher.ts
```diff
@@ -96,13 +112,17 @@ export const createIntegrationSyncBackWatcher = async (
...
+          graphApi,
+          vaultClient,
+        });
 
         return true;
       })
       .catch((err) => {
-        // eslint-disable-next-line no-console -- caught because this function loses ownership of the queue occasionally in dev
-        console.error(`Could not take message from queue: ${err.message}`);
+        // caught because this function loses ownership of the queue occasionally in dev
```
hashdotai, line 124: Now that you have access to a proper logger, you should add more context to this error message. Consider including information about which queue was being accessed.

libs/@local/hash-backend-utils/src/vault.ts
CiaranMn: I think the current approach is fine. The call to `sign` does not require a network request, it is simply adding certain headers to the request based on the presence of AWS credentials. It failing indicates there are no credentials.

libs/@local/hash-backend-utils/src/vault.ts
CiaranMn: The function purpose and parameters are self-explanatory.

libs/@local/hash-backend-utils/src/vault.ts
CiaranMn: 1 minute is plenty for a request to Vault, if it's taking more than that to respond something is seriously wrong. The tokens only last for 15 minutes, 5 minutes would be excessive.

libs/@local/hash-backend-utils/src/vault.ts
CiaranMn: This will only be called once when the process starts (in current usage) so I think helpful to have as `info` so it appears in remote server logs. It will not if it's set to `debug`.

libs/@local/hash-backend-utils/src/vault.ts
CiaranMn: I am worried about information leakage, the presence of an error is enough. We can add more details if there are errors in practice.

apps/hash-api/src/index.ts
CiaranMn: That sounds equivalent

apps/hash-api/src/integrations/sync-back-watcher.ts
CiaranMn: There's only one queue used in this function, which can be checked by looking at the value of `HASH_INTEGRATION_QUEUE_NAME`

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -15,18 +136,61 @@ type VaultSecret<D = unknown> = {
...
-        "X-Vault-Token": params.token,
+        "X-Vault-Token":
+          typeof params.token === "string"
+            ? params.token
+            : params.token.clientToken,
       },
     });
 
-    this.client.interceptors.response.use(
+    this.#client.interceptors.request.use(async (cfg) => {
```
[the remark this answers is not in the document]
CiaranMn: It only results in a network request if the token requires renewal, the synchronous work involved in checking is tiny.

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -1,6 +1,127 @@
...
+    iam_request_url: toBase64(`https://${signed.hostname}${signed.path}`),
+    iam_request_body: toBase64(signed.body as string),
+    iam_request_headers: toBase64(JSON.stringify(signed.headers)),
+  };
+
+  try {
+    const response = await axios.post<{ auth: VaultLoginResult }>(
+      `${opts.vaultAddr}/v1/auth/aws/login`,
+      payload,
+      { timeout: 5000 },
```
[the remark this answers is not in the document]
CiaranMn: `60_000` appears once and there is no need for the two timeouts to be synchronized, I think this is fine

libs/@local/hash-backend-utils/src/vault.ts
```diff
@@ -41,14 +205,55 @@ export class VaultClient {
     );
   }
 
+  private async ensureToken() {
```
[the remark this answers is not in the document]
CiaranMn: Good suggestion, applied in https://github.com/hashintel/hash/pull/7325/commits/640bc3b17f9a6c16c3e7d33f8a410447c70b7b7f

TimDiekmann (approved):

PR #7329, opened 2025-06-04
H-4712: Store policies in Snapshots

TimDiekmann:

When we move away from SpiceDB, the new policies needs to be covered instead of the old relationships.

This PR adds snapshot logic for dumping and restoring policies. The snapshot tests were updated to cover the most important policies.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/type-fetcher/src/store.rs
+++ b/libs/@local/graph/type-fetcher/src/store.rs
@@ -19,8 +19,8 @@ use hash_graph_authorization::{
     },
     schema::{
         DataTypeRelationAndSubject, DataTypeViewerSubject, EntityRelationAndSubject,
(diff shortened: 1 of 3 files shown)


apps/hash-graph/src/subcommand/snapshot.rs
```diff
@@ -27,6 +27,14 @@ pub struct SnapshotDumpArgs {
     #[clap(long)]
     pub no_principals: bool,
 
+    /// Whether to skip dumping the actions.
+    #[clap(long)]
+    pub no_actions: bool,
```
indietyp, line 32: sad that https://github.com/clap-rs/clap/issues/815 is still not a thing :/

indietyp (approved):

PR #7330, opened 2025-06-04
H-4714: HashQL: re-span types during type inference

indietyp:

Currently types point to the place they've been defined, instead of the place they're mentioned.  This makes many of the diagnostics more convoluted than they should be.

As drive-by completely strips ansi escapes have been removed to make the snapshots more readable.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/core/src/intern/map.rs
+++ b/libs/@local/hashql/core/src/intern/map.rs
@@ -657,40 +657,34 @@ where
     /// # Examples
     ///
     /// ```
-    /// # use hashql_core::intern::InternMap;
-    /// # use hashql_core::heap::Heap;
-    /// # use hashql_core::id::{HasId, Id};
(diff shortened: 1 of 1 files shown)


indietyp:
@TimDiekmann for a comparison, you can take a look at https://github.com/hashintel/hash/blob/e4af69df367a3a503ecd95a6fbfc0644752a7632/libs/@local/hashql/hir/tests/ui/lower/checking/closure-invalid-return-type.stderr#L1-L15 and similar.

libs/@local/hashql/hir/tests/ui/lower/checking/closure-invalid-return-type.stderr
indietyp, line 1: @TimDiekmann you can take a look at this file for a good comparison

TimDiekmann (approved):

PR #7339, opened 2025-06-05
H-4719: HashQL: restructure `kernel` and `math` modules

indietyp:

The math module is currently a catch-all of operations, instead we should move these into dedicated modules, as many are not really math operations.

`>`, `<=` etc aren't really math operations, but instead are compare operations. They should be - as such - be in a separate module.

This separates the current module structure into the following categories:

* language intrinsics and primitives (like e.g. `String`) are part of the `kernel` module
* the actual standard library (user constructible types), e.g. `Url` and such *are part of* the `core` module

This change is done now, instead of later, as any adjustment down the line will just make the move more painful, especially once specialisation of HIR nodes hits.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/module/std_lib.rs
```diff
@@ -16,6 +17,38 @@ use crate::{
...
+            closure = $this.ty.generic([
+                $(
+                    (${concat($generic, _arg)}, None $(.or(Some($generic_bound)))?)
+                ),*
+            ] as [(GenericArgumentId, Option<TypeId>); ${count($generic)}], closure)
+        }
+
+        $this.type_def(closure, &[$(${concat($generic, _ref)}),*])
+    }};
+}
```
hashdotai, line 50: There's a small issue with the implementation of the `decl!` macro. The macro is defined inside the implementation of `StandardLibrary`, which is unusual. It would be better to move this macro outside the impl block, preferably at the top of the file or in a separate module.

libs/@local/hashql/core/src/module/std_lib.rs
```diff
@@ -303,194 +260,374 @@
...
 
-            // Arithmetic
-            let funcs = &[
+            let items = [
                 (
-                    "::math::add",
+                    "::core::math::add",
                     &["+"] as &[&'static str],
-                    decl!(<T: Number, U: Number>(lhs: T, rhs: U) -> self.ty.union([T, U])),
+                    decl!(self; <T: Number, U: Number>(lhs: T, rhs: U) -> self.ty.union([T, U])),
```
hashdotai, line 372: In both `core_option_module` and the other similar methods, you're using hard-coded paths like `"::core::option::None"`. It would be more maintainable to build these paths programmatically based on the parent module ID, perhaps with a helper method that concatenates the module path.

libs/@local/hashql/core/src/module/std_lib.rs
```diff
@@ -303,194 +260,374 @@
...
-                    "::math::root",
+                    "::core::math::root",
                     &[], // cannot use `ⁿ√` because `ⁿ` is a letter, not a symbol
-                    decl!(<>(value: Number, root: Number) -> Number),
+                    decl!(self; <>(value: Number, root: Number) -> Number),
                 ),
-                // Bitwise operations
+            ];
+
+            let items: SmallVec<_> = items
```
hashdotai, line 422: Nice improvement using `SmallVec` instead of `Vec` for the collection of items. This is more memory efficient for small collections, which is likely the common case here.

libs/@local/hashql/core/src/module/std_lib.rs
```diff
@@ -303,194 +260,374 @@
...
                 ),
-                // Comparison operations
+            ];
+
+            let items: SmallVec<_> = items
+                .iter()
+                .flat_map(|(name, alias, def)| {
+                    self.alloc_intrinsic_value(id.value(), name, alias.iter().copied(), *def)
+                })
+                .collect();
```
hashdotai, line 482:
```suggestion
    .flat_map(|(name, alias, def)| {
        self.alloc_intrinsic_value(id.value(), name, alias.iter().copied(), *def)
    })
    ```

    You're using `id.value()` here, but in other places like line 375 you're using `id` directly. It would be better to be consistent across all the methods.

File: libs/@local/hashql/core/src/module/namespace.rs [hunk 5]
```diff
@@ -335,48 +333,51 @@ impl<'env, 'heap> ModuleNamespace<'env, 'heap> {
...
-        successful &= self.import_absolute_static("^", ["math", "bit_xor"]);
-        successful &= self.import_absolute_static("~", ["math", "bit_not"]);
-        successful &= self.import_absolute_static("<<", ["math", "bit_shl"]);
-        successful &= self.import_absolute_static(">>", ["math", "bit_shr"]);
+        successful &= self.import_absolute_static("&", ["core", "bits", "and"]);
+        successful &= self.import_absolute_static("|", ["core", "bits", "or"]);
+        successful &= self.import_absolute_static("^", ["core", "bits", "xor"]);
+        successful &= self.import_absolute_static("~", ["core", "bits", "not"]);
+        successful &= self.import_absolute_static("<<", ["core", "bits", "shl"]);
+        successful &= self.import_absolute_static(">>", ["core", "bits", "shr"]);
```
hashdotai, line 367: The restructuring of operators into their logical categories (math, compare, bits, bool) makes the code much more maintainable and follows better module organization principles. Good job on this reorganization.

libs/@local/hashql/core/src/module/std_lib.rs
indietyp: why? this is shrimply a stand-in until we have a real module system, until then the path is not expected to change.

TimDiekmann (approved):

PR #7340, opened 2025-06-05
H-4724: HashQL: track intrinsics through assignment

indietyp:

We currently have no way to track intrinsics through assignments. This is relatively straightforward and is simply an additional `HashMap` during inference.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -39,8 +39,15 @@ use crate::{
     visit::{self, Visitor},
 };
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Local<'heap> {
```
hashdotai, line 43: Consider adding a comment here explaining why we need to track intrinsics through assignments and what problem this solves. This would make it clearer for future developers looking at this code.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -39,8 +39,15 @@ use crate::{
...
+    pub r#type: TypeDef<'heap>,
```
hashdotai, line 44: The naming here is slightly confusing since `Local` now contains a `type` field. Consider using a more descriptive name for the struct or a more specific name for the field to avoid potential confusion between the container and its contents.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -159,7 +171,10 @@ impl<'heap> Visitor<'heap> for TypeInference<'_, 'heap> {
     fn visit_local_variable(&mut self, variable: &'heap LocalVariable<'heap>) {
         visit::walk_local_variable(self, variable);
 
-        let mut def = self.locals[Universe::Value][&variable.name.value];
+        let Local {
+            r#type: mut def,
+            intrinsic,
+        } = self.locals[Universe::Value][&variable.name.value];
```
hashdotai, line 177: There seems to be some unnecessary map traversals happening here. Consider getting both the intrinsic and type definition in a single lookup operation.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -169,6 +184,11 @@ impl<'heap> Visitor<'heap> for TypeInference<'_, 'heap> {
 
         self.types
             .insert_unique(Universe::Value, self.current, r#type);
+
+        if let Some(intrinsic) = intrinsic {
```
hashdotai, line 188: Consider adding a brief comment explaining how the intrinsic is propagated through variable assignments. This would help clarify the intent behind this change.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -54,10 +61,10 @@ pub struct TypeInference<'env, 'heap> {
     current: HirId,
 
     visited: FastHashSet<HirId>,
-    locals: FastRealmsMap<Symbol<'heap>, TypeDef<'heap>>,
+    locals: FastRealmsMap<Symbol<'heap>, Local<'heap>>,
     types: FastRealmsMap<HirId, TypeId>,
     arguments: FastRealmsMap<HirId, Interned<'heap, [GenericArgumentReference<'heap>]>>,
-    variables: FastRealmsMap<HirId, hashql_core::r#type::inference::Variable>,
+    intrinsics: FastRealmsMap<HirId, &'static str>,
```
hashdotai, line 67: You've removed `variables` from the struct, but it seems the functionality has been replaced by `intrinsics`. Consider adding a comment explaining this change to make it clearer for other developers why this field was removed.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -375,10 +398,9 @@ impl<'heap> Visitor<'heap> for TypeInference<'_, 'heap> {
         // For closure literals we invert the direction and collect `C <: F`
         self.inference.collect_constraints(function, closure);
 
-        self.variables
-            .insert_unique(Universe::Value, self.current, returns);
         self.types
             .insert_unique(Universe::Value, self.current, returns_id);
+
```
hashdotai, line 403: This previously had the `variables` assignment that was removed. I notice there are other places where similar code was removed without replacement. Could you add a comment explaining why these variables assignments are no longer needed?

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -39,8 +39,15 @@ use crate::{
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Local<'heap> {
+    pub r#type: TypeDef<'heap>,
+    pub intrinsic: Option<&'static str>,
+}
+
 pub struct TypeInferenceResidual<'heap> {
-    pub locals: FastHashMap<Symbol<'heap>, TypeDef<'heap>>,
+    pub locals: FastHashMap<Symbol<'heap>, Local<'heap>>,
+    pub intrinsics: FastHashMap<HirId, &'static str>,
```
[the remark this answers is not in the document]
indietyp: instrinsics are rn just static strings, in the future they will be replaced with symbols, once we actually define them in modules.

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: ?

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: ?

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: I think that's pretty clear, no?

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: because it's unused dummy

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: no

PR #7350, opened 2025-06-06
H-4726: Fix inconsistencies between `viewer` relationship and `viewEntity` policy action

TimDiekmann:

A few inconsistencies are remaining between the old SpiceDB based policy engine and the new engine:
- `incurred-in` entities should have the same policies as `usage-record` and `records-usage-of`
    - Web-admins should have access to these as well
- When manually creating relationships, policies are not updated (i.e. `viewer` permissions`
- synchronized linear link-entities get a permission, but the source entity does not
- view permissions for the system actor on
	- triggered-by-user
	- occurred-in-block
	- occurred-in-entity
	- occurred-in-comment
	- occurred-in-text

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
+++ b/apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
@@ -1,5 +1,6 @@
 import type {
   EntityId,
+  EntityUuid,
   OriginProvenance,
   ProvidedEntityEditionProvenance,
   UserId,
@@ -9,11 +10,13 @@ import { getInstanceAdminsTeam } from "@local/hash-backend-utils/hash-instance";
 import { createUsageRecord } from "@local/hash-backend-utils/service-usage";
 import type { GraphApi } from "@local/hash-graph-client";
 import { HashEntity } from "@local/hash-graph-sdk/entity";
+import { createPolicy } from "@local/hash-graph-sdk/policy";
 import type { FlowUsageRecordCustomMetadata } from "@local/hash-isomorphic-utils/flows/types";
 import { generateUuid } from "@local/hash-isomorphic-utils/generate-uuid";
 import { systemLinkEntityTypes } from "@local/hash-isomorphic-utils/ontology-type-ids";
 import { stringifyError } from "@local/hash-isomorphic-utils/stringify-error";
 import type { IncurredIn } from "@local/hash-isomorphic-utils/system-types/usagerecord";
+import type { PrincipalConstraint } from "@rust/hash-graph-authorization/types";
 // import { StatusCode } from "@local/status";
 import { backOff } from "exponential-backoff";
 
@@ -159,9 +162,6 @@ export const getLlmResponse = async <T extends LlmParams>(
           createUsageRecord(
             { graphApi: graphApiClient },
             {
-              additionalViewers: [
-                { actorType: "ai", id: aiAssistantAccountId },
-              ],
               assignUsageToWebId: webId,
               customMetadata,
               serviceName: isLlmParamsAnthropicLlmParams(llmParams)
@@ -173,6 +173,7 @@ export const getLlmResponse = async <T extends LlmParams>(
               inputUnitCount: usage.inputTokens,
               outputUnitCount: usage.outputTokens,
               userAccountId,
+              aiAssistantAccountId,
             },
           ),
         {
@@ -209,6 +210,7 @@ export const getLlmResponse = async <T extends LlmParams>(
       const errors = await Promise.all(
         incurredInEntities.map(async ({ entityId }) => {
           try {
+            const incurredInEntityUuid = generateUuid() as EntityUuid;
             await HashEntity.create<IncurredIn>(
               graphApiClient,
               { actorId: aiAssistantAccountId },
@@ -217,6 +219,7 @@ export const getLlmResponse = async <T extends LlmParams>(
                 properties: { value: {} },
                 provenance,
                 webId,
+                entityUuid: incurredInEntityUuid,
                 entityTypeIds: [
                   systemLinkEntityTypes.incurredIn.linkEntityTypeId,
                 ],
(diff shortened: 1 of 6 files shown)


apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -596,6 +605,10 @@ export const addEntityViewerResolver: ResolverFn<
   LoggedInGraphQLContext,
   MutationAddEntityViewerArgs
 > = async (_, { entityId, viewer }, graphQLContext) => {
+  if (viewer.kind !== AuthorizationSubjectKind.Public) {
```
hashdotai, line 608: This validation is being added to both `addEntityViewerResolver` and `removeEntityViewerResolver`, but the error message says "Only public viewers can be added to an entity". For the `removeEntityViewerResolver`, the message should be "Only public viewers can be removed from an entity".

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
```diff
@@ -622,6 +647,10 @@
   LoggedInGraphQLContext,
   MutationRemoveEntityViewerArgs
 > = async (_, { entityId, viewer }, graphQLContext) => {
+  if (viewer.kind !== AuthorizationSubjectKind.Public) {
+    throw new UserInputError("Only public viewers can be added to an entity");
```
hashdotai: The error message here should be "Only public viewers can be removed from an entity" since this is the `removeEntityViewerResolver`.

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -251,6 +254,38 @@ export const getLlmResponse = async <T extends LlmParams>(
...
+                id: userAccountId,
+              },
+              {
+                type: "actor",
+                actorType: "ai",
+                id: aiAssistantAccountId,
+              },
+            ];
+
+            // TODO: allow creating policies alongside entity creation
```
hashdotai, line 270: Great use of a TODO comment referencing the Linear ticket. However, this same TODO appears in multiple files. Consider creating a shared constant for the TODO message to ensure consistency if it needs to be updated.

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -251,6 +254,38 @@
...
+            //   see https://linear.app/hash/issue/H-4622/allow-creating-policies-alongside-entity-creation
+            for (const principal of viewPrincipals) {
+              await createPolicy(
+                graphApiClient,
+                { actorId: aiAssistantAccountId },
+                {
+                  name: `usage-record-view-entity-${incurredInEntityUuid}`,
```
hashdotai, line 277: This policy name uses a different format than other similar policies in the codebase (e.g., `public-view-entity-${entityUuid}` in other files). Consider standardizing the naming convention across the codebase for consistency.

libs/@local/hash-backend-utils/src/service-usage.ts
```diff
@@ -156,6 +152,7 @@ export const createUsageRecord = async (
      * Tracked separately from webId as usage may be attributed to an org, but we want to know which user incurred it.
      */
     userAccountId: UserId;
+    aiAssistantAccountId: AiId;
```
hashdotai, line 155: This parameter was previously in `additionalViewers` but is now a direct parameter. A comment explaining the expected type/format of `aiAssistantAccountId` would be helpful for maintainability, similar to other parameters in this interface.

apps/hash-api/src/graphql/resolvers/knowledge/entity/entity.ts
TimDiekmann:
```suggestion
    throw new UserInputError("Only public viewers can be removed from an entity");
```

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
TimDiekmann: it is consistent

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -251,6 +254,38 @@ export const getLlmResponse = async <T extends LlmParams>(
               },
             );
 
+            const viewPrincipals: PrincipalConstraint[] = [
+              {
+                type: "actor",
+                actorType: "user",
+                id: userAccountId,
```
CiaranMn, line 261: Is there not a 'web' equivalent yet? Because this is not quite the same as the existing permission (inherit view from web). If this is for consistency with the usage record itself, let's make them all view from web

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -91,6 +91,26 @@ fn system_actor_view_entity_policies(
...
+                    2,
+                ))
+                .chain(create_version_filters(
+                    base_url!("https://hash.ai/@h/types/entity-type/occurred-in-comment/"),
+                    1,
+                ))
+                .chain(create_version_filters(
+                    base_url!("https://hash.ai/@h/types/entity-type/occurred-in-text/"),
+                    1,
+                ))
```
CiaranMn: The system account does not currently have any permissions on these – instead, administrator permissions are given to the web bot that creates the notification. I'm not sure that is even necessary tbh, not sure what is relying on it.

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
TimDiekmann: The current viewers are:
- The web admins: https://github.com/hashintel/hash/blob/8a5757315a8fd094a9f5732b9d101677d888221f/libs/%40local/graph/postgres-store/src/store/postgres/seed_policies.rs#L240-L243
- The instance admins: https://github.com/hashintel/hash/blob/8a5757315a8fd094a9f5732b9d101677d888221f/libs/%40local/graph/postgres-store/src/store/postgres/seed_policies.rs#L296-L299
- The acting user (this line)
- the acting AI (the line below)

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
TimDiekmann: Fixed in [`8a57573` (#7350)](https://github.com/hashintel/hash/pull/7350/commits/8a5757315a8fd094a9f5732b9d101677d888221f)

apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
CiaranMn: ~Right, but I'm saying that the SpiceDB permissions are not this – the user is not granted a specific view permission, instead it's view from web~
```
                  {
                    relation: "setting",
                    subject: {
                      kind: "setting",
                      subjectId: "viewFromWeb",
                    },
                  },
```
~we can just make it view from web (and the same for the usage record) to avoid having to create additional permissions for the AI assistant.~

~Since we don't have any of these policies anyway we can change it again later in follow-up if we want.~

CiaranMn: Struck out comment as we already agreed to do something different, my bad, ignore it

PR #7351, opened 2025-06-06
H-4727: Remove usage of `viewer` and `viewFromWeb` relationships

TimDiekmann:

When the new policy engine is used, we don't need the old relationships. However, this was not done in a single PR so we have some time where both, the policies and relationships are maintained. In the case something goes wrong with the policies and we need to revert the policy PR, we still have a working permission system.

## 🚫 Blockers

- #7352 

## 🔍 What does this change?

Remove
- `viewer` and `viewFromWeb` entity relationships
- modification of entity owner/admins
- reading entity relationships

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-ai-worker-ts/src/activities/shared/get-llm-response.ts
```diff
@@ -235,21 +229,6 @@ export const getLlmResponse = async <T extends LlmParams>(
...
-                    },
-                  },
-                  {
-                    relation: "viewer",
-                    subject: {
-                      kind: "accountGroup",
-                      subjectId: hashInstanceAdminGroupId,
-                      subjectSet: "member",
-                    },
-                  },
```
CiaranMn, line 250: We were going to just comment these out, and others below, so that we can do a big check later once other Cedar permissions are implemented, that we have properly translated them (except where we decided to do something differently)

CiaranMn (dismissed):

TimDiekmann:
Superseded by https://github.com/hashintel/hash/pull/7587

PR #7352, opened 2025-06-06
H-847, H-4433: Use `viewEntity` policy over `viewer` relationships in the Graph

TimDiekmann:

I prepared everything in order to use the new `viewEntity` policy action and stop relying on SpiceDB for entity queries. This changes the implementation in the Graph.

## 🚫 Blocked by

- #7350 
- #7496 

## 🔍 What does this change?

All usages of `EntityPermission::View` has been replaced with the appropriated call to the new policy engine.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- https://github.com/hashintel/hash/pull/7351

## 🛡 What tests cover this?

Pretty much every test around entities implicitly tests this

## ❓ How to test this?

Start the application and make sure everything is still working as expected

### DIFF
--- a/libs/@local/graph/authorization/src/policies/components.rs
+++ b/libs/@local/graph/authorization/src/policies/components.rs
@@ -86,9 +86,17 @@ impl PolicyComponents {
     #[tracing::instrument(level = "info", skip(self))]
     pub fn extract_filter_policies(
         &self,
+        action: ActionName,
     ) -> impl Iterator<Item = (Effect, Option<&ResourceConstraint>)> {
+        debug_assert!(
+            self.tracked_actions.contains(&action),
+            "action `{}` not tracked",
+            action.to_string()
+        );
+
         self.policies
             .iter()
+            .filter(move |policy| policy.actions.contains(&action))
             .map(|policy| (policy.effect, policy.resource.as_ref()))
     }
 
@@ -584,12 +592,14 @@ mod tests {
                 Uuid::new_v4(),
             ))),
             policies: vec![policy],
-            tracked_actions: HashSet::new(),
+            tracked_actions: HashSet::from([ActionName::View]),
             context: Context::default(),
             optimization_data: OptimizationData::default(),
         };
 
-        let filter_policies: Vec<_> = policy_components.extract_filter_policies().collect();
+        let filter_policies: Vec<_> = policy_components
+            .extract_filter_policies(ActionName::View)
+            .collect();
 
         assert_eq!(
             filter_policies.len(),
--- a/libs/@local/graph/postgres-store/src/store/postgres/crud.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/crud.rs
@@ -17,8 +17,8 @@ use crate::store::{
 };
 
 pub struct QueryIndices<R: QueryRecordDecode, S: QueryRecordDecode> {
-    record_indices: R::Indices,
-    cursor_indices: S::Indices,
+    pub record_indices: R::Indices,
+    pub cursor_indices: S::Indices,
 }
 
 pub trait QueryRecordDecode {
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
@@ -20,8 +20,8 @@ use hash_graph_authorization::{
 };
 use hash_graph_store::{
     entity::{
-        CountEntitiesParams, CreateEntityParams, EmptyEntityTypes, EntityQueryPath,
-        EntityQuerySorting, EntityStore, EntityTypeRetrieval, EntityTypesError,
+        CountEntitiesParams, CreateEntityParams, EmptyEntityTypes, EntityQueryCursor,
+        EntityQueryPath, EntityQuerySorting, EntityStore, EntityTypeRetrieval, EntityTypesError,
         EntityValidationReport, EntityValidationType, GetEntitiesParams, GetEntitiesResponse,
         GetEntitySubgraphParams, GetEntitySubgraphResponse, PatchEntityParams, QueryConversion,
         UpdateEntityEmbeddingsParams, ValidateEntityComponents, ValidateEntityParams,
@@ -29,7 +29,7 @@ use hash_graph_store::{
     entity_type::{EntityTypeQueryPath, EntityTypeStore as _, IncludeEntityTypeOption},
     error::{InsertionError, QueryError, UpdateError},
     filter::{Filter, FilterExpression, Parameter, ParameterList},
-    query::{QueryResult as _, Read, ReadPaginated, Sorting as _},
+    query::{QueryResult as _, Read},
     subgraph::{
         Subgraph, SubgraphRecord as _,
         edges::{EdgeDirection, GraphResolveDepths, KnowledgeGraphEdgeKind, SharedEdgeKind},
@@ -88,9 +88,11 @@ use crate::store::{
     error::{DeletionError, EntityDoesNotExist, RaceConditionOnUpdate},
     postgres::{
         ResponseCountMap, TraversalContext,
+        crud::{QueryIndices, TypedRow},
         knowledge::entity::read::EntityEdgeTraversalData,
         query::{
-            InsertStatementBuilder, ReferenceTable, SelectCompiler, Table,
+            Distinctness, InsertStatementBuilder, PostgresRecord as _, PostgresSorting as _,
+            ReferenceTable, SelectCompiler, Table,
             rows::{
                 EntityDraftRow, EntityEditionRow, EntityHasLeftEntityRow, EntityHasRightEntityRow,
                 EntityIdRow, EntityIsOfTypeRow, EntityTemporalMetadataRow,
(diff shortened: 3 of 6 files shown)


libs/@local/graph/postgres-store/src/store/validation.rs
```diff
@@ -133,21 +138,74 @@ pub struct StoreCache {
 #[derive(Debug)]
 pub struct StoreProvider<'a, S> {
     pub store: &'a S,
-    pub cache: StoreCache,
-    pub authorization: Option<(ActorEntityUuid, Consistency<'static>)>,
+    pub cache: Box<StoreCache>,
```
TimDiekmann: `StoreCache` is quite big. I boxed it to not unnecessarily blow up futures.

libs/@local/graph/authorization/src/api.rs
```diff
@@ -136,7 +136,7 @@ pub trait AuthorizationApi: Send + Sync {
     ) -> impl Future<Output = Result<Vec<ActorIdOrPublic>, Report<ReadError>>> + Send;
 
     fn modify_entity_relations(
-        &mut self,
+        &self,
```
TimDiekmann: `mut` was not necessary because it's an external call, it was just used in case we want to implement another API without external services. Anyway, `mut self` was annoying when reusing the Postgres store. As we're going to remove it again, just making it non-mut makes the live a bit easier.

libs/@local/graph/store/src/entity/store.rs
```diff
@@ -433,7 +434,7 @@ pub trait EntityStore {
         actor_id: ActorEntityUuid,
         consistency: Consistency<'_>,
         params: Vec<ValidateEntityParams<'_>>,
-    ) -> impl Future<Output = HashMap<usize, EntityValidationReport>> + Send;
+    ) -> impl Future<Output = Result<HashMap<usize, EntityValidationReport>, Report<QueryError>>> + Send;
```
TimDiekmann: `validate_entity` is allowed to fail if the actor is invalid. This was not covered previously. The validation itself will still always return a proper structure.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -281,15 +291,105 @@ where
...
+                            Filter::for_entity_edition_ids(&entity_edition_ids);
+                        compiler
+                            .add_filter(&entity_editions_filter)
+                            .change_context(QueryError)?;
+
+                        let permission_filter =
+                            Filter::try_from(partial).change_context(QueryError)?;
+                        compiler
+                            .add_filter(&permission_filter)
+                            .change_context(QueryError)?;
```
TimDiekmann:
```suggestion
                        // TODO: Ideally, we'd incorporate the filter in the above query, but that's
                        //       not easily possible as the query above uses features that the query
                        //       compiler does not support yet.
                        let permission_filter =
                            Filter::try_from(partial).change_context(QueryError)?;
                        compiler
                            .add_filter(&permission_filter)
                            .change_context(QueryError)?;
```

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
```diff
@@ -556,11 +560,12 @@ where
...
-            .convert_parameters(&StoreProvider {
-                store: self,
-                cache: StoreCache::default(),
-                authorization: Some((actor_id, Consistency::FullyConsistent)),
-            })
+            .convert_parameters(
+                &StoreProvider::new(self)
+                    .with_authorization(actor_id, Consistency::FullyConsistent)
+                    .await
+                    .change_context(QueryError)?,
```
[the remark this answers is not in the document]
TimDiekmann: No.

libs/@local/graph/store/src/entity/store.rs
hashdotai: @TimDiekmann This is a good improvement to the error handling. Allowing `validate_entity` to properly report failures when the actor is invalid provides better diagnostics while still maintaining the validation structure. It makes the API more robust by explicitly handling this previously uncovered case.

libs/@local/graph/authorization/src/api.rs
hashdotai: @TimDiekmann Removing the `mut` requirement from the API makes the code more flexible and reduces friction when reusing the Postgres store. This is a good change, especially since we're planning to remove this code later anyway.

libs/@local/graph/postgres-store/src/store/validation.rs
hashdotai: @TimDiekmann Boxing `StoreCache` makes sense here. It helps avoid unnecessarily large futures on the stack, which is particularly important since it would otherwise be included in every future that uses a `StoreProvider`. Good optimization.

libs/@local/graph/postgres-store/src/store/postgres/ontology/data_type.rs
```diff
@@ -89,21 +87,37 @@ where
...
+            };
 
         Ok(ids
             .into_iter()
             .zip(data_types)
             .filter_map(move |(id, data_type)| {
+                let Some(permissions) = &permissions else {
+                    return Some(data_type);
+                };
+
```
hashdotai: This implementation is simpler now that you're checking if there are permissions rather than if the permission value is true. Good improvement.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -411,359 +369,288 @@ where
...
+                    },
+                    policy_context,
                 )
-            });
-
-            for filter in &filters {
-                compiler.add_filter(filter).change_context(QueryError)?;
+                .change_context(QueryError)?
+            {
+                Authorized::Always => {}
```
hashdotai: The way you've structured the code with the pattern-matching on `Authorized` results makes the authorization logic very clear. This is a good pattern to follow for all the authorization checks.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -91,6 +91,30 @@ fn system_actor_view_entity_policies(
                     base_url!("https://hash.ai/@h/types/entity-type/prospective-user/"),
                     1,
                 ))
+                .chain(create_version_filters(
```
hashdotai: I see you're adding support for notifications in the system actor view entity policies. This seems like a good addition to ensure the notification system works properly with the new policy engine.

libs/@local/graph/postgres-store/src/store/validation.rs
```diff
@@ -133,21 +138,74 @@ pub struct StoreCache {
...
     A: AuthorizationApi,
 {
+    /// Builds the policy context and policy set for the given actor, and stores them in the
+    /// `authorization` field of the provider.
+    ///
+    /// # Errors
+    ///
+    /// If the actor cannot be determined, or if the policy context or policy set cannot be built,
+    /// this function will return an error.
+    pub async fn with_authorization(
```
hashdotai: The `StoreProvider::with_authorization` method is well-designed. It properly handles all the error cases and provides clear context for failures. The addition of policy set evaluation makes the code more robust.

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -319,6 +319,16 @@ impl<'p> Filter<'p, Entity> {
             ),
         ])
     }
+
+    #[must_use]
+    pub const fn for_entity_edition_ids(entity_edition_ids: &'p [EntityEditionId]) -> Self {
```
hashdotai: This `for_entity_edition_ids` helper function is a nice addition for working with entity edition IDs. It makes the filtering code more readable and maintainable.

libs/@local/graph/postgres-store/src/store/validation.rs
```diff
@@ -591,20 +653,35 @@
         if let Some(cached) = self.cache.entities.get(&entity_id).await {
             return cached;
         }
-        if let Some((actor_id, consistency)) = self.authorization {
-            self.store
-                .authorization_api
-                .check_entity_permission(actor_id, EntityPermission::View, entity_id, consistency)
-                .await
+
+        let mut filters = vec![Filter::for_entity_by_entity_id(entity_id)];
```
hashdotai, line 612: Using the `authorize_entity` method here is cleaner than having the permission check logic duplicated across different methods. Good refactoring.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -411,359 +372,288 @@ where
...
-            created_by_ids,
-            edition_created_by_ids,
-            type_ids,
-            type_titles,
-        ) = if params.include_count
-            || params.include_web_ids
-            || params.include_created_by_ids
-            || params.include_edition_created_by_ids
-            || params.include_type_ids
+        if let Some((actor, _consistency, ref policy_set, ref policy_context)) =
```
indietyp: will we make use of the consistency at a later date?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1172,11 +1060,14 @@ where
             .await
             .change_context(InsertionError)?;
 
-        let validator_provider = StoreProvider {
-            store: &transaction,
-            cache: store_cache,
-            authorization: Some((actor_id, Consistency::FullyConsistent)),
-        };
+        // let validator_provider = StoreProvider {
```
indietyp: should this stay commented out?

indietyp (dismissed):
minor questions otherwise lgtm

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
indietyp: consistency is used somewhere else, so the comment is moot

TimDiekmann: Not anymore, no,  but I didn't want to create a new struct which I will remove in a follow-up just to save a few destructured fields. And I preferred `_consistency` over `_` to at least see what's going on of inline hints are not available.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Removed in https://github.com/hashintel/hash/pull/7352/commits/12733a765309583bd55fa59aef36e37d0c525d1b. This was the reason why I removed the `mut` in https://github.com/hashintel/hash/pull/7352#discussion_r2132034191 so we can re-use the already existing provider.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -27,7 +27,7 @@ use hash_graph_store::{
     entity_type::{EntityTypeQueryPath, EntityTypeStore as _, IncludeEntityTypeOption},
     error::{InsertionError, QueryError, UpdateError},
     filter::{Filter, FilterExpression, Parameter, ParameterList},
-    query::{QueryResult as _, Read, ReadPaginated, Sorting as _},
+    query::{QueryResult as _, Read, ReadPaginated},
```
hashdotai: You're removing the `Sorting` import from this line but it's not used elsewhere. This suggests the code no longer needs sorting capabilities in this module, which seems significant. Is this deliberate and is there a replacement for this functionality?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -385,280 +385,214 @@ where
...
-            created_by_ids,
-            edition_created_by_ids,
-            type_ids,
-            type_titles,
-        ) = if params.include_count
-            || params.include_web_ids
-            || params.include_created_by_ids
-            || params.include_edition_created_by_ids
-            || params.include_type_ids
+        let mut filters = vec![params.filter];
```
hashdotai: This change to use a `vec!` to collect filters is much cleaner than the previous fixed-size array approach. It makes the code more flexible for handling dynamic filtering conditions.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -385,280 +385,214 @@ where
...
-            let type_ids_idx = (params.include_type_ids || params.include_type_titles).then(|| {
-                (
-                    compiler.add_selection_path(&EntityQueryPath::TypeBaseUrls),
-                    compiler.add_selection_path(&EntityQueryPath::TypeVersions),
-                )
-            });
-
-            for filter in &filters {
-                compiler.add_filter(filter).change_context(QueryError)?;
+            Authorized::Always => {}
```
hashdotai: I like the pattern of handling the different `Authorized` states here. It's clear and makes the authorization logic easy to understand and maintain.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -385,280 +385,214 @@ where
...
+            &params.sorting,
+            params.limit,
+            params.include_drafts,
+        )
+        .await?;
+        let root_entities = rows
+            .into_iter()
+            .map(|row| (row.decode_record(&artifacts), row))
+            .collect::<Vec<_>>();
+        let cursor = if params.limit == Some(root_entities.len()) {
```
hashdotai: The cursor logic has been significantly simplified here. Instead of the complex loop that was trying to collect all entities and then filter by permission, you're now directly applying the permission filter at the query level. This should be more efficient.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1500,39 +1438,38 @@ where
...
+                &policy_components.context,
             )
-            .instrument(tracing::trace_span!("post_filter_entities"))
-            .await
             .change_context(QueryError)?
-            .0
-            .into_iter()
-            .filter_map(|(entity_id, has_permission)| has_permission.then_some(entity_id))
-            .collect::<HashSet<_>>();
+        {
```
hashdotai: The new pattern for handling permission evaluation is more concise than the previous approach and applies the filter directly in the query. This should lead to better performance since filtering happens at the database level rather than in application code.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1500,39 +1438,38 @@ where
...
+            Authorized::Always => {}
+            Authorized::Never => filters.push(Filter::Any(Vec::new())),
```
hashdotai: This approach of pushing an empty filter with `Filter::Any(Vec::new())` is a clever way to ensure no results are returned when authorization fails completely. It's more efficient than fetching results and then filtering them out.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1547,23 +1484,27 @@ where
...
+                    resource: Some(&PartialResourceId::Entity(None)),
+                    context: RequestContext::default(),
+                },
+                &policy_components.context,
             )
-            .await
             .change_context(QueryError)?
-            .assert_permission()
-            .change_context(QueryError)?;
+        {
```
hashdotai: The use of pattern matching for `Authorized` variants here is consistent with the approach used elsewhere in the codebase. This consistency improves readability and maintainability.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -1,32 +1,43 @@
 use alloc::borrow::Cow;
 use core::mem::swap;
+use std::collections::HashSet;
```
hashdotai, line 3: This HashSet import will be useful for the new permission filtering approach. Good addition.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -286,52 +292,102 @@ where
...
+                        resource: Some(&PartialResourceId::Entity(None)),
+                        context: RequestContext::default(),
+                    },
+                    &policy_components.context,
+                )
+                .change_context(QueryError)?
+            {
+                Authorized::Always => None,
+                Authorized::Never => Some(HashSet::new()),
+                Authorized::Partial(partial) => {
```
hashdotai: This new approach for dealing with entity permissions in the edges query is much more elegant. Instead of checking permissions after fetching the data, you're applying authorization filters directly to the query.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -286,52 +292,102 @@ where
...
+                    let entity_editions_filter =
+                        Filter::for_entity_edition_ids(&entity_edition_ids);
+                    compiler
+                        .add_filter(&entity_editions_filter)
+                        .change_context(QueryError)?;
+
+                    // TODO: Ideally, we'd incorporate the filter in the above query, but that's
+                    //       not easily possible as the query above uses features that the query
+                    //       compiler does not support yet.
+                    let permission_filter = Filter::try_from(partial).change_context(QueryError)?;
```
hashdotai: The pattern used here to evaluate permissions and create filters is consistent with other parts of the codebase. This consistency is good for maintainability.

libs/@local/graph/store/src/filter/mod.rs
hashdotai: This is a good utility function that makes it easy to filter by entity edition IDs. It follows the same pattern as other filter functions in this module.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: The `Sorting` trait was used to update the `Cursor`. As we don't loop anymore, we don't need the `Sorting::set_cursor` call anymore.
Good spot, though!

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -383,285 +385,225 @@
...
+            policy_components.optimization_data(),
+        );
 
-            let (statement, parameters) = compiler.compile();
+        let mut compiler = SelectCompiler::new(Some(temporal_axes), params.include_drafts);
+        compiler
+            .add_filter(&policy_filter)
+            .change_context(QueryError)?;
+        compiler
+            .add_filter(dbg!(&params.filter))
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security You're right, thank you for catching this. The `debug_assert!` is being used appropriately here as it verifies a condition that should always be true during development but doesn't need to be checked in production (since it would add overhead). It's not using the `dbg!` macro which is meant for temporary debugging output, so this is actually the correct approach for assertions that should only be verified during development.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -86,9 +86,17 @@ impl PolicyComponents {
     #[tracing::instrument(level = "info", skip(self))]
     pub fn extract_filter_policies(
         &self,
+        action: ActionName,
```
hashdotai, line 89: Great addition of an `action` parameter to `extract_filter_policies`. This ensures that only policies relevant to a specific action are considered, which should improve performance by filtering policies early.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -86,9 +86,17 @@
...
     ) -> impl Iterator<Item = (Effect, Option<&ResourceConstraint>)> {
+        debug_assert!(
+            self.tracked_actions.contains(&action),
+            "action `{}` not tracked",
+            action.to_string()
+        );
+
         self.policies
             .iter()
+            .filter(move |policy| policy.actions.contains(&action))
```
hashdotai, line 99: This filter to only include policies containing the specified action is a key part of the optimization. Now we're only extracting policies that are relevant to the current operation.

libs/@local/graph/postgres-store/src/store/postgres/crud.rs
```diff
@@ -17,8 +17,8 @@ use crate::store::{
 };
 
 pub struct QueryIndices<R: QueryRecordDecode, S: QueryRecordDecode> {
-    record_indices: R::Indices,
-    cursor_indices: S::Indices,
+    pub record_indices: R::Indices,
```
hashdotai, line 20: Making these fields public is a good decision as it allows for direct access to the indices without needing accessor methods. This improves composability when building more complex queries.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -383,285 +385,225 @@ where
...
-                (
-                    compiler.add_selection_path(&EntityQueryPath::TypeBaseUrls),
-                    compiler.add_selection_path(&EntityQueryPath::TypeVersions),
-                )
-            });
-
-            for filter in &filters {
-                compiler.add_filter(filter).change_context(QueryError)?;
-            }
+        let policy_filter = Filter::for_policies(
```
hashdotai, line 406: This is a key improvement - creating a policy filter directly from extracted policies rather than making separate API calls. This should significantly improve performance by eliminating the need for extra roundtrips to check permissions.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -383,285 +385,225 @@
...
-                            entity_uuid: row.get(entity_uuid_idx),
-                            draft_id: row.get(draft_id_idx),
-                        },
-                        row,
-                    )
-                })
-                .try_collect::<HashMap<_, _>>()
-                .await
-                .change_context(QueryError)?;
+        let (count, web_ids, created_by_ids, edition_created_by_ids, type_ids, type_titles) =
```
hashdotai, line 420: The restructuring of this code makes it more efficient. Previously, permissions were checked after fetching entities; now permission filters are applied directly in the query.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -383,285 +385,225 @@
...
-                .into_iter()
-                .map(|row| (row.decode_record(&artifacts), row))
-                .collect::<Vec<_>>();
-            if let Some(cursor) = entities
-                .last()
-                .map(|(_, row)| row.decode_cursor(&artifacts))
-            {
-                params.sorting.set_cursor(cursor);
-            }
+        if let Some(limit) = params.limit {
```
hashdotai, line 580: This approach of applying pagination limits at the SQL level is more efficient than the previous implementation which fetched potentially more records than needed and then filtered them.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -383,285 +385,225 @@
...
-            root_entities.extend(entities.into_iter().filter(|(entity, _)| {
-                permitted_ids.contains(&entity.metadata.record_id.entity_id.entity_uuid)
-            }));
+        let rows = self
+            .as_client()
+            .query(&statement, parameters)
+            .instrument(tracing::trace_span!("query"))
+            .await
+            .change_context(QueryError)?;
+        let artifacts = QueryIndices::<Entity, EntityQuerySorting> {
```
hashdotai, line 605: Using `QueryIndices` to manage both record and cursor indices is a cleaner approach than the previous implementation. It centralizes the mapping logic and makes the code more maintainable.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1525,39 +1471,31 @@
             .await
             .change_context(QueryError)?;
 
+        let policy_filter = Filter::for_policies(
```
hashdotai, line 1499: This is a significant improvement for the `count_entities` method. Using `policy_filter` directly in the query instead of fetching all entities and then filtering them should be much more efficient, especially with large datasets.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1569,26 +1507,18 @@
...
-                    .map_or_else(ActorEntityUuid::public_actor, ActorEntityUuid::from),
-                EntityPermission::View,
-                entity_id,
-                Consistency::FullyConsistent,
-            )
-            .await
-            .change_context(QueryError)?
-            .assert_permission()
-            .change_context(QueryError)?;
+        let filter = Filter::for_policies(
```
hashdotai, line 1545: Similar improvement for `get_entity_by_id` - applying the policy filter directly in the query instead of making a separate permission check afterward eliminates unnecessary API calls.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -286,52 +290,86 @@ where
...
-                        Consistency::FullyConsistent,
-                    )
-                    .await
-                    .change_context(QueryError)?
-                    .0,
-            )
-        } else {
-            None
-        };
+        let permitted_entity_edition_ids =
```
hashdotai, line 312: This refactoring of the permission checking for edge traversal is much better. Instead of fetching all entities and then filtering by permission, you're now applying permission filters directly in the query.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -286,52 +290,86 @@
...
-                return true;
+                let entity_editions_filter = Filter::for_entity_edition_ids(&entity_edition_ids);
+                compiler
+                    .add_filter(&entity_editions_filter)
+                    .change_context(QueryError)?;
+
+                // TODO: Ideally, we'd incorporate the filter in the above query, but that's
+                //       not easily possible as the query above uses features that the query
+                //       compiler does not support yet.
+                let permission_filter = Filter::for_policies(
```
hashdotai, line 339: Good use of the policy filter to restrict results at the database level. The TODO comment acknowledges future potential optimization, which is a helpful note for maintaining this code.

libs/@local/graph/postgres-store/src/store/postgres/query/compile.rs
```diff
@@ -398,12 +398,12 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
     }
 
     /// Adds a new filter to the selection.
-    pub fn add_filter(
+    pub fn add_filter<'f: 'q>(
```
hashdotai, line 405: Relaxing the lifetime constraints here (`'f: 'q` instead of requiring exactly `'q`) makes the filter method more flexible, allowing it to accept filters with different lifetimes as long as they live at least as long as needed.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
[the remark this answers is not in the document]
TimDiekmann: You commented on a fixed lint...

TimDiekmann:
VSCode being stupid with `git` rebasing...

PR #7354, opened 2025-06-06
H-4731, H-4732: HashQL: Improve readability of HIR snapshot output

indietyp:

The current HIR snapshot output is quite messy, in the sense that it makes it hard to decipher. We need to change that.

To show the effectiveness of the new snapshot, it has allowed me to find and isolate H-4732 (as it is a single line change has been included here).

---

The resulting type of a HIR `let x = <value> in <body>` node is currently that of the value and not of the body. This is incorrect. The resulting type of that expression should be the body. A let binding simply binds a value in a scope and during evaluation "vanishes".

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/compiletest/src/suite/common.rs
```diff
@@ -30,3 +32,202 @@ where
...
+
+        let len = self.0.len();
+
+        for _ in 0..4 {
+            fmt.write_char(HEADER)?;
+        }
+
+        write!(fmt, " {} ", self.0)?;
+
+        let remaining = (80_usize - 4).saturating_sub(len + 2);
```
hashdotai, line 56: The `Header` struct has a fixed width of 80 characters. Consider making this configurable or at least a constant at the top of the file for easier maintenance.

libs/@local/hashql/compiletest/src/suite/common.rs
```diff
@@ -30,3 +32,202 @@ where
...
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+        struct Prefix {
+            start: &'static str,
+            continuation: &'static str,
+        }
+
+        impl Prefix {
+            const ANNOTATION: Self = Self {
+                start: "└→ ",
```
hashdotai, line 86: The Unicode box drawing characters here might cause issues in environments with limited font support. Consider adding a fallback option using ASCII characters for environments where Unicode isn't well supported.

libs/@local/hashql/compiletest/src/suite/common.rs
```diff
@@ -30,3 +32,202 @@ where
...
+        let mut writer = Writer::new(fmt, Prefix::CONTENT);
+
+        let mut formatter = fmt::Formatter::new(&mut writer, options);
+        self.content.fmt(&mut formatter)?;
+        writer.write_str("\n")?;
+
+        writer.reset(Prefix::ANNOTATION);
+
+        let mut formatter = fmt::Formatter::new(&mut writer, options);
+        self.annotation.fmt(&mut formatter)?;
```
hashdotai, line 163: Consider adding whitespace around the type annotation display for better readability, e.g., `└→ Type: Number` instead of just `└→ Number`.

libs/@local/hashql/hir/tests/ui/lower/alias-replacement/complex-mixed.stdout
```diff
@@ -1,12 +1,14 @@
+════ Initial HIR ═══════════════════════════════════════════════════════════════
```
hashdotai, line 1: The Header string is hard-coded to all caps. Consider adding a parameter to control the case or consistently use Title Case for all headers.

libs/@local/hashql/compiletest/src/suite/common.rs
indietyp: ?

libs/@local/hashql/compiletest/src/suite/common.rs
indietyp: we're in 2025. i think that's fine.

libs/@local/hashql/compiletest/src/suite/common.rs
indietyp: just used for testing. no need to overengineer (and that's coming from me)

TimDiekmann (approved):

PR #7374, opened 2025-06-09
H-4734: HashQL: Modularize stdlib

indietyp:

The current way that the stdlib is setup makes any addition untenable. Especially considering dependencies across module boundaries. The current structure is both verbose and cannot easily be extended to have dependencies between modules.

There are currently three ways around this problem:

1. patch work the current implementation
2. create a mini-language (mini-hashql, akin to mini-rust) that is used only for the purpose of parsing the stdlib.
3. restructure the stdlib to make use of traits and dependencies instead.

I tried to make (1) work, but it's just makes the whole system a lot more unreadable. I then considered (2) and have disregarded it, as it would add (a) runtime complexity and (b) would add a whole new parser - granted said parser would not need to very resilient because the inputs rarely change - but still an additional parser is an additional parser.

I then settled on (3) instead of having a giant `StandardLibrary` struct we have a set of traits that are implemented:

```rust
enum ModuleKind { Newtype(TypeId), Type(TypeId), Intrinsic(IntrinsicItem) };
struct ModuleItem { name: Symbol<'heap>, kind: ModuleKind }

struct ModuleContents { items: SmallVec<ModuleItem> }

trait StandardLibraryModule {
  type Requires: StdLibModRequirement; // HList / Tuple
  type Children: StdLibModChildren; // HList / Tuple

  fn create(context: &mut StdLibContext, dependencies: Self::Requires::Output) -> ModuleContents;
}
```

The module items are then placed in places similar to the their actual definition, this makes it easy to map module definition to the real module.

Unlike the "real" module system, which has a separation of the universes into realms this doesn't happen in here, instead the module item (trading flexibility) can just be one of few choices that then is converted into a real item.

Also important to note: The module is assembled *outside* of the actual definition. This removes quite a bit of nesting and allows us to focus on what matters. It also makes the requirements work, as it allows us to access the module items outside of the hierarchy.

Items are in a `Vec` instead of a `HashMap` because we only expect a few amount of items, in which case a linear scan is simply faster.

---

The mini-hashql syntax would've looked like this:

```
// -------------------------- //
// core/math.hashql           //
// -------------------------- //

use ::kernel::type;

// bind is a poor-mans alias, makes it easier to parse, 
// items can otherwise only be mentioned by their absolute path
#[bind(::kernel::type::Number)]
type Number;

#[intrinsic(::core::math::add)]
#[alias(`+`)]
fn add<T: Number, U: Number>(lhs: T, rhs: U) -> T | U;

// -------------------------- //
// kernel/special_form.hashql //
// -------------------------- //

#[intrinsic(::kernel::special_form::if)]
fn if() -> !;

// -------------------------- //
// kernel/type.hashql         //
// -------------------------- //

#[primitive(Integer)] // similar-ish to #[rustc_doc_primitive = "never"]
type Integer;

#[intrinsic(::kernel::type::List)] // similar to #[rustc_intrinsic]
type List;

// -------------------------- //
// core/result.hashql         //
// -------------------------- //

newtype Ok<T> = T;
newtype Err<E> = E;

type Result<T, E> = Ok<T> | Err<E>;
```

resembling rust for most things, but lacking most of the nice things.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/module/std_lib/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+            #[expect(non_snake_case)]
+            let ${concat($generic, _arg)} = $context.ty.fresh_argument(stringify!($generic));
+            #[expect(non_snake_case)]
+            let ${concat($generic, _ref)} = $context.ty.hydrate_argument(${concat($generic, _arg)});
+            #[expect(non_snake_case)]
+            let $generic = $context.ty.param(${concat($generic, _arg)});
+        )*
+
+        let mut closure = $context.ty.closure([$($param_bound),*], $return);
+        if ${count($generic)} > 0 {
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer I agree that adding more comments to the `decl!` macro would improve maintainability. The macro is using some complex template expansion patterns with interpolations like `${concat($generic, _arg)}` that aren't immediately obvious to readers unfamiliar with macro metaprogramming.

Your suggested comments do a great job explaining how the generic parameters are processed, hydrated, and used to build the closure type. This makes it much clearer how the macro handles the creation of generic function types.

libs/@local/hashql/core/src/module/std_lib/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+        impl_submodules!(@expand @($($preceding)* $item); $($rest),*);
+        impl_submodules!(@impl; $($preceding)* $item);
+    };
+
+    (@impl; $($item:ident)*) => {
+        #[expect(non_snake_case, clippy::min_ident_chars)]
+        impl<'heap, $($item),*> Submodules<'heap> for ($($item,)*)
+        where
+            $($item: StandardLibraryModule<'heap>,)*
+        {
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Good point! Adding documentation for the macro template expansion would be helpful for future maintainers. The `${count($item)}` syntax is a special macro template expansion syntax that calculates the number of items passed to the macro and sets it as the `LENGTH` constant. This is useful context that should be documented.

libs/@local/hashql/core/src/module/std_lib/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+            let $generic = $context.ty.param(${concat($generic, _arg)});
+        )*
+
+        let mut closure = $context.ty.closure([$($param_bound),*], $return);
+        if ${count($generic)} > 0 {
+            closure = $context.ty.generic([
+                $(
+                    (${concat($generic, _arg)}, None $(.or(Some($generic_bound)))?)
+                ),*
+            ] as [(crate::r#type::kind::generic::GenericArgumentId, Option<crate::r#type::TypeId>); ${count($generic)}], closure)
```
hashdotai, line 326: Consider adding an explicit cast type annotation in the `impl_submodules!` macro for clarity, especially when working with complex generics. While the current code works, it's more readable with explicit type annotations for future maintainers:

```suggestion
            as [(crate::r#type::kind::generic::GenericArgumentId, Option<crate::r#type::TypeId>); ${count($generic)}] as _
```

libs/@local/hashql/core/src/module/std_lib/core/result.rs
```diff
@@ -0,0 +1,56 @@
...
+        def.push(
+            lib.heap.intern_symbol("Ok"),
+            ItemDef::newtype(lib.ty.env, ok_ty, &[t_ref]),
+        );
+
+        // newtype Err<E> = E;
+        let err_ty = lib.ty.generic(
+            [(e_arg, None)],
+            lib.ty.opaque("::core::result::Err", e_param),
+        );
```
hashdotai, line 41: The new modular organization significantly improves maintainability, but there's a small consistency issue in how error type parameters are handled here vs. in other modules. In `Option` module you use `t_param`, but here you're using the `TypeDef` directly. Consider using the same pattern across all modules for better consistency:

```suggestion
        // newtype Err<E> = E; 
        let err_ty = lib.ty.generic(
            [(e_arg, None)],
            lib.ty.opaque("::core::result::Err", e_param),
        );
```

libs/@local/hashql/core/src/module/std_lib/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn alias(&mut self, index: usize, alias: Symbol<'heap>) -> usize {
+        let item = self.0[index].alias(alias);
+
+        let index = self.0.len();
+        self.0.push(item);
+        index
+    }
+
+    fn find(&self, name: Symbol<'heap>) -> Option<ModuleEntry<'heap>> {
+        self.0.iter().find(|item| item.name == name).copied()
```
hashdotai, line 112: Simplify this find implementation with a slightly more idiomatic pattern:

```suggestion
        self.0.iter().find_map(|item| (item.name == name).then_some(*item))
```

libs/@local/hashql/core/src/module/std_lib/mod.rs
indietyp: what?

libs/@local/hashql/core/src/module/std_lib/core/result.rs
indietyp: what?

libs/@local/hashql/core/src/module/std_lib/mod.rs
indietyp: what

TimDiekmann (approved):

PR #7375, opened 2025-06-09
H-4733: HashQL: Specify initial graph stdlib

indietyp:

Specify and implement the first version of the graph stdlib, which only focuses on the minimal useful set required (for now).

Additionally, to catch cycles on the type level, rather than during e.g. tests each module has a `_dependencies` field. Any time we request a specific manifest of things we add it to the `_dependencies` field. If the code no longer compiles we have a cyclic error. This allows to get rid of any cyclic tracking, as it's guaranteed in the type system that there are no cycles.

> Holy moly was this... something

As it doesn't make sense to duplicate all the types, we should look into: https://linear.app/hash/issue/H-4735/hashql-convert-rust-types-into-hashql-types, it'd allow us to simply derive large parts of the std from the existing types, without duplicating efforts.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/module/std_lib/core/option.rs
```diff
@@ -2,9 +2,23 @@ use crate::{
...
     symbol::Symbol,
+    r#type::TypeId,
 };
 
-pub(in crate::module::std_lib) struct Option;
+const NONE_ABSOLUTE_PATH: &str = "::core::option::None";
+const SOME_ABSOLUTE_PATH: &str = "::core::option::Some";
+
+// create a concrete monomorphized instance of `Option`
+pub(in crate::module::std_lib) fn option(lib: &StandardLibrary<'_, '_>, value: TypeId) -> TypeId {
```
hashdotai, line 12: This helper function is a great addition for creating Options without duplicating code. Consider adding a similar helper for `Result` types in a future PR.

libs/@local/hashql/core/src/module/std_lib/core/graph/entity.rs
```diff
@@ -0,0 +1,75 @@
...
+        let versioned_url_ty = lib
+            .manifest::<std_lib::core::graph::types::ontology::Ontology>()
+            .expect_newtype(heap.intern_symbol("VersionedUrl"));
+
+        let json_path_ty = lib
+            .manifest::<std_lib::core::json::Json>()
+            .expect_type(heap.intern_symbol("JsonPath"));
+
+        // `is_of_type<T>(entity: Entity<T>, depth: Integer, type: VersionedUrl) -> Boolean`
+        // see: https://linear.app/hash/issue/H-4741/hashql-support-for-type-guards
```
hashdotai, line 47: Good future-proofing with the commented code that describes how this will evolve with type guards. Just adding a note to make sure that when H-4741 lands, you update this function signature.

libs/@local/hashql/core/src/module/std_lib/core/graph/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        // that the user cannot mention.
+        let t_arg = lib.ty.fresh_argument("T");
+        let t_ref = lib.ty.hydrate_argument(t_arg);
+        let t_param = lib.ty.param(t_arg);
+
+        let graph_ty = lib.ty.generic(
+            [t_arg],
+            lib.ty.opaque(
+                "::core::graph::Graph",
+                lib.ty.r#struct([("'marker", t_param)]),
```
hashdotai, line 53: This comment about data not being accessible by the user is important. Consider adding more documentation about this pattern when introducing opaque types with internal markers, as it might not be immediately clear to other developers why we're using this approach.

libs/@local/hashql/core/src/module/std_lib/core/option.rs
```diff
@@ -23,17 +37,16 @@
...
             ItemDef::newtype(lib.ty.env, none_ty, &[]),
         );
 
         // newtype Some<T> = T;
-        let some_ty = lib.ty.generic(
-            [(t_arg, None)],
-            lib.ty.opaque("::core::option::Some", t_param),
-        );
+        let some_ty = lib
+            .ty
```
hashdotai, line 48: You've simplified this code significantly compared to the previous implementation - using `lib.ty.generic()` with a chained approach is more readable.

libs/@local/hashql/core/src/module/std_lib/mod.rs
```diff
@@ -112,14 +114,31 @@ impl<'heap> ModuleDef<'heap> {
...
     }
 
-    #[expect(dead_code, reason = "follow up PR")]
+    #[track_caller]
     fn expect(&self, name: Symbol<'heap>) -> ModuleEntry<'heap> {
         self.find(name).expect("module item not found")
     }
+
+    #[track_caller]
+    fn expect_type(&self, name: Symbol<'heap>) -> TypeDef<'heap> {
```
hashdotai, line 123: These `expect_*` helper methods are useful additions to the codebase, but it looks like you're using direct pattern matching on ItemDef variants. Consider making this more robust with a match guard or separate handling for unexpected cases to provide better error messages.

libs/@local/hashql/core/src/module/std_lib/core/graph/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        def.push(
+            lib.heap.intern_symbol("Graph"),
+            // We don't push `newtype`, but `type` as the `Graph` is not purposefully constructible
+            // by the user
+            ItemDef::r#type(lib.ty.env, graph_ty, &[t_ref]),
+        );
+
+        // newtype TimeAxis = (:)
+        // ^ this will change in the future to be a constructible type
+        //   see: https://linear.app/hash/issue/H-4736/hashql-make-time-axis-constructible
```
hashdotai: Good note about making `TimeAxis` constructible in the future with H-4736. Consider adding a `// TODO: H-4736` tag to make it easier to find this comment when implementing that ticket.

libs/@local/hashql/core/src/module/std_lib/core/graph/types/mod.rs
```diff
@@ -0,0 +1,33 @@
+// This is in a separate module, to facilitate: https://linear.app/hash/issue/H-4735/hashql-convert-rust-types-into-hashql-types
```
hashdotai, line 1: This comment about H-4735 is helpful, but it would be even better to standardize how you reference Linear tickets in comments. Consider using a format like `// TODO(H-4735): Convert Rust types into HashQL types` throughout the codebase.

libs/@local/hashql/core/src/module/std_lib/core/option.rs
indietyp: if i'll need it yes.

indietyp:
follow up: https://linear.app/hash/issue/H-4744/hashql-add-documentation-about-the-standardlibrary-and-its

libs/@local/hashql/core/src/module/std_lib/core/graph/tmp.rs
```diff
@@ -0,0 +1,45 @@
...
+        heap.intern_symbol("tmp")
+    }
+
+    fn define(lib: &mut StandardLibrary<'_, 'heap>) -> ModuleDef<'heap> {
+        let mut def = ModuleDef::new();
+        let heap = lib.heap;
+
+        let time_axis_ty = lib
+            .manifest::<std_lib::core::graph::Graph>()
+            .expect_type(heap.intern_symbol("TimeAxis"));
```
TimDiekmann:
```suggestion
            .expect_type(heap.intern_symbol("QueryTemporalAxes"));
```

libs/@local/hashql/core/src/module/std_lib/core/graph/body.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        let graph = lib.manifest::<std_lib::core::graph::Graph>();
+
+        let mut graph_param = graph.expect_type(heap.intern_symbol("Graph"));
+        let mut graph_returns = graph_param;
+
+        graph_param.instantiate(&mut lib.instantiate);
+        graph_returns.instantiate(&mut lib.instantiate);
+
+        // `filter<T>(graph: Graph<T>, predicate: fn(entity: T) -> bool) -> Graph<T>;`
```
TimDiekmann: To prevent confusing with `Entity` we should use
```suggestion
        // `filter<T>(graph: Graph<T>, predicate: fn(element: T) -> bool) -> Graph<T>;`
```
or
```suggestion
        // `filter<T>(graph: Graph<T>, predicate: fn(vertex: T) -> bool) -> Graph<T>;`
```

libs/@local/hashql/core/src/module/std_lib/core/graph/types/principal/actor_group/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn name(heap: &'heap Heap) -> Symbol<'heap> {
+        heap.intern_symbol("actor_group")
+    }
+
+    fn define(lib: &mut StandardLibrary<'_, 'heap>) -> ModuleDef<'heap> {
+        let mut def = ModuleDef::new();
+        let heap = lib.heap;
+
+        // newtype ActorGroupEntityUuid = EntityUuid;
+        // (we just set it to Uuid to avoid any cycles)
```
TimDiekmann, line 28: Could you add `TODO` or something to make it more visible? A linear ticket ID or link would be great (H-4616?)

indietyp: done in [b2c69d4](https://github.com/hashintel/hash/pull/7375/commits/b2c69d46f4b8b9286a0199c255e823602efaf1e1)

libs/@local/hashql/core/src/module/std_lib/core/graph/body.rs
indietyp: done in [b2c69d4](https://github.com/hashintel/hash/pull/7375/commits/b2c69d46f4b8b9286a0199c255e823602efaf1e1)

libs/@local/hashql/core/src/module/std_lib/core/graph/tmp.rs
indietyp: done in [b2c69d4](https://github.com/hashintel/hash/pull/7375/commits/b2c69d46f4b8b9286a0199c255e823602efaf1e1)

PR #7376, opened 2025-06-09
H-4745: Fix URL generation for entity tabs when draftId present

CiaranMn:

When viewing a page for a draft entity, the URLs generated for the tabs (e.g. Overview, History) incorrectly duplicated the `draftId` param from the URL, causing a malformed request to the API (and the URL to be wrong).

This PR fixes that.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. View a draft entity, check that switching to the History tab doesn't end up with multiple `draftId` in URL

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
@@ -72,12 +72,16 @@ const getTabUrl = (tab: string, entityPath: string) => {
     .pathname;
 
   if (tab === defaultTab) {
-    return `${entityPathWithoutParams}?${searchParams.toString()}`;
+    return `${entityPathWithoutParams}${
+      searchParams.size ? `?${searchParams.toString()}` : ""
+    }`;
   }
 
   searchParams.set("tab", tab);
 
-  return `${entityPathWithoutParams}?${searchParams.toString()}`;
+  return `${entityPathWithoutParams}${
+    searchParams.size ? `?${searchParams.toString()}` : ""
+  }`;
 };
 
 export const EntityEditorTabs = ({
--- a/apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
@@ -58,25 +58,26 @@ export const useEntityEditorTab = () => {
   return context;
 };
 
-const getTabUrl = (tab: string, entityPath: string, isInSlide: boolean) => {
+const getTabUrl = (tab: string, entityPath: string) => {
   let searchParams = new URLSearchParams();
 
-  if (!isInSlide) {
-    /**
-     * Preserve the search params (except tab) if we're on the entity's page
-     */
-    const url = new URL(window.location.href);
-    searchParams = new URLSearchParams(url.search);
-    searchParams.delete("tab");
-  }
+  /**
+   * Preserve search params except any existing tab param
+   */
+  const url = new URL(entityPath, window.location.href);
+  searchParams = new URLSearchParams(url.search);
+  searchParams.delete("tab");
+
+  const entityPathWithoutParams = new URL(entityPath, window.location.href)
+    .pathname;
 
   if (tab === defaultTab) {
-    return `${entityPath}?${searchParams.toString()}`;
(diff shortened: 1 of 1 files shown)


apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
```diff
@@ -58,25 +58,30 @@ export const useEntityEditorTab = () => {
...
-     * Preserve the search params (except tab) if we're on the entity's page
-     */
-    const url = new URL(window.location.href);
-    searchParams = new URLSearchParams(url.search);
-    searchParams.delete("tab");
-  }
+  /**
+   * Preserve search params except any existing tab param
+   */
+  const url = new URL(entityPath, window.location.href);
```
hashdotai, line 67: This line constructs a URL using `entityPath` as the URL and `window.location.href` as the base. This looks like a potential issue because `entityPath` might already be a full URL or might be a relative path that shouldn't be resolved against the current location.

```suggestion
  const url = new URL(window.location.href);
```

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
```diff
@@ -58,25 +58,30 @@
...
+  searchParams = new URLSearchParams(url.search);
+  searchParams.delete("tab");
+
+  const entityPathWithoutParams = new URL(entityPath, window.location.href)
```
hashdotai, line 71: Similar to the issue above, constructing a URL with `entityPath` as the input URL might lead to unexpected behavior if `entityPath` is already a full URL. Consider using a more reliable approach to extract just the pathname.

```suggestion
  const entityPathWithoutParams = entityPath.split('?')[0];
```

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
```diff
@@ -58,25 +58,30 @@
...
-     */
-    const url = new URL(window.location.href);
-    searchParams = new URLSearchParams(url.search);
-    searchParams.delete("tab");
-  }
+  /**
+   * Preserve search params except any existing tab param
+   */
+  const url = new URL(entityPath, window.location.href);
+  searchParams = new URLSearchParams(url.search);
```
hashdotai, line 68: If you're using `url.search` to get search params, make sure `url` is correctly constructed as mentioned in my comments above.

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
```diff
@@ -58,25 +58,30 @@
   return context;
 };
 
-const getTabUrl = (tab: string, entityPath: string, isInSlide: boolean) => {
+const getTabUrl = (tab: string, entityPath: string) => {
```
hashdotai, line 61: The function signature has changed to remove the `isInSlide` parameter, but the code still handles the `isInSlide` condition in the onClick handlers below. Is this intentional? It appears the parameter was removed but its usage remains elsewhere in the component.

CiaranMn: Yes, this is intentional. `isInSlide` is used in `onClick` to determine whether to prevent the default behavior (opening the link) or not. It is no longer used in this function.

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
CiaranMn: `entityPath` is not a full URL, it is a path, and the current window location is the correct base (it is all the same app in the same location)

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
CiaranMn: `entityPath` is not a full URL, it is a path.

apps/hash-frontend/src/pages/shared/entity/shared/entity-editor-tabs.tsx
CiaranMn: `entityPath` is not a full URL, it is a path.

PR #7387, opened 2025-06-11
H-4769: Simplify entry point for permission checks

TimDiekmann:

Currently, some quite big boilerplate is required to kick-of a permission check. This PR aims to improve that and simplifies the building of the policy context, the policy set, and determination of the actor UUID.

## 🔍 What does this change?

- Move relevant policy code to `@rust/hash-graph-authorization`
- Introduce a `PolicyComponents` struct. The name is not great but currently reflects what's in it. It can be changed later.
- Allow to dynamically build a `PolicyComponets` struct using a builder pattern which then can be `.await`ed for building
- Don't re-utilize `StoreProvider`, instead, it takes a reference to the policy components

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/authorization/src/policies/components.rs
+++ b/libs/@local/graph/authorization/src/policies/components.rs
@@ -1,0 +1,189 @@
+use alloc::borrow::Cow;
+use std::collections::HashSet;
+
+use error_stack::{Report, ResultExt as _};
+use type_system::{
+    knowledge::entity::id::EntityEditionId, ontology::VersionedUrl, principal::actor::ActorId,
+};
+
+use super::{
+    Context, ContextBuilder, PolicySet,
+    principal::actor::AuthenticatedActor,
+    store::{PolicyStore, PrincipalStore, error::ContextCreationError},
+};
+
+#[derive(Debug)]
+pub struct PolicyComponents {
+    pub actor_id: Option<ActorId>,
+    pub policy_set: PolicySet,
+    pub context: Context,
+}
+
+impl PolicyComponents {
+    #[must_use]
+    pub fn builder<S>(store: &S) -> PolicyComponentsBuilder<'_, S> {
+        PolicyComponentsBuilder::new(store)
+    }
+}
+
+pub struct PolicyComponentsBuilder<'a, S> {
+    store: &'a S,
+    actor: AuthenticatedActor,
+    context: ContextBuilder,
+    entity_type_ids: HashSet<Cow<'a, VersionedUrl>>,
+    entity_edition_ids: HashSet<EntityEditionId>,
+}
+
+impl<'a, S> PolicyComponentsBuilder<'a, S> {
+    #[must_use]
+    pub fn new(store: &'a S) -> Self {
+        Self {
+            store,
+            actor: AuthenticatedActor::Public,
+            context: ContextBuilder::default(),
+            entity_type_ids: HashSet::new(),
+            entity_edition_ids: HashSet::new(),
+        }
+    }
+
+    pub fn set_actor(&mut self, actor: impl Into<AuthenticatedActor>) {
+        self.actor = actor.into();
+    }
+
+    #[must_use]
+    pub fn with_actor(mut self, actor: impl Into<AuthenticatedActor>) -> Self {
+        self.set_actor(actor);
+        self
+    }
+
+    pub fn add_entity_type_id(&mut self, entity_type: &'a VersionedUrl) {
+        self.entity_type_ids.insert(Cow::Borrowed(entity_type));
+    }
+
+    pub fn add_entity_type_ids(&mut self, entity_type: impl IntoIterator<Item = &'a VersionedUrl>) {
+        self.entity_type_ids
+            .extend(entity_type.into_iter().map(Cow::Borrowed));
+    }
+
+    #[must_use]
+    pub fn with_entity_type_id(mut self, entity_type: &'a VersionedUrl) -> Self {
+        self.add_entity_type_id(entity_type);
+        self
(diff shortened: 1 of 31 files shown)


libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1599,6 +1725,163 @@ where
...
+                    entity_uuid: row.get(1),
+                    draft_id: row.get(2),
+                },
+                entity_type: Cow::Owned(row.get(3)),
+            })
+            .try_collect()
+            .await
+            .change_context(BuildEntityContextError::StoreError)?)
+    }
 }
```
hashdotai, line 1881: This method should have a return type annotation to make it more clear that it returns a `Future`. It's better to be explicit with these complex return types.

libs/@local/graph/postgres-store/src/store/validation.rs
```diff
@@ -410,16 +368,16 @@ where
...
+                    policy_components
+                        .actor_id
+                        .map_or_else(ActorEntityUuid::public_actor, ActorEntityUuid::from),
                     PropertyTypePermission::View,
                     type_id,
-                    consistency,
+                    Consistency::FullyConsistent,
                 )
                 .await
                 .change_context(QueryError)?
```
hashdotai, line 380: Consider adding a brief inline comment about what `Ok(())` means in this context - perhaps indicating that the validation was successful or that no permission issues were found.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -710,14 +689,22 @@ where
...
+        let policy_components = PolicyComponents::builder(&transaction)
+            .with_actor(actor_id)
+            .with_entity_type_ids(
+                params
+                    .iter()
+                    .flat_map(|params| &params.entity_type_ids)
+                    .collect::<HashSet<_>>(),
+            )
             .await
             .change_context(InsertionError)?;
```
hashdotai, line 777: Consider using a more descriptive name for the `policy_components` variable. Something like `validation_context` or `policy_validation_context` would better communicate its purpose in this validation context.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+impl PolicyComponents {
+    #[must_use]
+    pub fn builder<S>(store: &S) -> PolicyComponentsBuilder<'_, S> {
+        PolicyComponentsBuilder::new(store)
+    }
+}
+
+pub struct PolicyComponentsBuilder<'a, S> {
```
hashdotai, line 29: The `PolicyComponentsBuilder` struct could benefit from explicit lifetime documentation. Since it manages borrowed references, it would be helpful to document the relationship between the lifetimes of the store and the entity types.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -843,6 +847,128 @@
...
+            .ok_or(BuildPrincipalContextError::ActorNotFound { actor_id })?;
+        context_builder.add_actor(&actor);
+
+        let group_ids = self
+            .get_actor_roles(actor_id)
+            .await
+            .change_context(BuildPrincipalContextError::StoreError)?
+            .into_values()
+            .map(|role| {
+                context_builder.add_role(&role);
```
hashdotai, line 898: Using `impl Future` here is good for flexibility, but you might want to add a comment about the expected behavior of this future - specifically what happens when it's awaited.

libs/@local/graph/authorization/src/policies/components.rs
indietyp, line 29: Would it maybe make sense to use something like https://docs.rs/bon/latest/bon/ here instead?

libs/@local/graph/authorization/tests/policies/definitions/mod.rs
```diff
@@ -61,13 +61,16 @@ pub(crate) fn permit_admin_web(web_id: WebId, admin_role: WebRoleId) -> Vec<Poli
...
     team_id: TeamId,
-    hash_instance_entity_id: EntityUuid,
+    hash_instance_entity_id: EntityId,
 ) -> Vec<Policy> {
     #[expect(clippy::literal_string_with_formatting_args)]
     read_policies(
         &include_str!("permit-hash-instance-admins.cedar")
             .replace("{team_id}", &team_id.to_string())
-            .replace("{entity_id}", &hash_instance_entity_id.to_string()),
+            .replace(
```
indietyp, line 70: if this _ever_ becomes a bottleneck (unlikely) (as in - we're replacing 100s of times and therefore allocating a String each time) it may make sense to take a look at https://docs.rs/aho-corasick/latest/aho_corasick/#example-replacing-matches-in-a-stream to do it more efficiently or take a look at https://github.com/askama-rs/askama.

> not important, just putting the ability of aho-corasick to do replacement on our radar that is all^^

indietyp (dismissed):
just some notes / ideas. Nothing major lgtm

libs/@local/graph/authorization/tests/policies/definitions/mod.rs
TimDiekmann: This is only for seeding some test policies ;)

indietyp: whoops completely missed that 😅

indietyp (approved):

PR #7404, opened 2025-06-13
H-4777: Allow exporting types from `hash-graph-store`

TimDiekmann:

As we mix entity creation with policy creation and policies don't have a corresponding utoipa definition, we need to export the types from hash-graph-store as well.

## 🔍 What does this change?

- Adds a codegen step for `hash-graph-store`
- drive-by: Add missing `test:integration` script for the type system to check the codegen

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/graph/store/tests/codegen.rs
```diff
@@ -0,0 +1,47 @@
+extern crate hash_graph_store;
+
```
hashdotai, line 2: Consider adding a module docstring explaining the purpose of this test file, especially since it's introducing new codegen functionality.

libs/@local/graph/store/tests/codegen.rs
```diff
@@ -0,0 +1,47 @@
...
+#[test]
+fn index() -> io::Result<()> {
+    let collection = TypeCollection::default();
+
+    let settings = TypeScriptGeneratorSettings::default();
+    let mut generator = TypeScriptGenerator::new(&settings, &collection);
+
+    for (type_id, _, type_definition) in collection.iter() {
+        if type_definition.module.starts_with("hash_graph_store") {
```
hashdotai, line 18: This filter only includes types from the `hash_graph_store` module but doesn't consider submodules. If there are types in submodules like `hash_graph_store::something`, they won't be included. Consider using `starts_with` to ensure all related modules are included:

```suggestion
        if type_definition.module.starts_with("hash_graph_store") {
```

libs/@local/graph/store/types/index.snap.d.ts
```diff
@@ -0,0 +1,2 @@
+// This file was generated from `libs/@local/graph/store/tests/codegen.rs`
+
```
hashdotai, line 2: The generated .d.ts file appears to be empty except for the header comment. You should verify that type generation is working correctly, as we would expect to see actual type definitions here.

libs/@local/graph/store/Cargo.toml
```diff
@@ -36,11 +37,24 @@ tracing        = { workspace = true }
...
 postgres = ["hash-graph-temporal-versioning/postgres", "type-system/postgres", "dep:postgres-types", "dep:tokio-postgres", "dep:bytes"]
 
 [lints]
 workspace = true
+
+
+[package.metadata.sync.turborepo]
+extra-dev-dependencies = [
+    { name = "@local/tsconfig", version = "0.0.0-private" },
+]
```
hashdotai, line 60: Consider adding more detailed metadata about the package.metadata.sync.turborepo section in a comment, to help future developers understand why these extra dependencies are needed specifically for this project.

libs/@local/graph/store/Cargo.toml
```diff
@@ -36,11 +37,24 @@
 uuid           = { workspace = true, features = ["v4"] }
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros"] }
+hash-codegen = { workspace = true }
```
hashdotai, line 40: Add a comment explaining what hash-codegen is used for in dev-dependencies to make it clear it's for the type generation process.

libs/@blockprotocol/type-system/rust/package.json
```diff
@@ -29,6 +29,7 @@
     "fix:clippy": "just clippy --fix",
     "lint:clippy": "just clippy",
     "lint:tsc": "tsc --noEmit",
+    "test:integration": "cargo hack nextest run --feature-powerset --all-targets --filterset \"kind(test) | kind(bench)\" --no-tests warn",
```
TimDiekmann: Actually not required as it's covered by unit tests
```suggestion
```

libs/@local/graph/store/package.json
```diff
@@ -3,10 +3,23 @@
...
+      ]
+    }
+  },
   "scripts": {
+    "build:types": "INSTA_UPDATE=always mise exec --env dev cargo:cargo-insta -- cargo-insta test --features codegen --test codegen",
     "doc:dependency-diagram": "cargo run -p hash-repo-chores -- dependency-diagram --output docs/dependency-diagram.mmd --root hash-graph-store --root-deps-and-dependents --link-mode non-roots --include-dev-deps --include-build-deps --logging-console-level info",
     "fix:clippy": "just clippy --fix",
     "lint:clippy": "just clippy",
+    "lint:tsc": "tsc --noEmit",
+    "test:integration": "cargo hack nextest run --feature-powerset --all-targets --filterset \"kind(test) | kind(bench)\" --no-tests warn",
```
TimDiekmann: Actually not required as it's covered by unit tests
```suggestion
```

indietyp (approved):

PR #7405, opened 2025-06-13
H-4622: Allow creating policies alongside entity creation

TimDiekmann:

In a few locations we require policies to be created together with the entity. To make this an ACID operation, we can move the policy creation to the graph and create the policy within the same transaction as the entity creation.

## 🔍 What does this change?

- Add an optional `policies` field to the entity creation functions

This PR:

The changes in this PR:

The changes in this PR:


apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -98,82 +113,58 @@ const createHashEntity = async (params: {
...
-        },
-        {
-          relation: "setting",
-          subject: { kind: "setting", subjectId: "administratorFromWeb" },
-        },
-        {
-          relation: "setting",
-          subject: { kind: "setting", subjectId: "updateFromWeb" },
+    params.outgoingLinks.map(({ linkEntityTypeId, destinationEntityId }) => {
+      const linkEntityUuid = extractEntityUuidFromEntityId(entity.entityId);
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Good catch! You're right - this is extracting the parent entity's UUID for each link entity, which would cause all links to share the same UUID. This would lead to unexpected behavior where policies might apply to the wrong entities. We should be generating a new unique UUID for each link entity as you suggested.

libs/@local/graph/store/src/entity/store.rs
```diff
@@ -145,6 +158,9 @@ pub struct CreateEntityParams<R> {
     pub link_data: Option<LinkData>,
     pub draft: bool,
     pub relationships: R,
+    #[cfg_attr(feature = "utoipa", schema(value_type = Vec<Object>))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're absolutely right. Since `policies` is declared as a `Vec<CreateEntityPolicyParams>` rather than an `Option<Vec<...>>`, we should use `skip_serializing_if = "Vec::is_empty"` instead. The current attribute won't work correctly since we're checking for `Option::is_none` on a non-optional value. The suggested change will correctly skip serialization when the policies vector is empty.

hashdotai: The `skip_serializing_if = "Option::is_none"` attribute targets an `Option`, but `policies` is declared as `Vec<CreateEntityPolicyParams>`. The condition will never be true.

```suggestion
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
```

apps/hash-integration-worker/src/linear-activities.ts
hashdotai: This extracts the parent entity's UUID for each link entity, causing all links to share the same UUID. You need a unique UUID for each link entity.

```suggestion
      const linkEntityUuid = uuidv4() as EntityUuid;
```

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -682,12 +684,14 @@ where
...
 
+        let mut policies = Vec::new();
+
         let mut entities = Vec::with_capacity(params.len());
         // TODO: There are expected to be duplicates but we currently don't have a way to identify
         //       multi-type entity types. We need a way to speed this up.
         let mut validation_params = Vec::with_capacity(params.len());
 
-        let transaction = self.transaction().await.change_context(InsertionError)?;
+        let mut transaction = self.transaction().await.change_context(InsertionError)?;
```
hashdotai, line 770: This is currently immutable which will make the transaction unusable for policy creation later. The `mut` keyword is necessary for the transaction to be modified when creating policies.

libs/@local/graph/store/src/entity/store.rs
```diff
@@ -145,6 +158,9 @@
...
+    pub policies: Vec<CreateEntityPolicyParams>,
```
hashdotai, line 163: Since you're now passing `policies` array explicitly, you need to provide a default value for backward compatibility when callers don't provide policies.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Correctly observed.

libs/@local/graph/store/src/entity/store.rs
TimDiekmann: First of all: we don't need backwards compatibility here, we address the caller site directly. Also, `#[serde(default)]` is used, so it _is_ using a default value.

apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -87,6 +90,18 @@ const createHashEntity = async (params: {
           subject: { kind: "setting", subjectId: "viewFromWeb" },
         },
       ],
+      policies: [
+        {
+          name: `linear-synced-administer-entity-${entityUuid}`,
```
hashdotai: You're updating the entity UUIDs correctly here, but it's important to ensure the policy name is unique for each entity type. Since multiple entities might have the same UUID for different purposes, consider including something type-specific in the policy name to avoid conflicts.

libs/@local/hash-backend-utils/src/service-usage.ts
```diff
@@ -317,6 +329,12 @@ export const createUsageRecord = async (
       provenance,
       entityTypeIds: [systemEntityTypes.usageRecord.entityTypeId],
       relationships: entityRelationships,
+      policies: viewPrincipals.map((principal) => ({
+        name: `usage-record-view-entity-${usageRecordEntityUuid}`,
+        principal,
+        effect: "permit",
+        actions: ["viewEntity"],
+      })),
```
hashdotai, line 350: The `resource` field is missing from this policy definition. Without specifying the resource, the policy won't be properly associated with the entity being created.

```suggestion
      policies: viewPrincipals.map((principal) => ({
        name: `usage-record-view-entity-${usageRecordEntityUuid}`,
        principal,
        effect: "permit",
        actions: ["viewEntity"],
        resource: {
          type: "entity",
          id: usageRecordEntityUuid,
        },
      })),
```

libs/@local/hash-backend-utils/src/service-usage.ts
```diff
@@ -330,38 +348,14 @@
...
         effect: "permit",
         actions: ["viewEntity"],
-        resource: {
-          type: "entity",
-          id: entityUuid,
-        },
-      });
-    }
-  }
+      })),
```
hashdotai, line 369: You need to add `resource` to this policy definition as well. The resource field is required for the policy to be associated with the correct entity.

```suggestion
      policies: viewPrincipals.map((principal) => ({
        name: `usage-record-view-entity-${recordsUsageOfEntityUuid}`,
        principal,
        effect: "permit",
        actions: ["viewEntity"],
        resource: {
          type: "entity",
          id: recordsUsageOfEntityUuid,
        },
      })),
```

apps/hash-api/src/graphql/resolvers/knowledge/user/submit-early-access-form.ts
```diff
@@ -134,26 +130,21 @@ export const submitEarlyAccessFormResolver: ResolverFn<
...
+          name: "prospective-user-view-entity",
+          principal: {
+            type: "actor",
+            actorType: "user",
+            id: user.accountId,
+          },
+          effect: "permit",
+          actions: ["viewEntity"],
+        },
+      ],
```
hashdotai, line 144: Similar to the above comments, the `resource` field is missing here. All policy definitions need a resource to properly associate the policy with the entity.

```suggestion
        policies: [
          {
            name: "prospective-user-view-entity",
            principal: {
              type: "actor",
              actorType: "user",
              id: user.accountId,
            },
            effect: "permit",
            actions: ["viewEntity"],
            resource: {
              type: "entity",
              id: <ENTITY_UUID>,
            },
          },
        ],
```

You'll need to either generate a UUID beforehand or extract it from the created entity afterward.

apps/hash-integration-worker/src/linear-activities.ts
```diff
@@ -87,6 +90,18 @@
...
+          effect: "permit",
+          principal: {
+            type: "actor",
+            actorType: "machine",
+            id: params.authentication.actorId,
+          },
+          actions: ["viewEntity"],
+        },
+      ],
```
hashdotai, line 123: Missing `resource` field in the policy definition. For consistency with other changes and to ensure the policy works correctly, add the resource field here:

```suggestion
          policies: [
            {
              name: `linear-synced-administer-entity-${entityUuid}`,
              effect: "permit",
              principal: {
                type: "actor",
                actorType: "machine",
                id: params.authentication.actorId,
              },
              actions: ["viewEntity"],
              resource: {
                type: "entity",
                id: entityUuid,
              },
            },
          ],
```

libs/@local/graph/store/src/entity/store.rs
```diff
@@ -115,6 +119,15 @@ impl Default for ValidateEntityComponents {
     }
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "codegen", derive(specta::Type))]
+pub struct CreateEntityPolicyParams {
+    pub name: String,
+    pub effect: Effect,
+    pub principal: Option<PrincipalConstraint>,
```
hashdotai, line 127: The `principal` field in `CreateEntityPolicyParams` is defined as optional (`Option<PrincipalConstraint>`), but in all your implementation examples, you're providing it as required. Consider whether it should be optional in the type definition or required in practice. If it's required in practice, update the type definition to reflect this.

apps/hash-integration-worker/src/linear-activities.ts
TimDiekmann: No, multiple entities ***don't*** share UUIDs...

apps/hash-ai-worker-ts/src/activities/flow-activities/shared/infer-summaries-then-claims-from-text/get-entity-summaries-from-text.ts
```diff
@@ -257,9 +257,9 @@ export const getEntitySummariesFromText = async (params: {
     };
 
     for (const { name, summary, type } of toolCallEntitySummaries) {
-      const entityUuid = generateUuid();
+      const entityUuid = generateUuid() as EntityUuid;
```
indietyp, line 260: would it make sense to move this `generateUuid` to `generateEntityUuid` (which is just `generateUuid` under the hood) to make sure that  it's easy to change the generation method in the future?

TimDiekmann: I'd like to have the hash-graph SDK to deal with that instead at some point. Conceptually you're right, yes.

PR #7406, opened 2025-06-13
H-4699: Support policy resource filter on the creator

TimDiekmann:

For some entity types we require that only the entity creator can see the entity, for example for the `usage-record`s, not every member of the web should see them but only the admins and the creators.

## 🔍 What does this change?

- Implement a resource filter `createdByPrincipal`
- Remove per-entity policy for `incurred-in` entities
- Remove per-entity policy for entity creator for `usage-record`, `user-secret`, and `records-usage-of` entities

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
@@ -11,8 +11,8 @@ use alloc::{borrow::Cow, sync::Arc};
 use core::{fmt::Debug, hash::Hash};
 use std::collections::{HashMap, HashSet};
 
-use error_stack::{Report, ReportSink, ResultExt as _, TryReportIteratorExt as _};
-use futures::TryStreamExt as _;
+use error_stack::{Report, ReportSink, ResultExt as _, TryReportStreamExt as _};
+use futures::{StreamExt as _, TryStreamExt as _};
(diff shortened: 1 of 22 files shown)


libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1728,20 +1728,20 @@ where
     ) -> Result<Vec<EntityTypeResource<'_>>, Report<[BuildEntityTypeContextError]>> {
         let () = self
             .as_client()
-            .query(
+            .query_raw(
```
indietyp, line 1731: why the change from `query` to `query_raw`?

indietyp (approved):

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: `query` returns a `Vec<Row>`, while `query_raw` returns a `Stream` over the rows. `query` more-or-less is the `collect()`-ed version of it. But as streams are harder to deal with, it's not unusual that we use `query` instead of `query_raw`. There are two other commonly used variants in the code base: `query_one`,  which fails if there is not exactly one row, and `query_opt` which returns `Option<Row>`.

PR #7431, opened 2025-06-15
H-4782, H-4784, H-4780: HashQL: Solve unconstrained variables through fix-point iteration

indietyp:

During inference one might discover nested inference variables (such as `List<T> <: List<Number>`) if that is the case we shouldn't error out immediately but instead record the newly created constraint.

This is what enables the removal of [H-4780](https://linear.app/hash/issue/H-4780/hashql-remove-structural-edges-during-type-inference), as we gradually discover new edges when required.

---

Structural edges are required for https://linear.app/hash/issue/H-4790/hashql-implement-polarized-µ-blocks-during-type-inference to work properly and for to create a proper order during solving of constraints. They cannot be unified like normal edges (as to why look at H-4790).

This changes it so that structural edges (dependencies) are considered during

---

There are cases in which we cannot propagate a bound during inference, for example given:

```
?1 <: ?2
Number <: ?2
```

During constraint generation when trying to determine the upper bound of `?1` we currently look at the upper bound of `?2`, but because there's no upper bound to speak of `?1` keeps being unconstrained.

This changes it so that after we've solved any constraints (setting `?2 = Number`) we go through any of the unconstraint variables and insert the previously unconstrained variables to their value.

> These have been folded into a single PR, as it was very hard to 

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/core/src/type/inference/solver/graph.rs
+++ b/libs/@local/hashql/core/src/type/inference/solver/graph.rs
@@ -1,4 +1,4 @@
-use core::{mem, mem::size_of, ops::RangeInclusive};
+use core::{mem, ops::RangeInclusive};
 
 use ena::unify::UnifyKey as _;
 use roaring::RoaringBitmap;
--- a/libs/@local/hashql/core/src/type/inference/solver/graph.rs
+++ b/libs/@local/hashql/core/src/type/inference/solver/graph.rs
@@ -168,8 +168,8 @@ impl Graph {
         // do not have to worry about overflow for the max sentinel value. In case of 32 bit
         // systems, if the length is exactly `u32::MAX`, we need to error out.
         assert!(
-            size_of::<usize>() != size_of::<u32>() || length < u32::MAX as usize,
-            "Too many variables, cannot use `usize::MAX` as sentinel value"
+            length < 0x3FFF_FFFF,
+            "Too many variables, expected a maximum of 1.073.741.823 variables"
         );
 
         let offset = self.lookup.len();
--- a/libs/@local/hashql/core/src/type/inference/solver/graph.rs
+++ b/libs/@local/hashql/core/src/type/inference/solver/graph.rs
@@ -9,16 +9,16 @@ use crate::r#type::inference::variable::VariableId;
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub(crate) enum EdgeKind {
     Any = 0,
-    Nominal = 1,
-    Structural = 2,
+    SubtypeOf = 1,
+    DependsOn = 2,
 }
 
 impl EdgeKind {
     const fn from_u32(value: u32) -> Self {
         match value {
             0 => Self::Any,
-            1 => Self::Nominal,
-            2 => Self::Structural,
+            1 => Self::SubtypeOf,
+            2 => Self::DependsOn,
             _ => unreachable!(),
         }
     }
@@ -439,11 +439,16 @@ impl Graph {
 mod tests {
     use core::borrow::Borrow;
 
+    use rstest::rstest;
+
     use super::Graph;
     use crate::r#type::{
         inference::{
             VariableKind,
-            solver::{Unification, graph::EdgeKind},
+            solver::{
+                Unification,
+                graph::{Edge, EdgeKind},
+            },
             variable::VariableId,
         },
         kind::infer::HoleId,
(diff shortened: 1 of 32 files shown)


libs/@local/hashql/core/src/type/inference/solver/mod.rs
```diff
@@ -973,6 +990,63 @@ impl<'env, 'heap> InferenceSolver<'env, 'heap> {
         is_satisfiable
     }
 
+    fn solve_constraints_inference(
```
hashdotai, line 993: This new `solve_constraints_inference` method is a nice addition that helps with constraint propagation between variables. It's well-structured to handle all possible combinations of constraints.

libs/@local/hashql/core/src/type/inference/solver/mod.rs
```diff
@@ -973,6 +990,63 @@
...
+            if let Some(equal) = equal {
+                discharge.push((equal, upper));
+            }
+
+            if let Some(lower) = lower {
+                discharge.push((lower, upper));
+            }
+        }
+
+        discharge.sort_unstable();
```
hashdotai, line 1042: The sorting and deduplication of the discharge list is a good optimization to avoid duplicate constraint processing.

libs/@local/hashql/core/src/type/inference/solver/mod.rs
```diff
@@ -1385,6 +1483,64 @@
...
+    /// unconstrained variables. It cannot be performed earlier due to the constraint
+    /// propagation strategy, which uses a forward-pass for lower bounds followed by a
+    /// backward-pass for upper bounds.
+    ///
+    /// The method iterates through unconstrained variables and checks their graph connections
+    /// (both incoming and outgoing edges) to see if any connected variables now have resolved
+    /// types in the substitution. If so, it adds appropriate bound constraints.
+    ///
+    /// Returns `true` if any progress was made (new constraints were added), `false` otherwise.
+    fn solve_unconstrained(
```
hashdotai, line 1507: The `solve_unconstrained` method is a key part of the fix-point iteration strategy and is well-documented. I particularly like how it clearly explains the scenarios it's designed to handle.

libs/@local/hashql/core/src/type/inference/solver/graph.rs
```diff
@@ -1,15 +1,137 @@
-use core::{mem, mem::size_of};
+use core::{mem, ops::RangeInclusive};
 
 use ena::unify::UnifyKey as _;
 use roaring::RoaringBitmap;
 
 use super::Unification;
 use crate::r#type::inference::variable::VariableId;
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
```
hashdotai, line 9: The `EdgeKind` enum and associated methods are well-designed. The bit manipulation approach for storing both edge type and target in a single u32 is an efficient representation.

libs/@local/hashql/core/src/type/inference/solver/graph.rs
```diff
@@ -46,8 +168,8 @@
         // do not have to worry about overflow for the max sentinel value. In case of 32 bit
         // systems, if the length is exactly `u32::MAX`, we need to error out.
         assert!(
-            size_of::<usize>() != size_of::<u32>() || length < u32::MAX as usize,
-            "Too many variables, cannot use `usize::MAX` as sentinel value"
```
hashdotai, line 50: The maximum variables assertion message could be more specific about where this limit comes from. Consider mentioning that it's due to the bit allocation in the Edge encoding (30 bits for the target value).

libs/@local/hashql/core/src/type/inference/visit.rs
```diff
@@ -0,0 +1,108 @@
...
+    r#type::{
+        Type,
+        environment::Environment,
+        kind::{GenericArgument, Infer, Param, generic::GenericSubstitution},
+        recursion::RecursionBoundary,
+        visit::{self, Visitor},
+    },
+};
+
+pub(crate) struct VariableVisitorFilter(!);
```
hashdotai, line 13: The `VariableDependencyCollector` is a cleaner approach to collecting dependencies compared to the previous structural edge system. It leverages the visitor pattern effectively.

libs/@local/hashql/core/src/type/inference/visit.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.visit_type(r#type);
+
+        self.variables.drain(..)
+    }
+}
+
+impl<'heap> Visitor<'heap> for VariableDependencyCollector<'_, 'heap> {
+    type Filter = VariableVisitorFilter;
+
+    fn env(&self) -> &Environment<'heap> {
```
hashdotai, line 53: Good implementation of the visitor pattern. Consider using `std::iter::once` instead of `self.variables.drain(..)` if you want to avoid modifying the state of the collector.

libs/@local/hashql/core/src/type/visit.rs
```diff
@@ -462,3 +469,41 @@ pub fn walk_generic<'heap, V: Visitor<'heap> + ?Sized>(
     visitor.visit_generic_arguments(arguments);
     visitor.visit_id(base);
 }
+
```
hashdotai, line 472: The new `walk_param` and `walk_infer` implementations are good additions for supporting substitutions in the visitor pattern. This makes the visitor more complete and consistent.

PR #7432, opened 2025-06-15
H-4783: HashQL: Improve compiletest panics

indietyp:

Panics in compiletest are currently quite opaque, as they do not have any backtraces and sometimes the output of tracing just removes them from the output.

Using `catch_unwind` and a custom hook we're able to make the error clearer.

> Sadly no output, because well... you need to have an error for that to happen.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/compiletest/src/executor/mod.rs
+++ b/libs/@local/hashql/compiletest/src/executor/mod.rs
@@ -3,7 +3,7 @@ mod trial;
 mod trial_group;
 
 use core::error;
-use std::{self, io, thread};
+use std::{self, io, sync::mpsc, thread};
 
 use anstream::StripStream;
 use ariadne::Source;
@@ -40,6 +40,8 @@ pub(crate) enum TrialError {
     TrialShouldFail,
     #[display("Expected trial to pass, but it failed")]
     TrialShouldPass,
+    #[display("Assertion failed for trial: {message}")]
+    AssertionFailed { message: String },
 }
 
 impl error::Error for TrialError {}
@@ -143,19 +145,28 @@ impl<'graph> TrialSet<'graph> {
 
     pub(crate) fn run(&self, context: &TrialContext) -> Vec<Report<[TrialError]>> {
         thread::scope(|scope| {
-            let mut handles = Vec::new();
+            let (tx, rx) = mpsc::channel();
 
             for group in &self.groups {
-                let handle = scope.spawn(|| group.run(context));
-                handles.push(handle);
+                let tx = tx.clone();
+                scope.spawn(move || {
+                    let result = group.run(context);
+
+                    tx.send(result).expect("should be able to send result");
+                });
             }
 
-            // First collect the values into a `Vec<_>`. This way we're able to say how many of the
-            // tasks failed
-            handles
-                .into_iter()
-                .flat_map(|handle| handle.join().expect("should be able to join thread"))
-                .collect()
+            // Dropping the sender here means that we will automatically shutdown the receiver once
+            // all senders are dropped.
+            drop(tx);
+
+            let mut reports = Vec::with_capacity(self.groups.len());
+
+            while let Ok(mut result) = rx.recv() {
+                reports.append(&mut result);
+            }
+
+            reports
         })
     }
 }
--- a/libs/@local/hashql/compiletest/src/executor/trial_group.rs
+++ b/libs/@local/hashql/compiletest/src/executor/trial_group.rs
@@ -1,4 +1,8 @@
-use std::io;
+use core::any::Any;
+use std::{
+    io,
+    panic::{self},
+};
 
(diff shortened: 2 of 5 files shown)


libs/@local/hashql/compiletest/src/executor/trial_group.rs
```diff
@@ -1,14 +1,28 @@
...
 use nextest_metadata::{RustBinaryId, RustTestBinaryKind};
 use rayon::iter::{IntoParallelRefIterator as _, ParallelIterator as _};
 
-use super::{TrialContext, TrialError, trial::Trial};
+use super::{TrialContext, TrialDescription, TrialError, trial::Trial};
 use crate::{TestGroup, reporter::Statistics};
 
+// Adapted from: https://github.com/rust-lang/rust/blob/6c8138de8f1c96b2f66adbbc0e37c73525444750/library/std/src/panicking.rs#L779-L787
+fn panic_payload_as_str(payload: Box<dyn Any + Send + 'static>) -> String {
+    match payload.downcast::<&'static str>() {
```
hashdotai, line 18: Consider adding a more descriptive error message when the payload is neither a string nor a &str. Instead of just returning "Box<dyn Any>", you could add some context like "Unknown panic payload: Box<dyn Any>".

libs/@local/hashql/compiletest/src/lib.rs
```diff
@@ -40,6 +44,21 @@ mod reporter;
 mod styles;
 mod suite;
 
+static PANICKED: AtomicBool = AtomicBool::new(false);
+
+fn panic_hook(panic_info: &PanicHookInfo) {
+    let message = panic_info
+        .payload_as_str()
+        .map_or_else(|| "Box<dyn Any>".to_owned(), ToOwned::to_owned);
```
hashdotai, line 52: This looks like it's duplicating logic from the `panic_payload_as_str` function in trial_group.rs. Consider extracting this into a shared utility function to avoid code duplication.

libs/@local/hashql/compiletest/src/lib.rs
```diff
@@ -110,6 +129,7 @@
                 let ignored = trials.ignored();
 
                 setup_progress_header!(reporter, Summary { total, ignored }, statistics);
+                panic::set_hook(Box::new(panic_hook));
```
hashdotai, line 132: You're setting a panic hook here, but it's important to consider that this will replace any existing panic hook. Consider saving the previous panic hook and chaining to it after your custom hook logic runs:

```suggestion
                let old_hook = panic::take_hook();
                panic::set_hook(Box::new(move |info| {
                    panic_hook(info);
                    old_hook(info);
                }));
```

libs/@local/hashql/compiletest/src/reporter.rs
```diff
@@ -193,7 +193,8 @@ impl Reporter {
 
             write!(stderr, "{delimiter} ---")?;
             write!(stderr, "\n\n")?;
-            write!(stderr, "{report}")?;
+            #[expect(clippy::use_debug)]
```
hashdotai, line 196: Why are you using `#[expect(clippy::use_debug)]` here? If you're intentionally using the debug output format because it provides more detailed error information, it would be helpful to add a comment explaining this choice.

libs/@local/hashql/compiletest/src/executor/mod.rs
```diff
@@ -143,19 +145,28 @@ impl<'graph> TrialSet<'graph> {
 
     pub(crate) fn run(&self, context: &TrialContext) -> Vec<Report<[TrialError]>> {
         thread::scope(|scope| {
-            let mut handles = Vec::new();
+            let (tx, rx) = mpsc::channel();
```
hashdotai, line 148: Consider adding a capacity to the channel if you know the size of `self.groups` to potentially optimize memory allocation:

```suggestion
            let (tx, rx) = mpsc::channel();
```

libs/@local/hashql/compiletest/src/reporter.rs
indietyp: because it's a report?!

libs/@local/hashql/compiletest/src/executor/mod.rs
indietyp: Then it'd be a sync-channel

libs/@local/hashql/compiletest/src/lib.rs
indietyp: Considering that we have full control and that the old hook would just output it to printf as well, I consciously decided against doing so.

libs/@local/hashql/compiletest/src/executor/trial_group.rs
indietyp: No. See the adapted from...

PR #7433, opened 2025-06-15
H-4747: HashQL: Specialize HIR Graph Expressions

indietyp:

Unlike the previous specialization pass, the graph specialization pass works in a backtracking fashion. We basically take a look if we're at a `tail` operation, if that is the case we start the conversion into a graph HIR expression and work our way up, while doing so we convert the nested definitions into a set of filters.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -1,53 +1,147 @@
...
+
+        let mut next = call.arguments[0].value;
+        loop {
+            // Follow any local variables
+            while let NodeKind::Variable(Variable {
+                kind: VariableKind::Local(local),
+                ..
+            }) = next.kind
+            {
+                next = self.locals[Universe::Value][&local.name.value];
```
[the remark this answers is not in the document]
indietyp: no, we have multiple checks that verify the local exists.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -202,6 +345,12 @@ impl<'heap> Fold<'heap> for Specialization<'_, 'heap> {
             self.types
                 .try_insert(node.id, r#type)
                 .expect("node id should be unique in types");
+
```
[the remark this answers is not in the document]
indietyp: I think that's clear enough from the context...

libs/@local/hashql/hir/src/lower/specialization/error.rs
```diff
@@ -113,3 +140,105 @@ pub(crate) fn unknown_intrinsic(span: SpanId, intrinsic_name: &str) -> Specializ
...
+
+    diagnostic
+        .labels
+        .push(Label::new(span, label_text).with_order(0));
+
+    diagnostic.add_help(Help::new(format!(
+        "{} is not a valid graph operation. Graph operation chains can only contain intrinsic \
+         functions that are part of the HashQL graph API. Higher-order functions (HOFs) and \
+         user-defined functions are not supported yet. To track support for user-defined \
+         functions see https://linear.app/hash/issue/H-4776/hashql-allow-user-defined-functions-in-graph-pipelines",
```
hashdotai, line 204: The error messages for graph operations provide clear guidance, but the line about user-defined functions links to a Linear ticket. Consider adding a version comment to ensure this link stays relevant, or using a more general description about future support instead of a specific ticket reference.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -1,53 +1,147 @@
...
+                    call.span,
+                    call.function,
+                ));
+
+                return None;
+            };
+
+            match intrinsic {
+                "::core::graph::body::filter" => {
+                    let &[follow, closure] = &*call.arguments else {
```
hashdotai, line 113: This matching pattern is quite effective for unpacking complex nested structures, but consider adding a comment explaining the overall pattern extraction strategy here for future maintainers.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -1,53 +1,147 @@
...
             intrinsics,
 
             nested: false,
             visited: FastHashMap::default(),
+            locals: FastRealmsMap::default(),
             diagnostics: Vec::new(),
         }
     }
 
+    fn fold_call_into_graph_read(
```
hashdotai, line 70: The naming of this backtracking approach could be clearer. Since you're working 'upward' through the chain (from tail to head), consider renaming to reflect this directionality, e.g., `fold_call_into_graph_read_backtrack` or similar to make the backtracking nature obvious.

libs/@local/hashql/core/src/type/kind/opaque.rs
```diff
@@ -333,7 +333,12 @@ impl<'heap> PrettyPrint<'heap> for OpaqueType<'heap> {
         env: &Environment<'heap>,
         boundary: &mut PrettyPrintBoundary,
     ) -> RcDoc<'heap, anstyle::Style> {
-        RcDoc::text(self.name.unwrap())
+        let name = self.name.unwrap();
+        let name = name
```
hashdotai, line 339: When adding specialized name formatting like this for OpaqueType, it would be helpful to add a brief comment explaining why this transformation is necessary (e.g., removing module prefixes for cleaner display in UI or error messages).

libs/@local/hashql/hir/src/lower/specialization/error.rs
indietyp: The link is there so that it keeps being the same

libs/@local/hashql/hir/src/lower/specialization/mod.rs
indietyp: I think it's pretty clear from the code no?

libs/@local/hashql/hir/src/lower/specialization/mod.rs
indietyp: Why not just call it:

`fold_call_into_graph_read_backtrack_upwards_but_only_if_no_error`

libs/@local/hashql/core/src/type/kind/opaque.rs
indietyp: done in [2c3b186](https://github.com/hashintel/hash/pull/7433/commits/2c3b18625a6c849c258c0f80a3ff3b606ba3bca7)

libs/@local/hashql/hir/src/lower/specialization/mod.rs
hashdotai: @indietyp Haha, good one! But jokes aside, I actually agree with the core suggestion. The current name `fold_call_into_graph_read` doesn't clearly convey the backtracking nature of the function.

Something like `fold_call_into_graph_read_backtrack` or even just `backtrack_graph_chain` would make the intent clearer to future readers. The function is doing something quite specific - walking backwards from a tail operation to construct a complete graph expression - and having that explicitly in the name would help with code readability.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -1,53 +1,147 @@
...
+        // The first argument is always the graph we're referring to.
+        let tail = match intrinsic {
+            "::core::graph::tail::collect" => GraphReadTail::Collect,
+            _ => unreachable!(),
+        };
+
+        let mut body = SmallVec::new();
+
+        let mut next = call.arguments[0].value;
+        loop {
```
hashdotai, line 84: This loop seems to run indefinitely if a cycle exists in the graph structure. Consider adding a cycle detection mechanism or a maximum iteration count to prevent potential infinite loops.

libs/@local/hashql/hir/src/lower/specialization/error.rs
```diff
@@ -113,3 +140,105 @@ pub(crate) fn unknown_intrinsic(span: SpanId, intrinsic_name: &str) -> Specializ
...
+    );
+
+    let label_text = "This is not a graph intrinsic operation";
+
+    diagnostic
+        .labels
+        .push(Label::new(span, label_text).with_order(0));
+
+    diagnostic.add_help(Help::new(format!(
+        "{} is not a valid graph operation. Graph operation chains can only contain intrinsic \
```
hashdotai, line 201: The error message format is good, but consider including examples of valid graph operations to make it more actionable for users troubleshooting their code.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -1,53 +1,147 @@
...
             diagnostics: Vec::new(),
         }
     }
 
+    fn fold_call_into_graph_read(
+        &mut self,
+        call: Call<'heap>,
+        intrinsic: &'static str,
+    ) -> Option<GraphRead<'heap>> {
+        // The first argument is always the graph we're referring to.
```
hashdotai, line 75: It would be helpful to add a comment explaining the structure of a graph operation chain (head -> body -> tail) and how this backtracking approach navigates it. This would make the code more maintainable for future developers.

libs/@local/hashql/hir/src/lower/specialization/mod.rs
indietyp: HIR is a tree

libs/@local/hashql/hir/tests/ui/lower/checking/collect-filter-graph.stdout
indietyp, line 1: I know I need to work on the output some more, because this is *not* easily readable

TimDiekmann:
I updated the branch due to failing CI. It should work with latest changes merged.

PR #7435, opened 2025-06-16
H-4621: Fix AI assistant retrieval from the graph

TimDiekmann:

We currently use getMachineByIdentifier, which does not work for AIs. Instead, we need to expose getAiByIdentifier.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/api/openapi/openapi.json
+++ b/libs/@local/graph/api/openapi/openapi.json
@@ -847,7 +847,7 @@
         ],
         "responses": {
           "200": {
-            "description": "The web was retrieved successfully",
+            "description": "The AI was retrieved successfully",
             "content": {
               "application/json": {
                 "schema": {
@@ -926,7 +926,7 @@
         ],
         "responses": {
           "200": {
-            "description": "The web was retrieved successfully",
+            "description": "The machine was retrieved successfully",
             "content": {
               "application/json": {
                 "schema": {
--- a/libs/@local/graph/api/src/rest/principal.rs
+++ b/libs/@local/graph/api/src/rest/principal.rs
@@ -222,7 +222,7 @@ where
         ("identifier" = String, Path, description = "The identifier of the machine to get"),
     ),
     responses(
-        (status = 200, content_type = "application/json", description = "The web was retrieved successfully", body = Option<Value>),
+        (status = 200, content_type = "application/json", description = "The machine was retrieved successfully", body = Option<Value>),
 
         (status = 500, description = "Store error occurred"),
     )
@@ -358,7 +358,7 @@ where
         ("identifier" = String, Path, description = "The identifier of the AI to get"),
     ),
     responses(
-        (status = 200, content_type = "application/json", description = "The web was retrieved successfully", body = Option<Value>),
+        (status = 200, content_type = "application/json", description = "The AI was retrieved successfully", body = Option<Value>),
 
         (status = 500, description = "Store error occurred"),
     )


libs/@local/hash-backend-utils/src/machine-actors.ts
```diff
@@ -51,7 +66,7 @@ export const getWebMachineId = async (
 export const getMachineEntityByIdentifier = async (
   context: { graphApi: GraphApi },
   authentication: { actorId: ActorEntityUuid },
-  { identifier }: { identifier: MachineActorIdentifier },
+  { identifier }: { identifier: MachineActorIdentifier | AiActorIdentifier },
 ): Promise<Entity<Machine> | null> => {
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You've made an excellent point. The current implementation of `getMachineEntityByIdentifier` still calls `getMachineByIdentifier` under the hood, even though we're now accepting `AiActorIdentifier` as part of the parameter type.

Your suggestion to branch the logic internally is spot on. This would ensure that AI identifiers are properly handled by the appropriate retrieval method. 

I'll implement this change in a follow-up PR to keep the scope of the current one focused. The current PR establishes the foundation by adding the `getAiByIdentifier` functionality, and the next step will be to properly use it within the existing functions that need to handle both machine and AI identifiers.

Thank you for catching this issue!

libs/@local/graph/api/openapi/openapi.json
```diff
@@ -818,6 +818,50 @@
...
+            "description": "The identifier of the AI to get",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "The web was retrieved successfully",
```
hashdotai: It looks like the description in the success response still refers to 'web' rather than 'AI'. Let's update it to be more accurate.

libs/@local/graph/api/src/rest/principal.rs
```diff
@@ -342,6 +349,52 @@ where
...
+#[utoipa::path(
+    get,
+    path = "/actors/ai/identifier/{identifier}",
+    tag = "Web",
+    params(
+        ("X-Authenticated-User-Actor-Id" = ActorEntityUuid, Header, description = "The ID of the actor which is used to authorize the request"),
+        ("identifier" = String, Path, description = "The identifier of the AI to get"),
+    ),
+    responses(
+        (status = 200, content_type = "application/json", description = "The web was retrieved successfully", body = Option<Value>),
```
hashdotai: Same issue here with the description referring to 'web' instead of 'AI'.

libs/@local/graph/sdk/typescript/src/principal/actor.ts
```diff
@@ -0,0 +1,34 @@
...
+): Promise<Machine | null> =>
+  graphAPI
+    .getMachineByIdentifier(authentication.actorId, identifier)
+    .then(({ data }) => {
+      const machine = data as Machine | null;
+      if (!machine) {
+        return null;
+      }
+      return machine;
+    });
```
hashdotai, line 19: These two implementation functions have very similar structure. Consider creating a generic helper function to reduce code duplication in the future.

libs/@local/hash-backend-utils/src/machine-actors.ts
```diff
@@ -24,12 +28,14 @@ import type { Logger } from "./logger.js";
 
 export type WebMachineActorIdentifier = `system-${WebId}`;
 
-export type GlobalMachineActorIdentifier = SystemTypeWebShortname | "hash-ai";
+export type GlobalMachineActorIdentifier = SystemTypeWebShortname;
```
hashdotai, line 31: Since 'hash-ai' is now strictly an AI identifier rather than a machine identifier, it should be removed from the `GlobalMachineActorIdentifier` type. You've correctly done this, but we should also update any existing references that may still be using 'hash-ai' as a machine identifier elsewhere in the codebase.

libs/@local/hash-backend-utils/src/machine-actors.ts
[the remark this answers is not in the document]
TimDiekmann: You're both wrong. The AI has the `machine` entity type (I don't know if we want to change that @CiaranMn?) so it's expected that both call the `getMachineEntityByIdentifier`.

libs/@local/graph/api/src/rest/principal.rs
TimDiekmann: https://github.com/hashintel/hash/pull/7435/commits/171dd53d8515c0423573ed454fbf231ead1658d1

libs/@local/graph/api/openapi/openapi.json
TimDiekmann: https://github.com/hashintel/hash/pull/7435/commits/171dd53d8515c0423573ed454fbf231ead1658d1

libs/@local/hash-backend-utils/src/machine-actors.ts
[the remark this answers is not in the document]
CiaranMn: No, we don't want to change this.

CiaranMn (approved):

PR #7436, opened 2025-06-16
H-4702: Bump transitive Cargo dependencies

TimDiekmann:

Renovate currently does not cover all transitive cargo dependency. To update them manually, the easiest way is to use

## 🔍 What does this change?

- `cargo update --ignore-rust-version --recursive`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,9 +31,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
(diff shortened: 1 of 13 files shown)


vilkinsons:
@TimDiekmann Could we use Renovate's [`postUpgradeTasks`](https://docs.renovatebot.com/configuration-options/#postupgradetasks) to run `cargo update --ignore-rust-version --recursive` automatically as part of creating its per-upgrade commits?

TimDiekmann:
In theory this would be possible, yes, but calling `cargo update --ignore-rust-version --recursive` updates _all_ transitive dependencies, not just for the dependencies which were updated. This means unrelated packages are bumped, which is annoying to track down if errors happen.
AFAIK renovate does not provide the list of updated packages to `postUpgradeTasks` which would be required here. And even if it would, it would not upgrade transitive dependencies for dependencies we don't update (e.g. because the dependency simply don't get updates).

We previously tried lock file maintenance of renovate. We might want to explore that again but previously it [had issues](https://hashintel.slack.com/archives/C02TWBTT3ED/p1740393809595489?thread_ts=1739918827.982519&cid=C02TWBTT3ED).

TimDiekmann:
I added a follow-up ticket to reduce the amount of `too_many_lines` lint: H-4805

TimDiekmann:
@vilkinsons: Another disadvantage of doing a transitive dependency update on regular updates is that multiple PRs will touch the same snippets, which is very likely resulting in merge conflicts. So if any, we should explore lock-file maintenance again for Rust (which is what I mostly do myself). I created a task for that: H-4807

PR #7453, opened 2025-06-19
H-4823: Read policies for actors filtered by actions

TimDiekmann:

We typically only need a few policies based on the actions we're going to do.

## 🔍 What does this change?

- Add functions to allow adding actions to the policy components
- The actions are persisted in the policy set, so if a request with a different action is being issued, it will fail if the action was not previously used. This helps in tracking issues, e.g. when it was forgotten to request the `instantiate` policy and the request evaluates to "cannot instantiate" it is hard to track down why a user may cannot instantiate an entity type. With tracked actions there is a clear error message that the `instantiate` action was not tracked at all

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
@@ -17,7 +17,7 @@ use hash_graph_authorization::{
     AuthorizationApi,
     backend::ModifyRelationshipOperation,
     policies::{
-        Authorized, ContextBuilder, Effect, PartialResourceId, Policy, PolicyId, PolicySet,
+        Authorized, ContextBuilder, Effect, PartialResourceId, Policy, PolicyComponents, PolicyId,
         Request, RequestContext,
         action::ActionName,
         principal::{PrincipalConstraint, actor::AuthenticatedActor},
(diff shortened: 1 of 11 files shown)


libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1368,32 +1360,48 @@ where
...
+                        policy_edition.effect,
+                        policy_edition.principal_id,
+                        policy_edition.principal_type,
+                        policy_edition.actor_type,
+                        policy_edition.resource_constraint,
+                        array_remove(array_agg(policy_action.action_name), NULL) AS actions
+                    FROM policy_edition
+                    LEFT JOIN policy_action
+                        ON policy_action.policy_id = policy_edition.id
+                        AND policy_action.transaction_time @> now()
```
indietyp, line 1376:
```suggestion
                    LEFT JOIN policy_action
                           ON policy_action.policy_id = policy_edition.id
                          AND policy_action.transaction_time @> now()
```

Make the formatting consistent with the previous iteration

indietyp (dismissed):
minor formatting related suggestion otherwise LGTM

indietyp (approved):

indietyp (approved):

PR #7454, opened 2025-06-19
Support enum variant skipping in codegen

TimDiekmann:

We currently have more actions present in the graph than being used. To only expose the used actions, we need variant skipping in codegen.


libs/@local/codegen/tests/standalone-types/enums.rs
```diff
@@ -90,3 +90,56 @@ pub(crate) enum EnumUntagged {
         simple: StructSimple,
     },
 }
+
+// Untagged - no discriminator field, purely based on shape
```
TimDiekmann:
```suggestion
```

PR #7480, opened 2025-06-22
H-4824, H-4847: Collect logs and metrics and send them via OTel

TimDiekmann:

Our OTel setup grew quite a bit without ever updating to current practices. This revamps the `hash-telemetry` crate and allows capturing of `logs` and `metrics` in addition to the traces.

As we don't have a setup to collect the traces except the `jaegertracing`, this also adds an OTel collector to the dev environment. Currently, this just forwards it to the existing `jaeger`, but we can extend this to send it to various services, such as DataDog.

## 🔍 What does this change?

- Overhaults the telemetry setup
- Add container for OTel collecting
- Update Sentry to 0.40

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/telemetry/src/lib.rs
```diff
@@ -10,19 +10,22 @@
 
 extern crate alloc;
 
-pub mod console;
-pub(crate) mod formatter;
-pub mod logging;
-pub mod opentelemetry;
-pub mod sentry;
+pub mod logs;
```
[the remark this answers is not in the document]
TimDiekmann: `turbo` is driving me crazy

libs/@local/telemetry/src/lib.rs
```diff
@@ -42,30 +45,71 @@ pub struct TracingConfig {
...
+            && let Err(error) = provider.shutdown()
+        {
+            tracing::warn!("Failed to shutdown OpenTelemetry logs: {error}");
+        }
+
+        if let Some(provider) = self.otlp_traces_provider.take()
+            && let Err(error) = provider.shutdown()
+        {
+            tracing::warn!("Failed to shutdown OpenTelemetry traces: {error}");
+        }
```
hashdotai, line 73: The tracing layer is now properly dropping resources when the guard is dropped. This is a significant improvement to ensure resources are properly cleaned up. Good implementation with a clear drop order for each provider.

libs/@local/telemetry/src/lib.rs
```diff
@@ -10,19 +10,22 @@
 
 extern crate alloc;
 
-pub mod console;
-pub(crate) mod formatter;
 pub mod logging;
-pub mod opentelemetry;
-pub mod sentry;
+pub mod metrics;
+pub mod traces;
```
hashdotai, line 15: This reorganization creates a cleaner separation of concerns between different telemetry types (logs, metrics, traces). The modular approach will make future extensions or modifications easier.

libs/@local/telemetry/src/otlp.rs
```diff
@@ -0,0 +1,29 @@
+/// Arguments for configuring the logging setup
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "clap", derive(clap::Args), clap(next_help_heading = Some("Open Telemetry")))]
+#[expect(
+    clippy::struct_field_names,
+    reason = "This is a configuration struct, field names are kept short for brevity"
+)]
+pub struct OtlpConfig {
```
hashdotai, line 8: Consider adding a small documentation comment here explaining what this configuration is capturing and how it relates to the three telemetry types (traces, logs, metrics).

apps/hash-external-services/docker-compose.dev.yml
```diff
@@ -59,17 +59,38 @@ services:
...
       - "5778:5778"
       # accept jaeger.thrift over Thrift-compact protocol (used by most SDKs)
       - "6831:6831"
     environment:
       COLLECTOR_OTLP_ENABLED: "true"
+    read_only: true
+    security_opt:
+      - no-new-privileges:true
+
+  otel-collector:
```
hashdotai, line 74: The Docker service reorganization seems clearer now. A few suggestions:

1. Consider adding a comment at the top of this file explaining the relationship between jaeger and otel-collector for future maintainers
2. It would be helpful to document why the healthcheck is needed in a comment

apps/hash-external-services/opentelemetry-collector/otel-collector-config.yaml
```diff
@@ -0,0 +1,33 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+
+processors:
+  batch:
+
+exporters:
```
hashdotai, line 10: The OpenTelemetry collector configuration is well-structured. For future maintainability, consider adding a few more comments explaining:

1. Why metrics and logs are using `nop` exporters (I assume they'll be connected to other systems later)
2. Any configuration options that might need to be changed when moving to production

libs/@local/telemetry/src/otlp.rs
TimDiekmann: The members are called `endpoint`, this means they refer to the endpoint. Also, there is a doc.

libs/@local/telemetry/src/logging/otlp.rs
```diff
@@ -0,0 +1,40 @@
...
+pub(crate) fn provider(
+    config: &OtlpConfig,
+) -> Result<Option<SdkLoggerProvider>, Report<ExporterBuildError>> {
+    let Some(endpoint) = config.logs_endpoint.as_deref() else {
+        return Ok(None);
+    };
+
+    Ok(Some(
+        SdkLoggerProvider::builder()
+            .with_resource(Resource::builder().with_service_name("graph").build())
```
indietyp, line 19: shouldn't the service name be configurable? e.g. the type-fetcher should have a different name than the graph. No?

libs/@local/telemetry/src/metrics/otlp.rs
```diff
@@ -0,0 +1,25 @@
...
+pub(crate) fn provider(
+    config: &OtlpConfig,
+) -> Result<Option<SdkMeterProvider>, Report<ExporterBuildError>> {
+    let Some(endpoint) = config.metrics_endpoint.as_deref() else {
+        return Ok(None);
+    };
+
+    Ok(Some(
+        SdkMeterProvider::builder()
+            .with_resource(Resource::builder().with_service_name("graph").build())
```
indietyp, line 16: same here

libs/@local/telemetry/src/traces/otlp.rs
```diff
@@ -0,0 +1,45 @@
...
+pub(crate) fn provider(
+    config: &OtlpConfig,
+) -> Result<Option<SdkTracerProvider>, Report<ExporterBuildError>> {
+    let Some(endpoint) = config.traces_endpoint.as_deref() else {
+        return Ok(None);
+    };
+
+    Ok(Some(
+        SdkTracerProvider::builder()
+            .with_resource(Resource::builder().with_service_name("graph").build())
```
indietyp, line 19: same here

indietyp (changes requested):
Thank you! One minor question / note otherwise LGTM

libs/@local/telemetry/src/logging/otlp.rs
TimDiekmann: Yes, but we need to configure more than just the service name so I decided to leave it as it is for now and cover that in a follow-up (it always has been like this).

indietyp (approved):

PR #7496, opened 2025-06-24
H-4868: Implement policy optimization for `OR`-to-`IN` clause conversion

TimDiekmann:

This PR implements a comprehensive policy optimization system that converts multiple OR conditions to efficient IN clauses for database queries. This significantly improves query performance when dealing with policies that have many permit conditions on the same resource type.

The optimization transforms database queries from patterns like:
`entity_uuid = a OR entity_uuid = b OR entity_uuid = c OR entity_uuid = d`
to:
`entity_uuid = ANY([a, b, c, d])`

This reduces query complexity from O(n) individual conditions to O(1) IN clause operations, providing substantial performance improvements for policy evaluation in large-scale deployments.

## 🔍 What does this change?

- **Add `OptimizationData` structure**: Holds extracted permit data for entity UUIDs and web IDs that can be optimized
- **Implement policy analysis**: New `analyze_optimization_opportunities()` method extracts optimizable patterns using efficient `swap_remove` and `HashSet` deduplication
- **Add optimized filter creation**: New `for_policies_with_optimization()` method creates efficient database filters with IN clauses
- **Support dual optimization**: Both entity UUID and web ID optimization patterns are supported
- **Add builder controls**: Opt-out functionality via `without_optimization()` and `with_optimization_enabled(false)` methods
- **Prevent `PolicySet` creation after optimization**: Safety measure to prevent action tracking issues
- **Comprehensive test coverage**: Tests for all optimization scenarios including edge cases
- **SQL parameter support**: Extended `ParameterList` enum with `EntityUuids` and `WebIds` variants

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- **Unit tests**: Comprehensive test suite covering all optimization scenarios
  - `optimization_data_empty_by_default()` - Verifies default state
  - `optimization_data_detects_entity_optimizations()` - Entity UUID detection
  - `build_policy_set_prevented_after_optimization()` - Safety constraints
  - `builder_optimization_opt_out_controls_policy_set()` - Opt-out functionality
  - `extract_filter_policies_converts_correctly()` - Policy extraction
  - `entity_optimization_creates_in_clause()` - Entity UUID optimization
  - `entity_optimization_with_forbids()` - Mixed permit/forbid scenarios
  - `entity_optimization_with_other_permits()` - Complex policy combinations
  - `web_id_optimization_creates_in_clause()` - Web ID optimization
  - `combined_entity_and_web_optimization()` - Dual optimization scenarios

## ❓ How to test this?

1. Checkout the branch
2. Run the test suite: `cargo test --package hash-graph-store --lib filter::tests::optimization`
3. Verify all tests pass and clippy shows no warnings: `cargo clippy --workspace --all-features --all-targets --no-deps`
4. Test policy creation with optimization enabled (default behavior)
5. Test policy creation with optimization disabled using `without_optimization()`
6. Confirm that `PolicySet` creation fails after optimization as expected

## 📹 Demo

Performance testing shows significant query improvements:
- Before: Multiple OR conditions create complex query plans
- After: Single IN clause operations with array parameters
- Expected performance improvement: 30-70% reduction in query execution time for policies with 5+ similar permits

### DIFF
--- a/libs/@local/graph/authorization/src/policies/components.rs
+++ b/libs/@local/graph/authorization/src/policies/components.rs
@@ -83,14 +83,6 @@ impl PolicyComponents {
     /// This method returns an iterator of `(Effect, Option<&ResourceConstraint>)` tuples
     /// that can be used directly with filter creation methods. This encapsulates the
     /// extraction logic and provides a clean interface for filter optimization.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// // Usage in filter creation (would be in store package):
-    /// let filter_data = policy_components.extract_filter_policies();
-    /// let filter = Filter::for_policies(filter_data, policy_components.actor_id());
-    /// ```
     #[tracing::instrument(level = "info", skip(self))]
     pub fn extract_filter_policies(
         &self,
@@ -116,8 +108,9 @@ impl PolicyComponents {
     ///
     /// This method examines the policies to find patterns that can be optimized
     /// for database query generation. It extracts optimizable policies from the
-    /// policies vec using [`swap_remove`] for efficiency and stores them in [`optimization_data`].
-    /// The remaining policies vec will only contain non-optimizable policies.
+    /// policies vec using [`Vec::swap_remove`] for efficiency and stores them in
+    /// [`OptimizationData`]. The remaining policies vec will only contain non-optimizable
+    /// policies.
     ///
     /// Uses [`HashSet`] during analysis for deduplication and fast lookups, then converts
     /// to [`Vec`] for efficient filter parameter usage.


libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -27,17 +51,138 @@ impl PolicyComponents {
...
+            if should_extract {
+                self.policies.swap_remove(i);
+                // Don't increment i since we just moved the last element to position i
+            } else {
+                i += 1;
+            }
+        }
+
+        // Convert HashSets to Vecs for efficient filter usage
+        self.optimization_data.permitted_entity_uuids = entity_uuids_set.into_iter().collect();
```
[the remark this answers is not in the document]
TimDiekmann: Ordering does not matter.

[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer @TimDiekmann is correct - the ordering doesn't matter in this case since the database query planner will treat the IN clause values as an unordered set regardless of the order they're provided in. The optimization is about reducing multiple OR conditions to a single IN clause, and the specific order within the IN clause parameters doesn't affect the query execution or results.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -27,17 +51,138 @@ impl PolicyComponents {
...
+    /// for exact entity UUIDs can be converted from:
+    /// `entity_uuid = a OR entity_uuid = b OR entity_uuid = c`
+    /// to: `entity_uuid = ANY([a, b, c])`
+    #[must_use]
+    pub const fn optimization_data(&self) -> &OptimizationData {
+        &self.optimization_data
+    }
+
+    /// Analyzes policies and extracts optimization opportunities.
+    ///
```
hashdotai, line 108: The optimization system looks well designed. Consider adding a simple performance metric or benchmark in this comment section to quantify the expected improvement (e.g., "This optimization can reduce query execution time by X% for policies with multiple entity permits"). This would help future developers understand the impact.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -27,17 +51,138 @@
...
+                }
+                (Effect::Permit, Some(ResourceConstraint::Web { web_id })) => {
+                    web_ids_set.insert(*web_id);
+                    true
+                }
+                _ => false,
+            };
+
+            if should_extract {
+                self.policies.swap_remove(i);
```
hashdotai, line 144: Using `swap_remove` here is a good optimization for removing elements during iteration. Consider adding a brief comment explaining why this is preferred over other removal methods, as it's a non-obvious performance optimization:
```suggestion
                // Use swap_remove for O(1) removal without preserving order
                self.policies.swap_remove(i);
```

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -423,122 +423,120 @@ impl<'p> Filter<'p, Entity> {
...
+                    }),
+                ));
+            }
+            _ => {
+                // Use the Vec directly for the IN clause
+                permits.push(Self::In(
+                    FilterExpression::Path {
+                        path: EntityQueryPath::Uuid,
+                    },
+                    ParameterList::EntityUuids(&optimization_data.permitted_entity_uuids),
```
hashdotai, line 490: Good use of direct reference to the optimization data here. For additional clarity, consider adding a brief comment about the lifetime relationship:
```suggestion
                    ParameterList::EntityUuids(&optimization_data.permitted_entity_uuids), // Use direct reference to avoid copying
```

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -423,122 +423,120 @@
...
+                    }),
+                ));
+            }
+            _ => {
+                // Use the Vec directly for the IN clause
+                permits.push(Self::In(
+                    FilterExpression::Path {
+                        path: EntityQueryPath::WebId,
+                    },
+                    ParameterList::WebIds(&optimization_data.permitted_web_ids),
```
hashdotai, line 516: Similar to the previous comment, add a note about using a reference:
```suggestion
                    ParameterList::WebIds(&optimization_data.permitted_web_ids), // Use direct reference to avoid copying
```

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -423,122 +423,120 @@
...
-                // permit
-                (true, false) => Self::Any(permits),
-
-                // We have permits and forbids, so we allow access if there is at least one permit
+                (false, _) => Self::Any(Vec::new()), // No permits = deny all
+                (true, false) => Self::Any(permits), // Only permits
                 (true, true) => Self::All(vec![
+                    // Both permits and forbids
                     Self::Any(permits),
                     Self::Not(Box::new(Self::Any(forbids))),
```
hashdotai, line 535: I see the removal of commented code which is a good practice. However, it seems like parts of that removed code were reimplemented in the new optimization system. Consider adding a brief comment explaining that this replaces/improves upon the previously commented-out approach.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: We currently don't have real-world data to measure this. From local tests this increased performance by up to 75%, but this will very much depend on the actual data so I won't add this.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: Unneeded verbose comment

libs/@local/graph/store/src/filter/mod.rs
TimDiekmann: Plain wrong. We require a reference, this is the reason why we danced around with `OptimizationData` in the first place.

libs/@local/graph/store/src/filter/mod.rs
TimDiekmann: I won't add comments for things which are not there anymore.

.cursor/rules/rust-documentation.mdc
```diff
@@ -12,7 +12,7 @@ alwaysApply: false
 - Begin doc comments with a single-line summary of the item's purpose
 - Include a blank line after the summary line for longer documentation
 - Structure complex documentation with Markdown headers
-- Always use intra-doc links (`[`Item`]`) whenever referencing other items in the codebase
+- Always use intra-doc links (`[`Item`]`) whenever referencing other items in the codebase, including standard library types like [`Vec`], [`HashMap`], etc.
```
indietyp, line 15:
```suggestion
- Always use intra-doc links (`[`Item`]`) whenever referencing other items in the codebase, including standard library types like [`Vec`], [`HashMap`], [`Option<T>`], etc.
```

I've experienced that even with this it likes to skip prelude items. So maybe it'd make sense to mention one (`Option`) explicitly.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -51,6 +189,7 @@ pub struct PolicyComponentsBuilder<'a, S> {
     entity_type_ids: HashSet<Cow<'a, VersionedUrl>>,
     entity_edition_ids: HashSet<EntityEditionId>,
     actions: HashSet<ActionName>,
+    enable_optimization: bool,
```
indietyp, line 192: Does it make sense to even gate this? What would be scenarios where we'd want to disable the optimizations?

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -423,122 +423,120 @@ impl<'p> Filter<'p, Entity> {
...
+                    FilterExpression::Path {
+                        path: EntityQueryPath::WebId,
+                    },
+                    ParameterList::WebIds(&optimization_data.permitted_web_ids),
+                ));
+            }
+        }
+
+        // Apply the same combination logic as for_policies
         if blank_permit {
```
indietyp, line 522: I feel like we're doing a lot of additional work even if we just end up with a blank permit. (not that it is blocking, just what I observed)

TimDiekmann: A blank permit does not guarantee that you have permissions. If you hit a `Forbid` permission it has a higher precedence. E.g. I could say "Bilal can see all entities in my web" and then in another policy I could say "Forbid access to this entity unless I'm the user"

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: I currently kept it to compare different scenarios

PR #7497, opened 2025-06-26
H-4829: HashQL: Create `hashql-eval` crate

indietyp:

This crate will house all of the query evaluation methods, such as, but not limited to the graph query executor.

The current mechanism will rely on the HIR, but will transition to a MIR/LIR representation at a later stage.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/eval/src/lib.rs
+++ b/libs/@local/hashql/eval/src/lib.rs
@@ -1,4 +1,4 @@
-//! # HashQL Core
+//! # HashQL Eval
 //!
 //! ## Workspace dependencies
 #![cfg_attr(doc, doc = simple_mermaid::mermaid!("../docs/dependency-diagram.mmd"))]


libs/@local/hashql/eval/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+//!
+//! ## Workspace dependencies
+#![cfg_attr(doc, doc = simple_mermaid::mermaid!("../docs/dependency-diagram.mmd"))]
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn it_works() {
+        assert_eq!(2, 2); // if this isn't true, then something went *horribly* wrong in the universe.
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Thanks for the suggestion. While I understand the reasoning behind maintaining a professional tone, I think this particular comment serves a useful purpose - it communicates to other developers that this is an intentional placeholder test rather than something incomplete. The lighthearted tone also contributes to a positive development culture.

Since this is just a test file in a newly created crate that's likely to be replaced with actual tests later, I'd suggest keeping it as is for now. When actual functionality is implemented, we can replace this placeholder with meaningful tests that follow a more standard approach.

libs/@local/hashql/eval/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! # HashQL Core
```
hashdotai: The crate name in the doc comment doesn't match the actual crate name. This appears to be copied from another crate without proper modification.
```suggestion
//! # HashQL Eval
```

libs/@local/hashql/eval/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name              = "hashql-eval"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+publish.workspace = true
+version.workspace = true
+
+[dependencies]
+
```
hashdotai, line 18: The dependencies section is empty, but according to the PR overview, this crate will need to depend on the HIR crate to function. Consider adding the necessary dependencies now to set up the foundation properly.

libs/@local/hashql/eval/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[lints]
+workspace = true
+
+[dev-dependencies]
+simple-mermaid.workspace = true
```
hashdotai: I notice `simple-mermaid` is in `dev-dependencies` but is not being used in the tests. It's only used for documentation. Consider moving it to a `[build-dependencies]` section or add a comment explaining why it's in dev-dependencies.

libs/@local/hashql/eval/src/lib.rs
indietyp: fixed in [09667d9](https://github.com/hashintel/hash/pull/7497/commits/09667d9f60e137757ce8410e1b86495ff2b1299d)

libs/@local/hashql/eval/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: please keep it.

TimDiekmann (approved):
@vilkinsons Bilal and I agreed offline, that this is probably a good crate name to house the actual evaluation stuff for HashQL. I wanted to avoid several crates, e.g. for postgres, conversions, etc. and just house pretty much everything here. Needs a legal-approval.

vilkinsons (approved):

PR #7498, opened 2025-06-26
H-4879: Improve documentation of `hash-graph-postgres-store`

indietyp:

Improves the documentation of several types in `hash-graph-postgres-store`, by adding error documentation and `must_use` where required.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/mod.rs
@@ -69,6 +69,7 @@ pub trait PostgresQueryPath: Sized {
     fn terminating_column(&self) -> (Column, Option<JsonField<'_>>);
 
     #[expect(unused_variables, reason = "No-op")]
+    #[must_use]
     fn label_property_path(inheritance_depth: Option<u32>) -> Option<Self> {
         None
     }
@@ -77,6 +78,10 @@ pub trait PostgresQueryPath: Sized {
 /// Renders the object into a Postgres compatible format.
 pub trait Transpile: 'static {
     /// Renders the value using the given [`Formatter`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the value cannot be formatted or written to the formatter.
     fn transpile(&self, fmt: &mut Formatter) -> fmt::Result;
 
     fn transpile_to_string(&self) -> String {
@@ -98,8 +103,18 @@ pub trait PostgresSorting<'s, R: QueryRecord>:
 
     type Error: Error + Send + Sync + 'static + Send + Sync + 'static;
 
+    /// Encodes the sorting parameters for use in the query.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the sorting parameters cannot be encoded.
     fn encode(&self) -> Result<Option<Self::CompilationParameters>, Self::Error>;
 
+    /// Compiles the sorting into column selections and ordering expressions.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the sorting cannot be compiled into valid SQL expressions.
     fn compile<'p, 'q: 'p>(
         &'p self,
         compiler: &mut SelectCompiler<'p, 'q, R>,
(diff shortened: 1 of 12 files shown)


libs/@local/graph/postgres-store/src/store/postgres/query/mod.rs
```diff
@@ -69,6 +69,7 @@ pub trait PostgresQueryPath: Sized {
     fn terminating_column(&self) -> (Column, Option<JsonField<'_>>);
 
     #[expect(unused_variables, reason = "No-op")]
+    #[must_use]
```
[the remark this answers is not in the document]
indietyp: clippy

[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer @indietyp I agree we should remove the `#[must_use]` annotation from the `label_property_path` method. The `#[must_use]` attribute is intended for functions that return a value that should be used, but this method always returns `None` by default, so the annotation doesn't provide value and might cause confusing warnings. 

Also, to clarify @indietyp's response: while Clippy might recommend adding `#[must_use]` for methods that return values, in this specific case where the default implementation always returns `None`, the annotation is unnecessary and potentially misleading.

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
```diff
@@ -1881,6 +1889,7 @@ pub enum ForeignKeyReference {
 }
 
 impl ForeignKeyReference {
+    #[must_use]
```
hashdotai, line 1893: Consider clarifying the `ForeignKeyReference::reverse` method's documentation. Since it's now part of a public API, it would be helpful to add a brief doc comment explaining what this method does.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -85,6 +85,7 @@ pub struct InsertStatementBuilder<'p> {
 struct SliceWrapper<'a, T>(&'a [T]);
 
 impl<'p> InsertStatementBuilder<'p> {
+    #[must_use]
```
hashdotai, line 88: Consider adding return type documentation for functions with `#[must_use]` attributes. For example, this method would benefit from explaining what the returned `InsertStatementBuilder` represents or how it should be used.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -85,6 +85,7 @@ pub struct InsertStatementBuilder<'p> {
 struct SliceWrapper<'a, T>(&'a [T]);
```
hashdotai, line 85: Consider adding inline documentation to explain the purpose of the `SliceWrapper` struct and why it's needed.

libs/@local/graph/postgres-store/src/store/postgres/query/mod.rs
```diff
@@ -98,8 +103,18 @@ pub trait PostgresSorting<'s, R: QueryRecord>:
 
     type Error: Error + Send + Sync + 'static + Send + Sync + 'static;
```
hashdotai: The `PostgresSorting` trait's `Error` type has `Send + Sync` specified twice. Consider removing the duplication:
```suggestion
    type Error: Error + Send + Sync + 'static;
```

indietyp: nice catch! [bb9d286](https://github.com/hashintel/hash/pull/7498/commits/bb9d286daeabc95e65a824d224a304c212f97857)

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
indietyp: This would need to be part of a larger push. Quite a few functions are undocumented right now, so just adding it here wouldn't make much sense.

vilkinsons (dismissed):

TimDiekmann (approved):

PR #7499, opened 2025-06-26
H-4880, H-4881:  Introduce `Value` type

indietyp:

The `Value` type is used during evaluation for any value in `HashQL`, due to the constraints of the language it's immutable and uses structural sharing where possible.

This PR:

The changes in this PR:

The changes in this PR:


indietyp:
@CodiumAI-Agent /review

libs/@local/hashql/core/src/value/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+                    {
+                        integer
+                    }
+                    _ => return Err(IndexAccessError::InvalidListIndexType(self.type_name())),
+                };
+
+                let Some(index) = integer.as_usize() else {
+                    return Ok(None);
+                };
+
```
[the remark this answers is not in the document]
indietyp: Why? It's the only place that uses it.

libs/@local/hashql/core/src/literal/float.rs
```diff
@@ -34,12 +38,12 @@ pub(crate) const PARSE: ParseFloatOptions = match ParseFloatOptionsBuilder::new(
 /// ```
 ///
 /// [JSON specification (RFC 8259)]: https://datatracker.ietf.org/doc/html/rfc8259#section-6
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
```
hashdotai, line 41: Since this module heavily relies on comparison of values, it's great that you've implemented `PartialOrd` and `Ord` for all the types. This will ensure consistent behavior in containers and sorting operations.

libs/@local/hashql/core/src/literal/float.rs
```diff
@@ -83,6 +87,34 @@
...
+
+    /// Converts the float literal to an integer literal.
+    ///
+    /// Conversion is lossless, meaning that the resulting [`IntegerLiteral`] value will be exactly
+    /// equal to the original value.
+    ///
+    /// Assumes that the float literal is a valid JSON-formatted floating-point number.
+    #[must_use]
+    pub fn as_integer(self) -> Option<IntegerLiteral<'heap>> {
+        let is_integer = memchr::memchr3(b'.', b'e', b'E', self.value.as_bytes()).is_none();
```
hashdotai, line 167: This is a good approach to handling conversion to integers, particularly using `memchr` to efficiently check for decimal points or exponents. It makes the conversion both robust and fast.

libs/@local/hashql/core/src/literal/float.rs
```diff
@@ -83,6 +87,34 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the float literal is not a valid JSON-formatted floating-point number.
+    /// This should never happen for properly constructed AST nodes.
+    #[must_use]
+    pub fn as_real(self) -> Real {
+        Real::from_str(self.value.as_str())
+            .expect("float literal should be formatted according to JSON specification")
+    }
```
hashdotai, line 209: Consider adding a test for this case with various real-world examples, especially edge cases like very large integers that can be represented exactly as floats.

libs/@local/hashql/core/src/value/struct.rs
```diff
@@ -0,0 +1,405 @@
...
+    ) -> Self {
+        let fields = fields.into_iter();
+
+        let (mut fields, mut values): (SmallVec<_>, SmallVec<_>) = fields.collect();
+
+        co_sort(&mut fields, &mut values);
+
+        // This is an assert, as previous stages in the compilation should have ensured that there
+        // are no duplicate fields.
+        assert!(
```
hashdotai, line 123: Good defensive coding pattern here - using constants and an assertion will help catch potential issues early in the development cycle while being clear about expectations.

libs/@local/hashql/core/src/value/struct.rs
```diff
@@ -0,0 +1,405 @@
...
+/// let mut values = ['c', 'a', 'd', 'b', 'e'];
+///
+/// co_sort(&mut keys, &mut values);
+///
+/// assert_eq!(keys, [1, 1, 3, 4, 5]);
+/// assert_eq!(values, ['a', 'b', 'c', 'd', 'e']);
+/// ```
+#[expect(unsafe_code)]
+#[inline]
+fn co_sort<T: Ord, U>(lhs: &mut [T], rhs: &mut [U]) {
```
hashdotai, line 329: Excellent implementation of `co_sort`. The use of unsafe code is well-justified here for performance reasons, and you've added proper safety comments. The inclusion of property-based tests is particularly commendable for such a complex algorithm.

libs/@local/hashql/core/src/value/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+                    &Self::Primitive(LiteralKind::Integer(integer)) => integer,
+                    Self::Primitive(LiteralKind::Float(float))
+                        if let Some(integer) = float.as_integer() =>
+                    {
+                        integer
+                    }
+                    _ => return Err(IndexAccessError::InvalidListIndexType(self.type_name())),
+                };
+
+                let Some(index) = integer.as_usize() else {
```
hashdotai, line 273: This if-let guard feature usage is elegant and makes the early-return pattern more readable. Good use of the feature flag.

libs/@local/hashql/core/src/value/opaque.rs
```diff
@@ -0,0 +1,105 @@
...
+/// let user_id = Opaque::new(heap.intern_symbol("UserId"), string("user_12345"));
+/// let score = Opaque::new(heap.intern_symbol("Score"), integer("95"));
+///
+/// assert_eq!(user_id.name(), heap.intern_symbol("UserId"));
+/// assert_eq!(user_id.value(), &string("user_12345"));
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Opaque<'heap> {
+    name: Symbol<'heap>,
+    value: Rc<Value<'heap>>,
```
hashdotai, line 34: The use of structural sharing through `Rc` here is a good performance optimization, especially for immutable values where reference counting has minimal overhead compared to deep cloning.

libs/@local/hashql/core/src/value/dict.rs
```diff
@@ -0,0 +1,327 @@
...
+/// let updated_person = person.insert(string("email"), string("alice@example.com"));
+///
+/// assert_eq!(person.len(), 2); // Original unchanged
+/// assert_eq!(updated_person.len(), 3); // New dict has additional entry
+/// ```
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
+pub struct Dict<'heap> {
+    // We use a `RedBlackTreeMap` here, because otherwise the type would be unhashable, another
+    // possibility would be to use an immutable chunkmap.
+    values: RedBlackTreeMap<Value<'heap>, Value<'heap>>,
```
hashdotai, line 43: It would be beneficial to add a note in the documentation explaining the choice of `RedBlackTreeMap` here, particularly why it was chosen over other potential implementations (like an immutable hash map) to give developers context on the tradeoffs.

indietyp: ?

indietyp:
@CodiumAI-Agent /improve

indietyp:
@CodiumAI-Agent /describe

libs/@local/hashql/core/src/literal/float.rs
indietyp: done in [c41d6ba](https://github.com/hashintel/hash/pull/7499/commits/c41d6ba41ecca3e4ac69e5eca0f28b7ca510eb5b)[c41d6ba](https://github.com/hashintel/hash/pull/7499/commits/c41d6ba41ecca3e4ac69e5eca0f28b7ca510eb5b)

libs/@local/codec/src/numeric/real.rs
```diff
@@ -63,10 +68,25 @@ pub struct ConversionError(dashu_base::ConversionError);
...
+/// # Implementation Notes
+///
+/// - **Decimal radix** – The significand is stored in base-10, so values such as `0.1` and `0.01`
+///   round-trip without error (`Real::from_str("0.1")?.to_string() == "0.1"`).
+/// - **Software fallback** – CPUs only accelerate binary FP; decimal math runs in software and is
+///   typically **3–10x** slower and larger than `f64`.
+/// - **Deterministic text I/O** – Converting to and from strings is loss-free, which ensures that
+///   serialized data, logs, and config files preserve every digit.
+/// - **When to pick `Real`** – Prefer it for financial or audit-grade workflows where correctness
+///   outweighs raw throughput. For hot paths that tolerate a few ULPs, stick to `f32`/`f64`.
```
TimDiekmann: Let's not put a recommendation here currently. In our context, correctness is pretty much always more important as we don't know how the data is being used.

```suggestion
```

libs/@local/codec/src/numeric/real.rs
```diff
@@ -63,10 +68,25 @@ pub struct ConversionError(dashu_base::ConversionError);
...
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 #[cfg_attr(feature = "codegen", derive(specta::Type))]
 pub struct Real(#[cfg_attr(feature = "codegen", specta(type = f64))] dashu_float::DBig);
 
+// Minimum precision for internal representation (64 bits). This balances accuracy and performance,
+// providing higher precision than f64 (53 bits) while maintaining reasonable computational costs.
+// This gives us ~19 decimal digits of precision vs ~15 for f64, reducing rounding errors in
+// financial calculations and iterative computations.
```
TimDiekmann: In `rustdoc` this breaks formatting. Please use a single intro sentence and a more detailed description in a separate paragraph.

libs/@local/codec/src/numeric/real.rs
```diff
@@ -321,14 +350,22 @@ macro_rules! impl_real_from_primitive_float {
...
             fn try_from(primitive: $primitive) -> Result<Self, Self::Error> {
-                Ok(Self(
-                    dashu_float::FBig::<mode::Zero>::try_from(primitive)
-                        .map_err(ConversionError)?
-                        .with_precision(MIN_PRECISION)
-                        .value()
-                        .to_decimal()
-                        .value(),
-                ))
+                // The order of operations is important here! If we raise the precision first, it
```
TimDiekmann:
```suggestion
                // The order of operations is important here. If we raise the precision first, it
```

libs/@local/codec/src/numeric/real.rs
indietyp: This is a comment, not rustdoc comment. I can convert it to a rustdoc comment tho (might be more appropriate anyway)

TimDiekmann: Oh, didn't noticed that. Yes, let's make this a doctoring 🙂

libs/@local/codec/src/numeric/real.rs
indietyp: fixed in [27ccf69](https://github.com/hashintel/hash/pull/7499/commits/27ccf69ee94f2e03f9703b9e88b6b8ae58413f5c)

libs/@local/codec/src/numeric/real.rs
indietyp: 👨‍⚕️ in [27ccf69](https://github.com/hashintel/hash/pull/7499/commits/27ccf69ee94f2e03f9703b9e88b6b8ae58413f5c)

libs/@local/codec/src/numeric/real.rs
indietyp: removed in [27ccf69](https://github.com/hashintel/hash/pull/7499/commits/27ccf69ee94f2e03f9703b9e88b6b8ae58413f5c)

TimDiekmann (dismissed):

libs/@local/codec/src/numeric/real.rs
TimDiekmann: Thanks!

libs/@local/codec/src/numeric/real.rs
```diff
@@ -530,6 +612,25 @@ mod tests {
...
-        assert!(large_value.to_f32().is_infinite());
+        assert!(large_value.to_f32_lossy().is_infinite());
+    }
+
+    #[test]
+    #[expect(clippy::float_cmp, clippy::decimal_literal_representation)]
+    fn lossy_vs_exact_conversion() {
+        let inside_f32 = Real::from_natural(123_456, 0); // 123456 - needs ~17 bits
+        let inside_f64 = Real::from_natural(16_777_217, 0); // 2^24 + 1 - needs 25 bits
+        let outside_f64 = Real::from_natural(123_456_789, 20); // 123456789 * 10^20
```
TimDiekmann, line 631: Can't we just pass the exponent? e.g. `from_natural(1, 17)`? That's more intuitive.

indietyp: in theory yes, but the issue described only happens if the significand is larger than eg 53 bits, which wouldn't happen if we use an exponent because the exponent is saved separately

PR #7500, opened 2025-06-26
Add more instrumentations for graph queries

TimDiekmann:

To properly track where time is spent in different queries (and also for more precise logging output), this PR adds several instrumentation attributes. It was a bit back and forth as it was not clear to me first why some spans are not shown in the final flamegraph, but I solved this. The reasoning can be seen in CLAUDE.md.

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
@@ -177,39 +177,19 @@ where
             for (entity_vertex_id, graph_resolve_depths, traversal_interval) in
                 entity_queue.drain(..)
             {
-                let _span = tracing::trace_span!(
+                tracing::trace_span!(
                     "traverse_edges",
                     entity_id = %entity_vertex_id.base_id,
                     entity_revision = %entity_vertex_id.revision_id,
                     graph_resolve_depths = ?graph_resolve_depths,
                     traversal_interval = ?traversal_interval
                 )
-                .entered();
-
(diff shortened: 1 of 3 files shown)


libs/@local/graph/store/src/query/mod.rs
```diff
@@ -110,6 +111,7 @@ pub trait Read<R: QueryRecord>: Sync {
     ) -> impl Future<Output = Result<Vec<R>, Report<QueryError>>> + Send {
         self.read(filters, temporal_axes, include_drafts)
             .and_then(TryStreamExt::try_collect)
+            .in_current_span()
     }
```
TimDiekmann, line 115: Due to this missing `.in_context_span`, some spans were not shown.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -176,6 +177,15 @@ where
             for (entity_vertex_id, graph_resolve_depths, traversal_interval) in
                 entity_queue.drain(..)
             {
+                let _span = tracing::trace_span!(
```
indietyp: Could we maybe move this into an external function (or even closure) that isn't `async` or use https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.in_scope ? This is motivated by https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code. I could imagine us changing this code in the future and then suddenly having rogue `traverse_edges` span in the tree.

indietyp (approved):
one small nit/comment. nothing blocking

PR #7502, opened 2025-06-27
H-4891: Introduce `CreateEntity` permission and fix policy optimization logic

TimDiekmann:

This PR implements Cedar-based authorization for entity creation by introducing the `CreateEntity` action. Previously, entity creation authorization was handled through Zanzibar permissions, but as part of the migration to Cedar-based authorization, we now have a dedicated Cedar action that provides more granular and consistent authorization control for entity creation operations.

## 🔍 What does this change?

- **Adds `CreateEntity` Cedar action**: New action in the Cedar schema that inherits from `Create` and applies to Web resources
- **Implements per-action optimization**: Refactors `PolicyComponents` to support per-action optimization instead of global optimization, allowing selective optimization of different actions
- **Updates entity creation authorization**: Replaces Zanzibar-based entity creation checks with Cedar-based authorization using the new `CreateEntity` action
- **Adds default policies**: Creates default policies in `seed_policies.rs` that grant `CreateEntity` permission to web members and administrators
- **Removes legacy Zanzibar code**: Eliminates the unused `WebPermission::CreateEntity` checks and related imports
- **Improves PolicyComponents flexibility**: Builder now accepts `optimize: bool` parameter per action, enabling fine-grained control over which actions get optimized
- **Fixes action filtering bug**: Ensures `analyze_optimization_opportunities` properly filters policies by the target action
- **Updates documentation**: Clarifies the relationship between policy filtering and PolicySet evaluation

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- None

## 🐾 Next steps

- Continue migration of other Zanzibar permissions to Cedar actions
- Monitor performance impact of the new authorization flow
- Consider implementing the optional policy filtering optimization in `build_policy_set` if needed for performance

## 🛡 What tests cover this?

- Existing Cedar action hierarchy tests cover the new `CreateEntity` action
- Updated unit tests for `PolicyComponents` per-action optimization
- Integration tests for entity creation flow continue to work with Cedar authorization
- Policy validation tests ensure Cedar schema correctness

## ❓ How to test this?

1. Checkout the branch
2. Run the entity creation tests: `cargo test --package hash-graph-postgres-store entity`
3. Verify that entity creation still works for authorized users
4. Confirm that unauthorized users are properly blocked from creating entities
5. Test the authorization in different web contexts

## 📹 Demo

Entity creation authorization now uses Cedar policies instead of Zanzibar permissions, providing more consistent and maintainable authorization decisions across the platform.

### DIFF
--- a/libs/@local/graph/authorization/src/policies/components.rs
+++ b/libs/@local/graph/authorization/src/policies/components.rs
@@ -1,5 +1,5 @@
 use alloc::borrow::Cow;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use error_stack::{Report, ResultExt as _};
 use type_system::{
@@ -28,21 +28,12 @@ pub struct OptimizationData {
     pub permitted_web_ids: Vec<WebId>,
 }
 
-impl OptimizationData {
-    /// Returns true if any optimization opportunities were found.
-    #[must_use]
-    pub const fn has_optimizations(&self) -> bool {
-        !self.permitted_entity_uuids.is_empty() || !self.permitted_web_ids.is_empty()
-    }
-}
-
 #[derive(Debug)]
 pub struct PolicyComponents {
     actor_id: Option<ActorId>,
     policies: Vec<Policy>,
-    tracked_actions: HashSet<ActionName>,
+    tracked_actions: HashMap<ActionName, Option<OptimizationData>>,
     context: Context,
-    optimization_data: OptimizationData,
 }
 
 impl PolicyComponents {
@@ -63,21 +54,6 @@ impl PolicyComponents {
         &self.context
     }
 
-    /// Returns a reference to the policies for direct access.
-    ///
-    /// This is primarily used for filter optimization where the raw policies
-    /// need to be analyzed without building a full [`PolicySet`].
-    #[must_use]
-    pub fn policies(&self) -> &[Policy] {
-        &self.policies
-    }
-
-    /// Returns a reference to the tracked actions.
-    #[must_use]
-    pub const fn tracked_actions(&self) -> &HashSet<ActionName> {
-        &self.tracked_actions
-    }
-
     /// Extracts policy data suitable for database filter creation.
     ///
     /// This method returns an iterator of `(Effect, Option<&ResourceConstraint>)` tuples
@@ -89,7 +65,7 @@ impl PolicyComponents {
         action: ActionName,
     ) -> impl Iterator<Item = (Effect, Option<&ResourceConstraint>)> {
         debug_assert!(
-            self.tracked_actions.contains(&action),
+            self.tracked_actions.contains_key(&action),
             "action `{}` not tracked",
             action.to_string()
         );
@@ -108,8 +84,27 @@ impl PolicyComponents {
     /// `entity_uuid = a OR entity_uuid = b OR entity_uuid = c`
     /// to: `entity_uuid = ANY([a, b, c])`
     #[must_use]
-    pub const fn optimization_data(&self) -> &OptimizationData {
-        &self.optimization_data
+    pub fn optimization_data(&self, action: ActionName) -> &OptimizationData {
+        const EMPTY_OPTIMIZATION_DATA: &OptimizationData = &OptimizationData {
+            permitted_entity_uuids: Vec::new(),
+            permitted_web_ids: Vec::new(),
+        };
+
+        self.tracked_actions
+            .get(&action)
+            .unwrap_or_else(|| {
+                unreachable!("Action `{action}` is not tracked in this `PolicyComponents`")
+            })
+            .as_ref()
+            .unwrap_or_else(|| {
+                if cfg!(debug_assertions) {
+                    unreachable!("Action `{action}` is not optimized in this `PolicyComponents`")
+                }
+
+                tracing::warn!("No optimization data for action `{action}`");
+
+                EMPTY_OPTIMIZATION_DATA
+            })
     }
 
     /// Analyzes policies and extracts optimization opportunities.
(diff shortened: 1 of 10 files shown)


libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -127,13 +122,19 @@ impl PolicyComponents {
...
         let mut i = 0;
         while i < self.policies.len() {
-            let should_extract = match (&self.policies[i].effect, &self.policies[i].resource) {
+            let policy = &mut self.policies[i];
+            if !policy.actions.contains(&action) {
+                i += 1;
+                continue;
+            }
+
+            let should_extract = match (policy.effect, &policy.resource) {
```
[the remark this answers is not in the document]
hashdotai: Good catch @copilot-pull-request-reviewer! You're absolutely right. The function identifies optimizable policies but doesn't actually extract the entity UUIDs or web IDs into their respective sets.

The `should_extract` logic correctly identifies candidates for optimization, but the actual values need to be inserted into the sets. Here's how this should be fixed:

```rust
let should_extract = match (policy.effect, &policy.resource) {
    (
        Effect::Permit,
        Some(ResourceConstraint::Entity(EntityResourceConstraint::Exact { id })),
    ) => {
        entity_uuids_set.insert(id.clone());
        true
    }
    (
        Effect::Permit,
        Some(ResourceConstraint::Web { web_id }),
    ) => {
        web_ids_set.insert(*web_id);
        true
    }
    _ => false,
};
```

Without this change, the optimization data would remain empty, making the optimization ineffective. Thanks for identifying this issue!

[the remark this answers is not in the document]
TimDiekmann: Literally 1 line below -- **you even copied the code!**
```rust
                (
                    Effect::Permit,
                    Some(ResourceConstraint::Entity(EntityResourceConstraint::Exact { id })),
                ) => {
                    entity_uuids_set.insert(*id);
                    true
                }
                (Effect::Permit, Some(ResourceConstraint::Web { web_id })) => {
                    web_ids_set.insert(*web_id);
                    true
                }
```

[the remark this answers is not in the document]
hashdotai: @TimDiekmann You're absolutely right, and I apologize for the confusion. I missed the actual implementation in the code below where the insertion is already happening correctly:

```rust
(
    Effect::Permit,
    Some(ResourceConstraint::Entity(EntityResourceConstraint::Exact { id })),
) => {
    entity_uuids_set.insert(*id);
    true
}
(Effect::Permit, Some(ResourceConstraint::Web { web_id })) => {
    web_ids_set.insert(*web_id);
    true
}
```

Thank you for pointing this out. The code is correctly implementing the optimization by inserting the entity UUIDs and web IDs into their respective sets.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -510,8 +511,10 @@ mod tests {
             "should show basic error message"
         );
         assert!(
-            debug_message.contains("Cannot create PolicySet when optimization has occurred"),
-            "should include detailed error information"
+            debug_message.contains(
```
hashdotai, line 550: The error message format could be more consistent with the rest of the codebase. Consider adding the action name to the format string rather than interpolating it in the message:

```suggestion
            debug_message.contains(
                "Action `view` has been optimized and cannot be used to create a `PolicySet`"
            ),
```

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -171,21 +185,46 @@
...
+                    )),
+                );
+            }
         }
 
+        // Note: We don't need to filter policies by actions here because the PolicySet
+        // only evaluates actions that are in `tracked_actions`. Including all policies
+        // is safe since untracked actions won't be evaluated anyway.
+        //
+        // Optional performance optimization: Filter policies to reduce PolicySet size:
```
hashdotai, line 217: This comment suggests filtering policies for better performance but doesn't provide a clear recommendation on whether it should be implemented now or later. Consider either implementing it now or creating a follow-up task with more details about expected performance benefits.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -108,8 +84,27 @@
     /// `entity_uuid = a OR entity_uuid = b OR entity_uuid = c`
     /// to: `entity_uuid = ANY([a, b, c])`
     #[must_use]
-    pub const fn optimization_data(&self) -> &OptimizationData {
-        &self.optimization_data
+    pub fn optimization_data(&self, action: ActionName) -> &OptimizationData {
+        const EMPTY_OPTIMIZATION_DATA: &OptimizationData = &OptimizationData {
```
hashdotai, line 88: The constant `EMPTY_OPTIMIZATION_DATA` is only used in this function. Consider moving it to module scope to avoid recreating it on each function call, which would be more efficient.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -274,39 +312,28 @@
         self
     }
 
-    pub fn add_action(&mut self, action: ActionName) {
-        self.actions.insert(action);
+    pub fn add_action(&mut self, action: ActionName, optimize: bool) {
```
hashdotai, line 324: The `optimize` parameter is now being passed, but we're still missing documentation on this parameter. Consider adding documentation to clarify when optimization should be enabled or disabled for specific actions.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -992,42 +995,43 @@ where
...
+                    forbidden_web_creations.push(web_id);
+                }
             }
         }
 
+        if !forbidden_web_creations.is_empty() {
+            return Err(Report::new(InsertionError)
+                .attach(StatusCode::PermissionDenied)
+                .attach_printable(
+                    "The actor does not have permission to create entities in one or more webs",
```
hashdotai, line 1024: The error message mentions "one or more webs" while previous message mentioned "one or more web ids". Consider keeping terminology consistent:

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: There is a clear recommendation to not implement this, yet. We don't know this speeds up or not and need to be benched.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: It's a `const`.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -108,8 +84,27 @@ impl PolicyComponents {
     /// `entity_uuid = a OR entity_uuid = b OR entity_uuid = c`
     /// to: `entity_uuid = ANY([a, b, c])`
     #[must_use]
-    pub const fn optimization_data(&self) -> &OptimizationData {
-        &self.optimization_data
+    pub fn optimization_data(&self, action: ActionName) -> &OptimizationData {
+        const EMPTY_OPTIMIZATION_DATA: &OptimizationData = &OptimizationData {
```
indietyp, line 88: Wouldn't this be better served through a `Default` or `new` implementation instead?

TimDiekmann: `Default` is not `const` and adding a `new` just for a single code path which should never be executed is not worth it IMO.

PR #7503, opened 2025-06-27
H-4882, H-4883, H-4897: HashQL: Save the name of the opaque type in the HIR

indietyp:
### HashQL: Save the name of the opaque type in the HIR

Currently the type constructor in the HIR is unaware of the opaque type it's constructing. To be able to construct a correct `Value` type, the name of the opaque type needs to be known.

### HashQL: Continue to lower the body of closures when specialising graph read operations

The initial implementation erroneously skipped this step.

### HashQL: Provide a general `lower` function in the HIR

This function is used to properly orchestrate and call the different lower operations outside of tests, where we're mostly interested in the output of all stages one after the other.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/hir/src/lower/mod.rs
+++ b/libs/@local/hashql/hir/src/lower/mod.rs
@@ -1,6 +1,83 @@
+use hashql_ast::lowering::ExtractedTypes;
+use hashql_core::{module::ModuleRegistry, r#type::environment::Environment};
+
+use self::{
+    alias::AliasReplacement,
+    checking::TypeChecking,
+    ctor::ConvertTypeConstructor,
+    error::{LoweringDiagnostic, LoweringDiagnosticCategory},
+    inference::TypeInference,
+    specialization::Specialization,
+};
+use crate::{fold::Fold as _, intern::Interner, node::Node, visit::Visitor as _};
+
 pub mod alias;
 pub mod checking;
 pub mod ctor;
 pub mod error;
 pub mod inference;
 pub mod specialization;
+
+/// Lowers the given node by performing different phases.
+///
+/// This will set the "substitution" field of the environment given.
+///
+/// # Errors
+///
+/// Returns a vector of `LoweringDiagnostic` if any errors occurred during lowering.
+///
+/// The vector is guaranteed to be non-empty.
+pub fn lower<'heap>(
+    node: Node<'heap>,
+    types: &ExtractedTypes<'heap>,
+    env: &mut Environment<'heap>,
+    registry: &ModuleRegistry<'heap>,
+    interner: &Interner<'heap>,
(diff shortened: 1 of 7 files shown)


libs/@local/hashql/hir/src/lower/specialization/mod.rs
```diff
@@ -206,6 +206,8 @@ impl<'env, 'heap> Specialization<'env, 'heap> {
                     return Ok(None);
                 };
 
+                let read = fold::walk_graph_read(self, read)?;
```
hashdotai, line 209: Good catch on adding the missing fold operation for graph read - this addresses the H-4882 ticket.

libs/@local/hashql/hir/tests/ui/lower/specialization/collect-filter-graph.stdout
```diff
@@ -21,7 +21,7 @@
     EntityUuid[Uuid[String]], web_id:
     WebId[ActorGroupEntityUuid[Uuid[String]]])])]]
   | None[Null]), properties: ?)]»): _1«Boolean» ->
-    ::core::cmp::eq(vertex:0.id.entity_id.entity_uuid, #ctor(fn(Uuid[String]) ->
+    (vertex:0.id.entity_id.entity_uuid == #ctor(fn(Uuid[String]) ->
```
hashdotai, line 24: It looks like the test output format has changed from using `::core::cmp::eq` to using the `==` operator directly. Is this an intentional change? If so, it might be worth mentioning in the PR description since it affects test output.

libs/@local/hashql/hir/src/node/operation/type.rs
```diff
@@ -47,6 +48,8 @@ pub struct TypeAssertion<'heap> {
 pub struct TypeConstructor<'heap> {
     pub span: SpanId,
 
+    /// The name of the opaque type being constructed.
```
hashdotai, line 51: Good documentation on the newly added field. This is clearly needed for the requirements in H-4883.

libs/@local/hashql/hir/src/lower/mod.rs
```diff
@@ -1,6 +1,83 @@
...
+use crate::{fold::Fold as _, intern::Interner, node::Node, visit::Visitor as _};
+
 pub mod alias;
 pub mod checking;
 pub mod ctor;
 pub mod error;
 pub mod inference;
 pub mod specialization;
+
+/// Lowers the given node by performing different phases.
```
hashdotai, line 21: I see you've moved the `lower` function implementation from another file. While this is likely a good organizational move, it would be helpful to mention this in the PR description to explain the large number of added lines that aren't directly related to the tickets.

TimDiekmann (approved):

PR #7504, opened 2025-06-27
H-4884: Allow resolving the query temporal axes relative to a specific now

indietyp:

Currently the `QueryTemporalAxesUnresolved` can only be resolved to the current time (`now`) as it means that we're unable to reliably use it during tests that make use of snapshots, as the time is **(usually)** monotonically increasing.

This changes it to introduce a new method that takes a `Timestamp` as `now`, and uses that instead.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/store/src/subgraph/temporal_axes.rs
+++ b/libs/@local/graph/store/src/subgraph/temporal_axes.rs
@@ -256,20 +256,20 @@ impl Default for QueryTemporalAxesUnresolved {
 }
 
 impl QueryTemporalAxesUnresolved {
-    /// Resolves temporal axes relative to the specified timestamp.
+    /// Resolves temporal axes using the provided timestamp.
     ///
-    /// Converts unresolved temporal axes into concrete temporal axes by resolving
-    /// relative timestamps and intervals against the provided reference time.
+    /// Converts unresolved temporal axes into concrete temporal axes by resolving unset values
+    /// with the provided timestamp.
     #[must_use]
-    pub fn resolve_relative_to(self, now: Timestamp<()>) -> QueryTemporalAxes {
+    pub fn resolve_with(self, timestamp: Timestamp<()>) -> QueryTemporalAxes {
         match self {
             Self::DecisionTime { pinned, variable } => QueryTemporalAxes::DecisionTime {
-                pinned: pinned.resolve(now),
-                variable: variable.resolve(now),
+                pinned: pinned.resolve(timestamp),
+                variable: variable.resolve(timestamp),
             },
             Self::TransactionTime { pinned, variable } => QueryTemporalAxes::TransactionTime {
-                pinned: pinned.resolve(now),
-                variable: variable.resolve(now),
+                pinned: pinned.resolve(timestamp),
+                variable: variable.resolve(timestamp),
             },
         }
     }
@@ -281,7 +281,7 @@ impl QueryTemporalAxesUnresolved {
     #[must_use]
     pub fn resolve(self) -> QueryTemporalAxes {
         let now = Timestamp::now();
-        self.resolve_relative_to(now)
+        self.resolve_with(now)
     }
 }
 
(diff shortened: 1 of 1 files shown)


libs/@local/graph/store/src/subgraph/temporal_axes.rs
```diff
@@ -257,8 +257,7 @@ impl Default for QueryTemporalAxesUnresolved {
 
 impl QueryTemporalAxesUnresolved {
     #[must_use]
-    pub fn resolve(self) -> QueryTemporalAxes {
```
hashdotai, line 260: The original function is being removed here, but it's better to maintain the original function and have it call the new function. I see you've done exactly that later in the diff, which is good - just noting this for context.

libs/@local/graph/store/src/subgraph/temporal_axes.rs
```diff
@@ -257,8 +257,7 @@
...
-        let now = Timestamp::now();
+    pub fn resolve_relative_to(self, now: Timestamp<()>) -> QueryTemporalAxes {
```
hashdotai: Good addition of the more flexible `resolve_relative_to` function. However, you should add a brief doc comment explaining what this function does and what the `now` parameter represents, especially since this is a public API.

libs/@local/graph/store/src/subgraph/temporal_axes.rs
```diff
@@ -270,6 +269,12 @@
             },
         }
     }
+
+    #[must_use]
+    pub fn resolve(self) -> QueryTemporalAxes {
```
hashdotai, line 282: Consider adding a brief doc comment here as well to clarify that this resolves relative to the current time. This helps maintain API documentation consistency with the new function.

libs/@local/graph/store/src/subgraph/temporal_axes.rs
indietyp: done in [02863c3](https://github.com/hashintel/hash/pull/7504/commits/02863c3598b6824822d185155760628578cf6625)

libs/@local/graph/store/src/subgraph/temporal_axes.rs
indietyp: done in [02863c3](https://github.com/hashintel/hash/pull/7504/commits/02863c3598b6824822d185155760628578cf6625)

libs/@local/graph/store/src/subgraph/temporal_axes.rs
```diff
@@ -256,9 +256,12 @@ impl Default for QueryTemporalAxesUnresolved {
...
 
 impl QueryTemporalAxesUnresolved {
+    /// Resolves temporal axes relative to the specified timestamp.
+    ///
+    /// Converts unresolved temporal axes into concrete temporal axes by resolving
+    /// relative timestamps and intervals against the provided reference time.
     #[must_use]
-    pub fn resolve(self) -> QueryTemporalAxes {
-        let now = Timestamp::now();
+    pub fn resolve_relative_to(self, now: Timestamp<()>) -> QueryTemporalAxes {
```
TimDiekmann: I don't understand the name. Why do you call it `relative`? Also, I would not call the variable `now` but `timestamp`.
Maybe `fn resolve_with(self, timestamp: Timestamp)`

TimDiekmann (changes requested):

libs/@local/graph/store/src/subgraph/temporal_axes.rs
indietyp: primarily because I couldn't come up with a better name than that. `resolve_with` is a lot better^^

indietyp: fixed in [4603e9e](https://github.com/hashintel/hash/pull/7504/commits/4603e9e4e621e5fd1108b577519a8d044f73d13c)

PR #7505, opened 2025-06-27
H-4885: Allow for incremental building of `JsonPath`

indietyp:

Currently a `JsonPath` is constructed in an all or nothing fashion from a vector of path tokens. This changes it so that the `JsonPath` can be built up incrementally by pushing tokens.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/store/src/filter/path.rs
+++ b/libs/@local/graph/store/src/filter/path.rs
@@ -47,6 +47,12 @@ pub struct JsonPath<'p> {
 }
 
 impl<'p> JsonPath<'p> {
+    /// Creates a new empty JSON path.
+    #[must_use]
+    pub const fn new() -> Self {
+        Self { path: Vec::new() }
+    }
+
     /// Creates a new JSON path from a sequence of path tokens.
     #[must_use]
     pub const fn from_path_tokens(path: Vec<PathToken<'p>>) -> Self {
@@ -86,6 +92,12 @@ impl<'p> JsonPath<'p> {
                 })
                 .collect(),
         }
+    }
+}
+
+impl Default for JsonPath<'_> {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/libs/@local/graph/store/src/filter/path.rs
+++ b/libs/@local/graph/store/src/filter/path.rs
@@ -9,6 +9,9 @@ use bytes::BytesMut;
 use postgres_types::{IsNull, ToSql, Type};
 use serde::{Deserialize, Serialize, Serializer};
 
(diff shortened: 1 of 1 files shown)


libs/@local/graph/store/src/filter/path.rs
```diff
@@ -27,6 +27,10 @@ impl<'p> JsonPath<'p> {
         Self { path }
     }
 
+    pub fn push(&mut self, token: PathToken<'p>) {
```
hashdotai, line 63: Consider adding a doc comment for the `push` method that explains its purpose and usage. This will make the API more self-documenting.

```suggestion
    /// Adds a new token to the end of the path.
    /// 
    /// This allows for incremental building of a JSON path by adding tokens one at a time.
    pub fn push(&mut self, token: PathToken<'p>) {
```

libs/@local/graph/store/src/filter/path.rs
```diff
@@ -27,6 +27,10 @@
...
+        self.path.push(token);
+    }
```
hashdotai, line 65: Consider adding a convenience method to create an empty `JsonPath` since we're now supporting the incremental building pattern. This would allow users to start with an empty path and then push tokens to it.

```suggestion
    }

    /// Creates a new empty JSON path with no tokens.
    /// 
    /// Use this with `push()` to build a path incrementally.
    pub fn new_empty() -> Self {
        Self { path: Vec::new() }
    }
```

libs/@local/graph/store/src/filter/path.rs
indietyp: done in [e45375d](https://github.com/hashintel/hash/pull/7505/commits/e45375d4abe1c96b085e0a0231f7db66260b9f63)

libs/@local/graph/store/src/filter/path.rs
indietyp: No. You just do `JsonPath::from_path_tokens(vec![])`.

TimDiekmann: As you added a `push` method, it would be natural to actually add a `new()` and `with_capacity(usize)` method so it feels more like a `Vec`.

TimDiekmann (dismissed):

libs/@local/graph/store/src/filter/path.rs
indietyp: added in [dfbec60](https://github.com/hashintel/hash/pull/7505/commits/dfbec60a3bfac44ca8e94d024674ec1efdfecbf5)

PR #7507, opened 2025-06-28
H-4821: Convert HIR read pipeline into query filters

indietyp:

Converts the HIR representation into a set of `Filter`s which can then be used in the `SelectCompiler` (e.g. through the API) to query the graph.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/eval/tests/compiletest.rs
```diff
@@ -0,0 +1,3 @@
+use hashql_compiletest::compiletest_main;
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security This is expected and not a real issue. The `compiletest_main` is a macro provided by the `hashql_compiletest` crate, which is correctly imported in the PR. The macro expands to the actual test harness code. The dependency is correctly added in the `Cargo.toml` file as a dev-dependency, and the code works correctly as can be seen in the PR tests that are passing. This is a pattern used across other packages in the codebase for handling compile tests.

libs/@local/hashql/eval/tests/compiletest.rs
```diff
@@ -0,0 +1,3 @@
...
+
+compiletest_main!();
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security This alert is a false positive. The `compiletest_main!()` macro call on line 3 expands to create the necessary `main` function for the test harness. This is a common pattern used throughout the codebase for test harnesses, and it's working correctly as shown by the successful test runs in the PR. The macro expansion happens at compile time, which is why the static analyzer is not able to detect the generated `main` function.

libs/@local/hashql/compiletest/src/suite/common.rs
```diff
@@ -33,6 +33,26 @@ where
     fatal.map_or(Ok(()), Err)
 }
 
+#[track_caller]
```
hashdotai, line 36: This function should include a brief doc comment explaining its purpose, especially since it's a non-trivial helper function that's used across the codebase.

libs/@local/hashql/compiletest/src/suite/common.rs
```diff
@@ -33,6 +33,26 @@
...
+pub(crate) fn process_result<C, T, I>(
+    output: &mut Vec<SuiteDiagnostic>,
+    result: Result<T, I>,
+) -> Result<T, SuiteDiagnostic>
```
hashdotai, line 40: For clarity, the return type could be more specific, something like `Result<T, DiagnosticError>` instead of using the generic `SuiteDiagnostic` type. This would make the error handling flow more obvious to readers.

libs/@local/hashql/eval/src/graph/read/path.rs
```diff
@@ -0,0 +1,356 @@
...
+pub(crate) enum PartialEntityIdQueryPath {
+    WebId,
+    EntityUuid,
+    DraftId,
+}
+
+impl<'heap> PartialQueryPath<'heap> for PartialEntityIdQueryPath {
+    type QueryPath = EntityQueryPath<'heap>;
+
+    fn from_field(heap: &'heap Heap, field: Symbol<'heap>) -> Option<Self> {
```
hashdotai, line 67: Consider adding error handling for unsupported field types in `from_field`. Currently it returns `None` for unrecognized fields, but there's no error or warning to help developers understand why their field might not be working.

libs/@local/hashql/eval/src/graph/read/path.rs
```diff
@@ -0,0 +1,356 @@
...
+        // see: https://linear.app/hash/issue/H-4512/hashql-use-symbol-sym-constants-instead-of-string-literals
+        let web_id = heap.intern_symbol("web_id");
```
hashdotai, line 69: This is using hard-coded strings for field names. Consider creating constants for these frequently used field names, or better yet, add a reference to the tracking issue for symbol constants (H-4512) as a TODO to remind future developers to update this when that work is completed.

libs/@local/hashql/eval/src/graph/read/path.rs
```diff
@@ -0,0 +1,356 @@
...
+                edge_kind: KnowledgeGraphEdgeKind::HasRightEntity,
+                path: Box::new(partial.finish()?),
+                direction: EdgeDirection::Outgoing,
+            }),
+            Self::RightEntityId(None) => None,
+        }
+    }
+}
+
+fn value_as_usize(value: &Value<'_>) -> Option<usize> {
```
hashdotai, line 233: The `value_as_usize` function is quite specific in its validation. Consider adding more robust error handling or at least a comment explaining what happens with negative numbers or large floats that don't fit in usize.

libs/@local/hashql/eval/tests/ui/graph/read/entity/unsupported-type-constructor.stderr
```diff
@@ -0,0 +1,35 @@
...
+    │ │            ──┬─  
+    │ │              ╰─── Cannot use constructor as value here
+    ┆ ┆   
+ 13 │ ├─▶       ]
+    │ │             
+    │ ╰───────────── ... within this filter expression
+    │     
+    │     Help: Constructor functions cannot currently be used as first-class values in filter expressions. You can still call constructors to create values (e.g., `SomeType(x)`), but you cannot use the constructor itself in comparisons or pass it as an argument within filter contexts.
+    │     
+    │     Note: This is a current limitation of the filter expression compiler. Constructors work as first-class values elsewhere in the language, and support for this in filter expressions is being tracked in [Linear Issue].
```
hashdotai: These tests include placeholder strings like `[Linear Issue]` in the error messages. Consider replacing them with actual issue links or tracking numbers, or adding a TODO comment to update them later.

libs/@local/hashql/compiletest/src/suite/common.rs
indietyp: This is an internal function.

libs/@local/hashql/eval/src/graph/read/path.rs
indietyp: Yes, because any transition would be a typecheck error.

libs/@local/hashql/eval/src/graph/read/path.rs
indietyp: No, because again - typecheck + it returns `Option` anyway.

libs/@local/hashql/eval/src/graph/read/path.rs
indietyp: These are quasi constants. The linked PR shows how we would convert it to proper constants.

libs/@local/hashql/eval/tests/ui/graph/read/entity/unsupported-type-constructor.stderr
indietyp: done in [12fc134](https://github.com/hashintel/hash/pull/7507/commits/12fc134ba10383208ba72866fcf0d774cf7ebde0)

libs/@local/hashql/compiletest/src/suite/common.rs
TimDiekmann: I don't thank the an internal function in general means we should skip docs, but I 100% agree with you in this case that this does not need a doctoring. It's clear what it's doing, even though the signature is not necessarily easy. Anyway, it's in the compiletest suite.

libs/@local/hashql/compiletest/src/suite/common.rs
TimDiekmann: And... why?

libs/@local/hashql/eval/Cargo.toml
```diff
@@ -8,13 +8,34 @@ version.workspace = true
...
 
 # Private third-party dependencies
+derive_more    = { workspace = true, features = ["display"] }
 simple-mermaid = { workspace = true }
 
+[dev-dependencies]
+hashql-compiletest = { workspace = true }
+
+[features]
+graph = ["dep:hash-graph-store", "dep:hashql-core", "dep:hashql-hir", "dep:type-system"]
```
TimDiekmann, line 29: I wonder if we want to start documenting the features somehow. I guess, that this enables evaluation to a `Filter` (reviewing the files in order, have not seen the implementation, yet), but that's not 100% clear.

libs/@local/hashql/eval/src/graph/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod read;
```
TimDiekmann, line 1: [Nitpick]
Suggestions on name. `read` is fine, no objections, but in alignment of H-4538, we might want to chose a different name
- `query`
- `select`
- `tail`

libs/@local/hashql/eval/Cargo.toml
indietyp: Yes, tracked in https://linear.app/hash/issue/H-4916/hashql-addupdate-readme-contents-of-crates.

I plan on updating the README, as well as creating several _flowcharts_ as part of the GB, which I'll be adding to the repository the GB is done/during it, as I need to document it either way.

libs/@local/hashql/eval/src/graph/mod.rs
indietyp: In that case, I'd gravitate towards `Select` (so that in the future we can have `Update` and `Delete`, normal db terminology). I would do that rename (as it also touches a fair part of the HIR) in a separate PR, is that okay?

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -0,0 +1,158 @@
...
+            }) => {
+                let func = match op.kind {
+                    BinOpKind::And => {
+                        let (left, right) = (
+                            self.compile_filter(context, left),
+                            self.compile_filter(context, right),
+                        );
+
+                        return Ok(Filter::All(vec![left?, right?]));
+                    }
```
TimDiekmann, line 89: I potentially see an issue here (which I expect can be solved later). If we have a filter with 100k entity Ids all OR-ed together, e.g. `entity_id == X OR entity_id == Y ...`, this nests _very_ deep. This is one of the reason why we have `All` as `Filter` and not `And` (and `Any` instead of `Or`). Any plans for optimizing this as a special node?

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -0,0 +1,158 @@
...
+                                span: _,
+                                value,
+                                r#type: _,
+                                force: _,
+                            }),
+                    }),
+            }) => self.compile_filter(context, value),
+            // If we came to this match arm using these nodes, then that means that the filter
+            // must've evaluated to a boolean expression. Therefore we can just check:
+            // did the expression evaluate to true?
```
TimDiekmann:
```suggestion
            // If we came to this match arm using these nodes, then that means that the filter
            // must have evaluated to a boolean expression. Therefore we can just check if the
            // expression evaluates to true.
```

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -0,0 +1,158 @@
...
+            | NodeKind::Graph(_) => Ok(Filter::Equal(
+                Some(
+                    self.compile_filter_expr(context, node)?
+                        .finish(context, &mut self.diagnostics)?,
+                ),
+                Some(FilterExpression::Parameter {
+                    parameter: Parameter::Boolean(true),
+                    convert: None,
+                }),
+            )),
```
TimDiekmann, line 155: Doesn't this recurse infinitely? The `node` is the same as the input, so it will hit this branch again.

TimDiekmann: Ah, I guess it's the `.finish()` which is doing the trick?

libs/@local/hashql/eval/tests/ui/graph/read/entity/arithmetic-comparisons.jsonc
```diff
@@ -0,0 +1,20 @@
...
+        [">", {"#literal": 1}, {"#literal": 2}],
+        ["||",
+          [">=", {"#literal": 1}, {"#literal": 2}],
+          ["||",
+            ["<", {"#literal": 1}, {"#literal": 2}],
+            ["<=", {"#literal": 1}, {"#literal": 2}]
+          ]
+        ]
+      ]
+    ]
```
TimDiekmann, line 18: As mentioned above: the nested structure of `||` will become an issue.

TimDiekmann (approved):
Amazing work!

A few comments, but all non-blocking.

For the module name I'd probably go with `query`, but we can change that later when we have more modules.

libs/@local/hashql/eval/src/graph/read/filter.rs
hashdotai: @TimDiekmann Yes, exactly. The `.finish()` is the key here - it transforms the intermediate representation into the final `FilterExpression` which is a terminal state. So we're not directly recursing with the same node, but rather going through the transformation pipeline.

Basically, that line says "if you've reached this point with a non-special node type, just compile it as a filter expression and check if it equals `true`".

libs/@local/hashql/eval/tests/ui/graph/read/entity/arithmetic-comparisons.jsonc
hashdotai: @TimDiekmann Agreed, as we discussed in the earlier comment, the nested OR structure will need optimization. I'll make sure the ticket I create addresses both the general case as well as this specific test case example. We'll need a strategy to flatten these nested logical operations for better performance and resource usage.

libs/@local/hashql/eval/src/graph/read/filter.rs
hashdotai: @TimDiekmann Thanks for the clarification! You're right, that's more accurate. I'll update the comment in a follow-up.

libs/@local/hashql/eval/src/graph/read/filter.rs
hashdotai: @TimDiekmann That's a very good point! You're right about the nesting issue with large OR conditions. I've been thinking about this and plan to implement an optimization for this specific case.

Specifically for entity IDs and other common equality patterns, we can detect patterns like `id == X OR id == Y OR id == Z...` and transform them into a more efficient `id IN [X, Y, Z, ...]` predicate, which would be a new filter type like `Filter::OneOf(path, values)`. 

I'll create a ticket to track this optimization so we can address it in a follow-up PR. We definitely don't want deeply nested OR trees with thousands of elements.

libs/@local/hashql/eval/Cargo.toml
hashdotai: @TimDiekmann You're right, documentation would be helpful here. As @indietyp mentioned, this is being tracked in H-4916. The plan to add README updates and flowcharts during/after the GB work sounds good - having visual documentation for these query transformations will be particularly valuable given the complexity of the conversion process.

libs/@local/hashql/eval/src/graph/mod.rs
hashdotai: @TimDiekmann @indietyp I agree that `Select` would be a good choice that aligns with standard DB terminology, making it more intuitive when we add `Update` and `Delete` in the future. Handling the rename in a separate PR makes sense given it would touch HIR and other components - it's a cleaner approach that keeps this PR focused on the filter conversion functionality.

libs/@local/hashql/compiletest/src/executor/trial.rs
```diff
@@ -210,7 +210,7 @@ impl Trial {
     }
 
     fn run_impl(&self, context: &TrialContext) -> Result<(), Report<[TrialError]>> {
-        let heap = Heap::new();
+        let heap = Heap::with_capacity(4 * 1024 * 1024); // 4MiB
```
hashdotai, line 215: The memory allocation is significantly increased here from the default. Since this is used for test execution, could you add a brief comment explaining why a 4MiB allocation is needed? Is it to accommodate large test cases or to prevent reallocations during complex test scenarios?

libs/@local/hashql/compiletest/src/suite/common.rs
hashdotai, line 36: This is a good addition! `process_result` provides a consistent error handling pattern that simplifies the code in several places. It would be good to add a brief docstring explaining its purpose, especially since it's being used across multiple test suites.

libs/@local/hashql/syntax-jexpr/src/parser/string/ident.rs
```diff
@@ -57,7 +57,7 @@ fn is_symbol(char: char) -> bool {
     match char.as_ascii().map(|char| char as u8) {
         Some(
             b'!' | b'#' | b'$' | b'%' | b'&' | b'*' | b'+' | b'.' | b'/' | b'<' | b'=' | b'>'
-            | b'?' | b'@' | b'\\' | b'^' | b'|' | b'-' | b'~',
+            | b'?' | b'@' | b'\\' | b'^' | b'|' | b'-' | b'~' | b'[' | b']',
```
hashdotai, line 60: The symbol set is expanded to include `[` and `]`. This is a good change that makes indexing operations more natural in expressions. Consider adding a test case that specifically verifies these new symbols work correctly in identifiers.

libs/@local/hashql/syntax-jexpr/src/parser/string/expr.rs
```diff
@@ -48,6 +48,7 @@ where
             }),
         )),
     )
+    .context(StrContext::Label("field"))
```
hashdotai, line 51: This is a helpful addition to provide better error messages by adding context to the field parsing. It will make debugging easier for users when they encounter syntax errors in field access expressions.

libs/@local/hashql/eval/src/graph/read/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+impl FilterCompilerContext<'_> {
+    const fn without_current_span(self) -> Self {
+        Self {
+            span: self.span,
+            current_span: None,
+            param_name: self.param_name,
+        }
+    }
+
+    const fn with_current_span(self, span: SpanId) -> Self {
```
hashdotai, line 61: The `FilterCompilerContext` provides a clean way to track contextual information during compilation. The `with_current_span` and `without_current_span` methods are particularly useful for maintaining the right error context during traversal. Good design!

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -0,0 +1,158 @@
...
+
+                        return Ok(Filter::All(vec![left?, right?]));
+                    }
+                    BinOpKind::Or => {
+                        let (left, right) = (
+                            self.compile_filter(context, left),
+                            self.compile_filter(context, right),
+                        );
+
+                        return Ok(Filter::Any(vec![left?, right?]));
```
hashdotai, line 96: When handling nested `Any`/`All` filter cases, consider future optimization for large sets of conditions. Deep nesting of OR conditions (e.g., with many entity IDs) could be inefficient. A future enhancement could flatten these into a more optimized structure like `IN` predicates.

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: I expected that the `SelectCompiler` optimizes these during the compliation into `Conditions`. I can instead take a look at doing that flattening directly inside this stage instead.

TimDiekmann: The query compiler cannot optimize this, because the data would need to be past as `&[EntityUuid]` instead of many `&EntityUuid` (if the slice contains references or not doesn't matter, the lifetime to the reference to the slice matters). This is why we add an [optimization step to the `Filter`](https://github.com/hashintel/hash/pull/7496). We should not try to optimize this as part of this PR, though.

TimDiekmann (approved):

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: Then my first pass of optimization would be to simply flatten the `And` and `Or` and then take a look at other optimization steps, but I am planning these more sophisticated in the DA.

TimDiekmann: That's a very low-hanging fruit, I think it makes sense. Also to move more sophisticated optimizations to the DA. Thank you!

indietyp: tracked in https://linear.app/hash/issue/H-4924/hashql-flatten-nested-and-and-or-operations-during-read-compilation

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: These are two different functions. `compile_filter` is for the top level (where `&&` and `||` are allowed) while `compile_filter_expr` is for filter expressions.

PR #7509, opened 2025-06-29
H-4899: HashQL: Fix contravariance applying unilaterally

indietyp:

Previously when we set something as being contravariant, we erroneously made it so that anytime e.g. `is_subtype_of` is called it'd change the arguments, instead this should've been done *once*.

This reworks the `Variance` into a `VarianceState` in conjunction with a `VarianceFlow`, which is used to switch the arguments only when an actual transition occurs.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/compiletest/src/executor/trial.rs
```diff
@@ -210,7 +210,10 @@ impl Trial {
     }
 
     fn run_impl(&self, context: &TrialContext) -> Result<(), Report<[TrialError]>> {
-        let heap = Heap::new();
+        // This is *way more* than we need, but allows us to avoid allocating a new slice, which we
+        // don't do for speed, but instead do because it keeps the indices stable during printing of
+        // types as we don't allocate a new block.
+        let heap = Heap::with_capacity(4 * 1024 * 1024); // 4MiB
```
hashdotai, line 218: Nice improvement with the heap capacity increase. Consider adding a constant for this value (e.g., `const HEAP_CAPACITY: usize = 4 * 1024 * 1024;`) to make it more maintainable and document its purpose better.

libs/@local/hashql/core/src/type/environment/context/variance.rs
```diff
@@ -1,219 +1,264 @@
...
+    /// Computes the composition of the current variance with `local` and updates
+    /// the internal state. Returns a guard that restores the previous variance
+    /// when dropped, along with the flow direction of this transition.
+    pub fn transition(&self, local: Variance) -> (VarianceGuard, VarianceFlow) {
+        let current = self.get();
+
+        let next = NEXT[current.into_index() * 3 + local.into_index()];
+        let flow = FLOW[current.into_index() * 3 + local.into_index()];
+
+        // replace is const, set isn't
```
hashdotai, line 71: Good use of a comment here explaining why you're using `replace` instead of `set`.

libs/@local/hashql/core/src/type/environment/context/variance.rs
```diff
@@ -1,219 +1,264 @@
...
+pub struct VarianceGuard {
+    state: Weak<Cell<Variance>>,
+    previous: Option<Variance>,
+}
+
+impl Drop for VarianceGuard {
+    fn drop(&mut self) {
+        // The take makes sure that we only ever restore the previous value once
+        if let Some(previous) = self.previous.take()
+            && let Some(value) = self.state.upgrade()
```
hashdotai, line 100: The pattern matching with `&&` could be rewritten with nested `if let` for better readability:
```suggestion
    if let Some(previous) = self.previous.take() {
        if let Some(value) = self.state.upgrade() {
            value.set(previous);
        }
    }
```

libs/@local/hashql/hir/src/lower/checking.rs
```diff
@@ -151,20 +156,25 @@ impl<'env, 'heap> TypeChecking<'env, 'heap> {
...
         // with diagnostics, in that case, if verification fails we'll have a diagnostic telling us
         // why. We just use the return type to ensure that said diagnostics have been emitted and we
         // don't silently swallow an error.
-        let compatible = self.analysis.is_subtype_of(subtype, supertype);
+        let compatible = self
+            .analysis
+            .is_subtype_of(Variance::Covariant, subtype, supertype);
 
-        if !compatible {
+        if compatible {
```
hashdotai, line 167: Nice improvement with cleaning diagnostics when verification passes. This should help prevent spurious error messages and improve the clarity of the diagnostics.

libs/@local/hashql/core/src/type/environment/analysis.rs
```diff
@@ -463,11 +481,18 @@ impl<'env, 'heap> AnalysisEnvironment<'env, 'heap> {
...
     }
 
-    pub fn is_subtype_of(&mut self, subtype: TypeId, supertype: TypeId) -> bool {
-        let (subtype, supertype) = match self.variance {
-            Variance::Covariant => (subtype, supertype),
-            Variance::Contravariant => (supertype, subtype),
-            Variance::Invariant => return self.is_equivalent(subtype, supertype),
+    pub fn is_subtype_of(
+        &mut self,
+        variance: Variance,
```
hashdotai, line 486: This signature change to include an explicit `variance` parameter is a key part of the fix. Now the variance is explicitly passed rather than implicitly applied through a context state.

libs/@local/hashql/compiletest/src/executor/trial.rs
indietyp: The next PR will remove this already

PR #7510, opened 2025-06-29
H-4910: HashQL: Make `TypeIdSet` output deterministic

indietyp:

The previous iteration of the `TypeIdSet` (while more space efficient) made use of list that is partitioned by `TypeId` and `Type` (materialized vs. not) which is then further sorted and deduped by their respective either `TypeKind` pointer or the `TypeId`.

The nice thing with this approach is that it is more space efficient, but has the problem that the ordering is dependent on the memory location of the type. For working with the types this isn't much of a problem. Rather this has been helpful to some degree in trying to find problems. The problem is that this makes formatting inconsistent in the case that we allocate a new chunk in the `Heap`, the reason is that we always bump down, meaning that a type that has been allocated in a new chunk, suddenly is at the front, making the location of types dependent on where exactly the system memory allocator decides to put things.

To circumvent this we simplify the `TypeIdSet` into two arrays:

* materialized
* provisioned

`provisioned` is for types that we don't have any information for (e.g. they're still in the process of simplification), while `materialized` are types that are already present and that can be compared by their interned kind.

(We compare by the reference of the `TypeKind`, as that makes it more "unique", as a `Type` is just a `TypeKind + SpanId`, so we're able to compare outside of the `Span` provided, allowing us to dedupe types that have the same information encoded but appear in different places)

The `TypeIdSet` is never intended to grow very large. At most \~8 elements, and most of the time \~1-2.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/type/collection/type_id_set.rs
```diff
@@ -105,55 +119,28 @@ impl<'env, 'heap, const CAPACITY: usize> TypeIdSet<'env, 'heap, CAPACITY> {
...
-        // What we now need to do is shift the materialized items to the excess, and then cut the
-        // excess items
-        self.items.copy_within(
-            len_provisioned..(len_provisioned + len_materialized - excess_materialized),
-            len_provisioned - excess_provisioned,
-        );
+        self.materialized
+            .sort_unstable_by(|lhs, rhs| ptr_cmp(lhs.kind, rhs.kind));
+        self.materialized
+            .dedup_by(|lhs, rhs| ptr_eq(lhs.kind, rhs.kind));
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer That's a good suggestion. Using `drain(..)` would indeed make the consumption more explicit and avoid moving the field. It's a more idiomatic approach in Rust when we want to take all elements from a collection while maintaining ownership of the collection itself. This would be particularly useful if we needed to reuse `self.materialized` after this operation.

libs/@local/hashql/core/src/type/collection/type_id_set.rs
```diff
@@ -105,55 +119,28 @@ impl<'env, 'heap, const CAPACITY: usize> TypeIdSet<'env, 'heap, CAPACITY> {
...
 
-        self.items.truncate(
-            len_provisioned + len_materialized - excess_provisioned - excess_materialized,
-        );
+        // We don't need to dedupe the provisioned upfront, because we can just do that once we've
+        // merged with the materialized types.
+
+        self.provisioned
+            .extend(self.materialized.into_iter().map(|r#type| r#type.id));
```
hashdotai, line 131: Consider using `drain(..)` instead of `into_iter()` here to make the consumption more explicit and avoid moving the field. It's more idiomatic when you want to take all elements from a collection while retaining ownership of the collection itself.

```suggestion
        self.materialized.drain(..)
```

libs/@local/hashql/core/src/type/kind/tests/intersection.rs
```diff
@@ -1595,7 +1595,7 @@ fn projection_propagate_pending() {
 
     let mut lattice = LatticeEnvironment::new(&env);
     let projection = lattice.projection(intersection, Ident::synthetic(heap.intern_symbol("foo")));
-    assert_eq!(lattice.diagnostics.len(), 0);
+    assert_eq!(lattice.diagnostics.len(), 1);
```
hashdotai, line 1598: The test `projection_propagate_pending` now expects a diagnostic when previously it expected none. Is this a deliberate change in behavior? If so, it would be good to add a comment explaining why this behavior changed.

libs/@local/hashql/core/src/type/kind/tests/union.rs
```diff
@@ -1683,7 +1683,7 @@ fn projection_propagate_pending() {
 
     let mut lattice = LatticeEnvironment::new(&env);
     let projection = lattice.projection(union, Ident::synthetic(heap.intern_symbol("foo")));
-    assert_eq!(lattice.diagnostics.len(), 0);
+    assert_eq!(lattice.diagnostics.len(), 1);
```
hashdotai, line 1686: Similar to the intersection.rs test, this test now expects a diagnostic when previously it expected none. Please ensure this is an intentional change and if so, add a comment explaining the change in behavior.

libs/@local/hashql/core/src/type/collection/type_id_set.rs
indietyp: I want to move the field tho.

libs/@local/hashql/core/src/type/kind/tests/intersection.rs
indietyp: Difference in order. This is expected.

libs/@local/hashql/core/src/type/kind/tests/union.rs
indietyp: Difference in order, this is expected.

TimDiekmann (approved):

PR #7526, opened 2025-06-29
H-4904: Enable `indexing-slicing` clippy lint

TimDiekmann:

Typically, index operation may panic. Mainly, there are three scenarios, where index operations are used:
- It's not checked, and the panic is simply propagated
- It's checked in the very function and should never happen
- It's checked outside of the function but in the same module

We mainly want to make sure the first thing never happens. If this is desired behavior (which may happen), an `.get(idx).unwrap_or_else(|| ...)` is desired with the body either
- `panic!()` if this is a bug in the driving library, or
- `unreachable!()` if this is a bug in the driven libirary

Generally, this PR implements this rule of thumb:
- If there is an index operation where the index comes from the outside, it's propagated as `panic` -- the caller has a bug (this only happens once)
- If there is an index operation where the index comes from the very module, it's propagated as `unreachable` -- it should never happen, but if it does, it's a bug on the library
- If the index is checked in the same function, the lint can be ignored.


libs/@blockprotocol/type-system/rust/src/ontology/json_schema/constraint/any_of.rs
```diff
@@ -20,9 +20,9 @@ pub struct AnyOfConstraints {
 
 impl From<AnyOfConstraints> for ValueConstraints {
     fn from(mut constraints: AnyOfConstraints) -> Self {
-        if constraints.any_of.len() == 1
-            && constraints.any_of[0].description.is_none()
-            && constraints.any_of[0].label.is_empty()
+        if let [any_of] = constraints.any_of.as_slice()
```
indietyp, line 23: I love this change!

libs/@local/harpc/codec/src/error.rs
```diff
@@ -56,7 +56,12 @@ impl NetworkError {
         );
         let length = length as u32;
 
-        buffer[..4].copy_from_slice(&length.to_be_bytes());
+        buffer
```
indietyp, line 59: Out of curiosity I just checked in godbolt (https://rust.godbolt.org/z/Mbc835sGz) if there would be any difference between both. The answer is: virtually no, the only difference is how they panic.

libs/@local/harpc/codec/src/error.rs
```diff
@@ -104,19 +109,19 @@ impl NetworkError {
...
+        clippy::indexing_slicing,
+        reason = "We error if the buffer is too small"
+    )]
     pub fn try_from_parts(code: ErrorCode, bytes: Bytes) -> Result<Self, Bytes> {
         let slice = bytes.as_ref();
         if slice.len() < 4 {
             return Err(bytes);
         }
 
-        // assert only exists to elide bounds checks and satisfy clippy
```
indietyp, line 117: verified in godbolt that this assert is no longer required: https://rust.godbolt.org/z/7fjYGvs8x

libs/error-stack/src/report.rs
```diff
@@ -403,9 +403,12 @@ impl<C> Report<C> {
     /// [`frames()`]: Self::frames
     #[must_use]
     pub fn current_frame(&self) -> &Frame {
-        // this never fails, because one cannot push an additional frame without making it a
-        // `Report<[C]>`
-        &self.frames[0]
+        self.frames.first().unwrap_or_else(|| {
```
indietyp, line 406: Maybe we should add a link to report this? like we did in other places of error-stack?

TimDiekmann: Done in [`e90876f` (#7526)](https://github.com/hashintel/hash/pull/7526/commits/e90876f1b109400545e6ebe67bb89816fb5202b9). Also updated the other message in [`28d09d8` (#7526)](https://github.com/hashintel/hash/pull/7526/commits/28d09d8beb185f9797263434d66088e5d91153dd) to align the messages.

indietyp (approved):

PR #7536, opened 2025-06-30
H-4599: Support `BaseUrl` filter of types in entity resource constraints

TimDiekmann:

This PR adds support for filtering entities by their base type URL (`BaseUrl`) in addition to the existing versioned type URL filtering. This enables more flexible entity resource constraints in the authorization system by allowing policies to match entities based on their type family rather than specific versions.

## 🔍 What does this change?

- **Authorization Schema**: Added `entity_base_types` field to the Entity schema in Cedar policies
- **Expression Tree**: Added `IsOfBaseType` variant to `PolicyExpressionTree` enum for parsing base URL filters
- **Resource Constraints**: 
  - Added `IsOfBaseType` variant to `EntityResourceFilter` enum
  - Added `entity_base_types` field to `EntityResource` struct alongside existing `entity_types`
- **Cedar Expression Generation**: Implemented Cedar expression generation for base type filtering using `entity_base_types` attribute
- **Database Integration**: Updated PostgreSQL store to populate `entity_base_types` by extracting base URLs from existing versioned entity types
- **Policy Simplification**: Refactored seed policies to use `IsOfBaseType` filters instead of creating multiple version-specific filters, significantly reducing policy complexity
- **Test Updates**: Updated all test cases to include `entity_base_types` data

The key benefit is that policies can now match entities like "all users" regardless of the specific user entity type version (v1, v2, v6, etc.) by filtering on the base URL `https://hash.ai/@h/types/entity-type/user/`.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

None

## 🛡 What tests cover this?

- Updated existing authorization policy tests to include `entity_base_types` data
- Tests cover both individual base type filtering and combined filtering scenarios
- Integration tests verify database population of base types from versioned URLs

## ❓ How to test this?

1. Checkout the branch
2. Run the authorization tests: `cargo nextest run --package hash-graph-authorization`
3. Verify policies can filter entities by base type instead of requiring specific versions
4. Test with different entity types that have multiple versions

## 📹 Demo

Not applicable - this is an internal authorization system enhancement.

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
@@ -1854,6 +1854,11 @@ where
     }
 
     #[tracing::instrument(level = "debug", skip(self, entity_edition_ids))]
+    #[expect(
+        clippy::too_many_lines,
+        reason = "A large part of this function is for a SQL query that is complex but necessary \
+                  for the context building."
+    )]
     async fn build_entity_context(
         &self,
         entity_edition_ids: &[EntityEditionId],
--- a/libs/@local/graph/authorization/src/policies/cedar/expression_tree.rs
+++ b/libs/@local/graph/authorization/src/policies/cedar/expression_tree.rs
@@ -144,6 +144,7 @@ impl PolicyExpressionTree {
             | Self::BaseUrl(_)
             | Self::OntologyTypeVersion(_)
             | Self::IsOfType(_)
+            | Self::IsOfBaseType(_)
             | Self::CreatedByPrincipal) => vec![expression],
         };
 
@@ -156,6 +157,7 @@ impl PolicyExpressionTree {
             | Self::BaseUrl(_)
             | Self::OntologyTypeVersion(_)
             | Self::IsOfType(_)
+            | Self::IsOfBaseType(_)
             | Self::CreatedByPrincipal) => all.push(expression),
         }
 
@@ -183,6 +185,7 @@ impl PolicyExpressionTree {
             | Self::BaseUrl(_)
             | Self::OntologyTypeVersion(_)
             | Self::IsOfType(_)
+            | Self::IsOfBaseType(_)
             | Self::CreatedByPrincipal) => vec![expression],
         };
 
@@ -195,6 +198,7 @@ impl PolicyExpressionTree {
             | Self::BaseUrl(_)
             | Self::OntologyTypeVersion(_)
             | Self::IsOfType(_)
+            | Self::IsOfBaseType(_)
             | Self::CreatedByPrincipal) => any.push(expression),
         }
 
--- a/libs/@local/graph/authorization/src/policies/resource/entity_type.rs
+++ b/libs/@local/graph/authorization/src/policies/resource/entity_type.rs
@@ -146,6 +146,7 @@ impl TryFrom<PolicyExpressionTree> for EntityTypeResourceFilter {
             condition @ (PolicyExpressionTree::Is(_)
             | PolicyExpressionTree::In(_)
             | PolicyExpressionTree::IsOfType(_)
+            | PolicyExpressionTree::IsOfBaseType(_)
             | PolicyExpressionTree::CreatedByPrincipal) => {
                 Err(Report::new(InvalidEntityTypeResourceFilter(condition)))
             }


libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1907,24 +1912,33 @@ where
...
-                created_by: ActorId::new(
-                    row.get::<_, ActorEntityUuid>(3),
-                    match row.get(4) {
-                        PrincipalType::User => ActorType::User,
-                        PrincipalType::Machine => ActorType::Machine,
-                        PrincipalType::Ai => ActorType::Ai,
-                        principal_type => unreachable!(
-                            "Unexpected actor type `{principal_type:?}` in entity context"
-                        ),
+            .map_ok(|row| {
```
hashdotai, line 1933: The SQL query here fetches base URLs but assigns them to a column named `base_urls`. However, when accessing the row data in the `map_ok` function, you're not actually using this column. Instead, you're deriving the base URLs from the entity types. While this still works functionally, it's inefficient to fetch data you don't use.

Either utilize the `base_urls` column from the query or remove it from the query to avoid unnecessary data retrieval.

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -292,28 +292,33 @@ impl<'p> Filter<'p, Entity> {
         }
     }
 
+    #[must_use]
```
hashdotai, line 295: For consistency with the rest of the codebase, this line should be `REQUIRED` in the function docstring. You've made this change to designate the method as an essential implementation detail, but the actual docstring annotation is missing.

libs/@local/graph/store/src/filter/mod.rs
```diff
@@ -292,28 +292,33 @@ impl<'p> Filter<'p, Entity> {
...
-                }),
-            ),
+            Self::for_entity_by_base_type_id(&entity_type_id.base_url),
             Filter::Equal(
                 Some(FilterExpression::Path {
                     path: EntityQueryPath::EntityTypeEdge {
                         edge_kind: SharedEdgeKind::IsOfType,
                         path: EntityTypeQueryPath::Version,
-                        inheritance_depth: Some(0),
+                        inheritance_depth: None,
```
hashdotai, line 321: It looks like you're removing the `inheritance_depth: Some(0)` here and replacing it with `inheritance_depth: None`. While this might be the correct approach for the base URL filter, please ensure this change won't affect existing behavior for version filters. If this is intentional, consider adding a comment explaining why `None` is now appropriate.

libs/@local/graph/store/src/filter/mod.rs
TimDiekmann: What are you talking about?

libs/@local/graph/store/src/filter/mod.rs
TimDiekmann: `Some(0)` was just wrong (and was only really used in tests). And I won't add comments on something which existed at some point.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: Good spot, fixed in [`dedd53b` (#7536)](https://github.com/hashintel/hash/pull/7536/commits/dedd53bba27d6a1cd1d515a4e7759ecfd291ea90)

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1837,6 +1837,7 @@
     }
 
     #[tracing::instrument(level = "debug", skip(self, entity_edition_ids))]
+    #[expect(clippy::too_many_lines)]
```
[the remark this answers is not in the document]
TimDiekmann:
```suggestion
```

libs/@local/graph/authorization/src/policies/resource/entity.rs
```diff
@@ -22,7 +22,8 @@ use crate::policies::cedar::{
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct EntityResource<'a> {
     pub id: EntityId,
-    pub entity_type: Cow<'a, [VersionedUrl]>,
+    pub entity_types: Cow<'a, [VersionedUrl]>,
+    pub entity_base_types: Cow<'a, [BaseUrl]>,
```
indietyp, line 26: just for clarification - the base types are the types without any version, correct? In that case, what happens if an entity type happens to inherit from 2 different versions of the same entity type? Will the `base_types` be deduplicated? Or will it always be `assert_eq!(entity_types.len(), entity_base_types.len())`.

indietyp (approved):
minor question, otherwise LGTM

libs/@local/graph/authorization/src/policies/resource/entity.rs
TimDiekmann: I'm not sure if we would even allow this scenario, but conceptually, this is converted to a `Set` in Cedar, so it would be deduplicated. The general goal is to have "Is this entity some kind of `User`", we mainly use a `contains` operation. Cedar also does not support closures, this is why we cannot use something like `.find(|type| type == base_url)` but need a dedicated list.

PR #7548, opened 2025-07-01
H-4892: Introduce `UpdateEntity` and `ArchiveEntity` permissions

TimDiekmann:

This PR introduces new authorization actions `UpdateEntity` and `ArchiveEntity` to provide fine-grained permissions for entity operations, replacing the SpiceDB `Update` permission with more specific entity-focused policies.

This change allows for more granular control over who can update and archive specific entities, improving security and enabling better permission management across the system.

## 🔍 What does this change?

- **Authorization Schema Updates**: Added `updateEntity` and `archiveEntity` actions to the Cedar schema alongside existing `viewEntity` action
- **Action Hierarchy**: Implemented proper action hierarchy where `UpdateEntity` inherits from `Update` and `ArchiveEntity` inherits from `Archive`
- **Permission Enforcement**: Updated entity update logic to use the new `UpdateEntity` and `ArchiveEntity` permissions instead of generic `Update` permission
- **Policy Updates**: Modified system policies to grant appropriate permissions:
  - System actors can now update and archive service features
  - Google bot can update Google account entities
  - Linear bot can update and archive Linear-created entities
  - User secrets now have proper update/archive permissions for managing bots
- **Entity-Specific Policies**: Added granular policies for specific entity types:
  - Comment text entities get creator update/archive permissions
  - Organization membership entities get member update/archive permissions
- **Removed Redundant Policies**: Cleaned up redundant view-only policies that are now covered by the system-wide policies

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
+++ b/apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
@@ -31,6 +31,7 @@ import { ApolloError } from "apollo-server-errors";
 import dedent from "dedent";
 
 import type { EmailTransporter } from "../../../../email/transporters";
+import { isActorGroupAdministrator } from "../../../../graph/account-permission-management";
 import {
   createEntity,
   getEntitySubgraphResponse,
@@ -218,14 +219,10 @@ export const inviteUserToOrgResolver: ResolverFn<
     );
   }
 
-  const isOrgAdmin = await context.graphApi
-    .hasActorGroupRole(
-      authentication.actorId,
-      org.webId,
-      "administrator",
-      authentication.actorId,
-    )
-    .then(({ data }) => data);
+  const isOrgAdmin = await isActorGroupAdministrator(context, authentication, {
+    actorId: authentication.actorId,
+    actorGroupId: org.webId,
+  });
 
   if (!isOrgAdmin) {
     throw new ApolloError(


libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -295,24 +606,32 @@ fn instance_admins_view_entity_policy(
...
+    role: &TeamRole,
+) -> impl Iterator<Item = PolicyCreationParams> {
+    iter::once(PolicyCreationParams {
+        name: Some("hash-admins-update-entity".to_owned()),
+        effect: Effect::Permit,
+        principal: Some(PrincipalConstraint::Role {
+            role: RoleId::Team(role.id),
+            actor_type: None,
+        }),
+        actions: vec![ActionName::ViewEntity],
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're absolutely right, good catch! The function name indicates it's for the "update" permission, but it's only granting the "view" permission. This is definitely a mistake that should be fixed to match the intended functionality.

Your suggestion to replace it with `ActionName::UpdateEntity, ActionName::ArchiveEntity` makes sense, as it would align with the function name and the pattern established in other similar policies. This would properly grant instance admins the ability to update and archive entities, which appears to be the intent based on the PR's description of introducing more granular permissions.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -27,14 +30,103 @@ use type_system::{
...
+});
+
+static GOOGLE_ACCOUNT: LazyLock<EntityTypeConfig> =
+    LazyLock::new(|| EntityTypeConfig::new("https://hash.ai/@google/types/entity-type/account/"));
+static LINEAR_INTEGRATION: LazyLock<EntityTypeConfig> = LazyLock::new(|| {
+    EntityTypeConfig::new("https://hash.ai/@h/types/entity-type/linear-integration/")
+});
+
+use super::PostgresStore;
+
```
hashdotai, line 129: Good change moving from hardcoded entity type URLs to the new `EntityTypeConfig` approach with `LazyLock`. This makes the code more maintainable and reduces duplication.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1643,11 +1631,80 @@ where
...
+                    return Err(Report::new(UpdateError)
+                        .attach(StatusCode::PermissionDenied)
+                        .attach_printable(format!(
+                            "The actor does not have permission to {} the entity",
+                            if archive { "archive" } else { "publish" },
+                        )));
+                }
+            }
+        }
+
```
hashdotai, line 1709: The action permission check for entity updates has been properly moved from a general `Update` permission to more specific `UpdateEntity` and `ArchiveEntity` permissions. This is a good architectural improvement that allows for more granular control.

apps/hash-api/src/graph/knowledge/system-types/comment.ts
```diff
@@ -196,6 +199,18 @@ export const createComment: ImpureGraphFunction<
...
+        name: `comment-text-creator-update-entity-${textEntityEntityUuid}`,
+        principal: {
+          type: "actor",
+          actorType: "user",
+          id: author.accountId,
+        },
+        effect: "permit",
+        actions: ["updateEntity", "archiveEntity"],
+      },
+    ],
```
hashdotai, line 213: This change implements granular permission control for comments by adding specific update/archive permissions for the comment text entity creator. This aligns well with the PR's goal of providing more fine-grained control.

apps/hash-api/src/graph/knowledge/system-types/org-membership.ts
```diff
@@ -85,6 +90,18 @@ export const createOrgMembershipLinkEntity: ImpureGraphFunction<
...
+        name: `org-membership-editor-${linkEntityEntityUuid}`,
+        principal: {
+          type: "actor",
+          actorType: "user",
+          id: userActorId,
+        },
+        effect: "permit",
+        actions: ["updateEntity", "archiveEntity"],
+      },
+    ],
```
hashdotai, line 103: Similarly, this implementation provides granular control for organization membership entities. Good addition of specific permissions for the member who should be able to update and archive their membership.

apps/hash-api/src/integrations/linear/oauth.ts
```diff
@@ -262,28 +262,6 @@ export const oAuthLinearCallback: RequestHandler<
...
-            name: `linear-integration-user-view-entity-${entityUuid}`,
-            principal: {
-              type: "actor",
-              actorType: "user",
-              id: userAccountId,
-            },
-            effect: "permit",
-            actions: ["viewEntity"],
-          },
-        ],
```
hashdotai, line 286: Removed redundant policies for linear-bot viewing entities that are now covered by the system-wide policies. This is a good cleanup that reduces policy duplication.

libs/@local/graph/authorization/schemas/policies.cedarschema
```diff
@@ -47,7 +47,7 @@ namespace HASH {
     resource: [Web],
   };
 
-  action viewEntity in [view] appliesTo {
+  action viewEntity, updateEntity, archiveEntity in [view] appliesTo {
```
hashdotai: When adding the new actions to the Cedar schema, there's a documentation opportunity here. Consider adding a comment or documentation that describes the relationship between these actions and what they are intended for, similar to how other permissions are documented.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1643,11 +1631,80 @@
...
+                Authorized::Never => {
+                    return Err(Report::new(UpdateError)
+                        .attach(StatusCode::PermissionDenied)
+                        .attach_printable(format!(
+                            "The actor does not have permission to {} the entity",
+                            if archive { "archive" } else { "publish" },
+                        )));
+                }
+            }
+        }
```
hashdotai, line 1708: The code added here to check `ArchiveEntity` permission is good, but there seems to be duplication between these two permission checks with very similar error message structures. Consider refactoring this into a helper function that can check a permission with the appropriate error message.

libs/@local/graph/authorization/types/index.snap.d.ts
```diff
@@ -13,7 +13,7 @@ export interface Policy {
 }
 import type { Brand } from "@local/advanced-types/brand";
 export type PolicyId = Brand<string, "PolicyId">;
-export type ActionName = "createEntity" | "createWeb" | "viewEntity" | "instantiate";
+export type ActionName = "createEntity" | "createWeb" | "viewEntity" | "updateEntity" | "archiveEntity" | "instantiate";
```
hashdotai, line 16: You've added `updateEntity` and `archiveEntity` to the exported action types, which is good. However, you might want to update the documentation for these new types to clarify their usage.

libs/@local/graph/authorization/schemas/policies.cedarschema
TimDiekmann: Well, `viewEntity` is to view entities. `updateEntity` is to ... update entities. And I imagine, `archiveEntity` should be somewhat clear as well.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: They are similar, but not identical. There are plans to improve this, yes, but that's more work than simply move two things into a function.

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
```diff
@@ -135,6 +139,12 @@ export const userBeforeEntityUpdateHookCallback: BeforeUpdateEntityHookCallback
 
       // Now that the user has completed signup, we can transfer the ownership of the web
       // allowing them to create entities and types.
+      await addActorGroupAdministrator(
+        context,
+        { actorId: systemAccountId },
+        { actorId: user.accountId, actorGroupId: user.accountId },
+      );
+
       await modifyWebAuthorizationRelationships(
```
CiaranMn, line 148: Presumably this is for deletion once we've fully switched over?

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
```diff
@@ -223,14 +223,14 @@ export const createUserSecret = async <
       relationships: botEditorUserViewerOnly,
       policies: [
         {
-          name: `user-secret-view-entity-${userSecretEntityUuid}`,
+          name: `user-secret-entity-${userSecretEntityUuid}`,
```
CiaranMn: Is it an issue that this is a duplicate of the above? I've tried to keep policy name uniques but don't know if it matters.

TimDiekmann: The name + principal combination needs to be unique. So it's fine to use the same name as long as the principal is different.

CiaranMn: The principal is the same in this case. Do you mean the resource?

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
TimDiekmann: Yes. It's not needed anymore, but kept it to keep the database consistent if we need to switch back.

apps/hash-api/src/graph/knowledge/system-types/user-secret.ts
TimDiekmann: Oh, the name is not correct

```suggestion
          name: `user-secret-entity-${usesUserSecretEntityUuid}`,
```

TimDiekmann: So regarding your question: yes, this would be an issue; it would fail in this case. Very good spot, thanks!

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -27,14 +30,103 @@ use type_system::{
...
+    EntityTypeConfig::new("https://hash.ai/@h/types/entity-type/occurred-in-text/")
+});
+
+static GOOGLE_ACCOUNT: LazyLock<EntityTypeConfig> =
+    LazyLock::new(|| EntityTypeConfig::new("https://hash.ai/@google/types/entity-type/account/"));
+static LINEAR_INTEGRATION: LazyLock<EntityTypeConfig> = LazyLock::new(|| {
+    EntityTypeConfig::new("https://hash.ai/@h/types/entity-type/linear-integration/")
+});
+
+use super::PostgresStore;
```
indietyp, line 128: odd place for a use statement

indietyp (dismissed):
lgtm minor comment

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
TimDiekmann: True. Will fix in a follow up 🙂

PR #7556, opened 2025-07-02
H-591, H-1385: Organization invites – invite via email, accept/decline invites

CiaranMn:

The PR makes several improvements to how organization invites work:
- Instead of automatically adding a user to an organization, org admins now instead issue invitations, which the user can accept or decline
- Invitations can be issued to an email address, as well as a username (the org admin cannot learn which username is associated with which email by issuing an invite, but obviously will figure it out if the user accepts it)
- Invitations can be emailed to users who aren't yet signed up, in which case there's a new signup flow (see demo video)
- Show pending invitations on org members page (if user is admin)
- Add org invitations as a new category of things in a users' Inbox, alongside actions (drafts for approval) and notifications (e.g. for comments)
- H-1385: add an 'invite' button when viewing an org's profile page (if user is an admin)

It also adds an SMTP mail transport option (in addition to the existing AWS one, and the dummy 'log to console' one for testing).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- None yet.

## ❓ How to test this?

1. [Locally] as `alice@example.com` try issuing org invitations to an email address, the invite link will be logged to the Node API's console, go through signup flow. Note you'll need to log out of Alice before clicking the link (or open in incognito).

## 📹 Demo

https://github.com/user-attachments/assets/60d091ed-b592-4e24-a12f-b8b5e6f4e10c


apps/hash-api/src/email/create-email-transporter.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  if (transporterType === "smtp") {
+    return new SmtpEmailTransporter({
+      from: `${getRequiredEnv(
+        "SYSTEM_EMAIL_SENDER_NAME",
+      )} <${getRequiredEnv("SYSTEM_EMAIL_ADDRESS")}>`,
+      subjectPrefix,
+    });
+  }
```
hashdotai, line 54: The fallback email transporter is missing the `Promise<void>` return type in the `sendMail` method. Consider adding this to match the `EmailTransporter` interface.

apps/hash-api/src/email/transporters/smtp-email-transporter.ts
```diff
@@ -0,0 +1,58 @@
...
+      ...(username && password
+        ? {
+            auth: {
+              type: "login",
+              user: username,
+              pass: password,
+            },
+          }
+        : {}),
+      secure: port === 465,
```
hashdotai, line 40: The `secure: port === 465` setting assumes port 465 will always use TLS. For better flexibility, consider adding a configuration option to override this, as some setups might use TLS on other ports.

apps/hash-frontend/src/pages/signup.page/accept-org-invitation.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <Box>
+      <Typography sx={{ fontSize: 32, fontWeight: 700, mb: 4 }}>
+        {invitedBy.displayName} has invited you to join {org.displayName}
+      </Typography>
+      <Typography sx={{ fontSize: 24, mb: 6 }}>
+        This invite was sent to <br />
+        <Box component="span" sx={{ fontWeight: 700 }}>
+          {"email" in invitation ? invitation.email : invitation.shortname}
```
hashdotai, line 22: For better security and to prevent information leakage, consider masking part of the email address when displaying it to the user, especially if this might be displayed to someone who doesn't yet have an account.

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
```diff
@@ -0,0 +1,210 @@
...
+      ] === user.shortname;
+  } else {
+    throw new ApolloError(
+      `Invalid invitation type ${invitation.metadata.entityTypeIds.join(", ")}`,
+      "INVALID_INVITATION_TYPE",
+    );
+  }
+
+  if (!isForUser) {
+    return {
```
hashdotai, line 83: Consider adding more detailed error messages for the user. Currently, it returns `notForUser: true` without explaining why the invitation might not be valid for them.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
```diff
@@ -0,0 +1,407 @@
...
+          }
+        : {
+            type: "shortname",
+            shortname: userShortname!,
+          },
+    ),
+  }).then((entities) => entities[0]);
+
+  if (existingInvitation) {
+    throw new ApolloError(
```
hashdotai, line 297: There should be a check for whether the invitation has expired before showing it as a pending invitation. Consider adding a filter to verify `invitation.properties["https://hash.ai/@h/types/property-type/expired-at/"] > new Date()`.

apps/hash-api/src/graphql/resolvers/knowledge/org/shared.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    if (
+      !linkEntity.metadata.entityTypeIds.includes(
+        systemLinkEntityTypes.hasIssuedInvitation.linkEntityTypeId,
+      )
+    ) {
+      throw new Error(
+        `Pending invitation with entityId ${root.entityId} has types ${linkEntity.metadata.entityTypeIds.join(
+          ", ",
```
hashdotai, line 79: For improved readability and maintainability, consider extracting the property path `"https://hash.ai/@h/types/property-type/expired-at/"` into a constant, either locally or imported from a shared location.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
```diff
@@ -0,0 +1,407 @@
...
+    relationships: invitationAuthorizationRelationships,
+    webId: orgWebId,
+  });
+
+  await sendOrgEmailInvitationToEmailAddress({
+    org,
+    invitationId: invitation.metadata.recordId.entityId,
+    emailAddress: existingUserToInvite
+      ? existingUserToInvite.emails[0]!
+      : userEmail!,
```
hashdotai, line 436: When sending an email, there's no validation of the email address format. Consider adding validation to ensure the email address is valid before attempting to send an invite.

apps/hash-api/src/email/create-email-transporter.ts
CiaranMn: The return type of the function is declared as `EmailTransporter`, and therefore it must necessarily match it.

apps/hash-api/src/email/transporters/smtp-email-transporter.ts
CiaranMn: I'll do this in follow-up when looking at email verification (H-2421)

apps/hash-frontend/src/pages/signup.page/accept-org-invitation.tsx
CiaranMn: They have arrived here via a link sent to their email address.

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
CiaranMn: This is sufficient, people aren't going to end up here unless they're making weird requests directly to the API.

apps/hash-api/src/graphql/resolvers/knowledge/org/shared.ts
CiaranMn: URL keys are a common feature of the app and they are type safe.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
CiaranMn: Good spot! Added in https://github.com/hashintel/hash/pull/7556/commits/f7ead8c7bddf3f9b1d18956762f3d190c002a1a0

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
CiaranMn: There is validation on the client. I appreciate users could bypass this to send an org invite but the impact is minimal (invite will not be sent)

apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -141,13 +146,62 @@ export const getUserById: ImpureGraphFunction<
...
+            systemEntityTypes.user.entityTypeId,
+            { ignoreParents: true },
+          ),
+          {
+            equal: [
+              {
+                path: [
+                  "properties",
+                  systemPropertyTypes.email.propertyTypeBaseUrl,
+                  0,
```
TimDiekmann, line 175: We may want to add a TODO here that we only search for the primary email (I know that we currently don’t allow updating these).

apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -560,3 +613,92 @@ export const isUserMemberOfOrg: ImpureGraphFunction<
...
+export const getUserPendingInvitations: ImpureGraphFunction<
+  { user: User },
+  Promise<PendingOrgInvitation[]>
+> = async (context, _authentication, { user }) => {
+  /**
+   * The system account is used to manage invitations on behalf of the user,
+   * because the user does not have permissions on them,
+   * to avoid accidentally leaking their identity before they have accepted the invitation.
+   *
+   * Otherwise an org admin could issue an invitation to an email address and check which user was given permission on the invitation.
```
TimDiekmann, line 629: Do we have a way to see which actors can access which entity?

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
```diff
@@ -0,0 +1,442 @@
...
+
+  if (isAlreadyAMember) {
+    throw new ApolloError(
+      "User is already a member of this organization",
+      "BAD_REQUEST",
+    );
+  }
+
+  const isOrgAdmin = await context.graphApi
+    .hasActorGroupRole(
```
TimDiekmann, line 222: There is a corresponding function in Node with the same name `isActorGroupAdministrator` which is more typesafe:
https://github.com/hashintel/hash/blob/47f25e58ff30ed30a8fdf2737e96cd1bb35e60ee/apps/hash-api/src/graph/account-permission-management.ts#L81-L92

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
```diff
@@ -0,0 +1,442 @@
...
+      authentication.actorId,
+    )
+    .then(({ data }) => data);
+
+  if (!isOrgAdmin) {
+    throw new ApolloError(
+      "You must be an administrator to invite users to this organization",
+      "UNAUTHORIZED",
+    );
+  }
```
TimDiekmann, line 235: Shouldn't this be covered by permission checks? i.e. `CreateEntity` is only granted for admins for this entity type? We would need to change how the `CreateEntity` permission check works slightly, but conceptually, I think this should be covered in the graph.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
```diff
@@ -0,0 +1,442 @@
...
+        relation: "administrator",
+        subject: {
+          kind: "account",
+          /**
+           * We use the system account to manage invitations (viewing and archiving) on the invited user's behalf.
+           * We cannot add the invited user directly, because:
+           *
+           * 1. They may not have an account yet (if invited via email with no account attached)
+           * 2. If they DO have an account and are invited by email, giving them permission on the entity would reveal
+           *    to the inviter which user account was associated with the email address (via inspection of the entity's permissions).
```
TimDiekmann, line 315: I cannot follow this logic, how would this expose it? Unless there is no link pointing to the user, this should not be an issue.

TimDiekmann (commented):
Generally looks good, thank you!

I have a few comments. Also, if the invitation only links to the org and not to the user, I don't think that the system user needs to manage the invitation link because AFAIK there is no way to get all actors who has view access to a specific entity. If I'm not mistaken, we can probably use the actual actor, which means it's also easier to track who actually created the invitation.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
TimDiekmann: With https://github.com/hashintel/hash/pull/7560 that's now possible

CiaranMn: It should be, but it would still be better to have a nice error message.

But wouldn't the permission check be something to do with 'can add to account group'?

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
CiaranMn: We previously (currently) have a query to expose the permissions on an entity (who can see it) and would assume we want this in future. We talked about this and agreed to leave it as-is for the moment.

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
CiaranMn: This function is in an unmerged PR

apps/hash-api/src/graph/knowledge/system-types/user.ts
CiaranMn: We currently do, and will want this again in the new system.

apps/hash-api/src/graph/knowledge/system-types/user.ts
CiaranMn: Added in https://github.com/hashintel/hash/pull/7556/commits/eb74829067ad751054efd3a89af5c06e831068d4

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
TimDiekmann: We could do both:
- Restrict the entity creation for `Invitation` - if a user cannot create an invitation, he cannot add someone as a member
- Add a whole new policy `AddAccountGroupMember` or `ChangeAccountGroupRole` or something. 

We probably want both longterm. The first one to generally restrict who can create entities of a certain type in the first place in a web (although, a large part of this could be handled by instantiation permission), and the second one for a probably more fine-grained system to change roles. Admins, e.g., should probably be able to change any role in a web/team, but there might be an HR team who may assign roles as well, but they must not allow to change the admin role.

initially, we probably don't need an extra permission if we can just use the `CreateEntity` permission (which is currently granted for all roles for the role's web)

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
TimDiekmann: Oh, you're right, sorry. Will update that when the PR is merged 🙂

apps/hash-api/src/graph/knowledge/system-types/user.ts
TimDiekmann: Great, thanks!

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
CiaranMn: Agree that we will need the `AddAccountGroupMember` to do any custom permissions, but don't need it until then.

TimDiekmann (approved):

apps/hash-api/src/graphql/resolvers/knowledge/org/invite-user-to-org.ts
TimDiekmann: [`58a6025` (#7548)](https://github.com/hashintel/hash/pull/7548/commits/58a60252d6a904f2e718cefc31cb38d501f32959)

PR #7560, opened 2025-07-03
H-4929: Change `CreateEntity` permission to use an `Entity`-resource

TimDiekmann:

This PR implements the changes described in H-4929 to transition from using `Web` resources to `Entity` resources for the `CreateEntity` permission. This change provides more granular control over entity creation by allowing filtering by entity type, actor type restrictions, and even specific entity creation permissions.

## 🔍 What does this change?

- **Authorization Schema**: Changed `createEntity` action in Cedar schema to use `Entity` resource instead of `Web` resource
- **Policy Components**: Added entity type tracking and entity resource context building capabilities
- **Entity Creation Logic**: Replaced web-based authorization with entity-based authorization, including early permission checks for both `Instantiate` and `CreateEntity` actions
- **Seed Policies**: Updated policy generation to use `EntityResourceConstraint::Web` with role-based filtering (`Administrator` vs `Member`)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

None

## 🐾 Next steps

- [H-4441: Have graph create user, machine, organization, and memberOf entities](https://linear.app/hash/issue/H-4441/have-graph-create-user-machine-organization-and-memberof-entities)

## 🛡 What tests cover this?

- Existing authorization tests cover the basic permission checking flow
- Entity creation integration tests verify the new authorization logic
- Policy evaluation tests ensure correct permission decisions

## ❓ How to test this?

1. Checkout the branch
2. Create entities with different user roles (Administrator, Member)
3. Verify that Members cannot create `memberOf` entities while Administrators can
4. Test entity type filtering and ensure appropriate permission errors are returned
5. Confirm that entity creation still works for permitted operations

## 📹 Demo

N/A - Internal authorization logic changes with no visible UI changes

### DIFF
--- a/libs/@local/graph/authorization/src/policies/components.rs
+++ b/libs/@local/graph/authorization/src/policies/components.rs
@@ -3,7 +3,10 @@ use std::collections::{HashMap, HashSet};
 
 use error_stack::{Report, ResultExt as _};
 use type_system::{
-    knowledge::entity::id::{EntityEditionId, EntityUuid},
+    knowledge::entity::{
+        EntityId,
+        id::{EntityEditionId, EntityUuid},
+    },
     ontology::VersionedUrl,
     principal::{actor::ActorId, actor_group::WebId},
 };
@@ -12,7 +15,7 @@ use super::{
     Context, ContextBuilder, Effect, Policy, PolicySet,
     action::ActionName,
     principal::actor::AuthenticatedActor,
-    resource::{EntityResourceConstraint, ResourceConstraint},
+    resource::{EntityResource, EntityResourceConstraint, ResourceConstraint},
     store::{PolicyStore, PrincipalStore, ResolvePoliciesParams, error::ContextCreationError},
 };
 
@@ -33,6 +36,7 @@ pub struct PolicyComponents {
     actor_id: Option<ActorId>,
     policies: Vec<Policy>,
     tracked_actions: HashMap<ActionName, Option<OptimizationData>>,
+    tracked_entity_types: HashSet<VersionedUrl>,
     context: Context,
 }
 
@@ -105,6 +109,11 @@ impl PolicyComponents {
 
                 EMPTY_OPTIMIZATION_DATA
             })
+    }
+
+    #[must_use]
+    pub const fn tracked_entity_types(&self) -> &HashSet<VersionedUrl> {
+        &self.tracked_entity_types
     }
 
     /// Analyzes policies and extracts optimization opportunities.
@@ -371,6 +380,25 @@ impl<'a, S> PolicyComponentsBuilder<'a, S> {
         self.add_actions(actions, optimize);
         self
     }
+
+    pub fn add_entity(
+        &mut self,
+        actor: ActorId,
+        id: EntityId,
+        entity_types: Cow<'a, [VersionedUrl]>,
+    ) {
+        self.context.add_entity(&EntityResource {
+            id,
+            entity_base_types: Cow::Owned(
+                entity_types
+                    .iter()
+                    .map(|url| url.base_url.clone())
+                    .collect(),
+            ),
+            entity_types,
+            created_by: actor,
+        });
+    }
 }
 
 impl<S> IntoFuture for PolicyComponentsBuilder<'_, S>
@@ -423,7 +451,9 @@ where
                 }
             }
 
-            if !entity_type_ids.is_empty() {
+            let tracked_entity_types = if entity_type_ids.is_empty() {
+                HashSet::new()
+            } else {
                 let entity_type_ids = entity_type_ids.iter().map(Cow::as_ref).collect::<Vec<_>>();
                 let entity_type_resources = self
                     .store
@@ -435,7 +465,11 @@ where
                 for entity_type_resource in &entity_type_resources {
                     self.context.add_entity_type(entity_type_resource);
                 }
-            }
+                entity_type_resources
+                    .into_iter()
+                    .map(|resource| resource.id.into_owned().into_url())
+                    .collect::<HashSet<_>>()
+            };
 
             let actions = self.actions.keys().copied().collect::<Vec<_>>();
             let policies = if actions.is_empty() {
@@ -457,6 +491,7 @@ where
                 actor_id,
                 policies,
                 tracked_actions: actions.iter().map(|action| (*action, None)).collect(),
+                tracked_entity_types,
                 context: self
                     .context
                     .build()
@@ -477,7 +512,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashMap;
+    use std::collections::{HashMap, HashSet};
 
     use type_system::{knowledge::entity::id::EntityUuid, principal::actor::ActorId};
     use uuid::Uuid;
@@ -521,6 +556,7 @@ mod tests {
             actor_id: None,
             policies,
             tracked_actions: HashMap::from([(ActionName::View, None)]),
+            tracked_entity_types: HashSet::new(),
             context: Context::default(),
         };
 
@@ -571,6 +607,7 @@ mod tests {
             actor_id: None,
             policies: policies_without_optimization,
             tracked_actions: HashMap::from([(ActionName::View, None)]),
+            tracked_entity_types: HashSet::new(),
             context: Context::default(),
         };
 
@@ -595,6 +632,7 @@ mod tests {
             actor_id: None,
             policies: policies_with_optimization,
             tracked_actions: HashMap::new(),
+            tracked_entity_types: HashSet::new(),
             context: Context::default(),
         };
 
@@ -627,6 +665,7 @@ mod tests {
             ))),
             policies: vec![policy],
             tracked_actions: HashMap::from([(ActionName::View, None)]),
+            tracked_entity_types: HashSet::new(),
             context: Context::default(),
         };
 
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
@@ -857,6 +857,14 @@ where
         let validator_provider = StoreProvider::new(&transaction, &policy_components);
 
         let mut validation_reports = HashMap::<usize, EntityValidationReport>::new();
+
+        debug_assert_eq!(
+            params.len(),
+            entity_ids.len(),
+            "Number of parameters ({}) and entity ids ({}) must be the same",
+            params.len(),
+            entity_ids.len(),
+        );
         for (index, (mut params, &entity_id)) in params.into_iter().zip(&entity_ids).enumerate() {
             let entity_type = ClosedMultiEntityType::from_multi_type_closed_schema(
                 stream::iter(&params.entity_type_ids)
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
@@ -746,26 +746,24 @@ where
         // We will use the added entity type IDs to check for the instantiation permission later.
         // This means that we need to make sure, that exactly the required entity types are passed
         // here.
-        let entity_type_id_set = params
-            .iter()
-            .flat_map(|params| {
-                let entity_id = EntityId {
-                    web_id: params.web_id,
-                    entity_uuid: params
-                        .entity_uuid
-                        .unwrap_or_else(|| EntityUuid::new(Uuid::new_v4())),
-                    draft_id: params.draft.then(|| DraftId::new(Uuid::new_v4())),
-                };
-                policy_components_builder.add_entity(
-                    actor_id,
-                    entity_id,
-                    Cow::Owned(params.entity_type_ids.iter().cloned().collect()),
-                );
-                entity_ids.push(entity_id);
+        let mut entity_type_id_set = HashSet::with_capacity(params.len());
+        for params in &params {
+            let entity_id = EntityId {
+                web_id: params.web_id,
+                entity_uuid: params
+                    .entity_uuid
+                    .unwrap_or_else(|| EntityUuid::new(Uuid::new_v4())),
+                draft_id: params.draft.then(|| DraftId::new(Uuid::new_v4())),
+            };
+            policy_components_builder.add_entity(
+                actor_id,
+                entity_id,
+                Cow::Owned(params.entity_type_ids.iter().cloned().collect()),
+            );
+            entity_ids.push(entity_id);
 
-                &params.entity_type_ids
-            })
-            .collect::<HashSet<_>>();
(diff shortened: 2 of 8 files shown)


libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,124 @@ where
...
+                policy_components_builder.add_entity(
+                    actor_id,
+                    entity_id,
+                    Cow::Owned(params.entity_type_ids.iter().cloned().collect()),
+                );
+                entity_ids.push(entity_id);
+
+                &params.entity_type_ids
+            })
+            .collect::<HashSet<_>>();
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're absolutely right. Using `flat_map` here primarily for side effects is not idiomatic Rust and makes the code less clear. A separate loop for the side effect of adding entities would improve readability by making the intent explicit.

Here's how it could be improved:

```rust
// First, collect entity IDs and entity types
let mut entity_ids = Vec::with_capacity(params.len());
for params in &params {
    let entity_id = EntityId {
        web_id: params.web_id,
        entity_uuid: params.entity_uuid.unwrap_or_else(|| EntityUuid::new(Uuid::new_v4())),
        draft_id: params.draft.then(|| DraftId::new(Uuid::new_v4())),
    };
    policy_components_builder.add_entity(
        actor_id,
        entity_id,
        Cow::Owned(params.entity_type_ids.iter().cloned().collect()),
    );
    entity_ids.push(entity_id);
}

// Then collect entity type IDs separately
let entity_type_id_set = params
    .iter()
    .flat_map(|params| &params.entity_type_ids)
    .collect::<HashSet<_>>();
```

This separates the side effect (adding entities) from the iteration for collecting entity types, making the code's intent clearer.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -802,21 +889,10 @@ where
...
-            };
-
-            if entity_id.entity_uuid != entity_id.web_id.into() {
-                checked_web_ids.insert(entity_id.web_id);
-            }
 
             let entity_provenance = EntityProvenance {
                 inferred: InferredEntityProvenance {
-                    created_by_id: actor_id,
+                    created_by_id: actor_id.into(),
```
hashdotai: Instead of setting `created_by_id: actor_id.into()`, consider using the original `actor_id` directly. The `.into()` conversion is unnecessary since you're already working with the correct type (`ActorId`) from the policy component actor resolution.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,124 @@
...
+                        resource: &ResourceId::Entity(entity_id.entity_uuid),
+                        context: RequestContext::default(),
+                    },
+                    policy_components.context(),
+                )
+                .change_context(InsertionError)?
+            {
+                Authorized::Always => {}
+                Authorized::Never => {
+                    forbidden_entity_creations.push(entity_id);
```
hashdotai, line 835: You're performing the permission checks for entity creation before actually constructing the entities, which is good for efficiency. However, you're storing the forbidden entity IDs in a Vec and then checking if it's empty. Consider returning an error immediately when you detect a forbidden entity creation to fail faster.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -392,9 +392,17 @@ fn web_create_entity_policies(role: &WebRole) -> impl Iterator<Item = PolicyCrea
             actor_type: None,
         }),
         actions: vec![ActionName::CreateEntity],
-        resource: Some(ResourceConstraint::Web {
+        resource: Some(ResourceConstraint::Entity(EntityResourceConstraint::Web {
             web_id: role.web_id,
-        }),
+            filter: match role.name {
+                RoleName::Administrator => EntityResourceFilter::All { filters: vec![] },
```
hashdotai, line 398: Consider adding a comment explaining the purpose of the filter in the Administrator vs Member roles. This would help document why Members can't create `memberOf` entities while Administrators can.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -107,6 +111,11 @@ impl PolicyComponents {
             })
     }
 
+    #[must_use]
+    pub const fn tracked_entity_types(&self) -> &HashSet<VersionedUrl> {
```
hashdotai, line 115: The method `tracked_entity_types` is defined as `const fn`, but it returns a reference to a non-trivial type (`HashSet`). Consider making this a regular function instead, as the performance benefit of `const fn` isn't applicable here and it restricts future modifications.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -371,6 +380,25 @@
         self.add_actions(actions, optimize);
         self
     }
+
+    pub fn add_entity(
+        &mut self,
+        actor: ActorId,
+        id: EntityId,
+        entity_types: Cow<'a, [VersionedUrl]>,
```
hashdotai, line 388: The parameter `entity_types` in `add_entity` is of type `Cow<'a, [VersionedUrl]>`, but you're always creating an owned value by collecting into a vector. Consider just taking a slice (`&[VersionedUrl]`) instead and handle the conversion internally.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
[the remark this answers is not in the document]
TimDiekmann: Why is this not idiomatic? You doubled the runtime of the function as you iterating twice.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: While returning an error immediately would be a different behavior (We want to check all entities and return all errored entities), but we can probably use a `try_collect_report` here, yes. It's not needed to store them into an error.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
TimDiekmann: Yes, I think it makes sense to document that.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: @indietyp how do you feel about this btw? It's not possible to const-create a `HashSet` (`HashSet::new` is not `const`), so it's never possible to use this function in a `const context. 
My take: Well, it does not hurt to have `const` here.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: Well, you _could_ pass a `Cow`. The function is directly using the `Cow`, so passing a slice potentially limits the lifetimes if you could pass a `Vec` instead.
@indietyp what do you typically do? The function signature is taking a `Cow` because it's stored as `Cow`.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Good point, will do

libs/@local/graph/authorization/src/policies/components.rs
indietyp: If possible, always do `const`, because it enables further compiler optimisations that might otherwise not be possible.

libs/@local/graph/authorization/src/policies/components.rs
indietyp: If it were only `entity_base_types`, I'd always use `impl IntoIterator` instead; in this case, here `Cow` is preferred. The rule of thumb is always: push the allocation outward if possible (staying ergonomic - don't require a scratch vector to be passed just because^^, you get what I mean). You shouldn't decide for the consumer. It also makes it less opaque and, as you've already alluded to, leads to lifetime issues down the road.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,130 @@ where
 
         let mut transaction = self.transaction().await.change_context(InsertionError)?;
 
-        let policy_components = PolicyComponents::builder(&transaction)
+        let actor_id = transaction
+            .determine_actor(actor_uuid)
+            .await
+            .change_context(InsertionError)?
+            .ok_or_else(|| Report::new(InsertionError).attach_printable("Actor not found"))?;
```
indietyp, line 740: Perhaps we should attach a dedicated type (or `String` here) that also indicates to the user *which* actor wasn't found?

Or is this already handled in the `determine_actor` function?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,130 @@ where
...
+
+        let mut policy_components_builder = PolicyComponents::builder(&transaction);
+
+        let mut entity_ids = Vec::with_capacity(params.len());
+
+        // We will use the added entity type IDs to check for the instantiation permission later.
+        // This means that we need to make sure, that exactly the required entity types are passed
+        // here.
+        let entity_type_id_set = params
```
indietyp: I am not a huge fan of having a `flat_map` that is primarily used to modify outside state. I think having a `for` loop here would help readability.

```rs
let mut entity_type_id_set = HashSet::with_capacity(params.len());

for param in &params {
	// what we had here before
	entity_type_id_set.extend(&params.entity_type_ids);
}

```

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,130 @@ where
...
-            .with_entity_type_ids(
-                params
-                    .iter()
-                    .flat_map(|params| &params.entity_type_ids)
-                    .collect::<HashSet<_>>(),
-            )
-            .with_action(ActionName::Instantiate, false)
-            .with_action(ActionName::CreateEntity, false)
+            .with_entity_type_ids(entity_type_id_set)
+            .with_actions([ActionName::Instantiate, ActionName::CreateEntity], false)
```
indietyp, line 773: Maybe it'd make sense to switch from a boolean to an enum in a later PR (or drive-by) because at the call-site it is not clear what the boolean is supposed to mean here.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,130 @@ where
...
+        if !forbidden_instantiations.is_empty() {
+            return Err(Report::new(InsertionError)
+                .attach(StatusCode::PermissionDenied)
+                .attach_printable(
+                    "The actor does not have permission to instantiate one or more entity types",
+                )
+                .attach_printable(
+                    forbidden_instantiations
+                        .into_iter()
+                        .map(ToString::to_string)
```
indietyp, line 812: in theory, you don't need to double allocate here, nightly has https://doc.rust-lang.org/stable/core/iter/trait.Iterator.html#method.intersperse so that you can do:

```rs
forbidding_instantiations.into_iter().map(ToString::to_string).map(Cow::Owned).intersperse(Cow::Borrowed(", ")).collect::<String>()
```

Not that it is required here, this is likely overkill, but just as a hint that `intersperse` exists as there are quite a few places where it is super helpful.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -735,24 +733,130 @@ where
...
+                        .collect::<Vec<_>>()
+                        .join(", "),
+                ));
+        }
+
         let validator_provider = StoreProvider::new(&transaction, &policy_components);
 
         let mut validation_reports = HashMap::<usize, EntityValidationReport>::new();
-        for (index, mut params) in params.into_iter().enumerate() {
+        for (index, (mut params, &entity_id)) in params.into_iter().zip(&entity_ids).enumerate() {
```
indietyp, line 866:
```suggestion
        debug_assert_eq!(params.len(), entity_ids.len(), "Params and entity ids must be the same length");
        for (index, (mut params, &entity_id)) in params.into_iter().zip(&entity_ids).enumerate() {
```

It may be sensible to add a debug assertion here. I know that this will never be hit (right now), but since `zip` stops once one of the iterators has been exhausted, I could imagine that in the future, this could become a silent troublemaker.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: The actor is captured in the instrument above

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: I'm going to replace it with an enum, yes. I didn't expect this to keep around for longer, but it's actually required. But won't do it here because the rebasing changes for follow-up work would be quite annoying (and in the IDE at least you get the `optimize: true` or `optimize: false` inline hint)

TimDiekmann: I created H-4952 as follow-up.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: I didn't know that `std` has it, I only knew about itertools (and for that, it's actually a decent reason to pull `itertools` because it's super handy). Thanks for the flag! 18566a5a8a49ccf5eb996cf532eaf5c19ac27d21

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: 9ea995a40166f35b7cc34ed3925a9f73b27632eb

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Made it a `for`-loop: cabfedc85c65384ad70c96b6081323045e97e6b7

PR #7575, opened 2025-07-07
H-4574, H-4933: Implement CRUD policies for entity types

TimDiekmann:

With `Entity`s being migrated to the new policy engine, I now need to migrate the entity type policies.

## 🔍 What does this change?

- Update actions and schemas to include entity type actions
- Add policies for entity types:
    - Everyone can view an entity type
    - Distinguish between local and remote types in policies (H-4574)
    - For remote entity types: Registered Machines and users can create them
    - For local entity types:
        - Web members and admins can create them
        - Web members and admins can update them
        - Web admins can archive them
- Replace SpiceDB permission checks with policy checks (main part of this PR)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The Node API currently still uses SpiceDB checks

## 🛡 What tests cover this?

Every test which is dealing with entity types now goes through the new policies


libs/@local/graph/store/src/entity_type/store.rs
```diff
@@ -246,6 +249,16 @@ pub struct UpdateEntityTypeEmbeddingParams<'a> {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct HasPermissionForEntityTypesParams<'a> {
+    #[cfg_attr(feature = "utoipa", schema(value_type = String))]
+    pub action: ActionName,
+    pub entity_type_ids: Cow<'a, [VersionedUrl]>,
+    pub include_parents: bool,
```
CiaranMn: Can we document how this changes the behavior of the check?

hashdotai: @CiaranMn The behavior change is quite significant. Previously, the `can_instantiate_entity_types` function returned a boolean array with results in the same order as the input entity type IDs, indicating whether the user could instantiate each type. 

The new `has_permission_for_entity_types` function is more flexible and powerful:

1. It works with any action type (not just instantiation), supporting `viewEntityType`, `updateEntityType`, `archiveEntityType`, etc.
2. It returns a set of entity type IDs that the user has permission for, rather than a boolean array matching the input order
3. It has an `include_parents` option that checks permissions on parent entity types when relevant (especially important for instantiation permissions which are inherited)

This new approach integrates better with the policy engine and provides a more consistent experience across different permission types.

TimDiekmann: @hashdotai can you prepare a doc-string for this? (Current features, not previous features)

libs/@local/graph/authorization/src/policies/action/mod.rs
```diff
@@ -35,21 +35,23 @@ pub enum ActionName {
     #[cfg_attr(feature = "codegen", specta(skip))]
     Create,
     CreateEntity,
+    CreateEntityType,
     CreateWeb,
 
     #[cfg_attr(feature = "codegen", specta(skip))]
     View,
     ViewEntity,
```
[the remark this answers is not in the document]
TimDiekmann: The point of this PR is to _not_ skip it anymore.

libs/@local/graph/store/src/entity_type/store.rs
hashdotai: @CiaranMn The function `has_permission_for_entity_types` now provides these key capabilities:

1. **Action flexibility**: Works with any entity type action (viewEntityType, updateEntityType, archiveEntityType, createEntityType, instantiate)
2. **Return format**: Returns a HashSet of VersionedUrls the actor has permission for, allowing for efficient lookups
3. **Parent inheritance**: With `include_parents: true`, it checks permissions on parent entity types when needed (critical for instantiation permissions that follow inheritance rules)
4. **Remote type support**: Properly handles permissions for remote/external entity types that may not exist in the local database yet

This approach integrates more seamlessly with the policy engine and provides a consistent permission-checking mechanism across all entity type operations.

libs/@local/graph/sdk/typescript/src/entity-type.ts
```diff
@@ -1,25 +1,31 @@
...
- * Returns whether the user can instantiate the given entity types.
- *
- * The returned array has the same length and order as the input array of `entityTypeIds`.
- * Each boolean value at index `i` indicates whether the user is permitted to instantiate the
- * entity type at `entityTypeIds[i]`.
- */
-export const canInstantiateEntityTypes = <
-  T extends readonly [VersionedUrl, ...VersionedUrl[]],
->(
+export const hasPermissionForEntityTypes = (
```
hashdotai, line 11: Consider adding a more descriptive docstring here to explain what this function does and how it should be used, similar to what you provided in the PR comment. This would make the code more self-documenting.

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
```diff
@@ -98,21 +101,42 @@ where
...
+
+            let entity_type_ids_filter = Filter::for_entity_type_uuids(&ids);
+            compiler
+                .add_filter(&entity_type_ids_filter)
+                .change_context(QueryError)?;
+
+            // TODO: Ideally, we'd incorporate the filter in the caller function, but that's not
+            //       easily possible as the query there uses features that the query compiler does
+            //       not support yet.
+            let permission_filter = Filter::<EntityTypeWithMetadata>::for_policies(
```
hashdotai, line 114: The optimization_data access is not null-checked here. If the action is not tracked in the components, this might still return a default empty OptimizationData rather than crashing, but it would be more explicit to add a null check or document this behavior.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -304,6 +304,29 @@ fn global_view_entity_policies() -> impl Iterator<Item = PolicyCreationParams> {
     public_policies.chain(authenticated_actor_policies)
 }
 
+fn global_view_ontology_policies() -> impl Iterator<Item = PolicyCreationParams> {
+    [ActorType::User, ActorType::Machine]
+        .into_iter()
+        .map(|actor_type| PolicyCreationParams {
+            name: Some("authenticated-create-external-ontology".to_owned()),
```
hashdotai, line 311: This policy should have a name that reflects its actual purpose (allowing remote entity type creation) rather than a generic 'authenticated-create-external-ontology' name.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -304,6 +304,29 @@
...
+            principal: Some(PrincipalConstraint::ActorType { actor_type }),
+            actions: vec![ActionName::CreateEntityType],
+            resource: Some(ResourceConstraint::EntityType(
+                EntityTypeResourceConstraint::Any {
+                    filter: EntityTypeResourceFilter::IsRemote,
+                },
+            )),
+        })
+        .chain(iter::once(PolicyCreationParams {
+            name: Some("public-view-ontology".to_owned()),
```
hashdotai, line 322: For maintenance clarity, consider adding a brief comment explaining that this view policy is for all entity types, regardless of whether they're remote or local. This makes the policy's scope clear.

libs/@local/graph/authorization/src/policies/components.rs
```diff
@@ -7,26 +7,35 @@ use type_system::{
...
 #[derive(Debug, Default)]
+#[expect(
+    clippy::struct_field_names,
+    reason = "We want to add `forbid` fields in the future as well"
+)]
 pub struct OptimizationData {
     /// Entity UUIDs that can be optimized to IN clause for exact entity permits
     pub permitted_entity_uuids: Vec<EntityUuid>,
+    /// Entity Type UUIDs that can be optimized to IN clause for exact entity type permits
+    pub permitted_entity_type_uuids: Vec<EntityTypeUuid>,
```
hashdotai, line 38: Since you're adding entity type fields to `OptimizationData`, you might want to add similar fields for property types in a future PR to maintain consistency across all ontology types.

libs/@local/graph/postgres-store/src/store/postgres/ontology/entity_type.rs
TimDiekmann: At most this should generate a warning, because then it could be optimized but it was not. However, if you check for `viewEntityType` in `hasPermissionForEntityTypes` and include parents, you'd need both, an optimized and an unoptimized policy set, which is currently not supported, thus, it would always error/warn.

libs/@local/graph/authorization/src/policies/components.rs
TimDiekmann: As you mentioned, in a future PR, yes: H-4934

libs/@local/graph/store/src/entity_type/store.rs
TimDiekmann: We agreed offline that we're going to remove this paramter. Instead, we will determine based on the action what we do. Further, we're going to remove the `instantiate` permission in the future: H-4956

CiaranMn (approved):

PR #7583, opened 2025-07-08
H-4934: Implement CRUD policies for entity types

TimDiekmann:

With `EntityType`s being migrated to the new policy engine, I now need to migrate the property type policies.

## 🔍 What does this change?

- Update actions and schemas to include property type actions
- Add policies for property types:
    - Everyone can view an property type
    - For remote entity types: Registered Machines and users can create them
    - For local entity types:
        - Web members and admins can create them
        - Web members and admins can update them
        - Web admins can archive them
- Replace SpiceDB permission checks with policy checks (main part of this PR)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The Node API currently still uses SpiceDB checks

## 🛡 What tests cover this?

Every test which is dealing with entity types now goes through the new policies


libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
```diff
@@ -108,100 +140,106 @@ where
...
+            let (statement, parameters) = compiler.compile();
+
+            let property_type_rows = self
+                .as_client()
+                .query(&statement, parameters)
+                .instrument(tracing::trace_span!("query"))
+                .await
+                .change_context(QueryError)?;
+
+            Some(property_type_rows.len())
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're absolutely right. Using the result set size isn't the correct way to get the total count when pagination is involved. The approach you suggest with a separate COUNT(*) query is much better for accurate counting.

Your suggested implementation looks good - it creates a separate compiler instance for the count query, sets it to select COUNT(*), and then executes that query to get the accurate total count rather than just counting the rows in the current page.

This would ensure that `include_count` returns the actual total number of matching property types across all pages, not just the count of the current page limited by pagination parameters.

libs/@local/graph/sdk/typescript/src/property-type.ts
```diff
@@ -0,0 +1,31 @@
...
+        "viewPropertyType" | "updatePropertyType" | "archivePropertyType"
+      >;
+    }
+  >,
+): Promise<{ [K in keyof T]: VersionedUrl }> =>
+  graphAPI
+    .hasPermissionForPropertyTypes(authentication.actorId, params)
+    .then(
+      ({ data: permitted }) => permitted as { [K in keyof T]: VersionedUrl },
+    );
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're right about the type safety issue here. The server doesn't guarantee preservation of input order, which makes the return type declaration potentially misleading. 

A better approach would be to return a simple `VersionedUrl[]` array and have the client match it back to their input if needed, or to build a map of results. Here's how I would improve this:

```typescript
export const hasPermissionForPropertyTypes = (
  graphAPI: GraphApi,
  authentication: AuthenticationContext,
  params: Subtype<
    HasPermissionForPropertyTypesParams,
    {
      propertyTypeIds: VersionedUrl[];
      action: Subtype<
        ActionName,
        "viewPropertyType" | "updatePropertyType" | "archivePropertyType"
      >;
    }
  >,
): Promise<VersionedUrl[]> =>
  graphAPI
    .hasPermissionForPropertyTypes(authentication.actorId, params)
    .then(({ data: permitted }) => permitted as VersionedUrl[]);
```

This way it's clear that the function returns just an array of permitted IDs, and the client can do any needed mapping afterward.

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
hashdotai, line 178: Using the page row count as the total count means `include_count` no longer returns the full count of matching property types. You should run a separate COUNT(*) query (or use a window function) to produce an accurate total.

```suggestion
            let mut count_compiler = compiler.clone();
            count_compiler.set_selection("COUNT(*)");
            let (count_statement, count_parameters) = count_compiler.compile();

            let count_rows = self
                .as_client()
                .query(&count_statement, count_parameters)
                .instrument(tracing::trace_span!("count_query"))
                .await
                .change_context(QueryError)?;

            count_rows
                .first()
                .and_then(|row| row.get::<_, i64>(0))
                .map(|count| count as usize)
```

libs/@local/graph/sdk/typescript/src/property-type.ts
hashdotai: The return type declaration assumes the server preserves input order, which it doesn't. This creates a potentially misleading type safety issue.

```suggestion
export const hasPermissionForPropertyTypes = (
  graphAPI: GraphApi,
  authentication: AuthenticationContext,
  params: Subtype<
    HasPermissionForPropertyTypesParams,
    {
      propertyTypeIds: VersionedUrl[];
      action: Subtype<
        ActionName,
        "viewPropertyType" | "updatePropertyType" | "archivePropertyType"
      >;
    }
  >,
): Promise<VersionedUrl[]> =>
  graphAPI
    .hasPermissionForPropertyTypes(authentication.actorId, params)
    .then(({ data: permitted }) => permitted as VersionedUrl[]);
```

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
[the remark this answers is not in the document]
TimDiekmann: Pagination is applied later.

libs/@local/graph/sdk/typescript/src/property-type.ts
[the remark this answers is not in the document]
TimDiekmann: It's a rebasing error, thanks for highlighting. Fixed in https://github.com/hashintel/hash/pull/7583/commits/479cdfe35ffafad768747da247f437702c6732f2

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -52,7 +56,7 @@
     error::{InsertionError, UpdateError},
     query::ConflictBehavior,
 };
-use hash_graph_temporal_versioning::{LeftClosedTemporalInterval, TransactionTime};
+use hash_graph_temporal_versioning::{LeftClosedTemporalInterval, Timestamp, TransactionTime};
```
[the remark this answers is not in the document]
TimDiekmann:
```suggestion
use hash_graph_temporal_versioning::{LeftClosedTemporalInterval, TransactionTime};
```

libs/@local/graph/authorization/src/policies/store/mod.rs
```diff
@@ -344,6 +345,18 @@ pub trait PolicyStore {
         Output = Result<Vec<EntityTypeResource<'_>>, Report<[BuildEntityTypeContextError]>>,
     > + Send;
 
+    /// Builds a context used to evaluate policies for a set of entity types.
```
CiaranMn:
```suggestion
    /// Builds a context used to evaluate policies for a set of property types.
```

libs/@local/graph/authorization/src/policies/resource/property_type.rs
```diff
@@ -0,0 +1,391 @@
...
+            HashSet::new(),
+            self.web_id
+                .as_ref()
+                .into_iter()
+                .map(WebId::to_euid)
+                .collect(),
+            iter::empty(),
+            Extensions::none(),
+        )
+        .expect("Entity type should be a valid Cedar entity")
```
CiaranMn:
```suggestion
        .expect("Property type should be a valid Cedar entity")
```

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
```diff
@@ -861,6 +1021,40 @@ where
...
+                policy_components.context(),
+            )
+            .change_context(UpdateError)?
+        {
+            Authorized::Always => {}
+            Authorized::Never => {
+                return Err(Report::new(UpdateError)
+                    .attach(StatusCode::PermissionDenied)
+                    .attach_printable(format!(
+                        "The actor does not have permission to publish the property type `{}`",
```
CiaranMn: Suggest we use 'unarchive' as 'publish' is not a concept we really use anywhere else... could also mean something like 'make public when it was private'. Easier to be explicit that this is unarchiving.

```suggestion
                        "The actor does not have permission to unarchive the property type `{}`",
```

libs/@local/graph/postgres-store/src/store/validation.rs
```diff
@@ -361,28 +361,56 @@ where
...
+            Report::new(QueryError).attach_printable(format!(
+                "Expected exactly one property type to be returned from the query but {} were \
+                 returned",
+                schemas.len(),
+            ))
+        );
+
+        schemas.pop().ok_or_else(|| {
+            Report::new(QueryError).attach_printable(
+                "Expected exactly one property type to be returned from the query but none was \
```
CiaranMn: Nit: none/zero is plural 
```suggestion
                "Expected exactly one property type to be returned from the query but none were \
```

CiaranMn (commented):
Looks good and works, just a few comment nits

TimDiekmann:
Thanks, @CiaranMn, applied your comments 🙂

CiaranMn (approved):

PR #7586, opened 2025-07-08
H-4945: Implement permission to add/remove from account groups

TimDiekmann:

This PR implements a proper permission system for adding and removing actors from account groups, replacing the previously hard-coded admin-only approach. The changes move from manual authorization relationship management to a cleaner role-based system using the graph SDK.

## 🔗 Related links

- [H-4945: Implement permission to add/remove from account groups](https://linear.app/hashintel/issue/H-4945)

## 🚫 Blocked by

- [ ] N/A

## 🔍 What does this change?

- **Consolidates permission management**: Creates a new `@local/hash-graph-sdk/principal/actor-group` module that wraps all actor group operations
- **Removes deprecated module**: Deletes `apps/hash-api/src/graph/account-permission-management.ts` and migrates all functions to the SDK
- **Simplifies permission checks**: Replaces manual `checkWebPermission` calls with `getActorGroupRole()` for cleaner role-based authorization
- **Updates AI assistant permissions**: Streamlines AI assistant permission granting to use `assignActorGroupRole()` instead of manual relationship creation
- **Standardizes API usage**: All actor group operations now consistently use the graph SDK instead of direct API calls
- **Updates OpenAPI schema**: Reflects the new role-checking API endpoints

Key changes include:
- New SDK functions: `getActorGroupRole` to replace `isActorGroupMember` and `isActorGroupAdmin`
- Updated all callers across the codebase to use the new SDK functions
- Simplified Linear integration and AI worker permission handling

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-ai-worker-ts/src/activities/flow-activities/user-has-permission-to-run-flow-in-web-activity.ts
+++ b/apps/hash-ai-worker-ts/src/activities/flow-activities/user-has-permission-to-run-flow-in-web-activity.ts
@@ -13,8 +13,8 @@ export const userHasPermissionToRunFlowInWebActivity = async (): Promise<
       status: "ok";
     }
   | {
-      status: "missing-permission";
-      missingPermissions: "create_entity"[];
+      status: "not-role-in-web";
+      errorMessage: string;
     }
 > => {
   const { userAuthentication, webId } = await getFlowContext();
@@ -26,8 +26,8 @@ export const userHasPermissionToRunFlowInWebActivity = async (): Promise<
 
   if (!webRole) {
     return {
-      status: "missing-permission",
-      missingPermissions: ["create_entity"],
+      status: "not-role-in-web",
+      errorMessage: "User is not assigned to any role in the web",
     };
   }
 
--- a/apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
+++ b/apps/hash-ai-worker-ts/src/workflows/run-flow-workflow.ts
@@ -228,9 +228,7 @@ export const runFlowWorkflow = async (
   const userHasPermissionToRunFlowInWeb = await userHasPermissionActivity();
 
   if (userHasPermissionToRunFlowInWeb.status !== "ok") {
-    const errorMessage = `User does not have permission to run flow in web ${webId}, because they are missing permissions: ${userHasPermissionToRunFlowInWeb.missingPermissions.join(
-      `,`,
-    )}`;
+    const errorMessage = `User does not have permission to run flow in web ${webId}: ${userHasPermissionToRunFlowInWeb.errorMessage}`;
     throw ApplicationFailure.create({
       message: errorMessage,
       details: [


tests/hash-backend-integration/src/tests/graph/knowledge/system-types/hash-instance.test.ts
```diff
@@ -75,17 +75,17 @@ describe("Hash Instance", () => {
...
       }),
-    ).toBeFalsy();
+    ).toBe(null);
   });
 
   it("can add a hash instance admin", async () => {
     expect(
       await getActorGroupMembers(
-        graphContext,
+        graphContext.graphApi,
```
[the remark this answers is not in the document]
TimDiekmann: It's there: `{ actorId: systemAccountId }`

apps/hash-ai-worker-ts/src/activities/flow-activities/user-has-permission-to-run-flow-in-web-activity.ts
```diff
@@ -17,15 +19,12 @@ export const userHasPermissionToRunFlowInWebActivity = async (): Promise<
...
+  const webRole = await getActorGroupRole(graphApiClient, userAuthentication, {
+    actorId: userAuthentication.actorId,
+    actorGroupId: webId,
+  });
 
-  if (!canCreateEntities) {
+  if (!webRole) {
     return {
       status: "missing-permission",
       missingPermissions: ["create_entity"],
```
CiaranMn: This is a slight downgrade in that a user could in future have a role in the web without the ability to create entities in it, and also a user could (somehow) be given permission to create entities in the web without being a member (not sure if this would ever happen though)... are we going to have an equivalent `checkWebPermission` function at some point?

The error return is also somewhat misleading now as the actual issue is that the user doesn't have a role in the web, not that they miss a `create_entity` permission in it.

apps/hash-ai-worker-ts/src/activities/get-ai-assistant-account-id-activity.ts
```diff
@@ -23,57 +21,10 @@ export const getAiAssistantAccountIdActivity = async (params: {
...
-    const aiAssistantHasPermission = await graphApiClient
-      .checkWebPermission(
-        aiAssistantAccountId,
-        grantCreatePermissionForWeb,
-        "update_entity",
-      )
-      .then((resp) => resp.data.has_permission);
-
-    if (!aiAssistantHasPermission) {
-      /** The AI Assistant does not have permission in the requested web, use the web-scoped bot to grant it */
```
CiaranMn, line 35: I'm assuming we deleted this branch because `addActorGroupMember` is a noop if the actor is already a member (rather than throwing an error as I guess it did before if I added this)

apps/hash-ai-worker-ts/src/activities/get-ai-assistant-account-id-activity.ts
```diff
@@ -23,57 +21,10 @@ export const getAiAssistantAccountIdActivity = async (params: {
...
-              relation: "entityEditor",
-            },
-          },
-        ],
-      );
-    }
+    await addActorGroupMember(graphApiClient, authentication, {
+      actorId: aiAssistantAccountId,
+      actorGroupId: grantCreatePermissionForWeb,
+    });
```
CiaranMn: Posted a longer message in Slack, but the implication of this is that normal members cannot initiate AI flows (at least the first one, because only admins can add members to groups, and it needs to happen the first time an AI flow is run for a web)

apps/hash-ai-worker-ts/src/activities/flow-activities/user-has-permission-to-run-flow-in-web-activity.ts
TimDiekmann: It's more complicated than "Can I create an entity in a Web?", it always has been. The `CreateEntity` permission takes into account the type of an entity and maybe other things in the future as well (previously, this was only modelled through the `instantiate` permission). It's possible to expose something like "Can I create _this_ entity" and pass an entity, but there would never be a general "A user in general can create entities in a web". Checking, if the actor is somehow related to the web is probably the best we can do here without more information about the entity, which is going to be created.

apps/hash-ai-worker-ts/src/activities/get-ai-assistant-account-id-activity.ts
TimDiekmann: Yes. If the user is already assigned to this role, it will ignore it (it will error if it's assigined to a different role as we don't want actors to be assigned to multiple roles).

apps/hash-ai-worker-ts/src/activities/flow-activities/user-has-permission-to-run-flow-in-web-activity.ts
TimDiekmann: I improved the error message to render
> User is not assigned to any role in the web

CiaranMn (approved):

PR #7630, opened 2025-07-15
H-4932: Simplify policy handling around Org Invitations

TimDiekmann:

The system user has access to the invitation for orgs. Currently, this is granted per invitation but it can be assigned by type.

## 🔍 What does this change?

- Remove manual assignment of policies for invitations
- Assign policies based on entity type


CiaranMn (approved):

TimDiekmann:
Do we also need the `UpdateEntity` permission?

CiaranMn (approved):
We don't need update permission, no, just view and archive

PR #7670, opened 2025-07-21
H-5025: Use an `ast::Value` instead of an `ast::Expr` for Cedar entity creation

TimDiekmann:

We encountered a panic in production when using `ast::Entity::new()`:

```
thread 'main' panicked at libs/@local/graph/authorization/src/policies/resource/meta.rs:62:10:
Policy should be a valid Cedar entity: EntityAttrEvaluationError { uid: EntityUID(EntityUIDImpl { ty: EntityType(Name(InternalName { id: Id("Policy"), path: [Id("HASH")], loc: None })), eid: Eid("521b8df8-662f-4730-b7c6-049ef7e4f0e6"), loc: None }), attr_or_tag: "actions", was_attr: true, err: RecursionLimit(RecursionLimitError { source_loc: None }) }
stack backtrace:
   0: __rustc::rust_begin_unwind
             at ./rustc/9982d6462bedf1e793f7b2dbd655a4e57cdf67d4/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at ./rustc/9982d6462bedf1e793f7b2dbd655a4e57cdf67d4/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at ./rustc/9982d6462bedf1e793f7b2dbd655a4e57cdf67d4/library/core/src/result.rs:1761:5
   3: hash_graph_authorization::policies::context::ContextBuilder::add_policy_meta_resource
   4: hash_graph_postgres_store::store::postgres::seed_policies::<impl hash_graph_postgres_store::store::postgres::PostgresStore<tokio_postgres::transaction::Transaction>>::update_seeded_policies::{{closure}}.20812
   5: <hash_graph_postgres_store::store::postgres::PostgresStore<C> as hash_graph_authorization::policies::store::PolicyStore>::seed_system_policies::{{closure}}.20772
   6: tokio::runtime::runtime::Runtime::block_on
   7: hash_graph::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

This PR avoids the issue by using another function instead. 

## 🔍 What does this change?

- Replace usage of `ast::Entity::new`
- Drive-by: Add more logging when seeding policies
- Drive-by: Fix snapshot restoring for archived policies
- Drive-by: Fix snapshot dumping (it waited for the remaining sender, which never was closed)

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/snapshot/mod.rs
+++ b/libs/@local/graph/postgres-store/src/snapshot/mod.rs
@@ -840,16 +840,6 @@ impl PostgresStorePool {
                 sink.sink_map_err(|report| report.change_context(SnapshotDumpError::Write)),
             ));
 
-            scope.spawn(
-                stream::once(ready(Ok(SnapshotEntry::Snapshot(SnapshotMetadata {
-                    block_protocol_module_versions: BlockProtocolModuleVersions {
-                        graph: semver::Version::new(0, 3, 0),
-                    },
-                    custom: CustomGlobalMetadata,
-                }))))
-                .forward(snapshot_record_tx.clone()),
-            );
-
             if settings.dump_principals {
                 scope.spawn(
                     self.read_users()
@@ -975,6 +965,16 @@ impl PostgresStorePool {
                         .forward(snapshot_record_tx.clone()),
                 );
             }
+
+            scope.spawn(
+                stream::once(ready(Ok(SnapshotEntry::Snapshot(SnapshotMetadata {
+                    block_protocol_module_versions: BlockProtocolModuleVersions {
+                        graph: semver::Version::new(0, 3, 0),
+                    },
+                    custom: CustomGlobalMetadata,
+                }))))
+                .forward(snapshot_record_tx),
+            );
         });
 
         for result in results {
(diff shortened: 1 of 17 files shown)


libs/@local/graph/postgres-store/src/snapshot/mod.rs
```diff
@@ -975,6 +965,16 @@ impl PostgresStorePool {
...
             }
+
+            scope.spawn(
+                stream::once(ready(Ok(SnapshotEntry::Snapshot(SnapshotMetadata {
+                    block_protocol_module_versions: BlockProtocolModuleVersions {
+                        graph: semver::Version::new(0, 3, 0),
+                    },
+                    custom: CustomGlobalMetadata,
+                }))))
+                .forward(snapshot_record_tx),
```
TimDiekmann, line 976: No `.clone()` anymore. Moved to the end as the clones above are conditional.

libs/@local/graph/postgres-store/src/snapshot/policy/batch.rs
```diff
@@ -55,6 +55,7 @@ where
                         "
                             INSERT INTO policy_tmp
                             SELECT DISTINCT * FROM UNNEST($1::policy[])
+                            ON CONFLICT DO NOTHING
```
TimDiekmann, line 58: When multiple policy editions exist in the snapshot there is a conflict if the amount of policies is larger than the batch size (10k policies)

CiaranMn (approved):

PR #7673, opened 2025-07-22
H-5029: Add `pdf-cmaps` explainer

vilkinsons:

Adds a README to the `pdf-cmaps` dir to explain their purpose.

### DIFF
--- a/apps/hash-frontend/public/pdf-cmaps/README.md
+++ b/apps/hash-frontend/public/pdf-cmaps/README.md
@@ -1,0 +1,7 @@
+# pdf-cmaps
+
+`.bcmap` files are "binary character maps" — text files used in PostScript and other Adobe products to map character codes to character glyphs in CID fonts. (See: [Stack Overflow](https://stackoverflow.com/a/40634723))
+
+[This spec document](https://www.adobe.com/content/dam/acom/en/devnet/font/pdfs/5014.CIDFont_Spec.pdf) by Adobe outlines what CID fonts are good for (primarily useful when dealing with East Asian writing systems; a legacy support technology).
+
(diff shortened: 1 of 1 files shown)


apps/hash-frontend/public/pdf-cmaps/README.md
```diff
@@ -0,0 +1,7 @@
+# pdf-cmaps
+
+`.bcmap` files are "binary character maps" — text files used in PostScript and other Adobe products to map character codes to character glyphs in CID fonts. (See: [Stack Overflow](https://stackoverflow.com/a/40634723))
```
hashdotai, line 3: Consider adding more context about why these files are in our codebase specifically. For example, which library or component requires them, and any specific configuration needed to make them work properly.

apps/hash-frontend/public/pdf-cmaps/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+[This spec document](https://www.adobe.com/content/dam/acom/en/devnet/font/pdfs/5014.CIDFont_Spec.pdf) by Adobe outlines what CID fonts are good for (primarily useful when dealing with East Asian writing systems; a legacy support technology).
+
+These files are used when displaying PDFs containing CID fonts and therefore need to be publicly accessible.
```
hashdotai: It would be helpful to add information about:
1. Where these files came from (e.g., did we generate them, download them from a specific source?)
2. If/when they need to be updated
3. How they relate to our PDF viewer implementation

vilkinsons:
```suggestion
These files are used when displaying PDFs containing CID fonts and therefore need to be publicly accessible. Within HASH, they are utilized within the in-app file viewer, specifically to support PDF rendering.
```

PR #7675, opened 2025-07-22
H-5033: Avoid checking permissions when database seeding happens

TimDiekmann:

We don't need to check permissions when seeding the default policies


indietyp (approved):
Sorry for the delay, took a bit longer than anticipated. It's a large PR afterall

PR #7689, opened 2025-07-29
H-4317: Improve `mise` config tool declaration

vilkinsons:

Moves CI-only tooling out of main `config.toml` file and into `config.ci.toml` 

We can likely also move other tools to the `config.dev.toml` file, but I haven't touched that here. Could be done on the same branch, or in follow-up.


TimDiekmann (approved):
> We can likely also move other tools to the config.dev.toml file, but I haven't touched that here. Could be done on the same branch, or in follow-up.

Please not in the same branch, this is more complicated because you need to specify what environment you want to use when installing the tools.

PR #7708, opened 2025-08-05
H-4628: Allow batch-creation of policies

TimDiekmann:

While we don't necessarily need batch-creation of policies in production, seeding tests and benchmarks are quite slow without it.

## 🔍 What does this change?

Change the internal `insert_policy_into_database` to take a batch of policies.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
@@ -197,7 +197,13 @@ impl PostgresStore<tokio_postgres::Transaction<'_>> {
                 peer.service = "Postgres",
             ))
             .await
-            .change_context(CreatePolicyError::StoreError)?;
+            .map_err(|error| {
+                let policy_error = match error.code() {
+                    Some(&SqlState::UNIQUE_VIOLATION) => CreatePolicyError::PoliciesAlreadyExist,
+                    _ => CreatePolicyError::StoreError,
+                };
+                Report::new(error).change_context(policy_error)
+            })?;
 
         client
             .execute(
@@ -238,7 +244,14 @@ impl PostgresStore<tokio_postgres::Transaction<'_>> {
                 peer.service = "Postgres",
             ))
             .await
-            .change_context(CreatePolicyError::StoreError)?;
+            .map_err(|error| {
+                let policy_error = match error.code() {
+                    Some(&SqlState::UNIQUE_VIOLATION) => CreatePolicyError::PoliciesAlreadyExist,
+                    Some(&SqlState::FOREIGN_KEY_VIOLATION) => CreatePolicyError::PrincipalsNotFound,
+                    _ => CreatePolicyError::StoreError,
+                };
+                Report::new(error).change_context(policy_error)
+            })?;
 
         client
             .execute(
@@ -261,7 +274,13 @@ impl PostgresStore<tokio_postgres::Transaction<'_>> {
                 peer.service = "Postgres"
             ))
             .await
-            .change_context(CreatePolicyError::StoreError)?;
+            .map_err(|error| {
+                let policy_error = match error.code() {
+                    Some(&SqlState::FOREIGN_KEY_VIOLATION) => CreatePolicyError::ActionsNotFound,
+                    _ => CreatePolicyError::StoreError,
+                };
+                Report::new(error).change_context(policy_error)
+            })?;
 
         Ok(policy_ids)
     }
@@ -1661,7 +1680,6 @@ where
                 FROM policy_edition
                     LEFT JOIN policy_action
                     ON policy_action.policy_id = policy_edition.id
-                    AND policy_action.action_name = ANY($3)
                     AND policy_action.transaction_time @> now()
                 GROUP BY
                     policy_edition.id,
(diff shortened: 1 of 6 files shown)


libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1353,7 +1403,10 @@ where
             .await
             .change_context(CreatePolicyError::StoreError)?;
 
-        let policy_id = transaction.insert_policy_into_database(&policy).await?;
+        let policy_ids = transaction.insert_policies_into_database([&policy]).await?;
+        let &[policy_id] = policy_ids.as_slice() else {
+            unreachable!("Expected exactly one policy ID");
+        };
```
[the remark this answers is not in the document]
TimDiekmann: We pass a single policy above, so we get exactly one policy ID back.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -1353,7 +1403,10 @@ where
             .await
             .change_context(CreatePolicyError::StoreError)?;
 
-        let policy_id = transaction.insert_policy_into_database(&policy).await?;
+        let policy_ids = transaction.insert_policies_into_database([&policy]).await?;
+        let &[policy_id] = policy_ids.as_slice() else {
+            unreachable!("Expected exactly one policy ID");
```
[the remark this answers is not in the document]
TimDiekmann: See above (this is the exact same code snippet)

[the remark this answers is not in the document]
hashdotai: @TimDiekmann You're correct - since you're passing in a slice with a single element, this is a reasonable assertion to make. 

The `unreachable!()` is actually appropriate here since it's enforcing an invariant of the code - you know with certainty that exactly one policy is being inserted, so anything else would represent a bug in the implementation rather than an expected runtime error condition.

Given the controlled input (a slice with one element), this is a reasonable design choice.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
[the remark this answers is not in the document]
hashdotai: @TimDiekmann You're right - since you know with certainty that you're passing a single policy and expect a single ID back, the pattern match is appropriate. This is technically an invariant of the function call and the assertion helps detect bugs during development.

If you ever decide to make the error handling more robust in the future, you could consider something like:
```rust
let policy_id = match policy_ids.as_slice() {
    [id] => *id,
    _ => unreachable!("Expected exactly one policy ID, got {}", policy_ids.len())
};
```

But your current approach is perfectly valid given the context.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -165,52 +238,32 @@ impl PostgresStore<tokio_postgres::Transaction<'_>> {
...
-                        CreatePolicyError::PolicyAlreadyExists { id: policy_id }
-                    }
-                    (Some(&SqlState::FOREIGN_KEY_VIOLATION), Some(principal_id)) => {
-                        CreatePolicyError::PrincipalNotFound { id: principal_id }
-                    }
-                    _ => CreatePolicyError::StoreError,
-                };
-                Report::new(error).change_context(policy_error)
-            })?;
+            .change_context(CreatePolicyError::StoreError)?;
```
hashdotai: This is a good optimization for handling batches of policies. However, the error handling in `insert_policies_into_database` doesn't preserve the specific error context that the previous implementation provided (like `PolicyAlreadyExists`, `PrincipalNotFound`, or `ActionNotFound`). Consider preserving these error contexts for better debugging and error reporting.

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -106,31 +106,90 @@
...
+        policies: I,
+    ) -> Result<Vec<PolicyId>, Report<CreatePolicyError>>
+    where
+        I: IntoIterator<Item = P>,
+        P: Borrow<PolicyCreationParams>,
+    {
+        // As `into_iter` consumes the iterator, we need to collect the policies first to later
+        // borrow them
+        let policies = policies.into_iter().collect::<Vec<_>>();
+        if policies.is_empty() {
```
hashdotai, line 129: Consider adding a limit to the number of policies that can be batch-inserted at once to prevent potential performance issues with very large batches. A reasonable upper bound could help protect against accidental resource exhaustion.

libs/@local/graph/postgres-store/src/store/postgres/seed_policies.rs
```diff
@@ -1040,16 +1040,14 @@ impl PostgresStore<Transaction<'_>> {
...
-        for (index, policy) in policies_to_add.iter().enumerate() {
-            tracing::debug!(
-                policy_name = policy.name,
-                "Adding policy to database {index}/{}",
-                policies_to_add.len()
-            );
-            self.insert_policy_into_database(policy)
+
+        if !policies_to_add.is_empty() {
+            tracing::debug!("Adding {} policies to database", policies_to_add.len());
```
hashdotai, line 1045: The tracing has been simplified in the batch version. Consider adding more detailed tracing that includes information about how many policies are being inserted to maintain observability.

TimDiekmann: The number of policies IS preserved

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
TimDiekmann: I added error handling but omitted information which policy exactly failed (which should be fine)

PR #7733, opened 2025-08-19
H-5190: Implement dependencies between different seeding producers

TimDiekmann:

To create data types (or other types) we need to create webs first. These seeding configs don't know about them, yet, as they are just about to be created, so ontology type producer require a way to chose from local webs.

This PR resolves dependencies between producers, so it's possible to read webs from both, local and remote sources. Also, it's prepared that user and org webs can be configured.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

Various test cases for this have been added


tests/graph/test-data/rust/src/seeding/distributions/adaptors/uniform.rs
```diff
@@ -48,7 +48,7 @@ where
     fn create_distribution(&self) -> Result<Self::Distribution, Self::Error> {
         Ok(UniformDistribution {
             index: Uniform::new(0, self.distributions.len())
-                .expect("`Uniform` should return a valid index"),
+                .change_context(UniformDistributionConfigError::Distribution)?,
```
indietyp: error conditions are (https://docs.rs/rand/latest/rand/distr/struct.Uniform.html#method.new_inclusive):
- low > high
- low is non-finite
- high is non-finite
- high - low is non-finite

In this case _none_ of these cases can happen, so I think an `unwrap_or_else(|_err| unreachable!())` would be more fitting.

tests/graph/test-data/rust/src/seeding/distributions/ontology/data_type/constraints/string/mod.rs
```diff
@@ -28,13 +28,13 @@ use crate::seeding::distributions::DistributionConfig;
...
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
-pub struct StringConstraintsDistributionConfig<'c> {
+pub struct StringConstraintsDistributionConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub min_length: Option<MinLengthDistributionConfig>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_length: Option<MaxLengthDistributionConfig>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub pattern: Option<PatternDistributionConfig<'c>>,
+    pub pattern: Option<PatternDistributionConfig>,
```
indietyp, line 37: Comment to `PatternDistributionConfig`:

It might be interesting to bridge `arbitrary` and `rand` here, use arbitrary to generate a https://docs.rs/regex-syntax/latest/regex_syntax/ using bytes from rand, then use that in the distribution. That should trivially work and give us a real stress test (`regex_syntax::Ast` provides a `Display` implementation so that it could be trivially turned into a `Regex`).

tests/graph/test-data/rust/src/seeding/distributions/ontology/shortname.rs
```diff
@@ -14,23 +14,23 @@ use crate::seeding::distributions::{
 };
 
 #[derive(Debug, Clone, PartialEq)]
-enum InnerShortnameDistribution<'s> {
-    Const(ConstDistribution<Cow<'s, str>>),
-    Weighted(WeightedDistribution<ConstDistribution<Cow<'s, str>>>),
-    Uniform(UniformDistribution<ConstDistribution<Cow<'s, str>>>),
+enum InnerShortnameDistribution {
```
indietyp, line 17: Wouldn't it make sense here to use something like https://docs.rs/fake/latest/fake/ to generate genuine usernames / shortnames?

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
```diff
@@ -46,76 +53,488 @@ pub enum DataTypeProducerConfigError {
...
+    catalog: &'a C,
+}
+
+impl<C: WebCatalog> core::fmt::Debug for LocalUniform<'_, C> {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        fmt.debug_struct("LocalUniform").finish()
+    }
+}
+
+// Shard chooser removed; flat global catalog is used
```
indietyp: remove superfluous comment

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
```diff
@@ -46,76 +53,488 @@ pub enum DataTypeProducerConfigError {
...
+#[derive(Clone)]
+struct LocalUniform<'a, C: WebCatalog> {
+    catalog: &'a C,
+}
+
+impl<C: WebCatalog> core::fmt::Debug for LocalUniform<'_, C> {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        fmt.debug_struct("LocalUniform").finish()
+    }
+}
```
indietyp:
```suggestion
#[derive(Clone, derive_more::Debug)]
struct LocalUniform<'a, C: WebCatalog> {
	#[debug(skip)]
    catalog: &'a C,
}
```

This should also work™

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
```diff
@@ -46,76 +53,488 @@ pub enum DataTypeProducerConfigError {
...
+
+pub trait WebCatalog: Sync + Send {
+    fn len(&self) -> usize;
+
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    fn get_entry(&self, index: usize) -> Option<(Arc<str>, Arc<str>, WebId)>;
```
indietyp, line 161: Why is the `WebCatalog` indexed by numerical index instead of `WebId`?

indietyp (changes requested):
One blocking question (`WebCatalog`), otherwise mostly suggestions.

tests/graph/test-data/rust/src/seeding/distributions/ontology/data_type/constraints/string/mod.rs
TimDiekmann: I think we can add `arbitrary` later to improve the seeding. For this work I think it's good enough that we can chose from a set of patterns.

tests/graph/test-data/rust/src/seeding/distributions/ontology/shortname.rs
TimDiekmann: Have not looked at name generators, yet, but yes, a proper distribution for names would be helpful. Although, for short names I used the local ID ans shard ID to make sure they are really unique by design.

tests/graph/test-data/rust/src/seeding/distributions/ontology/data_type/constraints/string/mod.rs
indietyp: Yep, just wanted to flag it 👍

tests/graph/test-data/rust/src/seeding/distributions/ontology/shortname.rs
indietyp: Yep, just wanted to flag it 👍

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
TimDiekmann: You know how many webs exist but not which webs. I have to ideas here:
- keep it as an index and maintain a list of web IDs
- Re-generate the web ID based on the producer context and global ID but this is more work.
For now ,the first option is definitely easier because you basically need to chose a web by "random"

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
TimDiekmann:
```suggestion
```

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
indietyp: I see. The API didn't make it clear that one would choose from an index like that. I would've expected either something like:

```rs
fn choose<R>(&self, rng: &mut R) -> Option<(Arc<str>, Arc<str>, WebId)>
```

or:

```rs
fn choose<R>(&self, rng: &mut R) -> WebId
fn get(&self, id: WebId) -> Option<(Arc<str>, Arc<str>)>
```

TimDiekmann: I'm still experimenting around with that. For property types I used the `choose` approach (PR to be opened). 
I will make the API consistent later.

tests/graph/test-data/rust/src/seeding/producer/data_type.rs
```diff
@@ -46,76 +53,482 @@
...
+                        shortname,
+                        web_id,
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[derive(Clone, derive_more::Debug)]
```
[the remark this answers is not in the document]
indietyp: I forgot to send in the patch to enable the `Debug` macro 🙃  sorry!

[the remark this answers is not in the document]
TimDiekmann: No worries, fixed 😉  https://github.com/hashintel/hash/pull/7733/commits/ee65fe4f4bb9e0773f3be4da29c218d9ab257c53
You also used a TAB 😄

indietyp (approved):

PR #7735, opened 2025-08-20
 H-5191: Add multi-stage scenario runner for deterministic test data generation

TimDiekmann:

Implements scenario-based test data generation infrastructure with deterministic seeding for multi-stage data creation workflows.

## 🚫 Blocked by

- https://github.com/hashintel/hash/pull/7733

## 🔍 What does this change?

- **Scenario runner**: JSON-defined workflows for multi-stage test data generation and persistence
- **Deterministic seeding**: Adds `StageId` to `GlobalId` (UUID v8 layout) for stage-aware RNG streams
- **Data generation stages**:
  - `GenerateUsers`/`PersistUsers`: User creation and database persistence
  - `WebCatalog`: Builds `InMemoryWebCatalog` from user collections
  - `GenerateDataTypes`/`PersistDataTypes`: Type generation with local/remote ownership
  - `ResetDb`: Database cleanup between runs
- **Parallel execution**: Uses rayon to distribute data generation across configurable shards
- **Database integration**: Connection pooling and migration setup for test data persistence
- **JSON configuration**: Producer configs and scenario definitions in `config/` directory

Technical additions:

- `StageId` integration into seeding context for reproducible data generation
- Resource passing between stages via `Runner.resources` HashMap
- Automatic shard count distribution for parallel generation
- Support for local and remote data type scenarios

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Remote type persistence requires `HASH_GRAPH_ALLOWED_URL_DOMAIN_PATTERN` environment variable (defaults to localhost:3000 only)
- Large data type counts load entirely into memory during generation phase

## 🛡 What tests cover this?

- Existing determinism tests in seeding infrastructure ensure stable outputs for identical contexts
- Scenario execution serves as integration tests for the full generation + persistence pipeline

## ❓ How to test this?

1. Run the scenario test: `cargo bench --bench graph`
2. Verify console output shows per-stage metrics (counts/duration)
3. For remote-only scenarios: `export HASH_GRAPH_ALLOWED_URL_DOMAIN_PATTERN='^https?://.*'`
4. Test different scenarios by modifying JSON files in `config/scenarios/`

## 📹 Demo

Console output from scenario execution shows detailed per-stage timing and generation counts.

Example:
```
Scenario passed: {
  "steps": [
    {
      "id": "reset",
      "produced": 5,
      "duration_ms": 613
    },
    {
      "id": "users",
      "produced": 10,
      "duration_ms": 1
    },
    {
      "id": "persist-users-basic",
      "produced": 10,
      "duration_ms": 359
    },
    {
      "id": "webs",
      "produced": 10,
      "duration_ms": 0
    },
    {
      "id": "local-datatypes",
      "produced": 100,
      "duration_ms": 1
    },
    {
      "id": "remote-datatypes",
      "produced": 100,
      "duration_ms": 0
    },
    {
      "id": "persist-local-datatypes",
      "produced": 200,
      "duration_ms": 5378
    }
  ]
}
```


tests/graph/benches/config/producers/data_types/basic.json
```diff
@@ -0,0 +1,51 @@
+{
```
indietyp, line 1: What is the decision behind putting these into dedicated JSON files instead of encoding them as Rust code? I feel like the DX for these files (as there is no schema) is a lot worse than just having a set of steps inside of a builder pattern of sorts.

tests/graph/test-data/rust/src/seeding/context.rs
```diff
@@ -198,8 +231,9 @@ impl GlobalId {
     /// # use hash_graph_test_data::seeding::context::{GlobalId, RunId, ShardId, LocalId, Scope, Provenance, ProducerId, SubScope};
     /// # use uuid::Uuid;
     /// let gid = GlobalId {
-    ///     run_id: RunId::new(0xAAAA_AAAA),
-    ///     shard_id: ShardId::new(0xBBBB),
+    ///     run_id: RunId::new(0xAAAA),
+    ///     stage_id: hash_graph_test_data::seeding::context::StageId::new(0xBBBB),
```
TimDiekmann, line 235:
```suggestion
   /// # use hash_graph_test_data::seeding::context::{GlobalId, RunId, ShardId, StageId, LocalId, Scope, Provenance, ProducerId, SubScope};
   /// # use uuid::Uuid;
   /// let gid = GlobalId {
    ///     run_id: RunId::new(0xAAAA),
    ///     stage_id: StageId::new(0xBBBB),
```

tests/graph/benches/graph/scenario/stages/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Debug, derive_more::Display)]
+#[display("Failed to execute stage")]
+pub struct StageError;
+
+impl Error for StageError {}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type", rename_all = "kebab-case")]
+pub enum Stage {
+    ResetDb(super::stages::reset_db::ResetDbStage),
```
indietyp: Why are these imports different from the other ones?

tests/graph/benches/graph/scenario/runner.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        self.setup_db().await?;
+        Ok(self.pool.as_ref().expect("pool set by setup_db"))
+    }
+
+    pub(super) async fn setup_db(&mut self) -> Result<(), Report<ScenarioError>> {
+        load_env(Environment::Test);
+
+        let user = std::env::var("HASH_GRAPH_PG_USER").unwrap_or_else(|_| "graph".to_owned());
```
indietyp, line 139: You should be able to just re-use the clap parser, it has `parse_from`, as clap automatically makes use of env variables you should just:

```
DatabaseConfig::parse_from(core::empty())
```

or something similar, you get what I mean.

indietyp: That way we dont need to hard code the defaults and env name twice.

tests/graph/benches/config/producers/data_types/basic.json
TimDiekmann: To quickly experiment around, JSON is easier because this does not require to recompile the code each time. In particular for benchmarks this is relevant as they are compiled in release mode typically. I will probably return to Rust code later, but for iterating on the approach is easier to just deserialize it from a simple JSON. 

JSON is much easier to share as well.

tests/graph/benches/graph/scenario/runner.rs
TimDiekmann: I'm not going to start to use `clap` as the general config parser, clap is a CLI argument parser. We use this approach in various location (and I'm super frustrated by this), but changing this now is out of scope. The relevant issue is H-4170.

tests/graph/benches/graph/scenario/stages/mod.rs
TimDiekmann: Fixed in https://github.com/hashintel/hash/pull/7735/commits/99ed0e7b8232815ac9e794fab524c9fbf2a962e3. Isn't there a lint for this?

tests/graph/benches/graph/scenario/runner.rs
indietyp: Gotcha and very understandable! I wanted to flag because I didn't want to duplicate configuration items. Keeping track of all of them, especially when changes occur, can be a bit of a nightmare, and re-using clap seemed to be the most straightforward solution of them all :P

TimDiekmann: It would be just another hack of which we have far too many around configuration and DX already. I prefer if we would just do it properly.

PR #7736, opened 2025-08-21
H-5198: Implement PropertyType generation and persistence MVP with known limitations

TimDiekmann:
## 🌟 What is the purpose of this PR?

This PR implements a **PropertyType seeding system** for benchmark scenarios, enabling deterministic generation and persistence of PropertyType datasets. The implementation follows the established seeding architecture patterns from DataType and User seeding, providing a foundation for large-scale benchmark testing.

While functional for current benchmark needs, the implementation includes acknowledged performance limitations and missing features that are tracked separately for future optimization.

### Core Implementation:

- **PropertyType Producer System** (`tests/graph/test-data/rust/src/seeding/producer/property_type.rs`)
  - Deterministic PropertyType generation with configurable parameters
  - Support for Local and Remote ownership types
  - URL construction with proper BaseUrl handling and error propagation
  - Two-phase approach: Configuration → Binding (with catalogs) → Distribution sampling

- **PropertyValues Distribution System** (`tests/graph/test-data/rust/src/seeding/distributions/ontology/property_type/values.rs`)
  - `DataTypeCatalog` trait for PropertyType value generation
  - `InMemoryDataTypeCatalog` implementation with proper error handling
  - Safe array access using `SliceRandom::choose()` instead of manual indexing
  - Empty catalog validation and comprehensive error propagation

- **Benchmark Integration** (`tests/graph/benches/graph/scenario/stages/property_type.rs`)
  - `GeneratePropertyTypesStage` - PropertyType generation from configs
  - `PersistPropertyTypesStage` - Database persistence with Local/Remote handling
  - `BuildDataTypeCatalogStage` - DataType catalog construction for PropertyType dependencies
  - Proper error handling for edge cases (empty catalogs, missing dependencies)

### Configuration & Infrastructure:

- **Configuration System** (`tests/graph/benches/graph/config/mod.rs`)
  - `PROPERTY_TYPE_PRODUCER_CONFIGS` LazyLock for config loading
  - Integration with existing config loading patterns

- **Scenario Updates** (`tests/graph/benches/config/scenarios/full_ontology_seeding.json`)
  - Complete PropertyType pipeline: Generation → Catalog Building → Persistence
  - Proper stage ordering to avoid resource consumption issues

- **Resource Management** (`tests/graph/benches/graph/scenario/runner.rs`)
  - Extended `Resources` struct with `property_types` and `data_type_catalogs`
  - Thread-safe resource sharing between stages

### 🚢 Has this modified a publishable library?

This PR:

[...]

PR #7745, opened 2025-08-23
H-5214: Implement distribution-based value sampling for string and number schemas

TimDiekmann:

This PR implements value generation capabilities for the test data seeding framework, enabling the creation of valid `PropertyValue` instances from `ValueConstraints` definitions. This is a critical component for generating realistic test data that respects JSON schema constraints defined in data types.

The implementation provides distribution-based sampling for primitive value types (null, boolean, number, string), laying the groundwork for comprehensive property type testing with deterministic, reproducible test data.

## 🔗 Related links

- Linear Issue: [H-5214](https://linear.app/hash/issue/H-5214)
- Related PRs:
  - [H-5198: PropertyType generation MVP](https://github.com/hashintel/hash/pull/7736)
  - [H-5174: Initial deterministic seeding functionality](https://github.com/hashintel/hash/pull/7727)
  - [H-5191: Multi-stage scenario runner](https://github.com/hashintel/hash/pull/7735)

## 🚫 Blocked by

_None_

### Core Implementation

- **Added `ValueProducer`** (`tests/graph/test-data/rust/src/seeding/producer/value.rs`)
  - Implements the `Producer<PropertyValue>` trait for generating values
  - Converts `ValueConstraints` into sampling distributions
  - Maintains deterministic generation through seeded RNG with `GlobalId`

- **Distribution modules** (`tests/graph/test-data/rust/src/seeding/distributions/value/`)
  - `NumberValueDistribution`: Handles numeric constraints including min/max ranges, exclusive bounds, and enum values
  - `StringValueDistribution`: Generates strings with length constraints and enum support
  - Both distributions properly handle edge cases like empty enums and invalid ranges

### Supporting Changes

- **Made `ProducerConfig` copyable** (commit 7401a4f7b)
  - Changed `ProduceContext` from using references to owned values
  - Enables easier context propagation between producers
  - Added `for_producer()` method for creating derived contexts

- **Extended scope system** (`tests/graph/test-data/rust/src/seeding/context.rs`)
  - Added new `Scope` variants: `Boolean`, `Number`, `String`, `Array`, `Object`
  - Added `ProducerId::Value` for identifying value producers
  - These scopes ensure deterministic RNG seeding for each value type

### Technical Design Decisions

1. **Distribution-based approach**: Rather than simple random generation, we use proper statistical distributions (Uniform, Bernoulli) to ensure values are well-distributed within constraints

2. **Borrowing strategy**: The distributions can work with both owned and borrowed schemas through separate `TryFrom` implementations, avoiding unnecessary clones

3. **Deterministic generation**: Each value gets a unique `GlobalId` based on its type scope, ensuring reproducible test data across runs

This PR:

The changes in this PR:

The changes in this PR:

### Not Yet Implemented (with Linear tracking)

- **Format strings**: [H-5236](https://linear.app/hash/issue/H-5236) - Support value seeding for format strings (email, uuid, etc.)
- **Pattern strings**: [H-5237](https://linear.app/hash/issue/H-5237) - Support regex pattern constraints
- **Multiple of numbers**: [H-5238](https://linear.app/hash/issue/H-5238) - Support multipleOf numeric constraint
- **Arrays**: [H-5239](https://linear.app/hash/issue/H-5239) - Support array value generation
- **Objects**: [H-5240](https://linear.app/hash/issue/H-5240) - Support object value generation
- **AnyOf types**: [H-5241](https://linear.app/hash/issue/H-5241) - Support union type constraints

### Current Limitations

- String generation uses only alphabetic characters (via `rand::distr::Alphabetic`)
- No support for complex string patterns or formats yet

## 🐾 Next steps

1. Integrate `ValueProducer` into entity generation pipeline for property values
2. Add support for array and object value types (most critical for realistic entities)
3. Implement format string generators for common formats (email, URL, UUID)
4. Add configurable string character sets beyond alphabetic

### DIFF
--- a/tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
+++ b/tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
@@ -28,6 +28,11 @@ pub enum StringValueDistribution<'e> {
 
 impl<'e> StringValueDistribution<'e> {
     #[expect(clippy::todo, reason = "Incomplete implementation")]
+    #[expect(
+        clippy::integer_division,
+        clippy::integer_division_remainder_used,
+        reason = "We need to fill missing ranges and precision is not important here"
+    )]
     fn from_constraints(
         constraints: &StringConstraints,
     ) -> Result<Self, Report<StringValueDistributionError>> {
@@ -41,8 +46,8 @@ impl<'e> StringValueDistribution<'e> {
 
         let (min, max) = match (constraints.min_length, constraints.max_length) {
             (Some(min), Some(max)) => (min, max),
-            (Some(min), None) => (min, min << 2),
-            (None, Some(max)) => (max >> 2, max),
+            (Some(min), None) => (min, min * 4),
+            (None, Some(max)) => (max / 4, max),
             (None, None) => (4, 16),
         };
 
@@ -99,7 +104,7 @@ impl Distribution<String> for StringValueDistribution<'_> {
             Self::Enum { choices } => {
                 let choice = choices
                     .choose(rng)
-                    .unwrap_or_else(|| unreachable!("Choices somehow became empty"));
+                    .expect("Enum should always have at least one choice");
                 choice.clone()
             }
         }
--- a/tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
+++ b/tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
@@ -1,0 +1,107 @@
+use alloc::borrow::Cow;
+use core::error::Error;
+
+use error_stack::{Report, ResultExt as _};
+use rand::{Rng, distr::Alphabetic, seq::IndexedRandom as _};
+use rand_distr::{Distribution, Uniform};
+use type_system::ontology::json_schema::{StringConstraints, StringSchema};
+
+#[derive(Debug, derive_more::Display)]
+pub enum StringValueDistributionError {
+    #[display("Enum is empty")]
+    EmptyEnum,
+    #[display("Invalid string length")]
+    StringLength,
+}
+
+impl Error for StringValueDistributionError {}
+
+pub enum StringValueDistribution<'e> {
+    Constraints {
+        length: Uniform<usize>,
+        alphabet: Alphabetic,
+    },
+    Enum {
+        choices: Cow<'e, [String]>,
+    },
+}
+
+impl<'e> StringValueDistribution<'e> {
+    #[expect(clippy::todo, reason = "Incomplete implementation")]
+    fn from_constraints(
+        constraints: &StringConstraints,
+    ) -> Result<Self, Report<StringValueDistributionError>> {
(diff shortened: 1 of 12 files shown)


tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
```diff
@@ -0,0 +1,107 @@
...
+            todo!("https://linear.app/hash/issue/H-5236/support-value-seeding-for-format-strings");
+        }
+
+        if constraints.pattern.is_some() {
+            todo!("https://linear.app/hash/issue/H-5237/support-value-seeding-for-pattern-strings");
+        }
+
+        let (min, max) = match (constraints.min_length, constraints.max_length) {
+            (Some(min), Some(max)) => (min, max),
+            (Some(min), None) => (min, min << 2),
```
indietyp: Why are you using `min << 2` instead of `min * 4`? I feel like that's a lot more intuitive here (and I mean it gets optimised to the same instructions anyway: https://rust.godbolt.org/z/xWx183h7K)

tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
```diff
@@ -0,0 +1,107 @@
...
+        match self {
+            Self::Constraints { length, alphabet } => {
+                let length = length.sample(rng);
+                String::from_utf8(alphabet.sample_iter(rng).take(length).collect())
+                    .expect("should be able to convert chars to string")
+            }
+            Self::Enum { choices } => {
+                let choice = choices
+                    .choose(rng)
+                    .unwrap_or_else(|| unreachable!("Choices somehow became empty"));
```
indietyp:
```suggestion
                    .unwrap_or_else(|| unreachable!("Enum choices are non-empty by construction"));
```

indietyp: This would only be the case if we can truly guarantee it. The enum is public, though, which means that the unreachable here is incorrect. Nothing is stopping (as the enum is public)

indietyp:
```suggestion
                    .expect("enum choices should be non-empty");
```

Alternative:

tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
TimDiekmann: Mainly did this to avoid clippy warnings. I reverted it back to mul/div and added the two required expects

tests/graph/test-data/rust/src/seeding/distributions/value/string.rs
TimDiekmann: Good point. I think an `expect` is clearer here.

indietyp (approved):

indietyp (dismissed):

indietyp (approved):

PR #7753, opened 2025-08-28
ES-2: Replace `attach` -> `attach_opaque` and `attach_printable` -> `attach`

TimDiekmann:

`attach_printable` is far more used than `attach`. Also `_lazy` is not really common while `_with` is widely used in Rust. Another reason to change these is that `attach` currently has less restrictions than `attach_printable`, which implies a higher chance of using the wrong function accidentally.

## 🔍 What does this change?

- Rename
	- `attach` -> `attach_opaque`
	- `attach_lazy` -> `attach_opaque_lazy`
	- `attach_printable` -> `attach`
	- `attach_printable_lazy` -> `attach_with`
- Add `deprecation` warnings where possible
- Add diagnostic-info
- Adjust codebase to use the new functions

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/apps/hash-frontend/vercel-install.sh
+++ b/apps/hash-frontend/vercel-install.sh
@@ -14,10 +14,11 @@ yum-config-manager --add-repo https://mise.jdx.dev/rpm/mise.repo
 yum install -y mise
 eval "$(mise activate bash --shims)"
 
-echo "Installing prerequisites"
-mise install node npm:turbo java biome npm:@redocly/cli cargo-binstall cargo:wasm-pack cargo:wasm-opt protoc yq
 echo "Installing Rust toolchain: $(yq '.toolchain.channel' rust-toolchain.toml)"
+mise install yq
 mise use --global rust[profile=minimal]@$(yq '.toolchain.channel' rust-toolchain.toml)
+echo "Installing prerequisites"
+mise install node npm:turbo java biome npm:@redocly/cli cargo-binstall cargo:wasm-pack cargo:wasm-opt protoc
 echo "Rust installation completed. Checking versions:"
 mise list rust
 rustc --version


libs/error-stack/src/context.rs
```diff
@@ -121,3 +121,17 @@ impl<C: Error + Send + Sync + 'static> Context for C {
...
     }
 }
+
+/// A trait for types that can be attached to a [`Report`] without being displayed.
+pub trait OpaqueAttachment: Send + Sync + 'static {}
+
+impl<T: Send + Sync + 'static> OpaqueAttachment for T {}
+
+/// A trait for types that can be attached to a [`Report`] and displayed.
+#[diagnostic::on_unimplemented(
```
indietyp, line 131: We need to bump our MSRV to 1.78.0, I believe.

indietyp (approved):
LGTM, just one minor comment

libs/error-stack/src/context.rs
TimDiekmann: We're covered 🙂 
https://github.com/hashintel/hash/blob/d8f700f37d3af966622f544c32c91bbb340f9162/libs/error-stack/Cargo.toml#L6

mkatychev:
@TimDiekmann thanks a lot for the quick turnaround, this is a fantastic change! 

Do you have any thoughts on publishing prereleases to crates.io for `0.6`? `0.6.0-alpha` would be nice. 

A big issue with using a git source for error-stack is that the entire hash monorepo has to be checked out and this really adds up in CI.

TimDiekmann:
Sure, will try to do it early next week. I might find some time over the weekend as well 👍🏻

mkatychev:
No rush at all, it's greatly appreciated. Doing releases is kind of a pain but it makes a big difference.

TimDiekmann:
I filed a [new release](https://github.com/hashintel/hash/releases/tag/error-stack%400.6.0) 🙂

mkatychev:
Much appreciated!

PR #7759, opened 2025-08-29
H-3255, H-4893: Enable pagination in entities table, speed up row processing

CiaranMn:

1. Enables pagination in the entities table
2. Speeds up processing row data in the entities table
3. Enable 'load more rows' behavior by combining page results
4. Removes row processing from web worker (not needed for only processing 500 results at a time, which is the 'load more' size at the moment)
5. Speeds up graph viz loading by not bothering to request linked entities (since we request all entities visible to the user anyway, which will already include the links)

## ⚠️ Known issues

1. With pagination added, filtering on columns is essentially broken and needs a rethink from scratch. The current behaviour is that changing a column filter will only filter the visible rows, but this is confusing. We could change it to filter the entire result set, but then need to decide what to do about counts displayed in various places, e.g. is the headline count at the top of the table all entities available to the user, or the count after filtering? What about counts against filter options in other columns? Tracked in H-3841.
6. It is only possible to click 'next page' – not to go to the previous page, or to reset to the first page.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Basic rendering tests.

## ❓ How to test this?

1. View `/entities` [on the preview deployment](https://hash-20fksqmpv.stage.hash.ai/entities)

## 📹 Demo

https://github.com/user-attachments/assets/3c8b7cbe-2e36-42fd-bb00-c316f66ff473

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
+++ b/apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
@@ -487,8 +487,8 @@ export const EntitiesVisualizer: FunctionComponent<{
   const tableHeight =
     maxHeight ??
     `min(600px, calc(100vh - (${
-      // The magic number accounts for the page header and the pagination controls
-      HEADER_HEIGHT + TOP_CONTEXT_BAR_HEIGHT + 290 + tableHeaderHeight
+      // The magic number accounts for the page header
+      HEADER_HEIGHT + TOP_CONTEXT_BAR_HEIGHT + 230 + tableHeaderHeight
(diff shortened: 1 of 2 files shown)


vilkinsons:
@CiaranMn Can we change this so it functions as "load more" rather than "next page"? Given inability to 'go back' (only hydrate further) this makes more sense IMO.

vilkinsons (changes requested):
See above

vilkinsons (approved):
Thank you! We can get the remaining things in in follow-up:

- Gradient fade behind button
- Only show this _when at the bottom of the list_, rather than omnipresently whenever there are unloaded results

PR #7764, opened 2025-08-30
SRE-25: Fix turbo-hack to bypass cyclic dependency issue

TimDiekmann:

`turbo` does not support cyclic dependencies. We missed a few dependencies in our prune-logic. This was the reason why previously we achieved a test-failure on main, which otherwise would have been caught.


TimDiekmann:
No worries at all, it's `turbo` which causes issues here and it was foreseeable that this will happen 🙂

PR #7768, opened 2025-08-30
H-5154: HashQL: Ensure that types are contractive

indietyp:

Ensure that `type T = T` is invalid, such a type is not contractive, because during type checking, this recursive type does not make any progress during unfolding (as described in the Großer Beleg, Def. 27)

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/ast/src/lowering/type_extractor/error.rs
```diff
@@ -705,3 +712,50 @@ pub(crate) fn unused_generic_parameter(
...
+    diagnostic.add_help(Help::new(format!(
+        "Add structure around the recursive reference. Such as, but not limited to, a struct, \
+         tuple, or union with at least one non-recursive alternative:\n- `type {type_name} = \
+         {type_name} | Null`\n- `type {type_name} = (value: {type_name})`\n- `type {type_name} = \
+         ({type_name}, String)`"
+    )));
+
+    diagnostic.add_note(Note::new(
+        "Recursive types need some 'structure' between the type and its selft prevent  be \
+         'contractive' to ensure type checking terminates. This means every recursive reference \
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Good catch on the grammatical errors in the error message. Your suggestion looks good, but I'd like to make a minor adjustment to maintain the full meaning of the original message:

```suggestion
        "Recursive types need some 'structure' between the type and itself to be 'contractive' \
         and ensure type checking terminates. This means every recursive reference \
```

This preserves the important relationship between the structure requirement and the contractive property while fixing the grammatical issues.

libs/@local/hashql/ast/src/lowering/type_extractor/contractive.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+fn is_contractive_kind<'heap>(
+    env: &Environment<'heap>,
+    root: Interned<'heap, TypeKind<'heap>>,
+    kind: Interned<'heap, TypeKind<'heap>>,
+    visited: &mut FastHashSet<Interned<'heap, TypeKind<'heap>>>,
+) -> Result<(), SpanId> {
+    if !visited.insert(kind) {
+        return Ok(()); // We're inside of a recursive type that isn't *us*
```
hashdotai, line 34: The comment for the `is_contractive_kind` function is slightly misleading. It states "We're inside of a recursive type that isn't *us*" but it would be clearer to say "We've already visited this type in our current traversal path, so we can safely terminate this branch."

This better explains why we return `Ok(())` when we encounter a type we've already visited.

libs/@local/hashql/ast/src/lowering/type_extractor/contractive.rs
```diff
@@ -0,0 +1,118 @@
...
+    let result = match &*kind {
+        // Contractive if any of it's variants are contractive
+        TypeKind::Union(UnionType { variants })
+        | TypeKind::Intersection(IntersectionType { variants }) => {
+            let mut result = Ok(());
+
+            for &variant in variants {
+                match is_contractive_type(env, root, variant, visited) {
+                    Ok(()) => return Ok(()), // Found at least one contractive variant
+                    Err(span) => result = Err(span),
```
hashdotai, line 46: The logic here can be improved. Since you're looking for at least one contractive variant, you should return `Ok(())` immediately when you find a contractive variant, and only return `Err(span)` if none are contractive. However, your current implementation does return early with `Ok(())` in line 45, but the code in line 46 assigns the error to `result` without checking if we already had a valid contractive path.

```suggestion
                match is_contractive_type(env, root, variant, visited) {
                    Ok(()) => return Ok(()), // Found at least one contractive variant
                    Err(span) if result.is_ok() => result = Err(span), // Only set error if we haven't found a valid path yet
                    _ => () // Keep current error
                }
```

libs/@local/hashql/ast/tests/ui/lowering/type-extractor/definition/contractive-union.jsonc
```diff
@@ -0,0 +1,7 @@
+//@ run: pass
+//@ description: A union with a contractive type should be accepted
+["type",
+  "A",
+  ["|", "A", "Null"],
+  "_"
+]
```
hashdotai: Consider adding a test case for a nested recursive type that's contractive, such as `type A = { value: A }` or `type A = (A, Int)`. This would help verify that your implementation correctly handles more complex cases where the recursion is inside a structure.

libs/@local/hashql/ast/tests/ui/lowering/type-extractor/definition/non-contractive-type.jsonc
```diff
@@ -0,0 +1,8 @@
+//@ run: fail
+//@ description: Non-contractive types (pure self references) should be rejected
+["type",
+  "A",
+  "A",
+  //~^ ERROR Type `A` cannot reference itself directly
+  "_"
+]
```
hashdotai, line 9: Consider adding another test case for a non-contractive union type where all variants are non-contractive, like `type A = A | A`. This would test the edge case mentioned in the documentation comment (line 104 in contractive.rs).

libs/@local/hashql/ast/src/lowering/type_extractor/contractive.rs
indietyp: Wrong, it literally means what it says. We're in a recursive type, but that recursive type isn't us.

given: `type B = (value: B)`, `type A = A | B`
this case would be hit inside `B`, a type that we do not care about.

libs/@local/hashql/ast/src/lowering/type_extractor/contractive.rs
indietyp: because a valid contractive path would've returned early

libs/@local/hashql/ast/tests/ui/lowering/type-extractor/definition/non-contractive-type.jsonc
indietyp: done in [570ffaa](https://github.com/hashintel/hash/pull/7768/commits/570ffaaabc6f1505608a1a7791ebc6fd52b01b18)

libs/@local/hashql/ast/tests/ui/lowering/type-extractor/definition/contractive-union.jsonc
indietyp: done in [570ffaa](https://github.com/hashintel/hash/pull/7768/commits/570ffaaabc6f1505608a1a7791ebc6fd52b01b18)

TimDiekmann (approved):

PR #7770, opened 2025-08-30
H-5016: HashQL: Rename `::kernel::special_form::is` to `::kernel::special_form::as`

indietyp:

`is` is currently doing a type widening/assertion. This is not what was initially intended. `is` is intended for use as a type guard in the future; therefore, we should rename the function.

The name reflects that commonly used in other languages, such as Rust and corresponds as well to the first two letters of the operation it represents **as**cription (Großer Beleg, Def. 15).

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/ast/src/node/expr/as.rs
```diff
@@ -5,7 +5,7 @@ use crate::node::{id::NodeId, r#type::Type};
 
 /// A type assertion expression in the HashQL Abstract Syntax Tree.
 ///
-/// Represents an `is` expression that checks at compile time whether a value conforms to a
+/// Represents an `as` expression that checks at compile time whether a value conforms to a
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You make a great point. The documentation comment should indeed be updated to clarify that this is for type assertion/widening rather than type checking, which would align with the PR description stating that `is` was intended for future type guarding while `as` is for type assertion/widening (ascription).

A better documentation comment could explicitly mention:
- That this performs type ascription/widening rather than type checking
- That it helps enforce compile-time type safety
- That it's different from type guards (which would validate at runtime)

This would make the distinction clearer and prevent future confusion about the purpose of this operator.

hashdotai: The documentation comment should be updated to clarify that this is for type assertion/widening (ascription) rather than type checking. This would align with the PR's stated purpose of distinguishing `as` from the future `is` type guarding function.

Something like:
```
/// Represents an `as` expression that performs type ascription, asserting at compile time that a value
/// can be widened to a specified type. Unlike type guards which would validate at runtime,
/// type ascription helps enforce static type safety within the language.
```

libs/@local/hashql/hir/tests/ui/lower/inference/as.jsonc
```diff
@@ -1,3 +1,3 @@
 //@ run: pass
 //@ description: Verifies that the placeholder type '_' in a type assertion ('is' expression) is correctly inferred from the value's type.
```
hashdotai: For consistency, we should update this test description to replace 'is/2' with 'as/2'.

libs/@local/hashql/hir/tests/ui/lower/checking/type-assertion-unrelated.jsonc
```diff
@@ -1,7 +1,7 @@
 //@ run: fail
 //@ description: Ensures type checking fails when asserting a value is of an unrelated type (Integer as String).
 [
-  "is",
+  "as",
   { "#literal": 42 },
   //~^ ERROR This is of type `Integer`
```
hashdotai, line 6: Consider updating this error message to use 'type ascription' terminology rather than just 'is of type' to reinforce the semantics of the `as` operator.

indietyp: that's clear in the stderr

libs/@local/hashql/hir/tests/ui/lower/inference/as.jsonc
indietyp: done in [0208b70](https://github.com/hashintel/hash/pull/7770/commits/0208b70b1b701efc8ee38532d7d0632f906654ce)

libs/@local/hashql/ast/src/node/expr/as.rs
indietyp: done in [0208b70](https://github.com/hashintel/hash/pull/7770/commits/0208b70b1b701efc8ee38532d7d0632f906654ce)

TimDiekmann (approved):

PR #7773, opened 2025-08-30
H-5115: HashQL: allow for integer literals during field access

indietyp:

Element access (tuple indexing) was currently only possible using the string notation in j-expr, e.g. using `"value.field"`, this is insufficient in cases where the target is of a dynamic nature. This remedies it by allowing for `[".", "foo", {"#literal": 1}]`, e.g. access via a single integer literal that fits into usize.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
+++ b/libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
@@ -120,6 +120,21 @@ const DUPLICATE_CLOSURE_PARAMETER: TerminalDiagnosticCategory = TerminalDiagnost
     name: "Duplicate closure parameter",
 };
 
+const FIELD_LITERAL_TYPE_ANNOTATION: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    id: "field-literal-type-annotation",
+    name: "Field literal with type annotation",
+};
+
+const INVALID_FIELD_LITERAL_TYPE: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    id: "invalid-field-literal-type",
+    name: "Invalid field literal type",
+};
+
+const FIELD_INDEX_OUT_OF_BOUNDS: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    id: "field-index-out-of-bounds",
+    name: "Field index out of bounds",
+};
+
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum SpecialFormExpanderDiagnosticCategory {
     UnknownSpecialForm,
@@ -142,6 +157,9 @@ pub enum SpecialFormExpanderDiagnosticCategory {
     InvalidGenericArgumentType,
     DuplicateGenericConstraint,
     DuplicateClosureParameter,
+    FieldLiteralTypeAnnotation,
+    InvalidFieldLiteralType,
+    FieldIndexOutOfBounds,
 }
 
 impl DiagnosticCategory for SpecialFormExpanderDiagnosticCategory {
@@ -175,6 +193,9 @@ impl DiagnosticCategory for SpecialFormExpanderDiagnosticCategory {
             Self::InvalidGenericArgumentType => Some(&INVALID_GENERIC_ARGUMENT_TYPE),
             Self::DuplicateGenericConstraint => Some(&DUPLICATE_GENERIC_CONSTRAINT),
             Self::DuplicateClosureParameter => Some(&DUPLICATE_CLOSURE_PARAMETER),
+            Self::FieldLiteralTypeAnnotation => Some(&FIELD_LITERAL_TYPE_ANNOTATION),
+            Self::InvalidFieldLiteralType => Some(&INVALID_FIELD_LITERAL_TYPE),
+            Self::FieldIndexOutOfBounds => Some(&FIELD_INDEX_OUT_OF_BOUNDS),
         }
     }
 }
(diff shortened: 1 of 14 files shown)


libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
```diff
@@ -1110,3 +1131,75 @@ pub(crate) fn duplicate_closure_generic(
...
+        Severity::Error,
+    );
+
+    diagnostic.labels.push(Label::new(
+        span,
+        "Use a valid field index within usize bounds",
+    ));
+
+    diagnostic.add_help(Help::new(
+        "Field indices must be non-negative integers that fit within the bounds of usize. Very \
```
TimDiekmann, line 1193: `usize` might be correct, but it differs between platforms. Do we actually parse it as `usize` or as `u64`?

hashdotai: @TimDiekmann Good question about platform differences. Looking at the code, the implementation is using `integer.as_usize().is_none()` to check if the literal fits within `usize`, which means it's explicitly checking platform-specific size limits. The diagnostic message is correctly using `usize::BITS` to communicate the architecture-dependent limit. The literal parsing itself is platform-independent, but the validation is specifically targeting `usize` bounds as shown in the error handling.

libs/@local/hashql/syntax-jexpr/src/parser/string/expr.rs
```diff
@@ -39,14 +42,20 @@ where
...
-                Access::Field(Ident {
-                    span: context.span(range),
-
-                    value: context.heap.intern_symbol(digit),
-                    kind: IdentKind::Lexical, // Do we need to specify a different kind here?
+            digit1
+                .with_span()
+                .try_map(|(digits, range): (&str, _)| {
+                    // Ensure the value is within bounds
+                    digits.parse::<usize>().map(|_value| (digits, range))
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Good catch! You're right, the underscore prefix suggests an unused variable that we're intentionally ignoring, but we could just omit the variable name entirely since we don't need the parsed value itself. Your suggestion makes the code cleaner and more consistent.

libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
```diff
@@ -1110,3 +1131,75 @@ pub(crate) fn duplicate_closure_generic(
...
+         large numbers or negative numbers cannot be used as field indices.",
+    ));
+
+    diagnostic.add_note(Note::new(format!(
+        "Field indexing in HashQL uses zero-based indexing where the first field is at index 0, \
+         the second at index 1, and so on. The maximum valid index depends on the system's \
+         architecture, which is {}-bit.",
+        usize::BITS
+    )));
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're right that using `format!` might be unnecessary here. However, your suggested approach using string concatenation with `to_owned()` and `to_string()` creates multiple allocations. For a diagnostic message, I think the `format!` macro is actually more readable and maintainable, even if it's slightly less efficient. Since this is error handling code that isn't in a hot path, I'd prioritize readability over minimal optimizations here.

libs/@local/hashql/syntax-jexpr/src/parser/string/expr.rs
hashdotai: The `.try_map()` function is handling the validation for integer literals, but it's not clear what happens when a negative integer is provided. Since the function is parsing with `usize`, negative integers would fail, but there's no explicit test case for this. Consider adding a test for negative integer literals to verify this behavior.

libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
hashdotai, line 1193: For consistency with the rest of the codebase, consider adding a constant for this diagnostic message pattern instead of using string literals directly. This would make it easier to maintain and reuse these patterns.

libs/@local/hashql/ast/src/lowering/special_form_expander/mod.rs
```diff
@@ -452,6 +454,39 @@ impl<'heap> SpecialFormExpander<'heap> {
         Some(name)
     }
 
+    fn lower_argument_to_field(
+        &mut self,
+        mode: BindingMode,
```
hashdotai, line 467: The parameter name `mode: BindingMode` is carried over from `lower_argument_to_ident`, but it's not clear from the implementation if this parameter affects the behavior of `lower_argument_to_field`. If it's not used for any conditional logic, consider adding a comment explaining its purpose or how it might be used in future extensions.

libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
indietyp: We always parse it as usize, because we still use ordinary collections, if we were to parse it as u64, on 32-bit platforms we'd crash when we try to index (or just couldn't index)

TimDiekmann: I guess that's fine. If someone needs bigger index than 4B on a 32 bit system they have different issues 😅

TimDiekmann (approved):

libs/@local/hashql/syntax-jexpr/src/parser/string/expr.rs
indietyp: digits1 does not allow for negative numbers....

libs/@local/hashql/ast/src/lowering/special_form_expander/error.rs
indietyp: the constant is only used if the diagnostic is used in multiple places, it isnt

libs/@local/hashql/syntax-jexpr/src/parser/string/expr.rs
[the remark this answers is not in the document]
indietyp: fixed in [e7bab1f](https://github.com/hashintel/hash/pull/7773/commits/e7bab1fab0a1b293d772f299c479ffe67c9e6cb7)

libs/@local/hashql/ast/src/lowering/special_form_expander/mod.rs
indietyp: done in [f206202](https://github.com/hashintel/hash/pull/7773/commits/f20620281de79d4b1ae80a60a6c4cd18ea8456db)

PR #7774, opened 2025-08-31
H-4924: HashQL: Flatten nested `And` and `Or` operations during read compilation

indietyp:

Does what it says on the tin.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/eval/src/graph/read/sink.rs
```diff
@@ -0,0 +1,51 @@
...
+            Err(_) => FilterSink::Fail,
+        }
+    }
+
+    pub(crate) fn and(&mut self) -> FilterSink<'_, 'b, R> {
+        match self {
+            FilterSink::And(inner) => FilterSink::And(inner),
+            FilterSink::Or(inner) => {
+                let Filter::All(inner) = inner.push_mut(Filter::All(Vec::new())) else {
+                    unreachable!()
```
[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer Good suggestion! Adding descriptive messages to `unreachable!()` calls is definitely a best practice. It helps future developers understand why a code path is considered unreachable and provides more context if the assumption is ever violated in the future. This makes debugging much easier and speeds up issue resolution.

hashdotai, line 35: The unreachable!() macro should include a descriptive message explaining why this code path should never be reached. This helps with debugging if the assumption is violated.
```suggestion
                    unreachable!("Expected the last element to be Filter::All after push, but found a different variant")
```

libs/@local/hashql/eval/src/graph/read/sink.rs
```diff
@@ -0,0 +1,51 @@
...
+            }
+            FilterSink::Fail => FilterSink::Fail,
+        }
+    }
+
+    pub(crate) fn or(&mut self) -> FilterSink<'_, 'b, R> {
+        match self {
+            FilterSink::And(inner) => {
+                let Filter::Any(inner) = inner.push_mut(Filter::Any(Vec::new())) else {
+                    unreachable!()
```
hashdotai, line 52: Similarly, add a descriptive message here as well:
```suggestion
                    unreachable!("Expected the last element to be Filter::Any after push, but found a different variant")
```

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -80,20 +81,32 @@ impl<'env, 'heap: 'env> GraphReadCompiler<'env, 'heap> {
...
+                            self.compile_filter(context, left, &mut sink),
+                            self.compile_filter(context, right, &mut sink),
                         );
 
-                        return Ok(Filter::All(vec![left?, right?]));
+                        // We defer the error handling to make sure we gather every diagnostic
+                        left?;
+                        right?;
+
+                        return Ok(());
```
hashdotai, line 95: The code now defers error handling to collect all diagnostics before returning. Consider adding a comment explaining this pattern at the module level rather than repeating similar comments in multiple places.

libs/@local/hashql/eval/src/graph/read/sink.rs
```diff
@@ -0,0 +1,51 @@
+use hash_graph_store::filter::{Filter, QueryRecord};
+
+pub(crate) enum FilterSink<'a, 'b, R: QueryRecord> {
+    And(&'a mut Vec<Filter<'b, R>>),
+    Or(&'a mut Vec<Filter<'b, R>>),
+    Fail,
+}
+
+impl<'a, 'b, R: QueryRecord> FilterSink<'a, 'b, R> {
+    pub(crate) const fn from_result<E>(result: &'a mut Result<Vec<Filter<'b, R>>, E>) -> Self {
```
hashdotai, line 17: Consider making this function take ownership of the result to avoid lifetime complexities. If there's a reason for the current design, please add a comment explaining why.

libs/@local/hashql/eval/src/graph/read/sink.rs
```diff
@@ -0,0 +1,51 @@
...
+                };
+
+                FilterSink::Or(inner)
+            }
+            FilterSink::Or(inner) => FilterSink::Or(inner),
+            FilterSink::Fail => FilterSink::Fail,
+        }
+    }
+
+    pub(crate) fn push(&mut self, filter: Filter<'b, R>) {
```
hashdotai, line 66: Add a brief doc comment explaining the purpose of this method and how it interacts with the different sink variants, especially the `Fail` case.

libs/@local/hashql/eval/src/graph/read/sink.rs
indietyp: It's self-evident from the code snippet and the line just above it. Additionally, what would even need to happen for this to be violated?! I think this would just add additional visual noise.

libs/@local/hashql/eval/src/graph/read/sink.rs
indietyp: see comment above

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: it always did that, and I am putting the comment where it makes sense.

libs/@local/hashql/eval/src/graph/read/sink.rs
indietyp: done in [a754dd7](https://github.com/hashintel/hash/pull/7774/commits/a754dd7762844e01ee22dee3564baeece75f4db3)

TimDiekmann (approved):
This should give a decent performance boost, thank you!

PR #7775, opened 2025-08-31
H-5116: HashQL: Move `graph` module out of `core`

indietyp:

`graph` should not be part of `core`, but instead be a separate module, as described in Großer Beleg 5.6.1.3.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/core/src/module/std_lib/core/mod.rs
```diff
@@ -10,15 +10,14 @@ use crate::{
...
-pub(in crate::module::std_lib) mod graph;
 pub(in crate::module::std_lib) mod json;
 pub(in crate::module::std_lib) mod math;
 pub(in crate::module::std_lib) mod option;
 pub(in crate::module::std_lib) mod result;
 pub(in crate::module::std_lib) mod url;
 pub(in crate::module::std_lib) mod uuid;
 
-fn func<'heap>(
+pub(crate) fn func<'heap>(
```
hashdotai, line 20: The `func` function in the core module should remain private to the module. Let's maintain the original visibility restriction:
```suggestion
fn func<'heap>(
```

libs/@local/hashql/core/src/module/std_lib/graph/entity.rs
```diff
@@ -53,13 +53,7 @@ impl<'heap> StandardLibraryModule<'heap> for Entity {
...
         );
 
-        func(
-            lib,
-            &mut def,
-            "::core::graph::entity::is_of_type",
-            &[],
-            decl,
-        );
+        func(lib, &mut def, "::graph::entity::is_of_type", &[], decl);
```
hashdotai, line 56: This function can be simplified with a more concise style to match the pattern used in other similar functions:
```suggestion
func(lib, &mut def, "::graph::entity::is_of_type", &[], decl);
```

libs/@local/hashql/core/src/module/std_lib/graph/types/ontology/mod.rs
```diff
@@ -25,10 +25,9 @@ impl<'heap> StandardLibraryModule<'heap> for Ontology {
         let heap = lib.heap;
 
         // newtype OntologyTypeVersion = String;
-        let ontology_type_version_ty = lib.ty.opaque(
-            "::core::graph::ontology::OntologyTypeVersion",
-            lib.ty.string(),
-        );
+        let ontology_type_version_ty = lib
+            .ty
+            .opaque("::graph::ontology::OntologyTypeVersion", lib.ty.string());
```
hashdotai, line 30: Consider using `.ty.opaque()` directly without breaking the chain to match the style in the rest of the file:
```suggestion
let ontology_type_version_ty = lib.ty.opaque("::graph::ontology::OntologyTypeVersion", lib.ty.string());
```

TimDiekmann (approved):

PR #7779, opened 2025-09-03
BE-33: Streamline telemetry setup to allow flamegraph generation per benchmark

TimDiekmann:

The current telemetry components are quite hardcoded. This PR aims to solve this issue by making all components optional. This allows easy creation of the telemetry setup and distinguish between per-thread and global initialization. Also, it supports flamegraphs as they were used in the benchmarks.

As a result, only the `hash_telemetry` crate defines how the layers are composed and downstream apps have a simple way to set up the telemetry as needed.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/telemetry/src/lib.rs
```diff
@@ -50,14 +60,200 @@ pub struct TracingConfig {
...
+    ///
+    /// # Errors
+    ///
+    /// - [`InitTracingError`], if initializing the [`tracing_subscriber::Registry`] fails.
+    pub fn init(self) -> Result<impl Drop, Report<InitTracingError>> {
+        #[expect(unused)]
+        struct DropGuard<T>(T, DefaultGuard);
+
+        #[expect(clippy::empty_drop)]
+        impl<T> Drop for DropGuard<T> {
```
indietyp, line 200: Why the empty drop? Couldn't you just force `DropGuard<T: Drop>` in the type definition? It's not like the type is used anywhere else.

tests/graph/benches/flamegraph.svg
indietyp, line 1: is this supposed to be here?

indietyp (changes requested):
Great improvement, thank you! just one minor comment about a file that seems to be out of place, otherwise just minor nit/question.

libs/@local/telemetry/src/lib.rs
TimDiekmann: I just need a type which `impl Drop` so I can return that as a guard. But `(T, U)` does not implement `Drop`, so I created a wrapper. It's just used as a marker.

tests/graph/benches/flamegraph.svg
TimDiekmann: Thanks, removed!

libs/@local/telemetry/src/lib.rs
indietyp: Yes, I get that. I just found the empty drop implementation a bit weird, considering that `Drop` is implemented automatically if one of the fields is `Drop`.

TimDiekmann: No, `Drop` is not implemented automatically.

indietyp: I mixed up two things, sorry!

indietyp (approved):

libs/@local/telemetry/src/lib.rs
TimDiekmann: No worries 🙂

PR #7781, opened 2025-09-05
FE-9: Improvements to pre/post user change hooks

CiaranMn:

Makes a few changes to the validation hooks the Node API uses on pre/post user save. This affects instances where someone attempts to modify their user record via the regular entity slideover (or directly in the API), rather than the dedicated profile page editor.

1. Correctly compare emails to check for no changes (was assuming they had changed in all cases, preventing making other changes)
2. Prevent editing feature flags
3. Check if provided email is actually different before attempting update in MailChimp/Kratos (updating makes no data changes but is pointless)

This PR:

The changes in this PR:

The changes in this PR:


vilkinsons (dismissed):

TimDiekmann (dismissed):
Ping me when you need a re-approval due to the linting error

PR #7782, opened 2025-09-05
BE-36: HashQL: Update `README` contents of HashQL

indietyp:

This updates the root README with condensed information from the Großer Beleg. This has primarily been generated through GPT-5 Pro, with minor manual touch-up to experiment with the viability of such a setup.

> Tagging Dei as it is README

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/README.md
+++ b/libs/@local/hashql/README.md
@@ -160,7 +160,7 @@ Special forms are **compiler-recognised macros** that lower to dedicated AST nod
 
 Conditional expression: executes `then` if `test` is `true`; otherwise executes `else`. Two‑argument `if` produces `Option<T>`; the three‑argument form is total and returns `T`.
 
-```
+```text
 if/2<T>(test: Expr<Boolean>, then: Expr<T>) -> Option<T>
 if/3<T>(test: Expr<Boolean>, then: Expr<T>, else: Expr<T>) -> T
 ```
@@ -169,7 +169,7 @@ if/3<T>(test: Expr<Boolean>, then: Expr<T>, else: Expr<T>) -> T
 
 Type ascription (assertion). Only **upcasts** are allowed; downcasts would require runtime checks. Operationally inert and erased after type checking.
 
-```
+```text
 as/2<T, U>(value: Expr<T>, type: Type<U>) -> Expr<U>   where T <: U
 ```
 
@@ -177,7 +177,7 @@ as/2<T, U>(value: Expr<T>, type: Type<U>) -> Expr<U>   where T <: U
 
 Lexically binds `value` to `name` within `body`. Bindings participate in standard shadowing.
 
-```
+```text
 let/3<T, U>(name: Ident, value: Expr<T>, body: Expr<U>) -> Expr<U>
 let/4<T, U, V>(name: Ident, type: Type<T>, value: Expr<U>, body: Expr<V>) -> Expr<V>  where U <: T
 ```
@@ -186,7 +186,7 @@ let/4<T, U, V>(name: Ident, type: Type<T>, value: Expr<U>, body: Expr<V>) -> Exp
 
 Introduces a **type alias** (possibly generic with bounds) visible within `body`.
 
-```
+```text
 type/3<T, U>(name: GenericIdent, value: Type<T>, body: Expr<U>) -> Expr<U>
 ```
 
@@ -194,7 +194,7 @@ type/3<T, U>(name: GenericIdent, value: Type<T>, body: Expr<U>) -> Expr<U>
 
 Declares a **nominal wrapper** over a structural type (opaque identity). Useful for domain types (`Url`, `Uuid`) and for tagging constructors used in sum types.
 
-```
+```text
 newtype/3<T, U>(name: GenericIdent, value: Type<T>, body: Expr<U>) -> Expr<U>
 ```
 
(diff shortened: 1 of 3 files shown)


vilkinsons (approved):
Very cool; thank you!

libs/@local/hashql/README.md
```diff
@@ -1,51 +1,357 @@
...
-- Variables: Variables are references to values that are defined elsewhere, they are defined using:
-  - `"symbol"` (any string is a reference to a variable)
-  - `{"var": name}` (this is a more explicit way of defining a variable)
-- Signature: Signature is a special type, which is parsed independently of variables, they have a specific form (specified in the CST crate) and are defined using:
-  - `{"sig": ...}`
-  - `"sig"`
+HashQL's compiler is organized into four phases. Only the frontend **pushes** into its successor; all other phases **pull** from their single predecessor. Each phase is side‑effect‑free and delimited by diagnostic boundaries.
 
-### Future Expansions
+```mermaid
```
TimDiekmann, line 18: Is it possible to align this vertically? It becomes very small horizontally.

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
TimDiekmann, line 1: - Does `type_atom` always require a trailing comma after the first `type`?
- You use `generics` for both, arguments and constraints. that seems to be very different. Shouldn't `generic argument` not be `<ident>`? as in `impl<T> Foo<T>` rather than `impl<T> Foo<Debug>`?
- You define `expr` before you define `underscore` and `path`

libs/@local/hashql/core/docs/identifier-grammar.svg
TimDiekmann, line 1: `baseUrl` does not have constraints such as `http`, it could be anything, such as `ftp://`. The only "real" constraint is that it can be a base (= trailing slash), otherwise, any URL is valid.

indietyp: Yes, but it is the current constraint of the parser only to allow for `http(s)://`, so it simply reflects what the parser currently allows. This is (by design) more strict than it could be to allow for future extension, while allowing for existing parsers.

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
indietyp: > Does type_atom always require a trailing comma after the first type?

If it is a tuple, yes. This mirrors the behaviour of Rust

> You use generics for both, arguments and constraints. that seems to be very different. Shouldn't generic argument not be <ident>? as in impl<T> Foo<T> rather than impl<T> Foo<Debug>

The parser does not differentiate between arguments and constraints; this is done in the AST. The lexer itself does not care which one you choose.

> You define expr before you define underscore and path

I chose to make it so that the top is the "top level" and then we go down, which mirrors how JSON defines it (https://www.json.org/json-en.html)

libs/@local/hashql/core/docs/identifier-grammar.svg
TimDiekmann: Fair enough!

libs/@local/hashql/README.md
indietyp: I had it both in vertical and horizontal orientations, but found that when used vertically, it occupied too much space. Happy to move it vertically as a drive-by.

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
TimDiekmann: > > You use generics for both, arguments and constraints. that seems to be very different. Shouldn't generic argument not be ? as in impl Foo rather than impl Foo
> 
> The parser does not differentiate between arguments and constraints; this is done in the AST. The lexer itself does not care which one you choose.

Right, but shouldn't it be `<ident: type> | <ident>` instead of `<ident: type> | <type>`? I might confuse two things here, though

libs/@local/hashql/README.md
TimDiekmann: Is a hybrid possible? So we group horizontally but have the different sections vertical?

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
indietyp: I thought that at first as well, but no, because `math::add<Graph<T>>` should still be possible, so the inner type would need to be a type.

libs/@local/hashql/README.md
TimDiekmann: I don't think there is something like "too much space" in a diagram in the readme which gives the overview. Usually, this is the most interesting part. It's super helpful, but when I open the readme currently I cannot read it.

indietyp: I'll fuzz around with it in the PR I am currently preparing. I just checked - I can change the direction in a subgraph.

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
TimDiekmann: Ah, it's on call-site. Don't we also have `<ident>` as standalone in definitions? In Rust: `fn foo<T>()` but never `fn foo<Graph<T>>()`.

libs/@local/hashql/README.md
TimDiekmann: Cool, thank you!

libs/@local/hashql/syntax-jexpr/docs/embedded-dsl.svg
indietyp: We have, `type` can also be just an ident. The differentiation you speak of happens in the AST (during special form macro expansion - so very early on). We first parse in valid positions both, and then error out (and discard to continue AST compilation) any constraint that is invalid. We basically check for `fn`, `type`, etc. definitions:<br><br>does this look like a constraint to us?

1. it's of form: `<ident>: <type>` so yes
2. it's a `<type>`
   1. is it just an identifier? (non-absolute path with a singular identifier?
      1. Yes? Convert it to a generic constraint with no bound
      2. No? Error out
   2. No? Error out

In places where we expect only generic arguments (e.g. in paths) we reject all `<ident>: <type>` forms.

indietyp: You can see this here in action: https://github.com/hashintel/hash/blob/bf98eb4fde65969ab171e98b224d4f4b38246887/libs/%40local/hashql/ast/src/lowering/special_form_expander/mod.rs#L516

PR #7783, opened 2025-09-06
BE-38: HashQL: `meet` and `join` should consider subtyping

indietyp:

`meet` and `join` currently indiscriminately choose `lhs` when recursion happens; instead, we should return the type depending on the subtype relationship. If `lhs <: rhs`, we return `lhs`, otherwise we should return `rhs` (on meet — reverse on join).

This has the downside that we traverse a recursive type essentially four times:

1. meet/join
2. Determine if lhs and rhs are recursive
3. subtype relationship
   1. Determine if `lhs` and `rhs` are recursive

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/core/src/type/environment/lattice.rs
+++ b/libs/@local/hashql/core/src/type/environment/lattice.rs
@@ -149,9 +149,15 @@ impl<'env, 'heap> LatticeEnvironment<'env, 'heap> {
         self.diagnostics
             .push(circular_type_reference(self.source, lhs, rhs));
 
-        if cycle.should_discharge() {
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
+            return rhs.id;
+        }
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, rhs.id, lhs.id) {
             return lhs.id;
         }
+
+        // If we're at this point and should still discharge, it means that A and B are unrelated to
+        // each other, therefore create a union type.
 
         // If they aren't in a subtyping relationship, create a union type
         let kind = self.environment.intern_kind(TypeKind::Union(UnionType {
@@ -257,9 +263,16 @@ impl<'env, 'heap> LatticeEnvironment<'env, 'heap> {
         self.diagnostics
             .push(circular_type_reference(self.source, lhs, rhs));
 
-        if cycle.should_discharge() {
+        // Check the subtyping relationship
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
             return lhs.id;
         }
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, rhs.id, lhs.id) {
+            return rhs.id;
+        }
+
+        // If we're at this point and should still discharge, it means that A and B are unrelated to
+        // each other, therefore create an intersection type.
 
         // If they aren't in a subtyping relationship, create an intersection type
         let kind = self
--- a/libs/@local/hashql/core/src/type/kind/tests/tuple.rs
+++ b/libs/@local/hashql/core/src/type/kind/tests/tuple.rs
@@ -21,7 +21,7 @@ use crate::{
             intersection::IntersectionType,
             primitive::PrimitiveType,
             test::{assert_equiv, generic, intersection, primitive, tuple, union},
-            tests::{assert_equivalent, assert_is_subtype, assert_join, assert_meet},
+            tests::{assert_is_subtype, assert_join, assert_meet},
             union::UnionType,
         },
         lattice::{Lattice as _, Projection, Subscript, test::assert_lattice_laws},
--- a/libs/@local/hashql/core/src/type/kind/tests/tuple.rs
+++ b/libs/@local/hashql/core/src/type/kind/tests/tuple.rs
@@ -7,6 +7,7 @@ use crate::{
     symbol::Ident,
     r#type::{
         PartialType,
+        builder::lazy,
         environment::{
             AnalysisEnvironment, Environment, InferenceEnvironment, LatticeEnvironment,
             SimplifyEnvironment, Variance, instantiate::InstantiateEnvironment,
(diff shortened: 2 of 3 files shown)


libs/@local/hashql/core/src/type/kind/tests/tuple.rs
```diff
@@ -1137,3 +1139,57 @@ fn subscript() {
...
+}
+
+#[test]
+fn join_recursive_tuple_unrelated() {
+    scaffold!(heap, env, builder, [lattice: lattice]);
+
+    let lhs = builder.tuple(lazy(|this, builder| [builder.integer(), this.value()]));
+    let rhs = builder.tuple(lazy(|this, builder| [builder.string(), this.value()]));
+
+    // The type itself is a bit larger, than one might light (this is of the recursive property of
```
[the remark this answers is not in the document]
indietyp: done in [65e4adb](https://github.com/hashintel/hash/pull/7783/commits/65e4adb574c9a198664261ea7c983590f3eb1a8e)

libs/@local/hashql/core/src/type/environment/lattice.rs
```diff
@@ -257,9 +263,16 @@ impl<'env, 'heap> LatticeEnvironment<'env, 'heap> {
...
-        if cycle.should_discharge() {
+        // Check the subtyping relationship
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
             return lhs.id;
         }
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, rhs.id, lhs.id) {
+            return rhs.id;
+        }
+
+        // If we're at this point and should still discharge, it means that A and B are unrelated to
```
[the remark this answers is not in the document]
indietyp: done in [65e4adb](https://github.com/hashintel/hash/pull/7783/commits/65e4adb574c9a198664261ea7c983590f3eb1a8e)

libs/@local/hashql/core/src/type/environment/lattice.rs
```diff
@@ -149,10 +149,16 @@ impl<'env, 'heap> LatticeEnvironment<'env, 'heap> {
...
 
-        if cycle.should_discharge() {
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
+            return rhs.id;
+        }
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, rhs.id, lhs.id) {
             return lhs.id;
         }
 
+        // If we're at this point and should still discharge, it means that A and B are unrelated to
```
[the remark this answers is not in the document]
indietyp: done in [65e4adb](https://github.com/hashintel/hash/pull/7783/commits/65e4adb574c9a198664261ea7c983590f3eb1a8e)

libs/@local/hashql/core/src/type/environment/lattice.rs
```diff
@@ -149,10 +149,16 @@ impl<'env, 'heap> LatticeEnvironment<'env, 'heap> {
         self.diagnostics
             .push(circular_type_reference(self.source, lhs, rhs));
 
-        if cycle.should_discharge() {
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
```
hashdotai, line 152: The condition logic here is improved by checking subtype relationships, but the error message from `circular_type_reference` might now be misleading since you're properly handling the subtype relationships. Consider updating the diagnostic to reflect that you're finding a solution through subtyping, not just warning about circularity.

libs/@local/hashql/core/src/type/environment/lattice.rs
```diff
@@ -257,9 +263,16 @@
         self.diagnostics
             .push(circular_type_reference(self.source, lhs, rhs));
 
-        if cycle.should_discharge() {
+        // Check the subtyping relationship
+        if cycle.should_discharge() && self.is_subtype_of(Variance::Covariant, lhs.id, rhs.id) {
```
hashdotai, line 267: It might be worth adding a comment explaining the logic reversal between meet and join operations with respect to subtyping. For `meet`, you return `rhs` when `lhs <: rhs`, but for `join`, you return `lhs` when `lhs <: rhs`. This is correct but could benefit from a brief explanation for future developers.

libs/@local/hashql/core/src/type/kind/tests/tuple.rs
```diff
@@ -1137,3 +1139,58 @@ fn subscript() {
...
+
+#[test]
+fn meet_recursive_tuple() {
+    scaffold!(heap, env, builder, [lattice: lattice]);
+
+    let lhs = builder.tuple(lazy(|this, builder| [builder.integer(), this.value()]));
+    let rhs = builder.tuple(lazy(|this, builder| [builder.number(), this.value()]));
+
+    assert_meet(&mut lattice, lhs, rhs, &[lhs]);
+    assert_meet(&mut lattice, rhs, lhs, &[lhs]);
```
hashdotai, line 1151: This test is well-constructed to validate the new subtyping logic in meet operations, but it might be helpful to add a small comment explaining why `lhs` is expected to be the result (because Integer is a subtype of Number).

libs/@local/hashql/core/src/type/kind/tests/tuple.rs
```diff
@@ -1137,3 +1139,58 @@
...
+
+#[test]
+fn join_recursive_tuple() {
+    scaffold!(heap, env, builder, [lattice: lattice]);
+
+    let lhs = builder.tuple(lazy(|this, builder| [builder.integer(), this.value()]));
+    let rhs = builder.tuple(lazy(|this, builder| [builder.number(), this.value()]));
+
+    assert_join(&mut lattice, lhs, rhs, &[rhs]);
+    assert_join(&mut lattice, rhs, lhs, &[rhs]);
```
hashdotai, line 1175: Similar to the meet test, consider adding a brief comment explaining why `rhs` is the expected result here (because Number is a supertype of Integer).

libs/@local/hashql/core/src/type/kind/tests/tuple.rs
```diff
@@ -1137,3 +1139,58 @@
...
+#[test]
+fn join_recursive_tuple_unrelated() {
+    scaffold!(heap, env, builder, [lattice: lattice]);
+
+    let lhs = builder.tuple(lazy(|this, builder| [builder.integer(), this.value()]));
+    let rhs = builder.tuple(lazy(|this, builder| [builder.string(), this.value()]));
+
+    // The type itself is a bit larger than one might like (this is of the recursive property of
+    // the match), but the type approximation is still valid, as proven below and is actually
+    // *smaller* than the naive approximation.
```
hashdotai, line 1187: The explanation about type approximation is very helpful. Consider adding a bit more detail about why the result is smaller than the naive approximation - this would make the test even more educational.

TimDiekmann (approved):

PR #7789, opened 2025-09-10
H-5331: Support CPU and Wall time profiling in benchmarks

TimDiekmann:

This implements CPU time profiling (using pprof) and Wall time profiling (using tracing spans). It can be hooked into the profiling by enabling them in the telemetry config.

## 🔍 What does this change?

- Implement Wall time profiling by introducing a layer. This creates a new layer instead of utilizing `tracing-flame` for three reasons:
	- We want to profile the idle time as well. This is important to also see the actual timings for asynchronous calls such as DB calls, which is one of the main reasons why we want this in the first place. Also, this aligns with the behavior of `tracing-opentelemetry`, which means that the resulting flame graph will be similar to the traces
	- `tracing-flame` only supports a `io::Write` interface but we want to collect them asynchronous in a dedicated thread using channels. While we could use `write` and `flush` to send messages, utilizing channels is easier, in particular because it allows us to use dedicated errors
	- `tracing-flame` creates a folded row for each enter/exit of a span resulting in a huge amount of data. It is enough to only generate a single row for flame graphs (we would need that for flame charts, but for that functionality, we have proper tracing)
- Implement CPU time profiling by using Pyroscopes' `pprof` implementation

Both implementation can be separately disabled.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This is currently disabled in production because:
- Wall time profiling results in timeouts in the app (probably related to the `std::thread` approach over `tokio::task`)
- CPU profiling does not start. Also, proper CPU profiling in the graph would need tags per endpoint which we don't have, yet


TimDiekmann:
Benchmarks don't finish, there is something going on which is not reproducible locally. Converting to draft.

vilkinsons:
How might the CodSpeed PR (https://github.com/hashintel/hash/pull/8204) interact with this open PR? @TimDiekmann @indietyp

PR #7790, opened 2025-09-11
H-5335: Bump transitive Cargo dependencies and Rust toolchain to 2025-09-10

TimDiekmann:

Recurring task on bumping transitive dependencies. Also bumps the toolchain as rust-analyzer, our renovate source, currently has a release freeze.


TimDiekmann:
Spurious unit test failure.

PR #7793, opened 2025-09-11
BE-46: HashQL: Register all variables mentioned during inference

indietyp:

Traverse a type to register any encountered generic variables. This is required so that in the case we have variables that are just not hit, we still record a diagnostic telling us to do so.

As a drive-by, it fixes the README diagram to be more readable.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/core/src/type/environment/infer.rs
+++ b/libs/@local/hashql/core/src/type/environment/infer.rs
@@ -170,8 +170,6 @@ impl<'env, 'heap> InferenceEnvironment<'env, 'heap> {
     }
 
     pub fn collect_constraints(&mut self, variance: Variance, subtype: TypeId, supertype: TypeId) {
-        // TODO: we need to add a way that "tops" out, meaning that is called from outside, and one
-        // from inside, outside also registers all variables.
         let (_guard, variance_flow) = self.variance.transition(variance);
 
         #[expect(
--- a/libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.jsonc
+++ b/libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.jsonc
@@ -1,5 +1,5 @@
 //@ run: fail
-//@ description: Collect variables that aren't explicitely mentioned
+//@ description: Collect variables that aren't explicitly mentioned
 [
   "let",
   "foo",


libs/@local/hashql/core/src/type/environment/infer.rs
```diff
@@ -156,6 +166,8 @@ impl<'env, 'heap> InferenceEnvironment<'env, 'heap> {
     }
 
     pub fn collect_constraints(&mut self, variance: Variance, subtype: TypeId, supertype: TypeId) {
+        // TODO: we need to add a way that "tops" out, meaning that is called from outside, and one
+        // from inside, outside also registers all variables.
```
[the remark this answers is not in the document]
indietyp: was a remnant from a previous integration attempt

libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.jsonc
```diff
@@ -0,0 +1,10 @@
+//@ run: fail
+//@ description: Collect variables that aren't explicitely mentioned
```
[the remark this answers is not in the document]
indietyp: fixed in [c6b897e](https://github.com/hashintel/hash/pull/7793/commits/c6b897e1c23e44a6ea5d80dc28c41055f2d96187)

libs/@local/hashql/core/src/type/environment/infer.rs
```diff
@@ -149,6 +149,16 @@ impl<'env, 'heap> InferenceEnvironment<'env, 'heap> {
         variable
     }
 
+    pub fn add_variables(&mut self, variables: impl IntoIterator<Item = Variable>) {
```
hashdotai, line 152: This method should include a clear comment explaining its purpose. From the context, it seems this is used to register variables that might not be used in a unification constraint, but it's not clear from the implementation alone.

libs/@local/hashql/hir/src/lower/inference.rs
```diff
@@ -123,6 +129,9 @@ impl<'env, 'heap> TypeInference<'env, 'heap> {
         TypeInferenceResidual<'heap>,
         Vec<TypeCheckDiagnostic>,
     ) {
+        let variables = self.collector.take_variables();
```
hashdotai, line 132: It looks like you're accessing the collector as a mutable reference, but then using `take_variables()` which consumes the collector. This means you can no longer use the collector after this point. It might be better to either:

1. Clone the variables instead of taking them
2. Add a method to get a reference to the variables
3. Document this behavior clearly if it's intentional

libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.jsonc
```diff
@@ -0,0 +1,10 @@
+//@ run: fail
+//@ description: Collect variables that aren't explicitly mentioned
```
hashdotai, line 2: Consider adding a test for the edge case where we have a type with variables that don't participate in any constraints. This test is a good start, but additional test cases with more complex scenarios would help ensure the robustness of this feature.

libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.stderr
```diff
@@ -0,0 +1,14 @@
+[type-check::unconstrained-type-variable] Error: Unconstrained type variable
+   ╭─[ <unknown>:6:19 ]
+   │
+ 6 │   "Dict<Integer | _, String>",
+   │                   ┬  
+   │                   ╰── Cannot infer type for this variable - no usage constraints available
```
hashdotai, line 6: You're updating the error message here, but it would be helpful to include a more specific diagnostic for cases where variables are mentioned but not constrained. Something like "Variable mentioned in type but not constrained by usage" would be more informative than the generic "Cannot infer type for this variable".

libs/@local/hashql/hir/src/lower/inference.rs
indietyp: no.

libs/@local/hashql/core/src/type/environment/infer.rs
indietyp: There's a comment in the implementation, also tracked in https://linear.app/hash/issue/BE-64/hashql-create-docs-for-publicly-accessible-environment-methods

libs/@local/hashql/hir/tests/ui/lower/inference/collect-variable.stderr
indietyp: ?

PR #7795, opened 2025-09-14
BE-45: HashQL: Expose HashQL in the Graph API

indietyp:

Adds HashQL to the entities endpoint via opt-in, to allow for request deserialization needed to change a bit to accommodate the use of `RawValue`.

This PR:

The changes in this PR:

The changes in this PR:

## 📹 Demo

```json
{
  "query": ["::graph::tail::collect",
    ["::graph::body::filter",
      ["::graph::head::entities", ["::graph::tmp::decision_time_now"]],
      ["fn", { "#tuple": [] }, { "#struct": { "vertex": "_" } }, "_",
        ["==",
          "vertex.id.entity_id.entity_uuid",
          ["::graph::types::knowledge::entity::EntityUuid",
            ["::core::uuid::Uuid", { "#literal": "2d3c9647-2d73-4374-a6bf-bda052e4f723" }]
          ]
        ]
      ]
    ]
  ],
  "sortingPaths": [
    {
      "path": [
        "label"
      ],
      "ordering": "ascending"
    }
  ],
  "temporalAxes": {
    "pinned": {
      "axis": "transactionTime",
      "timestamp": null
    },
    "variable": {
      "axis": "decisionTime",
      "interval": {
        "start": null,
        "end": null
      }
    }
  },
  "includeDrafts": false
}
```


libs/@local/hashql/core/src/heap.rs
```diff
@@ -84,27 +84,65 @@ pub struct Heap {
...
+    ///
+    /// # Usage Requirements
+    ///
+    /// The caller must call [`Self::prime_unchecked`] exactly once before using the heap
+    /// for any allocations or symbol interning operations. Using an unprimed heap may
+    /// result in missing essential symbols that other parts of the system expect to exist.
     #[must_use]
-    pub fn new() -> Self {
-        let this = Self {
+    pub fn empty_unchecked() -> Self {
```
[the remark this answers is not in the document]
indietyp: it's called `_unchecked` for a reason

libs/@local/graph/api/src/rest/entity_query_request.rs
```diff
@@ -0,0 +1,723 @@
...
+impl<'q> EntityQuery<'q> {
+    #[expect(
+        clippy::unnecessary_wraps,
+        clippy::panic_in_result_fn,
+        reason = "https://linear.app/hash/issue/BE-39/hashql-handle-errors-in-the-graph-api-properly"
+    )]
+    fn compile_query<'h>(
+        heap: &'h Heap,
+        query: &serde_json::value::RawValue,
+    ) -> Result<Filter<'h, Entity>, !> {
```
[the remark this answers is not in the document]
indietyp: yes, see todo

libs/@local/graph/api/src/rest/entity_query_request.rs
```diff
@@ -0,0 +1,723 @@
...
+        let spans = Arc::new(SpanStorage::new());
+
+        // Parse the query
+        let parser = hashql_syntax_jexpr::Parser::new(heap, Arc::clone(&spans));
+        let mut ast = parser.parse_expr(query.get().as_bytes()).expect(
+            "https://linear.app/hash/issue/BE-39/hashql-handle-errors-in-the-graph-api-properly",
```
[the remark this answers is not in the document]
indietyp: yes, see todo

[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer You're absolutely right. The current implementation with multiple `expect()` calls pointing to the same issue ticket (BE-39) is a temporary solution. This is a known limitation that we're planning to address properly in the follow-up ticket (https://linear.app/hash/issue/BE-39/hashql-handle-errors-in-the-graph-api-properly) which is specifically dedicated to improving error handling for HashQL in the Graph API.

Currently, we're just exposing the functionality with a basic implementation, and the proper error handling architecture will be implemented as part of that follow-up work. We'll replace these `expect()` calls with proper error types, propagation, and user-friendly error messages at that time.

indietyp:
I was unaware that ada-url has a C++ dependency. I am going to remove it in favor of servo/url.

The initial reason for choosing ada-irl was its speed as well as ability to just verify if a URL is valid without requiring parsing, which made it ideal for syntax-jexpr.

libs/@local/graph/api/src/rest/entity.rs
```diff
@@ -586,19 +437,38 @@ where
...
+        // The heap is going to be used in the compilation of the query and therefore needs to be
+        // primed.
+        // Doing this in a separate step allows us to be allocation free when not using HashQL
+        // queries.
+        heap.prime_unchecked();
+    }
+
+    let filter = query.compile(&heap).expect(
+        "https://linear.app/hash/issue/BE-39/hashql-handle-errors-in-the-graph-api-properly",
+    );
```
TimDiekmann, line 466: Can't you convert the error to a report for now?

libs/@local/graph/api/src/rest/entity.rs
```diff
@@ -44,7 +39,9 @@ use hash_graph_types::{
     },
 };
 use hash_temporal_client::TemporalClient;
-use serde::{Deserialize, Serialize};
+use hashql_core::heap::Heap;
+use serde::{Deserialize as _, Serialize};
+use serde_json::value::RawValue;
```
TimDiekmann: We typically alias `Value` to `JsonValue`. We should do the same for `RawValue` I think.

libs/@local/hashql/core/src/heap.rs
```diff
@@ -84,27 +84,65 @@ pub struct Heap {
...
+    ///
+    /// This is automatically called by [`Self::new`] and [`Self::reset`], so manual
+    /// invocation is only necessary when using the [`Self::empty_unchecked`] constructor.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the heap is already primed.
+    pub fn prime_unchecked(&self) {
+        self.prime_symbols();
+    }
```
TimDiekmann: I'm not a fan of `_unchecked` here, `unchecked` typically requires an `unsafe` block, which is not needed here. The more common approach would be to have `prime` and `try_prime`. As alternative, we could omit the `try_prime` if we don't need it. Also, what is the difference between `prime_unchecked` and `prime_symbols`? Can't we simply expose `prime_symbols` and document the panic?

libs/@local/hashql/core/src/heap.rs
TimDiekmann: Same reasoning as above, I think `empty` or `new_unprimed` would be a better name.

Cargo.lock
TimDiekmann, line 1: Were the downgrades by accident?

libs/@local/hashql/core/src/heap.rs
indietyp: `_unchecked` means that you need to check the invariants yourself; this is the reason this is unchecked. We have this quite a bit in the codebase, from `Interval` to `Valid<T>`.

This is unchecked because we don't check; the caller must verify that they only call it once. The reason `empty_unchecked` is also unchecked is that the caller must verify they call it; if they do not (and that is why it is `_unchecked`), it will lead to comparisons failing down the road, therefore leading to issues. (especially once https://linear.app/hash/issue/BE-60/hashql-use-symbol-sym-constants-instead-of-string-literals lands)

We make the distinction here very explicit.

I also don't want to expose any internal invariants here (in the future we might/will prime more things here) or cause panic. For the end user the only thing they are "allowed" to see is that: they need to call this once if they use it with `empty_unchecked`.

Cargo.lock
indietyp: Yes, seems to stem from the fact that I rebased the PR.

libs/@local/graph/api/src/rest/entity.rs
indietyp: I already have a PR (almost) ready. Additionally, most of these return a `Vec<Diagnostic>`, so conversion isn't really straightforward.

libs/@local/hashql/core/src/heap.rs
TimDiekmann: If we keep `_unchecked` (and I hear your reasoning, though, I still don't like it), we should have a checked variant

```rust
pub fn prime(&self) {
	if !self.is_primed() {
		self.prime_unchecked();
	}
}
```

and add a `debug_assert!` to `prime_unchecked`. 

In `Heap` it's not immediately clear from the API that `unchecked` means "We ensured it was not primed". For `Valid` and for `Interval` this is quite clear. For `empty_unchecked` (comment below) it's not clear what should have been checked. What is checked in that context?

libs/@local/graph/api/src/rest/entity.rs
TimDiekmann: Fair enough, thanks!

libs/@local/graph/api/src/rest/entity.rs
indietyp: addressed in [3676c2a](https://github.com/hashintel/hash/pull/7795/commits/3676c2a2120d5cfa8929c7a334af8ee5b92c38c5)

libs/@local/hashql/core/src/heap.rs
indietyp: discussed offline, culminating in [622ea22](https://github.com/hashintel/hash/pull/7795/commits/622ea22de0d42f999b73d53809cd118a77e6bc6b)

Cargo.lock
indietyp: fixed in [b0a77c5](https://github.com/hashintel/hash/pull/7795/commits/b0a77c56d01c07dfcd2c8ee5a7a66ab5b725c03a)

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -170,7 +171,9 @@ export const getEntities: ImpureGraphFunction<
   GetEntitiesRequest & { temporalClient?: TemporalClient },
   Promise<HashEntity[]>
 > = async ({ graphApi, temporalClient }, { actorId }, params) => {
-  await rewriteSemanticFilter(params.filter, temporalClient);
+  if (Predicate.hasProperty(params, "filter")) {
+    await rewriteSemanticFilter(params.filter, temporalClient);
+  }
```
TimDiekmann: I think we need to move this functionality to the Graph or deal otherwise with semantic searches. Could you add a TODO with [this linear ticket](https://linear.app/hash/issue/BE-108/consider-moving-semantic-filter-rewriting-to-the-graph)?

apps/hash-api/src/graph/knowledge/primitive/entity.ts
```diff
@@ -203,7 +206,9 @@ export const getEntitySubgraphResponse: ImpureGraphFunction<
     > & { subgraph: Subgraph<EntityRootType<HashEntity>> }
   >
 > = async ({ graphApi, temporalClient }, { actorId }, params) => {
-  await rewriteSemanticFilter(params.filter, temporalClient);
+  if (Predicate.hasProperty(params, "filter")) {
```
TimDiekmann, line 210: Same here

indietyp: done in [19d3d18](https://github.com/hashintel/hash/pull/7795/commits/19d3d18709c171a558e6b46916de774356270641)

apps/hash-api/src/graph/knowledge/primitive/entity.ts
indietyp: done in [19d3d18](https://github.com/hashintel/hash/pull/7795/commits/19d3d18709c171a558e6b46916de774356270641)

PR #7796, opened 2025-09-14
BE-47: HashQL: distribute on union and intersections when collecting constraints

indietyp:

Makes the relationship explicit by destructuring the match arms.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/core/src/type/kind/mod.rs
+++ b/libs/@local/hashql/core/src/type/kind/mod.rs
@@ -1965,7 +1965,9 @@ impl<'heap> Inference<'heap> for TypeKind<'heap> {
                 | Self::Intrinsic(_)
                 | Self::Struct(_)
                 | Self::Tuple(_)
-                | Self::Closure(_),
+                | Self::Closure(_)
+                | Self::Never
+                | Self::Unknown,
             ) => {}
 
             // Primitive <: _
@@ -1978,7 +1980,9 @@ impl<'heap> Inference<'heap> for TypeKind<'heap> {
                 | Self::Intrinsic(_)
                 | Self::Struct(_)
                 | Self::Tuple(_)
-                | Self::Closure(_),
+                | Self::Closure(_)
+                | Self::Never
+                | Self::Unknown,
             ) => {}
 
             // Intrinsic <: _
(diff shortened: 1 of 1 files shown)


libs/@local/hashql/core/src/type/kind/mod.rs
```diff
@@ -2271,11 +2339,19 @@ impl<'heap> Inference<'heap> for TypeKind<'heap> {
...
+            (
+                Self::Never | Self::Unknown,
+                Self::Opaque(_)
+                | Self::Primitive(_)
+                | Self::Intrinsic(_)
+                | Self::Struct(_)
+                | Self::Tuple(_)
+                | Self::Closure(_)
+                | Self::Never
+                | Self::Unknown,
```
hashdotai, line 2353: I notice that `Self::Never` and `Self::Unknown` appear in the match arms on both sides of the pattern (both as subtypes and supertypes). This creates a redundancy where `Never <: Never` and `Unknown <: Unknown` are matched twice - once here and once in the more general pattern. Consider removing these from one side of the pattern to ensure each case is handled exactly once.

libs/@local/hashql/core/src/type/kind/mod.rs
```diff
@@ -2271,11 +2339,19 @@
...
 
-            // `Never <: _` | `_ <: Never`
-            (Self::Never, _) | (_, Self::Never) => {}
-
-            // `_ <: Unknown` | `Unknown <: _`
-            (_, Self::Unknown) | (Self::Unknown, _) => {}
+            // `Never <: _`
+            // `Unknown <: _`
+            (
+                Self::Never | Self::Unknown,
```
hashdotai, line 2345: For consistency with the other match patterns throughout this file, consider separating `Self::Never` and `Self::Unknown` into their own match arms rather than combining them with a `|` operator. While they have the same behavior, keeping consistent structure improves readability and maintainability.

libs/@local/hashql/core/src/type/kind/mod.rs
```diff
@@ -2212,7 +2240,17 @@
             }
 
             // Infer <: _
-            (&Self::Infer(Infer { hole: self_id }), _) => {
+            (
```
hashdotai, line 2243: All of these match arms are now significantly more explicit about the patterns they match, which is good. However, the increased verbosity means there's some duplication across different match arms. Consider introducing a helper function or macro to reduce repetition in these patterns if this pattern continues across more of the codebase.

TimDiekmann (approved):

PR #7798, opened 2025-09-15
BE-103: Implement path-based query traversal depths

TimDiekmann:

This PR adds support for `TraversalPath`-based subgraph query traversal alongside the existing `GraphResolveDepths` system. Users can now choose between two approaches: the traditional depth-based flooding across edge types, or precise path-based traversal by specifying a sequence of edges with their directions.

This enhancement enables more sophisticated graph queries where specific traversal paths are required, while maintaining backward compatibility with existing depth-based queries.

## 🔗 Related links

- [BE-106: Expose traversal path types in REST API](https://linear.app/hash/issue/BE-106/expose-traversal-path-types-in-rest-api) *(internal)* - prerequisite that exposed the TraversalPath types

## 🚫 Blocked by

- [x] BE-106 completed (`TraversalPath` types are now available in the REST API)

## 🔍 What does this change?

- **Added `TraversalPath[]` parameter alongside `GraphResolveDepths`**: Subgraph query requests now support both the existing `graphResolveDepths` field and the new `traversalPaths` field, allowing users to choose their preferred traversal approach
- **Extended Rust graph store implementation**: Modified the PostgreSQL store implementation to handle both path-based traversal and the existing depth-based flooding
- **Refactored traversal context logic**: Updated `libs/@local/graph/postgres-store/src/store/postgres/traversal_context.rs` to work with the new path-based system
- **Updated all entity/ontology query interfaces**: Modified entity, entity type, property type, and data type query implementations to use the new traversal system
- **Updated integration tests**: Modified existing tests to demonstrate both `GraphResolveDepths` and equivalent `TraversalPath` configurations, ensuring both systems work correctly

### Key files modified:

- `libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs` - Entity query implementation
- `libs/@local/graph/postgres-store/src/store/postgres/ontology/` - All ontology type query implementations
- `libs/@local/graph/postgres-store/src/store/postgres/traversal_context.rs` - Core traversal logic
- `libs/@local/graph/store/src/subgraph/edges/traversal.rs` - Traversal type definitions
- Integration tests in `tests/hash-backend-integration/src/tests/subgraph/` - Test updates

### API Enhancement Example:

**Option 1 (existing approach):**
```typescript
{
  filter: entityFilter,
  graphResolveDepths: {
    hasLeftEntity: { incoming: 1, outgoing: 0 },
    hasRightEntity: { incoming: 0, outgoing: 1 },
    // ... other edge types with depths
  },
  temporalAxes: currentTimeInstantTemporalAxes,
}
```

**Option 2 (new path-based approach):**
```typescript
{
  filter: entityFilter,
  traversalPaths: [
    {
      edges: [
        { kind: "has-left-entity", direction: "incoming" },
        { kind: "has-right-entity", direction: "outgoing" }
      ]
    }
  ],
  temporalAxes: currentTimeInstantTemporalAxes,
}
```

This PR:

The graph store libraries will need version bumps for this breaking change.

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Monitor performance differences between depth-based and path-based queries in production
- Use `TraversalPaths` everywhere
- Remove `GraphResolveDepths` support

## 🛡 What tests cover this?

- **Enhanced integration tests**: Tests in `tests/hash-backend-integration/src/tests/subgraph/circular.test.ts` and `friendship.test.ts` now demonstrate both approaches
- **Preserved test logic**: The same graph traversal scenarios are tested, with examples showing both `GraphResolveDepths` and equivalent `TraversalPath` usage
- **Edge case coverage**: Complex circular reference scenarios and multi-hop traversals are covered


libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1354,12 +1390,33 @@ where
...
                     .keys()
-                    .map(|id| {
-                        (
-                            *id,
-                            Cow::Borrowed(&traversal_params),
-                            subgraph.temporal_axes.resolved.variable_interval(),
-                        )
+                    .flat_map(|id| {
+                        match &traversal_params {
+                            // TODO: The `vec` is not ideal as the flattening intermediate type but
```
hashdotai, line 1395: Consider adding more descriptive comments here to explain the purpose of this `vec!` intermediate and why it's needed for flattening. The TODO comment is helpful but adding context on the technical reasons would make it clearer for other developers.

libs/@local/graph/postgres-store/src/store/postgres/traversal_context.rs
```diff
@@ -175,31 +177,46 @@ impl<K: Eq + Hash + Copy> TraversalContextMap<K> {
...
+                    Some((
+                        key,
+                        BorrowedTraversalParams::ResolveDepths {
+                            graph_resolve_depths,
+                        },
+                        interval,
+                    ))
+                }
+            }
+            traversal_path @ BorrowedTraversalParams::Path { .. } => {
```
hashdotai, line 216: This pattern match for the traversal path doesn't handle path validation or optimization. You might want to consider adding some checks here, similar to what you have for ResolveDepths, to avoid redundant traversals when paths overlap. Even a basic implementation would be helpful for initial performance.

libs/@local/graph/postgres-store/src/store/postgres/ontology/data_type.rs
```diff
@@ -339,13 +342,42 @@ where
...
+                                direction: OntologyTraversalEdgeDirection::Outgoing,
+                            } => OntologyEdgeKind::ConstrainsValuesOn,
+                            TraversalEdge::ConstrainsPropertiesOn { .. }
+                            | TraversalEdge::ConstrainsLinksOn { .. }
+                            | TraversalEdge::ConstrainsLinkDestinationsOn { .. }
+                            | TraversalEdge::IsOfType { .. }
+                            | TraversalEdge::HasLeftEntity { .. }
+                            | TraversalEdge::HasRightEntity { .. } => continue,
+                        };
+
```
hashdotai, line 372: Consider extracting the edge kind selection logic into a helper function since this pattern is repeated across several files (entity_type.rs, property_type.rs). This would improve maintainability and ensure consistent behavior across all ontology types.

tests/hash-backend-integration/src/tests/subgraph/circular.test.ts
```diff
@@ -322,58 +336,37 @@ describe("Single linked list", () => {
...
-    expect(
-      verticesEquals(subgraph, [
-        entity_a,
-        link_ab,
-        entity_b,
-        link_da,
-        entity_d,
-      ]),
-    ).toBe(true);
+    expect(verticesEquals(subgraph, [entity_a, link_ab, entity_b])).toBe(true);
```
hashdotai, line 365: The vertices comparison here is different from the previous implementation (fewer entities included). If this is intentional due to the path-based traversal's more precise nature, it would be helpful to add a comment explaining why the expected vertices are different than before.

TimDiekmann: The subgraph returns fewer entities because we aimed for entity A, link AB, and entity B. There was no reason to also include other entities. This is more concise. 
And I'm **not** going to add a comment that a previous implementation was different. I have no Idea why AI always comes up with such a stupid idea.

libs/@local/graph/postgres-store/src/store/postgres/ontology/data_type.rs
TimDiekmann: What would be the resulting output? We would end up with basically the same and have an additional function to read edges.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -43,24 +43,24 @@ use crate::store::{
 };
 
 #[derive(Debug)]
-pub struct EntityEdgeTraversalData {
+pub struct EntityEdgeTraversalData<'edges> {
```
indietyp, line 46: big fan of giving the lifetime a proper name!

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -103,12 +103,12 @@ where
     C: AsClient,
 {
     #[tracing::instrument(level = "info", skip(self))]
-    pub(crate) async fn read_shared_edges<'t>(
+    pub(crate) async fn read_shared_edges<'edges, 't>(
```
indietyp: Non-blocking nitpick: the order is inconsistent between the two functions.

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
```diff
@@ -295,96 +297,125 @@ where
...
-                            traversal_data,
-                            ReferenceTable::PropertyTypeConstrainsValuesOn,
-                        )
-                        .await?,
-                        provider,
-                        subgraph.temporal_axes.resolved.clone(),
+            for (edge_kind, table) in [(
+                OntologyEdgeKind::ConstrainsValuesOn,
+                ReferenceTable::PropertyTypeConstrainsValuesOn,
+            )] {
```
indietyp, line 342: This loop seems a bit suspect to me

indietyp (changes requested):
one question, otherwise LGTM

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
TimDiekmann: It's just for consistency. The pattern is the same everywhere. No strong opinion to keep it or not

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#single_char_lifetime_names) against single-letter lifetimes. I tracked this under H-5356.

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
indietyp: It depends on whether you plan to add cases; if not, I'd remove it to decrease indentation. Otherwise, it's IMO quite distracting.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: Fixed in https://github.com/hashintel/hash/pull/7798/commits/318c238d0301902b327d0d0520f8ed3228fd63e4

libs/@local/graph/postgres-store/src/store/postgres/ontology/property_type.rs
TimDiekmann: I didn't plan to add cases yet, no, but I would keep the indentation because the next section is effectively the same but with slightly different parameters (but too different to unify them, in particular because we want to eventually remove these in the coming weeks)

indietyp: Okay, then let's keep it 🙂

indietyp (approved):

PR #7799, opened 2025-09-16
H-2592, BE-109: Add UUID validation to prevent malformed identifiers in graph data

TimDiekmann:

This PR addresses an issue in the UUID validation which prevents core functionality like the `IsEntityId` function to work.

The solution implements strict RFC4122 UUID validation at the serialization/deserialization boundary, ensuring that only properly formatted UUIDs with valid version and variant fields can enter the system.

## 🔗 Related links

- [H-2592: Enforce real UUIDs in the graph](https://linear.app/hashintel/issue/H-2592) - Original Linear issue
- [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122) - UUID specification being enforced

## 🚫 Blocked by

N/A - This PR is ready for review.

### Core Implementation
- **New validation module**: Added `libs/@local/codec/src/serde/valid_uuid.rs` with RFC4122-compliant UUID validation
  - Validates UUID version is present (not `None`)
  - Validates UUID variant is RFC4122 (rejecting other variants)
  - Provides clear error messages for validation failures

### Applied validation across critical types
- **Entity identifiers**: `EntityUuid`, `DraftId`, `EntityEditionId` in `@blockprotocol/type-system`
- **Ontology identifiers**: `OntologyTypeUuid`
- **Role identifiers**: `TeamRoleId`, `WebRoleId`
- **Authorization**: `PolicyId` in graph authorization
- **Query system**: `Parameter::Uuid` and `CursorField::Uuid` in graph store

### Test data normalization
- Updated test UUIDs from invalid formats like `0000000A-0001-0000-0000-000000000000`
- Changed to valid RFC4122 format like `0000000A-0001-8000-8000-000000000000`
- Updated test files: circular-links.http, circular.test.ts, friendship.test.ts, and corresponding JSONL fixtures

### Dependencies
- Added `uuid` dependency to `@local/codec` crate with serde feature enabled

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/codec/src/serde/valid_uuid.rs
+++ b/libs/@local/codec/src/serde/valid_uuid.rs
@@ -33,6 +33,10 @@ pub fn serialize<S: Serializer>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::
 /// - If the uuid is missing a version or variant.
 pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Uuid, D::Error> {
     let uuid = Uuid::deserialize(deserializer)?;
+    if uuid.is_nil() {
+        return Ok(uuid);
+    }
+
     if uuid.get_version().is_none() {
         return Err(de::Error::custom(InvalidUuid::VersionMissing));
     }


libs/@local/codec/src/serde/valid_uuid.rs
```diff
@@ -0,0 +1,47 @@
...
+pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Uuid, D::Error> {
+    let uuid = Uuid::deserialize(deserializer)?;
+    if uuid.is_nil() {
+        return Ok(uuid);
+    }
+
+    if uuid.get_version().is_none() {
+        return Err(de::Error::custom(InvalidUuid::VersionMissing));
+    }
+    if !matches!(uuid.get_variant(), Variant::RFC4122) {
```
indietyp, line 43: Variant has a `Display` implementation, might be worth to add it to the error and display what the actual variant is for better error reporting.

indietyp (approved):
LGTM just minor non-blocking improvement

libs/@local/codec/src/serde/valid_uuid.rs
TimDiekmann: The `Display` implementation basically is the `Debug` implementation, so not really great IMO. Effectively, `NCS` and `Microsoft` are both for backwards compatibility which we don't need so it does not matter what variant is set.

PR #7802, opened 2025-09-18
BE-50: HashQL: Introduce a `DiagnosticResult` type

indietyp:

Currently, we handle errors through the use of a `Result<…, Vec<Diagnostic>` or `(Option<…>, Vec<Diagnostic>)` (in case of cases where errors can happen.)

This makes handling the diagnostics quite painful and error-prone. Instead, we should switch to a more sophisticated approach, using a `sink` and `DiagnosticResult`, which looks somewhat like:

```rust
struct DiagnosticResult<T, C, S> {
  value: Result<T, Diagnostic<C, S>>,
  diagnostics: DiagnosticIssues
}
```

This would allow us to store the diagnostics and then retrieve the value. The error would always be a fatal diagnostic.

We introduce two new types: `DiagnosticIssues`, which replaces the `Diagnostics` type in the type system, and `DiagnosticResult` for error propagation.

This has been introduced before https://linear.app/hash/issue/BE-39/hashql-handle-errors-in-the-graph-api-properly, as otherwise the rendering of the diagnostics would have been more complicated than necessary.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/hashql/diagnostics/src/diagnostic.rs
+++ b/libs/@local/hashql/diagnostics/src/diagnostic.rs
@@ -43,9 +43,9 @@ impl<C, S> Diagnostic<C, S> {
     /// Initializes an empty diagnostic that can be populated with message, labels, notes,
     /// and help messages through the appropriate methods. The category and severity
     /// determine how the diagnostic will be classified and displayed.
-    pub fn new(category: impl Into<C>, severity: Severity) -> Self {
+    pub const fn new(category: C, severity: Severity) -> Self {
         Self {
-            category: category.into(),
+            category,
             severity,
             message: None,
             labels: Vec::new(),
--- a/libs/@local/hashql/diagnostics/src/issues.rs
+++ b/libs/@local/hashql/diagnostics/src/issues.rs
@@ -1,0 +1,120 @@
+use crate::{Diagnostic, category::DiagnosticCategory};
+
+pub type BoxedDiagnosticIssues<'category, S> =
+    DiagnosticIssues<Box<dyn DiagnosticCategory + 'category>, S>;
+
+pub struct DiagnosticIssues<C, S> {
+    diagnostics: Vec<Diagnostic<C, S>>,
+    fatal: usize,
+}
+
+impl<C, S> DiagnosticIssues<C, S> {
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            diagnostics: Vec::new(),
+            fatal: 0,
+        }
+    }
+
+    pub fn map<C2, S2>(
+        self,
+        func: impl FnMut(Diagnostic<C, S>) -> Diagnostic<C2, S2>,
+    ) -> DiagnosticIssues<C2, S2> {
+        DiagnosticIssues {
+            diagnostics: self.diagnostics.into_iter().map(func).collect(),
+            fatal: self.fatal,
+        }
+    }
+
+    pub fn map_category<C2>(self, mut func: impl FnMut(C) -> C2) -> DiagnosticIssues<C2, S> {
+        DiagnosticIssues {
+            diagnostics: self
+                .diagnostics
+                .into_iter()
+                .map(|diagnostic| diagnostic.map_category(&mut func))
+                .collect(),
+            fatal: self.fatal,
+        }
+    }
+
+    pub fn boxed<'category>(self) -> BoxedDiagnosticIssues<'category, S>
+    where
+        C: DiagnosticCategory + 'category,
+    {
+        DiagnosticIssues {
+            diagnostics: self
+                .diagnostics
+                .into_iter()
+                .map(Diagnostic::boxed)
+                .collect(),
+            fatal: self.fatal,
+        }
+    }
+
+    pub fn clear(&mut self) {
+        self.diagnostics.clear();
+        self.fatal = 0;
+    }
+
+    #[must_use]
+    pub const fn fatal(&self) -> usize {
+        self.fatal
+    }
+
+    #[must_use]
+    pub const fn len(&self) -> usize {
+        self.diagnostics.len()
+    }
+
+    #[must_use]
+    pub const fn is_empty(&self) -> bool {
+        self.diagnostics.is_empty()
+    }
+
+    pub fn push(&mut self, diagnostic: Diagnostic<C, S>) {
+        self.fatal += usize::from(diagnostic.severity.is_fatal());
+        self.diagnostics.push(diagnostic);
+    }
+
+    pub fn append(&mut self, other: &mut Self) {
+        self.fatal += other.fatal;
+        self.diagnostics.append(&mut other.diagnostics);
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &Diagnostic<C, S>> {
+        self.diagnostics.iter()
+    }
+}
+
+impl<C, S> Extend<Diagnostic<C, S>> for DiagnosticIssues<C, S> {
+    #[expect(clippy::indexing_slicing, reason = "indexing is checke")]
+    fn extend<T: IntoIterator<Item = Diagnostic<C, S>>>(&mut self, iter: T) {
+        let previous = self.diagnostics.len();
+        self.diagnostics.extend(iter);
+
+        if self.diagnostics.len() == previous {
+            return;
+        }
+
+        self.fatal += self.diagnostics[previous..]
+            .iter()
+            .filter(|diagnostic| diagnostic.severity.is_fatal())
+            .count();
+    }
+}
+
+impl<C, S> IntoIterator for DiagnosticIssues<C, S> {
+    type IntoIter = alloc::vec::IntoIter<Self::Item>;
+    type Item = Diagnostic<C, S>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.diagnostics.into_iter()
+    }
+}
+
+impl<C, S> Default for DiagnosticIssues<C, S> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/libs/@local/hashql/diagnostics/src/issues.rs
+++ b/libs/@local/hashql/diagnostics/src/issues.rs
@@ -1,8 +1,60 @@
 use crate::{Diagnostic, category::DiagnosticCategory};
 
+/// Type alias for [`DiagnosticIssues`] with type-erased diagnostic categories.
+///
+/// This convenience type allows working with collections of diagnostics where the
+/// specific category types may vary.
 pub type BoxedDiagnosticIssues<'category, S> =
     DiagnosticIssues<Box<dyn DiagnosticCategory + 'category>, S>;
 
+/// A collection of diagnostic messages for error reporting.
+///
+/// `DiagnosticIssues` collects diagnostic messages during compilation phases,
+/// allowing you to accumulate errors, warnings, and other messages before
+/// deciding how to handle them. Fatal diagnostics are errors that prevent
+/// successful compilation (severity codes ≥ 400).
+///
+/// # Examples
+///
+/// Basic usage:
+///
+/// ```
+/// use hashql_diagnostics::{Diagnostic, DiagnosticIssues, Severity};
+/// # use hashql_diagnostics::category::TerminalDiagnosticCategory;
+/// # const CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+/// #     id: "example", name: "Example"
+/// # };
+///
+/// let mut issues = DiagnosticIssues::new();
+/// issues.push(Diagnostic::new(CATEGORY, Severity::Error));
+/// issues.push(Diagnostic::new(CATEGORY, Severity::Warning));
+///
+/// assert_eq!(issues.len(), 2);
+/// assert_eq!(issues.fatal(), 1);
+/// ```
+///
+/// Working with the [`DiagnosticSink`] trait to process results:
+///
+/// ```
+/// use hashql_diagnostics::{Diagnostic, DiagnosticIssues, DiagnosticSink, Severity};
+/// # use hashql_diagnostics::category::TerminalDiagnosticCategory;
+/// # const CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+/// #     id: "example", name: "Example"
+/// # };
+///
+/// let mut issues = DiagnosticIssues::new();
+///
+/// // Success case - value is extracted
+/// let success: Result<i32, Diagnostic<_, ()>> = Ok(42);
+/// if let Some(value) = issues.sink(success) {
+///     assert_eq!(value, 42);
+/// }
+///
+/// // Error case - diagnostic is collected
+/// let error = Result::<i32, _>::Err(Diagnostic::new(CATEGORY, Severity::Error));
+/// assert!(issues.sink(error).is_none());
+/// assert_eq!(issues.len(), 1);
+/// ```
 #[must_use]
 #[derive(Debug)]
 pub struct DiagnosticIssues<C, S> {
@@ -11,6 +63,18 @@ pub struct DiagnosticIssues<C, S> {
 }
 
 impl<C, S> DiagnosticIssues<C, S> {
+    /// Creates a new, empty collection of diagnostic issues.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashql_diagnostics::DiagnosticIssues;
+    ///
+    /// let issues: DiagnosticIssues<(), ()> = DiagnosticIssues::new();
+    /// assert_eq!(issues.len(), 0);
+    /// assert_eq!(issues.fatal(), 0);
+    /// assert!(issues.is_empty());
+    /// ```
     pub const fn new() -> Self {
         Self {
             diagnostics: Vec::new(),
@@ -18,16 +82,62 @@ impl<C, S> DiagnosticIssues<C, S> {
         }
     }
 
+    /// Transforms each diagnostic using the provided function.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashql_diagnostics::{Diagnostic, DiagnosticIssues, Severity};
+    /// # use hashql_diagnostics::category::TerminalDiagnosticCategory;
+    /// # const OLD_CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "old", name: "Old"
+    /// # };
+    /// # const NEW_CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "new", name: "New"
+    /// # };
+    ///
+    /// let mut issues = DiagnosticIssues::new();
+    /// issues.push(Diagnostic::new(OLD_CATEGORY, Severity::Error));
+    ///
+    /// let transformed = issues.map(|diagnostic| Diagnostic::new(NEW_CATEGORY, diagnostic.severity));
+    /// ```
     pub fn map<C2, S2>(
         self,
         func: impl FnMut(Diagnostic<C, S>) -> Diagnostic<C2, S2>,
     ) -> DiagnosticIssues<C2, S2> {
-        DiagnosticIssues {
-            diagnostics: self.diagnostics.into_iter().map(func).collect(),
-            fatal: self.fatal,
-        }
+        let diagnostics: Vec<_> = self.diagnostics.into_iter().map(func).collect();
+
+        // re-calculate the fatal count as it might have changed during iteration
+        let fatal = diagnostics
+            .iter()
+            .filter(|diagnostic| diagnostic.severity.is_fatal())
+            .count();
+
+        DiagnosticIssues { diagnostics, fatal }
     }
 
+    /// Transforms the category of each diagnostic.
+    ///
+    /// This is useful when moving diagnostics between compilation phases that
+    /// use different category types.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashql_diagnostics::{Diagnostic, DiagnosticIssues, Severity};
+    /// # use hashql_diagnostics::category::TerminalDiagnosticCategory;
+    /// # const PARSER_CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "parser", name: "Parser"
+    /// # };
+    /// # const LOWERING_CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "lowering", name: "Lowering"
+    /// # };
+    ///
+    /// let mut issues = DiagnosticIssues::new();
+    /// issues.push(Diagnostic::new(PARSER_CATEGORY, Severity::Error));
+    ///
+    /// let lowering_issues = issues.map_category(|_| LOWERING_CATEGORY);
+    /// ```
     pub fn map_category<C2>(self, mut func: impl FnMut(C) -> C2) -> DiagnosticIssues<C2, S> {
         DiagnosticIssues {
             diagnostics: self
@@ -39,6 +149,25 @@ impl<C, S> DiagnosticIssues<C, S> {
         }
     }
 
+    /// Converts to a collection with type-erased categories.
+    ///
+    /// When combining diagnostics from different compilation phases that use
+    /// different category types, this method allows them to be stored together.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashql_diagnostics::{Diagnostic, DiagnosticIssues, Severity};
+    /// # use hashql_diagnostics::category::TerminalDiagnosticCategory;
+    /// # const CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "example", name: "Example"
+    /// # };
+    ///
+    /// let mut issues = DiagnosticIssues::new();
+    /// issues.push(Diagnostic::new(CATEGORY, Severity::Error));
+    ///
+    /// let boxed_issues = issues.boxed();
+    /// ```
     pub fn boxed<'category>(self) -> BoxedDiagnosticIssues<'category, S>
     where
         C: DiagnosticCategory + 'category,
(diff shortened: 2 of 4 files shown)


libs/@local/hashql/diagnostics/src/result.rs
```diff
@@ -0,0 +1,922 @@
...
+    /// assert_eq!(result.value, 42);
+    /// assert_eq!(result.diagnostics.len(), 1);
+    ///
+    /// // Error case
+    /// let error: DiagnosticResult<i32, _, ()> =
+    ///     DiagnosticResult::err(Diagnostic::new(CATEGORY, Severity::Error));
+    /// let error_result = error.into_result().expect_err("should be error");
+    /// assert!(error_result.primary.severity.is_fatal());
+    /// ```
+    pub fn into_result(self) -> Result<DiagnosticValue<T, C, S>, Box<DiagnosticError<C, S>>> {
```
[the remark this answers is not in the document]
indietyp: it literally is...

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+    ///     println!("Severity: {:?}", diagnostic.severity);
+    /// }
+    /// ```
+    pub fn iter(&self) -> impl Iterator<Item = &Diagnostic<C, S>> {
+        self.diagnostics.iter()
+    }
+}
+
+impl<C, S> Extend<Diagnostic<C, S>> for DiagnosticIssues<C, S> {
+    #[expect(clippy::indexing_slicing, reason = "indexing is checked")]
```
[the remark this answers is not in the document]
indietyp: It is, the only reason it could is if it is empty, that is checked before.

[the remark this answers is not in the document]
hashdotai: @copilot-pull-request-reviewer The indexing concern is valid, but the previous check at line 469-471 actually guarantees that `previous` is less than or equal to the current length:

```rust
if self.diagnostics.len() == previous {
    return;
}
```

This early return ensures we only proceed to indexing when `self.diagnostics.len() > previous`. However, for better safety and clarity, using a slice method like `get_unchecked(previous..)` with an explicit debug assertion or using a safe alternative like `.iter().skip(previous)` would be more robust and clearer about the invariant being maintained.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+    /// let first = issues.pop().expect("should have one diagnostic");
+    /// assert_eq!(first.severity, Severity::Error);
+    /// assert_eq!(issues.len(), 0);
+    ///
+    /// assert!(issues.pop().is_none());
+    /// ```
+    pub fn pop(&mut self) -> Option<Diagnostic<C, S>> {
+        let diagnostic = self.diagnostics.pop();
+
+        if let Some(diagnostic) = &diagnostic {
```
hashdotai: This function uses pattern matching with `let Some(diagnostic) = &diagnostic`, but it would be more idiomatic to directly check `if let Some(diagnostic) = self.diagnostics.pop()` to avoid the redundant binding.

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -0,0 +1,922 @@
...
+    ///
+    /// // Result became an error due to the fatal diagnostic
+    /// assert!(result.into_result().is_err());
+    /// assert!(additional.is_empty()); // Diagnostics were moved
+    /// ```
+    pub fn append_diagnostics(&mut self, diagnostics: &mut DiagnosticIssues<C, S>) {
+        self.diagnostics.append(diagnostics);
+
+        if self.result.is_ok()
+            && let Some(fatal) = self.diagnostics.pop_fatal()
```
hashdotai: This guard condition uses the new Rust `if let Some() && let` syntax pattern which requires a newer Rust version. While this is fine since you're already using nightly features, it might be worth documenting this requirement explicitly somewhere.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+    ///
+    /// let mut issues: DiagnosticIssues<_, ()> = DiagnosticIssues::new();
+    /// issues.push(Diagnostic::new(CATEGORY, Severity::Error));
+    /// issues.push(Diagnostic::new(CATEGORY, Severity::Warning));
+    ///
+    /// assert_eq!(issues.len(), 2);
+    /// assert_eq!(issues.fatal(), 1);
+    /// ```
+    pub fn push(&mut self, diagnostic: Diagnostic<C, S>) {
+        self.fatal += usize::from(diagnostic.severity.is_fatal());
```
hashdotai: Consider adding a helper method that combines `self.fatal += usize::from(diagnostic.severity.is_fatal())` since this pattern is repeated multiple times in the codebase. This would improve maintainability and reduce the chance of bugs in the future.

libs/@local/hashql/diagnostics/src/result.rs
```diff
@@ -0,0 +1,922 @@
...
+#[expect(
+    clippy::field_scoped_visibility_modifiers,
+    reason = "required for `DiagnosticIssues`"
+)]
+pub struct DiagnosticResult<T, C, S> {
+    pub(crate) diagnostics: DiagnosticIssues<C, S>,
+    pub(crate) result: Result<T, Diagnostic<C, S>>,
+}
+
+impl<T, C, S> DiagnosticResult<T, C, S> {
```
hashdotai: For better ergonomics, consider implementing `From<Result<T, Diagnostic<C, S>>>` for `DiagnosticResult<T, C, S>` to allow easy conversion from standard Result types.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+    #[expect(clippy::indexing_slicing, reason = "indexing is checked")]
+    fn extend<T: IntoIterator<Item = Diagnostic<C, S>>>(&mut self, iter: T) {
+        let previous = self.diagnostics.len();
+        self.diagnostics.extend(iter);
+
+        if self.diagnostics.len() == previous {
+            return;
+        }
+
+        self.fatal += self.diagnostics[previous..]
```
hashdotai: This function might not behave as expected if the slice indexing `self.diagnostics[previous..]` is empty. While the early return should prevent this, adding a debug assertion like `debug_assert!(self.diagnostics.len() > previous)` before the indexing would make the safety invariant more explicit.

indietyp: it literally does because of the previous statement (you even said it yourself...)

libs/@local/hashql/diagnostics/src/result.rs
indietyp: no, because the Result could have a non-fatal diagnostic in it's error.

libs/@local/hashql/diagnostics/src/result.rs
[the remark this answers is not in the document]
TimDiekmann: Is it actually that big that clippy would complain? (just curious, boxing definitely is correct here)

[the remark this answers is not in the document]
indietyp: Yes it is, that's why I did it

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+/// specific category types may vary.
+pub type BoxedDiagnosticIssues<'category, S> =
+    DiagnosticIssues<Box<dyn DiagnosticCategory + 'category>, S>;
+
+/// A collection of diagnostic messages for error reporting.
+///
+/// `DiagnosticIssues` collects diagnostic messages during compilation phases,
+/// allowing you to accumulate errors, warnings, and other messages before
+/// deciding how to handle them. Fatal diagnostics are errors that prevent
+/// successful compilation (severity codes ≥ 400).
```
TimDiekmann: Is this derived from HTTP codes or did it just happen?

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+/// # const CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+/// #     id: "example", name: "Example"
+/// # };
+///
+/// let mut issues: DiagnosticIssues<_, ()> = DiagnosticIssues::new();
+/// issues.push(Diagnostic::new(CATEGORY, Severity::Error));
+/// issues.push(Diagnostic::new(CATEGORY, Severity::Warning));
+///
+/// assert_eq!(issues.len(), 2);
+/// assert_eq!(issues.fatal(), 1);
```
TimDiekmann: It feels weird that you push an `Error` but check for `fatal`

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,737 @@
...
+
+impl<C, S> DiagnosticIssues<C, S> {
+    /// Creates a new, empty collection of diagnostic issues.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashql_diagnostics::DiagnosticIssues;
+    ///
+    /// let issues: DiagnosticIssues<(), ()> = DiagnosticIssues::new();
```
TimDiekmann, line 91: Does it make sense to always create a `<(), ()>` from `new()`?

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -0,0 +1,931 @@
...
+/// ```
+#[must_use]
+#[derive(Debug)]
+#[expect(
+    clippy::field_scoped_visibility_modifiers,
+    reason = "required for `DiagnosticIssues`"
+)]
+pub struct Status<T, C, S> {
+    pub(crate) diagnostics: DiagnosticIssues<C, S>,
+    pub(crate) result: Result<T, Diagnostic<C, S>>,
```
TimDiekmann: [nitpick]: I'd put `result` as first member, it's the interesting part
```suggestion
    pub(crate) result: Result<T, Diagnostic<C, S>>,
    pub(crate) diagnostics: DiagnosticIssues<C, S>,
```

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -0,0 +1,931 @@
...
+
+                Ok(Success {
+                    value,
+                    diagnostics: self.diagnostics,
+                })
+            }
+            Err(diagnostic) => {
+                debug_assert!(
+                    diagnostic.severity.is_fatal(),
+                    "Fatal diagnostics should only be present in error variants"
```
TimDiekmann:
```suggestion
                    "Only fatal diagnostics should be present in error variants"
```

libs/@local/hashql/diagnostics/src/diagnostic.rs
```diff
@@ -17,35 +17,137 @@ use crate::{
...
+    /// # const CATEGORY: TerminalDiagnosticCategory = TerminalDiagnosticCategory {
+    /// #     id: "syntax_error", name: "Syntax Error"
+    /// # };
+    ///
+    /// let diagnostic: Diagnostic<_, ()> = Diagnostic::new(CATEGORY, Severity::Error);
+    /// assert_eq!(diagnostic.severity, Severity::Error);
+    /// assert!(diagnostic.message.is_none());
+    /// assert!(diagnostic.labels.is_empty());
+    /// ```
+    pub const fn new(category: C, severity: Severity) -> Self {
```
TimDiekmann, line 148: Is it intended that you use `Severity` instead of `F` here?

libs/@local/hashql/diagnostics/src/diagnostic.rs
```diff
@@ -17,35 +17,137 @@ use crate::{
...
+///     .add_help(Help::new("Consider using a different approach"));
+///
+/// assert_eq!(diagnostic.notes.len(), 1);
+/// assert_eq!(diagnostic.help.len(), 1);
+/// ```
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[must_use = "A diagnostic must be reported"]
-pub struct Diagnostic<C, S> {
+pub struct Diagnostic<C, S, K = Severity> {
```
TimDiekmann, line 99: What does the `F` stands for?

libs/@local/hashql/diagnostics/src/severity.rs
```diff
@@ -1,9 +1,37 @@
...
+    /// assert!(Severity::Fatal.is_critical());
+    /// assert!(Severity::Error.is_critical());
+    ///
+    /// // Non-critical severities
+    /// assert!(!Severity::Warning.is_critical());
+    /// assert!(!Severity::Note.is_critical());
+    /// assert!(!Severity::Debug.is_critical());
+    /// ```
+    fn is_critical(self) -> bool;
+    fn is_advisory(self) -> bool;
```
TimDiekmann: We could move the non-critical examples to `is_advisory`. Also it's worth mentioning that a severity should either be critical OR advisory and that it's the responsibility of the implementor to upkeep this.
Probably not worth it but we could seal the trait.

TimDiekmann (commented):
I will properly review this later but it already looks very good. I like the trait approach much more, it feels more natural this way.

libs/@local/hashql/diagnostics/src/severity.rs
```diff
@@ -1,9 +1,37 @@
...
+use core::{
+    fmt::{self, Display},
+    mem,
+};
 
 use anstyle::Color;
 
 use crate::{Help, Note};
 
+pub const trait SeverityKind: Copy + const Into<Severity> {
```
TimDiekmann, line 35: [nitpick]: It feels a bit weird to be able to convert a `SeverityKind` into a `Severity`. I know that this trait only provides functionality to determine the kind of the severity, however, it might be more fitting to call this trait `Severity` and rename the current one to something else. If we only ever have two kinds we _could_ consider also making this an enum
```rust
enum Severity {
    Critical(CriticalSeverity),
    Advisory(AdvisorySeverity),
}
```

libs/@local/hashql/diagnostics/src/severity.rs
```diff
@@ -68,23 +96,23 @@ pub struct SeverityInfo {
...
 ///
 /// ```
-/// use hashql_diagnostics::Severity;
+/// use hashql_diagnostics::{Severity, severity::SeverityKind};
 ///
-/// // Fatal severity levels
+/// // Critical severity levels
 /// let bug = Severity::Bug; // Code: 600
 /// let fatal = Severity::Fatal; // Code: 500
 /// let error = Severity::Error; // Code: 400
```
TimDiekmann, line 139: What's the difference between the two?

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,839 @@
...
+/// assert_eq!(issues.len(), 1);
+/// ```
+pub trait DiagnosticSink<T> {
+    type Output;
+
+    /// Processes the input value, extracting success values and collecting diagnostics.
+    ///
+    /// Returns [`Some`] for successful operations and [`None`] when diagnostics
+    /// are collected instead.
+    fn sink(&mut self, value: T) -> Option<Self::Output>;
```
TimDiekmann, line 613: [nitpick]: I wonder if we want to come up with another function name than `sink`. But as it's designed to take different values I also don't know a better one.

libs/@local/hashql/diagnostics/src/label.rs
```diff
@@ -37,7 +38,10 @@ impl<S> Label<S> {
         &self.span
     }
 
-    pub fn message(&self) -> &str {
+    pub const fn message(&self) -> &str
+    where
+        String: [const] Borrow<str>,
```
TimDiekmann, line 43: Is it really worth it dealing with `const` here? We cannot create a const `String` anyway and it's unstable.

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -0,0 +1,521 @@
...
+    /// let result: Status<_, (), ()> = StatusExt::ok(42);
+    /// match result {
+    ///     Ok(success) => {
+    ///         assert_eq!(success.value, 42);
+    ///         assert_eq!(success.advisories.len(), 0);
+    ///     }
+    ///     Err(_) => panic!("should be successful"),
+    /// }
+    /// ```
+    fn ok(value: T) -> Self;
```
TimDiekmann: Wouldn't `fn success` be better suited given the type name?

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -0,0 +1,521 @@
...
+    ///     .expect_err("should be critical");
+    /// let result: Status<i32, _, ()> = StatusExt::err(critical_diagnostic);
+    /// match result {
+    ///     Ok(_) => panic!("should be error"),
+    ///     Err(error) => {
+    ///         assert!(error.primary.severity.is_critical());
+    ///     }
+    /// }
+    /// ```
+    fn err(error: Diagnostic<C, S, Critical>) -> Self;
```
TimDiekmann: Wouldn't `fn faillure` be better suited given the type name?

TimDiekmann (commented):
Mostly happy to approve it. I have a few things spotted and a few nitpicks (feel free to ignore them).

libs/@local/hashql/diagnostics/src/issues.rs
hashdotai: @TimDiekmann The severity code design appears to be inspired by HTTP status codes, though with a simplified approach. Similar to HTTP, we have ranges where codes ≥ 400 represent critical/error conditions (like HTTP 4xx and 5xx are error codes), while codes < 400 represent non-critical/advisory conditions (like HTTP 1xx, 2xx, 3xx). The specific values (100, 200, 300, 400, 500, 600) were likely chosen to provide clear separation between severity levels while maintaining an intuitive mental model familiar to developers who work with HTTP codes.

libs/@local/hashql/core/src/type/environment/context/diagnostics.rs
```diff
@@ -49,7 +49,7 @@
 
     /// Adds a new diagnostic to the collection.
     pub(crate) fn push(&mut self, diagnostic: TypeCheckDiagnostic) {
-        if diagnostic.severity.is_fatal() {
+        if diagnostic.severity.is_critical() {
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security This error occurs because we're replacing `is_fatal()` with `is_critical()` but the code scanning tool may be looking at an intermediate state where the method isn't defined on `Severity` yet. In the PR, we're adding the `SeverityKind` trait which defines `is_critical()` and implementing it for `Severity`, so the method will be available after these changes are applied. The new code does handle this properly.

libs/@local/hashql/diagnostics/src/diagnostic.rs
```diff
@@ -54,12 +59,79 @@
...
+            category: self.category,
+            severity: Advisory::new_unchecked(self.severity),
+            message: self.message,
+            labels: self.labels,
+            notes: self.notes,
+            help: self.help,
+        }
+    }
+
+    pub fn categorize(self) -> Result<Diagnostic<C, S, Advisory>, Diagnostic<C, S, Critical>> {
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security Good catch. The function `into_advisory_unchecked` is missing proper error documentation. This should be updated to include an `# Errors` section that documents the conditions under which the function might fail or panic, particularly since it's using `unchecked` in its name which suggests possible safety implications.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,797 @@
...
+            critical: self.critical,
+        }
+    }
+}
+
+impl<C, S> DiagnosticIssues<C, S, Severity> {
+    pub(crate) fn merge_into_advisories(
+        &mut self,
+        other: &mut DiagnosticIssues<C, S, Advisory>,
+    ) -> Result<(), (Diagnostic<C, S, Critical>, Self)> {
```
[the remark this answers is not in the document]
hashdotai: @github-advanced-security You're right about the large `Err` variant. The function `merge_into_advisories` returns an error type containing both a `Diagnostic` and the entire `DiagnosticIssues` collection, which can be quite large. We could consider optimizing this by:

1. Using references where possible
2. Returning just a reference to the critical diagnostic rather than the full collection
3. Or restructuring to avoid returning the entire collection at all

For a function that's only used internally (as indicated by the `pub(crate)` visibility), we should balance implementation convenience against memory efficiency.

libs/@local/hashql/diagnostics/src/diagnostic.rs
hashdotai: @TimDiekmann The `F` parameter likely stands for "Fatality" or "Fatal level" since it's used to categorize the severity level of diagnostics. It represents the generic severity type, which can be either the general `Severity` enum or one of the specialized types like `Critical` or `Advisory`. This enables compile-time enforcement of severity constraints for different diagnostic contexts.

libs/@local/hashql/diagnostics/src/severity.rs
hashdotai: @TimDiekmann It appears both methods (`is_critical` and `is_advisory`) are defined to be mutually exclusive - one returns `self.code() >= 400` while the other returns `!self.is_critical()`. The difference is just the condition they're checking:

- `is_critical()` returns true for high-severity diagnostics (Error, Fatal, Bug)
- `is_advisory()` returns true for low-severity diagnostics (Warning, Note, Debug)

They're complementary methods that provide different semantic meaning when used in code, even though they're checking opposite conditions.

libs/@local/hashql/diagnostics/src/status.rs
hashdotai: @TimDiekmann Yes, using `fn failure` would be more consistent with the type name `Failure`, just as we discussed for `success`. This would make the API more intuitive by aligning method names with the types they create.

The current `err` naming likely comes from consistency with `Result::Err`, but since we're using our specialized `Failure` type, having a matching method name would improve API clarity.

libs/@local/hashql/diagnostics/src/severity.rs
hashdotai: @TimDiekmann That's a good point. The current design does feel a bit awkward with the conversion. 

Your suggested enum-based approach would be more idiomatic Rust, providing a clearer type hierarchy and better semantic meaning. It would also enable exhaustive pattern matching, which could be valuable for handling different severity categories.

```rust
enum Severity {
    Critical(CriticalSeverity),
    Advisory(AdvisorySeverity),
}
```

This would make the distinction between severity kinds explicit at the type level, while still allowing for the specific levels within each category. It would also eliminate the need for runtime checks like `is_critical()`.

libs/@local/hashql/diagnostics/src/severity.rs
hashdotai: @TimDiekmann Good suggestions. Moving the non-critical examples to `is_advisory` would make the documentation more balanced and clearer. 

And yes, it would be valuable to explicitly document that these properties should be mutually exclusive - a severity should never be both critical AND advisory at the same time.

Sealing the trait could be worthwhile to prevent unexpected implementations, especially since this is a core part of the error handling system where correctness is essential. Since the trait is already marked with `pub const trait`, adding sealing wouldn't add much overhead but would provide additional safety guarantees.

libs/@local/hashql/diagnostics/src/label.rs
hashdotai: @TimDiekmann You're right, using `const` here seems unnecessarily complex, especially for a feature that's still unstable. The constraint `String: [const] Borrow<str>` is unlikely to be useful in practice since we can't create constant Strings anyway.

A simpler approach would be to just use a regular non-const function. If there's a specific need for const evaluation in some context, it could be addressed more directly rather than adding this complexity to the API.

libs/@local/hashql/diagnostics/src/diagnostic.rs
hashdotai: @TimDiekmann Yes, it appears intentional. In the `report` method, we're converting the potentially specialized severity type `K` to the general `Severity` type via `let severity: Severity = self.severity.into();`. This is necessary because the underlying `ariadne::Report::build` method expects a concrete severity kind, not a generic type parameter. The conversion ensures we always have a concrete `Severity` regardless of what `K` is.

libs/@local/hashql/diagnostics/src/issues.rs
hashdotai: @TimDiekmann I agree that `sink` is somewhat unusual terminology. Alternative names could be:

- `collect_diagnostic` - Emphasizes collecting/accumulating diagnostics
- `process` - Generic but clear about transforming the input
- `extract_value` - Focuses on what it returns (the value, if successful)
- `absorb` - Similar to sink but perhaps more intuitive

The challenge is finding a name that works well for all the different implementations (Result, DiagnosticIssues, etc.). `sink` does have the advantage of being concise and suggestive of its purpose - collecting/absorbing diagnostics while potentially passing through values.

libs/@local/hashql/diagnostics/src/status.rs
hashdotai: @TimDiekmann Yes, `fn success` would be more consistent with the type name `Success`. This would align the function name with the type it creates, making the API more intuitive. Using `ok` likely came from consistency with `Result::Ok`, but since we have our own specialized type, matching the method name to our type name would be clearer.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,839 @@
...
+
+impl<C, S> DiagnosticIssues<C, S, Severity> {
+    #[expect(
+        clippy::result_large_err,
+        reason = "The error is only used transiently"
+    )]
+    pub(crate) fn merge_into_advisories(
+        &mut self,
+        other: &mut DiagnosticIssues<C, S, Advisory>,
+    ) -> Result<(), (Diagnostic<C, S, Critical>, Self)> {
```
hashdotai, line 506: This function can be improved to better handle large error types. The `Err` variant here returns a tuple containing both a `Diagnostic` and the entire `DiagnosticIssues` collection, which can be memory inefficient.

Consider either:
1. Using Box for the error tuple to reduce stack size
2. Restructuring to avoid returning the entire collection
3. Using references where possible to avoid cloning large data structures

libs/@local/hashql/diagnostics/src/severity.rs
hashdotai: For the trait declaration, it would be valuable to explicitly document that `is_critical()` and `is_advisory()` should be mutually exclusive - a severity should never be both critical AND advisory at the same time. This would help implementors understand the invariants they need to maintain.

libs/@local/hashql/diagnostics/src/issues.rs
```diff
@@ -0,0 +1,839 @@
...
+    #[expect(clippy::indexing_slicing, reason = "indexing is checked")]
+    fn extend<T: IntoIterator<Item = Diagnostic<C, S, K>>>(&mut self, iter: T) {
+        let previous = self.diagnostics.len();
+        self.diagnostics.extend(iter);
+
+        if self.diagnostics.len() == previous {
+            return;
+        }
+
+        self.critical += self.diagnostics[previous..]
```
hashdotai, line 555: While the early return at lines 551-553 prevents issues with indexing, adding an explicit debug assertion like `debug_assert!(self.diagnostics.len() > previous)` before the indexing would make the safety invariant more obvious to future maintainers.

libs/@local/hashql/diagnostics/src/label.rs
hashdotai, line 43: The `const` constraint `String: [const] Borrow<str>` adds complexity for a feature that's still unstable. Since we can't create constant Strings anyway, consider simplifying this to a regular non-const function unless there's a specific need for const evaluation.

libs/@local/hashql/diagnostics/src/severity.rs
TimDiekmann: The question was about `fatal` vs `error`

libs/@local/hashql/diagnostics/src/issues.rs
indietyp: > Is this derived from HTTP codes or did it just happen?

Initially, they were based on HTTP status codes (in the very very first iteration) now the only thing that is the same is that we do jumps in 100s to allow for new codes in between.

libs/@local/hashql/diagnostics/src/diagnostic.rs
indietyp: > Is it intended that you use `Severity` instead of `F` here?

Yes, during construction, you pretty much always want to have a "normal" severity (which is then categorised during `Status` operations. To make it easier to use and type inference, play nice, we use normal inference here.

libs/@local/hashql/diagnostics/src/diagnostic.rs
indietyp: `K` stands for Severity**K**ind. In a perfect world I would've used `S`, but that is already taken for spans.

libs/@local/hashql/diagnostics/src/severity.rs
indietyp: > trait Severity and rename the current one to something else.

Thought about this, but didn't find a suitable name. I found this to be the most suitable way of using, additionally calling it Severity**K**ind allows us to have a generic letter that matches the trait :P

Adding two different types would IMO complicate the design, because now we need to implement the same methods multiple times instead of just dropping back to the default implementation.

Also note: the trait is const because I wanted to use it in all the places that currently `Severity` was used, as not to break things (and admittedly also experiment around with const traits), which is why this is const.

libs/@local/hashql/diagnostics/src/severity.rs
indietyp: `Error` may be recoverable in the application, e.g. something that the user can change to resolve the error. `Fatal` means that there has been a fatal non-recoverable error - something went seriously wrong, and `Bug` means that a program invariant has been violated.

examples:
* `Error`: query is malformed
* `Fatal`: connection to postgres cannot be established (any change from the user can't change that fact)
* `Bug`: invariant is violated - e.g. a special form still exists in the tree before reification, even though lowering should've removed them

libs/@local/hashql/diagnostics/src/label.rs
indietyp: > Is it really worth it dealing with const here? We cannot create a const String anyway and it's unstable.

Clippy yelled at me that it could be const D:, it also means that you can use `message` in a const context because `Cow::Borrowed(&'static str)`, so it isn't impossible to hit.

TimDiekmann: Clippy is wrong then. We should not add a nightly feature just to satisfy clippy. I'd prefer if we could leave out const traits here. In particular because we cannot create a `String` in a const context.

libs/@local/hashql/diagnostics/src/severity.rs
TimDiekmann: In the above example: If the connection to the database cannot be established, the user could retry. I cannot come up with an example where `Fatal` is needed (instead of `Error` or `Bug`)

libs/@local/hashql/diagnostics/src/label.rs
indietyp: > In particular because we cannot create a String in a const context.

Yes, but you can still create a `Cow::Borrowed` and use that then through `message`.

libs/@local/hashql/diagnostics/src/status.rs
indietyp: done in [ddf3c0f](https://github.com/hashintel/hash/pull/7802/commits/ddf3c0f93c3c4327a0839630bbc162ea5660b1ce)

libs/@local/hashql/diagnostics/src/status.rs
indietyp: done in [ddf3c0f](https://github.com/hashintel/hash/pull/7802/commits/ddf3c0f93c3c4327a0839630bbc162ea5660b1ce)

libs/@local/hashql/diagnostics/src/severity.rs
indietyp: done in [49ea34e](https://github.com/hashintel/hash/pull/7802/commits/49ea34e3509c9f1d5314f3fa99b9f9edaacf3732)

libs/@local/hashql/diagnostics/src/severity.rs
indietyp: > the user could retry. 

Yes, but it's not in his power to change anything. a `Error` indicates: it's your fault, not mine aka in HTTP speak: `Error` is 4xx, `Fatal` is 5xx.

libs/@local/hashql/diagnostics/src/label.rs
indietyp: (I had a couple of places where I had to de-const a couple of functions because of it AFAIR which is why I pursued the const design - pretty much everything possible surrounding `Severity` is used at compile time - otherwise I'd need to have `_const` or something functions in `Severity` to combat that and I want to avoid that as much as possible)

PR #7806, opened 2025-09-19
BE-122: HashQL: Adopt unified diagnostic infrastructure in `hashql-hir`

indietyp:

Adopts the new diagnostic infrastructure in HIR, to accommodate it, some new methods have been added to both `DiagnosticIssues` and `Status`.

> Important: Because the HIR tends to be *very* deep, we're not using the accumulation feature here. I have tried during reification, but found there to be a considerable slowdown. Instead the current approach is being kept.

A bonus is that the control-flow inside `HIR` `lower` is a lot clearer (IMO).

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -504,18 +644,75 @@ impl<T, C, S> StatusExt<T, C, S> for Status<T, C, S> {
...
         match self {
             Ok(success) => {
-                if let Err((critical, issues)) =
-                    diagnostics.merge_into_advisories(&mut success.advisories)
-                {
-                    *self = Err(Failure {
-                        primary: Box::new(critical),
-                        secondary: issues,
-                    });
+                if let Err(failure) = diagnostics.merge_into_advisories(&mut success.advisories) {
```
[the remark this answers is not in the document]
indietyp: done in [a64d67a](https://github.com/hashintel/hash/pull/7806/commits/a64d67a1d907709ec1b91d8529bea98c0014fd80)

libs/@local/hashql/hir/src/reify/error.rs
```diff
@@ -104,8 +107,10 @@ pub(crate) fn unsupported_construct(
 /// but the error wasn't properly handled.
 #[coverage(off)] // compiler bugs should never be hit
 pub(crate) fn dummy_expression(span: SpanId) -> ReificationDiagnostic {
-    let mut diagnostic =
-        Diagnostic::new(ReificationDiagnosticCategory::UnhandledError, Severity::Bug);
+    let mut diagnostic = Diagnostic::new(
+        ReificationDiagnosticCategory::UnhandledError,
+        Severity::Error,
```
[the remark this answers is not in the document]
indietyp: done in [a64d67a](https://github.com/hashintel/hash/pull/7806/commits/a64d67a1d907709ec1b91d8529bea98c0014fd80)

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -348,6 +386,46 @@ pub trait StatusExt<T, C, S>: sealed::Sealed {
...
+    ///
+    /// match doubled {
+    ///     Ok(result) => {
+    ///         assert_eq!(result.value, 84);
+    ///         assert_eq!(result.advisories.len(), 1);
+    ///     }
+    ///     Err(_) => panic!("should be successful"),
+    /// }
+    ///
+    /// // Error case - error is preserved
```
TimDiekmann: I think it makes sense to make this a separate test case/example.

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -447,6 +525,60 @@ pub trait StatusExt<T, C, S>: sealed::Sealed {
...
+    /// let error_diagnostic = Diagnostic::new(CATEGORY, Severity::Error);
+    /// let critical = error_diagnostic
+    ///     .specialize()
+    ///     .expect_err("should be critical");
+    /// let failure: Status<String, _, ()> = Status::failure(critical);
+    ///
+    /// let combined = Status::zip(success, failure);
+    /// assert!(combined.is_err());
+    /// ```
+    fn zip<B>(this: Self, other: Status<B, C, S>) -> Self::Zip<T, B>;
```
TimDiekmann, line 596: I'm not sure about the `name`, mainly because `Result` also can be converted into an iterator and zipping an iterator behaves differently (it omits the `Err` variant effectively). I wonder if `combine` would be more clear.

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -504,18 +644,79 @@ impl<T, C, S> StatusExt<T, C, S> for Status<T, C, S> {
...
-                    *self = Err(Failure {
-                        primary: Box::new(critical),
-                        secondary: issues,
-                    });
+                if let Err(failure) = diagnostics.merge_into_advisories(&mut success.advisories) {
+                    // Overwrite the entire status with failure: when a critical diagnostic is
+                    // encountered, the computation is no longer considered successful, so the
+                    // original value is discarded.
+                    // `merge_into_advisories` merges the advisories into the returned `failure`.
+                    *self = Err(failure);
```
TimDiekmann: Does it make sense to allow extracting the existing value and return `Option<T>`?

libs/@local/hashql/hir/tests/ui/lower/alias-replacement/generic-reassignment.jsonc
```diff
@@ -1,4 +1,4 @@
-//@ run: pass
+//@ run: fail
```
TimDiekmann, line 1: Why the change? Or: Why this wasn't covered before and we had an output?

indietyp: It always errored out, but we didn't collect the `AliasReplacement` diagnostics in the test correctly. The error was reported, but the test was falsely reported as passing.

libs/@local/hashql/diagnostics/src/status.rs
indietyp: I used `zip` because there's also `Option::zip`, which does something very similar.

libs/@local/hashql/diagnostics/src/status.rs
indietyp: I didn't think about that (primarily because I didn't need it) but i think that is a great idea!

indietyp: done in [2454cc8](https://github.com/hashintel/hash/pull/7806/commits/2454cc8a81ed41ba34526cda0f4ba748e58b8b73)

libs/@local/hashql/diagnostics/src/status.rs
indietyp: done in [2454cc8](https://github.com/hashintel/hash/pull/7806/commits/2454cc8a81ed41ba34526cda0f4ba748e58b8b73)

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -501,20 +656,88 @@ impl<T, C, S> StatusExt<T, C, S> for Status<T, C, S> {
...
-                        secondary: issues,
-                    });
+                if let Err(failure) = diagnostics.merge_into_advisories(&mut success.advisories) {
+                    // Overwrite the entire status with failure: when a critical diagnostic is
+                    // encountered, the computation is no longer considered successful, so the
+                    // original value is discarded.
+                    // `merge_into_advisories` merges the advisories into the returned `failure`.
+
+                    let success = mem::replace(self, Err(failure))
+                        .unwrap_or_else(|_err| unreachable!("Match arm has matched `Ok`"));
```
TimDiekmann:
```suggestion
                        .unwrap_or_else(|error| unreachable!("Match arm has matched `Err` but found {error:?}"));
```

TimDiekmann: You could also use `let Ok(success) = ... else { unreachable! };`

TimDiekmann: Nevermind, I was blind

PR #7811, opened 2025-09-20
BE-121: HashQL: Adopt unified diagnostic infrastructure in type-system

indietyp:

Continuation of the previous PRs.

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/hir/src/lower/checking.rs
```diff
@@ -190,14 +190,22 @@ impl<'env, 'heap> TypeChecking<'env, 'heap> {
...
+        }
+
+        if let Some(mut analysis) = self.analysis.take_diagnostics() {
+            self.analysis_diagnostics.append(&mut analysis);
+        }
+
+        self.diagnostics.append(
+            &mut self
+                .analysis_diagnostics
+                .map_category(LoweringDiagnosticCategory::TypeChecking),
```
hashdotai, line 207: You're mapping the category twice here. First you map it in line 204-208, then immediately after in the append call. This means the category is being transformed twice, which is likely not what you intended.

```suggestion
self.diagnostics.append(
    &mut self.analysis_diagnostics
);
```

libs/@local/hashql/diagnostics/src/status.rs
```diff
@@ -612,6 +641,19 @@ impl<T, C, S> StatusExt<T, C, S> for Status<T, C, S> {
...
+            Ok(Success { value, advisories }) => Ok(Success {
+                value,
+                advisories: advisories.map_category(func),
+            }),
+            Err(Failure { primary, secondary }) => Err(Failure {
+                primary: Box::new(primary.map_category(&mut func)),
+                secondary: secondary.map_category(func),
+            }),
+        }
+    }
```
hashdotai, line 670: You're creating a mutable borrowing in the closure argument that isn't needed anymore since you're using the `map_category` method. Either make this an immutable reference or remove the reference entirely:

```suggestion
        TypeId,
        error::{TypeCheckDiagnostic, TypeCheckDiagnosticStatus},
        inference::{InferenceSolver, VariableLookup},
```

libs/@local/hashql/hir/src/lower/checking.rs
```diff
@@ -190,14 +190,22 @@ impl<'env, 'heap> TypeChecking<'env, 'heap> {
...
+        );
```
hashdotai, line 208: The pattern of appending diagnostic issues after collecting them shows up in multiple places. Consider adding a helper method in the lower module to standardize this pattern and reduce duplication.

libs/@local/hashql/core/src/type/environment/analysis.rs
```diff
@@ -1,18 +1,22 @@
 use alloc::rc::Rc;
-use core::ops::{ControlFlow, Deref};
+use core::{
+    mem,
+    ops::{ControlFlow, Deref},
+};
 
+use hashql_diagnostics::DiagnosticIssues;
```
hashdotai, line 7: This change from a raw struct to using the proper diagnostic infrastructure is great! However, you still have a dependency on `hashql_diagnostics` that's only used here. Consider centralizing all diagnostic code in a common module to avoid scattered imports.

libs/@local/hashql/hir/src/lower/checking.rs
indietyp: That's exactly what this is?

PR #7815, opened 2025-09-23
H-5382: initialize @hashintel/ds-components

kube:

- Prototype of a new Component Library with refraction
- Storybook v9 (current is v7)
- Components based on ArkUI and PandaCSS
- Design Tokens

## 🔍 What does this change?

This PR adds three new packages, following PandaCSS [recommendations for Component Libraries in Monorepos](https://panda-css.com/docs/guides/component-library#why-should-my-component-library-use-an-external-package-styled-system) :

- `@hashintel/ds-theme`: Centralized design tokens for easy access and modification.
- `@hashintel/ds-helpers`: Uses `ds-theme` and allows to build styles.
- `@hashintel/ds-components`: Uses `ds-helpers` to style components.

```mermaid
graph BT
  subgraph DesignSystem
    Preset("@hashintel/ds-theme")
    StyledSystem("@hashintel/ds-helpers")
    ComponentLibrary("@hashintel/ds-components")
    
    StyledSystem --> Preset
    ComponentLibrary --> StyledSystem
  end
  
  subgraph Apps
    Frontend --> ComponentLibrary
    Frontend --> StyledSystem
    
    OtherApp --> ComponentLibrary
    OtherApp --> StyledSystem
  end
```

This lets apps create styles through `ds-helpers`, and get components from `ds-components`.

This PR:

- [ ] modifies an **npm**\-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**\-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**\-publishable library, but **it is not yet ready to publish**

The changes in this PR:

    - _Provide more detail here_

The changes in this PR:

- [ ] do not affect the execution graph

## ⚠️ Known issues

- These components only work in Chrome for now. See "Fallback" part in Storybook intro.
- FlexibleFilter backdrop can sometimes have rendering issues.
- CSS variables/classes can not directly control the refraction effect (which contains its own border radius). This could be achieved through watchers on CSS properties but might be overkill/laggy. Maybe we should define a standard props-based way to define radius on these components.

## 🐾 Next steps

- Cleanup and provide better interface to components
- Standardize the Optics-related properties (`radius`, `bezelWidth`, `refractiveIndex`, `glassHeight`, `bezelHeightFn`)
- More components
- Tests
- Fallback for Safari/Firefox

## 🛡 What tests cover this?

There are no automated tests for now. These components are still not usable in app, and will be subject to rework.

## ❓ How to test this?

There is no turbo.json for now, so to test it locally:

```
cd libs/@hashintel/styled-system
yarn build

cd libs/@hashintel/component-library
yarn storybook
```

## 📹 Demo

[](https://github.com/user-attachments/assets/17802b1b-a0ff-4c06-8c8c-439ef557da5a)

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Button/button.tsx
+++ b/libs/@hashintel/ds-components/src/components/Button/button.tsx
@@ -24,7 +24,7 @@ export const Button: React.FC<ButtonProps> = ({
             cursor: "not-allowed",
           },
         }),
-        className
+        className,
       )}
       {...props}
     >


libs/@hashintel/component-library/src/components/Button/button.tsx
```diff
@@ -0,0 +1,47 @@
...
+  radius?: number;
+  specularOpacity?: number;
+  specularSaturation?: number;
+  blurLevel?: number;
+  glassThickness?: number;
+  refractiveIndex?: number;
+  bezelWidth?: number;
+}
+
+export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
```
indietyp: `forwardRef` is deprecated since React 19: https://react.dev/reference/react/forwardRef

libs/@hashintel/component-library/src/components/Button/button.tsx
```diff
@@ -0,0 +1,47 @@
...
+        type="button"
+        className={cx(
+          css({
+            position: "relative",
+            display: "inline-flex",
+            alignItems: "center",
+            justifyContent: "center",
+            fontWeight: "semibold",
+            borderRadius: "md",
+            cursor: disabled ? "not-allowed" : "pointer",
```
indietyp: Shouldn't `disabled` be a button prop? and then a pandaCSS selector of `_disabled`?

libs/@hashintel/component-library/src/components/Button/button.tsx
```diff
@@ -0,0 +1,47 @@
...
+            transition: "all",
```
indietyp: I would advise against `transition: all` it tends to lead to bad performance, because you'll transition unintended properties.

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
```diff
@@ -0,0 +1,224 @@
...
+            boxShadow: "inset 1px 1px 1px rgba(0, 0, 0, 0.05)",
+          }}
+        />
+
+        {options.map((option) => (
+          <SegmentGroup.Item
+            key={option.value}
+            value={option.value}
+            className={itemStyles}
+            style={{ borderRadius: radius - ROOT_PADDING }}
```
indietyp, line 190: Couldn't you set a data property of `data-radius`, then css `calc` it to reduce the amount of inline styles required?

(Same applies to other `style` properties here)

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
```diff
@@ -0,0 +1,224 @@
...
+        className={cx(rootStyles, className)}
+        style={{
+          padding: ROOT_PADDING,
+          borderRadius: radius,
+          ...style,
+        }}
+      >
+        <div
+          className={rootBackdropStyles}
+          style={{
```
indietyp, line 179: Can't large parts of this be factored out into a CSS class?

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
```diff
@@ -0,0 +1,224 @@
...
+        onValueChange={(details) => {
+          if (details.value) {
+            onValueChange?.(details.value);
+          }
+        }}
+        disabled={disabled}
+        defaultValue={defaultValue}
+        className={cx(rootStyles, className)}
+        style={{
+          padding: ROOT_PADDING,
```
indietyp, line 172: Isn't `ROOT_PADDING` a constant? Shouldn't it go in `rootStyles` then?

libs/@hashintel/ds-components/src/components/Switch/switch.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  width: SLIDER_WIDTH,
+                  height: SLIDER_HEIGHT,
+                  borderRadius: SLIDER_RADIUS,
+                }}
+                className={css({
+                  position: "relative",
+                  display: "inline-block",
+                  cursor: disabled ? "not-allowed" : "pointer",
+                  transition: "all 0.2s ease",
+                  backgroundColor: context.checked ? "green.40" : "gray.40",
```
indietyp: Why are you using the context here? can't you use `data-state` state here as well? (same with `disabled`)

libs/@hashintel/ds-components/src/lib/displacement-map.ts
```diff
@@ -0,0 +1,143 @@
...
+        : 0;
+
+      const distanceToCenterSquared = x * x + y * y;
+
+      const isInBezel =
+        distanceToCenterSquared <= radiusPlusOneSquared &&
+        distanceToCenterSquared >= radiusMinusBezelSquared;
+
+      // Only write non-neutral displacements (when isInBezel)
+      if (isInBezel) {
```
indietyp: You're iterating here *a lot* of dead space, wouldn't it make sense to invert the condition instead? Only checking the pixels that are inside the bezel?

That should be pretty straightforward. For each corner, create a radius of pixels (or even an AABB), then iterate over those instead.

libs/@hashintel/ds-components/src/lib/displacement-map.ts
```diff
@@ -0,0 +1,143 @@
...
+        const opacity =
```
indietyp: Actually just spitballing here, but couldn't you just calculate the displacement map at one of the corners, then copy over the data to the other three corners?

libs/@hashintel/ds-components/src/lib/displacement-map.ts
```diff
@@ -0,0 +1,143 @@
...
+    ] as const;
+  }
+
+  return Array.from({ length: samples }, (_, i) => {
+    const x = i / samples;
+    const y = bezelHeightFn(x);
+
+    // Calculate derivative in x
+    const dx = x < 1 ? 0.0001 : -0.0001;
+    const y2 = bezelHeightFn(x + dx);
```
indietyp: Couldn't you save the second iteration here by instead precomputing the bezelHeight, then calculating the derivative by inspecting the previous or next step?

libs/@hashintel/ds-components/src/lib/displacement-map.ts
```diff
@@ -0,0 +1,143 @@
+/* eslint-disable id-length */
+import { createImageData } from "canvas";
+
+export function calculateDisplacementMapRadius(
+  glassThickness: number = 200,
+  bezelWidth: number = 50,
+  bezelHeightFn: (x: number) => number = (x) => x,
+  refractiveIndex: number = 1.5,
+  samples: number = 128
+): number[] {
```
indietyp: Would it make sense to instead return a `Float32Array` here?

libs/@hashintel/ds-components/src/lib/displacement-map.ts
indietyp: Something like:

```ts
const bezelHeight = new Float64Array(n);
for (let i = 0; i < n; i++) {
	bezelHeight[i] = bezelHeightFn(i * step);
}

const out = new Float64Array(n);
for (let i = 0; i < n; i++) {
	// Central difference where possible, otherwise one-sided
    let dy: number;
    if (i > 0 && i + 1 < n) {
      dy = (bezelHeight[i + 1] - bezelHeight[i - 1]) * inv2step;
    } else if (i === 0) {
      dy = (bezelHeight[1] - bezelHeight[0]) / step;
    } else {
      dy = (bezelHeight[n - 1] - bezelHeight[n - 2]) / step;
    }
}

return out;
```

libs/@hashintel/ds-components/src/lib/image-data-to-url.ts
```diff
@@ -0,0 +1,17 @@
...
+  x = 0,
+  y = 0
+): string {
+  const canvas = createCanvas(
+    width ?? imageData.width,
+    height ?? imageData.height
+  );
+  const ctx = canvas.getContext("2d");
+  ctx.putImageData(imageData, -x, -y);
+  return canvas.toDataURL();
```
indietyp, line 16: We should use `createObjectUrl` instead: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL

libs/@hashintel/ds-components/src/lib/use-is-liquid-glass-compatible.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * For now we only rely on frontend detection to see if the browser is compatible with Liquid Glass effects.
+ * Currently, only Chromium-based browsers (Chrome, Edge, Opera) are supported.
+ * Firefox and Safari have limited or no support for the required CSS features.
```
indietyp, line 4: Out of curiosity - what are those features? and can we work around that?

libs/@hashintel/ds-components/src/playground/example-article.tsx
```diff
@@ -0,0 +1,254 @@
...
+    >
+      <h1
+        style={{
+          fontSize: "7rem",
+          lineHeight: "7rem",
+          marginBottom: "2rem",
+          color: "#1a365d",
+        }}
+      >
+        Understanding Optics and Refraction
```
vilkinsons, line 21: @kube Where did this example article/content come from? (Fun!) 😅

libs/@hashintel/ds-components/src/stories/Intro.mdx
```diff
@@ -0,0 +1,164 @@
...
+- [Ark UI](https://ark-ui.com/) for accessible and unstyled components
+- [Panda CSS](https://panda-css.com/) for utility-first styling and design tokens
+
+> **Note:** This library is a work in progress and is currently just a proof-of-concept of feasability of a Design System with refractive effects.
+
+To understand technical aspect of these refraction effects, you can refer [to this article](https://kube.io/blog/liquid-glass-css-svg/).
+
+# Current Limitations
+
+- Only Chrome supports the CSS features required to render the refractive effects.
```
vilkinsons: Can we track/add links to the relevant [caniuse](https://caniuse.com/) pages for each of the relevant features?

vilkinsons: Let's add an inline jumplink to the later section on "fallback for non-supporting browsers" here (as I was wondering about this... but later on see it's been addressed... thus pre-empting eager beavers like me and improving discoverability for lazy folks).

libs/@hashintel/component-library/src/stories/Intro.mdx
```diff
@@ -0,0 +1,164 @@
...
+- Use classical gaussian blur for background blurring instead of the refraction effect.
+- Use progressive gaussian blur, where the blur increases with the distance from the center of the component.
+  > This can be achieved using multiple layers of gaussian blur, along with gradient-based masking.
+  > Though, this can be really slow on performance, particularly on Safari, so it should be used with caution.
+- Use a solid color or gradient background.
+
+## User Preference and Performance Considerations
+
+For preferences, accessibility, and performance considerations, especially on devices that may struggle with rendering refractive effects, it's important to provide a fallback design.
+It might be a good idea to provide a toggle/slider in the application settings to switch between the refractive design and the fallback design, so users can choose based on their preferences/accessibility and device performance.
```
vilkinsons:
```suggestion
For preferences, accessibility, and performance considerations, especially on devices that may struggle with rendering refractive effects, it's important to provide a fallback design.

Where practical and appropriate, users may be provided with a preferences toggle or slider in an application's "Appearance" settings enabling them to switch between fully refractive (Liquid Glass) and fallback (degraded) designs, allowing for choice based on personal style/accessibility preferences or according to device performance.
```

libs/@hashintel/ds-components/src/lib/maps/calculate-circle-map.ts
```diff
@@ -0,0 +1,100 @@
...
+        : 0;
+
+      const distanceToCenterSquared = x * x + y * y;
+
+      const isInBezel =
+        distanceToCenterSquared <= radiusPlusOneSquared &&
+        distanceToCenterSquared >= radiusMinusBezelSquared;
+
+      // Process pixels that are in bezel or near bezel edge for anti-aliasing
+      if (isInBezel) {
```
indietyp, line 75: This looks very similar to the displacement map code? Would it make sense to factor the common code out? Or does that not make sense?

libs/@hashintel/component-library/src/stories/Intro.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+This still needs to be debated with the design team.
```
vilkinsons:
```suggestion
The ability of users to opt-in or opt-out of Liquid Glass-style rendering may differ between usage environments. In HASH we will generally seek to provide this alongside other appearance-related user settings such as: **(information) density**, i.e. the compactness/spacing of content; **font size**, for smaller/larger display of text; **curvature**, the roundness of UI corner elements for sharper or softer displays; etc.

Similarly, usage of performance-intensive visual styles may be dynamically varied or reduced based on the application/context.
```

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
```diff
@@ -0,0 +1,483 @@
+import type { ImageData } from "canvas";
```
indietyp, line 1: don't these need the `"use client"` pragma because they can only be used on the client side and not be rendered during RSC? (primarily because of the `useEffect`)

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
```diff
@@ -0,0 +1,483 @@
...
+  width: MotionValue<number>;
+  height: MotionValue<number>;
+  result: string;
+  hideTop?: boolean;
+  hideBottom?: boolean;
+  hideLeft?: boolean;
+  hideRight?: boolean;
+};
+
+const CompositeParts: React.FC<CompositePartsProps> = memo(
```
indietyp, line 119: I wonder - do we need all properties to be motion values? Considering that e.g. `ImageData` is not something that's going to be animated I don't know why it would need to be a `MotionValue`.

Instead of manual memo, I think we should maybe use the react compiler: https://react.dev/reference/react-compiler/compiling-libraries

libs/@hashintel/ds-components/src/playground/example-article.tsx
kube: The whole article is full Copilot/Claude. :)

libs/@hashintel/component-library/src/components/Button/button.tsx
kube: Yes, scaffold of this component was done by Copilot, but I still need to clean it up.

Will do it the React 19 way.

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
kube: Yes they maybe need need `"use client"`, but for now I don't know how an external library built with Vite integrates with NextJS/RSC. (I mainly use React-Router, and haven't used NextJS in a while)

But in any case, they will be rendered on server, with classical SSR.

What would be a good NextJS package to start integrating this library and check the behaviour?

libs/@hashintel/ds-components/src/lib/displacement-map.ts
kube: Actually we iterate over very few dead zones when using the FlexibleFilter.

`<FlexibleFilter />` generates an image with:
- 4 corners (radius * radius)
- 4 sides (radius * 3)
- 3 dead-pixels in the middle (3 * 3 pixels) (could even be 1 pixel possibly)

So the dead zone in this case is only of 9 pixels, which do not enter the `isInBezel` condition.

libs/@hashintel/ds-components/src/lib/displacement-map.ts
kube: That's one of the optimizations I thought about, but it still requires to do:
- 1. Apply symmetry to the corner
- 2. Rotate colors

My initial idea was to do that in the SVG filter itself, but the only thing achievable with it is (2).

There are many ways to optimise this part, like leveraging SVG Filters operations, but they seem quite limited.

One idea is to instead only create one image with each pixel being the `distanceToBorder`, and map the displacement in the SVG Filter directly.

This would allow changing optics parameters, without recalculating the image.

But not sure how to achieve that, maybe with https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/feComponentTransfer.

Also, even the generation of the `distanceToBorder` could be possibly done with https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/feConvolveMatrix, but I never defined a Convolution Matrix.

Then we could handle arbitrary images, without even having to do image generation step.

libs/@hashintel/component-library/src/components/Bar/bar.tsx
```diff
@@ -0,0 +1,93 @@
+import { useId } from "react";
+
+import { Filter } from "../../lib/flexible-filter";
+import { CONVEX } from "../../lib/surface-equations";
+import { useMotionResizeObserver } from "../../lib/use-motion-resize-observer";
+
+export type BarProps = React.PropsWithChildren<{
```
vilkinsons: I'm not 100% sure `Bar` is the correct name for this component versus something like `Panel`,`BackgroundPanel`, or `Underlay`. cc @CiaranMn for thoughts

libs/@hashintel/ds-components/src/lib/maps/calculate-circle-map.ts
kube: Yeah I don't remember why I ended up just duplicating the code, certainly just to prototype quickly. But will factor it.

libs/@hashintel/ds-components/src/lib/use-is-liquid-glass-compatible.ts
kube: The main feature it relies on is the support of SVG filter for as `backdrop-filter`.

It seems to be in the CSS spec, but only Chrome supports it for now.

Firefox does not support it, but they opened a ticket:

https://bugzilla.mozilla.org/show_bug.cgi?id=1989910

Safari does not support it, and the browser itself is pretty slow with SVG/SVG Filters, even when used as `filter` property. (Which all browsers support)

libs/@hashintel/ds-components/src/lib/displacement-map.ts
indietyp: It's an interesting issue for sure, I was just looking for ways to do more/the same with less, considering that this is potentially running *a lot*, so every optimisation counts.

Another idea I had was using WebGL/WebGPU for the background instead of SVG, as that will be certainly faster, but I don't think you can achieve that as easily as you can with SVGs

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
indietyp: I don't think any of our current apps use RSC _yet_, so I don't think we have one for it. This will probably be something we'll just have to do one by one. Just a flag that with RSC it'd likely be a requirement.

> fun fact: react router does have experimental support for RSC - https://reactrouter.com/how-to/react-server-components

libs/@hashintel/ds-components/src/lib/image-data-to-url.ts
indietyp: discussed offline

libs/@hashintel/component-library/.storybook/main.ts
```diff
@@ -0,0 +1,20 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+import { dirname, join } from "path";
```
CiaranMn:
```suggestion
import { dirname, join } from "node:path";
```
linting will complain about this

libs/@hashintel/ds-components/package.json
```diff
@@ -0,0 +1,69 @@
...
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist",
```
CiaranMn, line 20: we'll want these in the published package, just adding now so I don't forget
```suggestion
    "dist",
    "CHANGELOG.md",
    "LICENSE-APACHE.md",
    "LICENSE-MIT.md",
    "LICENSE",
```

libs/@hashintel/component-library/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "lint:eslint": "eslint --report-unused-disable-directives .",
+    "lint:tsc": "tsc --noEmit",
+    "test": "vitest --run",
+    "test:watch": "vitest",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build"
+  },
+  "dependencies": {
+    "@ark-ui/react": "5.25.0",
+    "@hashintel/styled-system": "workspace:*",
```
CiaranMn: Once this is published, all dependencies will need to refer to a published package
```suggestion
    "@hashintel/styled-system": "0.0.1",
```

libs/@hashintel/component-library/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@pandacss/dev": "1.3.1",
+    "canvas": "3.2.0",
+    "motion": "12.23.19"
+  },
+  "devDependencies": {
+    "@local/eslint": "0.0.0-private",
+    "@local/tsconfig": "0.0.0-private",
+    "@storybook/addon-docs": "^9.1.8",
```
CiaranMn:
```suggestion
    "@storybook/addon-docs": "9.1.8",
```

libs/@hashintel/component-library/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@storybook/react-vite": "9.1.8",
+    "@testing-library/jest-dom": "^6.6.3",
+    "@testing-library/react": "^16.1.1",
+    "@testing-library/user-event": "^14.5.2",
```
CiaranMn: Prefer pinning
```suggestion
    "@testing-library/jest-dom": "6.6.3",
    "@testing-library/react": "16.1.1",
    "@testing-library/user-event": "14.5.2",
```

libs/@hashintel/ds-components/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@local/tsconfig": "0.0.0-private",
+    "@storybook/addon-docs": "^9.1.8",
+    "@storybook/react-vite": "9.1.8",
+    "@testing-library/jest-dom": "^6.6.3",
+    "@testing-library/react": "^16.1.1",
+    "@testing-library/user-event": "^14.5.2",
+    "@types/react": "19.1.13",
+    "@types/react-dom": "19.1.9",
+    "eslint": "9.36.0",
+    "eslint-plugin-storybook": "^9.1.8",
```
CiaranMn:
```suggestion
    "eslint-plugin-storybook": "9.1.8",
```

libs/@hashintel/component-library/panda.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from "@pandacss/dev";
+
+export default defineConfig({
+  // Whether to use css reset
+  preflight: true,
+
+  // Where to look for your css declarations
```
CiaranMn:
```suggestion
  // Where to look for css declarations
```

libs/@hashintel/component-library/src/components/Button/button.tsx
```diff
@@ -0,0 +1,47 @@
...
+            transitionDuration: "200ms",
+            overflow: "hidden",
+            paddingX: "3",
+            paddingY: "2",
+            fontSize: "base",
+          }),
+          className
+        )}
+        {...props}
+        style={style}
```
CiaranMn: nit: we're not using `style` anywhere else so there's no need to extract it from `props`, if we don't destructure it, it would be passed anyway

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+## Features
+- 🎨 Advanced visual effects (blur, specular highlights, refraction)
+- 📏 Customizable dimensions and radius
+- ⚙️ Configurable visual parameters
+- 🌟 Glass-like appearance with realistic lighting
+- 🔧 Flexible scale ratio for dynamic effects
+- 🖼️ Adjustable bezel width for frame customization
+- 🔍 Variable glass thickness for refraction depth control
+- ⚗️ Configurable refractive index for different material simulation
+- 🎛️ Built on Ark UI for accessibility and interaction
```
CiaranMn: I have a personal bias against emojis in lists because it's such an AI smell

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+const meta: Meta<typeof SegmentedControl> = {
+  title: "Component Library/SegmentedControl",
+  component: SegmentedControl,
+  tags: ["docsPage"],
+  parameters: {
+    layout: "centered",
+    docs: {
+      description: {
+        component: `
+A sophisticated segmented control component with advanced visual effects including blur,
```
CiaranMn: Personal bias against the flattering language AI injects into everything, I would keep it factual.
```suggestion
A segmented control component with visual effects including blur, 
```

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
```diff
@@ -0,0 +1,224 @@
...
+      backgroundColor: "whiteAlpha.10",
+    },
+    userSelect: "none",
+  });
+
+  const rootBackdropStyles = css({
+    position: "relative",
+    display: "flex",
+    alignItems: "center",
+    backgroundColor: "gray.10/20",
```
CiaranMn, line 79: What does the `/20` here do / mean?

libs/@hashintel/ds-components/src/lib/displacement-map.ts
kube: I considered WebGL/WebGPU, but the problem is that it has no access to the rendered DOM behind it, so people end up doing snapshots of the whole DOM and using it as background of the WebGL canvas. (which is far uglier in my opinion than the current SVG solution)

There is one solution that Chrome is working on, and with which I plan to do a blog post: https://github.com/WICG/html-in-canvas

But it's still a proposal, available behind a flag in Chrome though.

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { motion } from "motion/react";
+import { useId } from "react";
+
+import { Filter } from "../../lib/filter";
+import { CONVEX } from "../../lib/surface-equations";
+
+const BaseSliderThumb = motion.create(BaseSlider.Thumb);
+const BaseSliderRange = motion.create(BaseSlider.Range);
+
+const SLIDER_WIDTH = 330;
```
CiaranMn: I'm assuming this will become customisable

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
```diff
@@ -0,0 +1,189 @@
...
+const SLIDER_HEIGHT = 5;
+const THUMB_WIDTH = 18;
+const THUMB_HEIGHT = 16;
+const THUMB_RADIUS = THUMB_HEIGHT / 2;
+const THUMB_BEZEL_WIDTH = THUMB_RADIUS * 0.7;
+const THUMB_ACTIVE_SCALE = 2.2;
+
+const TRACK_COLOR = "#89898F66";
+const TRACK_ACTIVE = "#0f94c4ff";
```
CiaranMn: Will these be replaced by design system tokens? Presumably they'll need to change, e.g. depending on dark/light mode

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
```diff
@@ -0,0 +1,189 @@
...
+  onChange,
+}) => {
+  const filterId = `thumb-filter-${useId()}`;
+
+  const sliderHeight = SLIDER_HEIGHT;
+  const sliderWidth = SLIDER_WIDTH;
+  const thumbWidth = THUMB_WIDTH;
+  const thumbHeight = THUMB_HEIGHT;
+  const thumbRadius = THUMB_RADIUS;
+  const thumbBezelWidth = THUMB_BEZEL_WIDTH;
```
CiaranMn: What's the motivation for assigning these to new variables?

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
```diff
@@ -0,0 +1,189 @@
...
+  defaultValue?: number;
+  specularOpacity?: number;
+  specularSaturation?: number;
+  glassThickness?: number;
+  refractiveIndex?: number;
+  bezelWidth?: number;
+  blurLevel?: number;
+  label?: string;
+  showValueText?: boolean;
+  onChange?: (value: number[]) => void;
```
CiaranMn: Assuming these props will change, at the moment there's no `value` so it can't be used as a controlled component

libs/@hashintel/styled-system/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@hashintel/styled-system",
+  "version": "0.0.1",
+  "description": "HASH Panda Styled System",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hashintel/hash.git",
+    "directory": "libs/@hashintel/ui-preset"
```
CiaranMn:
```suggestion
    "directory": "libs/@hashintel/styled-system"
```

libs/@hashintel/styled-system/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@hashintel/styled-system",
+  "version": "0.0.1",
+  "description": "HASH Panda Styled System",
```
CiaranMn:
```suggestion
  "description": "HASH PandaCSS Styled System",
```

libs/@hashintel/styled-system/package.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "./jsx": {
+      "types": "./dist/jsx/index.d.ts",
+      "require": "./dist/jsx/index.mjs",
+      "import": "./dist/jsx/index.mjs"
+    },
+    "./styles.css": "./dist/styles.css"
+  },
+  "dependencies": {
+    "@hashintel/ui-preset": "0.0.0-private"
```
CiaranMn: Needs to be published if mentioned here
```suggestion
    "@hashintel/ui-preset": "0.0.1"
```

libs/@hashintel/ui-preset/README.md
```diff
@@ -0,0 +1,5 @@
+# @hashintel/ui-preset
+
+A Panda CSS preset for HashIntel's design system, providing a consistent and reusable set of design tokens and styles for building user interfaces.
```
CiaranMn:
```suggestion
A Panda CSS preset for HASH's design system, providing a consistent and reusable set of design tokens and styles for building user interfaces.
```

libs/@hashintel/ui-preset/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@hashintel/ui-preset",
+  "version": "0.0.0-private",
+  "private": true,
```
CiaranMn:
```suggestion
  "version": "0.0.1",
```

libs/@hashintel/ds-theme/src/index.ts
```diff
@@ -0,0 +1,559 @@
+import { definePreset } from "@pandacss/dev";
+
+export default definePreset({
```
CiaranMn, line 3: Not going to comment on this as I assume it still needs updating to populate `semanticTokens` and to rename core `tokens` to be private

package.json
```diff
@@ -102,6 +102,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "0.5.1",
     "@changesets/cli": "patch:@changesets/cli@npm%3A2.26.0#~/.yarn/patches/@changesets-cli-npm-2.26.0-49d5c5f72d.patch",
+    "@pandacss/dev": "1.3.1",
```
CiaranMn: Should this be at the root? What for if so?

libs/@hashintel/component-library/src/components/Bar/bar.tsx
CiaranMn: I would name it based on its function, I guess it's a `RefractiveSomething`? `Container`, `Underlay`, `Background`?

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
kube: `ImageData` is animated in the Storybook through the controls, and could possibly be animated in the future.

We should certainly use the React Compiler at some point, but last time I checked there were issues with third party libraries like Motion.

IMO React Compiler is good, but for animating properties of DOM nodes, Motion is better as it lives outside of React lifecycle, and does not require re-rendering (only mutates the DOM properties of the element it is in charge of).

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.stories.tsx
kube: Haha. I hate emojis too, but yeah this was generated by Copilot. Gonna clean it.

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
kube: `[colorToken]/[opacity]`. This is the same convention as in Tailwind.

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
indietyp: Those were two different points sorry for bunching them up 😅. Didn't know it was inside storybook! So doesn't apply thank you for the clarification.

re: react compiler - I haven't made as extensive use of motion in my projects, but since the rc (and release) I haven't had any problems so might be worth a shot in the future - there's also an eslint plugin that helps with issues

nothing that needs to be done now just wanted to flag it because I saw the memo!

libs/@hashintel/component-library/src/components/Bar/bar.tsx
kube: I renamed it to `RefractivePane`, WDYT?

libs/@hashintel/component-library/src/components/Button/button.tsx
kube: Removed `transition` for now.
I keep it simple for this first PR, and we will iterate.

libs/@hashintel/component-library/src/components/Button/button.tsx
kube: Thanks!
I used `_disabled`, and we will iterate to make this properly.

(I'm creating tickets in Linear for enhancements)

libs/@hashintel/component-library/src/components/Button/button.tsx
kube: Done

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
kube: The issue here is that we need to read the radius from React to calculate the refracted backdrop.

So `radius` needs to be defined in the JS world, and is passed to `style` to apply `borderRadius`.

I don't see how to do the other way around? (Except by watching computed properties, but would be quite heavy for many elements, as I don't think we can watch for a particular property)

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
kube: It could, but given that other elements also need properties calculated with `ROOT_PADDING` (e.g. `borderRadius: radius - ROOT_PADDING`), I try to put everything at same level.

And PandaCSS is better when given Design Tokens than constants/variables, where it just creates a className for this constant which seems just worse (IMO) than hardcoding in `style`.

kube: I will create a ticket about that.

libs/@hashintel/ds-theme/src/index.ts
kube: Yes we still haven't defined the semantic tokens, so it's still early to make the core tokens private.

I'm creating the ticket to do that.

package.json
kube: Thanks, removed!

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
kube: I moved everything in class definitions.

Just kept the `borderRadius`, `padding` and `backdropFilter` which depend only on the Component props.

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
kube: I was planning to make these parameters possibly configurable via props, to allow Ciarán to play with them in Storybook.

But we discussed that and for now he will change constants locally on the repo to test.

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
kube: Fixed.

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
kube: I removed this, and we will allow overriding colors via className.

libs/@hashintel/ds-components/src/components/Slider/slider.tsx
kube: Fixed

libs/@hashintel/ds-components/src/components/Switch/switch.tsx
kube: Removed. Thanks!

libs/@hashintel/ds-components/src/stories/Intro.mdx
kube: Done

kube: I could not find a simple Caniuse link for this particular sub-feature, so ended up adding bug trackers links:

> See [Firefox bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1989910), and [Safari bug report](https://bugs.webkit.org/show_bug.cgi?id=245510) to track progress on other browsers.

libs/@hashintel/ds-components/src/components/SegmentedControl/segmented-control.tsx
indietyp: I missed that part, thank you! I was primarily concerned about inline styles being defined, but if we end up using it anyway in JS it doesn't make too much sense to have it as a data property as well.

libs/@hashintel/ds-components/src/lib/displacement-map.ts
kube: Given the fact that this calculation is done only 127 times, it would not be worth it as here we would end up with a bad approximation of the normal, and the time taken by this reported to the global calculation in negligible.

Though, there are certainly ways to optimize the sampling itself here. (127 evenly distributed rays is not necessarily the best choice, as variations will be higher closer the corner).

libs/@hashintel/ds-components/src/components/RefractivePane/refractive-pane.tsx
```diff
@@ -0,0 +1,95 @@
...
+        height={trackedMotionHeight}
+        radius={radius}
+        bezelWidth={bezelWidth}
+        glassThickness={glassThickness}
+        refractiveIndex={refractiveIndex}
+        bezelHeightFn={CONVEX}
+      />
+
+      <div
+        ref={divRef}
```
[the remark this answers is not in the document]
kube: Thanks Vercel!

libs/@hashintel/ds-components/src/lib/maps/calculate-circle-map.ts
kube: I factored it by creating a `calculateCircleMap` that does all the heavy-lifting and accepts a `processPixel` callback.

Will be a bit less performant, but should be quickly optimized by JIT.

libs/@hashintel/ds-components/src/lib/displacement-map.ts
kube: It's an array of 128 numbers, not sure it would change much to change this here.

I kept it simple for now, the calculations certainly need to be optimized, but would have to be re-thought at a more global level.

libs/@hashintel/ds-theme/turbo.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "cache": false
+    }
```
TimDiekmann: What is the reasoning to disable caching for the `build` step?

.vscode/mcp.json
TimDiekmann, line 1: I'd prefer if we would not add this MCP servers globally. It is only relevant when building design system components, and not generally applicable.

libs/.gitignore
```diff
@@ -1 +1,5 @@
 Cargo.lock
+
+## Panda
+*storybook.log
+storybook-static
```
TimDiekmann: I think these should be in the root `.gitignore` instead, they are not specific to libraries.

.vscode/mcp.json
kube: Yeah sorry I messed with this PR at some point and don't know how ended up versioning this.

libs/.gitignore
kube: Moved

libs/@hashintel/ds-components/src/components/Button/button.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export interface ButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  radius?: number;
+  specularOpacity?: number;
+  specularSaturation?: number;
+  blurLevel?: number;
+  glassThickness?: number;
+  refractiveIndex?: number;
+  bezelWidth?: number;
```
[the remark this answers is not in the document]
kube: This will be tackled in H-5420.

libs/@hashintel/ds-theme/turbo.json
kube: I asked copilot help on generating `turbo.json` files, so maybe not the correct way to do it.

There is no build step for now in this package, dependent packages directly read the `index.ts` file. (only `ds-helpers` imports it)

It does not cache the build, as it directly reads from the versioned file.

libs/@hashintel/ds-components/src/lib/image-data-to-url.ts
```diff
@@ -0,0 +1,17 @@
+import { createCanvas, type ImageData } from "canvas";
```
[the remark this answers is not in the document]
kube: The SSR should work with `canvas`. (https://kube.io uses it for SSR/SSG)
If there is an issue at some point, we will fix it at this moment.

libs/@hashintel/ds-theme/turbo.json
TimDiekmann: We inherit everything from the top-level `turbo.json`. Unless you have a specific reason to change a configuration (such as not caching a step), no `turbo.json` is required. Even if you don't have a build-step specified, `turbo` generates one without content (to properly resolve dependencies, e.g. if `build` depends on `codegen` and you only have `codegen` specified it should still be executed if another tasks depends on `build` only).

In your case you probably can simply remove this file.

.vscode/mcp.json
TimDiekmann: No worries :)

libs/.gitignore
TimDiekmann: Great, thanks!

libs/@hashintel/ds-components/.gitignore
```diff
@@ -0,0 +1,2 @@
+# PandaCSS
+styled-system
```
TimDiekmann: nitpick: Trailing newline
```suggestion
styled-system

```

vilkinsons:
Will provide legal review once outstanding comments addressed/otherwise marked resolved.

libs/@hashintel/ds-theme/turbo.json
kube: Removed

libs/@hashintel/ds-components/src/components/Button/button.tsx
```diff
@@ -0,0 +1,43 @@
...
```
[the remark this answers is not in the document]
kube: This will be addressed in a dedicated ticket: H-5420

PR #7817, opened 2025-09-25
BE-130: HashQL: Move from `ariadne` to `annotate-snippets`

indietyp:

`ariadne` currently has had some longstanding issues, with near to no maintenance and a big refactor coming since at least Sep. '24, with no changes happening publicly.

Since our initial implementation a new crate has gained prominence `annotate-snippets`, which is now also used in the rust compiler. This gives us more features and more stability.

To be able to express these new capabilities the following changes have been done (to abstract away from any one rendering engine):

* rendering is now behind a `render` cfg
* we do not expose `annotate-snippets`
* reorganised the crate
* `Help` + `Note` => `Message`
* Messages can now have suggestions (not used yet, to be done in follow-up)
* `resolve` has been abstracted away
* `Source` is no longer an `ariadne` thing, but a `hashql-diagnostics` one
* `AbsoluteDiagnosticSpan` -> `SourceSpan`
* usage of newtypes
* rendering can no longer error out (even with resolving) this now will create a new diagnostic group with an error instead
* ability to fully style elements (where annotate-snippets allows for it)
* Primary labels (at least one label) are now enforced
* full documentation with examples

This PR:

The changes in this PR:

The changes in this PR:


libs/@local/hashql/syntax-jexpr/src/parser/object/error.rs
```diff
@@ -251,108 +249,86 @@ Empty objects don't have semantic meaning in J-Expr.
...
+    let (&first, rest) = spans.split_first().expect("Should have at least one span");
 
-    for (index, &span) in spans.iter().rev().enumerate() {
-        let message = if index == 0 {
-            "Remove this trailing comma"
-        } else {
-            "... and this trailing comma"
-        };
+    let mut diagnostic = Diagnostic::new(ObjectDiagnosticCategory::TrailingComma, Severity::Error)
+        .primary(Label::new(first, "Remove this trailing  comma"));
```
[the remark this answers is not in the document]
indietyp: tales of a broken keyboard 😔

libs/@local/hashql/syntax-jexpr/src/lexer/snapshots/hashql_syntax_jexpr__lexer__test__emoji_scientist.snap
```diff
@@ -3,12 +3,9 @@ source: libs/@local/hashql/syntax-jexpr/src/lexer/mod.rs
...
-   │
- 1 │ 👩🏾‍🔬
-   │ ┬─  
-   │ ╰─── Unrecognized character
-   │ 
-   │ Help: J-Expr only supports standard JSON syntax. Make sure you're using valid JSON tokens like {}, [], strings, numbers, true/false, or null.
-───╯
+error[lexer::invalid-character]: Invalid character
+  ╭▸ 
+1 │ 👩🏾🔬
```
TimDiekmann, line 8: That's a different emoji

Cargo.toml
```diff
@@ -94,27 +94,25 @@ hashql-syntax-jexpr.path            = "libs/@local/hashql/syntax-jexpr"
...
+aws-sdk-s3        = { version = "=1.104.0", default-features = false }
+aws-types         = { version = "=1.3.8", default-features = false }
+axum              = { version = "=0.7.5" }
+axum-core         = { version = "=0.5.0" }
+bitvec            = { version = "=1.0.1", default-features = false }
+bumpalo           = { version = "=3.17.0", default-features = false, features = ["allocator_api"] }
+bytes             = { version = "=1.10.1" }
+bytes-utils       = { version = "=0.1.4", default-features = false }
+camino            = { version = "=1.1.12", default-features = false }
+cargo_metadata    = { version = "=0.22.0", default-features = false }
```
TimDiekmann, line 115: Were `approx` and `anstyle-yansi` unused before or is this because of these changes?

libs/@local/hashql/diagnostics/src/diagnostic/label.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// let mut labels = Labels::new(primary);
+    /// labels.push(Label::new(20..25, "related issue"));
+    /// labels.push(Label::new(30..35, "context"));
+    ///
+    /// let messages: Vec<&str> = labels.iter().map(|label| label.message()).collect();
+    /// assert_eq!(messages, vec!["main error", "related issue", "context"]);
+    /// ```
+    pub fn iter(&self) -> impl Iterator<Item = &Label<S>> {
+        self.labels.iter()
+    }
```
TimDiekmann, line 301: Maybe worth implementing `impl IntoIterator` for `&Self` and `Self`

libs/@local/hashql/diagnostics/src/diagnostic/message.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// let mut messages: Messages<()> = Messages::new();
+    /// messages.push(Message::note("Background information"));
+    /// messages.push(Message::help("Actionable advice"));
+    ///
+    /// let texts: Vec<&str> = messages.iter().map(|msg| msg.message()).collect();
+    /// assert_eq!(texts, ["Background information", "Actionable advice"]);
+    /// ```
+    pub fn iter(&self) -> impl Iterator<Item = &Message<S>> {
+        self.messages.iter()
+    }
```
TimDiekmann, line 352: Same here, maybe worth implementing `IntoIterator`

libs/@local/hashql/diagnostics/src/diagnostic/suggestion.rs
```diff
@@ -0,0 +1,272 @@
...
+///
+/// // Add missing punctuation
+/// let punctuation = Patch::new(30..30, ";");
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Patch<S> {
+    span: S,
+    replacement: Cow<'static, str>,
+}
```
TimDiekmann, line 35: Why are the fields private?

Cargo.toml
indietyp: Approx seems to have been unused; when I use `cargo add`, it removes unused dependencies from the `Cargo.toml`.

libs/@local/hashql/syntax-jexpr/src/lexer/snapshots/hashql_syntax_jexpr__lexer__test__emoji_scientist.snap
indietyp: This is a regression in annotate-snippets, 👩🏾‍🔬 is:
WOMAN (👩) + EMOJI MODIFIER FITZPATRICK DARK + ZERO WIDTH JOINER +  MICROSCOPE (🔬)

It seems that annotate-snippets is unable to render ZWJ (although it supports some other Unicode sequences that Ariadne does _not_ support).

I just took a look, this is actually intended, with an explanation here: https://github.com/rust-lang/annotate-snippets-rs/blob/ebbafa89185863bebe35ac566a691b3bb74fe6a6/src/renderer/render.rs#L2536

It seems to be to used to increase compatibility across terminals (and spans) which may not be able to render ZWJ correctly.

libs/@local/hashql/diagnostics/src/diagnostic/message.rs
indietyp: tracked in https://linear.app/hash/issue/H-5404/hashql-implement-intoiterator-on-andself-in-hashql-diagnostics

libs/@local/hashql/diagnostics/src/diagnostic/label.rs
indietyp: tracked in https://linear.app/hash/issue/H-5404/hashql-implement-intoiterator-on-andself-in-hashql-diagnostics

libs/@local/hashql/diagnostics/src/diagnostic/suggestion.rs
indietyp: same reason they are in `Label`, to remove API surface, you can still access span (via `map_span`) if needed I would add other methods to it, or make them public, but would keep them private for now.

PR #7818, opened 2025-09-25
BE-127: Allow including permission in entity query responses

TimDiekmann:

This PR moves entity permission checking logic from the Node.js API layer down to the Graph database layer. Previously, when clients requested entity permissions alongside query results, the Node API would make separate permission checks for each entity after receiving the subgraph from the Graph API. This created N+1 query performance issues and violated separation of concerns. 

Now the Graph API natively supports including permissions in entity query responses through an `includePermissions` flag, making permission checks atomic with the main query and improving performance.

## 🔗 Related links

- [Linear Issue BE-127: Allow including permission in entity query responses](https://linear.app/hash/issue/BE-127/allow-including-permission-in-entity-query-responses)
- Related to the overall entity API refactoring effort (BE-118)

### Graph API & Store Layer
- **Added `includePermissions` flag** to `QueryEntitySubgraphRequest` and `QueryEntitiesRequest` in OpenAPI spec
- **Implemented `EntityPermissions` type** containing `update` permissions array in the Graph store
- **Added permission checking logic** directly in `query_entities()` and `query_entity_subgraph()` methods
- **Updated response structures** to include optional `permissions` and `entityPermissions` fields

### TypeScript SDK
- **Replaced `UserPermissionsOnEntities`** type with new `EntityPermissionsMap` type
- **Updated `queryEntities()` and `queryEntitySubgraph()`** to handle new permission structure
- **Removed restriction** that previously threw errors for `includePermissions: true`

### Node.js API Layer  
- **Eliminated manual permission checking** in `queryEntitySubgraphResolver`
- **Simplified resolvers** to use new Graph API permission responses directly
- **Updated `meResolver`** to use new query structure instead of separate permission calls

### Frontend Updates
- **Updated permission checks** across all components to use new format: 
  - Old: `permissions[entityId]?.edit` 
  - New: `(permissions[entityId]?.update ?? []).length > 0`
- **Updated type imports** from `@local/hash-graph-sdk/authorization` to `@local/hash-graph-sdk/entity`
- **Fixed permission logic** in block loader, entity editor, page components, and user management

This PR:

The changes in this PR:

- [x] are internal and do not require a docs change (API changes are backward-compatible)

The changes in this PR:

## ⚠️ Known issues

- The new permission structure only includes `update` permissions currently. Other permission types (read, delete) may need to be added in future iterations depending on frontend requirements.
- Some frontend code still uses optional chaining defensively even though the new structure should be more reliable.

## 🐾 Next steps

- Monitor performance improvements from eliminating N+1 permission queries
- Consider extending `EntityPermissions` to include other action types beyond `update` if needed
- Evaluate if similar permission optimization can be applied to other entity operations


apps/hash-api/src/graphql/resolvers/knowledge/user/me.ts
```diff
@@ -1,41 +1,62 @@
...
+            : []),
+        ],
+      },
       graphResolveDepths: {
+        ...zeroedGraphResolveDepths,
         hasLeftEntity,
         hasRightEntity,
       },
+      temporalAxes: currentTimeInstantTemporalAxes,
+      includeDrafts: false,
```
TimDiekmann:
```suggestion
      includeDrafts: !!draftId,
```

apps/hash-frontend/src/pages/shared/entity.tsx
```diff
@@ -366,9 +366,9 @@ export const Entity = ({
     !!draftEntity?.metadata.archived ||
     !!proposedEntitySubgraph ||
     (!draftLocalEntity &&
-      !queryEntitySubgraphData?.queryEntitySubgraph.entityPermissions?.[
+      !!queryEntitySubgraphData?.queryEntitySubgraph.entityPermissions?.[
```
TimDiekmann, line 369:
```suggestion
      !queryEntitySubgraphData?.queryEntitySubgraph.entityPermissions?.[
```

PR #7820, opened 2025-09-25
H-5371: HashQL: Create `hashql-mir` crate

indietyp:

Creates the `hashql-mir` crate, to be used for further lowering.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.github/actions/prune-repository/action.yml
+++ b/.github/actions/prune-repository/action.yml
@@ -26,6 +26,9 @@ runs:
           if echo "$DEPENDENCIES" | grep -q "@rust/hashql-hir"; then
             SCOPES="$SCOPES @rust/hashql-compiletest"
           fi
+          if echo "$DEPENDENCIES" | grep -q "@rust/hashql-mir"; then
+            SCOPES="$SCOPES @rust/hashql-compiletest"
+          fi
           if echo "$DEPENDENCIES" | grep -q "@rust/hashql-eval"; then
             SCOPES="$SCOPES @rust/hashql-compiletest"
           fi


libs/@local/hashql/mir/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn it_works() {
+        assert_eq!(2, 2); // if this isn't true, then something went *horribly* wrong in the universe.
+    }
+}
```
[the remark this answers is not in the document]
indietyp: currently a stub, to be removed once we have actual content to be tested

vilkinsons (approved):
Legal ✅

libs/@local/hashql/mir/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: I'm not convinced. We should add `assert_ne!(2, 3);`

libs/@local/hashql/mir/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+workspace = true
+
+[[test]]
+name    = "compiletest"
+harness = false
+
+[package.metadata.sync.turborepo]
+ignore-dev-dependencies = [
+    "@rust/hashql-compiletest",
+]
```
TimDiekmann, line 24: Could you please add this to the `prune-repository` action?
```bash
if echo "$DEPENDENCIES" | grep -q "@rust/hashql-mir"; then
  SCOPES="$SCOPES @rust/hashql-compiletest"
fi
```

indietyp: I knew I forgot something! That was it

TimDiekmann (changes requested):
Looks good, we just need the workaround for `turbo prune`

libs/@local/hashql/mir/Cargo.toml
indietyp: done in [c4cc1c1](https://github.com/hashintel/hash/pull/7820/commits/c4cc1c1c65166c10bc2491885dab8485e659a2c3)

TimDiekmann (approved):
We can ignore the semgrep warning. Thanks!

PR #7821, opened 2025-09-26
BE-131: Add `Exists` filter to replace null-check patterns in query system

TimDiekmann:

This PR introduces a dedicated `Exists` filter for queries, replacing the unintuitive pattern of using `Equal` with `None` to check for null values. The implementation also refactors the filter API to use non-optional expressions for `Equal` and `NotEqual` operations, making the query system more ergonomic and type-safe.

The primary motivation is to address codegen issues and improve developer experience when checking for the existence of values in query paths. Previously, developers had to use `Filter::Equal(Some(path), None)` to check if a field was null, which was counterintuitive and caused problems in code generation.

## 🔗 Related links

- Linear issue: [BE-131: Introduce an `exists` filter for queries](https://linear.app/hash/issue/BE-131/introduce-an-exists-filter-for-queries)

## 🔍 What does this change?

- **Adds `Exists` filter variant** to `Filter<'p, R>` enum with a `path` field for checking value existence
- **Refactors `Equal` and `NotEqual` filters** to use non-optional `FilterExpression` parameters instead of `Option<FilterExpression>`
- **Updates SQL transpilation** to handle `Exists` filter as `IS NULL` and `NOT(Exists(...))` as `IS NOT NULL`
- **Migrates existing null-check patterns** across the codebase to use the new `Exists` filter
- **Updates all filter constructors** throughout the graph store, API, and frontend to use the simplified signatures
- **Removes redundant null-handling logic** that was previously required for optional filter expressions
- **Updates OpenAPI specification** to reflect the new filter structure
- **Adds comprehensive test coverage** for the new `Exists` filter functionality

### Key architectural changes:

- **`libs/@local/graph/store/src/filter/mod.rs`**: Core filter enum definition and implementation
- **`libs/@local/graph/postgres-store/src/store/postgres/query/condition.rs`**: SQL transpilation logic for new filter types
- **`libs/@local/hashql/eval/src/graph/read/filter.rs`**: HashQL compilation updates
- **Frontend components**: Updated to use simplified filter constructors
- **Tests**: New test cases for `Exists` filter and updated existing tests

### Migration patterns applied:

```rust
// Before (checking for null/non-existence)
Filter::Equal(Some(FilterExpression::Path { path }), None)

// After (checking for existence)
Filter::Exists { path }

// Before (checking for not-null/existence)  
Filter::NotEqual(Some(FilterExpression::Path { path }), None)

// After (checking for non-existence)
Filter::Not(Box::new(Filter::Exists { path }))

// Before (regular equality)
Filter::Equal(Some(lhs), Some(rhs))

// After (regular equality - simplified)
Filter::Equal(lhs, rhs)
```

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/condition.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/condition.rs
@@ -10,8 +10,9 @@ pub enum Condition {
     All(Vec<Self>),
     Any(Vec<Self>),
     Not(Box<Self>),
-    Equal(Option<Expression>, Option<Expression>),
-    NotEqual(Option<Expression>, Option<Expression>),
+    Equal(Expression, Expression),
+    NotEqual(Expression, Expression),
+    Exists(Expression),
     Less(Expression, Expression),
     LessOrEqual(Expression, Expression),
     Greater(Expression, Expression),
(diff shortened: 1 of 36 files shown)


libs/@local/graph/postgres-store/src/store/postgres/query/condition.rs
```diff
@@ -65,23 +66,24 @@ impl Transpile for Condition {
                 Ok(())
             }
             Self::Not(condition) => {
-                fmt.write_str("NOT(")?;
-                condition.transpile(fmt)?;
-                fmt.write_char(')')
+                if let Self::Exists(path) = &**condition {
```
indietyp, line 69: This looks like it doesn't handle nested paths, is that correct? As in: `not(all([exists(...), ...]))` would fail.

TimDiekmann: Yes, that's intended. Semantically it's the same if you negate a whole expression or use the more readable version. I.e. `NOT (x IS NULL)` is the same as `x IS NOT NULL`.

indietyp: Great just wanted to verify. NULL in SQL is a bit weird so I wasn't sure.

indietyp (approved):

libs/@local/graph/postgres-store/src/store/postgres/query/condition.rs
TimDiekmann: It is indeed...

PR #7862, opened 2025-10-09
BE-133: HashQL: Convert HIR to HIR(ANF)

indietyp:

This PR implements HIR normalization, converting HIR to Administrative Normal Form (ANF). This transformation differentiates between values and computations, making the code more suitable for optimization and code generation.

## 🔍 What does this change?

- Adds a new `normalization.rs` module to the HIR codebase that transforms HIR into ANF
- Updates the `lower` function to apply normalization as the final step in the lowering process
- Modifies the `HirContext` parameter in the `lower` function to be mutable to support normalization
- Adds comprehensive test cases for the normalization transformation
- Updates existing test outputs to reflect the normalized form

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Added a new test suite `hir/lower/normalization` with multiple test cases covering different aspects of normalization
- Updated existing test outputs to reflect the normalized form

## ❓ How to test this?

1. Run the test suite to verify normalization works correctly
2. Check that existing functionality continues to work with the normalized HIR

### DIFF
--- a/libs/@local/hashql/compiletest/src/suite/hir_lower_normalization.rs
+++ b/libs/@local/hashql/compiletest/src/suite/hir_lower_normalization.rs
@@ -8,7 +8,10 @@ use hashql_core::{
     r#type::environment::Environment,
 };
 use hashql_hir::{
-    context::HirContext, intern::Interner, lower::normalization::Normalization, node::Node,
+    context::HirContext,
+    intern::Interner,
+    lower::normalization::{Normalization, NormalizationOptions},
+    node::Node,
     pretty::PrettyPrintEnvironment,
 };
 
@@ -27,7 +30,7 @@ pub(crate) fn hir_lower_normalization<'heap>(
 ) -> Result<Node<'heap>, SuiteDiagnostic> {
     let node = hir_lower_specialization(heap, expr, environment, context, options)?;
 
-    let normalization = Normalization::new(context, 4);
+    let normalization = Normalization::new(context, NormalizationOptions::default());
     let node = normalization.run(node);
 
     Ok(node)
--- a/libs/@local/hashql/hir/src/lower/mod.rs
+++ b/libs/@local/hashql/hir/src/lower/mod.rs
@@ -8,7 +8,7 @@ use self::{
     ctor::ConvertTypeConstructor,
     error::{LoweringDiagnosticCategory, LoweringDiagnosticStatus},
     inference::TypeInference,
-    normalization::Normalization,
+    normalization::{Normalization, NormalizationOptions},
     specialization::Specialization,
 };
 use crate::{context::HirContext, fold::Fold as _, node::Node, visit::Visitor as _};
@@ -95,7 +95,7 @@ pub fn lower<'heap>(
     );
     let Ok(node) = specialization.fold_node(node);
 
-    let normalization = Normalization::new(context, 4);
+    let normalization = Normalization::new(context, NormalizationOptions::default());
     let node = normalization.run(node);
 
     diagnostics.into_status(node)
--- a/libs/@local/hashql/hir/src/lower/normalization.rs
+++ b/libs/@local/hashql/hir/src/lower/normalization.rs
@@ -123,6 +123,34 @@ const fn is_projection(node: &Node<'_>) -> bool {
     }
 }
 
+/// Configuration options for HIR(ANF) normalization.
+///
+/// This struct controls various aspects of the normalization process, primarily
+/// focused on performance optimizations and resource management.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct NormalizationOptions {
+    /// Maximum number of binding vectors to cache in the recycler pool.
+    ///
+    /// The normalization process frequently creates temporary vectors to accumulate
+    /// `let` bindings at boundary points. To reduce allocation overhead, completed
+    /// vectors are cached in a recycler pool for reuse.
+    ///
+    /// A higher value reduces allocations at the cost of increased memory usage.
+    /// A value of 0 disables recycling entirely.
+    ///
+    /// # Default Value
+    ///
+    /// The default is 4, which provides a good balance between memory usage and
+    /// allocation reduction for typical workloads.
+    pub max_recycle: usize,
+}
+
+impl Default for NormalizationOptions {
+    fn default() -> Self {
+        Self { max_recycle: 4 }
+    }
+}
+
 /// HIR to HIR(ANF) normalization transformer.
 ///
 /// This struct implements the [`Fold`] trait to transform HIR nodes into Administrative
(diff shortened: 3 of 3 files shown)


indietyp:
* **#7862** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/hashql/compiletest/src/suite/hir_lower_normalization.rs
```diff
@@ -0,0 +1,83 @@
...
+pub(crate) fn hir_lower_normalization<'heap>(
+    heap: &'heap Heap,
+    expr: Expr<'heap>,
+    environment: &mut Environment<'heap>,
+    context: &mut HirContext<'_, 'heap>,
+    options: &mut TestOptions,
+) -> Result<Node<'heap>, SuiteDiagnostic> {
+    let node = hir_lower_specialization(heap, expr, environment, context, options)?;
+
+    let normalization = Normalization::new(context, 4);
```
TimDiekmann: What is `4`? (Would be good to have a constant instead, can be adjacent).

libs/@local/hashql/hir/src/lower/mod.rs
```diff
@@ -32,7 +34,7 @@ pub fn lower<'heap>(
     node: Node<'heap>,
     types: &ExtractedTypes<'heap>,
     env: &mut Environment<'heap>,
-    context: &HirContext<'_, 'heap>,
+    context: &mut HirContext<'_, 'heap>,
```
TimDiekmann, line 37: Is it intended that the normalization alters the context in place?

libs/@local/hashql/eval/src/graph/read/mod.rs
```diff
@@ -189,6 +192,23 @@ impl<'heap> Visitor<'heap> for GraphReadCompiler<'_, 'heap> {
...
+        }
+    }
+
+    fn visit_local_variable(&mut self, variable: &'heap LocalVariable<'heap>) {
+        visit::walk_local_variable(self, variable);
+
+        if let Some(output) = self.variables.get(&variable.id.value) {
+            self.output.insert(self.current, output.clone());
+        }
+    }
```
TimDiekmann, line 210: We track the variables, but what happens if the variables are in nested scopes? How does it deal with shadowing?

libs/@local/hashql/hir/src/lower/normalization.rs
```diff
@@ -0,0 +1,662 @@
...
+/// Non-atoms include control flow, operations, calls, and qualified variables
+/// (which require module thunking in the future).
+const fn is_atom(node: &Node<'_>) -> bool {
+    match node.kind {
+        NodeKind::Data(_) | NodeKind::Variable(Variable::Local(_)) | NodeKind::Access(_) => true,
+        // Qualified variables are *not* atoms, because of module thunking in the future
+        // see: https://linear.app/hash/issue/BE-67/hashql-implement-modules
+        NodeKind::Variable(Variable::Qualified(_))
+        | NodeKind::Let(_)
+        | NodeKind::Input(_)
```
TimDiekmann, line 94: Shouldn't an input be an atom as well?

libs/@local/hashql/eval/src/graph/read/mod.rs
```diff
@@ -189,6 +192,23 @@ impl<'heap> Visitor<'heap> for GraphReadCompiler<'_, 'heap> {
         self.current = previous;
     }
 
+    fn visit_binding(&mut self, binding: &'heap Binding<'heap>) {
+        visit::walk_binding(self, binding);
+
+        // Check if the binder has been assigned to an output
+        if let Some(output) = self.output.get(&binding.value.id) {
+            self.variables.insert(binding.binder.id, output.clone());
```
TimDiekmann, line 200: I guess we need to clone an cannot work with references?

libs/@local/hashql/eval/src/graph/read/mod.rs
TimDiekmann, line 210: Do I see correctly, that we need to call `visit_binding` before `visit_local_variable`? Is this guaranteed?

libs/@local/hashql/hir/src/lower/normalization.rs
```diff
@@ -0,0 +1,662 @@
...
+        &mut self,
+        access: IndexAccess<'heap>,
+    ) -> Self::Output<IndexAccess<'heap>> {
+        let Ok(IndexAccess { expr, index }) = fold::walk_index_access(self, access);
+
+        // To be a valid projection, the inner body must be an atom, and the index must be a local
+        // variable
+        Ok(IndexAccess {
+            expr: self.ensure_projection(expr),
+            index: self.ensure_local_variable(index),
```
TimDiekmann, line 604: Why is it required to be local? Wouldn't this imply that
```
list[0]
```
becomes this?
```
let %1 = 0;
list[%1]
```

libs/@local/hashql/hir/src/lower/mod.rs
indietyp: Yes, it uses the `Counter`, we can switch out the `Counter` (which requires `&mut`) with a producer (which requires `&`) in the future if we can switch. Primarily, I didn't want to introduce atomic operations unless necessary, as I can use a single number.

When allocating a new variable, we increment the counter, thereby modifying the context.

libs/@local/hashql/hir/src/lower/normalization.rs
indietyp: Conceptually, `input` is a function of `(input ident default)`; therefore, it's a function and not a value, in the future this might change, but it has primarily to do with placement of values, `input` will look up the value in the global context, instead of directly referencing it as a constant, therefore the differentiation.

libs/@local/hashql/hir/src/lower/normalization.rs
indietyp: That is correct and is intended. This is primarily to prevent projections from going here, so we always have a value to point to. This is partially influenced by how rustc MIR handles it, although I might broaden this in the future, also to include constants. It makes implementation easier since we only need to consider one case.

Basically, it turns it into a `FieldAccess` equivalent.

libs/@local/hashql/eval/src/graph/read/mod.rs
indietyp: yes, the AST removes variables duplication even in nested scopes using the name mangler (https://github.com/hashintel/hash/blob/835819929ab45afe2fb6075a5ee95db1c1ad1b0a/libs/%40local/hashql/ast/src/lowering/name_mangler.rs#L150), so the variable is guaranteed to be unique. The same thing applies to `VarId`, which is monotonically increasing and unique within a compilation unit.

indietyp: That is guaranteed through the `walk` methods and AST checking, the import resolver errors out if a variable isn't in scope, and the `walk` invocations at the top always make sure that we first walk the tree before doing anything.

libs/@local/hashql/eval/src/graph/read/mod.rs
indietyp: Well, `Clone` here is a bit misleading - the type is practically `Copy`, but can't implement it because `Range<usize>` isn't Copy (because of `IntoIterator`)

libs/@local/hashql/compiletest/src/suite/hir_lower_normalization.rs
indietyp: `4` is the number of vecs we recycle during bindings. I added a configuration struct in c0dea7a2021c926ef2a77af19508334383dd9273. The idea is simple: we are going to churn through quite a bit of vecs to collect all the bindings. Instead of recreating and tracking a max capacity, we just keep some in reverse and reuse them.

The thesis here is that the amount of bindings per branch quickly stabilises.

indietyp:
> Also, I'm a bit concerned, that this may results in quite a few let bindings, does it have performance implications?

In a previous version yes, but this is why we did https://linear.app/hash/issue/BE-137/hashql-allow-let-to-assign-multiple-variables, this way we reduce the amount of the tree and just relocate. In general the tree will depth will *shrink* after this normalization step.

We basically introduce a soft [TAC](https://en.wikipedia.org/wiki/Three-address_code) which makes translation in the MIR (which operates on a TAC) a lot easier.

PR #7868, opened 2025-10-12
BE-154: HashQL: Implement MIR data structures

indietyp:

Implement the core MIR (Mid-level Intermediate Representation) data structures for HashQL, providing a control-flow graph based representation that makes analysis and optimization easier than working with the original HIR(ANF).

## 🔍 What does this change?

- Adds the `body` module with core MIR structures including `BasicBlock`, `Local`, `Place`, `Operand`, `RValue`, `Statement`, and `Terminator`
- Implements various terminator types for control flow: `Goto`, `Branch`, `Call`, `Return`, `GraphRead`
- Adds `RValue` variants for different computation types: `Load`, `Binary`, `Unary`, `Aggregate`, `Input`
- Creates the `def` module with `DefId` for referencing definitions with associated MIR bodies
- Updates the `newtype!` macro in `hashql-core` to support different integer types and larger ranges
- Enhances `IdVec` to work with custom allocators and implements trait derivations manually

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The MIR structures are designed to be used by future compiler passes and will be tested through integration tests

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -14092,6 +14092,9 @@ __metadata:
 "@rust/hashql-mir@workspace:libs/@local/hashql/mir":
   version: 0.0.0-use.local
   resolution: "@rust/hashql-mir@workspace:libs/@local/hashql/mir"
+  dependencies:
+    "@rust/hashql-core": "npm:0.0.0-private"
+    "@rust/hashql-hir": "npm:0.0.0-private"
   languageName: unknown
   linkType: soft
 


indietyp:
* **#7868** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/hashql/core/src/id/vec.rs
```diff
@@ -25,7 +27,6 @@ use super::{Id, slice::IdSlice};
 /// let user_id = users.push("Alice".to_string());
 /// assert_eq!(users[user_id], "Alice");
 /// ```
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
TimDiekmann, line 28: Can we derive it through `derive_where` or similar instead or doesn’t it work?

indietyp: We can, and I contemplated doing so, but because `hashql_core` is at the bottom of the crate stack I wanted to avoid `syn` if it isn't super needed (although we're still using `derive_more` and derive macros in hashql-core, but I want to remove it)

(The trait bounds are also a bit more tricky to get right, because of the `Allocator + Clone`)

PR #7870, opened 2025-10-13
BE-151: Prepare entity traversal for recursive CTEs by replacing BFS with sequential path processing

TimDiekmann:

Refactor entity traversal logic from breadth-first search (BFS) to sequential path processing, preparing for recursive CTE implementation that will reduce SQL roundtrips from N (one per edge hop) to 1 (entire path in a single query).

This PR restructures how entity edge traversal works by changing from a "process all paths hop-by-hop in breadth-first manner" approach to a "process each complete path sequentially" approach. This architectural change is required before implementing recursive CTEs, which will batch all edge queries within a path into a single SQL call.

## 🔗 Related links

- Linear issue: [BE-151: Sequentialize SQL calls for entity traversal](https://linear.app/hash/issue/BE-151)

### Core Architectural Changes

- **Changed** **`traverse_entities`** **signature** to accept a single `BorrowedTraversalParams` per call instead of a vector of entities with different traversal parameters
    - Moved path grouping responsibility to the caller (`get_entity` method)
    - Each path is now processed independently in sequential calls
- **Replaced while-loop with sequential iteration** through traversal path edges
    - Old approach: BFS - collected all edges to traverse across all paths, then processed them in batches
    - New approach: For each path, sequentially resolve each edge one by one
    - This is a behavior-preserving refactoring that maintains identical performance while enabling future CTE optimization

### Code Organization Improvements

- **Extracted** **`resolve_entity_edge`** **helper** to encapsulate knowledge edge traversal logic
    - Handles `HasLeftEntity` and `HasRightEntity` edges
    - Returns discovered entities for next hop
- **Extracted** **`resolve_is_of_type_edge`** **helper** to encapsulate shared edge (entity-to-type) traversal
    - Returns entity types for further traversal
- **Simplified** **`EntityEdgeTraversalData`** **structure**
    - Removed `traversal_params` field (no longer needs to track per-entity traversal state)
    - Changed constructor to accept `&QueryTemporalAxes` instead of individual timestamp/axis parameters
- **Updated** **`add_entity_id`** **return type** from `Option<(EntityVertexId, BorrowedTraversalParams, Interval)>` to `bool`
    - Simpler API: returns whether the entity was newly added
    - Traversal parameters are now passed separately and applied uniformly

### Algorithm Flow Changes

Both `ResolveDepths` and `Path` traversal modes now work similarly:

1. Start with initial set of entity IDs
2. For each edge in the traversal path (sequentially):
    - Resolve that edge for all current entities
    - Replace current entities with discovered entities
    - Continue to next edge
3. Special handling for `IsOfType` edges:
    - In `ResolveDepths`: Accumulates ALL entities (original + discovered during each hop) for batch type resolution at end
    - In `Path`: Processes inline as part of the path sequence

### Files Modified

- `libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs` - Main traversal logic refactor
- `libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs` - Updated data structures and signatures

This PR:

The changes in this PR:

The changes in this PR:

### Performance Characteristics

This refactoring changes how edge queries are batched:

- **Old behavior (BFS)**: When multiple paths share identical edge sequences at the same hop depth, they were batched into a single query
    - Example: Path A `[HasLeftEntity-outgoing, HasRightEntity]` and Path B `[HasLeftEntity-outgoing, IsOfType]` would batch the first `HasLeftEntity-outgoing` query
    - Note: Different edge types or directions (e.g., `HasLeftEntity-outgoing` vs `HasRightEntity-incoming`) were never batched together
- **New behavior (Sequential per path)**: Each path is processed independently, which can result in duplicate queries for shared starting edges
    - Example: The same two paths now execute `HasLeftEntity` twice (once per path)

Within each path, entities are still batched at each hop (N queries for N edge hops per path).

**Impact in practice**: Analysis of the current codebase shows that all existing query patterns use paths with **different first edges** (e.g., `HasLeftEntity` vs `HasRightEntity`, or different directions). This means the theoretical performance regression **does not occur in practice with current usage patterns**.

**Trade-off rationale**: Separating paths enables the upcoming CTE optimization where each path will be resolved in a single recursive query. Even if future query patterns introduce shared first edges, the temporary duplicate queries will be eliminated when CTEs are implemented, reducing all queries to 1 query per path regardless of edge overlap.

The performance optimization (reducing to 1 query per path via recursive CTEs) will be implemented in a follow-up PR.

## 🐾 Next steps

- Implement recursive CTE queries to batch all edge queries within a path into a single SQL call
- This will reduce SQL roundtrips from N to 1 per traversal path
- Performance testing to validate the optimization impact

## 🛡 What tests cover this?

- Existing entity traversal integration tests continue to pass
- The refactor maintains identical behavior to the previous implementation
- No new tests added as this is a pure refactoring with no observable behavior changes

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
@@ -105,6 +105,27 @@ impl<C> PostgresStore<C>
 where
     C: AsClient,
 {
+    /// Resolves `is-of-type` edges from entities to their entity types.
+    ///
+    /// Queries the database for `is-of-type` edges connecting the provided entities to their
+    /// corresponding [`EntityType`]s, applies permission filtering, and inserts the discovered
+    /// edges into the subgraph. Returns the discovered entity types for further traversal.
+    ///
+    /// [`EntityType`]: type_system::ontology::entity_type::EntityType
+    ///
+    /// # Arguments
+    ///
+    /// * `entities` - Collection of entity vertex IDs with their temporal intervals to traverse
+    ///   from
+    /// * `next_traversal` - Traversal parameters to apply to discovered entity types
+    /// * `traversal_context` - Context tracking visited vertices to prevent duplicates
+    /// * `provider` - Store provider for permission checks
+    /// * `subgraph` - Subgraph to populate with discovered edges and vertices
+    ///
+    /// # Errors
+    ///
+    /// Returns [`QueryError`] if the database query fails or permission filtering encounters an
+    /// error.
     async fn resolve_is_of_type_edge<'edges>(
         &self,
         entities: impl IntoIterator<Item = (EntityVertexId, RightBoundedTemporalInterval<VariableAxis>)>,
@@ -165,6 +186,25 @@ where
         Ok(entity_type_queue)
     }
 
+    /// Resolves a single entity edge type for a collection of entities.
+    ///
+    /// Queries the database for edges of the specified [`EntityTraversalEdge`] connecting the
+    /// provided entities to other entities. Discovered edges are inserted into the subgraph, and
+    /// newly discovered entities are returned for the next traversal hop.
+    ///
+    /// # Arguments
+    ///
+    /// * `entities` - Collection of entity vertex IDs with their temporal intervals to traverse
+    ///   from
+    /// * `edge` - The type of entity edge to traverse
+    /// * `next_traversal` - Traversal parameters to apply to discovered entities
+    /// * `traversal_context` - Context tracking visited vertices to prevent duplicates
+    /// * `provider` - Store provider for permission checks
+    /// * `subgraph` - Subgraph to populate with discovered edges and vertices
+    ///
+    /// # Errors
+    ///
+    /// Returns [`QueryError`] if the database query fails.
     async fn resolve_entity_edge<'edges>(
         &self,
         entities: impl IntoIterator<Item = (EntityVertexId, RightBoundedTemporalInterval<VariableAxis>)>,


TimDiekmann:
* **#7870** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7870?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -105,227 +105,282 @@ impl<C> PostgresStore<C>
...
-    /// Internal method to read an [`Entity`] into a [`TraversalContext`].
-    ///
-    /// This is used to recursively resolve a type, so the result can be reused.
-    #[tracing::instrument(
-        level = "info",
-        skip(self, entity_queue, traversal_context, provider, subgraph)
-    )]
-    #[expect(clippy::too_many_lines)]
-    pub(crate) async fn traverse_entities<'edges>(
+    async fn resolve_is_of_type_edge<'edges>(
```
hashdotai, line 129: It's good that you're extracting helper methods to improve code organization. Consider adding a brief docstring to this method to document its purpose, especially since it's a core part of the new approach.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -105,227 +105,282 @@ impl<C> PostgresStore<C>
...
         &self,
-        mut entity_queue: Vec<(
-            EntityVertexId,
+        entities: impl IntoIterator<Item = (EntityVertexId, RightBoundedTemporalInterval<VariableAxis>)>,
```
hashdotai, line 131: Consider using more type-safe parameter definitions instead of `impl IntoIterator<Item = ...>` for better API clarity. While this is flexible, it can make function signatures harder to understand at a glance.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -43,57 +43,59 @@ use crate::store::{
...
         self.entity_revision_ids.push(vertex_id.revision_id);
         self.intervals.push(interval);
-        self.traversal_params.push(traversal_params);
+    }
+
+    pub const fn len(&self) -> usize {
+        self.web_ids.len()
+    }
+
+    pub const fn is_empty(&self) -> bool {
```
hashdotai, line 82:
```suggestion
    pub const fn is_empty(&self) -> bool {
        self.web_ids.is_empty()
    }
```
 This implementation is more direct and avoids an extra function call. It also uses the standard collection method pattern.

libs/@local/graph/postgres-store/src/store/postgres/traversal_context.rs
```diff
@@ -312,12 +312,9 @@ impl<'edges> TraversalContext<'edges> {
         edition_id: EntityEditionId,
         traversal_params: BorrowedTraversalParams<'edges>,
         traversal_interval: RightBoundedTemporalInterval<VariableAxis>,
-    ) -> Option<(
-        EntityEditionId,
-        BorrowedTraversalParams<'edges>,
-        RightBoundedTemporalInterval<VariableAxis>,
-    )> {
+    ) -> bool {
```
hashdotai, line 315: Great improvement to `add_entity_id` return type - returning a boolean is much cleaner than the previous tuple option.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -105,227 +105,282 @@ impl<C> PostgresStore<C>
...
+                if graph_resolve_depths.is_of_type {
+                    shared_edge_data.extend(entities.iter().copied());
+                }
 
-                    let next_traversal = traversal_context.add_entity_type_id(
-                        EntityTypeUuid::from(edge.right_endpoint_ontology_id),
-                        edge.traversal_params,
-                        edge.traversal_interval,
-                    );
+                // Sequentially traverse all entity edges in the path. Each iteration processes
```
hashdotai, line 306: The sequential approach makes the code much easier to follow. This is a good example of the "process each complete path sequentially" approach described in your PR overview.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1421,70 +1476,67 @@ where
...
-                                }
-                            }
+            let entity_ids = subgraph
+                .vertices
+                .entities
+                .keys()
+                .copied()
+                .collect::<Vec<_>>();
+
+            // Iterate over each traversal path and call traverse_entities separately
```
hashdotai, line 1526: This is a significant improvement to the calling code structure. Moving from a complex nested data collection to separate method calls for each path is much more maintainable.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: If we change where we get `len()` from we also need to change this. The function call will be optimized out.

PR #7872, opened 2025-10-15
BE-162: Refactor `OntologyTypeVersion` to support pre-release versions with lane-based drafts

TimDiekmann:

This PR refactors the ontology type version system to support semantic versioning with pre-release/draft versions. Previously, `OntologyTypeVersion` was a simple `u32` wrapper representing only published versions. Now it supports draft versions with lane-based isolation, enabling parallel development workflows while maintaining proper version ordering semantics.

The new URL format supports both published versions (`v/2`) and draft versions (`v/2-draft.lane.5`), following semantic versioning conventions where published versions have higher precedence than pre-release versions with the same major version.

## 🔗 Related links

- Linear issue: [BE-162](https://linear.app/hash/issue/BE-162) (Implement pre-release parsing for OntologyTypeVersion)
- Follow-up: [BE-161](https://linear.app/hash/issue/BE-161) (Allow IDs for pre-release types to be stored in Postgres)

### Core Type System Changes

- **Refactored** **`OntologyTypeVersion`** **structure** ([`libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs))
    - Changed from simple `u32` wrapper to struct with `major: u32` and `pre_release: Option<PreRelease>`
    - Created `PreRelease` enum to support extensible pre-release types (currently only `Draft` variant)
    - Draft format: `major-draft.lane.revision` (e.g., `2-draft.abcd1234.5`)
    - Lanes support dots for hierarchical naming (e.g., `v1.alpha.3`)
- **Implemented semantic versioning ordering** ([`libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs))
    - Published versions have higher precedence than drafts with same major version (v/2 > v/2-draft.x.y)
    - Draft versions ordered by lane (following SemVer rules) then revision (numeric)
    - Major version takes precedence across all comparisons
    - SemVer-compliant lane comparison: numeric identifiers compared numerically, alphanumeric compared lexicographically, numeric < alphanumeric
- **Enhanced error types with structured validation** ([`libs/@blockprotocol/type-system/rust/src/ontology/id/error.rs`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/rust/src/ontology/id/error.rs))
    - Added `ParseOntologyTypeVersionError` for version-specific errors with three variants:
        - `MissingVersion` - Empty version string
        - `ParseVersion(String)` - Invalid major version number
        - `InvalidPreRelease(String, ParseDraftInfoError)` - Draft format validation failures
    - Added `ParseDraftInfoError` for draft format validation with four variants:
        - `IncorrectFormatting` - Missing dot separator or empty lane
        - `InvalidLane(String)` - SemVer-invalid lane identifier
        - `MissingRevision` - Empty revision after dot
        - `InvalidRevision(String, String)` - Non-numeric or out-of-range revision
    - Changed `ParseVersionedUrlError::InvalidVersion` to use structured error types instead of strings
    - Error messages now provide detailed context about what's wrong and where

### TypeScript/JavaScript Changes

- **Updated URL validation** ([`libs/@blockprotocol/type-system/typescript/src/native/url.ts`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/typescript/src/native/url.ts))
    - Added regex pattern matching for draft versions with proper lane extraction
    - Validates both major version and revision numbers fit in `u32` range
    - Handles draft and published versions uniformly in validation
- **Implemented SemVer comparison logic** ([`libs/@blockprotocol/type-system/typescript/src/native/url.ts`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/typescript/src/native/url.ts))
    - `compareOntologyTypeVersions()` now properly handles draft versions
    - Added helper functions: `isDraftVersion()`, `extractMajorVersion()`, `haveSameMajorVersion()`
    - Matches Rust implementation's ordering semantics
- **Fixed TypeScript type generation** ([`libs/@blockprotocol/type-system/typescript/src/native/url.ts`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/typescript/src/native/url.ts))
    - Error types now properly export structured error information instead of tuples
    - Error format changed from `["version", "error string"]` to `["version", { reason: "...", inner: "..." }]`

### Database and API Updates

- **Updated all call sites** (34 files across `libs/@local/graph/`)
    - Changed `OntologyTypeVersion::new()` calls to struct initialization
    - Updated `.inner()` calls to `.major` field access
    - Modified `ToSql`/`FromSql` implementations to handle drafts (with TODO for full support)
    - Added errors for draft versions in Postgres (tracked in BE-161)

#### Comprehensive Rust Test Suite

**Refactored existing tests** to use `from_str()` and proper error handling:

- Changed from `OntologyTypeVersion::new()` to parsing from strings
- Updated function signatures to `Result<(), Box<dyn Error>>` for better error propagation
- Enhanced error validation to test actual error message content, not just error types

**Added 11 new comprehensive test functions** covering all edge cases:

1. **`ontology_version_parsing`** - Tests parsing both published (`1`, `42`) and draft (`2-draft.abcd1234.5`) version strings, verifies structure and roundtrip
2. **`ontology_version_roundtrip`** - Ensures parse → serialize → parse consistency for multiple version formats
3. **`ontology_version_ordering_same_major`** - Tests SemVer ordering rules for same major version (published > draft, revision ordering)
4. **`ontology_version_ordering_different_major`** - Tests major version precedence across published and draft versions
5. **`ontology_version_ordering_different_lanes`** - Tests lane-based ordering (lexicographic comparison)
6. **`ontology_version_ordering_semver_prerelease_rules`** - Comprehensive SemVer validation:
    - Rule 1: Numeric identifiers compared numerically (`2 < 10`)
    - Rule 2: Alphanumeric compared lexicographically (`alpha < beta`)
    - Rule 3: Numeric < alphanumeric (`999 < 123abc`)
    - Rule 4: Revision number comparison within same lane
7. **`ontology_version_equality`** - Tests equality comparisons for published and draft versions
8. **`draft_url_parsing`** - Tests full URL parsing with draft versions and roundtrip
9. **`invalid_prerelease_identifiers_rejected`** - Tests rejection of:
    - Special characters in lane (`abc,def`)
    - Unicode characters in lane (🚀)
    - Validates specific error types and messages
10. **`missing_lane_rejected`** - Tests edge case: `1-draft.123` (only revision, no lane separator)
    - Validates `ParseDraftInfoError::IncorrectFormatting`
11. **`missing_revision_rejected`** - Tests edge case: `1-draft.lane.` (lane present but empty revision)
    - Validates `ParseDraftInfoError::MissingRevision` with exact input tracking
12. **`invalid_revision_rejected`** - Tests rejection of:
    - Non-numeric revision (`lane.abc`) → `InvalidRevision` with message validation
    - Negative revision (`lane.-5`) → `InvalidRevision` with exact value tracking
    - Overflow cases (implicit in u32 parsing)
    - All cases validate error structure and message content
13. **`multiple_dots_in_lane_accepted`** - Tests hierarchical lanes (`lane.with.dots.5`, `v1.alpha.3.10`)
14. **`draft_in_lane_accepted`** - Tests lane names containing "draft" (`my-draft.2`)
15. **`build_is_rejected`** - Tests rejection of SemVer build metadata:
    - In draft versions (`1-draft.lane.1+build.123`)
    - In published versions (`1+build.123`)
    - In ambiguous formats (`1+build-draft.123`)

**Key testing improvements:**

- All error tests now validate the exact error variant, input tracking, and error message content
- Tests use pattern matching to ensure structured errors contain expected data
- Edge cases comprehensively covered: missing components, invalid formats, boundary conditions
- SemVer compliance thoroughly validated across all comparison rules

#### TypeScript Tests

**Added TypeScript tests** ([`libs/@blockprotocol/type-system/typescript/test/url.test.ts`](https://github.com/hashintel/hash/blob/t/be-162-implement-pre-release-parsing-for-ontologytypeversion/libs/@blockprotocol/type-system/typescript/test/url.test.ts))

- Draft URL validation tests
- Version comparison tests
- Helper function tests
- Error case validation
- Snapshot tests updated for new error format

#### Integration Coverage

- All existing integration tests pass with updated version handling
- 34 files updated across graph API, store, and benchmark code

## ⚠️ Known issues

- **Draft versions cannot be stored in Postgres yet**: The `ToSql` implementation for `OntologyTypeVersion` includes an error (via `todo!()`) when attempting to serialize draft versions. This is intentional - draft support in the database layer is tracked separately in [BE-161](https://linear.app/hash/issue/BE-161).
- **Error type incompatibility**: The error structure changed from simple strings to nested structured errors. This may affect error handling code that pattern matches on the old format.

## 🐾 Next steps

- BE-161: Implement Postgres storage for draft versions (requires schema changes and migration strategy)
- **Consider error-stack migration**: Multiple TODOs added indicating we should migrate to `error-stack` for better error handling (tracked in BE-160 subtask)

### Rust Tests (in `libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs`)

**Core functionality tests:**

- `ontology_version_parsing` - Parsing published and draft versions
- `ontology_version_roundtrip` - Serialize/deserialize consistency
- `ontology_version_ordering_same_major` - SemVer ordering (same major)
- `ontology_version_ordering_different_major` - Major version precedence
- `ontology_version_ordering_different_lanes` - Lane-based ordering
- `ontology_version_ordering_semver_prerelease_rules` - Complete SemVer compliance validation
- `ontology_version_equality` - Equality comparisons
- `draft_url_parsing` - Full URL parsing with draft versions

**Error validation tests:**

- `invalid_prerelease_identifiers_rejected` - Special characters and Unicode rejection
- `missing_lane_rejected` - Edge case: only revision without lane (`1-draft.123`)
- `missing_revision_rejected` - Edge case: lane without revision (`1-draft.lane.`)
- `invalid_revision_rejected` - Non-numeric (`abc`), negative (`-5`), and overflow cases

**Edge case tests:**

- `multiple_dots_in_lane_accepted` - Hierarchical lane names (`lane.with.dots`)
- `draft_in_lane_accepted` - Lane names containing "draft"
- `build_is_rejected` - SemVer build metadata rejection

**Updated existing tests:**

- `base_url_test` - Updated for new error types
- `versioned_url_test` - Updated for structured errors
- `record_id_conversions` - Updated for new structure

### TypeScript Tests (in `libs/@blockprotocol/type-system/typescript/test/url.test.ts`)

- Draft URL validation tests
- Version comparison tests covering all SemVer ordering scenarios
- Helper function tests (`isDraftVersion`, `extractMajorVersion`, etc.)
- Error message validation for invalid draft formats
- Snapshot tests updated for new error format

### Integration Coverage

- All existing integration tests pass with updated version handling
- 34 files updated across graph API, store, and benchmark code

### Test Published Versions (No Behavior Change)

1. Checkout the branch
2. Run Rust tests: `cargo nextest run --package type-system`
3. Run TypeScript tests: `turbo run test --filter '@blockprotocol/type-system'`
4. Verify existing functionality unchanged

### Test Draft Version Parsing

```
# Test Rust parsing
cargo test --package type-system ontology_version_parsing

# Test TypeScript validation
turbo run test --filter '@blockprotocol/type-system' -- url.test.ts
```

### Test Version Ordering

1. Run ordering tests:
2. Verify SemVer semantics:
    - Published > Draft (same major): `v/2` > `v/2-draft.x.y`
    - Major version precedence: `v/3` > `v/2` > `v/1`
    - Draft ordering: `v/2-draft.a.1` < `v/2-draft.b.1` < `v/2-draft.b.2`

### Test Error Handling and Edge Cases

```
# Run all error validation tests
cargo test --package type-system rejected

# Test specific edge cases
cargo test --package type-system missing_lane_rejected
cargo test --package type-system missing_revision_rejected
cargo test --package type-system invalid_revision_rejected
```

Verify error messages provide detailed context:

- Missing lane: `1-draft.123` → `IncorrectFormatting`
- Missing revision: `1-draft.lane.` → `MissingRevision`
- Invalid revision: `1-draft.lane.abc` → `InvalidRevision("abc", "invalid digit...")`
- Negative revision: `2-draft.lane.-5` → `InvalidRevision("-5", "invalid digit...")`

### Test Database Limitations (Expected Failure)

Attempting to serialize draft version to Postgres should panic with TODO message:

```
let draft = OntologyTypeVersion::from_str("2-draft.lane.5")?;
// This will panic with: "not yet implemented: https://linear.app/hash/issue/BE-161/..."
draft.to_sql(...)?;
```

## 📹 Demo

No visual demo needed - this is infrastructure work for the type system. The changes are visible in:

1. **Rust code** - New types and parsing logic
2. **Test output** - Comprehensive test coverage showing all scenarios
3. **Error messages** - Improved structured error reporting with detailed context

Example of new URL format support:

```
Published:  https://example.com/person/v/2
Draft:      https://example.com/person/v/2-draft.abcd1234.5
            ├─ major: 2
            ├─ pre_release.lane: "abcd1234"
            └─ pre_release.revision: 5

Ordering:
  v/1 < v/2-draft.lane.5 < v/2 < v/3

SemVer Lane Comparison:
  v/1-draft.2.1 < v/1-draft.10.1       (numeric: 2 < 10)
  v/1-draft.999.1 < v/1-draft.abc.1    (numeric < alphanumeric)
  v/1-draft.alpha.1 < v/1-draft.beta.1 (lexicographic)

Hierarchical Lanes:
  v/1-draft.lane.with.dots.5
  v/2-draft.v1.alpha.3.10
```

### DIFF
--- a/libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
+++ b/libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
@@ -1153,6 +1153,92 @@ mod tests {
     }
 
     #[test]
+    fn missing_lane_rejected() {
+        // Edge case: only revision without lane (no dot separator)
+        let err = OntologyTypeVersion::from_str("1-draft.123")
+            .expect_err("Should reject draft with only revision, no lane");
+        assert!(
+            matches!(
+                err,
+                ParseOntologyTypeVersionError::InvalidPreRelease(
+                    _,
+                    ParseDraftInfoError::IncorrectFormatting
+                )
+            ),
+            "Expected IncorrectFormatting error, got: {err:?}"
+        );
+    }
+
+    #[test]
+    fn missing_revision_rejected() {
+        // Edge case: lane present but revision missing
+        let err = OntologyTypeVersion::from_str("1-draft.lane.")
+            .expect_err("Should reject draft with missing revision");
+        let ParseOntologyTypeVersionError::InvalidPreRelease(
+            input,
+            ParseDraftInfoError::MissingRevision,
+        ) = err
+        else {
+            panic!("Expected MissingRevision error, got: {err:?}");
+        };
+        assert_eq!(input, "lane.");
+    }
+
+    #[test]
+    fn invalid_revision_rejected() {
+        // Non-numeric revision should be rejected
+        let err = OntologyTypeVersion::from_str("1-draft.lane.abc")
+            .expect_err("Should reject non-numeric revision");
+        let ParseOntologyTypeVersionError::InvalidPreRelease(
+            input,
+            ParseDraftInfoError::InvalidRevision(invalid_rev, msg),
+        ) = err
+        else {
+            panic!("Expected InvalidRevision error, got: {err:?}");
+        };
+        assert_eq!(input, "lane.abc");
+        assert_eq!(invalid_rev, "abc");
+        assert!(
+            msg.contains("invalid digit"),
+            "Error message should mention invalid digit, got: {msg}"
+        );
+
+        // Negative revision should be rejected
+        let err = OntologyTypeVersion::from_str("2-draft.lane.-5")
+            .expect_err("Should reject negative revision");
+        let ParseOntologyTypeVersionError::InvalidPreRelease(
+            input,
+            ParseDraftInfoError::InvalidRevision(invalid_rev, msg),
+        ) = err
+        else {
+            panic!("Expected InvalidRevision error, got: {err:?}");
+        };
+        assert_eq!(input, "lane.-5");
+        assert_eq!(invalid_rev, "-5");
+        assert!(
+            msg.contains("invalid digit"),
+            "Error message should mention invalid digit, got: {msg}"
+        );
+
+        // Overflow should be rejected
+        let err = OntologyTypeVersion::from_str("3-draft.lane.99999999999999999999")
+            .expect_err("Should reject revision overflow");
+        let ParseOntologyTypeVersionError::InvalidPreRelease(
+            input,
+            ParseDraftInfoError::InvalidRevision(invalid_rev, msg),
+        ) = err
+        else {
+            panic!("Expected InvalidRevision error, got: {err:?}");
+        };
+        assert_eq!(input, "lane.99999999999999999999");
+        assert_eq!(invalid_rev, "99999999999999999999");
+        assert!(
+            msg.contains("invalid digit") || msg.contains("too large"),
+            "Error message should mention overflow or invalid digit, got: {msg}"
+        );
+    }
+
+    #[test]
     fn multiple_dots_in_lane_accepted() {
         // SemVer allows multiple identifiers separated by dots
         // e.g., "draft.lane.with.dots" is valid as it becomes ["draft", "lane", "with", "dots"]
@@ -1236,19 +1322,5 @@ mod tests {
             error_msg.contains("invalid digit"),
             "Error message should mention invalid digit, got: {error_msg}"
         );
-    }
-
-    #[test]
-    fn absent_lane_is_rejected() {
-        let err = OntologyTypeVersion::from_str("1-draft.123")
-            .expect_err("Should reject build metadata in version string");
-        let ParseOntologyTypeVersionError::InvalidPreRelease(
-            input,
-            ParseDraftInfoError::IncorrectFormatting,
-        ) = err
-        else {
-            panic!("Expected `InvalidPreRelease` error for build metadata in draft, got: {err:?}");
-        };
-        assert_eq!(input, "123");
     }
 }


TimDiekmann:
* **#7872** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7872?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7872?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
```diff
@@ -289,11 +394,15 @@ impl ToSql for OntologyTypeVersion {
...
+    #[expect(clippy::todo, reason = "Draft versions are not yet supported")]
     fn to_sql(&self, ty: &Type, out: &mut BytesMut) -> Result<IsNull, Box<dyn Error + Sync + Send>>
     where
         Self: Sized,
     {
-        i64::from(self.0).to_sql(ty, out)
+        if self.pre_release.is_some() {
+            todo!("https://linear.app/hash/issue/BE-161/allow-ids-for-pre-release-type-to-be-stored-in-postgres");
+        }
+        i64::from(self.major).to_sql(ty, out)
```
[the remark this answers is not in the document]
TimDiekmann: I intentionally chose a `panic!` here. We should not use draft versions until they are implemented. This is intended to fail hard, it's a serious bug and probably quite a miscommunication.

libs/@blockprotocol/type-system/rust/src/ontology/id/error.rs
```diff
@@ -17,20 +19,50 @@ pub enum ParseBaseUrlError {
...
+// TODO: Use error-stack
+//   see https://linear.app/hash/issue/BE-160/simplify-error-handling-in-type-system-package
+#[cfg_attr(target_arch = "wasm32", derive(Tsify))]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Error)]
+#[serde(tag = "reason", content = "inner")]
+pub enum ParseDraftInfoError {
+    #[error("incorrect formatting")]
+    IncorrectFormatting,
+    #[error("missing lane")]
+    MissingVersion,
```
TimDiekmann:
```suggestion
```

libs/@blockprotocol/type-system/typescript/src/native/url.ts
```diff
@@ -284,17 +345,100 @@ export const incrementOntologyTypeVersion = (
...
+    if (a.preRelease.lane !== b.preRelease.lane) {
+      return a.preRelease.lane < b.preRelease.lane ? -1 : 1;
+    }
+    if (a.preRelease.revision !== b.preRelease.revision) {
+      return a.preRelease.revision < b.preRelease.revision ? -1 : 1;
+    }
     return 0;
   }
+
+  return 0;
```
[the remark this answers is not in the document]
TimDiekmann: Say that to `tsc`:
> Function lacks ending return statement and return type does not include 'undefined'.ts(2366)

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
```diff
@@ -217,37 +219,193 @@ impl<'a> FromSql<'a> for BaseUrl {
...
+        /// compared numerically when ordering drafts in the same lane.
+        revision: u32,
+    },
+}
+
+impl FromStr for PreRelease {
+    type Err = ParseDraftInfoError;
+
+    fn from_str(draft_info: &str) -> Result<Self, Self::Err> {
+        draft_info.rsplit_once('.').map_or(
```
indietyp: Would it also make sense to verify that it is a valid pre-release string by using https://docs.rs/semver/latest/semver/struct.Prerelease.html#method.new?

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
```diff
@@ -217,37 +219,193 @@ impl<'a> FromSql<'a> for BaseUrl {
...
+        /// Monotonic revision number within this lane.
+        ///
+        /// Should increment with each new draft in the same lane. Revisions are
+        /// compared numerically when ordering drafts in the same lane.
+        revision: u32,
+    },
+}
+
+impl FromStr for PreRelease {
+    type Err = ParseDraftInfoError;
```
indietyp, line 280: Is this the correct error? considering we're parsing `PreRelease` and not only drafts?

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
```diff
@@ -217,37 +219,193 @@ impl<'a> FromSql<'a> for BaseUrl {
...
+/// assert!(matches!(draft.pre_release, Some(PreRelease::Draft { .. })));
+///
+/// // Ordering: published > draft (same major)
+/// assert!(v1 < draft); // v/1 < v/2-draft.lane123.5
+/// let v2 = OntologyTypeVersion::from_str("2").unwrap();
+/// assert!(draft < v2); // v/2-draft.lane123.5 < v/2
+/// ```
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+#[cfg_attr(feature = "codegen", derive(specta::Type), specta(transparent))]
+pub struct OntologyTypeVersion {
```
indietyp, line 346: Again, would it make sense to simply wrap `semver::Version` here and ensure that the `minor` and `patch` are always `0`?

indietyp: This would mean that we wouldn't need to implement ordering, as ordering is defined per semver spec

TimDiekmann: `semver` does not make any assumption on what `pre_release` is, we can properly parse it to only allow `draft.<lane>.<rev>`

libs/@blockprotocol/type-system/typescript/src/native/url.ts
```diff
@@ -106,34 +107,94 @@ export const validateVersionedUrl = (
...
-    const versionNumber = Number(version);
-    if (versionNumber > 4294967295) {
-      return {
-        type: "Err",
-        inner: {
-          reason: "InvalidVersion",
-          inner: [version, "number too large to fit in target type"],
-        },
-      };
+    // Try to match draft version format first: "2-draft.lane.5"
```
indietyp: fwiw there's also: https://www.npmjs.com/package/semver that could be used here, additionally `[^.]` is too lax as per semver spec.

indietyp: It would also allow us to get automatic ordering.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
```diff
@@ -106,34 +107,94 @@ export const validateVersionedUrl = (
...
+    const draftMatch = version.match(/^(\d+)-draft\.([^.]+)\.(\d+)$/);
+
+    if (draftMatch) {
+      // It's a draft version
+      const [, majorStr, _lane, revisionStr] = draftMatch;
+      const major = Number(majorStr);
+      const revision = Number(revisionStr);
+
+      if (major > 4294967295) {
```
indietyp: Did you mean to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

tests/graph/integration/postgres/sorting.rs
```diff
@@ -78,37 +78,36 @@ async fn test_root_sorting(
...
-                        include_count: true,
-                        include_entity_types: None,
-                        include_drafts: false,
-                        include_web_ids: false,
-                        include_created_by_ids: false,
-                        include_edition_created_by_ids: false,
-                        include_type_ids: false,
-                        include_type_titles: false,
-                        include_permissions: false,
+        } = Box::pin(api.query_entity_subgraph(
```
indietyp, line 81: i guess this required box pinning because the future got too large?

TimDiekmann: Yes

libs/@blockprotocol/type-system/typescript/src/native/url.ts
TimDiekmann: `semver` does not parse `"1"` as `1.0.0` — it errors.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
TimDiekmann: No, that's 2^53-1 (The code was there before, I just moved/copied it)

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
indietyp: But it does, semver orders lexographically, which means that the ordering would be consistent. It would be _exactly_ the same as what you're doing right now.

11.4:

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
> 1. Identifiers consisting of only digits are compared numerically.
> 2. Identifiers with letters or hyphens are compared lexically in ASCII sort order.
> 3. Numeric identifiers always have lower precedence than non-numeric identifiers.
> 4. A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
TimDiekmann: We only parse drafts as pre-release. I don't see a reason to wrap this error in another error for wrappings sake. See BE-160

libs/@blockprotocol/type-system/typescript/src/native/url.ts
indietyp: Yes, I didn't know the reason, but it just seemed like a magic number. I know that this is `2^32-1` (the u32 limit), but I felt like the thing it was trying to do is fit it inside the max safe integer instead (or be compatible to the backend). Either way I would appreciate if we would factor out that number to e.g. `U32_MAX`, if we're not going to use `MAX_SAFE_INTEGER`.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
indietyp: I was sure it did, but just verified - it doesn't. I wonder if we want to expand `"1"` to `"1.0.0"`, as we're aiming for semver compatibility down the line anyway, but understand why one wouldn't want to do so.

TimDiekmann: We might want to at some point but 100% not as part of this PR.

TimDiekmann: What we could do is to parse it as `major-prerelease`, modify `major` to `${major}.0.0` and pass it on to semver, but that is an overhead.

indietyp: That was what I was advocating for yes, I personally think it is worth it, because it means that we wouldn't need to implement the sorting ourselves, but can instead rely on the semver one.

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
TimDiekmann: No, it wouldn't be the exact same. Currently, we violating rule 1 and 3. Examples:
- `1-draft.10.1` should be greater than `1-draft.2.1` (rule 1)
- `1-draft.900.1` should be smaller than `1-draft.123abc.1` (rule 3)
These don't really apply to us because we don't want to numbers as lanes, however, it's better to have it correct. I fixed that and added tests for the rules

Anyway, I would still not use the `semver` crate because that crate would not parse `lane` and `revision` for us. It does not assume anything what `prerelease` is, it could be `1-somecoolprerelease.foobar.idontcareaboutdrafts.9000` or [`shitshow`](https://crates.io/crates/sassers/0.11.0-shitshow) and would parse.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
TimDiekmann: I made it a constant.

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
TimDiekmann: We could extract the pre-release and parse it from here, yes. But I would not continue using the actual semver structure.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
TimDiekmann: I added a `toSemVer` (and simplified and removed unused bits)

libs/@blockprotocol/type-system/typescript/src/native/url.ts
```diff
@@ -9,6 +9,8 @@ import type {
   BaseUrl,
   OntologyTypeVersion,
 } from "@blockprotocol/type-system-rs/types";
+import type { SemVer } from "semver";
+import * as semver from "semver";
```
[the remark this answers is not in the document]
TimDiekmann: This was outdated before you posted it.

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
```diff
@@ -217,37 +219,202 @@ impl<'a> FromSql<'a> for BaseUrl {
...
+        Ok(())
+    }
+}
+
+impl FromStr for OntologyTypeVersion {
+    type Err = ParseOntologyTypeVersionError;
+
+    fn from_str(ontology_version: &str) -> Result<Self, Self::Err> {
+        let (version, draft_info) =
+            if let Some((version, draft_info)) = ontology_version.rsplit_once("-draft.") {
```
TimDiekmann: I wonder if we want to use left-split instead @indietyp

indietyp (dismissed):

libs/@blockprotocol/type-system/rust/src/ontology/id/mod.rs
indietyp: It doesn't _really_ matter, because `-` is only supposed to be in the version once, but `split_once` seems to be faster. In general using `split_once('.')` is the fastest because it uses `memchr` internally instead of sequentially going through the string.

TimDiekmann: It does matter because `-` is an allowed character. I changed that and added a test.

indietyp: Thank you yes, I totally missed in the BNF that:

```
<alphanumeric identifier> ::= <non-digit>
```

and then `non-digit`:

```
<non-digit> ::= <letter>
              | "-"
```

indietyp: Meaning that an absence of a `-` just proves that there's no pre-release or build present, but that the presence of a `-` may mean that a build ***or*** prerelease are present.

TimDiekmann: Yes, exactly. `-` can occur in `build` as well, so you would need to check both, `-` **and** `+`.

TimDiekmann: I don't know if it's worth adding a check that we don't want `+build` to be defined. I have added a small test regardless.

libs/@blockprotocol/type-system/typescript/src/native/url.ts
```diff
@@ -284,17 +362,23 @@ export const incrementOntologyTypeVersion = (
...
-  const parsedVersionB = Number.parseInt(versionB.toString(), 10);
-  if (parsedVersionA < parsedVersionB) {
-    return -1;
-  } else if (parsedVersionA > parsedVersionB) {
-    return 1;
-  } else {
-    return 0;
-  }
-};
+): -1 | 0 | 1 => toSemVer(versionA).compare(toSemVer(versionB));
```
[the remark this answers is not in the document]
TimDiekmann: Good spot, however, throwing here is fine because it's expected it's correct here.

PR #7873, opened 2025-10-15
BE-156: HashQL: Implement thunking for module-level bindings

indietyp:

This PR introduces thunking transformation for top-level bindings in HashQL, enabling lazy evaluation and proper handling of recursive definitions. The transformation wraps top-level bindings in thunks and inserts calls at reference sites, which is essential for implementing module-level lazy evaluation semantics and more importantly allows for **MIR lowering** of individual bindings without having to evaluate the whole body.

## 🔍 What does this change?

- Adds a new `Thunk` node type to represent delayed computations
- Implements `Thunking` transformer that wraps top-level bindings in thunks
- Modifies `Normalization` to handle thunks properly and maintain ANF properties
- Updates `Call` node to include a `PointerKind` field to distinguish between fat and thin function pointers
- Adds support for thunks in the evaluation engine
- Creates test suite for thunking transformation
- Refactors normalization state to be reusable across multiple passes

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- New test suite for thunking transformation
- Updated existing tests to account for thunking behavior
- Evaluation tests to ensure thunks work correctly in the runtime

## ❓ How to test this?

1. Run the test suite to verify thunking transformation works correctly
2. Check that recursive definitions are properly handled
3. Verify that lazy evaluation semantics are maintained

### DIFF
--- a/libs/@local/hashql/hir/src/lower/thunking.rs
+++ b/libs/@local/hashql/hir/src/lower/thunking.rs
@@ -247,14 +247,12 @@ impl<'ctx, 'env, 'heap> Thunking<'ctx, 'env, 'heap> {
     /// Panics if the trampoline is already set, which would indicate multiple
     /// replacement attempts for the same node.
     fn trampoline(&mut self, node: Node<'heap>) {
-        match &mut self.trampoline {
-            None => {
-                self.trampoline = Some(node);
-            }
-            Some(_) => {
-                panic!("trampoline has been inserted to multiple times");
-            }
-        }
+        assert!(
+            self.trampoline.is_none(),
+            "trampoline has been inserted to multiple times"
+        );
+
+        self.trampoline = Some(node);
     }
 }
 
--- a/libs/@local/hashql/hir/src/lower/normalization.rs
+++ b/libs/@local/hashql/hir/src/lower/normalization.rs
@@ -296,14 +296,12 @@ impl<'ctx, 'env, 'heap> Normalization<'ctx, 'env, 'heap> {
     /// Panics if the trampoline is already occupied, indicating multiple
     /// replacements attempted for the same node.
     fn trampoline(&mut self, node: Node<'heap>) {
-        match &mut self.trampoline {
-            None => {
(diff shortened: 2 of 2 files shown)


indietyp:
* **#7873** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7873?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/hashql/hir/src/lower/thunking.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// replacement attempts for the same node.
+    fn trampoline(&mut self, node: Node<'heap>) {
+        match &mut self.trampoline {
+            None => {
+                self.trampoline = Some(node);
+            }
+            Some(_) => {
+                panic!("trampoline has been inserted to multiple times");
+            }
+        }
```
TimDiekmann: `if self.trampoline.is_none()`?

indietyp:
> What happens if you thunk a variable with a heavy calculation and need that variable twice, wouldn't you calculate the heavy calculation twice?

That is correct, and this is precisely why we differentiate between `Thunk` and `Closure` (also in the MIR), the idea here is that the computation of a thunk only happens _once_, but happens lazily. This evaluation may happen during constant folding or during the runtime, but should only occur once (similarly to `OnceCell` or `const` in Rust)

> > modifies a Cargo-publishable library, but it is not yet ready to publish
> 
> That is not correct

Sorry, I overlooked that section when submitting. It has been changed.

libs/@local/hashql/hir/src/lower/thunking.rs
indietyp: done in a1ecbd0d8aa07e04ee1e44266104612a03cc2ad6

PR #7883, opened 2025-10-18
FE-42: Design System: Allow refraction Bezel > Radius

kube:

Currently, it's impossible to have a Bezel > Radius of the object we want to apply refraction to.

This causes issues in cases Border Radius is small, as the refraction effect won't be visible.

This PR allows to define Bezel > Radius.

## 🔗 Related links

An Observable notebook with visual explanation of how we achieve this:

- Use center of the square as target of every "normal" at border (not a normal anymore, but radius of the bezel)
- This defines a cone instead of a quarter circle
- For any point inside this cone:
    - Calculate its angle relative to center of square, and project linearly what it would be inside the quarter circle.
    - Intersection point in quarter circle = intersection point in cone.
- For any point outside the cone:
    - Just find the intersection at the square border (trivial)

https://observablehq.com/d/879ebd7e070ed87c

## 🔍 What does this change?

- Storybook: Adds internal story for Displacement Map. (for debugging)
- Add `calculate-rounded-square-map.ts`​, variation of `calculate-circle-map.ts`​ that computes rounded square maps, as its name implies.

This PR:

- [x] modifies an **npm**\-publishable library and **I have added a changeset file(s)**
- [ ] modifies a **Cargo**\-publishable library and **I have amended the version**
- [ ] modifies a **Cargo**\-publishable library, but **it is not yet ready to publish**

The changes in this PR:

    - _Provide more detail here_

The changes in this PR:

## ⚠️ Known issues

- When Border Radius comes close to zero, it shows some lighter diagonals.
    - This should not cause much problem for now as we won't have entirely squared objects, and could be enhanced later.
- Antialiasing still could be enhanced
    - Different antialiasing rendering between `bezel <= radius`​ and `bezel > radius`​)

## 🐾 Next steps

Maybe refine distance calculation later, to make things smoother.

## 🛡 What tests cover this?

- Not automatically tested for now, as will certainly evolve, and is purely graphical.
- We could maybe define unit tests later, by writing some assertions for points inside the cone and outside the cone. (Verify distance to border is properly calculated)

## ❓ How to test this?

1. Checkout the branch
2. Start the storybook: `yarn workspace @hashintel/ds-components run storybook`​
3. Navigate to either:
    1. Internal > Displacement Map
    2. Playground

### Interactive Playground Modal with Bezel > Radius

Radius was made smaller to highlight the fact that refraction effect is independent now.

### Observable Notebook

[Kapture 2025-10-19 at 15.41.11.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/7a5286d4-720d-4c4a-bb74-e13881496be4.mp4)

### DIFF
--- a/libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
+++ b/libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
@@ -13,7 +13,7 @@ function calculateBorderIntersection(
   radius: number,
   cornerWidth: number,
   x: number,
-  y: number
+  y: number,
 ) {
   const angleStart = Math.atan2(cornerWidth - radius, cornerWidth);
   const angleEnd = Math.atan2(cornerWidth, cornerWidth - radius);
@@ -81,7 +81,7 @@ export function calculateRoundedSquareMap(props: {
   const radius = Math.min(props.radius, width / 2, height / 2);
   const cornerWidth = Math.max(
     radius,
-    Math.min(props.maximumDistanceToBorder ?? 0, width / 2, height / 2)
+    Math.min(props.maximumDistanceToBorder ?? 0, width / 2, height / 2),
   );
 
   const widthBetweenCorners = width - cornerWidth * 2;
@@ -106,23 +106,23 @@ export function calculateRoundedSquareMap(props: {
       const x = isOnLeftSide
         ? bufferX - cornerWidth
         : isOnRightSide
-        ? bufferX - cornerWidth - widthBetweenCorners
-        : 0;
+          ? bufferX - cornerWidth - widthBetweenCorners
+          : 0;
 
       // Virtual y value
       // When not on sides, value is 0 to stretch circle into rounded rectangle.
       const y = isOnTopSide
         ? bufferY - cornerWidth
         : isOnBottomSide
-        ? bufferY - cornerWidth - heightBetweenCorners
-        : 0;
+          ? bufferY - cornerWidth - heightBetweenCorners
+          : 0;
 
       // Find the intersection point on the border of the rounded square
       const [intersectionX, intersectionY] = calculateBorderIntersection(
         radius,
         cornerWidth,
         x,
-        y
+        y,
       );
 
       const distanceToCenterSquared = x * x + y * y;
@@ -162,7 +162,7 @@ export function calculateRoundedSquareMap(props: {
           distanceFromBorder,
           distanceFromBorderRatio,
           angle,
-          opacity
+          opacity,
         );
       }
     }


kube:
* **#7886** <a href="https://app.graphite.com/github/pr/hashintel/hash/7886?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 4  dependent PRs ([#7887](https://github.com/hashintel/hash/pull/7887) <a href="https://app.graphite.com/github/pr/hashintel/hash/7887?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7888](https://github.com/hashintel/hash/pull/7888) <a href="https://app.graphite.com/github/pr/hashintel/hash/7888?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7911](https://github.com/hashintel/hash/pull/7911) <a href="https://app.graphite.com/github/pr/hashintel/hash/7911?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 1 other)
* **#7901** <a href="https://app.graphite.com/github/pr/hashintel/hash/7901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7897** <a href="https://app.graphite.com/github/pr/hashintel/hash/7897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7883** <a href="https://app.graphite.com/github/pr/hashintel/hash/7883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/7883?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
```diff
@@ -34,56 +34,68 @@ type Parts = {
 
 function splitImageDataToParts(
   imageData: ImageData,
-  radius_: number,
+  cornerWidth_: number,
   pixelRatio: number,
 ): Parts {
-  const radius = radius_ * pixelRatio;
+  const cornerWidth = cornerWidth_ * pixelRatio;
```
CiaranMn: Felt a bit weird to me to have a `cornerWidth` from the args and then to create a new `cornerWidth`... maybe we could call this `scaledCornerWidth` or something to make it immediately clearer?

The same applies to `lateralPartSize`

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
```diff
@@ -105,7 +117,7 @@ function splitImageDataToParts(
```
CiaranMn, line 118: It's in the line above, but the comment should be a JSDoc string so that it shows up in IDE hints, i.e.
```ts
/**
 * Component that renders the 8 parts of an image and composites them together.
 * Used internally by the Filter component, for DisplacementMap and SpecularMap.
 */
type CompositePartsProps = {
```

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
```diff
@@ -292,8 +304,17 @@ const FILTER: React.FC<FILTER_PROPS> = memo(
...
-    // Calculated image will always be a square that contains 4 corners + 3 pixels for middle
-    const imageSide = useTransform(() => radius.get() * 2 + LATERAL_PART_SIZE);
+    // Size of each corner area
+    // If bezelWidth < radius, corners will be in a circle shape
+    // If bezelWidth >= radius, corners will be in a rounded square shape
+    const cornerWidth = useTransform(() =>
+      Math.max(radius.get(), bezelWidth.get()),
+    );
+
+    // Calculated image will always be a square that contains 4 corners + 1 pixel for middle
```
CiaranMn: To avoid this comment and the actual value for `LATERAL_PART_SIZE` going out of sync. The value will be given on hover
```suggestion
    /** Calculated image will always be a square that contains 4 corners + {@link LATERAL_PART_SIZE} pixels for middle */
```

libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  // Calculate radiusPlusOne to antialias the border (1 pixel outside)
+  const radiusPlusOne = radius + 1;
+  const radiusPlusOneSquared = radiusPlusOne ** 2;
+
+  const angleStart = Math.atan2(cornerWidth - radius, cornerWidth);
+  const angleEnd = Math.atan2(cornerWidth, cornerWidth - radius);
+  const aperture = angleEnd - angleStart;
+
+  for (let y1 = 0; y1 < height; y1++) {
```
CiaranMn: Why `y1` / `x1` for these variables rather than `x` and `y` and call the inner thing something else, e.g. `virtualX`?

Not a big deal, just looked a bit odd

libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
```diff
@@ -0,0 +1,157 @@
...
+      // When not on sides, value is 0 to stretch circle into rounded rectangle.
+      const y = isOnTopSide
+        ? y1 - cornerWidth
+        : isOnBottomSide
+          ? y1 - cornerWidth - heightBetweenCorners
+          : 0;
+
+      const pointAngleInSquare = Math.atan2(Math.abs(y), Math.abs(x));
+
+      function calculateBorderIntersection() {
```
CiaranMn: At the moment we're creating x*y copies of this function, I might be better to declare it outside the loop and take `x`, `y` and `pointAngleInSquare` as arguments.

libs/@hashintel/ds-components/src/lib/maps/process-pixel.type.ts
```diff
@@ -0,0 +1,14 @@
+export type ProcessPixelFunction = (
```
CiaranMn, line 1: We can just call the file `.ts` – generally elsewhere in the repo `.d.ts` is only for generated declaration files, or type patching for imported libraries. Even if a file only contains types we can call it `.ts`

libs/@hashintel/ds-components/src/lib/maps/process-pixel.type.ts
```diff
@@ -0,0 +1,14 @@
...
+  buffer: Uint8ClampedArray<ArrayBufferLike>,
+  offset: number,
+  distanceFromCenter: number,
+  distanceFromBorder: number,
+  distanceFromBorderRatio: number,
+  /**
+   * Angle from center to pixel in radians.
+   */
+  angle: number,
+  opacity: number,
```
CiaranMn, line 13: At this number of arguments, most of which are the same type, I would definitely pass it as an object, otherwise it's easy to mix them up.

libs/@hashintel/ds-components/src/stories/displacement-map.stories.tsx
```diff
@@ -0,0 +1,258 @@
...
+  CONCAVE,
+  CONVEX,
+  LIP,
+  type SurfaceFnDef,
+} from "../lib/surface-equations";
+
+// Define the bezel height function types
+type BezelHeightType = "CONVEX" | "CONCAVE" | "LIP";
+
+// Define the args type for the displacement map function parameters
```
CiaranMn: These comments seem a bit redundant, the type names are good enough.

libs/@hashintel/ds-components/src/lib/maps/process-pixel.type.ts
kube: On this one it is on purpose, to prevent creating a new JavaScript object for every pixel, that will need to be garbage collected.

libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
kube: I was hoping that this was the kind of thing that V8 would inline given it only depends on the current scope, and is not exposed anywhere.

But yes, not sure, so I will do the update.

libs/@hashintel/ds-components/src/lib/maps/process-pixel.type.ts
CiaranMn: Makes sense, thanks!

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
kube: This was more to split the file, not for JSDoc.

The actual JSDoc should be on CompositeParts, this line is the Props type.
I'm adding one.

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
kube: I don't know how I never learned about `@link`.
Thanks!

libs/@hashintel/ds-components/src/lib/flexible-filter.tsx
kube: I replaced by an object props, and replaced this by:

```ts
  const cornerWidth = props.cornerWidth * props.pixelRatio;
  const lateralPartSize = LATERAL_PART_SIZE * props.pixelRatio;
```

My goal is to keep the variable naming inside the actual calculations simple.
(i.e. It should be abstracted of "scaling" inside this logic, and should be done upstream)

libs/@hashintel/ds-components/src/stories/displacement-map.stories.tsx
kube: Yes, the story was generated by Copilot.

libs/@hashintel/ds-components/src/lib/maps/calculate-rounded-square-map.ts
kube: I prefer to add the "name complexity" outside of the actual calculation.
So where the actual "equations" are defined, we use x/y and not virtualX/virtualY.

I replaced x1/y1 by bufferX/bufferY instead.

libs/@hashintel/ds-components/src/lib/maps/process-pixel.type.ts
kube: Renamed to process-pixel.type.ts

PR #7884, opened 2025-10-18
BE-165: HashQL: Vendor in rustc `BitSet` implementation

indietyp:
# 🌟 What is the purpose of this PR?

This PR adds efficient bit vector data structures from the Rust compiler codebase to our project. These provide optimized implementations for representing and manipulating sets of elements, which will be useful for various algorithms that need to track and operate on collections of IDs.

# 🔗 Related links

- Original source: https://github.com/rust-lang/rust/blob/f5242367f4acf0130c9b68e6956531b2cb73bd38/compiler/rustc_index/src/bit_set.rs

# 🔍 What does this change?

- Adds several bit vector implementations with different performance characteristics:
  - `DenseBitSet`: Fixed-size bitset with dense representation
  - `ChunkedBitSet`: Fixed-size bitset with partially dense/sparse representation
  - `MixedBitSet`: Combines DenseBitSet and ChunkedBitSet for optimal performance
  - `GrowableBitSet`: Resizable bitset with dense representation
  - `BitMatrix`: Fixed-size 2D bit matrix
  - `SparseBitMatrix`: Variable-row-size 2D bit matrix with sparse representation
  - `FiniteBitSet`: Fixed-size bitset represented by an integer type
- Includes comprehensive test suite for all implementations
- Adapts the code to use our `Id` trait instead of rustc's `Idx` trait

# Pre-Merge Checklist 🚀

## 🚢 Has this modified a publishable library?

This PR:

The changes in this PR:

## 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

# 🛡 What tests cover this?

- Comprehensive test suite ported from the original Rust compiler implementation
- Tests cover all data structures and their operations

### DIFF
--- a/libs/@local/hashql/core/src/id/bit_vec/tests.rs
+++ b/libs/@local/hashql/core/src/id/bit_vec/tests.rs
@@ -17,13 +17,11 @@
 //!
 //! Implementation and maintenance:
 //! - Applied clippy-driven fixes (no intended semantic changes).
+//! - Removed benchmarks
 #![expect(clippy::too_many_lines, clippy::min_ident_chars, clippy::unwrap_used)]
-extern crate test;
 
 use alloc::rc::Rc;
-use core::{hint::black_box, marker::PhantomData, ops::RangeBounds};
-
-use test::Bencher;
+use core::{marker::PhantomData, ops::RangeBounds};
 
 use super::GrowableBitSet;
 use crate::{
@@ -799,39 +797,4 @@ fn dense_contains_any() {
 
     set.insert(TestId::from_usize(22));
     assert!(set.contains_any(TestId::from_usize(20)..TestId::from_usize(59)));
-}
-
-newtype!(struct BenchId(u32 is 0..=u32::MAX));
-
-#[bench]
-fn bench_insert(b: &mut Bencher) {
-    let mut bs = DenseBitSet::new_filled(99999_usize);
-    b.iter(|| {
-        black_box(bs.insert(black_box(BenchId::from_u32(100))));
-    });
-}
-
-#[bench]
-fn bench_remove(b: &mut Bencher) {
-    let mut bs = DenseBitSet::new_filled(99999_usize);
-    b.iter(|| {
-        black_box(bs.remove(black_box(BenchId::from_u32(100))));
-    });
-}
-
-#[bench]
-fn bench_iter(b: &mut Bencher) {
-    let bs = DenseBitSet::new_filled(99999_usize);
-    b.iter(|| {
-        bs.iter().map(|b: BenchId| black_box(b)).for_each(drop);
-    });
-}
-
-#[bench]
-fn bench_intersect(b: &mut Bencher) {
-    let mut ba: DenseBitSet<BenchId> = DenseBitSet::new_filled(99999_usize);
-    let bb = DenseBitSet::new_filled(99999_usize);
-    b.iter(|| {
-        ba.intersect(black_box(&bb));
-    });
 }
--- a/libs/@local/hashql/core/src/lib.rs
+++ b/libs/@local/hashql/core/src/lib.rs
@@ -28,7 +28,6 @@
     variant_count,
     get_disjoint_mut_helpers
 )]
-#![cfg_attr(test, feature(test))]
 
 extern crate alloc;
 


indietyp:
* **#7916** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7915** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7910** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7910?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7885** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7884** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/hashql/core/src/id/bit_vec/mod.rs
```diff
@@ -0,0 +1,2025 @@
...
+}
+
+/// A fixed-size bitset type with a dense representation.
+///
+/// Note 1: Since this bitset is dense, if your domain is big, and/or relatively
+/// homogeneous (for example, with long runs of bits set or unset), then it may
+/// be preferable to instead use a [`MixedBitSet`], or an
+/// [`IntervalSet`](crate::interval::IntervalSet). They should be more suited to
+/// sparse, or highly-compressible, domains.
+///
```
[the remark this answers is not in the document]
indietyp: fixed in v2

libs/@local/hashql/core/src/id/bit_vec/tests.rs
```diff
@@ -0,0 +1,837 @@
...
+}
+
+#[bench]
+fn bench_intersect(b: &mut Bencher) {
+    let mut ba: DenseBitSet<BenchId> = DenseBitSet::new_filled(99999_usize);
+    let bb = DenseBitSet::new_filled(99999_usize);
+    b.iter(|| {
+        ba.intersect(black_box(&bb));
+    });
+}
```
TimDiekmann: These benches are not run as part of CI

libs/@local/hashql/core/src/id/bit_vec/mod.rs
```diff
@@ -0,0 +1,2026 @@
...
+//!
+//! Implementation and maintenance:
+//! - Removed rustc-specific serialization code.
+//! - Migrated to smallvec v2.
+//! - Replaced `unwrap` with `unwrap_or_else(|| unreachable!(...))` where applicable.
+//! - Applied clippy-driven fixes (no intended semantic changes).
+//! - Removed inherent impls for bit relations.
+//! - Removed used of `static_assert_size` macro in favor of `assert_eq!`.
+//! - Removed mentions of `IntervalSet`.
+#![coverage(off)] // Vendored in module with tests
```
TimDiekmann: Yes, this is vendored in, but we have tests for that, so let's keep this on, please. It would be useful to see in codecov if vendored-in code is **not** tested.

TimDiekmann (changes requested):
Do we really need all 7 bit sets? This seems overkill. We don't have benchmarks for HashQL yet, still, we optimize **_heavily_**. 
The PR description lacks the reason for what we need all the different bit sets.
What are the advantages of this vendored-in module in comparison to other bites implementation (such as [`bit_set`](https://docs.rs/bit-set/latest/bit_set/struct.BitSet.html))?

indietyp:
Currently we only really need `MixedBitSet`, but instead of just partially vendoring in, I thought it would be more beneficial to have the same file(s) with minimal modifications (as outlined in the code). The matrices can be helpful, but aren't used by us yet.
The reason why we need (performant) bitsets is that they play a central part in both type-analysis, graphs and dataflow analysis. Some of that being already used in the PR stacked above this.
I've done a large study of alternative crates and the ecosystem before, but it boils down to the following:

* The size of our domain is _static_, once created, the size of what we're actually interested in is unchanging, in a single e.g. dataflow run we keep the amount of variables the same. We do not add new variables (that is true for the MIR and the HIR(ANF)).
* bitset mutations over it are frequent
* dataflow often has regions which are large parts 0s or 1s and then some where it is largely mixed (logically this makes sense, given a function call, most of the time you are going to use the arguments that are right before it, but sometimes arguments that are from e.g. the upper scope, which then have a lower id)
* most of the time we're going to have less than 2048 items, which means we can benefit from a dense representation instead of a sparse one, which is magnitudes slower.

There are some crates that do *some* of this, but not all, and there is no crate that offers the hybrid approach. The hybrid approach is especially valuable if not required in the case where we have _a lot of variables_, but don't actually want to track all of them.

For the sparse implementation, there is:
* roaring
* croaring
* hibitset

Roaring has the problem that, due to its dynamic nature, random insertion performance is terrible (I know this from experience from my work with petgraph). `hibitset` solves some of the issues. Still, any insertion requires access to a 4-level tiered bitset. `hi_sparse_bitset` promises to fix some of these issues, but hasn't seen much adoption, so I am a bit sceptical there. `croaring` has better performance than `roaring`, but is a C dependency, and has the same random insertion performance problems.

For the dense implementations, there's a lot more to choose from:
* bitvec - problem: has been unmaintained for 2 years (although API is stable)
* vob
* bv
* smallbitvec

They all do the same thing: it is a growable vector of `usize` that is addressable over its individual indices.

The key insight is that none of them offer the ability to have both dense and sparse representations, despite having a different API surface. To address this, we would need to create wrapper types that work with our `Id` type. Coming with that, the potential need for a matrix (the type system would benefit from one for one) - where our `Graph` for inference solving currently has its own implementation. It made/makes sense to me to instead of vendor in a solution that has been proven to work instead of spending time on creating my own, as no crate currently exists that does both: sparse and dense implementations over an `Id` trait.

libs/@local/hashql/core/src/id/bit_vec/mod.rs
indietyp: fixed in v5

libs/@local/hashql/core/src/id/bit_vec/tests.rs
indietyp: I wasn't sure if we wanted to run the benches / remove them, or have them skipped. I just ported them over from rustc, but wasn't sure what to do with them.

TimDiekmann: I don't think it's worth keeping it. We won't improve the implementation anyway.

indietyp: removed in v6

libs/@local/hashql/core/src/id/bit_vec/mod.rs
```diff
@@ -0,0 +1,2025 @@
...
+    clippy::too_long_first_doc_paragraph
+)]
+
+use alloc::rc::Rc;
+use core::{
+    fmt, iter,
+    marker::PhantomData,
+    ops::{BitAnd, BitAndAssign, BitOrAssign, Bound, Not, Range, RangeBounds, Shl},
+    slice,
+};
```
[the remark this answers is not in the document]
indietyp: size_of is in the prelude

PR #7895, opened 2025-10-20
FE-53: Remove unused types and update Babel dependencies

indietyp:
## 🌟 What is the purpose of this PR?

Remove unused type dependencies and update several packages to their latest versions.

## 🔍 What does this change?

- Removed unused type dependencies:
  - `@types/set-interval-async` from `hash-realtime`
  - `@types/dompurify` from `person` block
  - `@types/uuid` from `hash-backend-utils`
- Updated Babel plugins to newer transform versions:
  - Replaced `@babel/plugin-proposal-class-properties` with `@babel/plugin-transform-class-properties`
  - Repalced `@babel/plugin-proposal-private-property-in-object` with `@babel/plugin-transform-private-property-in-object`
- Updated `babel-loader` from 9.2.1 to 10.0.0
- Updated `globals` from 15.15.0 to 16.4.0

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Existing tests should continue to pass with these dependency updates

## ❓ How to test this?

1. Checkout the branch
2. Run `yarn install` to update dependencies
3. Verify that all builds and tests pass

PR #7894, opened 2025-10-20
SRE-49: Update AWS SDK dependencies from 3.858.0 to 3.913.0

indietyp:
## 🌟 What is the purpose of this PR?

Update AWS SDK packages from version 3.858.0 to 3.913.0 in the hash-api and hash-backend-utils packages.

## 🔍 What does this change?

- Updates `@aws-sdk/client-s3` from 3.858.0 to 3.913.0
- Updates `@aws-sdk/client-ses` from 3.858.0 to 3.913.0
- Updates `@aws-sdk/credential-provider-node` from 3.858.0 to 3.913.0
- Updates `@aws-sdk/s3-presigned-post` from 3.858.0 to 3.913.0
- Updates `@aws-sdk/s3-request-presigner` from 3.858.0 to 3.913.0

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

Existing tests should cover this dependency update.

## ❓ How to test this?

1. Checkout the branch
2. Run the application and verify that AWS-related functionality (S3, SES) works as expected

PR #7896, opened 2025-10-20
BE-175, BE-176: Update Express and Apollo Server to v5

indietyp:

Update dependencies in the HASH API, including upgrading Express to v5, Snowplow Node Tracker to v4, and other related packages.

## 🔍 What does this change?

- Upgrades `@snowplow/node-tracker` from 3.24.6 to 4.6.7
- Updates Snowplow telemetry implementation to use the new API
- Upgrades `express` from 4.21.2 to 5.1.0
- Upgrades `express-handlebars` from 7.1.3 to 8.0.3
- Upgrades `express-rate-limit` from 7.5.1 to 8.1.0
- Upgrades `body-parser` from 1.20.3 to 2.2.0
- Upgrades `@apollo/server` to 5
- Upgrades `redis` to 5
- Removes unused `cross-env` dependency
- Removes `AsyncRedisClient`

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing tests should cover this functionality

## ❓ How to test this?

1. Checkout the branch
2. Start the API service
3. Confirm that the API functions correctly with the updated dependencies

### DIFF
--- a/apps/hash-api/src/graphql/error.ts
+++ b/apps/hash-api/src/graphql/error.ts
@@ -1,14 +1,12 @@
 import { ApolloServerErrorCode } from "@apollo/server/errors";
 import { GraphQLError } from "graphql";
 
-export const any = (message: string, extensions?: Record<string, unknown>) =>
-  new GraphQLError(message, { extensions });
 export const code = (
   // eslint-disable-next-line @typescript-eslint/no-shadow
   code: string,
   message: string,
   extensions?: Record<string, unknown>,
-) => any(message, { code, ...extensions });
+) => new GraphQLError(message, { extensions: { code, ...extensions } });
 
 export const badUserInput = (
   message: string,
--- a/apps/hash-api/src/graphql/resolvers/knowledge/block-collection/update-block-collection-contents.ts
+++ b/apps/hash-api/src/graphql/resolvers/knowledge/block-collection/update-block-collection-contents.ts
@@ -144,7 +144,7 @@ export const updateBlockCollectionContents: ResolverFn<
     };
   } catch (error) {
     if (error instanceof globalThis.Error) {
-      throw Error.any(`Could not apply update: ${error.message}`);
+      throw Error.internal(`Could not apply update: ${error.message}`);
     }
 
     throw Error.internal(`Could not apply update: ${JSON.stringify(error)}`);


indietyp:
* **#7939** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7939?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7903** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7902** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7896** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7895** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7895?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7894** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7893** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/hash-api/src/index.ts
```diff
@@ -192,11 +198,8 @@ const main = async () => {
...
 
   // Connect to Redis
-  const redis = new RedisCache(logger, {
-    host: redisHost,
-    port: redisPort,
-    tls: redisEncryptedTransit,
-  });
+  const redis = await createRedisClient({ url: redisUrl, logger }).connect();
+  const keyv = new Keyv({ store: new KeyvRedis(redis) });
   shutdown.addCleanup("Redis", async () => redis.close());
```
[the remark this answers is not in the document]
indietyp: it's the other way around...

apps/hash-realtime/src/config.ts
```diff
@@ -3,37 +3,32 @@ import type { Logger } from "@local/hash-backend-utils/logger";
...
+import { createRedisClient } from "@local/hash-backend-utils/redis";
 
 // The tables to monitor for changes
 export const MONITOR_TABLES = supportedRealtimeTables.map(
   (table) => `public.${table}`,
 );
 
 // The realtime service will push updates from the Postgres change stream to the following queues.
-export const generateQueues = (
+export const generateQueues = async (
```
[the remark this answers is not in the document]
indietyp: That is up to the caller, it wasn't before

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
```diff
@@ -148,7 +150,7 @@ export const userBeforeEntityUpdateHookCallback: BeforeUpdateEntityHookCallback
        * and prevent them from receiving ownership of the web.
        */
       if (!(await userHasAccessToHash(context, authentication, user))) {
-        throw new Error(
+        throw Error.any(
```
CiaranMn:
```suggestion
        throw Error.forbidden(
```

apps/hash-api/src/graphql/resolvers/knowledge/block-collection/update-block-collection-actions.ts
```diff
@@ -170,10 +171,14 @@ export const handleCreateNewEntity = async (params: {
...
-      throw new UserInputError(`action ${params.index}: ${error}`);
+    if (
+      error instanceof GraphQLError &&
+      error.extensions.code === ApolloServerErrorCode.BAD_USER_INPUT
+    ) {
+      throw Error.badUserInput(`action ${params.index}: ${error}`);
     }
-    throw new Error(
+
+    throw Error.any(
```
CiaranMn: In some places we translate `new Error` to `any`, in some places `internal`.

I'd just go for `internal` and possibly get rid of `any` entirely.
```suggestion
    throw Error.internal(
```

apps/hash-api/src/graphql/resolvers/knowledge/block-collection/update-block-collection-actions.ts
```diff
@@ -260,10 +265,14 @@ export const handleInsertNewBlock = async (
...
```
CiaranMn:
```suggestion
    throw Error.internal(
```

apps/hash-api/src/graphql/resolvers/knowledge/block-collection/update-block-collection-contents.ts
```diff
@@ -143,10 +143,10 @@ export const updateBlockCollectionContents: ResolverFn<
...
     };
   } catch (error) {
-    if (error instanceof Error) {
-      throw new Error(`Could not apply update: ${error.message}`);
+    if (error instanceof globalThis.Error) {
+      throw Error.any(`Could not apply update: ${error.message}`);
     }
 
-    throw new Error(`Could not apply update: ${JSON.stringify(error)}`);
+    throw Error.any(`Could not apply update: ${JSON.stringify(error)}`);
```
CiaranMn:
```suggestion
    throw Error.internal(`Could not apply update: ${JSON.stringify(error)}`);
```

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
```diff
@@ -6,8 +6,9 @@ import {
 import { addActorGroupAdministrator } from "@local/hash-graph-sdk/principal/actor-group";
 import { isUserHashInstanceAdmin } from "@local/hash-graph-sdk/principal/hash-instance-admins";
 import type { UserProperties } from "@local/hash-isomorphic-utils/system-types/user";
-import { ApolloError, UserInputError } from "apollo-server-express";
+import { GraphQLError } from "graphql";
 
+import * as Error from "../../../../../graphql/error";
```
CiaranMn, line 11: Generally not a fan of overwriting language tokens, but I'm guessing in these cases it's to stop people from calling `new Error`

apps/hash-api/src/graphql/error.ts
```diff
@@ -0,0 +1,40 @@
+import { ApolloServerErrorCode } from "@apollo/server/errors";
+import { GraphQLError } from "graphql";
+
+export const any = (message: string, extensions?: Record<string, unknown>) =>
+  new GraphQLError(message, { extensions });
```
CiaranMn: Suggest we delete this to force people to pick or specify a message.

Think Apollo might default to `INTERNAL_SERVER_ERROR` here anyway? I think it did before where codes weren't specified. If so, might as well make people explicitly choose `internal`.

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
```diff
@@ -186,9 +184,8 @@ export const acceptOrgInvitationResolver: ResolverFn<
   if (!creatorIsOrgAdmin) {
     await archiveInvitation();
 
-    throw new ApolloError(
+    throw Error.forbidden(
       "Invitation issuer is not an administrator of the organization",
-      "UNAUTHORIZED",
     );
```
CiaranMn, line 189: This is technically a behaviour change, but the codebase is already incorrectly using `FORBIDDEN` for both unauthenticated (i.e. `UNAUTHORIZED` in http terms, boo) and authenticated but not authorized (`FORBIDDEN`, which it should be in this case, so this change is fine). I don't suggest we try and fix that conflation in this PR. Tracked in BE-174.

apps/hash-api/src/graphql/create-apollo-server.ts
```diff
@@ -9,30 +20,141 @@ import type { VaultClient } from "@local/hash-backend-utils/vault";
...
+
+        // take the first part of the UA to help identify browser vs server requests
+        const userAgent = ctx.request.http?.headers
+          .get("user-agent")
+          ?.split(" ")[0];
+
+        const msg = {
+          operation: ctx.operationName,
+          elapsed: `${elapsed.toFixed(2)}ms`,
+          clientInfo: defaultGenerateClientInfo(ctx),
```
CiaranMn, line 116: A bit of a codebase inconsistency at the moment, but we actually have a `hashClientHeaderKey` we set in the FE (to `web-app`). But we should just use the `name` and `version` args when constructing `ApolloClient` and get rid of the custom header. Tracked in BE-174.

CiaranMn (commented):
Thanks Bilal! Just one suggested change to discuss, which is to get rid of `any` error type.

apps/hash-api/src/graphql/create-apollo-server.ts
indietyp: I didn't even know the custom header existed! (which is why I didn't include it in the change). Should I just keep it this way then and defer it to BE-174?

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
indietyp: Should I revert this change then and instead we do it in BE-174?

apps/hash-api/src/graphql/error.ts
indietyp: I wasn't sure either (which is why I added `any` in the first place. I think removing it and defaulting to `internal` instead would be a good thing. Agreed. Fixed in ceca8de21a8be0c845d83cd72fbc6928fd1727db

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
indietyp: I am usually very cautious about this as well, but yes that was the intention - it forces the user in the graphql context to choose an error by "automatically" erroring out when using `new Error`. Places where GraphQL is secondary, I have aliased this to `GraphQlError` instead. I can also do it here – although found it through the transition helpful enough that I thought we might want to adopt it in this case.

CiaranMn: The more robust way to do this would be with an ESLint rule applied to the relevant folders (if we can identify by folder), as the current approach won't stop anyone from calling `new Error` in files where the import statement isn't already present.

Not a big deal though.

Something like this.
```
  {
    files: [
      "**/src/graphql/**/*.ts",
      "**/src/graph/knowledge/**/*.ts",
      "**/src/graph/ontology/**/*.ts",
    ],
    rules: {
      "no-restricted-syntax": [
        "error",
        {
          selector: "NewExpression[callee.name='Error']",
          message:
            "Please construct a GraphQL error by importing from src/graphql/error instead",
        },
      ],
    },
  },
```

Screenshot 2025-10-24 at 16 29 43

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
CiaranMn: No, we can leave this as it is, at least this way everything is `FORBIDDEN` rather than having one random `UNAUTHORIZED`.

CiaranMn: I'm tempted to have `UNAUTHENTICATED` and `UNAUTHORIZED` as the errors, despite being inconsistent with HTTP codes (so annoying, the inconsistency means there's no great solution as there is cause for confusion in either direction)

apps/hash-api/src/graphql/create-apollo-server.ts
CiaranMn: Yes, let's just keep this. The confusion was already present because in this error handler we use the default client headers Apollo provides, but in some other logging context we take the client from the custom header.

apps/hash-api/src/graphql/resolvers/knowledge/org/accept-org-invitation.ts
indietyp: considering that Apollo and GraphQL are already outside of normal HTTP status codes (see `BAD_USER_INPUT`) I think using unauthenticated and unauthorized here would be better.

apps/hash-api/src/graph/knowledge/primitive/entity/before-update-entity-hooks/user-before-update-entity-hook-callback.ts
indietyp: I would potentially explore this in a follow PR if that is alright, I have my problems with this, because from last I remember working with eslint no-restricted-syntax is tricky in re to overrides and false positives, but I think in this case it should be fine.

CiaranMn: Yep, sounds good!

PR #7914, opened 2025-10-24
FE-61: Add Petrinaut dev mode

CiaranMn:

The main purpose of the PR is to add a `yarn dev` mode to Petrinaut for local development.

As a drive-by it also:
1. fixes a few potential bugs to do with the switch from `useLocalStorageState` from `rooks` to `useLocalStorage` from `@mantine/hooks`: the latter runs the initial request to local storage in a `useEffect`, which means it's _always_  empty on first render. This is for SSR compatibility, but may change behaviour in the places we used the old hook, which set the state synchronously. Since the places we're using it aren't going to be SSR'd anyway, I've added the option to not get the initial value in an effect.
2. Move `defaultTokenTypes` export into its own file to help hot module reloading.
3. Don't import the bundled JS file, but instead from the library.

This PR:

- [x] modifications do not need publishing. Can be rolled into next release, only very minor changes to internal Petrinaut implementation.

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. `yarn && yarn dev` in `libs/@hashintel/petrinaut`


libs/@hashintel/petrinaut/dev/vite.config.ts
```diff
@@ -0,0 +1,13 @@
...
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [react()],
+  root: "dev",
+  server: {
+    open: true,
+  },
+  define: {
+    "process.env.NODE_ENV": '"development"',
```
kube: You normally do not need that with Vite.

libs/@hashintel/petrinaut/dev/vite.config.ts
```diff
@@ -0,0 +1,13 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [react()],
+  root: "dev",
```
kube, line 7:
```suggestion
  mode: "development",
  plugins: [react()],
  root: "dev",
```

kube (approved):
Ideally we can just have one (possibly parameterized) config for dev and prod.

But that's just something we can iterate over.

libs/@hashintel/petrinaut/dev/vite.config.ts
CiaranMn: Thanks, removed!

vilkinsons (approved):

PR #7910, opened 2025-10-23
BE-171: HashQL: transfer type information in HIR transformations

indietyp:
## 🌟 What is the purpose of this PR?

Fix HIR node identity to survive tree transformations. The previous HIR architecture automatically generated new HirIds whenever nodes were modified, making it impossible to track auxiliary data (types, generic arguments) through HIR transformations like specialization and normalization. This PR decouples node identity from structure by making HirId assignment selective and introducing a centralized `HirMap` for auxiliary data storage.

## 🔍 What does this change?

- **Restructure Node architecture**: `Node<'heap>` is now `Interned<'heap, NodeData<'heap>>` instead of a direct struct
- **Remove automatic ID generation**: Eliminated `Decompose` trait and `PartialNode` - nodes are directly interned with selective HirId assignment
- **Introduce HirMap system**: New `HirMap` stores auxiliary data (type IDs, generic arguments) in `HirContext.map`, separate from node structure
- **Enable stable identity**: HirIds are reused during transformations (only assigned new IDs for expansion, freed only when nodes are completely erased)
- **Add explicit data transfer**: `HirMap::copy_to()` enables controlled auxiliary data transfer when identity changes are needed
- **Switch to dense storage**: Controlled ID space allows `HirIdVec` instead of sparse maps, reducing memory overhead
- **Update all visitors**: HIR passes now take `Node` by value and preserve auxiliary data across transformations
- **Centralize type access**: Pretty printing, type checking, and compilation now access type information through the unified map

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

PR #7915, opened 2025-10-24
BE-177: HashQL: Desguar complex binary operations into if expressions

indietyp:

This PR adds support for conditional expressions (`if/then/else`) in graph filter contexts by transforming them into equivalent boolean logic expressions. This enables more expressive filter conditions in graph queries.

Additionally, it desugars `A && B` and `A || B` into if expressions during normalization into if expressions, to aid in MIR transformations.

## 🔗 Related links

- N/A

## 🔍 What does this change?

- Adds support for `if` expressions in graph filter contexts by transforming them into equivalent boolean logic
- Implements optimizations for common patterns:
  - `if A then true else C` → `A || C`
  - `if A then B else false` → `A && B`
  - General case: `if A then B else C` → `(A && B) || (!A && C)`
- Updates the `Normalization` class to pass the environment to support these transformations
- Adds helper function `is_bool` to detect boolean literals for optimization purposes
- Improves documentation in the normalization module
- Normalize `A && B` and `A || B` when converting to HIR(ANF) if `B` isn't an atom into an `if` expression

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing tests have been updated to reflect the new behavior

## ❓ How to test this?

1. Checkout the branch
2. Run the test suite to verify that conditional expressions in filters now work
3. Try writing a graph query with an `if` expression in a filter condition

### DIFF
--- a/libs/@local/hashql/eval/src/graph/read/filter.rs
+++ b/libs/@local/hashql/eval/src/graph/read/filter.rs
@@ -28,16 +28,15 @@ use super::{
 ///
 /// ```ignore
 /// // Returns true if the node represents `true`
-/// assert!(is_bool(true_node, true));
+/// assert!(is_bool_literal(true_node, true));
 /// // Returns false if the node represents `false` when checking for `true`
-/// assert!(!is_bool(false_node, true));
+/// assert!(!is_bool_literal(false_node, true));
 /// ```
-#[expect(clippy::wildcard_enum_match_arm)]
-fn is_bool(node: Node<'_>, value: bool) -> bool {
-    match node.kind {
-        NodeKind::Data(Data::Primitive(Primitive::Boolean(constant))) => constant == value,
-        _ => false,
-    }
+fn is_bool_literal(node: Node<'_>, value: bool) -> bool {
+    matches!(
+        node.kind,
+        NodeKind::Data(Data::Primitive(Primitive::Boolean(constant))) if constant == value
+    )
 }
 
 impl<'env, 'heap: 'env> GraphReadCompiler<'env, 'heap> {
@@ -186,7 +185,7 @@ impl<'env, 'heap: 'env> GraphReadCompiler<'env, 'heap> {
             // - When A is true: `(true && B) || (false && C)` = `B || false` = `B`
             // - When A is false: `(false && B) || (true && C)` = `false || C` = `C`
             NodeKind::Branch(Branch::If(If { test, then, r#else })) => {
-                if is_bool(then, true) {
+                if is_bool_literal(then, true) {
                     // Optimization: `if A then true else C` → `A || C`
                     // When A is true, result is true; when A is false, result is C
                     let mut sink = sink.or();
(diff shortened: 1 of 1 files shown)


indietyp:
* **#7916** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7915** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7910** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7910?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7885** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7884** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -13,11 +16,30 @@ use hashql_hir::node::{
...
+/// # Examples
+///
+/// ```ignore
+/// // Returns true if the node represents `true`
+/// assert!(is_bool(true_node, true));
+/// // Returns false if the node represents `false` when checking for `true`
+/// assert!(!is_bool(false_node, true));
+/// ```
+#[expect(clippy::wildcard_enum_match_arm)]
+fn is_bool(node: Node<'_>, value: bool) -> bool {
```
TimDiekmann: I think `is_bool_primitive` or `is_bool_literal` would be more unambiguous. Otherwise, it could be interpreted if `node` evaluates to `value`

libs/@local/hashql/eval/src/graph/read/filter.rs
```diff
@@ -13,11 +16,30 @@ use hashql_hir::node::{
...
+    match node.kind {
+        NodeKind::Data(Data::Primitive(Primitive::Boolean(constant))) => constant == value,
+        _ => false,
+    }
```
TimDiekmann: This could be written using a `matches!` statement:

Something like should work:
```suggestion
    matches!(
        node.kind,
        NodeKind::Data(Data::Primitive(Primitive::Boolean(constant))) if constant == value
    );
```

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: done in 69eae4c0efe652749a4467755b7fcb17324dcf87

libs/@local/hashql/eval/src/graph/read/filter.rs
indietyp: done in 69eae4c0efe652749a4467755b7fcb17324dcf87

PR #7890, opened 2025-10-20
BE-167: Group policy calls for entity traversal

TimDiekmann:
## 🌟 What is the purpose of this PR?

In order to implement BE-142 without BE-141 we need to do the actual traversal calls before we check the policies.

It turns out, that in most cases this is a lot faster as in the common case we fetch only traverse entities which we also can see. This saves us quite a few queries.

## 🔗 Related links

- Performance optimization for complex graph traversals

## 🔍 What does this change?

- Refactors entity traversal to use a two-phase approach: first traverse all edges, then filter by permissions
- Adds a new `traverse_edges_cte` method that uses a recursive CTE for efficient traversal
- Implements a fallback `traverse_edges_sequential` method for cases where the CTE cannot be used
- Improves the `TraversalContext` to properly merge temporal intervals with union semantics
- Adds comprehensive tests for interval merging in the `TraversalContext`
- Adds a new `entity_with_intervals` method to the `Subgraph` for easier entity iteration
- Splits traversal path handling to separate entity edges from ontology edges

## 🛡 What tests cover this?

- Added unit tests for interval merging in `TraversalContext`

## ❓ How to test this?

1. Run complex graph traversal queries
2. Verify that performance is improved for deep traversals
3. Confirm that results are identical to the previous implementation

PR #7928, opened 2025-10-27
BE-142: Enable single-query entity traversal using PostgreSQL recursive CTEs

TimDiekmann:

This PR implements a recursive CTE (Common Table Expression) for entity traversal, replacing the N+1 query pattern where `read_knowledge_edges` was called sequentially for each edge in a traversal path.

The implementation executes a single PostgreSQL recursive CTE that traverses all edges at once, significantly reducing the number of database round-trips for graph traversal operations. This is particularly beneficial for deep entity traversals where the previous approach would make one query per hop.

**Key constraint**: To address performance limitations of the UNION ALL approach with deep recursion, this implementation currently applies the CTE optimization only to shallow traversals (1 hop = 2 edges). Deeper traversals fall back to sequential queries until a denormalized edge table can be implemented (see Next steps).

## 🔗 Related links

- Linear: [BE-142: Implement recursive CTEs when traversing entities](https://linear.app/hash/issue/BE-142)
- Related PR: [#7890 BE-167: Group policy calls for entity traversal](https://github.com/hashintel/hash/pull/7890) - sets up the two-phase traversal approach that this CTE implementation builds upon

### Main Implementation

- **Implements** **`read_knowledge_edges_recursive`**: A new method that executes a recursive CTE for entity edge traversal
- **Unified edge view**: Creates a UNION ALL over the four edge table/direction combinations:
    - `entity_has_left_entity` (outgoing/incoming)
    - `entity_has_right_entity` (outgoing/incoming)
- **Base case**: Starting entities from `EntityEdgeTraversalData`
- **Recursive case**: Follows edges based on configuration arrays (`edge_kinds`, `edge_directions`), with temporal metadata joins
- **Depth-limited application**: Currently only uses CTE for traversals with ≤2 edges (1 hop) due to UNION ALL recursion overhead

### Code Quality Improvements

- **Adds** **`Debug`** **derives**: Added to `SharedEdgeTraversal`, `KnowledgeEdgeTraversal`, `EdgeHopMetadata`, and `EntityTraversalResult` for better debugging
- **Adds** **`Default`** **derive**: Added to `EntityTraversalResult` for cleaner initialization in the streaming parser
- **Streaming result parsing**: Refactored result processing to use `try_fold` instead of collecting all results into memory first
    - More memory efficient for large result sets
    - Better error handling with immediate propagation
    - Cleaner code structure with the fold accumulator pattern

### Query Structure

The CTE implementation consists of:

1. **`edge_view`** **CTE**: Simple UNION ALL over edge tables (pure topology, no temporal joins yet)
2. **`traversal`** **CTE**: Recursive traversal with:
    - Base case: Input entities from traversal data
    - Recursive case: Join `edge_view`, filter by edge configuration at each depth, add temporal metadata joins
    - Depth limit to prevent array out-of-bounds access
3. **Result projection**: Returns all traversed edges grouped by depth with source entity tracking

### Temporal Handling

- Joins temporal metadata for both source and target entities
- Computes traversal interval intersections: `traversal_interval * target.variable_axis`
- Filters out empty intervals to prevent invalid traversal paths
- Tracks both edge intervals and traversal intervals separately

This PR:

The changes in this PR:

The changes in this PR:

### Performance Limitations

The current implementation using UNION ALL has measurable overhead:

- **UNION ALL evaluation**: Postgres must scan all four edge table branches in the view
- **Complex join patterns**: Query planning overhead is significant
- **Suboptimal index usage**: Indexes are less effective across the unified view
- **Recursion amplification**: With deep traversals (2+ hops), the UNION ALL is evaluated multiple times during recursion, causing 2x-3x performance regression

**Current mitigation**: Only apply CTE to shallow traversals (≤2 edges). Deeper traversals use the sequential fallback.

### Edge Configuration Encoding

Currently uses `TEXT[]` arrays for edge kinds and directions:

```rust
let edge_kinds = vec!["has_left", "has_right"];
let edge_directions = vec!["outgoing", "incoming"];
```

This works but has overhead:

- Text comparison is slower than integer comparison
- Array memory overhead is higher
- No type safety at the database level

**Future improvement**: Use INTEGER enums for better performance.

## 🐾 Next steps

1. **Implement denormalized** **`entity_edges`** **table** (the optimal solution):
    - Single table with explicit direction columns
    - Eliminates UNION ALL entirely
    - Proper indexes on `(web_id, entity_uuid, edge_kind, edge_direction)`
    - Expected: 3-5x performance improvement for deep traversals
    - Enables CTE usage for all traversal depths
2. **Replace TEXT with INTEGER enums** for edge configuration:
    - Define edge kind enum: `0=has_left, 1=has_right`
    - Define direction enum: `0=outgoing, 1=incoming`
    - Use `INTEGER[]` arrays instead of `TEXT[]`
    - Expected: 10-20% reduction in query planning overhead

## 🛡 What tests cover this?

- Existing entity traversal tests now exercise the CTE code path for shallow traversals
- Integration tests in the `read_scaling` benchmark cover traversals
- The CTE returns `None` for unsupported configurations, falling back to sequential queries (which have comprehensive test coverage)

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
@@ -81,6 +81,7 @@ impl EntityEdgeTraversalData {
 }
 
 /// The result of an entity-to-ontology edge traversal.
+#[derive(Debug)]
 pub struct SharedEdgeTraversal {
     pub left_endpoint: EntityVertexId,
     pub right_endpoint: EntityTypeVertexId,
@@ -88,6 +89,7 @@ pub struct SharedEdgeTraversal {
     pub traversal_interval: RightBoundedTemporalInterval<VariableAxis>,
 }
 
+#[derive(Debug)]
 pub struct KnowledgeEdgeTraversal {
     pub left_endpoint: EntityVertexId,
     pub right_endpoint: EntityVertexId,
@@ -97,6 +99,7 @@ pub struct KnowledgeEdgeTraversal {
 }
 
 /// Metadata for edges traversed in a single hop during entity traversal.
+#[derive(Debug)]
 pub struct EdgeHopMetadata {
     pub edge_kind: KnowledgeGraphEdgeKind,
     pub edge_direction: EdgeDirection,
@@ -104,6 +107,7 @@ pub struct EdgeHopMetadata {
 }
 
 /// Result of traversing entity edges, either via CTE or sequential queries.
+#[derive(Debug, Default)]
 pub struct EntityTraversalResult {
     /// All entity edition IDs encountered during traversal (for permission filtering)
     pub entity_edition_ids: Vec<EntityEditionId>,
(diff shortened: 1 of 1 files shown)


TimDiekmann:
* **#7932** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7932?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7928** <a href="https://app.graphite.dev/github/pr/hashintel/hash/7928?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/hashintel/hash/7928?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
...
+                    EntityTraversalEdge::HasLeftEntity { direction } => ("has_left", direction),
+                    EntityTraversalEdge::HasRightEntity { direction } => ("has_right", direction),
+                };
+                let direction_str = match direction {
+                    EdgeDirection::Outgoing => "outgoing",
+                    EdgeDirection::Incoming => "incoming",
+                };
+                (kind, direction_str)
+            })
+            .collect::<(Vec<_>, Vec<_>)>();
```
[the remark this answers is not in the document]
TimDiekmann: Yes, and we still need to use `(Vec<_>, Vec<_>)` as `unzip` is generic over the return. Actually, I much prefer `collect` over `unzip` since this is possible with tuples.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
...
+            .await
+            .change_context(QueryError)?
+            .try_fold(
+                EntityTraversalResult::default(),
+                |mut traversal_result, row| async move {
+                    #[expect(
+                        clippy::cast_sign_loss,
+                        reason = "depth is always positive in the query"
+                    )]
+                    let depth = row.get::<_, i32>(9) as usize - 1; // depth starts at 1, arrays at 0
```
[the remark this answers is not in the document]
TimDiekmann: There literally is a comment why this will never returns `0`

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
...
+                            ),
+                        };
+
+                        let edge_direction = match row.get(11) {
+                            "outgoing" => EdgeDirection::Outgoing,
+                            "incoming" => EdgeDirection::Incoming,
+                            _ => unreachable!(
+                                "edge_direction guaranteed by query to be outgoing or incoming"
+                            ),
+                        };
```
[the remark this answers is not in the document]
TimDiekmann: We extract them once due to the `while` statement. For each depth the row is extracted exactly once.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
 
     /// Attempts to traverse multiple entity edges using a recursive CTE query.
     ///
-    /// This function is designed to replace the N+1 query pattern where
-    /// [`Self::read_knowledge_edges`] is called sequentially for each edge. When implemented, it
-    /// will execute a single PostgreSQL recursive CTE that traverses all edges at once, performing
-    /// automatic interval merging and deduplication using PostgreSQL's multirange types.
+    /// This function replaces the N+1 query pattern where [`Self::read_knowledge_edges`] is
+    /// called sequentially for each edge. It executes a single PostgreSQL recursive CTE that
+    /// traverses all edges at once.
```
indietyp, line 442:
```suggestion
    /// Executes a recursive Common Table Expression (CTE) that follows entity edges across multiple
    /// hops in a single database query, instead of being `N + 1` ([`Self::read_knowledge_edges`]).
    /// The CTE starts from the provided entities and recursively follows edges based on the specified
    /// edge configuration (kind and direction for each hop).
```

I don't like the wording here; this sounds like "this replaces", which is very LLM-y. Maybe something like this instead?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
...
-        // 1. Build edge configuration arrays (edge_types, edge_directions)
-        // 2. Execute recursive CTE with:
-        //    - Base case: Starting entities from traversal_data
-        //    - Recursive case: UNION ALL over 4 edge combinations (left/right, incoming/outgoing)
-        //    - Deduplication: GROUP BY (entity, depth) with range_agg() for interval merging
-        // 3. Return all depths, group edges by hop
-
-        // For now, return None to use sequential fallback
-        Ok(None)
+        // TODO: Use INTEGER enums instead of TEXT for better performance (follow-up)
```
indietyp, line 489: Would it make sense to link to the issue here?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
```diff
@@ -433,35 +437,334 @@ where
...
+            return Ok(Some(EntityTraversalResult {
+                entity_edition_ids: Vec::new(),
+                edge_hops: Vec::new(),
+            }));
+        }
+
+        // TEMPORARY: Only use CTE for shallow traversals (1 hop = 2 edges)
+        // Deep traversals (2+ hops) show 2x-3x performance regression with CTE due to
+        // UNION ALL overhead being evaluated multiple times in recursion.
+        // TODO: Implement Option 6 (single edge table) to eliminate UNION ALL
```
indietyp, line 481: reference _where_ Option 6 is located (a link to the linear issue maybe?)

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: Whops, forgot to actually add the follow-up issue. But it's solved in the immediate follow-up PR anyway.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
TimDiekmann: It's located in [slack](https://hashintel.slack.com/archives/C03F7V6DU9M/p1760974707413359), but this is also removed in the follow-up.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/read.rs
indietyp: resolved upstack

TimDiekmann: Was just about to say that I changed the wording a bit in https://app.graphite.dev/github/pr/hashintel/hash/7932/BE-173-Create-%60entity_edge%60-table-to-replace-dynamic-view-creation-in-entity-traversal#comment-PRRC_kwDOC75-is6Te5qk :D

PR #8012, opened 2025-11-04
H-5557: Add HASH logotype to Storybook sidebar

lunelson:

This PR add some configuration code and static assets to the storybook configuration in `@hashintel/ds-components`, so that HASH branding is shown at the top of the sidebar instead of storybook's own default logo.

## 🔍 What does this change?

See Demo section

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## ⚠️ Known issues

By customizing the theme in this way, we are fixing the theme to "light" mode, where Storybook's default theme contains some additional code to auto-switch, to match the user's color scheme. However, since the design system is AFAIK so far light-mode only, this might be OK for the moment

## 🐾 Next steps

Code could be added to enable dynamic switching between the dark and light variants of this custom theme, if desired

## 📹 Demo

<div style="display:flex;gap:1rem">
CleanShot 2025-11-05 at 10 40 37@2x
CleanShot 2025-11-05 at 10 40 54@2x
</div>

### DIFF
--- a/libs/@hashintel/ds-components/.storybook/utils.ts
+++ b/libs/@hashintel/ds-components/.storybook/utils.ts
@@ -14,7 +14,7 @@ const { window: globalWindow } = global;
  */
 export const getPreferredColorScheme = () => {
   // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-  if (!globalWindow || !globalWindow.matchMedia) {
+  if (!globalWindow?.matchMedia) {
     return "light";
   }
 


libs/@hashintel/ds-components/.storybook/themes.ts
```diff
@@ -0,0 +1,20 @@
...
+    brandImage: '/hash_logo_black.svg',
+    brandTarget: '_self',
+  }),
+  dark: create({
+    ...themBase,
+    base: 'dark',
+    brandImage: '/hash_logo_white.svg',
+    brandTarget: '_self',
+  }),
+};
```
kube: Nice!

libs/@hashintel/ds-components/.storybook/static/hash_logo_white.svg
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="1577px" height="358px" viewBox="0 0 1577 358" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 63.1 (92452) - https://sketch.com -->
+    <title>hash_logo_white</title>
+    <desc>Created with Sketch.</desc>
```
kube: Maybe we can remove the Sketch metadata here?

libs/@hashintel/ds-components/.storybook/manager.ts
```diff
@@ -0,0 +1,6 @@
+import { addons } from 'storybook/manager-api';
+import { themes } from './themes';
+
+addons.setConfig({
+  theme: themes.light,
+});
```
kube, line 8: Indeed we don't need dark theme at the moment, given we don't handle the dark/light scheme properly yet.

But just out of curiosity what would it imply to add this toggle?

kube (dismissed):
I think you need to setup the formatter (Biome) properly.
(I still haven't properly configured it myself, I end up with issues on every commit)

Just run `yarn fix:format` manually and same for ESLint check.

lunelson:
@kube thanks, I spotted that afterward about the lint and format steps; seems I don't have feedback in VSCode configured correctly yet either

libs/@hashintel/ds-components/.storybook/static/hash_logo_white.svg
vilkinsons: Given we haven't had a Sketch subscription for something like 6 years it speaks to the age of that asset 😅 This would be good ^

libs/@hashintel/ds-components/.storybook/manager.ts
vilkinsons: I believe this was added as a result of my description in H-5557

vilkinsons:
Good work; thanks, Lu! This can be picked tomorrow to get it over the line. If you need help with linting/formatting please post in `#help` in the Slack.

libs/@hashintel/ds-components/.storybook/themes.ts
```diff
@@ -0,0 +1,20 @@
+import { create } from "storybook/theming";
+
+const themBase = {
```
kube:
```suggestion
const themeBase = {
```

libs/@hashintel/ds-components/.storybook/manager.ts
lunelson: I've gone ahead and added the code to make this auto-switch, same way storybook's own theme does it

libs/@hashintel/ds-components/.storybook/static/hash_logo_white.svg
lunelson: sketch cruft removed ✅

libs/@local/effect-dns/hickory/tests/dummy.test.ts
lunelson, line 1: @CiaranMn I'm not sure why this file is part of the diff. A result of running `fix:format` from the root maybe?

lunelson: Update: I ran `yarn fix` at the root and the eslint job is failing within this one package... so perhaps something is off in my environment? 

cc @TimDiekmann

lunelson: I've restored this file to what it was on `main` to see if lint passes now

libs/@hashintel/ds-components/.storybook/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Uses the CSS media query `prefers-color-scheme: dark` to detect if the user
+ * has configured their system to prefer dark themes. Falls back to 'light'
+ * if the media query API is not available or if the global window object
+ * is not accessible.
+ *
+ * @returns The preferred color scheme, either 'dark' or 'light'
+ */
+export const getPreferredColorScheme = () => {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!globalWindow || !globalWindow.matchMedia) {
```
CiaranMn: Non-blocking: preference for this syntax where appropriate, but not a big deal. Becomes more important if you're going deeper into an object.

```suggestion
  if (!globalWindow?.matchMedia) {
```

CiaranMn (dismissed):
Thanks @lunelson! One nit which you can take or leave as you prefer.

lunelson:
Thanks for the review @CiaranMn, I accepted your change, but I guess this revokes the approval 😐 so just need one more 😬

PR #8060, opened 2025-11-22
BE-208: HashQL: Introduce graph data structures

indietyp:
## 🌟 What is the purpose of this PR?

Add graph data structures and algorithms to the core library, including directed graph traits, a linked graph implementation, and traversal algorithms.

These graph structures (and fundamental algorithms - some will be added later) are required for many of the MIR problems, where the CFG (and other structures) revolve around graphs.

## 🔍 What does this change?

- Adds a comprehensive graph module with core traits for directed graphs
- Implements `LinkedGraph`, an efficient adjacency list implementation using intrusive linked lists
- Adds graph traversal algorithms including depth-first and breadth-first search
- Implements Tarjan's algorithm for finding strongly connected components
- Updates `fast_hash_set_in` to accept an optional capacity parameter

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Comprehensive test suite for Tarjan's algorithm
- Tests for graph traversal algorithms

PR #8051, opened 2025-11-21
BE-210: HashQL: Implement WorkQueue data structure

indietyp:
## 🌟 What is the purpose of this PR?

Adds a new `WorkQueue` data structure that prevents duplicate items from being enqueued simultaneously, making it ideal for work scheduling where fixed-point iteration is required.

## 🔍 What does this change?

- Implements a new `WorkQueue` collection that automatically deduplicates items in a FIFO queue
- Adds comprehensive documentation with usage examples
- Exposes the new collection through the module's public API

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- The implementation includes doc tests that verify the core functionality

## ❓ How to test this?

1. Checkout the branch
2. Run the doc tests to verify the `WorkQueue` functionality
3. Confirm that the queue properly handles item deduplication and FIFO ordering

PR #8062, opened 2025-11-22
BE-211: HashQL: Replace `scc` with thread-local lock abstraction for interning

indietyp:

Refactor the intern system to use a more efficient locking mechanism and improve performance by replacing `scc` with a custom lock abstraction.

This is also required inside of the MIR, to e.g. cache successors and predecessors of different graphs.

## 🔍 What does this change?

- Removes dependency on `scc` crate and replaces it with a custom lock abstraction
- Moves `similarity.rs` to `algorithms/similarity.rs` for better organization
- Implements `LocalLock` for thread-local synchronization with better performance
- Optimizes `InternMap` and `InternSet` to use raw hash map entries for faster lookups
- Adds `with_capacity` constructor to `InternMap` for pre-allocating space
- Improves documentation with more comprehensive examples

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing tests plus new tests for hash collisions and slice interning

## ❓ How to test this?

1. Run the test suite to verify the new locking mechanism works correctly
2. Verify that all imports of `did_you_mean` are updated to the new path

### DIFF
--- a/libs/@local/hashql/core/src/intern/map.rs
+++ b/libs/@local/hashql/core/src/intern/map.rs
@@ -8,7 +8,7 @@ use hashbrown::hash_map::RawEntryMut;
 
 use super::Interned;
 use crate::{
-    collections::FastHashMap,
+    collections::{FastHashMap, fast_hash_map_with_capacity},
     heap::Heap,
     id::{HasId, Id},
     sync::lock::LocalLock,
@@ -302,12 +302,10 @@ where
     /// }
     /// ```
     pub fn with_capacity(capacity: usize, heap: &'heap Heap) -> Self {
-        use crate::collections::fast_hash_map;
-
         Self {
             heap,
-            inner: LocalLock::new(fast_hash_map(capacity)),
-            lookup: LocalLock::new(fast_hash_map(capacity)),
+            inner: LocalLock::new(fast_hash_map_with_capacity(capacity)),
+            lookup: LocalLock::new(fast_hash_map_with_capacity(capacity)),
             next: AtomicU32::new(0),
         }
     }
--- a/libs/@local/hashql/core/src/intern/set.rs
+++ b/libs/@local/hashql/core/src/intern/set.rs
@@ -5,7 +5,7 @@ use hashbrown::hash_map::RawEntryMut;
 
 use super::Interned;
 use crate::{
-    collections::{FastHashMap, fast_hash_map},
+    collections::{FastHashMap, fast_hash_map_with_capacity},
     heap::Heap,
     sync::lock::LocalLock,
 };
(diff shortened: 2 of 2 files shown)


indietyp:
* **#8071** <a href="https://app.graphite.com/github/pr/hashintel/hash/8071?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8062** <a href="https://app.graphite.com/github/pr/hashintel/hash/8062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8060** <a href="https://app.graphite.com/github/pr/hashintel/hash/8060?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8051** <a href="https://app.graphite.com/github/pr/hashintel/hash/8051?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libs/@local/hashql/core/src/sync/lock.rs
```diff
@@ -0,0 +1,532 @@
...
+#[expect(unsafe_code)]
+unsafe impl<T: Send> Send for LockInner<T, Local> {}
+
+// SAFETY: The same safety invariants apply as the ones from `Mutex`
+#[expect(unsafe_code)]
+unsafe impl<T: Send> Send for LockInner<T, Shared> {}
+
+// SAFETY: The same safety invariants apply as the ones from `Mutex`
+#[expect(unsafe_code)]
+unsafe impl<T: Send> Sync for LockInner<T, Shared> {}
```
TimDiekmann, line 223: I actually had to look this up, but it's correct, it requires `T: Send` for the `Mutex` to be `Sync`

TimDiekmann (dismissed):
Good work!

Do you plan to replace `scc` in HaRPC as well btw.?

libs/@local/hashql/core/src/sync/lock.rs
indietyp: Me too when I was implementing it :sweat_smile:

TimDiekmann: That's a relief :smile:

I actually thought it would be something like `Send + Sync`​.

indietyp:
> Good work!
> 
> Do you plan to replace `scc` in HaRPC as well btw.?

Yes, once I resume HaRPC work, I plan to remove it, although arguably will be a bit harder, because the work we do there is a lot more parallized (but I think with a std mutex that doesn't go over an `await` boundary there shouldn't be much change – if any – tbh)

PR #8083, opened 2025-11-26
H-5631: Restore Petrinaut's external interface, split out dev/demo site

CiaranMn:

1. Adds props to Petrinaut, replicating the external interface of the old Petrinaut, so that the new one can be used in applications.
2. Split the demo/dev site out into a separate folder, and make the package export a React component.

Note that it looks like lots of LOC changed but this is mostly whitespace from indent changes, enable 'hide whitespace' when viewing the diff to cut down on noise.

## 🔍 What does this change?

- Reimplement Petrinaut props.
- Split out `title` and `id` from SDCPN definition (external consumers often control these)
- Switch 'update net' methods to be by mutation (for compatibility with e.g. Automerge). **Worth particularly checking the new context to make sure the new implementation looks okay.** 
- Move the SPA into a `demo-site` folder, make the package build and export a React component
- Move local storage sync to demo site
- Add a `vercel.json` to control Vercel deployment settings from the codebase
- Rename a few things in the type (to discuss, can be reverted/changed easily)
  - `SDCPNType` –> `Color` (found it a bit confusing with `SDCPN` and `SDCPNType`, but may be a better name)
  -  `iconID` –> `iconSlug` to avoid confusion with other `xId`s (which refer to other entities)
  - `colorCode` –> `displayColor` to avoid confusion with color/type
  - `type.elements.id` –> `type.elements.elementId` because there's another `id` one level up, on `type.id` (generally prefer not having bare `id`s, but can discuss)
  - `differentialEquationCode` –> `differentialEquationId` for consistency, remove unused provision for inline code string
- Add conversion between old format and new for local storage sync and file import, so that people don't lose existing work. We can remove this at some point.

This PR:

- [x] modifies an **npm**-publishable library but **I have not yet added a changeset file(s)** (to discuss when we want to publish the new version)

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing unit tests.

## ❓ How to test this?

1. Visit preview deployment

### DIFF
--- a/libs/@hashintel/petrinaut/demo-site/main/app.tsx
+++ b/libs/@hashintel/petrinaut/demo-site/main/app.tsx
@@ -23,10 +23,14 @@ export const DevApp = () => {
   }, [currentNetId, storedSDCPNs]);
 
   const existingNets: MinimalNetMetadata[] = useMemo(() => {
-    return Object.values(storedSDCPNs).map((net) => ({
-      netId: net.id,
-      title: net.title,
-    }));
+    return Object.values(storedSDCPNs)
+      .filter(
+        (net): net is SDCPNInLocalStorage => !isOldFormatInLocalStorage(net),
+      )
+      .map((net) => ({
+        netId: net.id,
+        title: net.title,
+      }));
   }, [storedSDCPNs]);
 
   const createNewNet = useCallback(
@@ -59,7 +63,7 @@ export const DevApp = () => {
 
       setStoredSDCPNs((prev) =>
         produce(prev, (draft) => {
-          if (draft[currentNetId]) {
+          if (draft[currentNetId] && "title" in draft[currentNetId]) {
             draft[currentNetId].title = title;
           }
         }),


libs/@hashintel/petrinaut/src/state/editor-store.ts
```diff
@@ -59,152 +46,99 @@ export type EditorState = {
...
-              isLeftSidebarOpen: true,
-              selectedResourceId: null,
-              selectedItemIds: new Set(),
-              draggingStateByNodeId: {},
-            },
-            false,
-            { type: "initializeEditorStore" },
-          );
-        },
-      }),
```
CiaranMn: For some reason the type checking in `create<>` isn't showing a compiler error if a function in the type isn't provided in the create call... this does catch it. Not sure what's going on there.

libs/@hashintel/petrinaut/demo-site/main/app.tsx
```diff
@@ -0,0 +1,158 @@
...
+  type SDCPNInLocalStorage,
+  useLocalStorageSDCPNs,
+} from "./app/use-local-storage-sdcpns";
+
+export const DevApp = () => {
+  const { storedSDCPNs, setStoredSDCPNs } = useLocalStorageSDCPNs();
+
+  const [currentNetId, setCurrentNetId] = useState<string | null>(null);
+
+  const currentNet = useMemo(() => {
```
kube, line 18: We have React Compiler set up, so no need for `useMemo` or `useCallback`.

kube: That's a detail, no need to change that now, it won't break anything to have it.

libs/@hashintel/petrinaut/src/views/Editor/editor-view.tsx
```diff
@@ -148,51 +155,57 @@ export const EditorView: React.FC = () => {
...
+                   *       transition depending on which was randomly selected.
+                   */
+                  // {
+                  //   id: "load-example-supply-chain",
+                  //   label: "Supply Chain",
+                  //   onClick: () => {
+                  //     createNewNet(supplyChainSDCPN);
+                  //     clearSelection();
+                  //   },
+                  // },
```
kube, line 171: What I understood from the probabilistic Transition Kernel discussion with Jade:
- We don't define a probability between two places
- We define a probability distribution of the new token value inside this place.

So my ticket about Probabilistic Transition Kernel wouldn't help tackle a probabilistic split between Good Product and Bad Product.

kube: And just after posting I understood that you want to add an intermediate place, and do the Dispatch after that, which absolutely makes sense.

CiaranMn: Yes, exactly – and the reason I commented this out for now is that at the moment it will put a token in both the 'Hospital' and 'Disposal' token after each manufacture, which doesn't make sense, so better to leave it out for now

libs/@hashintel/petrinaut/package.json
```diff
@@ -10,14 +10,17 @@
   "license": "(MIT OR Apache-2.0)",
   "sideEffects": false,
   "type": "module",
+  "main": "src/main.ts",
+  "types": "src/main.ts",
   "scripts": {
-    "build": "vite build",
-    "dev": "vite",
+    "build": "rimraf dist && tsc -p tsconfig.build.json",
```
kube: I guess in the future it would be better to build using Vite?

libs/@hashintel/petrinaut/demo-site/main/app/use-local-storage-sdcpns.ts
```diff
@@ -1,12 +1,14 @@
 import { useLocalStorage } from "@mantine/hooks";
 
-import type { SDCPN } from "../core/types/sdcpn";
+import type { SDCPN } from "../../../src/core/types/sdcpn";
 
 const rootLocalStorageKey = "petrinaut-sdcpn";
 
-type SDCPNInLocalStorage = {
+export type SDCPNInLocalStorage = {
+  id: string;
```
kube, line 12: I understand why you want to put `id` here rather than in the SDCPN definition itself, but we're gonna need at some point to identify the SDCPN used for a given Simulation, to get the list of simulations done for one SDCPN, not necessarily with the latest version. (or simulations done in previous versions would not be accessible anymore)

kube: Not necessary now, just a thought for the future.

libs/@hashintel/petrinaut/src/core/types/sdcpn.ts
```diff
@@ -31,13 +29,13 @@ export type Place = {
   height?: number;
 };
 
-export type SDCPNType = {
+export type Color = {
```
kube, line 32: I agree that it's better to use `Color` rather than `Type` here, but maybe keeping the prefix would help understanding that this does not represent an actual screen color.

```suggestion
export type SDCPNColor = {
```

Again just a remark.

kube (approved):
Good job at being able to iterate over the current state of the code. :)

Just some comments, but more for future discussions.

libs/@hashintel/petrinaut/demo-site/main/app/use-local-storage-sdcpns.ts
CiaranMn: Agreed, future update: simulation state will need to store the id as well as the definition (since the id is no longer in the definition)

libs/@hashintel/petrinaut/demo-site/main/app.tsx
CiaranMn: Cool, we'll review the whole package for opportunities to remove these in a future refactor

libs/@hashintel/petrinaut/src/core/types/sdcpn.ts
CiaranMn: We need to figure out a good name for this – since Petrinaut will handle other types of PNs, we want to avoid calling everything 'SDCPNx'. We agreed to come back to this naming later.

libs/@hashintel/petrinaut/package.json
CiaranMn: We talked about this – `vite` (or some other bundling step) is needed to do the PandaCSS compilation. I'll add it back in

libs/@hashintel/petrinaut/demo-site/main/app.tsx
lunelson: I'm generally in favour of not using memoization hooks; but FTR the React Compiler still doesn't handle every possible case, so checking the rendering is still something we are responsible for, and it's still valid to use these hooks 😐 explicitly

lunelson (approved):

libs/@hashintel/petrinaut/tsconfig.json
```diff
@@ -1,12 +1,12 @@
...
-    "jsx": "preserve",
+    "jsx": "react-jsx",
     "lib": ["dom", "dom.iterable", "ESNext"],
-    "rootDir": "src/",
     "moduleResolution": "bundler",
     "noEmit": true,
     "skipLibCheck": true
   },
-  "include": ["./src/**/*"]
+  "include": ["src", "demo-site", "vite.config.ts"],
```
kube: You should not need `vite.config.ts` here, because it is "meta" to the project, and was added to ESLint configuration as an exception.

CiaranMn: My editor was giving some warning about this but now it isn't, removed it 🤷

kube: Yeah it seems ESLint breaks easily.

kube (approved):

PR #8137, opened 2025-12-01
H-5651: Extract Refractive library and provide minimal API

kube:

Extract the Refractive effects from Design System as a standalone library, and provide a simpler API, with minimal surface, to let us update the internals progressively.

## 🔍 What does this change?

- New package: `@hashintel/refractive`
- Compiled with React Compiler
- Remove `<Filter />`, and keep only `<FlexibleFilter />` (renamed to Filter)
- Remove Motion entirely at the moment
- Remove superfluous code
- Only one export, an Higher-Order Component API called `refractive`:
  ```tsx
  import { refractive } from "@hashintel/refractive"
  
  const RefractiveButton = refractive(Button)
  
  // This adds a new `refraction` prop, to customize borderRadius and refraction.
  <RefractiveButton
    className="my-classname"
    refraction={{ radius: 20, bezelWidth: 22 }}
  />

  // To directly use JSX intrinsic elements:
  <refractive.div
    className="my-classname"
    refraction={{ radius: 20, bezelWidth: 22 }}
  />
  ```

- Migrate `@hashintel/ds-components` to use this

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## 🐾 Next steps

- FE-43: Try to remove entirely usage of ResizeObserver, which requires JS to render/update the filter dimensions properly.

## 🛡 What tests cover this?

None at the moment.

### DIFF
--- a/apps/hash-api/package.json
+++ b/apps/hash-api/package.json
@@ -78,7 +78,7 @@
     "dedent": "1.7.0",
     "effect": "3.18.4",
     "exponential-backoff": "3.1.3",
-    "express": "5.1.0",
+    "express": "5.2.0",
     "express-handlebars": "8.0.4",
     "express-rate-limit": "8.2.1",
     "fractional-indexing": "3.2.0",
--- a/libs/@local/hash-isomorphic-utils/package.json
+++ b/libs/@local/hash-isomorphic-utils/package.json
@@ -34,7 +34,7 @@
     "@sentry/browser": "10.22.0",
     "@sindresorhus/slugify": "3.0.0",
     "big.js": "7.0.1",
-    "express": "5.1.0",
+    "express": "5.2.0",
     "fix-esm-import-path": "1.10.3",
     "fractional-indexing": "3.2.0",
     "graphql-tag": "2.12.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -512,7 +512,7 @@ __metadata:
     effect: "npm:3.18.4"
     eslint: "npm:9.38.0"
     exponential-backoff: "npm:3.1.3"
-    express: "npm:5.1.0"
+    express: "npm:5.2.0"
     express-handlebars: "npm:8.0.4"
     express-rate-limit: "npm:8.2.1"
     fractional-indexing: "npm:3.2.0"
@@ -10249,7 +10249,7 @@ __metadata:
     "@vitest/coverage-istanbul": "npm:3.2.4"
     big.js: "npm:7.0.1"
     eslint: "npm:9.38.0"
-    express: "npm:5.1.0"
+    express: "npm:5.2.0"
     fix-esm-import-path: "npm:1.10.3"
     fractional-indexing: "npm:3.2.0"
     graphql: "npm:16.11.0"
@@ -23239,7 +23239,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"body-parser@npm:2.2.1, body-parser@npm:^2.2.0":
+"body-parser@npm:2.2.1, body-parser@npm:^2.2.0, body-parser@npm:^2.2.1":
   version: 2.2.1
   resolution: "body-parser@npm:2.2.1"
   dependencies:
@@ -28426,17 +28426,18 @@ __metadata:
   languageName: node
   linkType: hard
 
-"express@npm:5.1.0, express@npm:^5.0.1":
-  version: 5.1.0
-  resolution: "express@npm:5.1.0"
+"express@npm:5.2.0, express@npm:^5.0.1":
+  version: 5.2.0
+  resolution: "express@npm:5.2.0"
   dependencies:
     accepts: "npm:^2.0.0"
-    body-parser: "npm:^2.2.0"
+    body-parser: "npm:^2.2.1"
     content-disposition: "npm:^1.0.0"
     content-type: "npm:^1.0.5"
     cookie: "npm:^0.7.1"
     cookie-signature: "npm:^1.2.1"
     debug: "npm:^4.4.0"
+    depd: "npm:^2.0.0"
     encodeurl: "npm:^2.0.0"
     escape-html: "npm:^1.0.3"
     etag: "npm:^1.8.1"
@@ -28457,7 +28458,7 @@ __metadata:
     statuses: "npm:^2.0.1"
     type-is: "npm:^2.0.1"
     vary: "npm:^1.1.2"
-  checksum: 10c0/80ce7c53c5f56887d759b94c3f2283e2e51066c98d4b72a4cc1338e832b77f1e54f30d0239cc10815a0f849bdb753e6a284d2fa48d4ab56faf9c501f55d751d6
+  checksum: 10c0/be184dbd5eadffd899811bcd91b22df96d968b29e31076e8bbafcaa5c25dafd0012b8daece4135b876e4576407e4ca811c089ead2bb45f9e08df812ba0966a82
   languageName: node
   linkType: hard
 


libs/@hashintel/refractive/src/components/composite-parts.tsx
```diff
@@ -0,0 +1,272 @@
...
+  right: string;
+  bottomLeft: string;
+  bottom: string;
+  bottomRight: string;
+};
+
+//
+// Splits an ImageData into 8 parts and returns them as Base64-encoded PNG images.
+// The parts can then be rendered and composited together to form a scalable image with correct corners and edges.
+//
```
CiaranMn: This looks like it's supposed to be a docstring on the function, is it? If so should be in JSDoc format to be picked up properly

libs/@hashintel/refractive/src/components/composite-parts.tsx
```diff
@@ -0,0 +1,272 @@
...
+    bottomLeft,
+    bottom,
+    bottomRight,
+  };
+}
+
+//
+// Component that renders the 8 parts of an image and composites them together.
+// Used internally by the Filter component, for DisplacementMap and SpecularMap.
+//
```
CiaranMn: Not sure this comment is needed as it's duplicated by the docstring below

libs/@hashintel/refractive/src/hoc/refractive.tsx
```diff
@@ -0,0 +1,116 @@
...
+          radius={refraction.radius ?? 4}
+          glassThickness={refraction.glassThickness ?? 70}
+          bezelWidth={refraction.bezelWidth ?? 0}
+          refractiveIndex={refraction.refractiveIndex ?? 1.5}
+          specularOpacity={refraction.specularOpacity ?? 0}
+          specularAngle={refraction.specularAngle ?? 0}
+          bezelHeightFn={refraction.bezelHeightFn ?? CONVEX}
+        />
+
+        {/* @ts-ignore Need to fix types in this file */}
```
CiaranMn: Always prefer `@ts-expect-error`, never `@ts-ignore`, because then we get linting errors once there's no longer an error (if we just ignore it, we don't)
```suggestion
        {/* @ts-expect-error Need to fix types in this file */}
```

libs/@hashintel/refractive/src/hoc/refractive.tsx
```diff
@@ -0,0 +1,116 @@
...
+};
+
+type RefractiveFunction = (<P extends object>(
+  Component: ComponentType<P>,
+) => ComponentType<P & RefractionProps>) &
+  HTMLElements;
+
+const CACHE = new Map<string, ComponentType<any>>();
+
+export const refractive = new Proxy(createRefractiveComponent as any, {
```
CiaranMn: As this is the single export from the library, can we add a JSDoc comment please explaining how to use it?

libs/@hashintel/refractive/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "dist",
+    "CHANGELOG.md",
+    "LICENSE-APACHE.md",
+    "LICENSE-MIT.md",
+    "LICENSE",
+    "README.md"
+  ],
+  "scripts": {
+    "build": "vite build",
+    "clean": "rm -rf dist",
```
CiaranMn: Can you use `rimraf dist` instead (will need adding to `devDependencies`)? So it works in non-*nix environments.

And add this command to build as it's currently unused a, i.e. `"build": "clean && vite build"`. We could do the same for `ds-components` at the same time

libs/@hashintel/refractive/src/components/composite-parts.tsx
kube: Yes sorry, this was moved without further cleaning.
Was to split sections when everything was in only one file.
I'm updating it.

libs/@hashintel/refractive/package.json
kube: I'm gonna add `rimraf`, but if we use `&&` in command, we will break Windows compatibility.

CiaranMn: Pretty sure `&&` works in Windows (`;` doesn't)

kube: Oh yeah indeed.
It was not working in Powershell, but seems that it was fixed in 2019, which says a lot about my usage of Windows for programming.

CiaranMn: In any case I think `vite` will empty the out dir if you specify it as part of the config? So this command may be unnecessary

kube: It cleans it already yes, so no need to add it to it to `build` indeed.

kube: I just removed it, and in `@hashintel/ds-components` too.

We don't have this command anywhere else in the repo, except in `@blockprotocol/type-system`, where it is used in `build` command.

I guess nobody uses `yarn clean` commands.
This was just Copilot doing package.json as if it was doing a Makefile.

libs/@hashintel/refractive/src/hoc/refractive.tsx
kube: 👍 Sorry about that.

libs/@hashintel/refractive/src/hoc/refractive.tsx
```diff
@@ -0,0 +1,160 @@
...
+          }
+        }
+      });
+
+      resizeObserver.observe(element);
+
+      return () => {
+        resizeObserver.disconnect();
+      };
+    }, []);
```
lunelson: setting state inside a resize handler hammers React by forcing everything here to re-render many times per second. It would be better to use a ref for both the element and the filter, and use event listeners to directly update the size of one based on the size of the other, not go through React state

libs/@hashintel/refractive/src/hoc/refractive.tsx
```diff
@@ -0,0 +1,160 @@
...
+          refractiveIndex={refraction.refractiveIndex ?? 1.5}
+          specularOpacity={refraction.specularOpacity ?? 0}
+          specularAngle={refraction.specularAngle ?? 0}
+          bezelHeightFn={refraction.bezelHeightFn ?? CONVEX}
+        />
+
+        {/* @ts-expect-error Need to fix types in this file */}
+        <Component
+          {...componentProps}
+          ref={elementRef}
```
lunelson, line 99: the graphite comment is correct I think, that you need to handle ref forwarding

libs/@hashintel/refractive/src/hoc/refractive.tsx
```diff
@@ -0,0 +1,160 @@
...
+      ({ children, ...props }: any) =>
+        createElement(elementName, props, children),
+    );
+    CACHE.set(elementName, refractiveComponent);
+    return refractiveComponent;
+  },
+  apply: (target, _thisArg, argArray) => {
+    return target(...argArray);
+  },
+}) as RefractiveFunction;
```
lunelson, line 176: This doesn't seem right to use an `as` here. Is there not a generic parameter on `Proxy`?

libs/@hashintel/refractive/src/hoc/refractive.tsx
kube: Yes, forgot to write that in PR description.
I still need to forward the ref, but need to do it correctly relative to typings, and React 19 compatibility.

libs/@hashintel/refractive/src/hoc/refractive.tsx
kube: For now I'd like to keep it as is, given this is a simple extraction, and removal of Motion.

I hesitated to either add a Signal/Observable library, or to pass this as a subscribe function down to `<CompositeParts>` (which will be the component where we will need to update the DOM).

But given FE-43 is coming soon, and it will possibly remove the need for `ResizeObserver`, I'll do that at this moment rather than optimizing this part.

For now the performance is acceptable.

libs/@hashintel/refractive/src/hoc/refractive.tsx
kube: Possibly we will remove the need for `Proxy`, but Proxy is hard to type, given its Generic parameter is the one of the first parameter.

Here we modify the output, so we need to cast.
Not a really typesafe way, but will try to find a better one later.

The goal here was to provide a first minimal version of the API, and update internals in iterations.

libs/@hashintel/refractive/src/hoc/refractive.tsx
lunelson: My point is you can just pass a `ref` down to the `Filter`, so you can capture the `filterElement` from that and imperatively update `height` and `width` on it in this resize handler, rather than updating state and having Filter update itself from the state. React state shouldn't be used for this kind of update loop

kube: It's more complex than that.

Every part needs to be updated (CompositeParts), and for multiple layers (RefractionMap and Specular).

So you would need multiple refs, and it would become a mess.

kube: Here is the file in question: https://github.com/hashintel/hash/blob/cf/h-5651-extract-refractive-library-and-publish-to-npm/libs/%40hashintel/refractive/src/components/composite-parts.tsx

Most of `<feImage>` elements need to be updated with positions derived from object dimensions.

kube: So just to be clear: I know that using `useState` is a bad solution.

That's why Motion was used before, but the goal here was to remove as much dependencies as possible, have a simple implementation with minimal interface, and iterate on performance later.

`useState` won't be kept here for long, but implementing a clean and performant solution to directly remove it on the next ticket is a waste of time.

lunelson: So what are you going to replace it with?

kube: FE-43

kube: I updated the ticket to explicitly say we will optimize current solution if `objectBoundingBox` solution is not successful.

kube: And just to reassure you on the fact that this is not going to stay:

https://linear.app/hash/issue/H-5413/refractive-optimize-usage-of-motion-or-find-alternative-solution

This ticket is part of Refractive milestone too, but `objectBoundingBox` solution would just remove the need for it. (at least for calculation of position for filter parts)

PR #8180, opened 2025-12-12
H-5837: Add temp/scratch/holding directory ignore patterns (ad-hoc DX)

lunelson:

This PR makes small additions to the root `.gitignore` and the `.dockerignore` to allow: 

1. placement of directories matching `_temp/`, `_trash/`, `_scratch/` or `_holding/` anywhere in the file tree to have their contents ignored
2. use of Jujutsu VCS' `.jj/` directory

...without causing false inclusions in other tools.

Background: most repo tools have the behaviour of ignoring whatever is ignored in `.gitignore`, but some do not successfully ignore patterns like `my-directory/.gitignore` where the nested ignore file just contains `*` (which is what Jujutsu does in the `.jj` directory, and what was suggested below in the comments for `_temp` etc. directories).

This is just a DX improvement.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -177,3 +177,6 @@ out/
 **/_trash/
 **/_scratch/
 **/_holding/
+
+# Jujutsu
+.jj/
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,6 @@ out/
 **/_trash/
 **/_scratch/
 **/_holding/
+
+# Jujutsu
+.jj/


lunelson:
@hashintel/devops ...could this change be causing the terraform action to fail 🤔 😬 ?

TimDiekmann:
No worries, the issue with Terraform is that it tries to downgrade Postgres from 18 to 17 again (A change I did today morning), but we don't have auto-apply enabled currently. But better flagging that than having a silent infra-issue 😅

lunelson:
> Put a .gitignore with * in it in these folders

@TimDiekmann Oh 🤯 huh I never thought of that, nice one!

I do have these configured as a global ignore on my system... but the issue I'm finding is that LLMs which try to "follow .gitignore rules", don't pick up the global config and then are scanning those folders for context and `@-`references etc.

TimDiekmann:
I thought that LLMs (or at least the tool around them) would use commands which actually respect the `.gitignore`. such as `rg` instead of `grep`, or `git grep`, but I might be wrong. But at least I've seen the LLM itself trying to use `find` and explicitly excluding folders such as `node_modules/` or `target/` (it should know how to use the correct tools, though).

lunelson:
@TimDiekmann they do tend to correctly see the local .gitignore file they just don't seem to look at whatever you might have globally configured—and IIRC having a file at `~/.gitignore` isn't actually a standard convention, i.e. it won't take effect automatically you have to declare global ignore file on your system (and technically you could name it anything AFAIK?)

PR #8181, opened 2025-12-12
H-5746: Build NER workflow in Mastra-first way

lunelson:

This PR implements the "extraction" phase of an NER workflow in an idiomatic Mastra way, aiming to establish some essential simplifications over the previous patterns used in `"@apps/hash-integration-worker"`

The moving parts include 
- an `nerAgent`
- an `nerPeopleWorkflow` including an `nerPeopleStep`
- an `nerScorer` for evals
- and some (newly) dereferenced entity type schema fixtures...

NOTE: the entity-type schemas have been generated in this branch using `"@apidevtools/json-schema-ref-parser"` and manipulated with `"@sinclair/typebox"` before passing in to the inference call's `structuredOutput` prop.

The inputs to this workflow are simply a source-text and a research-goal. The interest in _people_ as entities is assumed (it could be inferred in a precedent step, from the research goal). See the [demo](#demo) section below for what this looks like

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

The following are notes on what the next key pieces would be for a more robust/realistic NER workflow

### (A) provide entity types dynamically (or infer, based on goal)
    
Right now the `nerPeopleWorkflow` assumes the entity types that are of interest; obviously this should be dynamic, either in the sense that the user can choose from desired entity types somehow and those are passed through, or the goal text is taken as a basis for inferring with an LLM, what they should be. Possibly the latter could be a fallback for the former.
    
In either case, a more flexible dynamic approach here might be best if the desired types were grouped in their URL form with TypeBox, and this was then dynamically fetched and dereferenced AND compiled to TS types. Specifically the latter step would allow inlining the TS types in the prompt as an additional reinforcement, and it is probably the most semantically rich way of doing this compactly

### (B) implement chunked looping extraction

The current implementation is a toy, in the sense that it is only working with relatively small texts. An obvious next step would be to turn to larger texts, and to process these in chunks, in a loop.

Mastra provides primitives to: 
- chunk documents
- carry provenance via metadata
- orchestrate multi-step passes
- maintain scoped state between chunks for things like pronoun hand-offs;

...so in theory an implementation could go something like:

1. (Chunk): Create `MDocument`, chunk with sentence or structure-aware strategy, set maxSize/overlap, and emit chunks + metadata
2. (Looped NER): For each chunk, call NER agent with chunk text + provenance + prior “handoff” entity-vs-pronoun context; output entities with offsets and chunkId.
3. (Merge): Consolidate entities across chunks, resolve coreference, and produce a final entity set with aggregated provenance

### (C) implement post-extraction steps

After a successful extraction, there are still concerns of merging/de-duplicating, as well as enriching the extractions with more observations (claims).

A sketch I captured earlier about how merging might work via clustering and using some non-LLM tools as well

1. produce clusters of merge-candidates
 - `sentence-transformers` with embedding cache
 - similarity matrix + clusters
2. have a validation gate
 - return `{merge: true/false, confidence: 0.0-1.0, reason: string}`
 - only merge above threshold e.g. 0.8
3. track non-merged items as "related" entities

## 🛡 What tests cover this?

- `apps/hash-ai-agent/src/mastra/workflows/ner-people-workflow.test.ts`

## ❓ How to test this?

Test run with the normal Vitest runner

## 📹 Demo

https://github.com/user-attachments/assets/8b415cf5-8193-4284-89db-abd3fc089066

### DIFF
--- a/apps/hash-ai-agent/src/mastra/constants.ts
+++ b/apps/hash-ai-agent/src/mastra/constants.ts
@@ -4,16 +4,17 @@
  * Centralizes configuration values used across agents, scorers, and workflows.
  */
 
-import type { ModelForProvider } from '@mastra/core/llm/model';
+import type { ModelForProvider } from "@mastra/core/llm/model";
 
-export type OpenRouterModelId = `openrouter/${ModelForProvider<'openrouter'>}`;
+export type OpenRouterModelId = `openrouter/${ModelForProvider<"openrouter">}`;
 
 /**
  * Default LLM model identifier for agent inference.
  *
  * Uses OpenRouter routing to Google's Gemini 2.5 Flash Lite model.
  */
-export const DEFAULT_MODEL = 'openrouter/google/gemini-2.5-flash-lite' as const satisfies OpenRouterModelId;
+export const DEFAULT_MODEL =
+  "openrouter/google/gemini-2.5-flash-lite" as const satisfies OpenRouterModelId;
 
 /**
  * Block Protocol property type URL for entity names.
@@ -21,4 +22,5 @@ export const DEFAULT_MODEL = 'openrouter/google/gemini-2.5-flash-lite' as const 
  * Used as the canonical key for extracting/matching person and organization names
  * in structured entity output.
  */
-export const NAME_PROPERTY = 'https://blockprotocol.org/@blockprotocol/types/property-type/name/';
+export const NAME_PROPERTY =
+  "https://blockprotocol.org/@blockprotocol/types/property-type/name/";


apps/hash-ai-agent/src/mastra/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Centralizes configuration values used across agents, scorers, and workflows.
+ */
+
+/**
+ * Default LLM model identifier for agent inference.
+ *
+ * Uses OpenRouter routing to Google's Gemini 2.5 Flash Lite model.
+ */
+export const DEFAULT_MODEL = "openrouter/google/gemini-2.5-flash-lite";
```
indietyp: Should this maybe be:

```suggestion
export const DEFAULT_MODEL = "openrouter/google/gemini-2.5-flash-lite" as const satisfies ModelRouterModelId;
```

to ensure that the type is indeed satisfiable. At the end this is practically the same as having it in `new Agent`, but this makes sure that the error is localized to this specific place instead of spread out, which should make debugging in the future easier.

apps/hash-ai-agent/src/mastra/fixtures/generate-schemas.ts
```diff
@@ -0,0 +1,244 @@
+/* eslint-disable no-console */
```
indietyp, line 1: If we generate the fixtures, do we need to commit into the repo? or could be use a `codegen` step with turborepo config instead?

apps/hash-ai-agent/src/mastra/index.ts
```diff
@@ -1,29 +1,52 @@
...
+const resolveDbUrl = (): string => {
+  const envPath = process.env.MASTRA_DB_PATH;
+  if (envPath) {
+    const resolved = path.isAbsolute(envPath)
+      ? envPath
+      : path.resolve(process.cwd(), envPath);
+    return pathToFileURL(resolved).toString();
+  }
+  // Default: relative to this file (works in dev, may need adjustment post-build)
+  const defaultPath = path.resolve(metaDirname, "../../data/mastra.db");
```
indietyp: I'd rather have it that we fail loud, instead of (fail) silently here and have the application work in unexpected ways. Maybe we could use/add it to our env variables instead and require it here?

apps/hash-ai-agent/src/mastra/utils/json-schema-to-zod.quokka.ts
indietyp, line 1: Out of curiosity: What is quokka?

package.json
```diff
@@ -85,7 +78,7 @@
     "@playwright/test": "1.57.0",
     "@tldraw/editor@npm:2.0.0-alpha.12": "patch:@tldraw/editor@npm%3A2.0.0-alpha.12#~/.yarn/patches/@tldraw-editor-npm-2.0.0-alpha.12-ba59bf001c.patch",
     "@tldraw/tlschema@npm:2.0.0-alpha.12": "patch:@tldraw/tlschema@npm%3A2.0.0-alpha.12#~/.yarn/patches/@tldraw-tlschema-npm-2.0.0-alpha.12-13bf88407b.patch",
-    "@types/react": "19.2.7",
+    "@types/react": "19.1.9",
```
indietyp: This change seems a bit odd.

apps/hash-ai-agent/src/mastra/index.ts
lunelson: Yeah this was a bit of GPT 5.2 PR review-derived stuff. My original code was simpler, just pretty much what you see below. IMO we don't need to overthink it, the DB only exists locally so far anyway, and the whole thing is not yet deployed anywhere

```ts
const metaFilename = fileURLToPath(import.meta.url);
const metaDirname = path.dirname(metaFilename);
const dbFilename = path.resolve(metaDirname, "../../data/mastra.db");
```

lunelson: (The resolution is relative to the `.mastra/output/` directory)

apps/hash-ai-agent/src/mastra/utils/json-schema-to-zod.quokka.ts
lunelson: Ah yes, good catch 🙏🏼 . This is a thing I do, to work out patterns and draft things that may become tests. but I should probably de-index and ignore these files. Quokka is [this](https://quokkajs.com/)

indietyp: Ah, okay, good to know! 😁

package.json
lunelson: I agree, I don't know where that crept in, probably an LLM modification when I was modifying other resolutions 😐, will revert

apps/hash-ai-agent/src/mastra/fixtures/generate-schemas.ts
lunelson: ✅ agreed and done!

apps/hash-ai-agent/src/mastra/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Block Protocol property type URL for entity names.
+ *
+ * Used as the canonical key for extracting/matching person and organization names
+ * in structured entity output.
+ */
+export const NAME_PROPERTY =
+  "https://blockprotocol.org/@blockprotocol/types/property-type/name/" as const satisfies ModelRouterModelId;
```
indietyp: Added to the wrong item, I am pretty sure, but it also shows a critical flaw here, the type definition is:

```ts
export type ModelRouterModelId =
  | {
      [P in Provider]: `${P}/${ProviderModelsMap[P][number]}`;
    }[Provider]
  | (string & {});
```

the `string` makes it so that _any_ value would be okay, which is something we do not want (and which would also not be caught by typescript if we have a typo inside of `Agent`. We would need to have a new type:

```ts
type AvailableModels = {[P in keyof ProviderModelsMap]: `${P}/${ProviderModelsMap[P][number]}`}[keyof ProviderModelsMap]
```

and then use that (it removes the `string & {}` requirement)

lunelson: well caught 🤦🏼 ! I made the type stricter: turns out there's another utility type in the library that lets me restrict it to a certain provider, which is even better in this case 🙏🏼

PR #8202, opened 2025-12-19
H-5854: Fix disabling search indexing for non-production deployments

CiaranMn:

As title. 

Screenshot 2025-12-19 at 11 52 56

### DIFF
--- a/apps/hash-frontend/src/shared/layout/plain-layout.tsx
+++ b/apps/hash-frontend/src/shared/layout/plain-layout.tsx
@@ -1,5 +1,4 @@
 import { useTheme } from "@mui/material";
-import Head from "next/head";
 import { useRouter } from "next/router";
 import type { DefaultSeoProps } from "next-seo";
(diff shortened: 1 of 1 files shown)


apps/hash-frontend/src/shared/layout/plain-layout.tsx
```diff
@@ -49,6 +45,7 @@ export const PlainLayout: FunctionComponent<{
             href: `/favicon.png?v=${router.asPath}`, // force favicon refresh on route change
           },
         ]}
+        dangerouslySetAllPagesToNoIndex={!isProduction}
```
vilkinsons, line 48: Appreciate this has already merged, but might we want to set `dangerouslySetAllPagesToNoFollow` for non-prod deployments, too?

CiaranMn: We can do but I don't think it makes much difference – the robot can follow links to discover other pages but they're all noindexed anyway (or on other sites)

PR #8241, opened 2026-01-06
H-5764: Setup Sentry UserFeedback + ErrorTracker

kube:
## Demo

[Kapture 2026-01-07 at 20.22.59.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/4f12f7d0-0518-4716-bbc6-b2f6d30cb05b.mp4)

This PR adds Sentry User Feedback + Error Tracker inside Petrinaut demo website.

### Error Tracker

Because we don't want Petrinaut to be directly coupled to Sentry:

- Adds React Context is added `ErrorTrackerContext`
- Adds ErrorTrackerContext Provider for Sentry `<SentryErrorTrackerProvider>` (only used in demo-website)

This `SentryErrorTrackerProvider` is still not used for now in the app, but will be extended later to add more granular tracking of user experience in the demo app.

### User Feedback

For now the User Feedback widget is the default one, this keeps it simple and prevents having to deal with React component props, or with design.

If we get a good Figma design, we'll be able customize this feedback button to place it properly in the UI.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Given the User Feedback widget is the default one, it's on top of the UI and could possibly hide some things.

## 🐾 Next steps

In the future, we'll do more granular tracking of what happens in the app.

## ❓ How to test this?

Use latest Petrinaut deployment

### DIFF
--- a/libs/@hashintel/petrinaut/demo-site/sentry/instrument.ts
+++ b/libs/@hashintel/petrinaut/demo-site/sentry/instrument.ts
@@ -1,3 +1,6 @@
+// Sentry is only used in the demo site, which is bundled by Vite,
+// so we can keep @sentry/react a dev dependency.
+// eslint-disable-next-line import/no-extraneous-dependencies
 import * as Sentry from "@sentry/react";
 
 Sentry.init({


kube:
* **#8241** <a href="https://app.graphite.com/github/pr/hashintel/hash/8241?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8241?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libs/@hashintel/petrinaut/.env
```diff
@@ -0,0 +1 @@
+SENTRY_DSN=https://5c0a691258164a68eac390cd10574484@o146262.ingest.us.sentry.io/4510663776141312
```
CiaranMn: This feels like it shouldn't be committed to the repo. I've added it as an environment variable in Vercel so it will be picked up when built in Vercel.

For local testing we can add a non-committed `.env.local` with the value.

libs/@hashintel/petrinaut/demo-site/sentry/instrument.ts
```diff
@@ -0,0 +1,31 @@
...
+// The value is read from .env files and exposed via import.meta.env.SENTRY_DSN
+const sentryDsn =
+  typeof import.meta.env.SENTRY_DSN === "string"
+    ? import.meta.env.SENTRY_DSN
+    : undefined;
+
+const environment =
+  typeof import.meta.env.MODE === "string"
+    ? import.meta.env.MODE
+    : "development";
```
CiaranMn: Let's call this 'environment' as it's what it tends to be called elsewhere.

I know that Vite sets a special 'mode' variable, but we don't actually want to rely on this – it means that if someone does a production build locally and runs it, any errors will be sent to Sentry as production errors.

The only 'production' errors in Sentry should be those from the main production site, so we need to handle setting the environment manually (via Vercel environment variables, added in a file below).
```suggestion
  typeof import.meta.env.ENVIRONMENT === "string"
    ? import.meta.env.ENVIRONMENT
    : "development";
```

libs/@hashintel/petrinaut/demo-site/vite-env.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="vite/client" />
+
+interface ImportMetaEnv {
+  readonly SENTRY_DSN?: string;
+  readonly MODE: string;
```
CiaranMn:
```suggestion
  readonly ENVIRONMENT: string;
```

libs/@hashintel/petrinaut/vite.config.ts
```diff
@@ -1,12 +1,20 @@
...
 
 export default defineConfig(({ mode }) => {
   const isLibMode = mode === "lib" || !!process.env.VITEST;
 
+  // Load environment variables from .env files
+  // For demo-site builds, load from the package root directory
+  const envDir = isLibMode ? undefined : path.resolve(__dirname);
+  const env = loadEnv(mode, envDir ?? process.cwd(), "");
+
+  // Get SENTRY_DSN from environment variables
```
CiaranMn:
```suggestion
```

libs/@hashintel/petrinaut/vite.config.ts
```diff
@@ -1,12 +1,20 @@
...
 import react from "@vitejs/plugin-react";
-import { defineConfig } from "vite";
+import { defineConfig, loadEnv } from "vite";
 import dts from "vite-plugin-dts";
 
 export default defineConfig(({ mode }) => {
   const isLibMode = mode === "lib" || !!process.env.VITEST;
 
+  // Load environment variables from .env files
+  // For demo-site builds, load from the package root directory
```
CiaranMn:
```suggestion
```

libs/@hashintel/petrinaut/vite.config.ts
```diff
@@ -1,12 +1,20 @@
...
+  const envDir = isLibMode ? undefined : path.resolve(__dirname);
+  const env = loadEnv(mode, envDir ?? process.cwd(), "");
+
+  // Get SENTRY_DSN from environment variables
+  const sentryDsn = env.SENTRY_DSN ?? process.env.SENTRY_DSN;
```
CiaranMn: Load the environment from Vercel (which will be either `production` for the main site or `preview` for preview builds)

See https://vercel.com/docs/frameworks/frontend/vite#environment-variables
```suggestion
  const sentryDsn = env.SENTRY_DSN ?? process.env.SENTRY_DSN;
  const environment = process.env.VITE_VERCEL_ENV ?? "development";
```

libs/@hashintel/petrinaut/vite.config.ts
```diff
@@ -72,6 +80,8 @@ export default defineConfig(({ mode }) => {
     define: {
       // Provide minimal process shim for TypeScript language service in browser
       "process.versions": JSON.stringify({ pnp: undefined }),
+      // Expose SENTRY_DSN at build time
+      "import.meta.env.SENTRY_DSN": JSON.stringify(sentryDsn),
```
CiaranMn:
```suggestion
      "import.meta.env.SENTRY_DSN": JSON.stringify(sentryDsn),
      "import.meta.env.ENVIRONMENT": environment,
```

libs/@hashintel/petrinaut/demo-site/sentry/instrument.ts
kube: `MODE` is a constant provided by Vite:  
  
https://vite.dev/guide/env-and-mode#built-in-constants  
  
We could provide our own but it would be repeating what is already available, or use `PROD === true`.

libs/@hashintel/petrinaut/vite.config.ts
CiaranMn: This line only makes sense for the demo site build so maybe we should only inject it in there somehow?

CiaranMn: We can leave it in here for now

libs/@hashintel/petrinaut/.env
kube: Yeah, I remove it, environment variable is certainly better.

libs/@hashintel/petrinaut/package.json
```diff
@@ -48,6 +48,7 @@
     "@hashintel/ds-helpers": "0.0.1-b",
     "@local/eslint": "0.0.0-private",
     "@pandacss/dev": "1.4.3",
+    "@sentry/react": "10.22.0",
```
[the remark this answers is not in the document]
kube: This is on purpose, because only used by the Demo website, and we don't want this as a dependency of the distributed library.

libs/@hashintel/petrinaut/demo-site/sentry/instrument.ts
kube: Thanks and sorry for misreading/misunderstanding your point.

libs/@hashintel/petrinaut/vite.config.ts
kube: `define` in the config works by text replacement (if I'm not mistaken)

So this won't be bundled in the library, because there won't be no reference to the global variable in the library.

kube: I'm gonna use `define` for that.

kube: @CiaranMn I updated to use `define`, with `__SENTRY_DSN__` and `__ENVIRONMENT__`

CiaranMn (approved):

PR #8254, opened 2026-01-14
SRE-260: Disable immutable installs for changeset version script

CiaranMn:

#8253 fixed one issue in the Release workflow but it uncovered another, which this PR should fix. The release workflow needs to change `yarn.lock` in CI because converting the changeset files to package bumps should also result in a `yarn.lock` update, but Yarn 4's default behaviour is to error if a `yarn` in CI causes a diff in `yarn.lock`.


TimDiekmann (dismissed):
Great how all three AIs don't get the point 😆

CiaranMn:
> Great how all three AIs don't get the point 😆

They're actually right, it's the `yarn` we need to be immutable 😁 

I didn't know that it would only apply to the first command

PR #8255, opened 2026-01-14
Version Packages

hashdotai:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#8083](https://github.com/hashintel/hash/pull/8083) [`469eb01`](https://github.com/hashintel/hash/commit/469eb01a176e94f9ba881701e9d41f3c4390ca67) Thanks [@CiaranMn](https://github.com/CiaranMn)! - major overhaul that removes temporal/non-temporal split (all now temporal) and implements latest type system

### Patch Changes

-   Updated dependencies \[[`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6)]:
    -   @blockprotocol/type-system@0.2.0

### Minor Changes

-   [#7714](https://github.com/hashintel/hash/pull/7714) [`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6) Thanks [@CiaranMn](https://github.com/CiaranMn)! - major overhaul including new metadata, data type inheritance

### Patch Changes

-   Updated dependencies \[[`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6), [`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6), [`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6), [`469eb01`](https://github.com/hashintel/hash/commit/469eb01a176e94f9ba881701e9d41f3c4390ca67)]:
    -   @blockprotocol/type-system@0.2.0
    -   @hashintel/design-system@0.0.9
    -   @blockprotocol/graph@0.4.0

### Patch Changes

-   [#7714](https://github.com/hashintel/hash/pull/7714) [`f7a47f1`](https://github.com/hashintel/hash/commit/f7a47f15f97eb5ef79a35edf76078a45bdc36cc6) Thanks [@CiaranMn](https://github.com/CiaranMn)! - add data type selector, more icons

[...]

PR #8261, opened 2026-01-15
H-5998: Version fixes for JS package publishing

CiaranMn:

Some fixes which will allow us to merge #8255 when we choose to, and publish public JS packages:
1. Change a Petrinaut version bump from `minor` to `patch` (at this early `< 1.0.0` stage we'll just keep patching until something seems significant enough for a minor, by some fuzzy criteria TBC)
2. Set `@hashintel/ds-helpers` and `@hashintel/ds-components` version to `0.0.1` so that changesets correctly bumps them to `0.0.2` – it is currently setting them to `1.0.0`, it doesn't understand the weird not-really-semver I used to publish manually a few versions before.
3. Ignore `@hashintel/petrinaut-old` for the purpose of changeset, so it doesn't attempt to publish it. This means that publishing merging #8255 is not blocked on removing that package from the repo.

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

Once it's merged we'll need to check the diff on #8255 to see if the desired changes mentioned above are correctly implemented:
1. `ds-helpers` and `ds-components` bumped to `0.0.2`
2. No mention of `@hashintel/petrinaut-old`

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,7 +6,13 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@apps/*", "@local/*", "@rust/*", "@tests/*"],
+  "ignore": [
+    "@apps/*",
+    "@local/*",
+    "@rust/*",
+    "@tests/*",
+    "@hashintel/petrinaut-old"
+  ],
   "snapshot": {
     "useCalculatedVersion": true
   },


.changeset/config.json
```diff
@@ -6,7 +6,13 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@apps/*", "@local/*", "@rust/*", "@tests/*"],
+  "ignore": [
+    "@apps/*",
+    "@local/*",
+    "@rust/*",
+    "@tests/*",
+    "@hashintel/petrinaut-old"
```
kube, line 14: 👍

.changeset/empty-forks-flow.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@hashintel/petrinaut": minor
```
kube, line 2: Sorry for this. Indeed better to keep patches only for now.

PR #8262, opened 2026-01-16
FE-204: Improve Transitions and Arcs animation

kube:
#### Simple strokeWidth animation

https://github.com/user-attachments/assets/6af7d3f6-8639-484d-a792-e0d0b2bd4419

Because we do not necessarily show all frames during Playback, we possibly skip the frames where transitions happen, and no transition animation is shown to the user.

This PR solves that by not relying on the current frame only (check if current frames fires a given transition), but instead relying a count of transitions since previous frame.

### Arc Animation

This PR leverages the same idea to show a "flow" inside arcs when their linked transition fired.

Because a transition firing is a discrete event (not even happening inside one frame, but between two), we display an animation for a certain amount of time to make it clear to the user that something happened in it.

But because an other firing can happen during the animation of a previous one, we cumulate the remaining of animations so the arc width increases to reflect this accumulation.

### Other fixes

- Fix token count display in `PlaceNode` inside `SDCPNView`

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

This is just a first step in animating the "flow" inside the Arcs and Transitions.

Many better animations are possible, we'll iterate on that later to make it better.

> Currently orchestration of every animation still relies on React, and this can lead to some performance issues.
>
> Though there are many performance bottlenecks to address upstream prior to enhancing this part alone.

### DIFF
--- a/libs/@hashintel/petrinaut/src/views/SDCPN/components/arc.tsx
+++ b/libs/@hashintel/petrinaut/src/views/SDCPN/components/arc.tsx
@@ -163,17 +163,8 @@ export const Arc: React.FC<EdgeProps<ArcData>> = ({
         />
       )}
 
-      {/* Main edge with marker */}
-      <path
-        id={id}
-        d={arcPath}
-        markerEnd={markerEnd}
-        fill="none"
-        stroke={style?.stroke ?? "#b1b1b7"}
-        strokeWidth={style?.strokeWidth ?? BASE_STROKE_WIDTH}
-        className="react-flow__edge-path"
-        style={style}
-      />
+      {/* Main edge with marker - using BaseEdge for proper interaction handling */}
+      <BaseEdge id={id} path={arcPath} markerEnd={markerEnd} style={style} />
 
       {/* Animated overlay path for firing visualization (no marker) */}
       <path


kube:
* **#8262** <a href="https://app.graphite.com/github/pr/hashintel/hash/8262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8258** <a href="https://app.graphite.com/github/pr/hashintel/hash/8258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libs/@hashintel/petrinaut/src/views/SDCPN/hooks/use-firing-delta.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef } from "react";
+
+/**
+ * Hook to track the previous firingCount and compute the delta.
+ * Returns the delta and sign symbol when firingCount changes.
```
kube:
```suggestion
```

libs/@hashintel/petrinaut/src/simulation/simulator/types.ts
```diff
@@ -47,7 +53,15 @@ export type SimulationFrame = {
...
     { instance: Place; offset: number; count: number; dimensions: number }
   >;
-  transitions: Map<ID, { instance: Transition; timeSinceLastFiring: number }>;
+  transitions: Map<
+    ID,
+    {
+      instance: Transition;
+      timeSinceLastFiring: number;
+      justFired: boolean;
+      firingCount: number;
```
CiaranMn: I think these would benefit from docstrings – I see you add them below, maybe we can make a type with these three fields and then share it so they don't need repeating?

libs/@hashintel/petrinaut/src/simulation/context.ts
```diff
@@ -9,6 +9,34 @@ export type SimulationState =
...
+  };
+  transitions: {
+    [transitionId: string]:
+      | {
+          /** Time since last firing of the transition at the time of the frame. */
+          timeSinceLastFiring: number;
+          /** Whether this transition fired in this frame. */
+          justFired: boolean;
+          /** Total number of times this transition has fired up to this frame. */
+          firingCount: number;
```
CiaranMn: Some comments which I think make things a bit clearer, happy to discuss. 

1. I generally prefer putting the units in things like durations unless it's otherwise obvious. Alternatively we can add it to the docstring.
2. I'd just call `firedInThisFrame` so then you don't have to explain what `justFired` means (just could mean 'now' or 'very recently'
3. Presumably `firingCount` is the number of times the transition has fired up to this frame _since it was last animated_? Or similar? it's not literally the total number of firings from Frame 1 to current frame
```suggestion
          timeSinceLastFiringMs: number;
          /** Whether this transition fired in this frame. */
          firedInThisFrame: boolean;
          /** Total number of times this transition has fired between this frame and the previously selected frame (if it was the past). */
          firingCount: number;
```

CiaranMn (commented):
Just one comment on variable naming and documentation.

I prefer the arc animation without lines. Looks good, and great simulation speed up!

libs/@hashintel/petrinaut/src/simulation/context.ts
kube: > I'd just call firedInThisFrame so then you don't have to explain what justFired means (just could mean 'now' or 'very recently'

The thing is it actually does not fire in the frame, but between two frames, because it is a discrete event. (I only named the variables, comments were added by Claude)

I'm renaming it as you suggest because there is no "good" way of naming it, but these discrete events are a bit subtle, as there is no frame in which you "see" the transition firing, only a difference between a previous and current frame.

kube: > Presumably firingCount is the number of times the transition has fired up to this frame since it was last animated? Or similar? it's not literally the total number of firings from Frame 1 to current frame

`firingCount` is now the total number of firings of the transition since the beginning of the simulation.

It is the best way to both compute the delta between two frames (for animation purpose), and to later visualize this in the timeline too.

- `firedInThisFrame` tells you for the "instant".
- `firingCount` tells you for the cumulated count.

And there is no notion of "delta between two frames" directly in the SimulationContext anymore. The delta is directly computed by the component doing the rendering, and could be possibly moved to a new `PlaybackContext` later.

This prevents SimulationContext having to know anything about the playback.

libs/@hashintel/petrinaut/src/views/SDCPN/hooks/use-firing-delta.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (firingCount === prevFiringCountRef.current) {
+    return null;
+  }
+
+  const prevCount = prevFiringCountRef.current ?? 0;
+  const delta = firingCount - prevCount;
+
+  return delta;
+}
```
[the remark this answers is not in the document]
kube: That's by design.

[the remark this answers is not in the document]
kube: If user scrubs through the Timeline, we animate the flow of Arcs to give a sense of what happened between these frames.

The animation duration is always the same, and the stroke width uses a logarithmic scale of the number of transitions, to prevent having something too "big" on screen.

libs/@hashintel/petrinaut/src/simulation/context.ts
CiaranMn: Ahh got it, fair enough! Thank you

libs/@hashintel/petrinaut/src/simulation/simulator/types.ts
kube: > maybe we can make a type with these three fields and then share it so they don't need repeating?

At the moment I'm trying to decouple implementation (Provider) and interface (Context), and it's still not finished as `SimulationContext` exposes the raw `SimulationInstance` from current `simulator`.

So I'd like if possible not starting sharing types between them, I know it's redundant and not elegant, but it has the benefit of not creating complex dependency graphs.

Maybe we can do this once we have better interfaces defined? WDYT?

kube: I create a dedicated type and moved types that were imported from Simulator, directly into SimulationContext.

I created a technical ticket to make them more generic later. FE-207

This will be useful when adding HASH simulation.

PR #8297, opened 2026-01-23
Revert "Update Rust toolchains to nightly-2026-01-12"

TimDiekmann:
Reverts hashintel/hash#8235

The [BE Deployment Job](https://github.com/hashintel/hash/actions/workflows/hash-backend-cd.yml) is failing with an ICE.


TimDiekmann:
We're going to bump the toolchain version instead. A local test showed that the latest toolchain fixed it.

PR #8325, opened 2026-01-26
H-6083: Fix `npm` pre-publishing scripts, mark some packages as private

CiaranMn:

Attempts to fix [the issues](https://github.com/hashintel/hash/actions/runs/21359488258/job/61484041479) with package publishing by:

1. Fixing the pre-publish/prepare scripts in packages we are actively developing
2. Mark those which are deprecated/don't need publishing as `private`

### DIFF
--- a/libs/@hashintel/petrinaut/package.json
+++ b/libs/@hashintel/petrinaut/package.json
@@ -30,7 +30,7 @@
     "fix:eslint": "eslint --fix .",
     "lint:eslint": "eslint --report-unused-disable-directives .",
     "lint:tsc": "tsc --noEmit",
-    "prepare": "turbo run build",
+    "prepublishOnly": "turbo run build",
     "preview": "vite preview",
     "test:unit": "vitest"
   },


libs/@hashintel/petrinaut/package.json
```diff
@@ -21,6 +30,7 @@
     "fix:eslint": "eslint --fix .",
     "lint:eslint": "eslint --report-unused-disable-directives .",
     "lint:tsc": "tsc --noEmit",
+    "prepare": "turbo run build",
```
kube: Does this mean that we will have `turbo run build` of `@hashintel/petrinaut` run every time we do a `yarn install`?

Are you sure it's safe to directly call `turbo` here?

I guess you want to make the CI build things properly with dependencies, but I wonder what the impacts will be on local development, and how it could collide with building from workspace's root.

CiaranMn:
> Looks sensible. I can imagine petrinaut is still a bit of a tangle; I'd like to convert it to `tsdown` but I guess you also want to separate the library from the demo site, yeah?

@lunelson yes – we have a somewhat analogous situation in `ds-helpers` in that we have a `build` command that doesn't build the library but instead the Ladle site. Ideally this would be more separated also (or at least not take up the `build` command, which we should reserve for building the main thing). Can sort out later though, just want to get publishing working.

libs/@hashintel/petrinaut/package.json
CiaranMn: `turbo` will work here, yes.

On running on `yarn install` – yes, that's true. I've switched it to `prepublishOnly` in https://github.com/hashintel/hash/pull/8325/commits/5464e2065b14b1159e2583d2487a4dd8e3cba701

I've left `ds-*` as `prepare` for now, we can look to strip some of them back to `prepublishOnly` if possible (I think it definitely is _not_ possible for `ds-theme` as it needs to do codegen stuff for the TS files to even exist)

PR #8342, opened 2026-02-01
SRE-415: Fix vercel not building

TimDiekmann:
## 🌟 What is the purpose of this PR?

Remove the sourcing of Cargo environment variables in the Vercel build script, as it's no longer needed with mise handling environment setup.

## 🔗 Related links

- https://linear.app/hash/issue/H-3212/clean-up-env-files (internal)
- https://linear.app/hash/issue/H-3213/use-consistent-naming-schema-for-environment-variables (internal)
- https://linear.app/hash/issue/H-4202/sort-out-which-environment-variables-are-defined-where (internal)

## 🔍 What does this change?

- Removes `source "$HOME/.cargo/env"` from the Vercel build script
- Adds a reference to H-3212 in the environment variable cleanup comments

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Vercel build process will validate this change works correctly

## ❓ How to test this?

1. Checkout the branch
2. Verify the Vercel build completes successfully without the Cargo environment sourcing

PR #8345, opened 2026-02-02
BE-359: Refactor entity embedding generation to use entity IDs instead of full entities

TimDiekmann:

This PR improves entity embedding security by moving property filtering from the backend to the worker. Instead of sending full entities to the worker, we now only send entity IDs and let the worker fetch and filter the entities, ensuring sensitive data is properly excluded from embeddings.

## 🔍 What does this change?

- Refactors the embedding workflow to accept entity IDs instead of full entity objects
- Moves property filtering logic from Rust to TypeScript in the worker
- Adds support for configurable embedding exclusions based on entity type
- Improves filtering by applying exclusions before embedding generation
- Increases chunk size for entity ID batches (from 100 entities to 10,000 IDs)

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Existing tests for embedding generation

## ❓ How to test this?

1. Checkout the branch
2. Create entities with properties that should be excluded from embeddings
3. Verify that the excluded properties are not included in the generated embeddings

### DIFF
--- a/apps/hash-ai-worker-ts/src/workflows.ts
+++ b/apps/hash-ai-worker-ts/src/workflows.ts
@@ -315,6 +315,11 @@ export const updateEntityEmbeddings = async (
   // Build filter from entity IDs if provided
   let filter: Filter;
   if ("entityIds" in params) {
+    // Early return if no entity IDs provided - avoids ambiguous empty `any` filter
+    if (params.entityIds.length === 0) {
+      return usage;
+    }
+
     // Build a filter matching any of the entity IDs, excluding FlowRun entities
     filter = {
       all: [


TimDiekmann:
* **#8368** <a href="https://app.graphite.com/github/pr/hashintel/hash/8368?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8345** <a href="https://app.graphite.com/github/pr/hashintel/hash/8345?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8345?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8332** <a href="https://app.graphite.com/github/pr/hashintel/hash/8332?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8342** <a href="https://app.graphite.com/github/pr/hashintel/hash/8342?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/hash-ai-worker-ts/src/workflows.ts
```diff
@@ -340,6 +366,8 @@ export const updateEntityEmbeddings = async (
        * Don't try to create embeddings for `FlowRun` entities, due to the size
        * of their property values.
        *
+       * This is a safety fallback - the filter should already exclude these.
```
indietyp, line 369: Would it make sense to log this? Seems like something we should be alerted at.

TimDiekmann: This is mainly useful if we provide a filter e.g. when manually triggering embeddings. Then it would just be noise. In any other circumstance, this would just test if the filter above works.

apps/hash-ai-worker-ts/src/workflows.ts
```diff
@@ -375,9 +403,24 @@ export const updateEntityEmbeddings = async (
...
+              delete filteredProperties[propertyBaseUrl];
+            }
+          }
+        }
+      }
+
       const generatedEmbeddings =
         await aiActivities.createEntityEmbeddingsActivity({
-          entityProperties: entity.properties,
+          entityProperties: filteredProperties,
```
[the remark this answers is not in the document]
TimDiekmann: We're going to move this endpoint to the graph.

PR #8369, opened 2026-02-08
FE-223, H-5780, H-6227: Setup Petrinaut inside HASH + Build with Vite 8 and TypeScript 7

kube:
https://github.com/user-attachments/assets/e4664d61-0ed1-4bc1-a519-af6e25c25e07

Remove `@hashintel/petrinaut-old`, integrate the new `@hashintel/petrinaut` into Hash frontend, fix CSS scoping for embedding, and migrate the build to **Vite 8 (Rolldown) + TSGo**.

### 1\. Remove `@hashintel/petrinaut-old`

- Deletes `libs/@hashintel/petrinaut-old/` and removes it from workspace config

### 2\. Integrate new Petrinaut into Hash frontend

- Switches `apps/hash-frontend` from `petrinaut-old` to `@hashintel/petrinaut`
- Adds `convert-net-formats.ts` (Hash entities ↔ SDCPN format conversion)
- Refactors `use-process-save-and-load.tsx` for the new API

### 3\. Migrate to Vite 8 + Rolldown + TSGo

- Upgrades to **Vite 8** (beta) with **Rolldown** bundler (`rolldownOptions` replaces `rollupOptions`)
- Replaces `typescript` 5.9 with **TSGo 7.0** (`@typescript/native-preview`) for type-checking
- Makes `tsconfig.json` autonomous (no longer extends monorepo shared config) for TSGo compatibility

### 4\. Split Vite configs

- **`vite.config.ts`** — library build only (previously handled both via `--mode=lib`)
- **`vite.site.config.ts`** — demo site dev/build (will become a separate package)

### 5\. Vite external dependencies

- Externalizes heavy deps (`typescript`, `monaco-editor`, `@babel/standalone`) from the library bundle
- Uses `esmExternalRequirePlugin` for externals bundled in library still relying on `require`
    - (`elkjs`, `react/jsx-runtime`, `react/compiler-runtime`)

### 6\. Fix panel positioning and worker URL

- Panel positioning: `position: fixed` → `absolute` for proper integration into an existing UI layout
- Worker URL: uses `experimental.renderBuiltUrl` to fix resolution — Vite's default `config.base` adds a `"" +` prefix that breaks worker URLs in consumers

### 7\. Replace MUI Tooltip with Ark UI

- Removes `@mui/material` — replaces `MuiTooltip` / `SvgIcon` with the existing Ark UI `Tooltip` and a plain `<svg>`
- Adds `flip`, `slide`, and `overflowPadding` so tooltips stay within the viewport
- Resets inherited text styles (`fontWeight`, `textTransform`, `letterSpacing`) and caps width via `maxWidth: min(300px, var(--available-width))`

### 8\. Fix CSS scoping for HASH embedding (⚠️ important)

Two PandaCSS scoping issues fixed in `panda.config.ts`:

**A — Color tokens not on** **`.petrinaut-root`:**
The `@hashintel/ds-theme` preset scopes `light`/`dark` conditions to `:root &`, so PandaCSS put color variables on `:where(:root,...)` instead of `.petrinaut-root` — undefined when embedded.

#### **Fix**

Override conditions to scope to `.petrinaut-root`:

```ts
conditions: {
  extend: {
    light: ".petrinaut-root &",
    dark: ".dark .petrinaut-root &, [data-theme='dark'] .petrinaut-root &",
  },
},
```

**B — CSS** **`@layer`** **vs unlayered global reset:**
Hash's unlayered global reset (`* { padding: 0; ... }`) always beats PandaCSS's `@layer`\-wrapped utilities, per the CSS cascade spec.

#### **Fix**

Enable `polyfill: true` — transforms `@layer` into `:not(#\#)` specificity chains so normal cascade rules apply. Adds ~14 kB uncompressed (~0.6 kB gzipped).

> **Note:** Ideally Hash's global reset would be wrapped in a CSS layer instead. This polyfill is a workaround.

#### **Problem**

Webpack rewrites `typeof window` to `"object"` (browser optimization), which breaks in Web Workers — `window` is not defined there, so guarded code paths like `typeof window !== "undefined"` evaluate to `true` and crash with `ReferenceError`.

#### Fix

Rolldown `replacePlugin` in the Vite `worker` config pre-resolves `typeof window` to `"undefined"` at build time, so environment checks work correctly regardless of the consumer's bundler.

- [x] No publishable changes

- [x] Internal changes only

- [x] No execution graph changes

## ⚠️ Known issues

- CSS `@layer` polyfill is a workaround — long-term fix is wrapping Hash's global CSS in a layer
- `:not(#\#)` boosts add ~14 kB uncompressed (~0.6 kB gzipped)
- Bundle is really heavy, which makes the Petrinaut component slow to load
- Vite 8 is in beta

## 🐾 Next steps

- Wrap Hash's global CSS reset in a `@layer` to avoid the polyfill
- Lazy-loading of some heavy external libraries, like Monaco and TypeScript
- WebWorker for linting, to remove both heavy external (TypeScript) and runtime pressure (linting) from Main Thread.
- Extract demo site into a separate package

## ❓ How to test this?

- Go to latest HASH deployment and login.
- Navigate to `/process`
- Petrinaut should be usable
    - Should create/save properly (shows a bar at top of screen)
    - Should run simulations properly

### DIFF
--- a/apps/hash-frontend/src/lib/csp.ts
+++ b/apps/hash-frontend/src/lib/csp.ts
@@ -24,6 +24,7 @@ export const buildCspHeader = (nonce: string): string => {
       "https://apis.google.com",
       // Vercel toolbar / live preview widget
       "https://vercel.live",
+      // @todo FE-488 will make this unnecessary
       // Monaco Editor loaded from CDN by @monaco-editor/react (used by petrinaut)
       "https://cdn.jsdelivr.net",
     ],
@@ -33,6 +34,7 @@ export const buildCspHeader = (nonce: string): string => {
       // Required for Emotion/MUI CSS-in-JS inline style injection.
       // @todo Use nonce-based approach via Emotion's cache `nonce` option.
       "'unsafe-inline'",
+      // @todo FE-488 will make this unnecessary
       // Monaco Editor stylesheet loaded from CDN by @monaco-editor/react (used by petrinaut)
       "https://cdn.jsdelivr.net",
     ],
@@ -51,6 +53,7 @@ export const buildCspHeader = (nonce: string): string => {
 
     "font-src": [
       "'self'",
+      // @todo FE-488 will make this unnecessary
       // Monaco Editor CSS embeds the Codicon icon font as an inline base64 data URI
       "data:",
     ],


kube:
* **#8472** <a href="https://app.graphite.com/github/pr/hashintel/hash/8472?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#8463](https://github.com/hashintel/hash/pull/8463) <a href="https://app.graphite.com/github/pr/hashintel/hash/8463?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#8509](https://github.com/hashintel/hash/pull/8509) <a href="https://app.graphite.com/github/pr/hashintel/hash/8509?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#8300** <a href="https://app.graphite.com/github/pr/hashintel/hash/8300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8369** <a href="https://app.graphite.com/github/pr/hashintel/hash/8369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TimDiekmann:
CI Errors should be fixed, updated the branch.

apps/hash-frontend/src/pages/process.page/process-editor-wrapper/use-process-save-and-load/use-persisted-nets.ts
```diff
@@ -37,142 +23,24 @@ export const getPersistedNetsFromSubgraph = (
...
-        );
-      },
-    );
+    const rawDefinition =
+      net.properties[
+        "https://hash.ai/@h/types/property-type/definition-object/"
+      ];
 
-    const parentProcess = outgoingLinkAndRightEntities[0]?.rightEntity[0];
+    const definition = rawDefinition as SDCPN;
```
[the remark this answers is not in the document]
kube: Won't fix, we don't have any data in db yet.

apps/hash-frontend/src/lib/csp.ts
```diff
@@ -24,13 +24,17 @@ export const buildCspHeader = (nonce: string): string => {
       "https://apis.google.com",
       // Vercel toolbar / live preview widget
       "https://vercel.live",
+      // Monaco Editor loaded from CDN by @monaco-editor/react (used by petrinaut)
+      "https://cdn.jsdelivr.net",
```
CiaranMn, line 29: Will this become unnecessary with your next-up work?

Can we please add a `@todo H-XXXX will make this unnecessary` comment here? 

CI will then enforce that we address the `@todo` when merging a PR that fixes H-XXXX.

apps/hash-frontend/src/lib/csp.ts
```diff
@@ -24,13 +24,17 @@ export const buildCspHeader = (nonce: string): string => {
...
     ],
 
     "style-src": [
       "'self'",
       // Required for Emotion/MUI CSS-in-JS inline style injection.
       // @todo Use nonce-based approach via Emotion's cache `nonce` option.
       "'unsafe-inline'",
+      // Monaco Editor stylesheet loaded from CDN by @monaco-editor/react (used by petrinaut)
+      "https://cdn.jsdelivr.net",
```
CiaranMn, line 39: Same as above, can we please add a @todo H-XXXX will make this unnecessary comment here?

apps/hash-frontend/src/lib/csp.ts
```diff
@@ -45,7 +49,11 @@ export const buildCspHeader = (nonce: string): string => {
       ...(process.env.NODE_ENV === "development" ? ["http:"] : []),
     ],
 
-    "font-src": ["'self'"],
+    "font-src": [
+      "'self'",
+      // Monaco Editor CSS embeds the Codicon icon font as an inline base64 data URI
+      "data:",
```
CiaranMn, line 58: Will this become unnecessary with a future ticket? If so please mention here.

If it needs to stick around, is there a more specific glob pattern we can use?

libs/@hashintel/petrinaut/tsconfig.json
```diff
@@ -1,11 +1,21 @@
 {
-  "extends": "@local/tsconfig/legacy-base-tsconfig-to-refactor.json",
```
CiaranMn, line 2: Why do we need to do this?

apps/hash-frontend/src/lib/csp.ts
kube: I will add a comment in this PR, but it already fixed in the next PR:

https://github.com/hashintel/hash/pull/8300/changes#diff-29bc5726843df5f7d74590b559f6b58ae16ce0f25965784d79239d21b3455734

libs/@hashintel/petrinaut/tsconfig.json
kube: This PR migrates to TypeScript 7, which does not support everything defined in this legacy-base-tsconfig.

I did not want to impact other packages, but maybe it would be better to have a ticket to re-unify everything.

kube: TypeScript 6 was released in Beta last week, its main purpose is to let projects prepare to migration for TypeScript 7.

In Petrinaut the goal was only to reduce typechecking time, so I migrated to version 7 directly.

In other projects, it would be worth doing a first migration to TypeScript 6 to fix the tsconfig:

https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-beta/

vilkinsons (approved):

PR #8398, opened 2026-02-11
BE-390: Update Prune GH action; Kratos hook error handling; framed block `postMessage`

CiaranMn:

A few minor security improvements:
1. Inject `scope` for GitHub 'prune repository' action via `ENV`
2. Don't send full error back to client if Kratos hook fails
3. Limit target `origin` for `FramedBlock` `postMessage` back to parent

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/package.json
+++ b/libs/@hashintel/ds-components/package.json
@@ -74,7 +74,7 @@
     "@vitest/browser-playwright": "^4.0.16",
     "bumpp": "^10.3.2",
     "eslint": "9.39.2",
-    "eslint-plugin-storybook": "10.1.11",
+    "eslint-plugin-storybook": "10.2.7",
     "lucide-react": "0.544.0",
     "npm-run-all": "4.1.5",
     "storybook": "9.1.17",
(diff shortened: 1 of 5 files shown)


apps/hash-frontend/src/components/sandbox/framed-block/framed-block.tsx
```diff
@@ -38,13 +38,20 @@ export const FramedBlock: FunctionComponent = () => {
...
+      // The sandbox iframe is served from the same Next.js app, so the
+      // parent origin matches our own. This prevents cross-origin injection.
+      if (event.origin !== window.location.origin) {
+        return;
+      }
+
+      switch (event.data.type) {
         case "newData":
-          setBlockProperties(data.payload as HashEntity);
+          setBlockProperties(event.data.payload as HashEntity);
```
TimDiekmann: `HashEntity` is a class, do we need to use `new HashEntity` here?

CiaranMn: I've switched to the interface rather than the class, thanks!

PR #8438, opened 2026-02-18
SRE-382: Enforce `workspace:` version specifiers for workspace deps

lunelson:

Migrates all internal workspace dependency ranges to use `workspace:*` (or `workspace:^` for `peerDependencies`) protocol and makes sure that yarn constraints enforce this going forward. This ensures consistent resolution of internal packages across the monorepo and prevents version drift between workspace packages.

## 🔗 Related links

- [SRE-382](https://linear.app/hash/issue/SRE-382) _(internal)_

## 🚫 Blocked by

_Nothing — ready to merge._

## 🔍 What does this change?

- Disable transparent workspaces in Yarn configuration (`18254f2118`)
- Improve Yarn constraint handling for workspace dependencies (`c0e9836400`)
- Update `repo-chores` codegen to emit `workspace:*` for local deps in generated `package.json` files (`a54497227f`)
- Regenerate all `package.json` files with `workspace:*` protocol (`13bd050797`)
- Migrate all remaining internal deps to `workspace:*` (`c48b264fd4`)
- Refine Yarn dependency ignore logic for consistency with the new protocol (`16b7b56ea0`)
- Normalize any remaining internal dependency ranges to `workspace:*` (`c2b97f8f67`)
- ~~Add `yarn lint:workspace-protocol` check to enforce `workspace:*` ranges on all internal dependencies (`5fb9075d15`)~~

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

None.

## 🐾 Next steps

- Consider adding `yarn lint:workspace-protocol` to CI pipeline.

## 🛡 What tests cover this?

- ~~`yarn lint:workspace-protocol` — verifies all internal deps use `workspace:*`~~
- `yarn constraints` — verifies Yarn constraint rules pass
- `yarn install --immutable --mode=skip-build` — verifies lockfile integrity

## ❓ How to test this?

1. Checkout the branch: `git checkout open-workspace-protocol-pr`
2. Run `yarn install --immutable --mode=skip-build` to verify lockfile integrity
3. Run `yarn constraints` to verify Yarn constraints pass
4. ~~Run `yarn lint:workspace-protocol` to verify all internal deps use `workspace:*`~~
5. Confirm all three commands exit with code 0

## 📹 Demo

_No UI changes — internal tooling only._

### DIFF
--- a/libs/@local/repo-chores/rust/src/sync_turborepo/mod.rs
+++ b/libs/@local/repo-chores/rust/src/sync_turborepo/mod.rs
@@ -340,11 +340,9 @@ async fn read_package_json(path: &Utf8Path) -> Result<PackageJson, Report<SyncTu
 }
 
 fn version_protocol_from_str(version: &str) -> VersionProtocol {
-    if version == "workspace:*" {
-        VersionProtocol::Workspace(WorkspaceProtocol::Any { alias: None })
-    } else {
-        VersionProtocol::Tag(version.to_owned())
-    }
+    version
+        .parse()
+        .unwrap_or_else(|_| VersionProtocol::Tag(version.to_owned()))
 }
 
 fn compute_package_json(


libs/@local/repo-chores/node/scripts/check-workspace-protocol.ts
indietyp, line 1: Why is this a separate command instead of integrating with the already existing – and for this exact purpose yarn feature – of yarn constraints?

libs/@local/repo-chores/rust/src/sync_turborepo/mod.rs
```diff
@@ -383,22 +383,36 @@ fn compute_package_json(
...
-                VersionProtocol::Version(package_version(dependency.to())),
+                VersionProtocol::Workspace(WorkspaceProtocol::Any { alias: None }),
             );
         }
     }
 
     let ExtraDependencies { normal, dev } = ExtraDependencies::new(metadata)?;
     for js_dep in normal {
-        dependencies.insert(js_dep.name, VersionProtocol::Tag(js_dep.version));
+        if local_yarn_packages.contains(&js_dep.name) {
```
indietyp: Tho should probably require adjustment in parsing so that we require workspace:* inside of the Cargo.toml instead of silently overwriting it.

yarn.config.cjs
```diff
@@ -95,8 +115,10 @@ function enforceProtocols({ Yarn }) {
     );
 
     if (workspaceDependency) {
-      // turbo doesn't support the `workspace:` protocol when rewriting lockfiles, leading to inconsistent lockfiles
-      dependency.update(workspaceDependency.manifest.version);
+      // Internal workspace dependencies must use the workspace:* protocol
+      if (dependency.range !== "workspace:*") {
```
indietyp: This already does the workspace change so why do we have a separate lint command if this does lifting and fixing?

libs/@local/repo-chores/node/scripts/check-workspace-protocol.ts
lunelson: Thanks, that was over-exuberance from augment. Removed

libs/@local/repo-chores/rust/src/sync_turborepo/mod.rs
lunelson: Updated

yarn.config.cjs
lunelson: 🙏

libs/@local/repo-chores/node/scripts/check-workspace-protocol.ts
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+
+  console.log();
+  process.exit(1);
+};
+
+void (async () => {
+  await script();
+})();
```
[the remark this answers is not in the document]
lunelson: fixed

libs/@hashintel/ds-components/package.json
```diff
@@ -87,8 +87,8 @@
   },
   "peerDependencies": {
     "@ark-ui/react": "^5.26.2",
-    "@hashintel/ds-helpers": "0.0.4",
-    "@hashintel/refractive": "0.0.2",
+    "@hashintel/ds-helpers": "workspace:*",
+    "@hashintel/refractive": "workspace:*",
```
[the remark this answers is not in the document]
lunelson: @indietyp I'm curious what your opinion is on this one 👆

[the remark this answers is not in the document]
indietyp: You can just use

```
workspace:^
```

instead.<br><br>Which I think we should do.

yarn.config.cjs
```diff
@@ -95,8 +115,12 @@ function enforceProtocols({ Yarn }) {
     );
 
     if (workspaceDependency) {
-      // turbo doesn't support the `workspace:` protocol when rewriting lockfiles, leading to inconsistent lockfiles
-      dependency.update(workspaceDependency.manifest.version);
+      const expectedRange =
+        dependency.type === "peerDependencies" ? "workspace:^" : "workspace:*";
```
indietyp: Would it make sense to force this for any published workspace dependencies? Instead of allowing both? aka: force `workspace:*` for internal, and `workspace:^` for published libraries?

lunelson: Interesting point! AFAICT we don't seem to have super clear ways of identifying what is published though:

**Private packages** (`"private": true`):

* `@hashintel/type-editor`
* `@hashintel/query-editor`
* `@hashintel/block-design-system`
* `@hashintel/petrinaut-old`
* `@hashintel/design-system`

**Has** `publishConfig` (explicitly public):

* `@hashintel/ds-components`
* `@hashintel/ds-helpers`
* `@hashintel/ds-theme`

**Neither private nor publishConfig**:

* `@hashintel/petrinaut`
* `@hashintel/refractive`

So the basic reliable heuristic would be **"not private"?** Maybe this is too broad…but the API would seem to support `dependency.workspace.manifest.private !== true` so it should be doable

indietyp: I think that would be it yes. That should cover all our libraries.

lunelson: OK good. Also: the original comment specifically called out `peerDependencies`, but now that I think about it, the "exact-pin" problem applies equally to `dependencies` in publishable packages 🤔 right? —which would mean:

* for publishable packages, use `workspace:^` for both `dependencies` and `peerDependencies` (i.e., everything except `devDependencies`)
* for private packages, `workspace:*` everywhere

```javascript
if (workspaceDependency) {
  const isPublishable = dependency.workspace.manifest.private !== true;
  const expectedRange =
    isPublishable && dependency.type !== "devDependencies"
      ? "workspace:^"
      : "workspace:*";
  if (dependency.range !== expectedRange) {
    dependency.update(expectedRange);
  }
}
```

Does that make sense to you?

indietyp: yep makes sense!

PR #8495, opened 2026-02-27
BE-427: Add hashql-macros crate with procedural macros for ID types and symbols

indietyp:
## 🌟 What is the purpose of this PR?

This PR introduces a new `hashql-macros` crate that provides procedural macros for generating ID types and symbol tables, replacing the existing declarative macro implementations. The new macros offer better error reporting, more flexible syntax, and improved maintainability.

## 🔗 Related links

- Internal refactoring to improve macro ergonomics and error handling

## 🔍 What does this change?

- **Adds new `hashql-macros` crate** with procedural macro implementations for `define_id`, `id`, and `define_symbols`
- **Replaces declarative `newtype!` macro** with `hashql_macros::define_id` that supports both struct and enum ID types
- **Updates all ID type definitions** to use the new macro syntax with `#[id(...)]` attributes
- **Migrates symbol table generation** from declarative macro to `define_symbols!` procedural macro
- **Changes integer backing types** from `usize` to `u32` for most ID types to improve memory efficiency
- **Adds proper error reporting** with span information for better developer experience
- **Introduces attribute-based configuration** for ID types (e.g., `derive(Step)`, `display`, `crate`)
- **Updates Cargo.lock** to include new dependencies: `unsynn`, `mutants`, and `proc-macro2`/`quote`

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Existing unit tests continue to pass with the new macro implementations
- Compile-time assertions verify ID type roundtrip behavior
- Integration tests ensure backward compatibility

## ❓ How to test this?

1. Checkout the branch
2. Run `cargo build` to verify all ID types compile correctly
3. Run existing test suites to confirm functionality is preserved
4. Try creating new ID types using the new syntax to verify macro functionality

PR #8498, opened 2026-03-02
BE-428: HashQL: Simplify traversal tracking to path recording

indietyp:

Replaces the traversal extraction transform pass (which rewrote MIR bodies by splicing load statements and creating new locals) with a pure analysis pass that records which entity paths each block accesses as bitsets. Introduces per-block path transfer cost charging so the placement solver can account for data locality when assigning blocks to backends.

## 🔍 What does this change?

- **Traversal analysis** (`traversal/analysis/`): walks the body, resolves vertex projections via `EntityPath::resolve()`, records accessed paths as `TraversalPathBitSet`. No MIR mutation.
- **Per-block cost analysis** (`cost/analysis.rs`): new `BasicBlockCostAnalysis` combines per-statement costs with a path transfer premium. For each block, charges a transfer premium on targets that are not the natural origin for the accessed paths (e.g., Interpreter pays a premium for `Vectors` because Embedding is the origin). Premiums are charged once per block, not per statement.
- **`BasicBlockCostVec`**: the placement solver now operates on precomputed per-block costs. `PlacementSolverContext` takes `blocks: &BasicBlockCostVec` instead of separate statement costs and assignment arrays.
- **`BlockPartitionedVec<T>`**: generic offset-table-backed per-block storage, extracted from `StatementCostVec` internals.
- **`TraversalLattice`**: lattice structure carrying `VertexType` for dataflow analysis over path bitsets.
- **`EntityPath::origin()`**, **`EntityPath::estimate_size()`**: per-path origin backend and transfer size estimation.
- **`TransferCostConfig`**: variable-size parameters (properties, embeddings, provenance sizes, per-target cost multiplier).
- **`FiniteBitSet::negate()`** in hashql-core.
- **Removed** `pass/transform/traversal_extraction/` (the MIR transform and its tests).

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Path transfer premiums use pure over-approximation: every non-origin backend pays the full premium with no discounting for co-location. This can over-penalize non-origin assignment when multiple blocks access the same path on the same backend (cross-block dedup is not implemented). This is intentional: separable costs are a prerequisite for the PBQP solver planned in BE-436.

The ideal cost model is non-separable: the first block on a backend pays the full data fetch cost, subsequent co-labeled blocks pay zero ("shared-fetch" / pay-once-per-group). This maps to submodular clique costs in the optimization literature. Two frameworks formalize this:

- **Jegelka et al. 2014** ([arXiv:1402.0240](https://arxiv.org/abs/1402.0240), Cooperative Graph Cuts): the "label cost" special case is mathematically identical to shared-fetch. Algorithms include polymatroidal flow (exact for separable surrogates) and semigradient methods. However, all are **binary-only** (k=2 labels).
- **Kolmogorov et al. 2010** ([arXiv:1006.1990](https://arxiv.org/abs/1006.1990), Sum of Submodular Functions): general framework via submodular flow with capacity scaling. Also **binary-only**.

The multi-label extension via alpha-expansion (Boykov, Veksler, Zabih 2001) reduces k-label to a sequence of binary cooperative cuts (each solvable exactly), but only guarantees a local minimum. Our `TransMatrix` is not metric, so no global approximation factor holds. For k=3 backends (growing to 6-7), the overhead of iterative cooperative cut calls on ~100-node graphs is not justified when PBQP R1/R2 reductions on the separable over-approximation give exact solutions in microseconds.

Pure over-approximation does not distort the relative ranking between non-origin backends (the premium cancels in pairwise comparisons) but inflates the absolute cost of non-origin assignment, biasing toward origin backends. Intra-block dedup (implemented here) is the first mitigation; cross-block dedup remains an open problem for k > 2 on cyclic graphs.
:
## 🛡 What tests cover this?

- Unit tests for `BasicBlockCostAnalysis` covering: no vertex access, single origin/non-origin paths, multiple path accumulation, composite path expansion, restricted target domains, cross-block independence
- Unit tests for `TraversalAnalysisVisitor` covering path resolution from MIR projections
- Unit tests for `EntityPathBitSet` covering composite swallowing, lattice operations, normalization
- Existing solver tests updated to use `BasicBlockCostVec`; new `path_premiums_influence_placement` and `provenance_variants_produce_different_premiums` integration tests
- End-to-end `mixed_postgres_embedding_interpreter` test verifying three-backend splitting

## ❓ How to test this?

1. `cargo nextest run --package hashql-mir`
2. `cargo test --package hashql-mir --doc`


indietyp:
* **#8827** <a href="https://app.graphite.com/github/pr/hashintel/hash/8827?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8697** <a href="https://app.graphite.com/github/pr/hashintel/hash/8697?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8696** <a href="https://app.graphite.com/github/pr/hashintel/hash/8696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8694** <a href="https://app.graphite.com/github/pr/hashintel/hash/8694?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8693** <a href="https://app.graphite.com/github/pr/hashintel/hash/8693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8650** <a href="https://app.graphite.com/github/pr/hashintel/hash/8650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8635** <a href="https://app.graphite.com/github/pr/hashintel/hash/8635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8607** <a href="https://app.graphite.com/github/pr/hashintel/hash/8607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8600** <a href="https://app.graphite.com/github/pr/hashintel/hash/8600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8595** <a href="https://app.graphite.com/github/pr/hashintel/hash/8595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8587** <a href="https://app.graphite.com/github/pr/hashintel/hash/8587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8586** <a href="https://app.graphite.com/github/pr/hashintel/hash/8586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8675** <a href="https://app.graphite.com/github/pr/hashintel/hash/8675?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8674** <a href="https://app.graphite.com/github/pr/hashintel/hash/8674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8526** <a href="https://app.graphite.com/github/pr/hashintel/hash/8526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8525** <a href="https://app.graphite.com/github/pr/hashintel/hash/8525?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8524** <a href="https://app.graphite.com/github/pr/hashintel/hash/8524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8501** <a href="https://app.graphite.com/github/pr/hashintel/hash/8501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8498** <a href="https://app.graphite.com/github/pr/hashintel/hash/8498?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8498?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8496** <a href="https://app.graphite.com/github/pr/hashintel/hash/8496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8495** <a href="https://app.graphite.com/github/pr/hashintel/hash/8495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

TimDiekmann (dismissed):

libs/@local/hashql/core/src/id/bit_vec/finite.rs
```diff
@@ -265,14 +281,40 @@ impl<I: Id, T: FiniteBitSetIntegral> FiniteBitSet<I, T> {
     /// Returns `false` if `index` is out of bounds (rather than panicking).
     #[inline]
     #[must_use]
-    pub fn contains(&self, index: I) -> bool {
+    pub const fn contains(&self, index: I) -> bool
+    where
+        I: [const] Id,
```
thehabbos007, line 286: I have not seen the `[const]` syntax before for trait bounds. Is this to specify that the `Id` instance of `T` must be `impl const`?

indietyp: `[const]` is for conditional const bounds, it's the new syntax for const traits, it basically states: this function is `const` if all the traits that are marked `[const]` are also const.

libs/@local/hashql/mir/src/pass/analysis/dataflow/lattice/impls.rs
```diff
@@ -195,6 +195,80 @@ macro_rules! impl_bitset {
```
thehabbos007, line 195: <p class='graphite__hidden'><i>[Re: lines +175 to +177]</i></p>
This was not part of the diff in this file, but i was reading thorugh the implementation and noticed that we are comparing the `value` to the `value.domain_size()` here. Don't we need to compare against the domain size of the lattice itself? as in `self.domain_size`?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/hashintel/hash/8498?utm_source=unchanged-line-comment">Graphite</a>.</p>

libs/@local/hashql/mir/src/pass/analysis/dataflow/lattice/impls.rs
```diff
@@ -195,6 +195,80 @@ macro_rules! impl_bitset {
...
+{
+    #[inline]
+    fn join_owned(&self, mut lhs: (T, U), rhs: &(T, U)) -> (T, U)
+    where
+        (T, U): Sized,
+    {
+        self.0.join(&mut lhs.0, &rhs.0);
+        self.1.join(&mut lhs.1, &rhs.1);
+        lhs
+    }
```
thehabbos007, line 243: Can you help me understand the use for `join_owned over join`? Is the purpose to return back the owed lhs, or should we instead call `join_owned` here for `self.0` and `self.1`?

libs/@local/hashql/mir/src/pass/execution/traversal/entity.rs
```diff
@@ -0,0 +1,651 @@
...
+}
+
+/// Resolved entity field path.
+///
+/// Each variant identifies a specific storage location in the entity schema. Consumers can
+/// exhaustively match on this to generate backend-specific access (SQL expressions, placement
+/// decisions, etc.) without duplicating path resolution logic.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Id)]
+#[id(const)]
+pub enum EntityPath {
```
thehabbos007, line 39: I love this abstraction, well done!

thehabbos007 (approved):
Looks good to me. I have some things here for my own understanding, but I suppose it is spotted in later PRs?

libs/@local/hashql/mir/src/pass/analysis/dataflow/lattice/impls.rs
indietyp: It's developer convenience, sometimes you just want to join and get the result back. `join_owned` does that by default. That's why the default implementation just delegates. But just like `Clone::clone_from` we specialize to be faster depending on what the inner type is.

libs/@local/hashql/mir/src/pass/analysis/dataflow/lattice/impls.rs
indietyp: Yes, both would work, implicitly by contract we ensure that `domain_size` for both is the same, here it's just defensive programming. The domain size is set on construction and cannot be changed, so *should for some reason* a bitset be given to the PowersetLattice that doesn't have the same domain size it still correct, but(!) this could be better, I have added https://linear.app/hash/issue/BE-558/hashql-powersetlattice-debug-assertion-if-domain-size-is-not-the-same to make this more obvious.

PR #8500, opened 2026-03-03
BE-423: Add JWT authentication and actor resolution for Graph admin API

TimDiekmann:

Adds JWT-based authentication to the Graph admin API, with actor ID resolution from JWT email claims for provenance tracking. Designed for Cloudflare Access but compatible with any OIDC-compliant JWT issuer.

## 🔗 Related links

- [BE-423](https://linear.app/hash/issue/BE-423) _(internal)_

## 🚫 Blocked by

- #8497

## 🔍 What does this change?

- **JWT validation module** (`jwt.rs`): JWKS-based token validation with caching (10m TTL), key rotation support, algorithm confusion attack prevention (asymmetric-only allowlist checked before JWKS fetch), 10s HTTP client timeout, and dual token extraction (`Cf-Access-Jwt-Assertion` / `Authorization: Bearer`, case-insensitive per RFC 7235)
- **`JwtAuthentication` / `OptionalJwtAuthentication` extractors**: Axum extractors for mandatory and optional JWT enforcement. When a validator is configured, missing/invalid tokens return 401; when unconfigured (dev mode), auth is skipped
- **`AdminActorId` extractor** (`admin.rs`): Resolves JWT email → `UserId` → `AuthenticatedActor` via DB lookup. Falls back to `X-Authenticated-User-Actor-Id` header in dev mode (no JWT configured). Returns proper 500 on misconfiguration instead of panicking
- **`get_user_id_by_email`**: New `AccountStore` trait method + Postgres implementation that looks up a user's ID by email in entity properties (JSONB containment query through `entity_temporal_metadata` → `entity_editions`)
- **`AuthenticatedActor` in entity deletion**: `delete_entities` now takes `AuthenticatedActor` instead of raw `ActorEntityUuid` for consistent actor typing across the store layer
- **JWT configuration**: Three env vars (`HASH_GRAPH_JWT_JWKS_URL`, `HASH_GRAPH_JWT_AUDIENCE`, `HASH_GRAPH_JWT_ISSUER`) on both `admin-server` and `server --embed-admin`. All three must be provided together or not at all (enforced by clap `requires` + runtime guard)
- **JWKS fetch outside write lock**: Concurrent JWT validations are not blocked during cache refresh

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Email lookup bypasses property masking by design (admin-only operation querying raw JSONB)
- Clap does not support `Option<JwtConfig>` with required fields ([clap-rs/clap#5292](https://github.com/clap-rs/clap/issues/5292)), so fields are `Option` with bidirectional `requires`

## 🐾 Next steps

- Wire `AdminActorId` into remaining admin handlers (currently only `delete_entities` uses it)
- Add integration tests for JWT auth flow
- Configure Cloudflare Access env vars in staging/production

## 🛡 What tests cover this?

- Manually tested end-to-end with real Cloudflare Access JWT tokens against local admin server
- Existing deletion tests updated for `AuthenticatedActor` type change
- Clippy passes clean on all affected packages

## ❓ How to test this?

1. Start the admin server: `cargo run --bin hash-graph --all-features -- admin-server`
2. Without JWT config, endpoints work with `X-Authenticated-User-Actor-Id` header (dev mode)
3. With JWT config (`HASH_GRAPH_JWT_JWKS_URL`, `HASH_GRAPH_JWT_AUDIENCE`, `HASH_GRAPH_JWT_ISSUER`), endpoints require a valid JWT token via `Cf-Access-Jwt-Assertion` or `Authorization: Bearer` header
4. Verify actor resolution: the resolved actor ID is logged on authenticated requests to admin endpoints

### DIFF
--- a/libs/@local/graph/api/src/rest/jwt.rs
+++ b/libs/@local/graph/api/src/rest/jwt.rs
@@ -11,7 +11,10 @@
 
 use alloc::{borrow::Cow, sync::Arc};
 use core::time::Duration;
-use std::{sync::RwLock, time::Instant};
+use std::{
+    sync::{PoisonError, RwLock},
+    time::Instant,
+};
 
 use axum::{extract::FromRequestParts, http::request::Parts};
 use error_stack::{Report, ResultExt as _};
@@ -242,9 +245,11 @@ impl JwtValidator {
             .get(self.jwks_url.clone())
             .send()
             .await
-            .change_context(JwtError::JwksFetch)?
+            .change_context(JwtError::JwksFetch)
+            .attach(format!("JWKS URL: {}", self.jwks_url))?
             .error_for_status()
-            .change_context(JwtError::JwksFetch)?;
+            .change_context(JwtError::JwksFetch)
+            .attach(format!("JWKS URL: {}", self.jwks_url))?;
 
         let jwks: JwkSet = response
             .json()
@@ -252,7 +257,7 @@ impl JwtValidator {
             .change_context(JwtError::JwksFetch)
             .attach("failed to deserialize JWKS response")?;
 
-        *self.cache.write().expect("JWKS cache lock poisoned") =
+        *self.cache.write().unwrap_or_else(PoisonError::into_inner) =
             Some((Instant::now(), jwks.clone()));
 
         // http_client (mutex guard) is dropped here, after the cache is updated, so concurrent
@@ -264,7 +269,7 @@ impl JwtValidator {
 
     /// Returns cached JWKS if still valid, or `None` if a fetch is needed.
     fn cached_jwks(&self, force_refresh: bool) -> Option<JwkSet> {
-        let cache = self.cache.read().expect("JWKS cache lock poisoned");
+        let cache = self.cache.read().unwrap_or_else(PoisonError::into_inner);
         let (fetched_at, jwks) = (*cache).as_ref()?;
         let age = fetched_at.elapsed();
         if !force_refresh && age < self.jwks_cache_ttl
@@ -298,13 +303,15 @@ fn extract_token_from_headers(headers: &HeaderMap) -> Result<Cow<'_, str>, Repor
             .change_context(JwtError::InvalidTokenEncoding)?;
 
         // RFC 7235: authentication schemes are case-insensitive
-        if let Some(token) = value
+        return value
             .get(..7)
             .filter(|prefix| prefix.eq_ignore_ascii_case("bearer "))
             .and_then(|_| value.get(7..))
-        {
-            return Ok(Cow::Borrowed(token));
-        }
+            .map(Cow::Borrowed)
+            .ok_or_else(|| {
+                Report::new(JwtError::MissingToken)
+                    .attach("Authorization header present but scheme is not Bearer")
+            });
     }
 
     Err(Report::new(JwtError::MissingToken))
@@ -321,7 +328,7 @@ fn extract_token_from_headers(headers: &HeaderMap) -> Result<Cow<'_, str>, Repor
 /// async fn protected_handler(
 ///     JwtAuthentication(claims): JwtAuthentication,
 /// ) -> impl IntoResponse {
-///     format!("Hello, {}", claims.email().unwrap_or_default())
+///     format!("Hello, {}", claims.email.as_deref().unwrap_or_default())
 /// }
 /// ```
 pub struct JwtAuthentication(pub JwtClaims);
--- a/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/mod.rs
@@ -3453,11 +3453,11 @@ impl<C: AsClient> AccountStore for PostgresStore<C> {
         &self,
         email: &str,
     ) -> Result<Option<UserId>, Report<GetActorError>> {
-        Ok(self
+        let rows = self
             .as_client()
-            .query_opt(
+            .query(
                 "
-                SELECT user_actor.id
+                SELECT DISTINCT user_actor.id
                 FROM user_actor
                 JOIN entity_temporal_metadata ON user_actor.id = entity_temporal_metadata.entity_uuid
                 JOIN entity_editions ON entity_temporal_metadata.entity_edition_id = entity_editions.entity_edition_id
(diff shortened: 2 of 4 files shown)


libs/@local/graph/api/src/rest/jwt.rs
```diff
@@ -0,0 +1,341 @@
...
+use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header, jwk::JwkSet};
+use reqwest::{Client, Url};
+use serde::Deserialize;
+use time::OffsetDateTime;
+use tokio::sync::RwLock;
+
+use crate::rest::status::{BoxedResponse, report_to_response};
+
+/// How long to cache JWKS keys before re-fetching.
+const JWKS_CACHE_TTL: Duration = Duration::from_hours(1);
```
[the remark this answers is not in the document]
TimDiekmann: https://doc.rust-lang.org/stable/core/time/struct.Duration.html#method.from_hours

libs/@local/graph/api/src/rest/jwt.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        let claims = validator
+            .validate(&token)
+            .await
+            .attach_opaque(StatusCode::Unauthenticated)
+            .map_err(report_to_response)?;
+
+        Ok(Self(claims))
+    }
+}
```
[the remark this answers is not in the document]
TimDiekmann: This is acceptable.

libs/@local/graph/api/src/rest/jwt.rs
```diff
@@ -0,0 +1,378 @@
...
+            .change_context(JwtError::JwksFetch)
+            .attach("failed to deserialize JWKS response")?;
+
+        *self.cache.write().await = Some((Instant::now(), jwks.clone()));
+        Ok(jwks)
+    }
+
+    /// Returns cached JWKS if still valid, or `None` if a fetch is needed.
+    async fn cached_jwks(&self, force_refresh: bool) -> Option<JwkSet> {
+        let cache = self.cache.read().await;
```
indietyp: if you don't hold the lock across an await point it's considered better to instead just use the std equivalent. See: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

> Contrary to popular belief, it is ok and often preferred to use the ordinary Mutex from the standard library in asynchronous code.

The feature that the async mutex offers over the blocking mutex is the ability to keep it locked across an .await point. This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used. The primary use case for the async mutex is to provide shared mutable access to IO resources such as a database connection. If the value behind the mutex is just data, it’s usually appropriate to use a blocking mutex such as the one in the standard library or parking_lot.

libs/@local/graph/api/src/rest/jwt.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        // Serialize fetches — only one task fetches at a time.
+        let _fetch_guard = self.fetch_mutex.lock().await;
+
+        // Re-check after acquiring the mutex: another task may have refreshed while we waited.
+        if let Some(jwks) = self.cached_jwks(force_refresh).await {
+            return Ok(jwks);
+        }
+
+        let response = self
```
indietyp: Why don't you just guard the HTTP client behind the mutex, instead of using an adjacent mutex? Also wouldn't a semaphore be better here, allowing us to gate the amount of concurrent access instead?

TimDiekmann: The dual-lock decouples these concerns: `std::sync::RwLock` for cache reads (fast, concurrent), `Mutex` for fetch serialization (held across the HTTP round-trip). Readers never contend with the fetch.

This matters specifically during key rotation: a new `kid` triggers one force refresh, but all requests using the previous (still-valid) key continue to be served without blocking. Same for attacker-crafted `kid`s hitting the cooldown path — legitimate traffic is unaffected.

The `Mutex` itself is never used in the hot-path so it's a net win (except the small overhead in the struct size, which is a one-time overhead).

indietyp: That is not really what I meant. I meant: why don't you just guard the HTTP client behind a `Mutex`, essentially the same, but Rust-idiomatic. You access the serialised resource (the remote workload) via the HTTP client, so the mutex should be over that resource.

libs/@local/graph/api/src/rest/jwt.rs
TimDiekmann: Good flag, fixed in [`f20fb0a` (this PR)](https://github.com/hashintel/hash/pull/8500/commits/f20fb0a4fe0f173d02ab59dd876a67fc85240fe2). Note: you commented on `cache` (`RwLock`) but linked the mutex docs. I changed it to use `RwLock` from `std`

libs/@local/graph/api/src/rest/jwt.rs
TimDiekmann: Ah, true, thanks!
Fixed in [`f20fb0a` (this PR)](https://github.com/hashintel/hash/pull/8500/commits/f20fb0a4fe0f173d02ab59dd876a67fc85240fe2)

indietyp (approved):

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -3449,6 +3449,48 @@ impl<C: AsClient> AccountStore for PostgresStore<C> {
...
+                JOIN entity_editions ON entity_temporal_metadata.entity_edition_id = entity_editions.entity_edition_id
+                WHERE entity_temporal_metadata.decision_time @> now()
+                  AND entity_temporal_metadata.transaction_time @> now()
+                  AND EXISTS (
+                      SELECT 1
+                      FROM jsonb_array_elements_text(
+                          entity_editions.properties -> 'https://hash.ai/@h/types/property-type/email/'
+                      ) AS stored_email
+                      WHERE LOWER(stored_email) = LOWER($1)
+                  )",
```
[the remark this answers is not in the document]
TimDiekmann: This is always an array, it's enforced through the underlying property type.

apps/hash-graph/src/subcommand/admin_server.rs
```diff
@@ -37,14 +40,114 @@ impl fmt::Display for AdminAddress {
...
+    )]
+    pub issuer: Option<String>,
+
+    /// How long to cache JWKS keys before re-fetching (in seconds).
+    #[clap(
+        long = "jwt-jwks-cache-ttl",
+        env = "HASH_GRAPH_JWT_JWKS_CACHE_TTL",
+        default_value_t = 600
+    )]
+    pub jwks_cache_ttl_secs: u64,
```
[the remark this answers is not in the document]
TimDiekmann: updated the PR description

PR #8556, opened 2026-03-17
Add back pre-2025-11-28 old format import support

kube:

Re-adds support for importing SDCPN files saved in the pre-2025-11-28 format, which used different field names (`type` instead of `colorId`, `differentialEquationCode` instead of `differentialEquationId`, `iconId` instead of `iconSlug`, etc.).

## 🔗 Related links

- Stacked on #8536

## 🔍 What does this change?

- **Adds `old-format.ts`** — a single consolidated file containing the Zod schema, types, and conversion logic for the pre-2025-11-28 format (previously spread across 4 files)
- **Integrates as last-resort fallback** in `parseSDCPNFile` — tried after versioned and legacy formats both fail

This PR:

The changes in this PR:

The changes in this PR:

## ❓ How to test this?

1. Export a net in the current format
2. Manually edit the JSON to use old field names (`type` instead of `colorId`, etc.)
3. Import — should successfully parse and convert


CiaranMn:
No-one has raised these old formats so I think we can assume it is not needed. We can always resurrect the PR if they do.

PR #8590, opened 2026-03-26
H-6376: Improve Petrinaut style scoping, font loading

CiaranMn:

When trying to use Petrinaut in another application there were issues which this PR fixes:

1. Not all CSS was properly scoped. To simplify this for future libraries, the PR introduces a scoping utility to `ds-helpers`.
2. Fix font loading
3. Don't set an explicit `undefined` value when creating a place (not valid JSON)
4. Hide additional menu controls when net management is controlled by the consuming application (specifically, `New`, `Load Example`, and `Import`, all of which rely on creating new records).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The regressions would be stylistic – no tests for this.

## ❓ How to test this?

1. View the preview deployment for Petrinaut to check no visual regressions (the only change should be that Inter is now correctly loaded as a font, whereas it isn't in demo.petrinaut.org, although the difference is subtle)

### DIFF
--- a/libs/@hashintel/petrinaut/src/fontsource.d.ts
+++ b/libs/@hashintel/petrinaut/src/fontsource.d.ts
@@ -1,3 +1,5 @@
 // tsgo (unlike tsc) requires type declarations for side-effect-only imports.
 // Fontsource packages ship CSS only, so we declare the module manually.
 declare module "@fontsource-variable/jetbrains-mono";
+declare module "@fontsource-variable/inter";
+declare module "@fontsource-variable/inter-tight";


libs/@hashintel/ds-theme/src/main.ts
```diff
@@ -59,220 +60,284 @@ const scaledLineHeights = scaleTokenValues(
...
+  const surfaceStyles = scope
+    ? { ...fontPipelineCssVars, ...documentSurfaceStyles }
+    : documentSurfaceStyles;
 
-      inert: "&:is([inert], [inert] *, [data-inert])",
+  const lightCondition = scope
+    ? `${scope} &`
+    : ':where(:root:not(.dark, [data-theme="dark"])) &, .light &, [data-theme=light] &';
+  const darkCondition = scope
+    ? `.dark ${scope} &, [data-theme='dark'] ${scope} &`
```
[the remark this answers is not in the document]
CiaranMn: The scoped root is inside Petrinaut so consumers shouldn't be modifying it, they can set `.dark` on their own wrapper

libs/@hashintel/ds-theme/src/main.ts
```diff
@@ -59,220 +60,284 @@ const scaledLineHeights = scaleTokenValues(
...
+ * });
+ * ```
+ */
+export function scopedThemeConfig(scope: string) {
+  return {
+    preflight: { scope } as const,
+    cssVarRoot: scope,
+    presets: [createPreset({ scope })],
+  };
+}
```
lunelson, line 336: if downstream consumers want to pass their own presets in to their config, this function would need to accept them; or, this function could return `preset` and `config` items, the former places in `presets` and the latter spread in

CiaranMn: Consumers can just add the `presets` key in their config, like Petrinaut was doing previously (this removed line), or spread + add to the returned presets array. All the PR is doing is adding another function which saves them having to do all the scoping stuff, the default preset export is identical.

Image

Having said that, I'm happy to add another argument to the function which does this for them, if you think worth doing. Is your suggestion that whatever preset they provide would be added to the existing one, or would it be a complete overwrite?

kube:
👍

libs/@hashintel/ds-theme/src/main.ts
lunelson: Right, yeah true enough, they could just add it back. I guess as long as this is clearly documented it should be fine. Basically it's "if you need to add your own presets, you must also include `"@hashintel/ds-theme"` in the array

PR #8622, opened 2026-04-13
H-6219: Restore TOTP MFA and fix AAL2 login redirect

TimDiekmann:

Re-enables the TOTP MFA settings UI that was shipped disabled in H-2421 (hidden behind `{false && …}` with a `@todo H-6219 restore this` note) and fixes the underlying issue that had broken it on staging.

The blocker was Kratos emitting `redirect_browser_to: app.hash.ai/self-service/login/browser?aal=aal2` after a password-only login from a TOTP-enabled user. Kratos builds that URL from `SERVE_PUBLIC_BASE_URL` (set to the frontend origin) but no route on the frontend actually serves Kratos-native paths, so following it verbatim dead-ended on 404. The AAL2 form was never shown.

## 🔗 Related links

- https://linear.app/hash/issue/H-6219 _(internal)_

## 🚫 Blocked by

_None._

### Frontend
- `apps/hash-frontend/src/pages/settings/security.page.tsx`
  - Reactivates the TOTP enrolment / disable / backup-code UI block.
  - TOTP disable now also submits `lookup_secret_disable: true`. With `required_aal: highest_available` a leftover backup-code credential kept forcing AAL2 at next login with no TOTP to answer it.
  - Removes the pre-existing placebo "enter a current TOTP code to disable" step — Kratos v26 ignores `totp_code` for an already-enrolled identity, so it gated nothing. The disable form now shows a plain "yes, disable 2FA" confirmation and a warning that backup codes will also be removed.
  - Surfaces previously-silent failures on the clipboard-copy, backup-code regeneration, and "I've saved my codes" paths so the user sees an error instead of getting a half-broken account.
  - Backup-code extraction regex now accepts Kratos v26's no-dash 8-char codes (`XXXXXXXX`) in addition to the old `XXXX-XXXX` layout. Without this the UI rendered all 12 codes as one block and the AAL2-login test submitted the whole comma-separated string.
- `apps/hash-frontend/src/pages/shared/ory-kratos.ts`
  - New `flowMetadata` record mapping each Kratos self-service flow to `uiPath` + `kratosBrowserPath`.
  - New `uiPathForKratosBrowserRedirect` helper that rewrites Kratos-native `/self-service/*/browser` URLs to the matching frontend route.
- `apps/hash-frontend/src/pages/shared/use-kratos-flow-error-handler.ts`
  - `session_aal2_required`, `session_refresh_required`, `browser_location_change_required` now route through the rewriter before calling `router.replace`.
  - The old ad-hoc `router.push(\`/${flowType}\`)` fallbacks use `flowMetadata[flowType].uiPath`. Fixes a latent bug where `flowType: "registration"` would have pushed to a non-existent `/registration` route.
- `apps/hash-frontend/src/pages/signin.page.tsx`
  - Flow-reuse guard compares `flow.requested_aal` against the URL `?aal` param. Without this, the AAL1 flow loaded at `/signin` persisted through the client-side navigation to `/signin?aal=aal2` and the AAL2 form never rendered.
  - The two remaining raw `redirect_browser_to` follows (login-success `continue_with` branch, toSession 403 fallback) now go through `uiPathForKratosBrowserRedirect` too.

### Backend (API proxy)
- `apps/hash-api/src/index.ts`
  - Replaces the single `authRouteRateLimiter` on the Kratos `/auth/*` proxy with tiered limiters: non-GET credentials (30/10s), GETs other than whoami (60/10s), whoami exempted. The redundant whoamis the frontend emits during MFA flows were consuming the old 12/10s IP budget and surfacing as 429s.
  - `userIdentifierRateLimiter` only applies when `body.identifier` is actually present (its documented purpose), instead of falling back to IP.
  - Safe `ipKey` keyGenerator replaces the `req.ip!` assertion used by the previous default generator.

### Test harness
- `tests/hash-playwright/tests/shared/totp-utils.ts`: rewrites `decodeBase32` so the accumulator is truncated after every emitted byte. The old implementation overflowed JavaScript number precision past 53 bits and silently produced a half-zero key, so generated TOTP codes were rejected by Kratos.
- `tests/hash-playwright/tests/shared/get-kratos-verification-code.ts`: parses Mailslurper's timezone-less `dateSent` strings as UTC (they were being interpreted as local time, causing the 5-second window filter to drop every legitimate email on CEST machines).
- `tests/hash-playwright/tests/shared/signup-utils.ts`: adds a per-process suffix to test emails and shortnames. `resetDb()` is still a no-op and `POST /users/delete` intentionally preserves the user's web principal (so shortnames survive), which broke re-runs with the static names.
- `tests/hash-playwright/tests/shared/runtime.ts`: replaces the blanket status-code regex for the 401 tolerance with a URL-aware credit-counter. Each tolerated 4xx response on a specific Kratos path grants exactly one `Failed to load resource` console suppression, so real failures from other endpoints still flag.
- `tests/hash-playwright/tests/mfa.spec.ts`: unskips all five TOTP tests; the disable test additionally asserts the `regenerate-backup-codes-button` is gone so a regression in the `lookup_secret_disable` step would fail the assertion.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The `SERVE_PUBLIC_BASE_URL` misconfiguration that caused the redirect mismatch is worked around in the frontend rather than corrected at the source. Proper config fix tracked in H-6419.
- Disabling 2FA currently relies only on the logged-in privileged session — no explicit re-authentication. A proper SSO-compatible session-refresh step is tracked in H-6417 (same note applies to the password-change flow).
- SSO-only users cannot change their password today (pre-existing limitation of `handlePasswordSubmit`). Documented and tracked in H-6418.
- The test runtime's URL-aware 4xx tolerance consumes suppression credits by status code only, so a rare interleaving of a real and an expected failure at the same status can mask the real one. Acceptable trade-off against blanket global tolerance.

## 🐾 Next steps

- H-6417 — Forcible session refresh for sensitive settings changes (SSO-compatible)
- H-6418 — Password setup / reset flow for SSO-only users
- H-6419 — Fix Kratos \`SERVE_PUBLIC_BASE_URL\` / frontend-proxy mismatch

## 🛡 What tests cover this?

- `tests/hash-playwright/tests/mfa.spec.ts` — five E2E tests covering: enable TOTP, AAL2 login with TOTP, AAL2 login with backup code, disable TOTP, rejection of wrong code. All green locally.

## ❓ How to test this?

1. Check out the branch, deploy FE + BE together.
2. Sign in with an existing user, go to `/settings/security`, click "Enable TOTP".
3. Scan the QR code, enter the 6-digit code, save the generated backup codes.
4. Log out, log back in with password alone — confirm the AAL2 form appears, enter a TOTP code or toggle to backup-code mode.
5. From `/settings/security`, click "Disable TOTP" and confirm — log out, log back in with password; no AAL2 prompt should appear.

## 📹 Demo

_To add once manual staging verification is done._

### DIFF
--- a/apps/hash-frontend/src/pages/settings/security.page.tsx
+++ b/apps/hash-frontend/src/pages/settings/security.page.tsx
@@ -102,7 +102,6 @@ const SecurityPage: NextPageWithLayout = () => {
   const [isRecoveryFlow, setIsRecoveryFlow] = useState(false);
   const [currentPasswordError, setCurrentPasswordError] = useState<string>();
   const [totpCode, setTotpCode] = useState("");
-  const [disableTotpCode, setDisableTotpCode] = useState("");
   const [showTotpSetupForm, setShowTotpSetupForm] = useState(false);
   const [showTotpDisableForm, setShowTotpDisableForm] = useState(false);
   const [backupCodes, setBackupCodes] = useState<string[]>([]);
@@ -384,7 +383,7 @@ const SecurityPage: NextPageWithLayout = () => {
   ) => {
     event.preventDefault();
 
-    if (!flow || !disableTotpCode) {
+    if (!flow) {
       return;
     }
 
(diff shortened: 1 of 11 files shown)


apps/hash-frontend/src/pages/settings/security.page.tsx
```diff
@@ -537,222 +563,214 @@ const SecurityPage: NextPageWithLayout = () => {
 
           <Divider />
 
-          {/* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- @todo H-6219 restore this */}
```
TimDiekmann, line 540: This section was only commented in, no changes made.

CiaranMn (approved):

PR #8623, opened 2026-04-13
FE-583: better petrinaut text wrapping

alex-e-leon:

Improves the text-wrapping functionality in petrinaut nodes.
I've broken this PR into 2 commits. The first is what I think are uncontroversial css changes that simply handles multi-line overflows + ellipsis better in the classic nodes.
The second is a new approach of handling the pascal case splitting that behaves slightly differently to the current implementations and I'm open to throwing away if its not liked. I'll explain the implementation + differences with some examples below:

Screenshot 2026-04-13 at 5 13 25 pm
Screenshot 2026-04-13 at 5 12 50 pm

In the original version, we manually split pascal case into 2 lines. This means that if we have text that is too long, each individual line will ellipsify as in the first image. It also means that if we have a short pascal case it still splits into multiple lines.

In my new version, instead I add a zero width space between every pascal case character, which will then only break between pascal case characters if they do not fit on one line. If a very long word without a new pascal case section appears it uses overflow wrap to break anyway. Then uses line-clamp to keep the total max lines at 3 before ellipsifying.

The behaviour in most cases is identical, but how it handles very long words is slightly different.

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
+++ b/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
@@ -31,7 +31,3 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 
   return segments;
 };
-
-export const addPascalCaseBreakPoints = (str: string) => {
-  return splitPascalCase(str).join("\u200B");
-};
--- a/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
+++ b/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
@@ -31,3 +31,7 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 
   return segments;
 };
+
+export const addPascalCaseBreakPoints = (str: string) => {
+  return str.slice(0, 1) + str.slice(1).replaceAll(/[A-Z]/g, "\u200B$&");
+};
--- a/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
+++ b/libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
@@ -33,5 +33,5 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 };
 
 export const addPascalCaseBreakPoints = (str: string) => {
-  return str.slice(0, 1) + str.slice(1).replaceAll(/[A-Z]/g, "\u200B$&");
+  return splitPascalCase(str).join("\u200B");
 };
--- a/libs/@hashintel/petrinaut/src/views/SDCPN/components/classic-place-node.tsx
+++ b/libs/@hashintel/petrinaut/src/views/SDCPN/components/classic-place-node.tsx
@@ -4,7 +4,7 @@ import { use } from "react";
 import { TbMathFunction } from "react-icons/tb";
 
 import { hexToHsl } from "../../../lib/hsl-color";
-import { addPascalCaseBreakPoints } from "../../../lib/split-pascal-case";
+import { splitPascalCase } from "../../../lib/split-pascal-case";
 import { PlaybackContext } from "../../../playback/context";
 import { SimulationContext } from "../../../simulation/context";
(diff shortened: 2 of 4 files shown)


kube (dismissed):
That is a better solution :+1:

libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
```diff
@@ -31,3 +31,7 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 
   return segments;
 };
+
+export const addPascalCaseBreakPoints = (str: string) => {
+  return str.slice(0, 1) + str.slice(1).replaceAll(/[A-Z]/g, "\u200B$&");
```
kube: Please use splitPascalCase here, as it has the unit tests for where we want to split on the PascalCase.

kube (changes requested):

libs/@hashintel/petrinaut/src/views/SDCPN/components/classic-transition-node.tsx
```diff
@@ -75,6 +75,8 @@ const contentWrapperStyle = css({
   flexDirection: "column",
   alignItems: "center",
   gap: "3",
+  overflowWrap: "break-word",
+  lineClamp: "2",
```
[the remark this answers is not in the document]
alex-e-leon: Good catch - flex removed, its not being used.

libs/@hashintel/petrinaut/src/views/SDCPN/components/classic-place-node.tsx
```diff
@@ -74,20 +75,24 @@ const contentWrapperStyle = css({
...
+  padding: "[12px 0]",
   lineHeight: "[1.1]",
+  maxWidth: "[100%]",
 });
 
 const labelSegmentStyle = css({
   display: "inline-block",
-  whiteSpace: "nowrap",
+  overflowWrap: "break-word",
+  lineClamp: "3",
```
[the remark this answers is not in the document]
alex-e-leon: This is wrong, it applies the other css properties like  `display: -webkit-box;`

libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
```diff
@@ -31,3 +31,7 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 
   return segments;
 };
+
+export const addPascalCaseBreakPoints = (str: string) => {
```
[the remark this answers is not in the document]
alex-e-leon: Switched to use splitPascalCase under the hood

libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
alex-e-leon: Yeah, on reflection my pascal case regex was also not great, as it didn't handle all caps sections well. Switched to use splitPascalCase under the hood, but kept it as a seperate export.

libs/@hashintel/petrinaut/src/lib/split-pascal-case.ts
```diff
@@ -33,5 +33,5 @@ export const splitPascalCase = (pascalCaseString: string): string[] => {
 };
 
 export const addPascalCaseBreakPoints = (str: string) => {
-  return str.slice(0, 1) + str.slice(1).replaceAll(/[A-Z]/g, "\u200B$&");
+  return splitPascalCase(str).join("\u200B");
```
kube: Just last comment I promise, but given this is specific to the component now (defining how to render the break between two segments), this could just be inlined the component.

alex-e-leon: sure thing - no problem

alex-e-leon: Done

kube (approved):
:+1:

PR #8632, opened 2026-04-15
FE-585: Petrinaut – add hover + selected emphasis for nodes

alex-e-leon:

This PR adds some additional styling when hovering + selecting nodes to help focus on on the currently selected nodes and its connections. Note, because we are using context to manage global state, it means that every time a node is hovered/selected every node is re-rendering, even ones that haven't actually changed state. There's no real way around this without switching to a state library that has render optimizations built in (or building that ourselves, but this seems like overkill). My ideal approach to implementing this would have been to add an overlay, and use z-index to float the focused nodes above it, but react-flow sets its own stacking contexts and z-indexes across a couple layers making this approach impossible.

This PR:

The changes in this PR:


kube (approved):
:+1:

PR #8637, opened 2026-04-16
FE-593: pan viewport

alex-e-leon:

When opening a panel, pans the toolbar if there is a selected item that would be moved off screen after the panel opens.
I also adjust the min zoom to have a minimum value of 0.75 so that the min zoom when only 1 item is selected isn't horrible.

This PR:

The changes in this PR:


libs/@hashintel/petrinaut/src/lib/viewport.ts
```diff
@@ -0,0 +1,90 @@
...
+type Viewport = {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+};
+
+// returns the amount offscreen as a postive integer for each direction
+const getOffscreenAmount = (
+  reactFlowInstance: PetrinautReactFlowInstance,
```
kube, line 15: Please do not propagate `reactFlowInstance` everywhere.

These functions just do simple math, and should just receive something like:
- `{ x, y, width, height }`
- `{ x1, x2, y1, y2 }`

kube (approved):
I approve as it works, and it's cool to have this behaviour.

Though, we need to discuss that next week, as being able to determine dimensions and compute dimensions/positioning should be part of our "lib", not delegated to React Flow.

Dimensions and positions are things controlled by Petrinaut, React Flow only draws them for us, and adds interactivity on top of it.

Delegating that to React Flow adds an unnecessary overhead/indirection, that prevents us from computing anything before the component is entirely mounted, and requires a second render.

libs/@hashintel/petrinaut/src/lib/viewport.ts
alex-e-leon: Thanks. I'll merge it and we can revisit it and talk about it in person when I get to london. So that I don't forget, I'll quickly share my reasoning for doing it this way here as well.

First off, in general, I completely agree with you - since we control the data, we normally also want to control the operations on that data, and eliminating a re-render is always good.

However in this case I've gone with react flows solution because it is doing the actual drawing and so the positioning might not be 100% in sync with our data model. To illustrate, react-flow actually exports getNodeBounds directly, so originally I didn't pass the react-flow instance around at all, however when I did that it gave me a warning `Please use `getNodesBounds` from `useReactFlow`/`useSvelteFlow` hook to ensure correct values for sub flows. If not possible, you have to provide a nodeLookup to support sub flows.` (https://github.com/xyflow/xyflow/blob/a58568f11bc0e1a1bdca1b3549e959e2e1ca0cdd/packages/system/src/utils/graph.ts#L201)

Now we don't currently use subFlows, so its probably not an issue but it suggests to me that our model and react-flows model may not always be in sync - so if we do the measurement calculations ourselves, it might be different to whats actually drawn on the screen, leading to subtle errors. So I felt like it was safest to just use react-flows functions.

PR #8659, opened 2026-04-27
FE-623: DS cleanup

alex-e-leon:

- Removes storybook in favor of just using ladle in DS components and converts all stories to a valid Ladle format
- Removes the playground files that were used for refractive
- Removes the figma files from ds-components

Each of the above changes is in its own commit in this PR

This PR:

The changes in this PR:


vilkinsons:
@alex-e-leon @lunelson So no Figma Code Connect anymore?

alex-e-leon:
> @alex-e-leon @lunelson So no Figma Code Connect anymore?

I don't have a strong view either way, but if we're not using it (and the designers aren't using it) then it makes sense to clean it up.

Lu was intending to remove it and thinks its unnecessary, so maybe he can chime in.

lunelson (approved):
Nice one!

PR #8661, opened 2026-04-28
FE-620: DS icon component

alex-e-leon:

Adds an Icon component to the DS. Once I've created a button component and updated petrinaut to use this component I may need to come back and update some properties.

Because we are using font awesome's pro icon set we can't re-export (or import) all of font-awesomes Icon set, so instead I've manually copied over the subset that we use. We also won't be able to allow users to select an icon in our UI's so we'll need to ban that ui pattern somehow. 

## Pre-Merge Checklist 🚀

This PR:

The changes in this PR:


libs/@hashintel/ds-components/.ladle/types.d.ts
lunelson, line 1: This declaration is now required by package source code (`src/components/Icon/icon.tsx` imports `*.svg`), not just by the Ladle harness. Could we move the `*.svg` module declaration to a package-owned type file, e.g. `src/types/svg.d.ts` or a root `svg.d.ts`, so the library build contract is not owned by `.ladle`?

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
lunelson, line 1: Could we give these props a named exported `IconProps` type and export it from `src/main.ts` alongside `Icon`? Consumers of `@hashintel/ds-components` will likely need the prop type for wrappers, and an inline object type makes that harder to import or project from.

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
```diff
@@ -0,0 +1,207 @@
...
+import ThoughtBubble from "./svgs/light/thought-bubble.svg";
+import Trash from "./svgs/regular/trash.svg";
+import TriangleExclamation from "./svgs/solid/triangle-exclamation.svg";
+import UserPlus from "./svgs/regular/user-plus.svg";
+import WandMagicSparkles from "./svgs/regular/wand-magic-sparkles.svg";
+import Xmark from "./svgs/regular/xmark.svg";
+
+import { styles } from "./icon.recipe";
+
+export const sizes = ["xs", "sm", "md", "lg"] as const;
```
lunelson, line 91: This repeats the `size` variants defined in `icon.recipe.ts`. Could we make one of these the source of truth? For example, export a single `iconSizes` tuple used by the recipe and stories, or derive the prop type from the recipe variant type (preferred)

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
```diff
@@ -0,0 +1,207 @@
...
+}) => {
+  const IconSvg = IconMap[name];
+
+  return (
+    <IconSvg
+      className={cx(styles({ size }), className)}
+      aria-label={alt}
+      role={alt ? "img" : undefined}
+      aria-hidden={alt ? undefined : "true"}
+      viewBox="0 0 640 640"
```
lunelson: Do we need to override `viewBox` here? The imported SVG asset already owns its intrinsic viewBox. Hardcoding `0 0 640 640` works for the current Font Awesome set, but it makes the wrapper silently wrong if we add any non-640 asset later.

lunelson (changes requested):

lunelson:
So, FTR I must say I'm not convinced `svgr` should be our solution for integrating Font Awesome Pro icons in the DS. It provides good local ergonomics, but the SVG markup becomes part of the shipped JS and part of React’s rendered tree, which costs in bundle size, parse/compile work, and more React-managed markup and memory overhead

Personally I’d strongly prefer something like:

- maintain an explicit approved icon manifest;
- generate subset sprite files containing only icons actually used by the product;
- expose a very small `Icon` component that renders `<svg><use href="...#icon-name" /></svg>`;
- style via panda tokens / `currentColor`, not icon-specific component APIs

lunelson:
Some context: [Breaking Up with SVG-in-JS](https://kurtextrem.de/posts/svg-in-js)

libs/@hashintel/ds-components/.ladle/types.d.ts
alex-e-leon: Good idea!

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
alex-e-leon: Can't consumers do `React.ComponentProps<typeof Icon>`?

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
alex-e-leon: The reason I've broken this out is because size is actually going to be a shared type between a couple different components. The idea is effectively that icons, buttons + inputs all need to share the same size api, and have their sizes visually aligned as well so that they compose nicely together.

You can see that I've actually already done this in the LoadingSpinner PR:
https://github.com/hashintel/hash/pull/8664/changes#diff-b610fc5161a73a4c6249a821e7324e4582ca89348760ef76508613929cfc7474

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
lunelson: They can for sure, it's a DX policy issue I guess

lunelson: Not major

libs/@hashintel/ds-components/src/components/Icon/icon.tsx
alex-e-leon: Good call - I'll remove it.

alex-e-leon:
> So, FTR I must say I'm not convinced `svgr` should be our solution for integrating Font Awesome Pro icons in the DS. It provides good local ergonomics, but the SVG markup becomes part of the shipped JS and part of React’s rendered tree, which costs in bundle size, parse/compile work, and more React-managed markup and memory overhead
> 
> Personally I’d strongly prefer something like:
> 
> * maintain an explicit approved icon manifest;
> * generate subset sprite files containing only icons actually used by the product;
> * expose a very small `Icon` component that renders `<svg><use href="...#icon-name" /></svg>`;
> * style via panda tokens / `currentColor`, not icon-specific component APIs

I 100% agree - I plan on implementing spriting down the line.

I've used svgr for now because its much simpler to setup and get working, while I'm a little uncertain how the pipeline for spriting will work for consumers right now.

My plan was once I've finished a couple of these components and got a consumer (likely petrinaut) actually consuming the icons properly to then come back and implement spriting so that I have a real consumer to test the spriting pipeline.

alex-e-leon:
> * maintain an explicit approved icon manifest;

What exactly do you mean by a manifest?

> * style via panda tokens / `currentColor`, not icon-specific component APIs

Agree - this is how the icon component works at the moment.

lunelson:
> * What exactly do you mean by a manifest?

Just a central definition of which icons should go in to the sprite

lunelson (approved):

PR #8650, opened 2026-04-21
BE-515: HashQL: Remove `OFFSET 0` materialization fence from generated query

indietyp:
## 🌟 What is the purpose of this PR?

# HashQL: Remove `OFFSET 0` materialization fence from generated SQL

## Problem

Each PostgreSQL island is wrapped in a `CROSS JOIN LATERAL` subquery. The subquery previously included `OFFSET 0` as a materialization fence to prevent PostgreSQL's planner from inlining the CASE tree expression and duplicating it across each composite field access in the SELECT list.

The planner's cost estimate favors the fenced plan (lower estimated cost), but empirical benchmarks show the opposite for queries with conditional CASE trees:

| Query | Change (fenced → unfenced) |
| -- | -- |
| all_entities | −11% (improved) |
| filter_by_uuid | +10% (regressed) |
| diamond_cfg | −5% (improved) |
| sequential_filters | −72% (improved) |
| stress_test | −77% (improved) |

For multi-body queries with conditional branches, the materialization overhead exceeds the duplication cost by 3–4×. The fence helps modestly for simple single-expression queries without branching (filter_by_uuid), but is catastrophic for the common case.

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Existing test coverage for the HashQL eval Postgres module.

## ❓ How to test this?

1. Checkout the branch.
2. Run the existing test suite.
3. Confirm all tests pass and query behavior is unchanged.

PR #8587, opened 2026-03-26
BE-474: HashQL: Take into account terminator target eligibility in execution backend placement

indietyp:
## 🔗 Related links

- ...

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

-

## ❓ How to test this?

1. Checkout the branch / view the deployment
2. Try X
3. Confirm that Y

PR #8501, opened 2026-03-03
BE-429: HashQL: Add island dependency graph with data requirement resolution

indietyp:
## 🌟 What is the purpose of this PR?

This PR implements the island dependency graph construction and scheduling system for the HashQL MIR execution engine. It builds a directed graph over computation islands, resolves data requirements between islands, and computes a topological schedule with parallelism levels for execution.

## 🔗 Related links

- Related to the island placement and execution planning work

## 🔍 What does this change?

- **Adds `first_set()` method to `FiniteBitSet`** - Returns the first set bit using `trailing_zeros()`, with comprehensive test coverage for empty sets, single bits, multiple bits, and wide integral types
- **Enhances `IdVec` with new utility methods** - Adds `from_raw()`, `from_domain_derive()`, `extend_from_slice()`, `append()`, `into_iter_enumerated()`, and `copy_within()` methods with detailed documentation and examples
- **Implements island dependency graph (`IslandGraph`)** - Creates a directed graph over `IslandNode`s connected by three edge types: `ControlFlow` (execution ordering), `DataFlow` (data dependencies), and `Inherits` (path inheritance between same-target dominators)
- **Adds data requirement resolution** - Automatically inserts synthetic data islands when traversal paths cannot be satisfied by upstream providers, using dominator analysis to find optimal providers
- **Implements island scheduling (`IslandSchedule`)** - Computes topological ordering with parallelism levels using Kahn's algorithm, allowing islands at the same level to execute concurrently
- **Adds comprehensive test suites** - Tests cover single islands, data edges between predecessors, fetch island insertion, diamond CFG patterns, inheritance edges, data island reuse, and scheduling correctness
- **Enables `option_into_flat_iter` feature** - Adds the unstable feature flag for iterator functionality

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## 🛡 What tests cover this?

- Unit tests for `FiniteBitSet::first_set()` covering empty sets, single/multiple bits, edge cases
- Integration tests for island graph construction with various CFG patterns
- Tests for data requirement resolution including fetch island insertion and reuse
- Schedule computation tests verifying topological ordering and parallelism levels
- Tests for edge deduplication and dominator-based inheritance

[...]

PR #8675, opened 2026-04-29
BE-523: HashQL: Suspendable interpreter with structured values and input system

indietyp:

Rework the MIR interpreter to support suspend/resume execution, richer value representations, and a proper input system. Together these let the interpreter pause when it needs external data (e.g. a database query), hand back a structured description of what it needs, and resume once the caller provides the result.

## 🔍 What does this change?

**Suspension model** (`interpret/suspension/`)
- Introduce a suspend/resume protocol: when the interpreter hits a `GraphRead` terminator it yields a `Suspension` instead of blocking. Callers drive a `start`/`resume` loop until they get a `Return`.
- `GraphReadSuspension` extracts temporal axes, filter parameters, and entity type from the suspended state.
- `Continuation` carries the response value that resumes execution.
- `TemporalAxesInterval` / `Timestamp` give structured representations of the temporal query window.
- `Runtime::run` convenience wrapper for callers that can fulfill suspensions synchronously.

**Input system** (`interpret/inputs.rs`)
- Add `Inputs<'heap, A>`: a typed map from interned symbols to `Value`s, consulted for `InputOp::Load` and `InputOp::Exists`. Replaces the unstructured `FastHashMap` previously threaded through the runtime. Supports both global and custom allocators.

**Value representation** (`interpret/value/`)
- `Int`: carry bit-width (1-bit boolean vs 128-bit integer) so serialization can distinguish `true`/`false` from `0`/`1` without external type info. Arithmetic promotes to 128-bit; bitwise boolean ops preserve 1-bit.
- `Struct`: add `StructBuilder<N>` for stack-allocated fixed-field construction. Enforce sorted field invariant. Add `merge` and `project_by_name`.
- Minor improvements to `Opaque`, `List` allocator parameterization, `Str` display.

**SSA repair fix** (`pass/transform/ssa_repair`)
- Fix block parameters not being renamed during SSA repair.
- Extend use-before-def detection to terminators, not just statements.

**Entity path type resolution** (`pass/execution/traversal`)
- Add `field_path` and `resolve_type` to `EntityPath` for navigating from an entity path (e.g. `WebId`, `Properties`) to its concrete stdlib type.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

_None_

## 🐾 Next steps

- The orchestrator in `hashql-eval` consumes the suspension model and input system introduced here.

## 🛡 What tests cover this?

- Interpreter unit tests (`interpret/tests.rs`): cover the suspend/resume loop, input resolution, and value operations.
- SSA repair UI tests (`tests/ui/pass/ssa_repair`): cover the block parameter and terminator fixes.
- Existing pass UI tests re-blessed for snapshot changes from the `Int` representation change.

## ❓ How to test this?

1. `cargo nextest run --package hashql-mir`
2. `cargo test --package hashql-mir --doc`


indietyp:
* **#8827** <a href="https://app.graphite.com/github/pr/hashintel/hash/8827?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8697** <a href="https://app.graphite.com/github/pr/hashintel/hash/8697?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8696** <a href="https://app.graphite.com/github/pr/hashintel/hash/8696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8694** <a href="https://app.graphite.com/github/pr/hashintel/hash/8694?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8693** <a href="https://app.graphite.com/github/pr/hashintel/hash/8693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8650** <a href="https://app.graphite.com/github/pr/hashintel/hash/8650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8635** <a href="https://app.graphite.com/github/pr/hashintel/hash/8635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8607** <a href="https://app.graphite.com/github/pr/hashintel/hash/8607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8600** <a href="https://app.graphite.com/github/pr/hashintel/hash/8600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8595** <a href="https://app.graphite.com/github/pr/hashintel/hash/8595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8587** <a href="https://app.graphite.com/github/pr/hashintel/hash/8587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8586** <a href="https://app.graphite.com/github/pr/hashintel/hash/8586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8675** <a href="https://app.graphite.com/github/pr/hashintel/hash/8675?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/hashintel/hash/8675?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8674** <a href="https://app.graphite.com/github/pr/hashintel/hash/8674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8526** <a href="https://app.graphite.com/github/pr/hashintel/hash/8526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8525** <a href="https://app.graphite.com/github/pr/hashintel/hash/8525?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8524** <a href="https://app.graphite.com/github/pr/hashintel/hash/8524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8501** <a href="https://app.graphite.com/github/pr/hashintel/hash/8501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8498** <a href="https://app.graphite.com/github/pr/hashintel/hash/8498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8496** <a href="https://app.graphite.com/github/pr/hashintel/hash/8496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8495** <a href="https://app.graphite.com/github/pr/hashintel/hash/8495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libs/@local/hashql/mir/src/interpret/suspension/graph_read.rs
```diff
@@ -0,0 +1,134 @@
...
+fn extract_timestamp<'heap, E, A: Allocator>(
+    value: &Value<'heap, A>,
+) -> Result<Timestamp, RuntimeError<'heap, E, A>> {
+    let Value::Opaque(opaque) = value else {
+        return Err(RuntimeError::UnexpectedValueType {
+            expected: TypeName::terse("Opaque"),
+            actual: value.type_name().into(),
+        });
+    };
+    debug_assert_eq!(opaque.name(), sym::path::Timestamp);
```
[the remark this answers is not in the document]
indietyp: typchk

libs/@local/hashql/mir/src/pass/transform/inst_simplify/mod.rs
```diff
@@ -252,42 +252,32 @@ impl<'heap, A: Allocator> InstSimplifyVisitor<'_, 'heap, A> {
...
-            BinOp::Lte => i128::from(lhs <= rhs),
-            BinOp::Gt => i128::from(lhs > rhs),
-            BinOp::Gte => i128::from(lhs >= rhs),
+            BinOp::Add => return lhs.checked_add(rhs),
+            BinOp::Sub => return lhs.checked_sub(rhs),
+            // Bitwise ops preserve bool provenance via the Int operators
+            BinOp::BitAnd => return Some(lhs & rhs),
+            BinOp::BitOr => return Some(lhs | rhs),
+            // Comparisons produce booleans
+            BinOp::Eq => lhs.as_int() == rhs.as_int(),
```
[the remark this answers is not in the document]
indietyp: fixed upstream

libs/@local/hashql/mir/src/interpret/suspension/graph_read.rs
```diff
@@ -0,0 +1,134 @@
...
+            expected: TypeName::terse("Opaque"),
+            actual: value.type_name().into(),
+        });
+    };
+    debug_assert_eq!(opaque.name(), sym::path::Interval);
+
+    let value = opaque.value();
+
+    let start = value.project_by_name(sym::start)?;
+    let end = value.project_by_name(sym::end)?;
```
thehabbos007, line 66: Just curious here, for struct values, the inner struct is essentially `{fields: [], values: []}`  right? And project by name searches through the fields for a sym eq match, and then the value index is returned from  the values field, is that correct?

libs/@local/hashql/mir/src/interpret/suspension/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Writes the continuation's result value into the target block's parameter
+    /// and advances the frame to that block.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`RuntimeError::CallstackEmpty`] if the call stack has no frames.
+    pub fn apply<E>(
+        self,
+        callstack: &mut CallStack<'ctx, 'heap, A>,
+    ) -> Result<(), RuntimeError<'heap, E, A>> {
```
thehabbos007, line 94: I was about to ask how the contiuation is applied to the runtime, but this is a very direct way to discover/understand it. Neat!

libs/@local/hashql/mir/src/interpret/suspension/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+                        current_block.block.terminator.kind,
+                        TerminatorKind::GraphRead(_)
+                    );
+                }
+
+                let next_block = &frame.body.basic_blocks[read.target];
+                let params = next_block.params;
+                debug_assert_eq!(params.len(), 1);
+
+                frame.locals.insert(params[0], value);
```
thehabbos007, line 119: Just curious, can these insertions be expensive when we have many locals? or is it always constant time?

libs/@local/hashql/mir/src/interpret/value/int.rs
```diff
@@ -13,442 +35,235 @@ use crate::{
...
-/// assert_eq!(negative.as_u8(), None);
-///
-/// // Raw value access always succeeds
-/// assert_eq!(large.as_int(), 1000);
+/// // Bool provenance is metadata, not identity: from(true) == from(1)
+/// assert_eq!(Int::from(true), Int::from(1_i32));
 /// ```
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+// Uses `#[repr(packed)]` to avoid alignment padding, which would duplicate size, same as
+// rust-lang's ScalarInt.
```
thehabbos007, line 68: So we get a 17 byte struct size because of the repr, right? And no 15 byte alignment padding. Does this cause performance problems because of almost guaranteed unaligned access and potential false sharing?

thehabbos007 (commented):
Just leaving partial review so you don't get mega comment block

libs/@local/hashql/mir/src/interpret/value/int.rs
indietyp: No, it's actually a performance gain, funnily enough. We rarely ever access the size field, and mostly just operate on the value, so it is faster. I introduced the packed repr here when I was chasing benchmarking to get it as down as possible, but I'd be interested in a general benchmark for this, tracked in https://linear.app/hash/issue/BE-569/hashql-benchmark-packed-vs-unpacked-repr-for-int

libs/@local/hashql/mir/src/interpret/suspension/mod.rs
indietyp: It's always constant time. Underneath it's a `Vec<Option<T>>` so it grows when required. It's a pseudo hashmap.

libs/@local/hashql/mir/src/interpret/value/int.rs
```diff
@@ -13,442 +35,235 @@ use crate::{
...
     #[must_use]
-    pub const fn as_i64(self) -> Option<i64> {
-        if self.value >= i64::MIN as i128 && self.value <= i64::MAX as i128 {
-            Some(self.value as i64)
-        } else {
-            None
+    pub const fn as_bool(self) -> Option<bool> {
+        if !self.is_bool() {
+            return None;
         }
```
thehabbos007, line 176: Sort of a semantic question. The `as_int`, `as_uint` both seem to be a sort of infalible cast. Is there a reason why as_bool couldn't be a similar cast semantically? I'm just thinking that if i am able to go from bool to uint, but not uint back to bool, it seems like we are drawing the line at lossy/unlossy conversions. Should there then be a way to losslessly (if that's a word) cast to bool?

libs/@local/hashql/mir/src/interpret/suspension/graph_read.rs
indietyp: depends, it's faster than binary lookup because most structs are very small, one future possibility is to allow for numeric indexing during MIR lowering on closed structs, but I never captured that as a ticket: https://linear.app/hash/issue/BE-570/hashql-lower-to-positional-access-over-closed-structs

libs/@local/hashql/mir/src/interpret/value/int.rs
indietyp: ah yes, correct. There's a reason for that: you can see that in what was there previously `as_X` for types are fallible and inspect the value/size itself, whereas `as_int` and `as_uint`, are moreso an escape hatch, they circumvent the machinery and instead return just the raw underlying value, hence the name `as_int` instead of `as_u128`. Does that make sense?

libs/@local/hashql/mir/src/interpret/value/int.rs
```diff
@@ -633,14 +504,14 @@ impl Neg for Int {
...
-        let (value, overflow) = self.as_int().overflowing_neg();
+        let value = self.as_int();
+        let (result, overflow) = value.overflowing_neg();
 
-        if overflow {
-            // There's only a single reason why this overflowed: the value was i128::MIN, in this
-            // case we return `i128::MAX + 1` as a Num.
+        if hint::unlikely(overflow) {
+            // Only i128::MIN overflows: return i128::MAX + 1 as float.
             Numeric::Num(Num::from((i128::MAX as f64) + 1.0))
```
thehabbos007, line 512: This is a fun trick. Why does this work? wow lol

libs/@local/hashql/mir/src/interpret/suspension/graph_read.rs
thehabbos007: Makes sense, yeah the structs are probably so small these checks shouldn't be an issue, just checking :)

thehabbos007 (commented):
And the last comment :)

libs/@local/hashql/mir/src/interpret/value/int.rs
thehabbos007: That's so interesting. Would a Vec of these also be unpadded in memory? Or is the packed repr just for the bytes within the struct?

libs/@local/hashql/mir/src/interpret/value/int.rs
indietyp: This issues a hint to LLVM, in turn this influences the instructions that are generated, so that the branch predictor has an easier time. This gave me like 20% in the fibonacci benchmark for some reason.

libs/@local/hashql/mir/src/interpret/value/int.rs
indietyp: That would be AFAIK still padded, we mostly use it to keep size of the `Value` down as well, because it means the enum has "more space to work with", and the actual load destination (aka the size) is rarely ever inspected.

thehabbos007 (approved):

PR #8687, opened 2026-05-01
FE-647: Add `button` component

alex-e-leon:

- Adds Button component to the DS.
- Adds a new loading spinner variation to be used in Button.
- Updates Button usage in petrinaut to match the new api.

There are more custom buttons in petrinaut that we can port over to this new Button - including all IconButtons, but I'll do that in a follow up PR to keep the PR size more manageable.

I'll also be looking to do a design review with CiaranH. Depending on the discussion there may be follow on updates.

## 🔗 Related links

[Designs](https://www.figma.com/design/WmnosvOvi4Blw0HK0jh1mG/Design-System?node-id=42292-58967&t=PpEthB0HTO70ohxv-0)

This PR:

The changes in this PR:


libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -0,0 +1,734 @@
...
+  },
+  variants: {
+    size: {
+      xs: {
+        button: {
+          paddingX: "2",
+          paddingY: "0",
+          borderRadius: "md",
+          textStyle: "xs",
+          minWidth: `[calc(1em * ${1.6} * var(--leading-factor, 1) + 2px)]`,
```
alex-e-leon: @lunelson - this line needs to be fixed, and wondering if you have any suggestions.

Background:
Buttons of all sizes should have:
- consistent heights
- consistent alignment of children/icons/text inside them
- support dynamic text content, and break cleanly to multiple lines if the text content is too long, especially to support translations
- ideally have their height defined intrinsically so that if we change our font stack/spacing tokens (required if we want to support theming or white-labeling with different fonts) everything works.

In order to support all these I've used a trick: I add a zero-width space before any icons inside the buttons. This means that regardless of the icon height the alignment + height of the buttons always matches the text line-height.

The final part to bring this together is to set a min-width based on the line-height so that icon only buttons will always look square.

Which brings us finally to the problem here: the calc function here should be `calc(var(--text-xs-line-height) + var(--x-padding) * 2 + var(--border-width) * 2)` - But I can't figure out a nice way to get the equivalent of `var(--text-xs-line-height)`.

I tried to export it out of tokens as a seperate token, but since its value itself is a `calc` function, panda-css won't compose a calc inside a calc (though it is valid css). So for now I've just copied the calc function used for leading here instead, but this will break if we change our font-stack down the line.

Maybe an option is to switch the token from a calc function to straight JS?
Thoughts?

libs/@hashintel/ds-components/src/components/Button/button.tsx
```diff
@@ -1,434 +1,230 @@
...
+  };
+
+  if ("href" in props) {
+    return (
+      <a
+        {...sharedProps}
+        {...(sharedEventHandlers as React.DOMAttributes<HTMLAnchorElement>)}
+        ref={(props as AnchorElementOnlyProps).ref}
+        href={href}
+        target={props.target}
```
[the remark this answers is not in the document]
alex-e-leon: This is not true any more in modern browsers

libs/@hashintel/ds-components/src/components/Button/button.tsx
```diff
@@ -1,434 +1,230 @@
...
-          opacity: isLoading ? 0 : 1,
-        })}
+        <span className={classes.loadingContent}>{content}</span>
+      </>
+    );
+  }
+
+  const sharedProps = {
+    className: cx(classes.button, className),
+    title: tooltip,
```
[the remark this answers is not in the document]
alex-e-leon: This is valid, but the plan is to switch away from title to an actual tooltip component once the tooltip is built.

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -0,0 +1,766 @@
...
+      },
+      error: {
+        button: {
+          "&:focus-visible": {
+            outlineColor: "red.a60",
+          },
+        },
+      },
+    },
+    variant: {
```
lunelson, line 112: The panda way here, is to use the ["virtual color"](https://panda-css.com/docs/docs/concepts/virtual-color) mechanism, and just use the `colorPalette` special property within the recipe. In the component then, your `tone` prop would accept a regular or semantic color key, and pass that through to the `colorPalette` prop on button element, which means it has to be a `styled(button)` component so it can accept this prop

This should spare you from ever having to write out `tone` variations inside the recipe

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -0,0 +1,766 @@
...
+  },
+  variants: {
+    size: {
+      xxs: {
+        button: {
+          paddingX: "2",
+          paddingY: "[1px]",
+          borderRadius: "md",
+          textStyle: "xxs",
+          minWidth: `[calc(1em * ${1.6} * var(--leading-factor, 1) + 1px * 2 + var(--button-border-width) * 2)]`,
```
lunelson, line 40: This seems very complicated, why not a simple em value? Also, why is it using the `--leading-factor` variable in a horizontal dimension?

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -0,0 +1,766 @@
...
+      },
+    },
+    {
+      size: "lg",
+      hasIconLeft: true,
+      css: {
+        iconText: { marginLeft: "2" },
+      },
+    },
+    {
```
lunelson: This would be more compact if you defined a `--icon-margin` within `size` variants, and consumed that here: you would need only two compound variants, one for `hasIconLeft` and one for `hasIconRight`

OTOH couldn't you also place the icons and iconText in a single flex row and specify a gap value? Then you wouldn't need the compoundVariants and the variable could be `--icon-gap`, for example

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -0,0 +1,766 @@
...
+      tone: "neutral",
+      isPressed: true,
+      css: {
+        button: {
+          background: "neutral.s115",
+          borderColor: "neutral.s115",
+          boxShadow: "[inset 0 2px 4px rgba(0,0,0,0.35)]",
+        },
+      },
+    },
```
lunelson, line 245: All of this color assignment is encoded in to the semantic color aliases, and shouldn't need to be laid out again here by variant/tone combination, only by variant.

aliases like `"colorPalette.bg.solid"` and `"colorPalette.bg.solid.active"` for example, would work in this case

It's possible not all of those assignments are correct, but the move is then to edit the assignments so they are correct, not re-specify them in each component

Check `libs/@hashintel/ds-components/src/preset/tokens/utils.ts:43~159`

lunelson (changes requested):

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Do you know if its possible to use virtual colors AND still overwrite with specific colors under specific situations?

The issue is that the color mappings aren't always 1:1 in my implementation (or the figma designs) - and I think this is what we want.

Different colors that are the same lightness don't always have the same perceptual weight - as an example the brand subtle hover effect sets the background to blue.a30. while the error one sets the background to red.a25. Lining them up to the same value makes the hover effect in one case feel a little strong/weak.

So we could simplify the logic out here with virtual colors, but I think the designs would not be as strong for it. But if I can use virtual colors AND still overwrite, then that definitely feels a lot cleaner.

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: See my comment here https://github.com/hashintel/hash/pull/8687#discussion_r3172984593

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Yeah good point around simplifying with css variables - will update.
Re using flex + a gap: because I'm adding a zero-width space for alignment (again see my old comment https://github.com/hashintel/hash/pull/8687#discussion_r3172984593 for why) adding a gap would create extra space after the zero-width space, so its just easier to use margin instead.

Also note that there's more margin/padding adjustments on their way from the designers to visually align the end paddings for icon combinations. IE its eventually going to look something like this M1-Icon-M2-Text-M3 Where M1 and M2 + M3 are all different spacings

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Yeah agreed -  the issue is that they aren't correct. I tried using them initially but the values were quite far off. I think updating them should come as a separate piece of work. I'd also need to double check that those values aren't being used anywhere else before updating them.

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: Right, OK I see the trick but I would rather just use a fixed height, for the reason that buttons, inputs, toggles, and other input-/button-like elements should have compatible heights across `size` variants, so that if they are placed in a line they look aligned

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: You can still overwrite yes, with a single `compoundVariant`; but for a half-step difference like a25 vs a30 I would push back to the designers personally

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: > IE its eventually going to look something like this Where M1 and M2 + M3 are all different spacings

What's M1 M2 and M3? we have a `--spacing-factor` custom prop for themable spacing as we have a `--roundness-factor` for themable roundness IIRC

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: No, I think that work should be folded in here: the point was to correct those assignments *in practice when building the components*, since that's the ideal time to validate if they are correct. Deferring it leaves existing debt and creates new debt on top

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: M1, M2 and M3 are just made up shorthands for different spacing values - they don't map to any tokens or anything I was just trying to use text to explain that the spacing values are going to be different in all 3 of those places

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Fair - how about I update them in a new PR after this one to keep the reviews simpler + easier and that functionally its going to be the same output?

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: 100% - I plan on making sure that they're aligned - just rather than aligning by making them all share the same height variable they'll be aligned by sharing the same line-height/padding/border values. I'll also be creating a story that includes them all so that it's easy to visually check that the heights stay aligned.

In my experience doing it this way is slightly more complex, but it also is much much more robust during usage and prevents a lot of ui issues - especially once theming or translations are brought in.

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: Ah I see. Well probably they could still be CSS props like `--outer-margin-icon`, `--outer-margin-text`, `--inner-margin` in that case, and be defined in the `size` variants. Would still reduce repetition

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
lunelson: In this case sure, but for future components let's not separate them that's just more PRs

alex-e-leon: I'm also expecting some of these values to change over the next couple of days once ciearanH completes his full design review - but sure in the future I'll wrap these kinds of changes into a single PR.

lunelson: Yeah subsequent changes should be fine, and should also be easier since they would require fewer changes

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Done

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: Updated so that the line-height is calculated in js, and using that value directly now.

libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
alex-e-leon: I'm going to wrap this one in the follow on PR that corrects the semantic color layer, since it makes more sense to do it together.

lunelson (changes requested):
A couple more comments on TS patterns

libs/@hashintel/ds-components/src/preset/tokens.ts
```diff
@@ -85,7 +81,18 @@ export const tokens = {
...
+    xxs: { value: "1.6em" },
+    xs: { value: "1.6em" },
+    sm: { value: "1.6em" },
+    base: { value: "1.5em" },
+    lg: { value: "1.5em" },
+    xl: { value: "1.4em" },
+    "2xl": { value: "1.4em" },
+    "3xl": { value: "1.3em" },
+    "4xl": { value: "1.3em" },
+  },
```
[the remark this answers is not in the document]
lunelson: Yes please restore this

lunelson (changes requested):
I had a lot of comments about the button.tsx file and the type declarations but Github is acting up again so maybe we can address on the next pass

libs/@hashintel/ds-components/src/preset/tokens.ts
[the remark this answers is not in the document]
alex-e-leon: Sure thing - reverted.

alex-e-leon:
> I had a lot of comments about the button.tsx file and the type declarations but Github is acting up again so maybe we can address on the next pass

> A couple more comments on TS patterns

@lunelson Looks like github may have eaten your comments - the only one I saw since the previous rounds was the one about reverting the logic for line-height calculations which I've actioned.

libs/@hashintel/ds-components/scripts/figma-variables.json
lunelson, line 1: just FYI this was originally exported from figma... which is now probably a flow that should be done differently, if we don't want automatic updates in that direction

lunelson (dismissed):

lunelson (dismissed):

lunelson (dismissed):

TimDiekmann (approved):

PR #8699, opened 2026-05-05
FE-689: Add tooltip component to DS

alex-e-leon:

Adds a tooltip component to the DS based on ark-ui. In addition to the tooltip features provided by ark-ui we also add logic to handle when a tooltip wraps both focusable and non-focusable elements - adding focus handlers when wrapping non-focusable content.

## Pre-Merge Checklist 🚀

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.tsx
+++ b/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.tsx
@@ -1,0 +1,38 @@
+import type { RequireExactlyOne } from "type-fest";
+
+type Direction = "bottom" | "top" | "left" | "right";
+type Position = Direction | `${Direction}-${'start' | 'end'}`;
+
+export const Tooltip = ({
+  ...props,
+}: {
+  className?: string;
+  /** The tooltip trigger */
+  children: React.ReactNode;
+  /** The content that triggers the tooltip */
+  content: React.ReactNode;
+  /** The preferred position of the tooltip - depending on the viewport, trigger and content another position may be chosen for better fit */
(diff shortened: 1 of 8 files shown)


lunelson (commented):
I guess my first question is why didn't you adapt the tooltip in `libs/@hashintel/ds-components/src/beta/tooltip/`?

alex-e-leon:
> I guess my first question is why didn't you adapt the tooltip in `libs/@hashintel/ds-components/src/beta/tooltip/`?

Do you mean why didn't I use the the exact same tooltip props exposed in that component? Or why didn't I use their stories? Or something else?

libs/@hashintel/ds-components/src/components/Tooltip/tooltip.tsx
lunelson, line 1: Can you explain why this implementation differs so completely from [this one](https://github.com/hashintel/hash/blob/main/libs/%40hashintel/ds-components/src/beta/tooltip/tooltip.tsx)?

alex-e-leon: Mostly its logic that handles focus states for different tooltip children. In the version you linked if you wrap a plain string in a tooltip the string never renders - you'd have to know to wrap the string in a span. Also if you wrap an element that isn't focusable like an icon the tooltip isn't accessible through focus - again you would have to know to manually configure your own focus handlers on the child.

I wrote logic to fix all those issues so that tooltip works properly across all those kinds of elements out of the box.

lunelson (approved):

PR #8708, opened 2026-05-08
FE-627: Update tsgo/vite/rolldown + better React hooks linting in Petrinaut

kube:

Update the Petrinaut frontend toolchain and linting after extracting the library.

This is just a temporary update to have React Hooks linting in Petrinaut and Refractive (the only packages using Oxlint at the moment), and is not meant to be the final config every package will have to use once we migrate all packages to Oxlint.

## 🔍 What does this change?

- Upgrades Petrinaut, Refractive, and the Petrinaut website dev tooling around tsgo, Vite, Rolldown, Oxlint, and related packages.
- Adds shared Oxlint config for React Hooks, React Compiler diagnostics, and React Refresh checks.
- Fixes Petrinaut React Hooks exhaustive-deps issues surfaced by the new config.
- Updates `@apps/petrinaut-website` `main/app.tsx` to keep one active document handle, persist localStorage metadata correctly, and hide the ephemeral default net from the Open menu.
- Updates `@hashintel/petrinaut` `sdcpn-view.tsx` to use `useEffectEvent` so the load-time fit reads the latest `minZoom` without refitting on every `minZoom` change.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Follow up on React Refresh `only-export-components` refactors in Petrinaut.

## 🛡 What tests cover this?

- `yarn fix:eslint`
- `yarn lint:tsc`

## ❓ How to test this?

1. Check CI.
2. Run the Petrinaut lint/typecheck scripts locally.


libs/@hashintel/petrinaut/src/ui/views/SDCPN/sdcpn-view.tsx
```diff
@@ -116,7 +116,7 @@ export const SDCPNView: React.FC<{
       minZoom: minZoom,
       maxZoom: 1.1,
     });
-  }, [reactFlowInstance, petriNetId]);
+  }, [minZoom, reactFlowInstance, petriNetId]);
```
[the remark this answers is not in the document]
kube: Fixed by add usage of `useEffectEvent`.

alex-e-leon (dismissed):
Haven't run/tested the changes locally, but as far as I can tell from a quick skim it all looks good.

If you preferred not to include/write your own oxlint rule .json file for react-compiler though this project has all the eslint-recommended packs transformed for oxlint:
https://github.com/popup-plus/oxlint-config-presets

kube:
> Haven't run/tested the changes locally, but as far as I can tell from a quick skim it all looks good.
> 
> If you preferred not to include/write your own oxlint rule .json file for react-compiler though this project has all the eslint-recommended packs transformed for oxlint: https://github.com/popup-plus/oxlint-config-presets

@alex-e-leon Yeah would definitely be better to use existing configs.
I don't have the time to do these changes right now, as I know the choice of the preset will be subject to debate.

If you want to create a PR for this go ahead!

.config/oxlint/react-compiler.json
```diff
@@ -0,0 +1,31 @@
...
+  "$schema": "../../node_modules/oxlint/configuration_schema.json",
+  "jsPlugins": [
+    {
+      "name": "react-hooks-js",
+      "specifier": "eslint-plugin-react-hooks"
+    }
+  ],
+  "rules": {
+    "react/rules-of-hooks": "off",
+    "react/exhaustive-deps": "off",
```
lunelson: why not leave these on? they are the same rule and will run faster if you use the oxlint version, no?

alex-e-leon:
> @alex-e-leon Yeah would definitely be better to use existing configs. I don't have the time to do these changes right now, as I know the choice of the preset will be subject to debate.
> 
> If you want to create a PR for this go ahead!

For eslint-plugin-react-hooks there's only really one preset option to choose from which is `react-hooks/recommended` so I think that should prevent any debate : ) (technically there is also `react-hooks/recommended-latest` but I think we can all agree we don't want to be guinea pigs for rules that are still experiments)

Happy to do it some time later as well though - totally understand if you don't have time to action right now - neither do I! : )

.config/oxlint/react-compiler.json
kube: Updated.

I asked an agent to check quickly and they should be the same indeed, and it is a bit faster to use the native version.

But hopefully a native version will come out soon and we'll be able to remove the JS/Babel linting/compilation.

alex-e-leon (approved):

PR #8718, opened 2026-05-15
FE-715: Fix petrinaut react-compiler issues in dev build pipeline

alex-e-leon:

React Compiler's `validateNoRefPassedToFunction` rule was firing on the bundled output of @hashintel/ds-components. The library's source uses standard JSX (<button ref={ref} ... />), which the compiler permits — but tsup (via esbuild) compiles JSX into explicit `jsx("button", { ref, ... })` calls - which causes `validateNoRefPassedToFunction` to fail.

Since Vite follows the workspace symlink from node_modules/@hashintel/ds-components to libs/@hashintel/ds-components/dist/, the default node_modules exclude didn't match and the babel plugin ran the compiler over the bundled output.

This fixes the issue by having react-compiler ignore the pre-built ds-components export.

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/petrinaut/vite.config.ts
+++ b/libs/@hashintel/petrinaut/vite.config.ts
@@ -96,6 +96,15 @@ export default defineConfig(({ command }) => ({
 
     react(),
     babel({
+      // Default excludes node_modules. Also skip workspace `dist/` outputs:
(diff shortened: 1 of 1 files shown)


kube (approved):
Thanks!
This should even make the compilation step faster if it was compiling so many things.

PR #8720, opened 2026-05-15
FE-706: Add design system components to Petrinaut

alex-e-leon:

Replaces all Buttons, IconButtons and Icons with DS components.

This PR:

The changes in this PR:


kube (approved):
:+1:
Thanks!

Cool to know that these are all in the Design System now.

A little remark: `name` in `<Icon name=""` is a bit strange because you could think it is the HTML property, where it is actually the id of the icon (`id` wouldn't be better though)

alex-e-leon:
> :+1:
> Thanks!
> 
> Cool to know that these are all in the Design System now.
> 
> A little remark: `name` in `<Icon name=""` is a bit strange because you could think it is the HTML property, where it is actually the id of the icon (`id` wouldn't be better though)

Agree it's not perfect - I'll mull over it and see if I can come up with anything better, or if you think of anything let me know. I took a look at what some of the other major design systems use in case any of them have a better property name, but the only one which uses an <Icon> api I've found so far is Atlaskit and they also use the term "name".

PR #8736, opened 2026-05-21
H-6429: TypeScript dependency updates

CiaranMn:

Dependency updates to address security reports.

~**Includes a major NextJS version bump.**~ This forced a migration from webpack to turbo, so I'm going to come back to it. The latest Next 16 doesn't even fix the vulnerable `postcss` version.


apps/hash-frontend/src/proxy.ts
```diff
@@ -30,7 +30,7 @@ const applyCspHeaders = (
   return response;
 };
 
-export const middleware = async (request: NextRequest) => {
+export const proxy = async (request: NextRequest) => {
```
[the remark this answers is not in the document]
CiaranMn: https://nextjs.org/blog/next-16#proxyts-formerly-middlewarets

apps/hash-frontend/next.config.js
```diff
@@ -143,10 +143,6 @@ export default withSentryConfig(
       },
       pageExtensions: ["page.tsx", "page.ts", "page.jsx", "page.jsx", "api.ts"],
 
-      // We call linters in GitHub Actions for all pull requests. By not linting
```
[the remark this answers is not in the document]
CiaranMn: https://github.com/hashintel/hash/pull/8736/commits/8ebaec186c245c3ac87c3b2ac4c5110a1816f7f1

libs/@hashintel/ds-components/src/preset/stories/tokens.color-layering.story.tsx
```diff
@@ -66,7 +66,7 @@ const NestedLayers = ({ palette }: { palette: PaletteName }) => (
 );
 
 const NestedSurface = ({ palette }: { palette: PaletteName }) => (
-  <Box colorPalette={palette as "blue"}>
+  <Box colorPalette={palette as "bg.min"}>
```
kube: Are we sure about this?

apps/hash-frontend/src/middleware.ts
kube, line 1: I don't remember NextJS, but:
- I suppose this was not acting as a middleware if it had the `.page` suffix.
- Doesn't removing the `.page` suffix make it act as an actual middleware, and thus possibly change the behaviour of some pages?

kube (dismissed):
LGTM.
Just a few comments.

kube (dismissed):

libs/@hashintel/ds-components/src/preset/stories/tokens.color-layering.story.tsx
kube: Re-reading it, and it's actually broken even before.

I don't understand why we cast as `blue` here, not more than I understand why we cast as `bg.min`.

kube: But it's just a Storybook story so should be okay.

apps/hash-frontend/src/middleware.ts
CiaranMn: That was my bad – it got renamed to `proxy` when upgrading to Next 16 but then I got the name wrong when renaming it back again.

https://github.com/hashintel/hash/pull/8736/commits/a372a741a23bfcba2dd92c0744fabbbea0790e59

libs/@hashintel/ds-components/src/preset/stories/tokens.color-layering.story.tsx
kube: I suppose `colorPalette` is supposed to reference a ColorPalette by name.

`blue` is a plausible ColorPalette name, `bg.min` is not.

CiaranMn: The type assertions shouldn't change runtime behaviour. I've replaced them with proper types in https://github.com/hashintel/hash/pull/8736/commits/29c13649ed2d31b4fd2cef414f0eb058b4f52e39 anyway

BUT the color variants story is not showing correctly in the preview deployment, whereas it is in the production deployment. This file (color layering) doesn't show correctly in either. It might be that the PandaCSS broke variants, or it might be that it will Just Work in production. Since a lot of the storybook doesn't show correctly I'd like to just create a ticket to investigate after.

kube: Yeah there won't be any runtime impact.
Just wondering what our types do.

kube (dismissed):

libs/@hashintel/ds-components/src/preset/stories/tokens.color-layering.story.tsx
CiaranMn: `tsc` failed because `red` isn't assignable to the type for whatever reason.

I'll create a ticket for it, I've gone back to asserting a type that matches.

kube (approved):

PR #8750, opened 2026-05-25
H-6487: Add AI assistant to Petrinaut + demo website

CiaranMn:

1. Add the Petrinaut AI assistant UI
2. Adds the endpoint to the demo website that proxies to the AI provider (OpenAI for now)
3. Adds a 'fractal file structure' skill (since I made the AI part of `@hashintel/petrinaut` organised in this way)

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- Some new added ones.

## ❓ How to test this?

1. Open the preview deployment
2. For local testing, `turbo run dev` in `petrinaut-website` after copying `.env.example` to `.env` and setting an `OPENAI_API_KEY`. If you want simulation mode to work – it doesn't when running the website in dev mode – you need to use `npx vercel dev` instead (which will build the site and serve it, as well as serving the chat API endpoint)


libs/@hashintel/petrinaut-core/src/schemas/entity-schemas.ts
```diff
@@ -200,16 +225,22 @@ export const differentialEquationSchema = z
...
+        "ID of the colour/type whose token attributes this dynamics function updates. MUST match the `colorId` of every place that references this equation via `differentialEquationId`.",
     }),
     code: z.string().meta({
-      description:
-        "JavaScript module code exporting Dynamics(...). Return derivatives for each token attribute that changes continuously.",
+      description: [
+        "Module: `export default Dynamics((tokens, parameters) => …)`.",
+        "`tokens` is THIS place's current tokens only — `Array<{ [elementName]: number }>` — NOT all places' tokens.",
+        "MUST return an array of the SAME LENGTH where each entry is `{ [elementName]: derivative }` (i.e. dx/dt, NOT the new value).",
+        "The engine integrates with Euler: `next = current + derivative * dt`.",
```
kube: This is a detail of implementation of the Engine.

I understand that it's helpful for the LLM to know how what it outputs will be handled by the engine though.

Let's keep it like this for now.
Maybe worth adding a comment or something to let us know it should be rephrased later?

CiaranMn: I'll just remove it, I don't know if this actually helps the AI write better dynamics.

CiaranMn: By which I mean "I can't see how it would"

CiaranMn:
```suggestion
```

apps/petrinaut-website/api/chat.ts
```diff
@@ -0,0 +1,225 @@
...
+  headers.set("content-type", "application/json");
+  return new Response(JSON.stringify(body), { ...init, headers });
+};
+
+const logChatFailure = (
+  reason: string,
+  context: Record<string, unknown> = {},
+) => {
+  // oxlint-disable-next-line no-console
+  console.error(`[Petrinaut AI] ${reason}`, context);
```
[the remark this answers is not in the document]
CiaranMn: The reason comes from an LLM provider.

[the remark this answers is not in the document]
CiaranMn: The reason comes from an LLM provider.

libs/@hashintel/petrinaut/src/ui/views/Editor/panels/ai-assistant-panel.tsx
```diff
@@ -0,0 +1,627 @@
...
+  useEffect(() => {
+    diagnosticsVersionRef.current += 1;
+  }, [diagnosticsByUri]);
+
+  useEffect(() => {
+    diagnosticsContextRef.current = formatDiagnosticsForAi({
+      definition: petriNetDefinition,
+      diagnosticsByUri,
+    });
+  }, [diagnosticsByUri, petriNetDefinition]);
```
kube, line 236: This is the part for which we'll need to bring the diagnostics directly into the `PetrinautInstance`.

kube (approved):
Good job!

PR #8753, opened 2026-05-26
FE-749 add ds tooltip to petrinaut

alex-e-leon:

Switch to the DS <Tooltip> component in petrinaut.

This PR:

The changes in this PR:


kube (dismissed):
Good job!

It's cleaner, and solves many issues.

Just too little things, we can do later:

Screenshot 2026-05-27 at 11 19 18

Can we just make the Tooltips on the left tabs align to the right of the element so they don't hide the other tabs?

Image

When Tooltip is on a Dropdown button, maybe we can disable the tooltip when the Dropdown is open.

libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
```diff
@@ -18,7 +18,7 @@ export const contentStyles = cva({
     borderRadius: "md",
     paddingX: "2",
     paddingY: "1",
-    textStyle: "sm",
+    textStyle: "xs",
```
kube, line 21: Good design choice!

alex-e-leon:
> Can we just make the Tooltips on the left tabs align to the right of the element so they don't hide the other tabs?
Updated

> When Tooltip is on a Dropdown button, maybe we can disable the tooltip when the Dropdown is open.
Yes agree - I haven't built the dropdown into the DS yet, but when I do this behaviour will be included.

kube (approved):

PR #8754, opened 2026-05-26
FE-750: Fix color variants in panda

alex-e-leon:

Fixes broken color variants being exported from the design system.

This PR:

The changes in this PR:


CiaranMn (commented):
Does this allow us to remove the type assertions in the stories? the `as "bg.min"`

alex-e-leon:
> Does this allow us to remove the type assertions in the stories? the `as "bg.min"`

Nope - the way the stories are currently constructed its not enumerating over the token values but instead constructing them out of partial token name parts through string concatenation, so they need to be typecast.

Eventually I'd love to clean up the token generation logic since it seems more complex than it needs to be but not right now.

CiaranMn (approved):

PR #8757, opened 2026-05-26
FE-748: Switch ds from tsup to tsdown

alex-e-leon:

- Switches ds-components from tsup to tsdown. tsup is deprecated and no longer maintained, so switching to tsdown is recommended.
- Disables the nextjs webpack build cache in production for apps/hash-frontend - which fixes docker builds that were breaking in scenarios where the webpack build cache was already populated during other tasks.

This PR:

The changes in this PR:

The changes in this PR:


alex-e-leon:
After discussion with @lunelson he prefers not to switch back to tsdown at this time so closing.

PR #8758, opened 2026-05-26
SRE-726, H-6526: Single profiled Compose stack for self-hosting HASH

TimDiekmann:

Lay the groundwork for self-hosting HASH: consolidate every Docker/Compose asset into one profiled stack under `infra/compose/`, dockerise all app services, and make the stack self-contained so `docker compose --profile hash up` brings up a complete HASH instance with no env-file wrangling.

This began as "move external services + dockerise apps" and grew (same Linear issue) into the full overhaul below.

## 🔗 Related links

- [SRE-726](https://linear.app/hash/issue/SRE-726) _(internal)_
- [FE-752](https://linear.app/hash/issue/FE-752) — clean server/browser API-origin split; the dockerised-frontend networking here is a workaround for it
- [SRE-728](https://linear.app/hash/issue/SRE-728) — hosting images on GHCR; works today, but the frontend image isn't origin-portable until FE-752

## 🔍 What does this change?

**Layout**
- Move `infra/docker/{api,frontend}/prod/Dockerfile` → `apps/hash-{api,frontend}/docker/Dockerfile`; add Dockerfiles for `hash-graph`, `hash-ai-worker-ts`, `hash-integration-worker`
- Move `apps/hash-external-services/` → `infra/compose/`; drop the vestigial `@apps/hash-external-services` workspace
- Rename `docker-compose.*` → `compose.*`, single project name `hash`
- Drop `infra/docker/{citus,statsd}/` and the statsd-override file

**One profiled compose file** — merge the dev/test/prod overlays into a single `compose.yml` with profiles:
- `dev` — infra for local development
- `hgres` — graph layer + its deps (Postgres, Temporal) only
- `hash` — the full app stack (graph, api, migrate, frontend, workers)
- `observability` — Grafana/Loki/Tempo/Mimir
- Add a `hash-api-migrate` one-shot that seeds the system graph (mirrors the ECS migrate task)
- Self-contained: every value has an inline `${VAR:-default}`, so the stack runs without an env-file; trim the now-redundant vars from `.env`/`.env.test`
- Wire every long-running app service to the otel-collector
- Switch `temporal-migrate`/`temporal-setup` to the pulled admin-tools image

**Dockerised frontend** — the frontend container shares the api's network namespace (`network_mode: service:api`) so its server-side `localhost:5001` reaches the api. Compose-only workaround for the single-`API_ORIGIN` assumption (browser, CSP and SSR all share one origin); the clean fix is FE-752.

**CI** — call `docker compose --profile dev` directly (working-directory `infra/compose`) instead of the yarn wrapper, using `up -d` (the dev profile's one-shot bucket-init trips `--wait`); `prune-repository` preserves `infra/`.

**Docs** — rewrite the README's local-run section (dev quickstart + profiles) and point self-hosting / full-stack readers to hash.dev/docs. (A companion change in the internal-sites repo rewrites the get-started setup page.)

- [x] documentation changes are included in this PR

- [x] affected the execution graph, and `turbo.json` has been updated

## ⚠️ Known issues

- The dockerised-frontend networking is a `network_mode` workaround, not the real fix (FE-752); it works for local/self-hosting but couples frontend↔api.
- Images are still built locally. Pushing them to GHCR works, but the frontend image bakes `NEXT_PUBLIC_*` at build time, so the published image is pinned to one origin until FE-752 makes it runtime-configurable (SRE-728).

## 🛡 What tests cover this?

No new tests. The compose stack is exercised by the existing `test.yml` / `bench.yml` workflows (now on `--profile dev`).

## ❓ How to test this?

**Full stack (self-hosting):**

1. `cd infra/compose && docker compose --profile hash up -d`
2. Sign up at http://localhost:3000 (verification mail at http://localhost:4436)

**Local dev:**

1. `yarn compose --profile hgres up -d` (infra + graph in Docker)
2. `yarn dev` (api + frontend natively); log in as `admin@example.com` / `password`


infra/compose/compose.yml
```diff
@@ -0,0 +1,803 @@
...
+      - hash
+    image: otel/opentelemetry-collector-contrib:0.128.0
+    command: ["--config=/etc/otelcol/config.yaml"]
+    volumes:
+      - ./opentelemetry-collector/otel-collector-config.yaml:/etc/otelcol/config.yaml:ro
+    ports:
+      - 4317:4317
+    read_only: true
+    security_opt:
+      - no-new-privileges:true
```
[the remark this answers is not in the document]
TimDiekmann: @CiaranMn I noticed this while developing the new format. Open to any route, possible options:

1. Status quo: let the Otel collector error (no traces/logs/metrics until `--profile observability` is enabled)
2. Run `Loki`, `Tempo`, and `Mimir` all the time, only enable Grafana with `--profile observability`: More things to run, but no errors and retroactive logs
3. Same as above + Grafana (always running everything, no `--profile observability`).
4. 
Note, that this will only affect runs with `docker compose up` from `infra/compose`. The `yarn compose up` includes both, `--profile dev` and `--profile observability` all the time.

I lean towards 2., but also see 1. as very viable.

infra/compose/compose.yml
```diff
@@ -0,0 +1,798 @@
...
+        condition: service_completed_successfully
+    restart: unless-stopped
+    ports:
+      - "4433:4433" # public API
+      - "4434:4434" # admin API
+    environment:
+      SECRETS_COOKIE: "VERY-INSECURE-AND-SHOULD-ONLY-BE-USED-IN-DEV"
+      SECRETS_CIPHER: "32-LONG-SECRET-NOT-SECURE-AT-ALL"
+      DSN: "postgres://${HASH_KRATOS_PG_USER:-kratos}:${HASH_KRATOS_PG_PASSWORD:-kratos}@postgres:${POSTGRES_PORT:-5432}/${HASH_KRATOS_PG_DATABASE:-kratos}"
+      OAUTH2_PROVIDER_URL: "http://hydra:4445"
```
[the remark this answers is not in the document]
TimDiekmann: Discussed internally

TimDiekmann:
@copilot fix the linting errors please.

apps/hash-api/src/setup-admin.ts
thehabbos007, line 1: Nice setup flow

.env
```diff
@@ -13,42 +13,30 @@ HASH_INTEGRATION_QUEUE_NAME=integration
...
 KRATOS_API_KEY=secret
 KRATOS_COOKIE_DOMAIN=localhost
 
 HASH_HYDRA_PUBLIC_URL=http://127.0.0.1:4444
 HASH_HYDRA_ADMIN_URL=http://127.0.0.1:4445
-HASH_HYDRA_PG_USER=hydra
-HASH_HYDRA_PG_PASSWORD=hydra
-HASH_HYDRA_PG_DATABASE=hydra
-HASH_HYDRA_SECRETS_SYSTEM=VERY-INSECURE-AND-SHOULD-ONLY-BE-USED-IN-DEV
-HASH_HYDRA_SECRETS_COOKIE=VERY-INSECURE-AND-SHOULD-ONLY-BE-USED-IN-DEV
```
thehabbos007, line 35: I see that we now default these values in the new compose stack, but I would like to supply all configurable env vars in one shared spot (`.env`). To know that `HASH_HYDRA_SECRETS_SYSTEM` is configurable, one needs to be aware of it or look into the compose rather than looking here in the `.env`. What are our thoughts on this?

turbo.json
```diff
@@ -125,9 +125,7 @@
     "package.json",
     "pyproject.toml",
     "rust-toolchain.toml",
-    "yarn*",
-    // External services
-    "apps/hash-external-services/**"
```
thehabbos007, line 130: Should the compose stack be added here to be part of the graph?

infra/compose/compose.yml
```diff
@@ -0,0 +1,782 @@
...
+    network_mode: "service:api"
+    depends_on:
+      api:
+        condition: service_healthy
+    deploy:
+      restart_policy:
+        condition: on-failure
+    environment:
+      FRONTEND_URL: "${FRONTEND_URL:-http://localhost:3000}"
+      API_ORIGIN: "${API_ORIGIN:-http://localhost:5001}"
```
thehabbos007, line 725: Just a question regarding these, I know there was some discussion on build vs runtime for these. If i were to pull the image from ghcr.io/hashintel/hash/frontend, would the `FRONTEND_URL`/`API_ORIGIN` be changeable? Or is it baked in anywhere there?

README.md
```diff
@@ -59,156 +57,119 @@ When you first create an account you may be placed on a waitlist. To jump the qu
...
+If you're **not** working on the graph itself, run it in Docker instead of compiling Rust locally — add the `hgres` profile and start only the api + frontend natively:
 
-   Note: seeding only runs when `NODE_ENV=development`, start the Graph API separately using `yarn start:graph` then launch the app using `yarn dev` in a separate terminal to start it in development environment.
+```sh
+yarn compose --profile hgres up -d   # infrastructure + graph in Docker
+yarn dev                             # api + frontend only
+```
 
-##### Running the browser plugin
+Add `--build` when pulling changes that touch the graph. For the full application stack in Docker, see [the setup guide](https://hash.dev/docs/get-started/setup#local-hash).
```
thehabbos007: Is this correct? We don't have any `build:` keys in the new compose stack for graph, api or frontend, so would `--build` change anything? We do have it for kratos, postgres, etc.

thehabbos007 (commented):
Looks good to me. I have some comments that we should consider

turbo.json
TimDiekmann: I removed it because it only caused issues without an advantage. While we orchestrate the Rust-code as turbo-packages to determine dependencies, the compose file does not have any dependencies. Previously (way before this PR) we had the Graph as service in here as well so we required a dependency, but now it was just an artifact. 
I would leave it out.

infra/compose/compose.yml
TimDiekmann: Currently it's baked into the binary. The relevant issue is FE-752. This is not ideal, but it's at least a way of providing it at ghcr and local compilation.

README.md
TimDiekmann: Ah, good spot. I retroactively changed this PR to use `ghcr` already (it had a `build:` section before) so this is outdated. Removed it.

.env
TimDiekmann: I don't feel strongly about this, but I removed them from `.env` because `.env` became way too bug and the configs are not used anywhere except in the compose-file. 
I like to get rid of the global .env file at some point (and move to proper configuration files). And if you want to host it yourself, you need to set these variables anyway. I probably would keep them out, unless you feel strongly.

thehabbos007: I do not feel strongly about it. I just did something similar in a previous project where it was of utmost importance that all env vars were transparently applied in the right order by env and deployment 'segment'. In that case, it helped us to create a shared env file that we could refer back to and statically reason about. But that is overkill here. So let's just keep in mind that having centralized env vars may be helpful, but not something we need to do here right now :)

turbo.json
thehabbos007: Gotcha, leave it out :)

infra/compose/compose.yml
thehabbos007: Okay, so we will consider this problem later on. That makes sense to me!

thehabbos007: Gotcha, so we will deal with it in a follow up, no problem

thehabbos007 (dismissed):
LGTM now, thanks Tim! ping me if you need reapproval after rebase/merge

infra/compose/README.md
```diff
@@ -43,7 +43,7 @@ Start all development services:
 
 ```bash
 # From the repository root
-yarn external-services up --wait
+yarn compose up --wait
```
indietyp, line 46: calling this compose feels very weird to me. I had to do a double take that it's not the docker compose binary or anything like that.

apps/hash-api/src/setup-admin.ts
```diff
@@ -0,0 +1,163 @@
...
+    rl.close();
+  }
+};
+
+/**
+ * Reads a line from stdin without echoing characters, masking each typed
+ * character with `*`. Falls back to a visible read if stdin isn't a TTY
+ * (piped/non-interactive input).
+ */
+const promptHidden = (label: string): Promise<string> => {
```
indietyp, line 36: Rolling our own here seems odd, especially because you first instantiate readline, and then immediately use process stdin and stdout.

This should also just work:

```
const rl = readline.createInterface({
  input: process.stdin,
  output: process.stdout,
  terminal: true,
});

rl._writeToOutput = function (str) {
  if (rl.stdoutMuted) rl.output.write("*");
  else rl.output.write(str);
};

rl.question("Password: ", (password) => {
  rl.close();
  process.stdout.write("\n");
});
rl.stdoutMuted = true;
```

(also applies to promises, just means no callback)

if you want to instead just remove the output there's also: https://stackoverflow.com/a/33500118

infra/compose/compose.yml
```diff
@@ -0,0 +1,782 @@
...
+    build:
+      context: ./kratos
+    depends_on:
+      postgres:
+        condition: service_healthy
+    read_only: true
+    security_opt:
+      - no-new-privileges:true
+    environment:
+      - DSN=postgres://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD:-postgres}@postgres:${POSTGRES_PORT:-5432}/${HASH_KRATOS_PG_DATABASE:-kratos}
```
indietyp: why is this the only place where we use the list form, not the key-value form for the environment?

infra/compose/compose.yml
```diff
@@ -0,0 +1,782 @@
...
+        condition: on-failure
+    healthcheck:
+      test: ["CMD-SHELL", "redis-cli ping | grep PONG"]
+      interval: 2s
+      timeout: 2s
+      retries: 5
+    security_opt:
+      - no-new-privileges:true
+    ports:
+      - "6379:6379"
```
indietyp, line 233: I really dislike that we're just opening the ports here, because
A. it's a nightmare to set up then
B. it means that because it this publishes it to _any_ host address, meaning that inside of internal networks people can now access the redis instance etc.

README.md
```diff
@@ -59,156 +57,117 @@ When you first create an account you may be placed on a waitlist. To jump the qu
...
-##### Running the app
+Install these on the host first:
 
-1. Make sure you have, [Git](https://git-scm.com), [Rust](https://www.rust-lang.org), [Docker](https://docs.docker.com/get-docker/), and [Protobuf](https://github.com/protocolbuffers/protobuf). Building the Docker containers requires [Docker Buildx](https://docs.docker.com/build/install-buildx/).
-   Run each of these version commands and make sure the output is expected:
+- **[Git](https://git-scm.com)** (≥ 2.17)
+- **[Docker](https://docs.docker.com/get-docker/)** (≥ 20.10), including:
+  - **[Docker Compose](https://docs.docker.com/compose/)** (≥ 2.17) — orchestrates the stack
+  - **[Docker Buildx](https://docs.docker.com/build/install-buildx/)** (≥ 0.10) — builds the images
+- **[mise](https://mise.jdx.dev/)** — installs and pins Node, Rust, protoc and the other required tool versions
```
indietyp, line 70: You don't need mise if you just want to use the docker compose no?

infra/compose/compose.yml
TimDiekmann: I did not change this, GitHub was just not able to detect that this file was moved from `docker-compose.yml`

```suggestion
      DSN: "postgres://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD:-postgres}@postgres:${POSTGRES_PORT:-5432}/${HASH_KRATOS_PG_DATABASE:-kratos}"
```

infra/compose/README.md
TimDiekmann: I don't understand. It's called `yarn compose` and all it does is calling `docker compose` with a few more arguments. I don't see the issue here. You can literally pass the same arguments as you would to `docker compose` and it's inside the `infra/compose` folder.

apps/hash-api/src/setup-admin.ts
TimDiekmann: Your proposal uses an internal node function. Also, it doesn't handle `CTRL+C`, backspace, echo, or non TTY stdin. I'd rather keep it as is.

infra/compose/compose.yml
TimDiekmann: A: Not sure what you mean by "nightmare", these are the standard ports for each services and devs expect them.

B: On Docker Desktop this goes through a VM and on linux it's bridged but typically firewalled. If you expose all ports to the network you have other issues. We potentially can switch to `127.0.0.1:<port>:<port>` or make it configurable, however, this is a bigger change and needs proper testing. Will leave it as it is for now.

And the port is required by the Node API.

README.md
TimDiekmann: The section literally says

> These instructions set up HASH for **working on the codebase**

and for this, `mise` is required (or at least the recommended option, it should work without it if you have everything installed).

Further, the section even mentions a link where instructions are found without installing the toolchain:

> To run the whole stack from container images instead (no toolchain required on the host), follow [the setup guide](https://hash.dev/docs/get-started/setup#local-hash).

infra/compose/compose.yml
```diff
@@ -0,0 +1,782 @@
...
+      - "3000:3000"
+
+  frontend:
+    profiles:
+      - hash
+    image: ghcr.io/hashintel/hash/frontend
+    network_mode: "service:api"
+    depends_on:
+      api:
+        condition: service_healthy
```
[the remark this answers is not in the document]
TimDiekmann: healthcheck is baked into the Dockerfile.

infra/compose/compose.yml
```diff
@@ -0,0 +1,782 @@
...
+        condition: service_completed_successfully
+    restart: unless-stopped
+    ports:
+      - "4433:4433" # public API
+      - "4434:4434" # admin API
+    environment:
+      SECRETS_COOKIE: "VERY-INSECURE-AND-SHOULD-ONLY-BE-USED-IN-DEV"
+      SECRETS_CIPHER: "32-LONG-SECRET-NOT-SECURE-AT-ALL"
+      DSN: "postgres://${HASH_KRATOS_PG_USER:-kratos}:${HASH_KRATOS_PG_PASSWORD:-kratos}@postgres:${POSTGRES_PORT:-5432}/${HASH_KRATOS_PG_DATABASE:-kratos}"
+      OAUTH2_PROVIDER_URL: "http://hydra:4445"
```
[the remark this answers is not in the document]
TimDiekmann: It's working fine.

apps/hash-api/src/setup-admin.ts
indietyp: You use readline just above, hence the suggestion, it's also not true, readline is made for *exactly* this thing (I had to modify the function a bit so that it works properly w/ display), but proof in video:


https://github.com/user-attachments/assets/5e8e0336-2599-44f3-a36a-8e95e5682876

This specific API has also been available (and used in packages like [`read`](https://github.com/npm/read), hence why I suggested it in the first place), also fine to keep it, but just wanted to flag.

hashdotai (approved):

apps/hash-api/src/setup-admin.ts
TimDiekmann: We can still change this in follow-up, but the current thing is working and the PR needs to be merged.

infra/compose/compose.yml
indietyp: A. sure, but it means that I cannot have e.g. a system redis running while I have the docker compose up, breaking encapsulation. This has happened to me before (in this very project) many times.

B. That is not true. Per spec (https://github.com/compose-spec/compose-spec/blob/main/spec.md#long-syntax-3): "host_ip: The host IP mapping. If it is not set, it binds to all network interfaces (0.0.0.0)." I am fine with deferring it, but I wanted to mention it explicitly.

> And the port is required by the Node API.

That is fine, that's usually done through the default network inside of the compose file anyway, instead of routing through the host. If it runs outside during dev that's fine, but i am specifically looking at this from the prod angle^^

indietyp: ^ You can test this yourself:
1. grab your local ip (`ipconfig getifaddr en0`)
2. spawn an nginx container: `docker run --rm -p 8080:80 nginx:alpine`
3. use your phone to navigate to `[ip]:8080`.

(verified just now using orbstack)

TimDiekmann: > If it runs outside during dev that's fine, but i am specifically looking at this from the prod angle^^

I want to emphasize, that the docker-file is _not_ a production-grade configuration and the issue you describe applies to all services, not only to redis. We currently don't have something to "just-run" for self-hosted production, see https://hash.dev/docs/get-started/setup/self-hosted. In addition to the additional steps required and properly hardening the environment, the `SELF_HOSTED_HASH` env-var also must be set, which is currently not tested.

TimDiekmann: Your test, btw, failed for me. I was not able to connect to my mac from my phone that way. I'm sure you can configure it to make it work, but out of the box it did not work.

image

indietyp: it worked out of the box, I just tried it in a configuration where I changed nothing. This is the default configuration that shipped.

Screenshot 2026-05-31 at 1 22 03 PM

indietyp: > I want to emphasize, that the docker-file is not a production-grade configuration and the issue you describe applies to all services, not only to redis.

Yes but I didn't want to annotate each ports configuration. The issue with binding to `0.0.0.0` still stands, I also didn't do a pure prod audit (then the notes would be probably a lot more) but rather: what would be the problems that I would run into when running this on a large internal network (on my dev machine) and the issue with `0.0.0.0` and colliding with a redis on host both stand the test of time here.

TimDiekmann: And this is why I propose to change it to `127.0.0.1`:

> we potentially can switch to 127.0.0.1:<port>:<port> or make it configurable

But this is not the correct PR to change the behavior which already existed before.

PR #8767, opened 2026-05-27
FE-728: Fix toast styling in petrinaut website

alex-e-leon:

Due to the fact that panda css styling gets scoped to a specific dom subtree, any elements that open portals (dialogs, popovers, toasts etc) need to be portaled into that subtree or else the scoped styles won't be applied. The way that this has been implemented is to add a `<PortalContextProvider>` element that scopes all portals beneath it to the portal context element instead of the root html element.

Since `<PetrinautProvider>` was recently updated to call toasts during some operations, it needs to sit underneath this `<PortalContextProvider>` or else the toasts it triggers won't get styled. This PR moves the PortalContextProvider to now be at the app root to fix the issue.

This PR:

The changes in this PR:


kube (approved):
Thanks! Works like a charm

PR #8774, opened 2026-05-27
FE-740: Add Read Arcs

kube:

This PR introduces Read Arcs in Petrinaut.

Read Arcs work like normal input arcs for enablement: the source place must contain at least `weight` tokens. For coloured places, those tokens are provided to `Lambda` and `TransitionKernel` under `input.<PlaceName>`, with tuple length equal to the arc weight.

The difference from standard arcs is consumption. When a transition fires, standard input arc tokens are removed immediately, while read arc tokens remain in their source place. This lets a transition inspect shared state or token attributes without moving that token through the transition.

They are almost syntactic sugar for a self-loop (same Place as input and output of transition with same token), but they are convenient in the sense that they prevent the user from accidentally modifying the data they are trying to read.

### Change in runtime behaviour

Currently, for simulating the same behaviour you could use this self-loop pattern.

```mermaid
graph LR
  InputPlace(InputPlace)
  TransitionA["Reader A"]
  TransitionB["Reader B"]
  
  TransitionA --> InputPlace
  TransitionB --> InputPlace
  
  InputPlace --> TransitionA
  InputPlace --> TransitionB
```

Though given the tokens of Reader A would be consumed temporarily in the frame during which the transition would be happening, Reader B would not be able to consume it until the next frame.

With Read Arcs, both these transitions would be triggered in the same frame.

### Impacted areas

- Core SDCPN model: `InputArc.type` now accepts `"read"` alongside `"standard"` and `"inhibitor"`.
- Validation/import paths: runtime schemas, file import, and clipboard import preserve read arcs while continuing to default older payloads to `"standard"`.
- Mutations and AI tools: `addArc` can create typed input arcs directly, and `updateArcType` can switch an input arc to read.
- Simulation engines: read arcs participate in enablement and token combination enumeration, but are excluded from token removal in both standard and Monte Carlo execution.
- LSP/UI/docs: read arcs contribute typed input properties, are selectable/rendered distinctly in the editor, and are documented alongside standard and inhibitor arcs.

### Edge cases

- Uncoloured read arcs still gate enablement but do not appear in lambda/kernel input, matching uncoloured standard arc behaviour.
- A transition with only read arcs can fire repeatedly while inspecting the same tokens unless a predicate, output effect, time limit, or another input stops it.
- Arc identity remains place-to-transition based; this does not add support for parallel input arcs between the same place and transition.

This PR:

- [X] modifies an **npm**-publishable library and **I have added a changeset file(s)**

The changes in this PR:

- [X] require changes to docs which **are made** as part of this PR

The changes in this PR:

- [X] do not affect the execution graph

## 🛡 What tests cover this?

- Structural enablement tests for read arcs.
- Single-step and full-frame simulation tests proving read arc tokens reach lambda/kernel input and are not removed.
- Monte Carlo coverage for non-consuming read arcs.
- LSP completion coverage for read arc token inputs.
- File and clipboard parsing coverage for `"read"` arc types.
- AI tool/schema tests for typed `addArc` inputs.

Verified with:

- `yarn workspace @hashintel/petrinaut-core test:unit`
- `yarn workspace @hashintel/petrinaut-core lint:tsc`
- `yarn workspace @hashintel/petrinaut-core lint:eslint`
- `yarn workspace @hashintel/petrinaut-core build`
- `yarn workspace @hashintel/petrinaut test:unit`
- `yarn workspace @hashintel/petrinaut lint:tsc`
- `yarn workspace @hashintel/petrinaut lint:eslint`

## ❓ How to test this?

1. Checkout the branch or view the deployment.
2. Create an input arc from a place to a transition and set its type to **Read** in the arc properties panel.
3. Run a simulation where the transition reads a coloured token from that place.
4. Confirm the transition can use the token data while the source token remains in place after firing.
5. For shared-state behaviour, add two transitions reading the same place and confirm both can fire in the same frame.


libs/@hashintel/petrinaut/src/ui/views/SDCPN/components/arc.tsx
```diff
@@ -236,31 +252,57 @@ export const Arc: React.FC<EdgeProps<ArcEdgeType>> = ({
...
-            <circle
-              cx={INHIBITOR_MARKER_RADIUS + BASE_STROKE_WIDTH}
-              cy={INHIBITOR_MARKER_RADIUS + BASE_STROKE_WIDTH}
-              r={INHIBITOR_MARKER_RADIUS}
-              fill="white"
-              stroke={strokeColor}
-              strokeWidth={BASE_STROKE_WIDTH}
-            />
-          </marker>
+          {data.arcType === "inhibitor" ? (
```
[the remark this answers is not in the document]
kube: Thanks Copilot, you are really bad.

libs/@hashintel/petrinaut-core/src/schemas/entity-schemas.ts
```diff
@@ -49,16 +50,20 @@ export const inputArcSchema = z
...
     }),
   })
   .meta({
     description: "Input arc from a place into a transition.",
-  });
+  }) satisfies z.ZodType<{
+  placeId: string;
+  weight: number;
+  type: InputArcType;
+}>;
```
CiaranMn: Keeps it in sync with the type in `sdcpn/types.ts`
```suggestion
  }) satisfies z.ZodType<InputArc>;
```

CiaranMn (approved):

libs/@hashintel/petrinaut-core/src/schemas/entity-schemas.ts
kube: Done

PR #8779, opened 2026-05-28
FE-765: Remove line-height on tooltip wrapper

alex-e-leon:

Sets the line-height on tooltip wrappers to 0 so that small items inside tooltips are properly aligned. This is safe to do, because the internal size/line-height of an element is still used to set the min height of the line-box, so it doesn't change the alignment/height of items at all.

## 🔗 Related links
- https://drafts.csswg.org/css-inline-3/

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
+++ b/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
@@ -1,6 +1,8 @@
 import { css, cva } from "@hashintel/ds-helpers/css";
 
 export const triggerStyles = css({
+  lineHeight: "[0]",
+
   "&:focus-visible": {
     outline: "[2px solid]",
     outlineColor: "neutral.s30",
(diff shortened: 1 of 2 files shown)


libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
```diff
@@ -1,6 +1,8 @@
 import { css, cva } from "@hashintel/ds-helpers/css";
 
 export const triggerStyles = css({
+  lineHeight: "[0]",
```
kube, line 4: As suggested by AI, isn't there a way to reset the line-height for descendants? To be sure that descendants inherit their line-height properly if needed?

alex-e-leon: Had another deeper look and there's no nice way to fix this short of using JS. This should be safe though, because consumers should usually be explicitly setting line-heights on items rather than inheriting. If not, we can re-assess later.

kube (approved):
Thank you!
We'll need to be careful when applying Tooltip to components that have no line-height, but is acceptable.

PR #8780, opened 2026-05-28
FE-766: Add blur to subtle buttons

alex-e-leon:

Adds a blur to subtle buttons so that button text is readable when floating over busy content/backgrounds.

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
+++ b/libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
@@ -114,11 +114,7 @@ export const styles = sva({
     },
     variant: {
       solid: {},
-      subtle: {
-        button: {
-          backdropFilter: "[blur(10px)]",
-        },
-      },
+      subtle: {},
       ghost: {
         button: {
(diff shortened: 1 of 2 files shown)


libs/@hashintel/ds-components/src/components/Button/button.recipe.ts
```diff
@@ -114,7 +114,11 @@ export const styles = sva({
     },
     variant: {
       solid: {},
-      subtle: {},
+      subtle: {
+        button: {
+          backdropFilter: "[blur(10px)]",
+        },
+      },
```
kube: I don't know how many `subtle` variants we have in the codebase, but backdropFilter is kinda heavy, so maybe not worth for all buttons.

kube (dismissed):
Thank you!
Just a little remark/question, mostly because if we have a lot of `"subtle"` buttons it's gonna be heavy to render.

kube (dismissed):

kube (dismissed):

kube (approved):

PR #8786, opened 2026-05-29
FE-840: Add Property Filter To Entities Visualizer

yusufkinatas:

Follow-up for https://github.com/hashintel/hash/pull/8783 based on [the task](https://docs.google.com/document/d/1uVpMqrFaIqs-w7xI5-YTJwEZcCva3KzvvrXNhrjydwU)
Adds **per-property value filtering** to the Entities Visualizer. Previously the ribbon only supported filtering by web, type, and archived status; you could narrow the table by which entities were shown but not by the values of their properties.

This lets a user filter the table on the properties they can see — e.g. `Age > 13`, `Name contains "Acme"`, `Active is true`, or `Email has any value` — by adding property-filter "pills" to the existing filter ribbon. Each pill picks a property, an operator appropriate to the property's value kind, and (where relevant) a value; the resulting clauses are ANDed with the existing web / type / archived filters and pushed down into the graph query.

## 🔍 What does this change?

**Filter model & query building**

- Extends `EntitiesFilterState` with a `propertyFilters` array (`data/types.ts`).
- Adds `data/property-filters/types.ts` defining `PropertyFilter`, the `PropertyFilterOperator` set, the `FilterValueKind` (`number` / `string` / `boolean`), and the `FilterableProperty` shape (filterable, or disabled-with-reason).
- `buildPropertyFilterClause` (`data/property-filters/build-property-filter-clause.ts`) translates a single filter into a graph `Filter` clause as a pure, unit-testable function. It owns parameter typing (numbers coerced to JS numbers so comparisons are numeric, strings kept untrimmed for case-sensitive matching) and the `exists` / `not exists` composition. An incomplete or invalid filter yields `null` and contributes no clause, so an unfinished pill is inert rather than matching nothing. `buildEntitiesFilter` now folds these clauses into the query.
- `deriveFilterableProperties` (`data/property-filters/derive-filterable-properties.ts`) derives the pickable properties from the same closed-entity-type / definitions data that builds the visible columns ("filter on the columns you see"). Properties that can't be filtered in v1 (lists, nested objects, multiple data types, unsupported kinds) are either omitted or surfaced disabled with a reason. The filterable interpretation wins when a base URL appears in multiple shapes across types.
- `get-operators-for-kind.ts` catalogs the operators available per value kind, their dropdown labels, pill connectors, and whether they require a value, with a sensible default operator per kind.

**UI**

- `AddFiltersMenu` gains a "Property…" entry that opens an in-menu `PropertyFilterPicker` (searchable list of filterable properties, disabled rows with explanatory tooltips).
- `PropertyFilterPill` renders each property filter in the ribbon with a kind icon, operator/value editor popover (operator changes apply immediately; value input is debounced before hitting shared state to avoid refetching on every keystroke), an active/incomplete visual state, and removal.
- `FilterRibbon` wires up add/change/remove handlers, auto-opens the editor for a just-added pill, wraps to multiple lines, and includes property filters in the "filters are default" / clear-all logic.
- `incompletePillSx` added for the muted placeholder look of pills that don't yet contribute a clause.

**Data-type pool robustness (supporting change)**

- `EntitiesTableData` now bundles the `dataTypeDefinitions` pool alongside the rows it was generated from, unioned across paginated pages. This guarantees a row's value can always resolve its data type and fixes `formatValue` throwing when a refetch narrowed the result set (e.g. a property filter matching nothing) before rows were regenerated. `EntitiesTable` no longer takes `definitions` directly and reads the pool from the table data.
- `getReferencedDataTypeIds` (`format-value.ts`) is exposed so the table can cheaply check that every data type a value depends on is present before rendering, falling back to a "Not Found" cell instead of crashing the grid.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- v1 only supports filtering `number`, `string`, and `boolean` properties. List, nested, and multiple-data-type properties are listed in the picker but disabled with an explanatory tooltip.
- String matching (`equals` / `contains` / `startsWith` / `endsWith`) is case-sensitive.

## 🐾 Next steps

- Broaden support to currently-unfilterable kinds (lists, nested properties, multiple data types).
- Consider case-insensitive string matching as an option.

## 🛡 What tests cover this?

- none

## ❓ How to test this?

1. Checkout the branch / view the deployment.
2. Open the Entities Visualizer for a set of entities with properties of varying kinds.
3. In the filter ribbon, click the add-filter control and choose **Property…**.
4. Search for and select a property; confirm the operator list matches its value kind and a pill appears with its editor open.
5. Choose an operator and enter a value (where required); confirm the table refetches and narrows accordingly, and that an incomplete pill is shown muted and applies no filter.
6. Add multiple property filters and confirm they are ANDed together; remove them and confirm "Clear filters" resets everything.
7. Confirm a filter matching nothing shows an empty table without crashing the grid.

## 📹 Demo

https://github.com/user-attachments/assets/6d375964-8f0d-4a62-a18d-499ca3f26659


apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
```diff
@@ -93,10 +121,37 @@ export const FilterRibbon: FunctionComponent<FilterRibbonProps> = ({
...
+        />
       )}
-      {!filtersAreDefault && <ClearFiltersButton onClear={handleClear} />}
+      <AddFiltersMenu
+        canAddIncludeArchived={!filterState.includeArchived}
+        onAddIncludeArchived={() => setIncludeArchived(true)}
+        filterableProperties={filterableProperties}
+        propertiesLoading={propertiesLoading}
+        onAddPropertyFilter={handleAddPropertyFilter}
+      />
```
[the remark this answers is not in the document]
CiaranMn: I know. The omnipresent 'clear' button was often pushing the filter onto two lines and we have limited space there. It's not really that big a deal for users to change the filters back.

vilkinsons (dismissed):

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
```diff
@@ -23,56 +31,76 @@ type FilterRibbonProps = {
...
-  const typeIsDefault =
-    isTypePinned || isTypeFilterDefault(filterState.type, availableTypes);
-  const archivedIsDefault = !filterState.includeArchived;
+  const setPropertyFilters = (
+    updater: (prev: PropertyFilter[]) => PropertyFilter[],
+  ) =>
+    setFilterState((prev) => ({
+      ...prev,
+      propertyFilters: updater(prev.propertyFilters),
+    }));
```
TimDiekmann, line 59: Just a word of warning: Without a web-filter and without a type filter, this will slow down the frontend like crazy for big graphs. Also for web filters with webs with very many entities. Properties are not indexed (due to the nature of JSONb in Postgres). The single `archived` property before already implied this effect as Postgres requires to do a full scan. We need to be careful with adding filters to random things.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
```diff
@@ -0,0 +1,192 @@
...
+ * Properties that fail the gate are still returned, but annotated with a
+ * {@link FilterableProperty.disabledReason} so the picker can list them disabled
+ * with a reason-specific tooltip.
+ *
+ * When the same property base URL appears across several entity types in
+ * different shapes (e.g. a list on one type, a single value on another), the
+ * **filterable** interpretation wins, so the user can still filter on the column
+ * they see. Among unfilterable interpretations the first encountered wins.
+ *
+ * Pure function – authored so it can be unit-tested in isolation.
```
TimDiekmann: It would be good to also add a few unit tests in this case

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/build-property-filter-clause.ts
```diff
@@ -0,0 +1,115 @@
...
+    case "lessThanOrEqual":
+      return { lessOrEqual: [{ path }, { parameter }] };
+    case "contains":
+      return { containsSegment: [{ path }, { parameter }] };
+    case "startsWith":
+      return { startsWith: [{ path }, { parameter }] };
+    case "endsWith":
+      return { endsWith: [{ path }, { parameter }] };
+    default:
+      return null;
```
TimDiekmann: If we add a new operator, we very likely miss to add It here. It would be good if the `switch` statement would be exhaustive (and add `satisfies never`)

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/build-filter.ts
TimDiekmann, line 100: This is not a valid filter. An always-true filter is `all: []` ("every") and an always-false filter is `any: []` ("at least one").

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/use-available-types.ts
TimDiekmann: This only `null`s the type, but it retains the property filters. This means other types cannot be selected anymore. It also implies the quite expensive call with `includeTypeIds`. I would not restrict the type IDs at all. If you search for a type, e.g. `Pdf`, but don't have any in your web, the result set is empty. That's expected UX.

apps/hash-frontend/src/pages/shared/entities-visualizer/header/type-filter-pill.tsx
TimDiekmann, line 42: This compares the available list to the current list. This is wrong if types are filtered, this is not "all types". It's a false claim.

Image

This is not even true if we don't filter further because we never include all types (we exclude a few types by default + if a web doesn't have all types used, "All Types" is still wrong.

Additionally, as this function is used to determine if we want to exclude the noise-types, this will wrongly include the filter (which will slow down the query!).

TimDiekmann, line 239: When filtering by a property without matching the filter, this results in very odd behavior, such as 

Image

I feel quite strongly right now, that we should _not_ use the return graph to determine what filters we allow.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
```diff
@@ -0,0 +1,192 @@
...
+      kind = resolveValueKind(mergedSchema.anyOf[0]!.type);
+    } else {
+      kind = resolveValueKind(mergedSchema.type);
+    }
+  } catch {
+    // The data type has a shape `getMergedDataTypeSchema` can't reduce to a
+    // single set of constraints (e.g. mixed primitives) – treat it as a
+    // multiple-data-types case.
+    return disabled("multiple-data-types");
+  }
```
TimDiekmann: This catches everything, such as graph errors.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
```diff
@@ -0,0 +1,192 @@
...
+  // list of values – neither is filterable in v1.
+  if (!("$ref" in valueDefinition)) {
+    return disabled(valueDefinition.type === "object" ? "nested" : "list");
+  }
+
+  const dataType = definitions.dataTypes[valueDefinition.$ref];
+
+  if (!dataType) {
+    return null;
+  }
```
TimDiekmann: This at least should emit a warning, if not a sentry error.

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table.tsx
```diff
@@ -388,6 +394,21 @@ export const EntitiesTable: FunctionComponent<
...
+               * bundled into `tableData` this should not happen in normal flow,
+               * but it still guards against a genuinely inconsistent response.
+               */
+              const someDataTypeUnresolved = getReferencedDataTypeIds(
+                propertyMetadata,
+              ).some((dataTypeId) => !dataTypeDefinitions[dataTypeId]);
+
+              if (someDataTypeUnresolved) {
+                return notFoundCell;
+              }
```
TimDiekmann, line 409: At least a warning, if not a Sentry error here would be helpful.

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table.tsx
```diff
@@ -341,22 +345,24 @@ export const EntitiesTable: FunctionComponent<
...
              * this by briefly not displaying anything in the cell.
              */
-            return {
-              kind: GridCellKind.Text,
-              allowOverlay: false,
-              readonly: true,
-              displayData: String("Not Found"),
-              data: "Not Found",
-            };
+            return notFoundCell;
```
TimDiekmann: It is displaying "Not Found", wrong comment.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/property-filter.ts
```diff
@@ -0,0 +1,78 @@
...
+  /** Resolved value kind, determines the available operators and parameter typing. */
+  kind: FilterValueKind;
+  /** The chosen operator. */
+  operator: PropertyFilterOperator;
+  /**
+   * The raw value from the editor's input. Absent (or empty / invalid for the
+   * kind) means the filter is incomplete and contributes no clause. Unused by
+   * value-less operators (boolean / existence).
+   */
+  value?: string;
```
TimDiekmann, line 53: The value for things like `greaterThan` is not a string but a number.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/get-operators-for-kind.ts
```diff
@@ -0,0 +1,157 @@
...
+ * booleans).
+ */
+export const getOperatorsForKind = (
+  kind: FilterValueKind,
+): OperatorDescriptor[] => operatorsByKind[kind];
+
+/** The default operator for a kind (the first in its catalog). */
+export const getDefaultOperatorForKind = (
+  kind: FilterValueKind,
+): PropertyFilterOperator => operatorsByKind[kind][0]!.operator;
```
TimDiekmann: If the operator is missing int he above list, this fails.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
```diff
@@ -0,0 +1,192 @@
...
+ * - it is not used as a list/array on the entity type;
+ * - its property type has exactly one value definition (`oneOf.length === 1`);
+ * - that single definition is a direct data-type reference (not a nested
+ *   property object, not a list);
+ * - the resolved data type permits exactly one data type; and
+ * - that data type resolves to a `number`, `string`, or `boolean` kind.
+ *
+ * Properties that fail the gate are still returned, but annotated with a
+ * {@link FilterableProperty.disabledReason} so the picker can list them disabled
+ * with a reason-specific tooltip.
```
TimDiekmann, line 197: Missing definitions or null-kinds are not shown at all

apps/hash-frontend/src/pages/shared/format-value.ts
```diff
@@ -12,12 +12,50 @@ import {
...
+
+    // Object metadata is rendered as JSON and needs no data type.
+    if (isValueMetadata(metadata) && metadata.metadata.dataTypeId) {
+      dataTypeIds.push(metadata.metadata.dataTypeId);
+    }
+  };
+
+  collect(valueMetadata);
+
+  return dataTypeIds;
```
TimDiekmann, line 56: If a value doesn't have a dataTypeId (happens in AI inference), this returns an empty list.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/filter-state.ts
```diff
@@ -9,6 +10,12 @@ export type EntitiesFilterState = {
     selectedTypeIds: Set<VersionedUrl> | null;
   };
   includeArchived: boolean;
+  /**
+   * Per-property value filters (e.g. `Age > 13`). Each is translated into a
+   * graph filter clause and ANDed with the web / type / archived filters – see
+   * {@link buildPropertyFilterClause} and {@link buildEntitiesFilter}.
```
TimDiekmann: The links do not resolve

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table-data.ts
```diff
@@ -130,6 +130,17 @@ export type VisibleDataTypeIdsByPropertyBaseUrl = Record<
 
 export type EntitiesTableData = {
   columns: EntitiesTableColumn[];
+  /**
+   * The data type definitions referenced by {@link rows}, captured from the
+   * same query result the rows were generated from.
+   *
+   * Bundling the pool with the rows guarantees that anything drawing a row's
+   * value can always resolve its data type: the two can never come from
+   * different query results (which previously caused `formatValue` to throw
```
TimDiekmann: Isn't pagination exactly for this so the _do_ come from different results?

apps/hash-frontend/src/pages/shared/entities-visualizer/header/property-filter-pill.tsx
```diff
@@ -0,0 +1,296 @@
...
+  string: FontCaseRegularIcon,
+  boolean: CheckRegularIcon,
+};
+
+type PropertyFilterPillProps = {
+  filter: PropertyFilter;
+  mode: "add" | "edit";
+  /** Whether to open the editor automatically (just-added draft). */
+  autoOpen: boolean;
+  onAutoOpenHandled: () => void;
```
TimDiekmann: We never use this, right?

apps/hash-frontend/src/pages/shared/entities-visualizer/header/property-filter-pill.tsx
```diff
@@ -0,0 +1,296 @@
...
+  number: HashtagRegularIcon,
+  string: FontCaseRegularIcon,
+  boolean: CheckRegularIcon,
+};
+
+type PropertyFilterPillProps = {
+  filter: PropertyFilter;
+  mode: "add" | "edit";
+  /** Whether to open the editor automatically (just-added draft). */
+  autoOpen: boolean;
```
TimDiekmann, line 47: Do we ever flip this?

apps/hash-frontend/src/pages/shared/entities-visualizer/header/add-filters-menu.tsx
```diff
@@ -4,27 +4,48 @@ import {
...
 type AddFiltersMenuProps = {
+  canAddIncludeArchived: boolean;
   onAddIncludeArchived: () => void;
+  filterableProperties: FilterableProperty[];
+  propertiesLoading: boolean;
+  onAddPropertyFilter: (property: {
+    baseUrl: BaseUrl;
+    title: string;
+    kind: FilterValueKind;
+  }) => void;
```
TimDiekmann: This is `FilterableProperty`

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
```diff
@@ -23,56 +31,76 @@ type FilterRibbonProps = {
...
 
-  const filtersAreDefault = webIsDefault && typeIsDefault && archivedIsDefault;
+  const handleAddPropertyFilter = (property: {
+    baseUrl: BaseUrl;
+    title: string;
+    kind: FilterValueKind;
+  }) => {
```
TimDiekmann: This is `FilterableProperty`

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
```diff
@@ -93,10 +121,37 @@ export const FilterRibbon: FunctionComponent<FilterRibbonProps> = ({
...
+      {draftPropertyFilter && (
+        <PropertyFilterPill
+          key={draftPropertyFilter.id}
+          filter={draftPropertyFilter}
+          mode="add"
+          autoOpen
+          onAutoOpenHandled={() => {}}
+          onCommit={handleCommitDraftPropertyFilter}
+          onRemove={() => setDraftPropertyFilter(null)}
+        />
```
TimDiekmann, line 139: It's not clear, what filters apply how

Image

The property filter could mean they **both** needs to match or **any** needs to match

I appreciate, that we don't have a way to create other filters (such as `(X and Y) or (A and B)` but it should be clear from the UI what is filtered. 

Also, it's not possible to only see archived entities.

TimDiekmann (dismissed):
Given how many edge cases we have here (and I really don't like the UX, from a filter I expect it to behave differently): I would _**not**_ filter the types based on the currently shown entities.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/build-filter.ts
CiaranMn: It is a valid filter, it just won't match anything. 

I can short-circuit the whole thing with an `any: []` if no types are selected if that's what you mean?

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/property-filter.ts
CiaranMn: This is the raw input value (always a string, even for number inputs), it gets converted into a number before being sent to the Graph (in `coerceValueParameter`)

apps/hash-frontend/src/pages/shared/entities-visualizer/header/add-filters-menu.tsx
CiaranMn: It's not exactly because `FilterableProperty` was a union of things, I've split it out in https://github.com/hashintel/hash/pull/8786/commits/6fe9094a9fafe9d25c25cb359a492248499027d3

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
CiaranMn: Appreciate that. How about we add a type filter to the query which limited the types to only the one(s) we know have the property in question?

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
```diff
@@ -23,56 +31,76 @@ type FilterRibbonProps = {
...
+    kind: FilterValueKind;
+  }) => {
+    setDraftPropertyFilter({
+      id: generatePropertyFilterId(),
+      baseUrl: property.baseUrl,
+      title: property.title,
+      kind: property.kind,
+      operator: getDefaultOperatorForKind(property.kind),
+    });
+  };
```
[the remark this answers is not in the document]
CiaranMn: That's fine, clicking off the in-progress thing will discard it regardless

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
CiaranMn: It's not exactly because FilterableProperty was a union of things, I've split it out in 6fe9094

apps/hash-frontend/src/pages/shared/entities-visualizer/header/filter-ribbon.tsx
CiaranMn: Not being able to see only archived entities was the existing behaviour. We could potentially have an 'Archived only' option although I'd rather do it later..

These filters are cumulative. They must all match. We can talk about some improved design to make that clearer

apps/hash-frontend/src/pages/shared/entities-visualizer/header/property-filter-pill.tsx
CiaranMn: Yes, it is either true or false when this component is loaded in `filter-ribbon.tsx` depending on whether it's a newly added draft or not

apps/hash-frontend/src/pages/shared/entities-visualizer/header/property-filter-pill.tsx
CiaranMn: No – removed in https://github.com/hashintel/hash/commit/6fe9094a9fafe9d25c25cb359a492248499027d3

apps/hash-frontend/src/pages/shared/entities-visualizer/header/property-filter-pill.tsx
```diff
@@ -0,0 +1,296 @@
...
+            {operators.map((operatorDescriptor) => (
+              <MenuItem
+                key={operatorDescriptor.operator}
+                value={operatorDescriptor.operator}
+                sx={{ fontSize: 13 }}
+              >
+                {operatorDescriptor.label}
+              </MenuItem>
+            ))}
+          </Select>
```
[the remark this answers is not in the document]
CiaranMn: This does not appear to be true

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/use-available-types.ts
```diff
@@ -32,21 +32,18 @@ export const useAvailableTypes = ({
...
+        filterState: {
+          web: filterState.web,
+          type: { selectedTypeIds: null },
+          includeArchived: filterState.includeArchived,
+          propertyFilters: [],
+        },
         internalWebIds,
       }),
-    [filterStateWithoutType, internalWebIds],
+    [filterState, internalWebIds],
```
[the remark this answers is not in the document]
CiaranMn: This is intended

apps/hash-frontend/src/pages/shared/entities-visualizer/header/type-filter-pill.tsx
CiaranMn: `allAvailableIds` here is all types returned from a query for entiites in the webs. What you are seeing here is a bug from including the property filters in `useAvailableTypes` which I have fixed in https://github.com/hashintel/hash/pull/8786/changes/d4dd48a247099d4587f0bc0e7b0be59cc64f6e46

I know it isn't 'All types' but it's everything in the list we choose to show the user (which is what is available in the webs they've selected, excluding some system stuff). Open to suggestions for something better to write here. "Any in web(s)"?

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/build-property-filter-clause.ts
CiaranMn: I just removed the default, then TypeScript will complain about the return type being wrong (since it will fall through to `undefined` if a case is missed)

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
CiaranMn: I'm still not sure about showing these disabled properties at all tbh.

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/derive-filterable-properties.ts
CiaranMn: AI comment, I have removed in https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table-data.ts
CiaranMn: AI comment, removed in https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entities-visualizer/header/type-filter-pill.tsx
CiaranMn: Same issue fixed in https://github.com/hashintel/hash/commit/d4dd48a247099d4587f0bc0e7b0be59cc64f6e46

We were also using the return graph to determine what properties to show, which missed properties not in the page. I have updated in https://github.com/hashintel/hash/pull/8786/changes/9a235199fcbd17c963a4d46337cae71a89b06018

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/get-operators-for-kind.ts
CiaranMn: It shouldn't be but I have thrown an error in https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/build-filter.ts
```diff
@@ -144,12 +146,18 @@ export const buildEntitiesFilter = ({
     clauses.push(typeClause);
   }
 
-  const userPickedSpecificTypes =
-    !isTypePinned && filterState.type.selectedTypeIds !== null;
-
-  if (!isTypePinned && !userPickedSpecificTypes) {
```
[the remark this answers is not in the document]
CiaranMn: This is intended. We want to exclude those types from the visualizer always, unless the user is looking at it in the context of a specific type's page (in which case the type filter is not even available)

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/filter-state.ts
CiaranMn: Removed in https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table.tsx
CiaranMn: Updated to an actually blank cell in https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entities-visualizer/entities-table.tsx
CiaranMn: https://github.com/hashintel/hash/pull/8786/changes/fd93b310a719b1d61def21f19a10c7371f9743ad

apps/hash-frontend/src/pages/shared/entity/entity-editor/links-section.tsx
```diff
@@ -38,9 +35,6 @@ export const LinksSection = ({ isLinkEntity }: { isLinkEntity: boolean }) => {
       !draftLinksToArchive.includes(
         incomingLinkAndSource.linkEntity[0].entityId,
       ) &&
-      !incomingLinkAndSource.linkEntity[0].metadata.entityTypeIds.some(
-        (typeId) => noisySystemTypeIds.includes(typeId as NoisySystemTypeId),
```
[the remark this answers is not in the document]
CiaranMn: Yes, that's fine, I think in the context of looking at a single entity we should faithfully show all incoming/outcoming links

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/build-filter.ts
```diff
@@ -144,12 +146,18 @@ export const buildEntitiesFilter = ({
...
+  if (!isTypePinned) {
     clauses.push(ignoreNoisySystemTypesFilter);
   }
```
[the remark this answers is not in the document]
CiaranMn: The types should not show in the type filter at all (because they are excluded from the initial query).

If the user arrives at the viz via a specific type's page, only entities of that type (even noisy ones) will be included, and the type filter does not appear.

This is the intended behaviour

apps/hash-frontend/src/pages/shared/entities-visualizer/shared/property-filters/build-property-filter-clause.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+};
+
+/**
+ * Whether a filter currently contributes a clause to the query – i.e. whether
+ * it is "active". A filter that builds no clause (incomplete / invalid) is not
+ * active and is shown in a muted state in the ribbon.
+ */
+export const isPropertyFilterActive = (filter: PropertyFilter): boolean =>
+  buildPropertyFilterClause(filter) !== null;
```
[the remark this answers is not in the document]
CiaranMn: tsc handles this, it will error because it does not list `undefined` in the return type

apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
```diff
@@ -332,9 +346,88 @@ export const EntitiesVisualizer: FunctionComponent<{
...
+        ...filterState.type.selectedTypeIds,
+      ].filter((selectedTypeId) => availableEntityTypeIds.has(selectedTypeId));
+
+      if (
+        retainedSelectedTypeIds.length !== filterState.type.selectedTypeIds.size
+      ) {
+        nextSelectedTypeIds =
+          retainedSelectedTypeIds.length === 0
+            ? null
+            : new Set(retainedSelectedTypeIds);
```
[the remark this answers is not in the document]
CiaranMn: This is by intention. The user did not have 'no types' selected previously (which would be pointless anyway)

TimDiekmann (approved):

PR #8789, opened 2026-05-29
FE-756: petrinaut product walkthrough

alex-e-leon:

Adds a product walkthrough to the petrinaut website. The logic for the new walkthrough lives inside the petrinaut library so that it can be kept in sync with any changes to the petrinaut ui and/or interact with other modals - however the content (copy + any video animations + their urls) is passed in by the consuming application by including the WalkthroughProvider - not including a walkthrough provider means that no preview will be rendered.

We track whether the user has seen the walkthrough in its own localstorage variable rather than the userSettings variable in order to keep the data better co-located with the context + provider and avoid situations where a provider is added in the wrong order.

The animations in this PR are not final yet, and still need polish @kube has graciously offered to record the experiments animation over the weekend. Will update the PR once they are done.

This PR:

The changes in this PR:


libs/@hashintel/petrinaut/src/ui/components/walkthrough/walkthrough-provider.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    if (!hasSeenWalkthrough) {
+      return;
+    }
+    try {
+      localStorage.setItem(STORAGE_KEY, "true");
+    } catch {
+      // Ignore write failures (e.g. quota exceeded)
+    }
+  }, [hasSeenWalkthrough]);
```
kube: Could I ask you to use `UserSettings` since now this code live in the Petrinaut component.

You should create the UserSetting value + setter in the UserSettingContext then:

```ts
  const {
    showWalkthrough
    setShowWalkthrough,
  } = use(UserSettingsContext);
```

Or something with a better name.

And add it in the ViewportSettings component too.

It is a beautiful Walkthrough you created, and I want to be able to re-open it from time to time.

kube (commented):
Beautiful, just a comment on the LocalStorage usage.

libs/@hashintel/petrinaut/src/ui/components/walkthrough/walkthrough-provider.tsx
kube: Ha yes, and instead of putting that in the Provider, put it in the Dialog so you have access to UserSettingsContext.

kube (dismissed):

libs/@hashintel/petrinaut/src/ui/components/walkthrough/walkthrough-provider.tsx
alex-e-leon: Done

vilkinsons (approved):

PR #8793, opened 2026-05-31
SRE-735: Quote AWS account IDs in deploy.yml to preserve leading zero

TimDiekmann:

Quote the AWS account IDs in `deploy.yml` so the staging ID keeps its leading zero. Unquoted, `054238437032` was parsed as an integer (`54238437032`), producing an invalid role ARN — OIDC role assumption failed with `No OpenIDConnect provider found`.

## 🔗 Related links

- SRE-735

## 🔍 What does this change?

- Quote `AWS_DEPLOY_ACCOUNT_ID` and `AWS_ECR_ACCOUNT_ID` so YAML parses them as strings, not integers.


indietyp (approved):
@indietyp approved this PR from Slack with [Graphite](https://app.graphite.com/github/pr/hashintel/hash/8793)

PR #8801, opened 2026-06-01
SRE-739: Fix backend Deploy concurrency — dropped builds and out-of-order mutable tags

TimDiekmann:

The backend `Deploy` workflow was intermittently cancelling Docker builds on `main` (e.g. `Build kratos (arm64)` → "The operation was canceled"). Root cause: the shared `…-publish` concurrency group dropped pending main-push runs (GitHub keeps only one pending run per group). Combined with the per-commit `turbo affected` change detection and the promote walk-back, a dropped run for a service-changing commit could let `promote` ship a stale image to production **with no error**.

This reworks the workflow's concurrency, manifest assembly, mutable-tag handling, and build cache so that no run is dropped and mutable tags only ever move forward.

## 🔗 Related links

- [SRE-739](https://linear.app/hash/issue/SRE-739) _(internal)_

## 🚫 Blocked by

- none

## 🔍 What does this change?

- **Concurrency**: the workflow-level group is now `github.run_id` (unique per run → never evicted); serialization moved to job-level `queue: max` per resource. Builds run **unserialized** (their pushes are immutable), so the slow Rust builds no longer block each other. Only the mutable tags/rollouts serialize — `:staging` in a dedicated `stage` job, `:latest` in `manifest`, ECS in `deploy` (which runs `needs: stage`, so the redeploy pulls the freshly-advanced `:staging` rather than the previous image).
- **by-digest manifests**: GHCR images are pushed by digest; the multi-arch manifest is assembled from those exact digests instead of reading back the moving `:latest-<arch>` tags. Removes the "Frankenstein manifest" race (arm64 from one commit + amd64 from another) and drops the per-arch GHCR tags.
- **Mutable-tag guard** (`tag-if-newer` action): `:staging` / `:latest` advance onto a commit only if it is a strict git-descendant of the currently-tagged commit (resolved via the `org.opencontainers.image.revision` label). Closes the "`queue: max` FIFO not guaranteed" gap and protects a manual `:staging` rollback against older/parallel runs.
- **Build cache → GHCR registry**: moved off the GHA cache backend (`type=gha,mode=max` was 504-ing on the large Rust builds) to a dedicated **private** `…/<service>/cache:<arch>` package. Written on `push`/`workflow_dispatch` (trusted events) only; PR/merge_group read it but never write it.
- **Retry** (`nick-fields/retry`, repo convention): wraps the `imagetools create` calls (manifest assemble + guard) against transient GHCR 5xx.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- **The `is_main` path is not exercised by CI before merge.** ECR push (`:sha`/`:run`/`:staging`), the `:staging` guard, the `stage` job, and the ECS `deploy` job all require `is_main`, so they are skipped on `pull_request` (`PUSH=false`) and `workflow_dispatch` (non-main). They run for the first time on the first `main` push — **that run should be watched**. The riskiest unverified assumption is `imagetools create` against **ECR** for `:staging` (ECR's image-index media-type quirk); the fallback is `aws ecr put-image`.
- The out-of-order guard is logically correct but **not empirically tested** with two overlapping runs of different commits (the `:latest` guard only ran single-run).
- First `main` run: existing `:staging`/`:latest` images have no `revision` label yet → the guard advances them once unconditionally (by design).

## 🐾 Next steps

- **Rollback pin**: the guard only moves `:staging` *forward*, so a manual rollback isn't durable on its own (the next merge overtakes it). Needs a separate pin/freeze mechanism if "freeze staging" is ever required.
- Move `:staging` tagging into `internal-infra` via `workflow_call` (keep the `:latest` guard in this OSS repo — it can't go closed-source).

## 🛡 What tests cover this?

- No automated tests (this is CI workflow config). Validated by a `workflow_dispatch` run on this branch: all 12 builds parallel, GHCR by-digest + registry cache (no 504), 5 multi-arch manifests assembled from digests, `:latest` guard. A separate throwaway probe confirmed job-level `queue: max` serializes without dropping runs. The `is_main` path is covered only by the first `main` run after merge.

## ❓ How to test this?

1. `gh workflow run deploy.yml --ref <this branch>` → exercises the GHCR path (build-parallel, by-digest, cache, manifest, `:latest` guard). `stage`/`deploy` are skipped (non-main).
2. On merge, watch the first `main` run for: ECR push, the `stage` job (`:staging` guard), and the ECS redeploy — these cannot be exercised beforehand.

## 📹 Demo

n/a — CI / infra change.


.github/workflows/deploy.yml
```diff
@@ -370,25 +465,54 @@ jobs:
...
-          docker buildx imagetools create \
-            --tag "$IMAGE:sha-$SHA" \
-            "$IMAGE:sha-$SHA-arm64" \
-            "$IMAGE:sha-$SHA-amd64"
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          retry_wait_seconds: 10
+          command: |
+            cd "$RUNNER_TEMP/digests"
```
indietyp, line 488: I guess this relies on the fact that it runs on the same runner, which I think is a reasonable assumption. Am I understanding this correctly that you chose to use the runner's temporary storage, instead of using outputs, because of the required task serialisation?

TimDiekmann: Neither, we download the artifacts in the step above.

indietyp: Yes, I am aware of that. My question was not about that, but rather to confirm the reasoning for using `$RUNNER_TEMP` (in conjunction with the prior write to that location) instead of using task outputs.

TimDiekmann: I'm not sure what you're proposing. Do you mean why we don't use task output from the build step? That's not possible, it's a matrix.

indietyp: I didn't want to propose anything^^ I just wanted to confirm the intention. Thank you!

indietyp (approved):

PR #8808, opened 2026-06-02
FE-790: Switch petrinaut to use DS dialog

alex-e-leon:

- Use the design system dialog in petrinaut
- I also update the z-index values across petrinaut to use the proper z-index values where appropriate, and fix up the containerRef zIndex so that items stack properly.

@kube I've tested all the dialogs/drawers/panels where I adjusted the zIndex - except for the Toast, I couldn't figure out how to trigger one. Would you be able to double check the toast zIndex is correct?

Also a note on the panel zIndex ordering. If in the future you do need to manually adjust the zIndex of items - its usually better to subtract rather than add, as that way DS components on the same layer that use the plain zIndex will still appear on top when they are triggered in.

This PR:

The changes in this PR:


libs/@hashintel/petrinaut/src/ui/views/Editor/panels/BottomPanel/panel.tsx
```diff
@@ -27,7 +27,7 @@ import {
 
 const glassPanelBaseStyle = css({
   position: "absolute",
-  zIndex: 1001,
+  zIndex: "[calc(var(--z-index-sticky) - 2)]",
```
kube, line 30: Wouldn't we gain readability if we centralized these zIndexes in the Petrinaut Panda config?

kube (approved):
Love that!
Image

Thanks it's good work!

libs/@hashintel/petrinaut/src/ui/views/Editor/panels/BottomPanel/panel.tsx
alex-e-leon: It would be nice, but these z-indexes are petrinaut specific. The zIndex variables in the design tokens are z-index "layers". If elements are rendered in the correct order on the correct layer, everything just works - but in petrinaut all the toolbars are on the "sticky" layer, but even if they are rendered out of order there's a specific ordering we want them to appear as (or so I assume).

If the layouts are set correctly + the panels don't overlap, then the ordering also doesn't matter, and they can just be position sticky or sticky - 1, but I didn't want to dive deep to find out in this PR.

Also I've just used the zIndex design tokens as is, but in the future I think we could refine them - not sure all the layers are necessary and some of the names could be a little better IMO.

PR #8812, opened 2026-06-03
FE-797: Fix tooltip zIndex

alex-e-leon:

Panda was not applying the zIndex style to tooltips. Why does panda not apply the style that I've set? Who knows. This PR fixes this by adding !important which seems to force panda to apply the style.
I also add styled-system to petrinauts gitignore file to avoid it being committed during development.
I also remove the tooltip from the ai-cta close button, as its obvious + redundant (but leave the aria label)

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
+++ b/libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
@@ -12,7 +12,7 @@ export const triggerStyles = css({
 });
 
 export const positionerStyles = css({
-  zIndex: "tooltip",
+  zIndex: "tooltip !important",
 });
 
 export const contentStyles = cva({
--- a/libs/@hashintel/petrinaut/.gitignore
+++ b/libs/@hashintel/petrinaut/.gitignore
@@ -1,0 +1,2 @@
+# PandaCSS
+styled-system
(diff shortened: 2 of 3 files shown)


libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
```diff
@@ -12,7 +12,7 @@ export const triggerStyles = css({
 });
 
 export const positionerStyles = css({
-  zIndex: "tooltip",
+  zIndex: "tooltip !important",
```
kube, line 15: Are we sure this is necessary?

kube (approved):

libs/@hashintel/petrinaut/.gitignore
```diff
@@ -0,0 +1,2 @@
+# PandaCSS
+styled-system
```
kube, line 2: :+1: For when Claude automatically builds it. :)

libs/@hashintel/ds-components/src/components/Tooltip/tooltip.recipe.ts
alex-e-leon: panda css's specificity rules are a total black box to me. Sometimes it just decides not to apply css rules and I have no idea why. In this case I think ark-ui already sets a z-index and for some reason it decided ark-ui's z-index was more important than our style - but setting !important changed its mind. 🤷

PR #8813, opened 2026-06-03
FE-798: Add a touch of padding to vertical container sub views

alex-e-leon:

The scrolling issue identified looks like it was actually just some missing padding causing the editor to look like it was going off screen. Adding a little bit of padding looks fixes the issue.

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/petrinaut/src/ui/components/sub-view/vertical/vertical-sub-views-container.tsx
+++ b/libs/@hashintel/petrinaut/src/ui/components/sub-view/vertical/vertical-sub-views-container.tsx
@@ -60,6 +60,7 @@ const sectionContentStyle = css({
   display: "flex",
   flexDirection: "column",
   flex: "[1]",
+  paddingBottom: "3",
 });
 
 const scrollContainerStyle = css({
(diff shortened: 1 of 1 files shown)


kube (commented):
There is no padding at the bottom of vertical views on purpose.

Mostly for cases where the child component is a scrollable list, and the padding should be handled by this list. (for instance in the Lists displayed in the LeftSidePanel)

Here I would personally add padding to Subview that had the problem, rather than to the SubViewContainer directly.

If you think that there is a better way, do not hesitate to propose, I just don't think the solution in this PR is taking everything into  account.

alex-e-leon:
> There is no padding at the bottom of vertical views on purpose.
> 
> Mostly for cases where the child component is a scrollable list, and the padding should be handled by this list. (for instance in the Lists displayed in the LeftSidePanel)
> 
> Here I would personally add padding to Subview that had the problem, rather than to the SubViewContainer directly.
> 
> If you think that there is a better way, do not hesitate to propose, I just don't think the solution in this PR is taking everything into account.

Good catch - I've moved the padding into the scrollable container instead.

kube (approved):
Careful just about the impact on existing scrollable lists that possibly have a padding-bottom themselves.

alex-e-leon:
> Careful just about the impact on existing scrollable lists that possibly have a padding-bottom themselves.

Yeah - we may need to remove some padding on existing sections. Its not a lot of extra padding though, so it shouldn't cause any big issues on sections that already have padding - they will just have a bit more now.

PR #8817, opened 2026-06-03
FE-801: Fix cropped selects in petrinaut by removing portaling behaviour

alex-e-leon:

Removes the ability to turn select portalling off in petrinaut which was leading to selects being cropped in overflow hidden containers.
Screenshot 2026-06-03 at 16 46 41

This PR:

The changes in this PR:


kube (approved):
Thank you! 🚀

PR #8833, opened 2026-06-06
FE-821: add menu to ds

alex-e-leon:

Adds menu component to design system

This PR:

The changes in this PR:


lunelson (approved):
Looks good, although I notice in Ladle I can't actually "check" the kitchen sink checkbox... clicking it closes the menu without changing its checked state

alex-e-leon:
> Looks good, although I notice in Ladle I can't actually "check" the kitchen sink checkbox... clicking it closes the menu without changing its checked state

Good catch - I'll clean up some of the stories in the next PR or 2.

PR #8842, opened 2026-06-08
FE-822: Add Select to ds

alex-e-leon:

Adds a Select component to the design system

This PR:

The changes in this PR:


lunelson (approved):
Seems to work very well; 

IMO unusual design choice to allow choosing an empty option... I would make that a non-default escape hatch I think

Image

PR #8847, opened 2026-06-09
FE-828: petrinaut use ds menu

alex-e-leon:

Switch petrinaut to use ds dropdown

This PR:

The changes in this PR:


libs/@hashintel/petrinaut/src/ui/views/Editor/components/TopBar/top-bar.tsx
```diff
@@ -104,7 +103,6 @@ export const TopBar: React.FC<TopBarProps> = ({
             />
           }
           items={menuItems}
-          animated
```
kube, line 107: We have no animation anymore when dropdown opens.

libs/@hashintel/ds-components/src/components/Menu/SelectableList/selectable-list-util.ts
```diff
@@ -33,7 +40,7 @@ export type Item = ItemBase &
         onClick: (id: string) => void;
       }
     | {
-        nestedItems: ItemOrGroup<Item>;
+        nestedItems: Array<ItemOrGroup<Item>>;
```
kube: Not introduced by this PR, but nestedItems seems strange IMO.

kube (commented):
We need to keep animations on open/close.

libs/@hashintel/petrinaut/src/ui/views/Editor/components/TopBar/top-bar.tsx
alex-e-leon: I've added animations back in.

kube (approved):

PR #8857, opened 2026-06-11
FE-751: Style username in signup-registration messaging

alex-e-leon:

Adds a small amount of styling to help improve the comprehension of the messaging on the user signup page by making it clear that @pizza is a username.

Although personally, I still find that the copy reads uncomfortably and I think updating the copy would be a better fix. Here's a couple alternatives I cobbled together that I think could work better:
- Before someone else grabs @pizza _(if we want to run with a subtle pizza joke)_
- To get a short handle like @pizza _(neutral, but clear)_
- Be @you, not @you2087 _(if we want something with a bit more urgency/marketing copy - may need to reserve those usernames though)_

@vilkinsons let me know if you prefer any of them and I'll switch them over - otherwise I'll just keep the original messaging.

Finally - the new styles:
Screenshot 2026-06-11 at 15 37 42

(And old for reference):
Screenshot 2026-06-11 at 15 33 37

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-frontend/src/pages/signup.page/signup-registration-right-info.tsx
+++ b/apps/hash-frontend/src/pages/signup.page/signup-registration-right-info.tsx
@@ -10,7 +10,11 @@ import { PizzaSolidIcon } from "../../shared/icons/pizza-solid-icon";
 
 import type { FunctionComponent, ReactNode } from "react";
 
-const infoItems: { title: string; description: string; icon: ReactNode }[] = [
+const infoItems: {
+  title: string;
+  description: React.ReactNode;
+  icon: ReactNode;
+}[] = [
   /** @todo: add this info item when email verification is supported */
   // {
   //   title: "Skip email verification",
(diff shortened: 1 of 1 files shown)


TimDiekmann (approved):
Agreed with you that the copy could be better. But this already is an improvement -> ✅

vilkinsons:
"Before someone else grabs pizza" reads best burial bleeds over 2 lines therefore is undesirable?

alex-e-leon:
@vilkinsons it only breaks on 2 lines between 1100px > 900px screen sizes (ie tablets). On my laptop its all one line:

Screenshot 2026-06-11 at 16 18 43

And here's the tablet version - it looks fine IMO.
Screenshot 2026-06-11 at 16 21 18

CiaranMn (approved):

alex-e-leon:
@vilkinsons - I've merged without the copy change, but put up a new PR with just the copy change here and tagged you in it: https://github.com/hashintel/hash/pull/8862

Just hit approve or close the PR depending on which outcome you prefer : )

PR #8861, opened 2026-06-12
FE-472: fix type header appearance

alex-e-leon:

Fixes issues with layout in entity-header and introduces panda-css to hash-frontend so that ds-components can be loaded in. A couple of notes:

- I'm not across the hash-frontend CI, so while it looks good to me, might be worth double checking that the panda-css step works properly for prod builds as well as dev.
- The layout strategy could be much improved for headers, drawers + pages IMO, but in order to fix the issue without blowing up the scope + to keep the changeset minimal I've just made some manual adjustments to padding in some cases instead of completely fixing the layout strategy, and applied some of second-choice strategies for handling large content: in particular in retrospect I'd probably keep the link icon aligned + right buttons aligned top, possibly even float the title, and consider breaking long words instead of ellipsifying them - but I think the approach is good enough for a first pass.
- ~I also have no idea why the mui-button drops its padding when under stress.~ Fixed
- I also fixed up an issue where the breadcrumb menu was sitting underneath the close button.
Screenshot 2026-06-12 at 12 49 00

- And finally here's a screenshot of how it looks with extreme content now:
Screenshot 2026-06-12 at 13 09 42
 
This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
@@ -107,8 +107,14 @@ export const EntityTypeHeader = ({
           alignItems="center"
           justifyContent="space-between"
         >
-          <Stack direction="row" alignItems="center" gap={5}>
-            <Stack direction="row" alignItems="flex-start" mt={1} mb={3}>
+          <Stack direction="row" alignItems="center" gap={5} minWidth={0}>
+            <Stack
+              direction="row"
+              alignItems="flex-start"
+              mt={1}
+              mb={3}
+              minWidth={0}
+            >
               {entityTypeNameSize !== null && (
                 <Controller
                   control={control}
@@ -166,7 +172,7 @@ export const EntityTypeHeader = ({
                   }}
                 />
               )}
-              <Box sx={{ ml: 2.5 }}>
+              <Box sx={{ ml: 2.5, minWidth: 0 }}>
                 <Tooltip
                   placement="top-start"
                   componentsProps={{
@@ -196,9 +202,7 @@ export const EntityTypeHeader = ({
                     variant="h1"
                     fontWeight="bold"
                     ref={entityTypeNameTextRef}
-                    sx={{
-                      lineHeight: 1.2,
-                    }}
+                    sx={{ overflow: "hidden", textOverflow: "ellipsis" }}
                   >
                     {entityTypeSchema.title}
                     {isInSlide && (
--- a/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
@@ -215,8 +215,12 @@ export const EntityTypeHeader = ({
                               fill: ({ palette }) => palette.blue[50],
                               fontSize: 24,
                               verticalAlign: "middle",
+                              // Nudge the icon up slightly relative to the
+                              // text, without affecting line-box layout
+                              position: "relative",
+                              top: -8,
                               "&:hover": {
-                                fill: ({ palette }) => palette.blue[70],
+                                fill: ({ palette }) => palette.blue[60],
                               },
                             }}
                           />
--- a/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
@@ -236,7 +236,7 @@ export const EntityTypeHeader = ({
               gap={1.5}
               sx={{ position: "relative", top: 5 }}
             >
-              <EntityTypePlural isLinkType={isLink} readonly={isReadonly} />
+              <EntityTypePlural isLinkType readonly={isReadonly} />
               {isLink && <EntityTypeInverse readonly={isReadonly} />}
             </Stack>
           </Stack>
--- a/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
+++ b/apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
@@ -106,15 +106,10 @@ export const EntityTypeHeader = ({
           direction="row"
           alignItems="center"
           justifyContent="space-between"
+          mb={3}
         >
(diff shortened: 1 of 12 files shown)


apps/hash-frontend/src/pages/shared/entity-type-page/shared/entity-type-header.tsx
```diff
@@ -195,39 +202,37 @@ export const EntityTypeHeader = ({
...
+                              position: "relative",
+                              top: -8,
+                              "&:hover": {
+                                fill: ({ palette }) => palette.blue[60],
+                              },
+                            }}
+                          />
+                        </Link>
+                      </TextMark>
+                    )}
```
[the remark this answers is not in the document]
alex-e-leon: true - but unlikely to happen, and IMO not worth fixing right now. Requires that the last line of content be super long and not be able to be broken onto 2 lines. Could be fixed by allowing word-break (ugly) or moving the icon back out (requires re-aligning all the content)

apps/hash-frontend/src/pages/shared/entity-type.tsx
```diff
@@ -480,7 +481,25 @@ export const EntityType = ({
...
+                 * instead of the sidebar-state based padding used on regular
+                 * pages.
+                 */
+                isInSlide
+                  ? {
+                      pt: `${backForwardHeight}px`,
+                      pl: 4,
+                      pr: 3,
+                    }
+                  : {},
```
CiaranMn: The breadcrumbs are intentionally vertically aligned with the slide buttons (i.e. at the top), can we instead push the context dots left if in a slide? If that's a pain (as margin would have to be dynamic based on presence of back/forward buttons), just omit the context dots in the slideover. People can go to the main page to archive.

apps/hash-frontend/src/pages/shared/entity-type.tsx
CiaranMn: Not related to your changes, but something I noticed we might as well fix now (diff doesn't show up properly but you get the idea)
```suggestion
                  href: `/types/${isLink ? "link" : "entity"}-type`,
```

CiaranMn (commented):
Definitely an improvement! One comment on the diff and a couple of styling things that we'd ideally fix while touching this:

1. Title text can be cut off (due to line height probably and no overflow?) – see the 'g' below
2. Needs some margin with the description

Image

3. For non link types where the plural title spans multiple rows, 'plural' looks a bit close to the actual plural title... maybe we should just use the layout for link types (above) always, i.e. just stack 'plural' and the plural title vertically rather than putting on the same horizontal row

Image

apps/hash-frontend/src/pages/shared/entity-type.tsx
alex-e-leon: Sure thing - only reason I did it this way was to keep it aligned to your og screenshot (and minorly because the close button is absolutely positioned, the padding has to be hand tweaked, ie if the close button size changes, it could introduce bugs - but not a big deal in the short term, and long term we'll just have a better dialog component without the issue)

apps/hash-frontend/src/pages/shared/entity-type.tsx
alex-e-leon: np - I'll apply it

alex-e-leon:
> Definitely an improvement! One comment on the diff and a couple of styling things that we'd ideally fix while touching this:

Sure thing, I'll fix these up. How is is that the base line-height is so bad : (

vilkinsons:
Can we make sure the vertical ellipsis icon/breadcrumbs always sit on the same row as the close button when the latter is present? I thought we'd fixed this prior.

apps/hash-frontend/src/pages/shared/slide-stack/slide-back-forward-close-bar.tsx
```diff
@@ -29,7 +29,6 @@ export const SlideBackForwardCloseBar = ({
         pointerEvents: "auto",
         pt: 1.5,
         pr: 3,
-        zIndex: 2,
```
alex-e-leon, line 32: @CiaranMn - I've removed this, as otherwise the close button sits on top of the breadcrumbs and menu, making the menu unclickable.

IMO this is a better approach, but it looks like in a previous PR you introduced this css - so its possible that you did this to fix a stacking issue in another drawer?

There's other approaches I could use to fix this, so wanted to check whether you had any context on this css and whether this change might be a regression - in which case I'd probably just remove the flex container from the close button so that it does not prevent clicks on the whole top row with zIndex of 2 set.

libs/@hashintel/design-system/src/theme/typography.ts
```diff
@@ -19,7 +19,7 @@ export const typography: ThemeOptions["typography"] = {
   h1: {
     fontFamily: headingFonts,
     fontSize: "var(--step-5)",
-    lineHeight: 1.1,
+    lineHeight: 1.2,
```
alex-e-leon, line 22: @CiaranMn - this updates the line-height for h1s used across the app. IMO the line-height of a text should never be so small that ellipsifying the text causes text cropping - so this feels better - but this is a global change. Looks like this was set by Ugbala back when setting up the DS, so I think this was just an early typography choice, but wanted to check before applying a global change like this if you had any thoughts.

Option b is I could just fix the line-height locally in entity-header - but I think that this is a better approach if there's no issues with it.

alex-e-leon:
@CiaranMn I've fixed up the other issues you mentioned - let me know if there's anything else. Also left a few comments where I made some choices with non-local impacts. LMK your thoughts on those before I apply them.

apps/hash-frontend/src/pages/shared/slide-stack/slide-back-forward-close-bar.tsx
CiaranMn: I don't think this is required, I can't find any context in which its lack is a problem

CiaranMn (commented):
One outstanding issue, which is that the context menu can overlap the back/forward/close bar. I mentioned before that we could potentially just hide the context menu dots when in a slide. Otherwise its right offset will need to vary based upon which additional buttons are visible (back, forward, close)

Image

alex-e-leon:
> One outstanding issue, which is that the context menu can overlap the back/forward/close bar. I mentioned before that we could potentially just hide the context menu dots when in a slide. Otherwise its right offset will need to vary based upon which additional buttons are visible (back, forward, close)

Ok fixed in the last commit.

CiaranMn (approved):

PR #8863, opened 2026-06-12
FE-866: Update Petrinaut + refractive toolchain

kube:
> Back to Draft: Will wait for TypeScript 7 RC and Rolldown with React Compiler support to be available next week.

## Purpose

Toolchain and dependency sweep for the Petrinaut packages: `@hashintel/petrinaut`, `@hashintel/petrinaut-core`, `@hashintel/refractive`, `@apps/petrinaut-website`.

## What changed

**Toolchain (all 4 packages)**
- Vite 8.0.12 → 8.0.16, rolldown 1.0.3 (pinned to Vite's internal version), `@vitejs/plugin-react` 6.0.2
- `@typescript/native-preview` (tsgo) 20260511 → 20260610 (nightly; 5-day age gate applied)
- `rolldown-plugin-dts` 0.25.2, `@rolldown/plugin-babel` 0.2.3
- Storybook 10.4.3, vitest 4.1.8
- oxlint 1.69.0 + oxlint-tsgolint 0.23.0
- `@babel/core` added explicitly (undeclared peer of the React Compiler babel setup)

**React and types**
- react/react-dom 19.2.6 → 19.2.7 (monorepo-wide — root `resolutions` pin forces a single copy)
- `@types/react` 19.2.17

**petrinaut-core**
- `typescript` 5.9.3 → 6.0.3: this is the in-browser LSP language service bundled into the worker, not the monorepo type-checker (which stays on tsgo)

**Runtime deps (petrinaut-scope-only)**
- `@ai-sdk/react` 3.0.201, `@xyflow/react` 12.11.0, `react-resizable-panels` 4.11.2, `@babel/standalone` 7.29.7
- `react-icons` 5.6.0, `@whatwg-node/server` 0.11.0 (website devDep)

**Accessibility fixes surfaced by oxlint 1.69.0**

5 `aria-label` additions in petrinaut — the new oxlint version fires `jsx-a11y/control-has-associated-label` on elements the previous version missed:
- Spreadsheet cell editor (`type="number"`)
- Search panel input
- Simulation timeline range scrubber
- Scenario/metric description textareas (false-positives: both have a properly wired `htmlFor`/`id` pair, but oxlint can't verify template-literal equality statically)

**Why the dts `enforce: "pre"` remap stays**

Tested the upstream-recommended `oxc: { exclude }` alternative (rolldown-plugin-dts#201) with vite 8.0.16 + rolldown-plugin-dts 0.25.2 — it produces silent broken output: `petrinaut`'s `main.d.ts` comes out as a bare re-export stub, and `petrinaut-core`'s `index.d.d.ts` is not emitted at all. The existing `enforce: "pre"` remap works correctly. Both vite configs now have a comment documenting the test.

**React Compiler stays on Babel**

The Rust port is vendored in oxc 0.135 but rolldown's integration PR (rolldown#9671) is still a draft; stable Vite pins rolldown 1.0.3 anyway.

## Skipped / out of scope

| Package | Reason |
|---|---|
| `vscode-languageserver-types` 3.18.0 | Breaking: `Diagnostic.message` widened to `string \| MarkupContent`; needs code changes in Monaco/JSX layers first |
| `immer` 10→11, `@mantine/hooks` 8→9 | Major version bumps, also used by `hash-frontend`/`ds-components`/`hash-api` |
| `ai`, `@ai-sdk/openai`, `@sentry/react`, `elkjs` | Also in `hash-frontend` — constraint conflict |
| `@ark-ui/react`, `@tanstack/react-form`, `@pandacss/dev` | Also in `ds-components` — constraint conflict |
| `@monaco-editor/react` | On `next` tag (4.8.0-rc.3); stable 4.7.0 would be a downgrade |
| `vitest` 4.1.9, `storybook` 10.4.4+, `rolldown` 1.1.x | Don't clear the 5-day age gate |

This PR:

The changes in this PR:

The changes in this PR:

## Tests

`turbo run build test:unit lint:tsc lint:eslint` for all 4 packages — green. Storybook `--smoke-test` passes on petrinaut and refractive.


yarn.config.cjs
```diff
@@ -32,6 +32,12 @@ const ignoredDependencies = [
   "vitest",
   "@dnd-kit/sortable",
   "@babel/core",
+  // Petrinaut + refractive use @types/react 19.2.17; other workspaces are on 19.2.14
+  "@types/react",
+  // petrinaut-core bundles TypeScript 6 into its LSP worker; the rest of the repo is on 5.9
+  "typescript",
+  // Petrinaut tests run on jsdom 29; hash-ai-worker-ts is still on 24
+  "jsdom",
```
CiaranMn, line 40: Why can't the other workspaces also be upgraded?

Adding things to this array is a very blunt instrument which means we don't get any linting against packages ending up on different versions, we should only use it if there is some real difficulty in just upgrading everything else at the same time.

libs/@hashintel/petrinaut/package.json
```diff
@@ -63,41 +63,42 @@
     "@hashintel/refractive": "workspace:^",
     "@monaco-editor/react": "4.8.0-rc.3",
     "@tanstack/react-form": "1.29.0",
-    "@xyflow/react": "12.10.1",
+    "@xyflow/react": "12.11.0",
     "ai": "6.0.182",
```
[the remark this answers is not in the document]
CiaranMn: Agreed

yarn.config.cjs
kube: Because it would be a lot of work (for TypeScript) to make sure that all packages still pass the linting, and update them without worrying that we wouldn't break things.

For Petrinaut, it's just a runtime dependency.

CiaranMn: And for the others? And what is the motivation for upgrading Petrinaut to TypeScript 6?

CiaranMn: There should be really good reasons for just giving up on version consistency for packages

alex-e-leon:
@kube - I'm super comfortable for you to update any deps/dev deps in ds-components any time you like. The only 2 danger zone dependencies in ds-components in my view are ark-ui + panda-css. The risk of breaking things with any of the other deps is close to none.

yarn.config.cjs
kube: - Petrinaut lints User Code with TypeScript
- Other packages lint their own code with TypeScript

So I think it's safe to assume they are not the same version, and make sure that new user-code is linted with latest language.

PR #8866, opened 2026-06-12
BE-597: Array-based type filtering in the SelectCompiler

TimDiekmann:

Type filters (`type.versionedUrl` / `type.baseUrl` without an inheritance depth) compiled to one `entity_is_of_type → ontology_temporal_metadata → ontology_ids` join chain **per predicate**. On the trace that started this investigation (7 type exclusions), that meant 7 multiplying join groups, a planner misestimate, and a row-explosion the query had to `DISTINCT` away again.

This PR compiles those filters to array predicates (`@>` / `&&`) on the `entity_edition_cache` type arrays (from #8854) instead. Equality filters in the same `All`/`Any` group bundle into a single predicate over one parameter array, which the GIN index on the cache columns can serve.

## 🔗 Related links

- [BE-597](https://linear.app/hash/issue/BE-597/array-based-type-filtering-in-selectcompiler)

## 🚫 Blocked by

- #8864

## 🔍 What does this change?

- `BinaryOperator::ArrayContains` (`@>`) in the expression AST; `&&` (overlap) already existed.
- Path mapping (`query/entity.rs`): `EntityTypeEdge { IsOfType, BaseUrl | VersionedUrl, inheritance_depth: None }` resolves directly to the `entity_edition_cache` array columns. Explicit inheritance depths keep the join chain (the cache arrays are all-depth).
- The redundant `EntityQueryPath::TypeBaseUrls` / `TypeVersionedUrls` variants are deleted; the edge form is the single canonical path (they were documented as not queryable from the API, so no public surface changes).
- Compiler rewrite (`query/compile.rs`), type-driven via the column's declared `ParameterType` rather than hardcoded paths:
  - `Equal`/`NotEqual`/`In(parameter, path)` on text-array columns compile to containment predicates.
  - `All`/`Any` groups bundle same-column equality filters: `All`+equals → `@>`, `Any`+equals → `&&`, `All`+notEquals → `NOT(&&)`, `Any`+notEquals → `NOT(@>)` — each an exact logical equivalent of the unbundled forms.
  - Bundles are keyed on the **aliased** column, so an entity's own type and a linked entity's type (same column, different join chain) stay separate predicates.
  - `StartsWith`/`EndsWith`/`ContainsSegment` on array-backed paths fail at compile time (`UnsupportedTextArrayOperation`) instead of producing a Postgres type error.
- `Filter::for_entity_by_type_id` builds a single `versionedUrl` equality instead of `All[baseUrl, version]` (the old shape was only correct through shared join aliases). All `IsOfType`/`IsOfBaseType` policy filters now go through the array path.
- The property-protection filters (`In(param, type base URLs)`) compile to `base_urls @> ARRAY[$n]` instead of `$n = ANY(base_urls)` — semantically equivalent, but GIN-indexable.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- **Intentional behaviour change:** `notEqual` on type paths now uses set semantics — "the entity has **no** such type" — instead of the old per-join-row semantics, under which a multi-typed entity matched `notEqual(A)` as long as it had *any* other type. The old behaviour made type exclusions effectively useless for multi-typed entities (an entity with types `{notification, X}` stayed visible despite a `notEqual(notification)` filter); the new behaviour is what the filters are meant to express. The same set semantics applies to separate `baseUrl`/`version` equality filters within one `all` block.
- Nested paths through link edges (`leftEntity.type.versionedUrl` etc.) intentionally keep the join chain for the *link* traversal; only the terminal type lookup uses the cache.

## 🐾 Next steps

- BE-598: keys-then-hydrate split (the remaining cost is the label-sort scan, not the type predicates).
- Integration test pinning the `notEqual` exclusion semantics on a multi-typed entity.
- Filter normalization pass (flattening nested `All`/`Any`, sibling dedup) to widen bundling across artificial nesting and improve planner estimates — separate ticket.

## 🛡 What tests cover this?

- New compiler snapshot tests in `statement/select.rs`: single equality (constructor + raw filter forms), `Any`/`All` bundling, bundled `NOT(&&)` exclusion, base-URL policy form, and the own-type vs. linked-type alias-separation regression test.
- Updated property-masking snapshot (`= ANY` → `@>`).
- `email_filter_protection.rs` integration tests cover the rewritten protection-filter path in CI.
- Manually validated against an 8.8M-entity dataset: counts and all summary maps byte-identical to the join-based compilation.

## ❓ How to test this?

1. Checkout the branch, start the graph with a seeded database.
2. Run a `queryEntities` request with type filters (`equal`/`notEqual` on `["type", "versionedUrl"]`, with and without an `any`/`all` group).
3. Confirm the compiled SQL (debug logs) contains `entity_edition_cache` array predicates instead of `entity_is_of_type` joins, and results match the base branch — except multi-typed entities under `notEqual`, which are now correctly excluded.

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/compile.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/compile.rs
@@ -16,7 +16,7 @@ use postgres_types::ToSql;
 use tracing::instrument;
 use type_system::knowledge::Entity;
 
-use super::expression::{JoinType, TableName, TableReference};
+use super::expression::{ColumnReference, JoinType, TableName, TableReference};
 use crate::store::postgres::query::{
     Alias, Column, Distinctness, EqualityOperator, Expression, Function, Identifier,
     PostgresQueryPath, PostgresRecord, SelectExpression, SelectStatement, Table, Transpile as _,
@@ -92,6 +92,8 @@ pub enum SelectCompilerError {
     UnsupportedDistanceExpression,
     #[display("Cannot add a cursor: {reason}")]
     CursorDisallowed { reason: &'static str },
+    #[display("String operations are not supported on paths backed by materialized array columns")]
+    UnsupportedTextArrayOperation,
 }
 
 impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
@@ -429,18 +431,8 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
         }
 
         Ok(match filter {
-            Filter::All(filters) => Expression::all(
-                filters
-                    .iter()
-                    .map(|filter| self.compile_filter(filter))
-                    .collect::<Result<_, _>>()?,
-            ),
-            Filter::Any(filters) => Expression::any(
-                filters
-                    .iter()
-                    .map(|filter| self.compile_filter(filter))
-                    .collect::<Result<_, _>>()?,
-            ),
+            Filter::All(filters) => Expression::all(self.compile_filter_group(filters, true)?),
+            Filter::Any(filters) => Expression::any(self.compile_filter_group(filters, false)?),
             Filter::Not(filter) => self.compile_filter(filter)?.not(),
             Filter::Equal(lhs, rhs) => Expression::equal(
                 self.compile_filter_expression(lhs).0,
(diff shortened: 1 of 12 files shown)


libs/@local/graph/postgres-store/src/store/postgres/query/compile.rs
```diff
@@ -697,6 +695,202 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
...
+    /// Compiles the filters of an `All`/`Any` group, bundling equality filters backed by
+    /// the same materialized array column into a single array predicate.
+    ///
+    /// Bundles are keyed on the *aliased* column: paths terminating in the same column
+    /// through different join chains (e.g. an entity's own types vs. a linked entity's
+    /// types) resolve to different aliases and stay separate predicates.
+    fn compile_filter_group<'f: 'q>(
+        &mut self,
+        filters: &'p [Filter<'f, R>],
+        conjunction: bool,
```
indietyp: would it make sense to move this to an enum instead to increase readability?

TimDiekmann: Yes, good point. Fixed in [`8ef9c06` (this PR)](https://github.com/hashintel/hash/pull/8866/commits/8ef9c0649a199ed92dceda312350bdaadef71639)

indietyp (approved):

PR #8877, opened 2026-06-17
FE-851: seed data for hash app

alex-e-leon:

The previous seed data sets that I used/were provided to me didn't include enough edge case data for stress testing the ui, so I've gone and built a new seed-data set from scratch that includes more edge cases.

I ended up opting to design the data set around a sales CRM, as a) I think its a potential realistic use-case for hash, with the potential to model various sales/lead generation pipelines down the line to test those features, and b) its an easy category to create fake data for.

I've also opted for generating completely fake data so that there's no issues around data-provenence.

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/package.json
+++ b/apps/hash-api/package.json
@@ -9,7 +9,7 @@
     "build:docker": "docker buildx build --tag hash-api --file docker/Dockerfile ../../ --load",
     "codegen": "rimraf './src/**/*.gen.*'; graphql-codegen --config codegen.config.ts",
     "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=2048 tsx watch --clear-screen=false --import ./src/instrument.mjs ./src/index.ts",
-    "dev:seed-data": "NODE_ENV=development tsx ./src/seed-data/seed-crm-data.ts",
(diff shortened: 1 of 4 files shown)


apps/hash-api/package.json
```diff
@@ -9,6 +9,7 @@
     "build:docker": "docker buildx build --tag hash-api --file docker/Dockerfile ../../ --load",
     "codegen": "rimraf './src/**/*.gen.*'; graphql-codegen --config codegen.config.ts",
     "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=2048 tsx watch --clear-screen=false --import ./src/instrument.mjs ./src/index.ts",
+    "dev:seed-data": "NODE_ENV=development tsx ./src/seed-data/seed-crm-data.ts",
```
CiaranMn: There's potential for confusion here with the rest of the content of the `seed-data` folder, which runs whenever the API is started in development. 

I would rename this to `dev:seed-crm-data` to avoid that potential confusion and for alignment with the script being run.

CiaranMn (commented):
Very helpful addition! One comment on the command

apps/hash-api/package.json
alex-e-leon: Good call - done.

CiaranMn (approved):

PR #8891, opened 2026-06-22
FE-1048: Petrinaut use ds toggle

alex-e-leon:

Replaces Toggle in Petrinaut with toggle from the DS
@kube, the new Toggle has a few new tones(colors): neutral + brand, with neutral being the default - I've switched over a couple of the toggles to be neutral as a result, in particular, where the toggle behaves more like a form element rather than a setting, or where the success (green) variation felt too heavy/distracting.

The one controversial change in tone is likely in the user-settings dialog, since there are so many toggles on that dialog, making them neutral felt less visually distracting than all success.

Let me know if you prefer me to change any of the tones, and any other feedback also appreciated.

This PR:

The changes in this PR:


kube (commented):
## Alignment

The thumbs are not properly aligned with the "rail":

Image

## Green vs Dark Gray

I personally prefer green for enablement, but this is just my opinion.

Image

Would be worth checking with designers.
If you did it already, go with the dark gray.


## "Refraction"

Refraction was already removed, now we remove the transform while dragging.

I wonder how we'll put refraction back.

kube (approved):
Perfect thanks!

PR #8901, opened 2026-06-24
BE-444: Fix email verification failing for uppercase signups

TimDiekmann:

Signing up with a mixed-case email (e.g. `T@HASH.ai`) left the user stuck on the email-verification screen: Kratos delivered the code to the lowercased address and accepted it (consuming it), but the account never unlocked. The cause was **case-sensitive email comparisons** between Kratos `verifiable_addresses[].value` (the casing typed at signup) and a normalized/lowercased source — and that comparison logic was **duplicated across the frontend and backend** and had drifted.

This PR makes all email comparisons case-insensitive via a single shared helper, fixing the reported signup flow and several adjacent latent bugs of the same class.

## 🔗 Related links

- [BE-444](https://linear.app/hash/issue/BE-444/email-verification-code-fails-when-signing-up-with-uppercase-email) _(internal)_
- Original report & Kratos/app logs in the Slack thread _(internal)_

## 🔍 What does this change?

- Adds a shared `normalizeEmail` helper (`@local/hash-isomorphic-utils/normalize`, `.trim().toLowerCase()`) imported by both frontend and backend, so the comparison logic can no longer drift.
- **Frontend (the user-facing gate):** normalizes both sides of the verification check in `getPrimaryEmailVerificationStatus` (`_app.page.tsx`) and `isPrimaryEmailAddressVerified` (`user-and-org.ts`). This is what unblocks the stuck verification screen.
- **Backend:** normalizes emails read from Kratos at the single `getUserFromEntity` funnel and in `getVerifiedEmailsFromKratosIdentity`, the allowlist on parse (`user-has-access-to-hash.ts`), and `checkEmailVerificationAndUsageStatus`.
- **Cleanup job:** `isPrimaryEmailVerified` now compares case-insensitively — previously a verified mixed-case account could be classed as unverified and **archived + its Kratos identity deleted** (data loss).
- Removes the dead `req.primaryEmailVerified` machinery (computed in `create-auth-handlers`, never read anywhere; `express.d.ts` field included).

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Emails are normalized **on read** (`getUserFromEntity`, the Kratos read helpers) rather than **on write** — the user entity still stores the email in its original casing. This is safe because every comparison path now normalizes (TS via `normalizeEmail`, and the graph admin delete-by-email already lowercases both sides in SQL). Flagged so a future exact-match graph filter against the stored `email` property doesn't get caught out.

## 🐾 Next steps

- If stored canonicalization is ever wanted, normalize at write time in `createUser` (would also cover any future stored-value exact-match query).

## 🛡 What tests cover this?

- **New e2e** (`signup.spec.ts`): registers with an uppercased allowlisted email, verifies via Mailslurper, completes signup. Demonstrably failed pre-fix (stuck at the verify gate, then at completion).
- **New unit** (`normalize.test.ts`): pins the `normalizeEmail` contract (trim + lowercase, idempotent, no local-part canonicalization).
- **New unit** (`create-unverified-email-cleanup-job.test.ts`): guards the data-loss path — a verified mixed-case identity is recognized as verified.
- Existing signup e2e (lowercase + waitlist) still pass — no regression.

## ❓ How to test this?

1. Check out the branch and run the test stack (Kratos + Mailslurper + API + frontend).
2. Sign up with a mixed-case email such as `Foo@Example.com`.
3. Enter the verification code from Mailslurper.
4. Confirm you advance past the verification screen and can complete signup (set shortname + display name) rather than being stuck on "Verify your email address".

Automated: `turbo run test:unit --filter '@local/hash-isomorphic-utils' --filter '@apps/hash-api'` and the Playwright `signup` spec on `account-chromium`.

## 📹 Demo

n/a — backend + frontend logic fix; covered by the e2e regression test above.

### DIFF
--- a/apps/hash-api/src/graph/knowledge/system-types/user.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/user.ts
@@ -172,11 +172,18 @@ export const getUserFromEntity: PureGraphFunction<
 
   const {
     displayName,
-    email: emails,
+    email,
     enabledFeatureFlags: maybeFeatureFlags,
     kratosIdentityId,
     shortname,
   } = simplifyProperties(entity.properties);
+
+  // `email` is typed as a required property, but property-level permission
+  // masking can drop it at runtime for non-owners (see `getUser`'s Kratos
+  // back-fill). Model that nullability, then canonicalise so every `User` built
+  // from an entity compares case-insensitively regardless of signup casing.
+  const emails =
+    (email as [string, ...string[]] | undefined)?.map(normalizeEmail) ?? [];
 
   const isAccountSignupComplete = !!shortname && !!displayName;
 
@@ -190,11 +197,7 @@ export const getUserFromEntity: PureGraphFunction<
       entity.metadata.recordId.entityId,
     ) as UserId,
     displayName,
-    // Canonicalise at the single point every `User` is built from an entity, so
-    // case-insensitive comparisons hold regardless of the casing stored at
-    // signup.
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- the non-nullable type lies; property-level masking drops `email` at runtime
-    emails: emails?.map(normalizeEmail) as typeof emails,
+    emails,
     enabledFeatureFlags,
     entity,
     isAccountSignupComplete,
--- a/apps/hash-api/src/graph/knowledge/system-types/user.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/user.ts
@@ -183,7 +183,7 @@ export const getUserFromEntity: PureGraphFunction<
   // back-fill). Model that nullability, then canonicalise so every `User` built
   // from an entity compares case-insensitively regardless of signup casing.
   const emails =
-    (email as [string, ...string[]] | undefined)?.map(normalizeEmail) ?? [];
+(email as typeof email | undefined)?.map(normalizeEmail) ?? [];
 
   const isAccountSignupComplete = !!shortname && !!displayName;
 
--- a/apps/hash-api/src/graph/knowledge/system-types/user.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/user.ts
@@ -182,8 +182,7 @@ export const getUserFromEntity: PureGraphFunction<
   // masking can drop it at runtime for non-owners (see `getUser`'s Kratos
   // back-fill). Model that nullability, then canonicalise so every `User` built
   // from an entity compares case-insensitively regardless of signup casing.
-  const emails =
-    (email as typeof email | undefined)?.map(normalizeEmail) ?? [];
+  const emails = (email as typeof email | undefined)?.map(normalizeEmail) ?? [];
 
   const isAccountSignupComplete = !!shortname && !!displayName;
 
--- a/apps/hash-api/src/graph/knowledge/system-types/user.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/user.ts
@@ -183,7 +183,7 @@ export const getUserFromEntity: PureGraphFunction<
   // back-fill). Model that nullability, then canonicalise so every `User` built
   // from an entity compares case-insensitively regardless of signup casing.
   const emails =
-(email as typeof email | undefined)?.map(normalizeEmail) ?? [];
+    (email as typeof email | undefined)?.map(normalizeEmail) ?? [];
 
   const isAccountSignupComplete = !!shortname && !!displayName;
 
--- a/apps/hash-api/src/graph/knowledge/system-types/user.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/user.ts
@@ -24,6 +24,7 @@ import {
   currentTimeInstantTemporalAxes,
   generateVersionedUrlMatchingFilter,
 } from "@local/hash-isomorphic-utils/graph-queries";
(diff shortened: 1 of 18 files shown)


apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -187,7 +190,11 @@ export const getUserFromEntity: PureGraphFunction<
       entity.metadata.recordId.entityId,
     ) as UserId,
     displayName,
-    emails,
+    // Canonicalise at the single point every `User` is built from an entity, so
+    // case-insensitive comparisons hold regardless of the casing stored at
+    // signup.
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- the non-nullable type lies; property-level masking drops `email` at runtime
+    emails: emails?.map(normalizeEmail) as typeof emails,
```
alex-e-leon: Rather than adding a type coercion, can we just fix this type? Otherwise this is just a bug waiting to happen in my experience.

tests/hash-playwright/tests/shared/get-kratos-verification-code.ts
```diff
@@ -83,6 +83,15 @@ const pollForKratosCode = async ({
   const maxWaitMs = 10_000;
   const pollIntervalMs = 250;
   const timestampBufferMs = 5_000;
+  // Match recipients case-insensitively: a user may sign up with an uppercased
+  // address while Kratos delivers the email to the lowercased one.
+  const normalizedTarget = emailAddress.toLowerCase();
```
alex-e-leon: Why not use normalizeEmail here?

alex-e-leon (dismissed):
A few minor comments, but LGTM otherwise - might still be worth getting a review from someone more familiar with this code though.

apps/hash-api/src/graph/knowledge/system-types/user.ts
TimDiekmann: Done in [`4b289d5` (this PR)](https://github.com/hashintel/hash/pull/8901/commits/4b289d5c6fcfcfd59f7057dfc880fb052521c060)

tests/hash-playwright/tests/shared/get-kratos-verification-code.ts
TimDiekmann: I didn't implemented the `normalizeEmail` yet when I have changed this code 😅 Thanks for spotting!
[`4b289d5` (this PR)](https://github.com/hashintel/hash/pull/8901/commits/4b289d5c6fcfcfd59f7057dfc880fb052521c060)

alex-e-leon (dismissed):

apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -172,12 +172,19 @@ export const getUserFromEntity: PureGraphFunction<
...
     kratosIdentityId,
     shortname,
   } = simplifyProperties(entity.properties);
 
+  // `email` is typed as a required property, but property-level permission
+  // masking can drop it at runtime for non-owners (see `getUser`'s Kratos
+  // back-fill). Model that nullability, then canonicalise so every `User` built
+  // from an entity compares case-insensitively regardless of signup casing.
+  const emails =
+    (email as [string, ...string[]] | undefined)?.map(normalizeEmail) ?? [];
```
CiaranMn: Guards against the type of `email` changing
```suggestion
(email as typeof email | undefined)?.map(normalizeEmail) ?? [];
```

apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -169,12 +172,19 @@ export const getUserFromEntity: PureGraphFunction<
...
     kratosIdentityId,
     shortname,
   } = simplifyProperties(entity.properties);
 
+  // `email` is typed as a required property, but property-level permission
+  // masking can drop it at runtime for non-owners (see `getUser`'s Kratos
+  // back-fill). Model that nullability, then canonicalise so every `User` built
+  // from an entity compares case-insensitively regardless of signup casing.
+  const emails =
+(email as typeof email | undefined)?.map(normalizeEmail) ?? [];
```
CiaranMn:
```suggestion
    (email as typeof email | undefined)?.map(normalizeEmail) ?? [];
```

CiaranMn: Sorry, misformatted because GitHub didn't give me the suggestion option first and I typed it manually

CiaranMn (dismissed):

apps/hash-api/src/graph/knowledge/system-types/user.ts
```diff
@@ -169,12 +172,19 @@ export const getUserFromEntity: PureGraphFunction<
...
     shortname,
   } = simplifyProperties(entity.properties);
 
+  // `email` is typed as a required property, but property-level permission
+  // masking can drop it at runtime for non-owners (see `getUser`'s Kratos
+  // back-fill). Model that nullability, then canonicalise so every `User` built
+  // from an entity compares case-insensitively regardless of signup casing.
+  const emails =
+    (email as typeof email | undefined)?.map(normalizeEmail) ?? [];
+
```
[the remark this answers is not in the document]
TimDiekmann: Added an integration test to prove this is a false positive.

CiaranMn (approved):

PR #8902, opened 2026-06-24
H-3665: Paginate search bar

alex-e-leon:

Implements pagination on the search bar + refactors accumulate-pagination to be much simpler to use and reason about

This PR:

The changes in this PR:


alex-e-leon:
Closing, as the search api is not able to be paginated. Instead we are applying a simple limit to the search query for now in https://github.com/hashintel/hash/pull/8886

PR #8903, opened 2026-06-25
BE-580: Inline type filters onto the edition cache; rework summarizeEntities aggregation

TimDiekmann:

Two changes to how entity queries are compiled in the Postgres store:

- The `summarizeEntities` type aggregation no longer expands the cached type arrays with a per-row `CROSS JOIN LATERAL unnest`, and skips the `DISTINCT ON` deduplication when the compiled query cannot emit duplicate rows.
- Type-matching filters across `apps/` and `libs/` are inlined so they resolve against the materialized `entity_edition_cache` columns instead of the `entity_is_of_type` join.

## 🔗 Related links

- BE-580 _(internal)_

## 🔍 What does this change?

**summarizeEntities aggregation** (`hash-graph-postgres-store`):

- The type-ids branch expands the cache arrays via set-returning `unnest` in the `SELECT` list (`ProjectSet`) instead of `CROSS JOIN LATERAL unnest(...)`.
- Adds `Relation::is_to_many` and `SelectCompiler::has_to_many_join`. The `hits` CTE omits `DISTINCT ON` when no to-many (fan-out) join was added and the variable temporal axis is a collapsed point; otherwise it deduplicates as before. Selection is via a `Deduplication` enum (no boolean parameter).

**Type-filter inlining** (`apps/` + `libs/`):

- Removes `generateVersionedUrlMatchingFilter`. Every call site is inlined as a direct `{ equal: [{ path: ["type", "baseUrl"] }, { parameter: …entityTypeBaseUrl }] }` — `["type", "versionedUrl"]` for raw `VersionedUrl` inputs and version-sensitive sites, `…linkEntityTypeBaseUrl` for link types.
- Drops the `inheritanceDepth = 0` path qualifier: a bare `type` path resolves to the GIN-indexed `entity_edition_cache` `base_urls`/`versioned_urls` columns rather than the `entity_is_of_type` join. The remaining hand-written `inheritanceDepth = 0` filters (`machine-actors`, `org`) and a rustdoc example are migrated to the same form.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `hash-graph-postgres-store` unit tests: `statement_all_dimensions`, `statement_count_only`, `statement_skips_dedup`, `has_to_many_join_flag`.
- The inlined filters are covered by type-checking (`lint:tsc`).

## ❓ How to test this?

1. `cargo nextest run -p hash-graph-postgres-store`
2. `turbo run lint:tsc` for the affected TypeScript packages.

### DIFF
--- a/apps/hash-api/src/graph/knowledge/system-types/page.ts
+++ b/apps/hash-api/src/graph/knowledge/system-types/page.ts
@@ -16,7 +16,6 @@ import {
 } from "@local/hash-isomorphic-utils/ontology-type-ids";
 import {
   includesPageEntityTypeId,
-  pageEntityTypeFilter,
   pageEntityTypeIds,
 } from "@local/hash-isomorphic-utils/page-entity-type-ids";
 import { simplifyProperties } from "@local/hash-isomorphic-utils/simplify-properties";
@@ -278,7 +277,14 @@ export const getAllPagesInWorkspace: ImpureGraphFunction<
   const { entities: pageEntities } = await queryEntities(ctx, authentication, {
     filter: {
       all: [
-        pageEntityTypeFilter,
+        {
+          equal: [
+            { path: ["type", "baseUrl"] },
+            {
+              parameter: systemEntityTypes.page.entityTypeBaseUrl,
+            },
+          ],
+        },
         {
           equal: [{ path: ["webId"] }, { parameter: webId }],
         },
--- a/libs/@local/hash-isomorphic-utils/src/page-entity-type-ids.ts
+++ b/libs/@local/hash-isomorphic-utils/src/page-entity-type-ids.ts
@@ -1,4 +1,3 @@
-import { generateVersionedUrlMatchingFilter } from "./graph-queries.js";
 import {
   systemEntityTypes,
   systemLinkEntityTypes,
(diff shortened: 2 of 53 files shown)


libs/@local/hash-isomorphic-utils/src/page-entity-type-ids.ts
```diff
@@ -35,12 +34,15 @@ export const pageEntityTypeFilter = {
   /**
    * We specify each of these page types individually rather than Page, which they both inherit from,
    * because checking against types involving inheritance is currently slow.
```
CiaranMn: This comment is no longer valid. We could replace the filter below with a single check against `systemEntityTypes.page.entityTypeId`. Or just delete this whole comment.

TimDiekmann: I removed the function and replaced the caller by a proper filter.

apps/hash-api/src/graph/knowledge/system-types/page.ts
```diff
@@ -278,7 +277,14 @@ export const getAllPagesInWorkspace: ImpureGraphFunction<
...
       all: [
-        pageEntityTypeFilter,
+        {
+          equal: [
+            { path: ["type", "baseUrl"] },
+            {
+              parameter: systemEntityTypes.page.entityTypeBaseUrl,
+            },
+          ],
+        },
```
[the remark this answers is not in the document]
TimDiekmann: I agree the assertion is not ideal. However, it still checks `includesPageEntityTypeId` and should result in the same behavior.

CiaranMn (approved):

PR #8904, opened 2026-06-26
BE-620: Add read-only entity functionality

TimDiekmann:

Let integrations seed entities that users can read but not modify, without creating a policy per entity (which would be millions). Adds a `read_only` flag to entities, enforced by a single authorization policy.

## 🔍 What does this change?

- New first-class `EntityMetadata.read_only` flag, stored on `entity_ids` (entity-level, set once at creation, preserved across editions). Deliberately not in provenance — it's a policy attribute, not history.
- Set at creation via `CreateEntityParams.read_only`.
- Global `forbid-user-modify-readonly-entity` policy: user actors cannot `UpdateEntity`/`ArchiveEntity` a read-only entity. Machine actors are unaffected, and `CreateEntity` stays open so users can still create their own entities of the same types.
- New `EntityResourceFilter::IsReadOnly` and `EntityQueryPath::ReadOnly` (internal, no public query token).
- Migration `V53` adds the column, plus inline in the `v009` shadow migration.

## 🛡 What tests cover this?

- `tests/graph/integration/postgres/entity.rs::read_only_entity_cannot_be_modified_by_user` — create a read-only entity, then assert a user actor's modification is denied.
- Cedar filter round-trip unit test for `IsReadOnly` in `hash-graph-authorization`.

## ❓ How to test this?

1. `cargo nextest run -p hash-graph-integration read_only_entity_cannot_be_modified_by_user` (test DB must be migrated to `V53`).
2. Confirm the user patch returns a permission error.


libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
```diff
@@ -2,6 +2,7 @@ CREATE TABLE entity_ids (
     web_id UUID NOT NULL REFERENCES web,
     entity_uuid UUID NOT NULL,
     provenance JSONB NOT NULL,
+    read_only BOOLEAN NOT NULL,
```
thehabbos007, line 5: I am not too up to date on the migrator, but wouldn't editing this existing migration be wrong, shouldn't we `alter table` to add this column in V53 instead? And probably provide a down sql snippet too?

thehabbos007 (commented):
Looks good, Just a single comment i would like to know the answer to.

libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
TimDiekmann: The `graph-migrations` folder was introduced as part of BE-21 but as we never finished that, I have to keep this and the `postgres-migrations` folder in sync. The latter is [actually altering the table](https://github.com/hashintel/hash/pull/8904/changes#diff-8c9fa254dd74efd8f4b0287475ef9d46262387852004aa7524b38a92a858d9b8). 
`graph-migrations` were supposed to replace the old plain-sql migrations with more functionality (more than just raw SQL). It currently is basically a shadow-copy of the other migrations. The postgres-integration-tests are using these migrations while the app still use the old migration approach.

I know it's confusing and really suboptimal, but we cannot really do much about it.

thehabbos007 (approved):
Works as expected within an integration run. Thank you so much!

PR #8912, opened 2026-06-29
FE-684: Add checkbox to DS

alex-e-leon:

Updates the checkbox component in the DS with a new implementation
Also removed a few unused dependencies in ds-components

This PR:

The changes in this PR:

### DIFF
--- a/.changeset/purple-lights-grin.md
+++ b/.changeset/purple-lights-grin.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/ds-components": minor
+---
+
+New UI and api for Checkbox component (breaking change)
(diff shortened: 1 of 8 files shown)


kube (approved):
LGTM.
Only checked in Petrinaut though.
Don't know how this `indeterminate` issue Cursor bot was talking about actually is. :)

alex-e-leon:
> LGTM.
> Only checked in Petrinaut though.
> Don't know how this `indeterminate` issue Cursor bot was talking about actually is. :)

In my experience indeterminate is never really used so it's kind of moot - but with this api you just need to remember to manually set indeterminate false if the user makes a selection. I've gone with this api, since it matches the w3c api better where you can have indeterminate + checked or unchecked

PR #8923, opened 2026-07-01
FE-1106: Add CheckboxGroup + RadioGroup to DS

alex-e-leon:

Added CheckboxGroup and RadioGroup to the design system.

I also upgraded ark-ui, as the previous version did not support some fields on radio group that we use like "required". Upgrading also fixes some existing minor issues I noticed (like menu's closing during some hover operations), and will be required anyway for some upcoming components like DateInput.

I've read through the ark-ui changelog, and tested the DS, petrinaut + the hash frontend to make sure that no breaking changes apply to our ark-ui usage.

Including @CiaranMn + @kube in this PR in case you have any concerns with the ark-ui upgrade.

This PR:

The changes in this PR:


kube (approved):
I checked Petrinaut and did not find any issue. 👍

kube:
I removed from Merge Queue, given it was my approval that triggered it, and I suppose the real review you want is from @lunelson.

lunelson (approved):

PR #8950, opened 2026-07-03
FE-1132: Add RightClickMenu to DS

alex-e-leon:

Adds a right click menu component to the DS

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/components/Menu/SelectableList/selectable-list.recipe.ts
+++ b/libs/@hashintel/ds-components/src/components/Menu/SelectableList/selectable-list.recipe.ts
@@ -19,12 +19,6 @@ export const styles = sva({
       minWidth: "[140px]",
       zIndex: "popover",
       transformOrigin: "var(--transform-origin)",
-      '&[data-state="open"]': {
-        animation: "popoverIn 120ms ease-out",
-      },
-      '&[data-state="closed"]': {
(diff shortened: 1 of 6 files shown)


lunelson (changes requested):
Two things jump out at me

1. you can't do it in a single press-release cycle, i.e. mouse-down, drag to the position of the item you want, release to select (as normal context menus can do)
2. I'm not a fan of zoom transitions on drop-down menus, I would use position only, or no transition at all

alex-e-leon:
> Two things jump out at me
> 
> 1. you can't do it in a single press-release cycle, i.e. mouse-down, drag to the position of the item you want, release to select (as normal context menus can do)
> 2. I'm not a fan of zoom transitions on drop-down menus, I would use position only, or no transition at all

@lunelson - thanks for the feedback. Agree with you about the menu animation, I've updated it so that Menu only has a subtle fade in/out animation instead.

Re the right click issue, agree with you again, but this behaviour is actually driven by ark-ui - to fix it I'd need to override/re-implement the behaviour, which took claude ~150 loc. Claude also mentioned that right click behaves the way we expect it on windows + linux, and that this is an OSX only issue. I haven't/can't test to see if thats correct, but assuming it is, I think all things considered its not worth fixing this for now.

lunelson (approved):

PR #8541, opened 2026-03-13
FE-514: Compile TypeScript expressions to SymPy

kube:
## 🌟 What is the purpose of this PR?

Adds the ability to compile Petrinaut's TypeScript expressions (rate laws, guard conditions, dynamics) into SymPy Python code — enabling interoperability with external scientific computing tools (TOPOS, Coherence Research). This is a demo-quality implementation proving the concept and establishing the TS→SymPy mapping.

## 🔗 Related links

- [FE-514: Compile TypeScript expressions to SymPy](https://linear.app/hash/issue/FE-514) _(internal)_

## 🔍 What does this change?

- **`compile-to-sympy.ts`** — AST-walking compiler that parses TypeScript expressions and emits SymPy equivalents. Supports arithmetic, comparisons, logical operators, ternary→`Piecewise`, `Math.*` functions/constants, `Distribution.*`, token/parameter access, object literals, `const` bindings, and type assertion unwrapping. Rejects unsupported syntax with positioned error diagnostics (`start`/`length`).
- **`compile-to-sympy.test.ts`** — 51 tests covering all supported syntax, real-world expressions (SIR model, orbital dynamics, transition kernels), and error cases with position assertions.
- **`checker.ts`** — Integrates SymPy compilation into the LSP checker. Runs `compileToSymPy` on all differential equations, transition lambdas, and transition kernels. Failures surface as warnings (code `99000`) with accurate source positions in the Diagnostics tab.
- **`export-sympy.ts`** — New export function that converts all model expressions to SymPy and produces a JSON file containing both the original SDCPN and a `sympy_expressions` array.
- **`editor-view.tsx`** — Adds "JSON with SymPy expressions" to the Export submenu.

### 🚢 Has this modified a publishable library?

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

### 📜 Does this require a change to the docs?

The changes in this PR:

- [x] are internal and do not require a docs change

### 🕸️ Does this require a change to the Turbo Graph?

The changes in this PR:

- [x] do not affect the execution graph

## ⚠️ Known issues

- This is a restricted subset compiler — not all valid TypeScript will produce SymPy output. Unsupported constructs (control flow, `let`, arbitrary function calls) are rejected with warnings rather than errors, since the TS code itself is still valid.

## 🐾 Next steps

- Python language support ([H-6323](https://linear.app/hash/issue/H-6323)) would complement this by providing the runtime for SymPy expressions.

## 🛡 What tests cover this?

[...]

PR #8955, opened 2026-07-04
FE-629: Compile Petrinaut user code through an HIR to buffer-native programs

kube:

Petrinaut user code (dynamics, firing rates, transition kernels, metrics) was executed by stripping TypeScript with Babel and `eval`-ing it against per-call decoded token objects. That made the code impossible to analyze statically — e.g. knowing the probability-distribution structure of a kernel, or whether a rate depends on parameters — and slow, since every evaluation decoded packed frame bytes into records and back.

This PR introduces an **HIR (high-level intermediate representation)**: a typed, source-spanned, JSON-serializable expression tree that every user-code surface compiles through. Goals:

- **One analyzable representation** for all user code, schema-checked against the net (token attribute types, places, arc weights, parameters).
- **Static analysis for free**: distribution DAGs inside kernels (derivation edges, output sinks, shared draws), dependency sets (which parameters/attributes a rate reads), determinism classification — surfaced as LSP diagnostics with exact source ranges.
- **Fast execution**: the simulator runs *only* HIR-compiled buffer programs that read/write the packed frame buffers directly, with all attribute offsets and strides resolved at compile time and baked in as literal constants. No Babel, no per-call token decoding, no `eval` of raw user text. Simulation worker bundles shrank from ~3 MB to ~40 kB.

And what it enables next: a **Petrinaut DSL** frontend lowering to the same HIR (TS ⇄ DSL migration becomes pretty-printing), **WASM/GPU backends** (the buffer ABI is already the right shape: scalars, static offsets, no GC), engine optimizations informed by analyses (e.g. constant-rate detection), and artifact caching.

> [!NOTE]
> **This HIR is a first version and should not be considered definitive.** The node set, the accepted TypeScript subset, the artifact format, and the buffer ABI are all expected to evolve (they already went through three artifact revisions inside this PR). It is deliberately private API: nothing outside `petrinaut-core`/`petrinaut` depends on its shape.

### How it fits together

```mermaid
flowchart LR
    subgraph authoring["Authoring"]
        TS["TypeScript user code<br/>dynamics · rates · kernels · metrics"]
        DSL["Petrinaut DSL<br/>(future)"]
    end

    subgraph hir["HIR pipeline (this PR — runs in the language worker)"]
        LOWER["Lower + typecheck<br/>against the token schema"]
        IR[("HIR<br/>typed · spanned · JSON")]
        ANALYZE["Analyses<br/>distribution DAG · dependencies · lints"]
        EMIT["Emit buffer programs<br/>offsets/strides baked as constants"]
    end

    subgraph runtime["Simulation workers"]
        ENGINE["Engine hot loops<br/>packed frame buffers (format v2)"]
        FUTURE["WASM / GPU backends<br/>(future)"]
    end

    LSPD["Editor diagnostics<br/>exact source ranges"]

    TS --> LOWER
    DSL -.-> LOWER
    LOWER --> IR
    IR --> ANALYZE --> LSPD
    IR --> EMIT -->|"artifacts (plain JS sources)"| ENGINE
    EMIT -.-> FUTURE

    style DSL stroke-dasharray: 5 5
    style FUTURE stroke-dasharray: 5 5
```

**Touched by the HIR**: the four code surfaces above, the LSP checker/worker (new lints + compile request), the engine hot loops (buffer-ABI execution), and the simulation/experiment start flows (artifact threading). **Not touched**: the frame format itself, the net editor, scenarios (still sandboxed `new Function`, run once at start) and visualizers (JSX in the UI) — those are candidates for later HIR/DSL surfaces.

## 🔗 Related links

- [FE-629: Add an intermediate representation for Petrinaut](https://linear.app/hash/issue/FE-629) _(internal)_
- [FE-514](https://linear.app/hash/issue/FE-514) _(internal)_ — the earlier expression-IR prototype (#8541) this supersedes
- [FE-928: Python language support](https://linear.app/hash/issue/FE-928) _(internal)_ — alternative frontends lower to the same HIR

## 🚫 Blocked by

- [x] Stacked on #8956 (FE-769 string type) — base is set to that branch; GitHub retargets to `main` when the stack merges.

## 🔍 What does this change?

- **HIR pipeline** (`petrinaut-core/src/hir/`, entries `./hir` + dependency-free `./hir-runtime`): lowering (incl. destructuring, guard `if`s, `.map`/`.reduce`/`.concat`, `Distribution.*`, `Uuid.*`, string predicates), schema typechecking, analyses, lints, and buffer-program emitters for all four surfaces. Design doc: `src/hir/README.md`; DSL sketch: `src/hir/dsl-sketch.md`.
- **Buffer ABI**: lambdas read attributes via `f64/u64/u8` views at literal offsets (`placeBases` + baked strides); kernels write per-transition staging with RNG-consuming slots deferred through declaration-order sinks (exact RNG-stream preservation); dynamics compile to allocation-free loops; metrics emit reduction loops over the token region (the per-frame `MetricState` full decode is gone). Strings resolve via the per-run intern pool; uuids as bigints from two u64 lanes.
- **Engine**: `CompiledTransition` carries only buffer programs + reusable scratch; Babel (`@babel/standalone`), `compile-user-code.ts`, `compile-metric.ts`, and all record decode/encode adapters are deleted.
- **Artifacts**: compiled in the LSP worker (`sdcpn/compileHirArtifacts` request), threaded through simulation/Monte-Carlo configs; items with no artifact fail per-item at build.
- **LSP/UX**: HIR lints with codes `99001+` and `source: "hir"`; out-of-subset or non-compilable code is a blocking **error**; Play now gates on error-severity diagnostics only (warnings/hints show an amber indicator).
- **Examples** are the coverage gate: every dynamics/lambda/kernel/metric of all five shipped models must compile to buffer programs (two pre-existing H-6519 violations in the supply-chain example were fixed along the way).
- **HIR Playground** Storybook story: type code + edit the schema, inspect the HIR tree, diagnostics, types, dependency sets, distribution DAG, and both emitted program forms live.

This PR:

The changes in this PR:

  - `petri-net-extensions.md` documents the supported code subset (now incl. metrics) and the error/warning split; `simulation.md` updated for error-only Play gating. Screenshots showing the diagnostics indicator may want refreshing (new amber warnings-only state).

The changes in this PR:

## ⚠️ Known issues

- **Behaviour change (deliberate)**: user TypeScript outside the analyzable subset no longer runs — it surfaces as an error diagnostic with a rewrite hint. The subset is what makes analysis and static-offset compilation possible; it covers all shipped examples and common idioms (`const` destructuring, guard clauses, `.map`/`.reduce`/`.concat`, `Math.*`, distributions, `Uuid.*`).
- Scenarios keep their sandboxed `new Function` path (cold, once per run); visualizers keep JSX/Babel in the UI package.
- A `.reduce` inside a conditional branch is evaluated eagerly (pure, so semantics hold — just possible wasted work).
- Static `storybook build` output fails to render for **all** stories in this package (pre-existing `react/jsx-runtime` issue); use `yarn dev`.

## 🐾 Next steps

- Benchmark vs the pre-HIR baseline (`8d623ca2ad`) — harness plan exists
- Skip `materializeEngineFrame` in the single-run engine (buffer path needs only offsets/counts + token bytes)
- Scenario surface on the HIR; then the DSL frontend (`src/hir/dsl-sketch.md`)
- WASM backend, GPU dynamics/Monte-Carlo
- Artifact caching keyed by content hash

## 🛡 What tests cover this?

- 707 petrinaut-core + 141 petrinaut tests, including:
  - determinism gate: identical inputs → identical frames over 50 stochastic steps (RNG stream through distribution sampling and uuid generation)
  - coverage gate: every example model item (all four surfaces) compiles to a buffer program
  - unit tests for lowering (spans, destructuring, guards, reduce/concat, metric bodies), schema typechecking, analyses, and all emitters against hand-packed format-v2 buffers (strings, uuids, booleans)
- The playground story was verified headless (Playwright): all panels populate, surface switching works, schema edits produce live diagnostics.

## ❓ How to test this?

1. `cd libs/@hashintel/petrinaut && yarn dev` → open **HIR → Playground**: type code across the four surfaces, break it (add a `for` loop), edit the schema (make an attribute `"integer"` and feed it a Distribution)
2. Open the editor, load an example (e.g. Supply Chain), check the Diagnostics tab, run a simulation and a Monte-Carlo experiment with expression metrics
3. `yarn workspace @hashintel/petrinaut-core test:unit run`


kube:
Stacked on #8764 (H-6519 discrete attribute types) — this PR's base is set to that branch so the diff shows only the HIR work; GitHub will retarget to `main` when #8764 merges.

kube:
Rebased onto the top of the stack (#8956, `cf/fe-769-add-string-discrete-type-support`) and retargeted the base accordingly. The HIR now supports the v2 packed token layout plus **uuid** and **string** attribute types end-to-end: lowering (string literals, `.startsWith`/`.endsWith`/`.includes`, `Uuid.generate()`/`Uuid.from(...)`), schema typechecking (optional uuid outputs, string-writable kernels), and buffer-ABI lambdas/dynamics reading packed structs directly (f64/u64/u8 views, pool-resolved strings, bigint uuids from the two u64 lanes). Kernel buffer emission is deferred pending an RNG-ordinal sink design (kernels run the object program, once per firing).

kube:
The simulator now runs **exclusively on HIR buffer programs** — the object-convention path (record decode → user-shaped JS → re-encode) is gone. Kernel buffer emission landed via declaration-order sink calls (distributions / UUID generation defer to the engine, preserving the seeded RNG stream; strings intern inline through the per-run pool). All dimension offsets **and strides** are resolved once at compile time and baked into the emitted programs as literal constants — lambdas receive `(placeBases, indices)` and compute token bases inline, replacing the generic per-slot offset loop. Non-scalarizable shapes are now compile errors (`hir:not-compilable`); every example model (kernels included) compiles fully, enforced by the coverage gate. 694 tests green.

kube:
Metrics are now the fourth HIR surface, and the `new Function` metric path is deleted. New `.reduce`/`.concat` nodes; metric bodies lower with spans mapped to the raw text; compiled metric programs read the packed frame directly — attribute offsets/strides are compile-time literals, place ordinals resolve once at instantiation, and reductions emit as loops over the token region (the per-frame `buildMetricState` decode of every place is gone). Single-run timeline and actual-mode metric paths use the same evaluator. LSP metric sessions gain HIR lint including `hir:not-compilable` as blocking errors. All 18 example metrics compile; 707 core + 141 UI tests green.

PR #8959, opened 2026-07-06
BE-635: Replace composite-type inserts to support zero-downtime additive migrations

TimDiekmann:

Schema changes must survive the rollout window (graph-migrate done, old binary still serving). Composite row types on the wire encode the table's column count, so **any** column change on a composite-inserted table broke every insert from the old binary — rolling migrations were impossible and every schema change required a maintenance window.

This PR replaces all composite-type inserts (live write paths and snapshot restore) with explicit column lists and one array parameter per column via `unnest`. An old binary simply doesn't mention a new nullable/defaulted column and keeps working, so additive migrations roll out with zero downtime.

## 🔗 Related links

- [BE-635](https://linear.app/hash/issue/BE-635) _(internal)_
- [BE-625](https://linear.app/hash/issue/BE-625) _(internal)_ — unblocked by this PR
- [BE-634](https://linear.app/hash/issue/BE-634) _(internal)_ — pre-existing restore bug found while validating

## 🔍 What does this change?

- `PostgresRow` produces column–parameter pairs; `InsertStatement` derives statement, casts, and parameters from that single sequence, so they cannot fall out of order
- `InsertStatementOptions` with table-name override, `DISTINCT`, and an extensible `OnConflict` enum; the `<table>_tmp` staging convention lives in the snapshot's `insert_rows_batch` helper
- all 32 snapshot batch inserts converted; every composite `ToSql` derive removed
- `PostgresType` becomes the single storage-type authority (built-in names sourced from `postgres_types::Type`); `DatabaseColumn` now carries the physical column facts, filter typing moves to the new `FilterColumn` trait
- dead code removed along the way (unused `nullable()`, eight unused row structs)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Column removals/renames still require a two-release drain — only additive changes become free

## 🐾 Next steps

- `information_schema` conformance test for the hand-written column/type mappings
- `DatabaseTable` trait linking table → columns → row, separating physical from computed columns
- snapshot fixtures for teams/AI actors (the only two batch arms without test data)

## 🛡 What tests cover this?

- transpile unit tests for the generated insert statements
- `hash-graph-postgres-store` integration suite (277 tests) against live Postgres
- `hash-backend-integration` snapshot-restore roundtrip (48 tests) and the graph HTTP suite
- manual `dump → wipe → restore → dump` equivalence check (canonical JSON comparison)

## ❓ How to test this?

1. `cargo nextest run --package hash-graph-postgres-store --all-features` (needs the compose Postgres + plain `hash-graph-migrations`)
2. `cd tests/hash-backend-integration && yarn vitest --run src/tests/subgraph` (needs the graph with `--embed-admin`)


libs/@local/graph/postgres-store/src/snapshot/action/table.rs
```diff
@@ -1,16 +1,117 @@
...
-#[derive(Debug, ToSql)]
-#[postgres(name = "action_hierarchy")]
+#[derive(Debug)]
 pub struct ActionHierarchyRow {
     pub parent_name: String,
     pub child_name: String,
     pub depth: i32,
 }
+
+use crate::store::postgres::query::{
```
indietyp: please move the use statement ot the top

libs/@local/graph/postgres-store/src/snapshot/action/table.rs
```diff
@@ -1,16 +1,117 @@
...
+
+impl PostgresRow for ActionRow {
+    type Column = Action;
+
+    fn table() -> TableName<'static> {
+        Table::Action.into()
+    }
+
+    fn columnar_parameters<'r>(
+        rows: &'r [Self],
```
indietyp: We started using descriptive names for lifetimes, so I reckon this should be `'rows` instead.

libs/@local/graph/postgres-store/src/snapshot/policy/table.rs
```diff
@@ -22,10 +20,198 @@ pub struct PolicyEditionRow {
...
-#[derive(Debug, ToSql)]
-#[postgres(name = "policy_action")]
+#[derive(Debug)]
 pub struct PolicyActionRow {
     pub policy_id: PolicyId,
     pub action_name: String,
     pub transaction_time: LeftClosedTemporalInterval<TransactionTime>,
 }
+
+use crate::store::postgres::query::{
```
indietyp: please move the use statement to the top

libs/@local/graph/postgres-store/src/store/postgres/query/rows.rs
```diff
@@ -26,68 +26,215 @@ use type_system::{
...
+pub trait PostgresRow: Sized {
+    /// The column type of the table this row is inserted into.
+    type Column: DatabaseColumn + 'static;
+
+    /// The table this row is inserted into.
+    fn table() -> TableName<'static>;
+
+    /// Transposes `rows` into one array parameter per column.
+    ///
+    /// Implementations destructure `Self` exhaustively so a new field is a compile error
```
indietyp: I would promote this section to a proper `# Implementation Note` instead of just a description. The conversion into `columnar_properties` may lead to data corruption if not done carefully, such as properties missing or misalignment (as the description itself states). Additionally, a small declarative macro may help reduce repetitive code throughout this change, which should also reduce the potential hazard of a wrong implementation in the future.

libs/@local/graph/postgres-store/src/snapshot/principal/table.rs
```diff
@@ -7,53 +7,377 @@ use type_system::principal::{
...
 
-#[derive(Debug, ToSql)]
-#[postgres(name = "actor_role")]
+#[derive(Debug)]
 pub struct ActorRoleRow {
     pub actor_id: ActorEntityUuid,
     pub role_id: Uuid,
 }
+
+use crate::store::postgres::query::{
```
indietyp: please move these at the top (I will stop to mention them all now, but there are a bunch of others)

libs/@local/graph/postgres-store/src/snapshot/principal/table.rs
indietyp, line 1: It seems odd to me that we're deferring snapshot tests for this, even though we explicitly add the implementation here. Feels like a potential footgun.

libs/@local/graph/postgres-store/src/snapshot/action/table.rs
TimDiekmann: Whops. In another project I had a lint for this and I got used to that 😄 
Fixed.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
-            Self::Query(query) => {
-                fmt.write_str("(\n    ")?;
-                query.transpile(fmt)?;
-                fmt.write_str("\n)")
-            }
-        }
-    }
+/// Options for compiling a bulk [`InsertStatement`].
+#[derive(Debug, Clone, Default)]
+pub struct InsertStatementOptions {
```
indietyp: I am not a huge fan of re-using the `InsertStatement` for an operation which is no longer a pure insert statement, but rather a `INSERT SELECT`, as 1) it makes intent unclear, and 2) some of the operations and options that are given are not insert-related, but select-subquery-related. If we don't need the old `InsertStatement`, then it may make sense to 1) convert this into multiple structs, such that it becomes clear which options operate on the select, and which on the outer insert and 2) rename it to something like `BulkInsertStatement`.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
-    pub fn add_expression(
-        &mut self,
-        column: impl Into<Column>,
-        expression: Expression,
-    ) -> &mut Self {
-        self.statement.columns.push(column.into());
-        // TODO: Use a builder which knows `values` at compile time
-        let InsertValueItem::Values(values) = &mut self.statement.values else {
-            unreachable!()
+    pub fn compile_rows_with<'r, R: PostgresRow>(
```
indietyp: Why do we even have `InsertStatement`, if there is no way to directly invoke it? Right now, we have two types that do the same thing, but differently. Either we should use bon for a builder-like pattern, or consolidate it.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
+        rows: &'r [R],
+        options: InsertStatementOptions,
+    ) -> (String, Vec<Box<dyn ToSql + Send + Sync + 'r>>) {
+        let (columns, parameters): (Vec<_>, Vec<_>) =
+            R::columnar_parameters(rows).into_iter().unzip();
```
indietyp: considering that the result from columnar properties is always a Vec turned Box, it may make sense to specialize the type to be more general, or have a lightweight wrapper type that delegates and implements ToSql that gives us the length as an `Option<usize>`, that way we can `debug_assert` that indeed the length of the parameters are uniform across the board, which would allow us to circumvent silent data corruption, at least during testing.

indietyp: nit: can't you use `.collect` here now? considering `FromIterator` is implemented for tuples?

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
+
+        (statement.transpile_to_string(), parameters)
     }
+}
 
-    #[must_use]
-    pub fn with_value(mut self, column: impl Into<Column>, value: &'p (impl ToSql + Sync)) -> Self {
-        self.add_value(column, value);
-        self
+impl Transpile for InsertStatement {
```
indietyp: There is a style regression here: the previous transpile operations all tried to generate somewhat human-readable SQL; this no longer does.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
+/// See `PostgresRow` for why parallel arrays are used instead of the table's composite
+/// row type.
+#[derive(Debug)]
 pub struct InsertStatement {
-    pub table: Table,
-    pub alias: Option<Alias>,
-    pub columns: Vec<Column>,
-    pub values: InsertValueItem,
+    pub table: TableName<'static>,
+    pub columns: Vec<String>,
```
indietyp: This seems extremely dangerous to me; we previously used `Column` here to make sure that no column is used which isn't declared as such. This just _screams_ to be abused with a potential SQL injection. Either escape the column name, or go back to the `Column` type.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
+///
+/// See `PostgresRow` for why parallel arrays are used instead of the table's composite
+/// row type.
+#[derive(Debug)]
 pub struct InsertStatement {
-    pub table: Table,
-    pub alias: Option<Alias>,
-    pub columns: Vec<Column>,
-    pub values: InsertValueItem,
+    pub table: TableName<'static>,
```
indietyp: Why the move from `Table` to `TableName` here?

libs/@local/graph/postgres-store/src/snapshot/mod.rs
```diff
@@ -56,9 +56,58 @@ use uuid::Uuid;
...
+pub(crate) struct SnapshotInsertOptions {
+    /// Deduplicate the rows with `SELECT DISTINCT`.
+    pub distinct: bool,
+    /// How to handle rows conflicting with existing ones.
+    pub on_conflict: OnConflict,
+}
+
+/// Bulk-inserts `rows` into the temporary staging table created by
+/// [`WriteBatch::begin`] and returns the number of inserted rows.
+pub(crate) async fn insert_rows_batch<R: PostgresRow + Sync>(
```
indietyp: Please document the invariants this function participates in, concretely that it inserts into `R_tmp`, then we commit from tmp -> real. This is only possible, AFAIR, because we create the temporary table via `LIKE <table> INCLUDING ALL`. Any change to this would break this specific function.

indietyp: We only ever return the value here to log it. In the case of `distinct: false`, `OnConflict::Error`, should we assert/error out if the row count going in and the row count that has been inserted mismatch? As we do that everywhere else.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
```diff
@@ -1029,13 +1047,22 @@ where
                 ))
                 .await
                 .change_context(InsertionError)?;
+            if inserted_rows != *expected_rows as u64 {
```
indietyp, line 1055: This would benefit from the table name being additionally attached, so that we know which table this happened in.

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
```diff
@@ -401,10 +428,18 @@ pub enum StaticJsonField {
     StaticJson(&'static str),
 }
 
-pub trait DatabaseColumn: Copy {
-    fn parameter_type(self) -> ParameterType;
-    fn nullable(self) -> bool;
-    fn as_str(self) -> &'static str;
+/// A physical column of a database table: its SQL name and storage type.
```
indietyp: That is incorrect; it is also implemented for, e.g. `DataTypeEmbeddings::Distance`, which is not a physical column.

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
```diff
@@ -401,10 +428,18 @@ pub enum StaticJsonField {
...
+pub trait DatabaseColumn {
+    fn as_str(&self) -> &str;
```
indietyp: Why did this degrade to `&str` instead of `&static str`? The idea behind using `&'static str` was that we could ensure that the value is well... static and has been properly escaped. We cannot do that in this case anymore.

thehabbos007 (dismissed):
Looks good to me. Using `unnest` should also have slightly better performance in some situations as far as I know. I suggest we also await Bilal's review

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -3,177 +3,159 @@ use core::{fmt, fmt::Formatter};
...
-    ) -> &mut Self {
-        self.statement.columns.push(column.into());
-        // TODO: Use a builder which knows `values` at compile time
-        let InsertValueItem::Values(values) = &mut self.statement.values else {
-            unreachable!()
+    pub fn compile_rows_with<'r, R: PostgresRow>(
+        rows: &'r [R],
+        options: InsertStatementOptions,
+    ) -> (String, Vec<Box<dyn ToSql + Send + Sync + 'r>>) {
+        let (columns, parameters): (Vec<_>, Vec<_>) =
```
indietyp: We may also want to add a debug assert to ensure that no column is named twice and that the number of columns matches the expected count.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
indietyp: Isn't there a potential footgun here with arrays as arguments and `UNNEST`? Nothing prevents the user from supplying a `PostgresType::Array`, which is then unnested.

libs/@local/graph/postgres-store/src/store/postgres/query/rows.rs
```diff
@@ -109,42 +301,60 @@ pub struct EntityEmbeddingRow {
...
-impl PostgresRow for EntityHasRightEntityRow {
-    fn table() -> Table {
-        Table::EntityHasRightEntity
+    fn columnar_parameters<'r>(
+        rows: &'r [Self],
+    ) -> Vec<(EntityEmbeddings, Box<dyn ToSql + Send + Sync + 'r>)> {
+        let mut web_ids = Vec::with_capacity(rows.len());
+        let mut entity_uuids = Vec::with_capacity(rows.len());
+        let mut draft_ids = Vec::with_capacity(rows.len());
+        let mut propertys = Vec::with_capacity(rows.len());
```
indietyp: `properties` 😉

indietyp (changes requested):
I'm uncomfortable with the size of the unverified, hand-authored surface here: the 33 `columnar_parameters` impls are ~950 lines of manual AoS-to-SoA transposition over ~300 hand-written column/type pairs, none of it const-, macro-, or otherwise generated, with virtually 0 coverage reported by codecov over these files. While I appreciate the fact that a future PR may create conformance tests over information schemas, such tests would not catch issues in e.g. `columnar_parameters` column attribution during transposition.

The benchmarks have also shown a major regression; is this just noise or something more substantial here?

libs/@local/graph/postgres-store/src/store/postgres/query/rows.rs
TimDiekmann: I should stop using `sed` 😂

libs/@local/graph/postgres-store/src/snapshot/action/table.rs
TimDiekmann: The lifetime documents itself, the parameter name is literally `rows`. And unless we enforce it somewhere, this will happen again. 
Fixed, although I see this as a mechanical nitpick.

libs/@local/graph/postgres-store/src/snapshot/policy/table.rs
TimDiekmann: Fixed

libs/@local/graph/postgres-store/src/snapshot/principal/table.rs
TimDiekmann: We have snapshot tests in the BE integration test suite. Not the ideal place but they exist.

libs/@local/graph/postgres-store/src/snapshot/principal/table.rs
TimDiekmann: Fixed

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/mod.rs
TimDiekmann: Fixed

libs/@local/graph/postgres-store/src/snapshot/mod.rs
TimDiekmann: No, we shouldn't. Postgres should handle these cases properly and error before we even check the returned number.

TimDiekmann: Well, this would be possible with any table. `LIKE <table> INCLUDING ALL` is just the way how we create the temporary table here. I extended the description (and made the function private).

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: I have not kept the old behavior of the single-insertion statement as it's currently unused and I don't like implementing dead code. I renamed it to `BulkInsertStatement` to make the intend clear.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: Because we need to support tables which are not in `Table` (e.g. in snapshots). `TableName` is either an identifier or a `Table`.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: Same reason as above, we cannot use `Column`. And the risk of injections are low, because this is not user-facing code. I changed it to `ColumnName` instead, which is either an identifier or `Column`.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: Assertion added.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: You nailed it: "somewhat" readable. It always was quite bad 😄 
I inserted a few newlines to make it at least a bit better.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: I added a small wrapper, `ColumnParameters` which is constructed from `Vec`.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
TimDiekmann: I iterated over this struct multiple times and did not really noticed, that it became "unused" except for the sake of a name. I did not have the capacity, yet, but I wanted to improve the overall interface of calling the compilers, this I created BE-641 and BE-642 for that.
To retain the DX, I used a simple `bon` wrapper.

libs/@local/graph/postgres-store/src/store/postgres/query/rows.rs
TimDiekmann: Done

libs/@local/graph/postgres-store/src/store/postgres/query/table.rs
TimDiekmann: Fixed

TimDiekmann (commented):
> I'm uncomfortable with the size of the unverified, hand-authored surface here: the 33 columnar_parameters impls are ~950 lines of manual AoS-to-SoA transposition over ~300 hand-written column/type pairs, none of it const-, macro-, or otherwise generated, with virtually 0 coverage reported by codecov over these files. While I appreciate the fact that a future PR may create conformance tests over information schemas, such tests would not catch issues in e.g. columnar_parameters column attribution during transposition.

> The benchmarks have also shown a major regression; is this just noise or something more substantial here?

We have hundreds of test using these endpoints to insert things into the database. Codecov is known to be very inaccurate, in particular when it comes to cross-language tests. Also, we very well know, that the benchmarks on GitHub are very flakey.

I appreciate that a macro would clean up a lot of the boilerplate, but it's _far_ more work to implement and test a proc-macro and I definitely don't have the capacity. I need to get this PR in in order to get performance improvements into prod, this is a hard blocker.

libs/@local/graph/postgres-store/src/store/postgres/query/statement/insert.rs
```diff
@@ -1,179 +1,175 @@
...
+    let ((columns, casts), parameters): ((Vec<_>, Vec<_>), Vec<_>) = R::columnar_parameters(rows)
+        .into_iter()
+        .map(|(column, parameters)| {
+            let name = column.name();
+            debug_assert_eq!(
+                parameters.len(),
+                rows.len(),
+                "column `{}` must contain one element per row",
+                name.as_str()
+            );
```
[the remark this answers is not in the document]
TimDiekmann: This is not a runtime error, the input is never provided by the user. `debug_assert` is enough here.

libs/@local/graph/postgres-store/src/snapshot/mod.rs
```diff
@@ -56,9 +56,59 @@ use uuid::Uuid;
...
+            "INSERT",
+            otel.kind = "client",
+            db.system = "postgresql",
+            peer.service = "Postgres",
+            db.query.text = statement,
+        ))
+        .await
+        .change_context(InsertionError)
+        .attach(format!("could not insert into `{table_name}`"))
+}
```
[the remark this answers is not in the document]
TimDiekmann: Fine for snapshots.

indietyp:
> We have hundreds of test using these endpoints to insert things into the database. Codecov is known to be very inaccurate, in particular when it comes to cross-language tests. Also, we very well know, that the benchmarks on GitHub are very flakey.

While that may be true, I'm uneasy relying on integration tests instead of proper localised tests. I appreciate that it's urgent. Still, it may be worth creating a ticket to track more exhaustive testing and/or a declarative macro for column creation (no need for a proc-macro here tbh; this should be similar in complexity to the `bitflags` macro). If we ever have an error in those outer-layer tests, it's going to be a _pain_ and non-obvious to track down the bug this far down if it's not tested independently.

indietyp (approved):

PR #8971, opened 2026-07-07
FE-1133: Add drawer to ds

alex-e-leon:

Adds a new Drawer component to the DS.
Updates Dialog to share many of the same internal components/styling + improves some of the nesting and mounting/unmounting animations.
Add a use-avoid-scroll-width-change hook so that expandable content inside drawers/dialogs doesn't trigger reflow when a scrollbar gets added/removed.
Also adds the duration tokens to panda so that some ark animations work properly (and fixes new panda css errors caused by adding the tokens)

This PR:

The changes in this PR:

### DIFF
--- a/libs/@hashintel/ds-components/src/beta/spinner/speed.story.tsx
+++ b/libs/@hashintel/ds-components/src/beta/spinner/speed.story.tsx
@@ -1,5 +1,5 @@
 import { Spinner } from "../spinner";
 
 export const App = () => {
-  return <Spinner animationDuration=".8s" />;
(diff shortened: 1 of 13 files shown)


lunelson (approved):
Looks good; but I don't see any difference between the two versions, i.e. with/without partitioned footer?

alex-e-leon:
The partitioned footer version has a grey background, otherwise white if a footer is provided - ideally would move this to a prop on the footer component but i then have to  do some nasty data shuffling so that the parent can get the right css

PR #8979, opened 2026-07-08
BE-638: Denormalize entity creator and creation timestamps into columns

TimDiekmann:

Pull the three write-once provenance values that authorization and sorting touch on every entity query out of the `provenance` JSONB into dedicated columns on `entity_ids` (plus the edition creator on `entity_editions`), so those reads no longer extract from the JSONB per row.

Done in a rolling-deploy-safe way (nullable + backfill + dual-write + read fallback), building on the zero-downtime insert work from BE-635.

## 🔗 Related links

- [BE-638](https://linear.app/hash/issue/BE-638) — this issue
- [BE-639](https://linear.app/hash/issue/BE-639) — follow-up: NOT NULL, index, drop the JSONB keys

## 🔍 What does this change?

- Adds nullable columns to `entity_ids`: `created_by_id`, `created_at_transaction_time`, `created_at_decision_time`; and `created_by_id` to `entity_editions` (one `V54` migration + `v009` inline, single backfill pass).
- Populates them on create and snapshot restore; the `CreatedByPrincipal` auth filter and the "Created" sort read `COALESCE(column, provenance ->> …)` so rows written by pre-migration binaries stay correct during the rollout window.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The columns are nullable and reads still fall back to the JSONB via `COALESCE`, so the query planner cannot use a column index yet — the auth filter and sort still avoid the per-row JSONB extraction, but the "Created" sort remains a full sort. Eliminating the sort (index scan) and dropping the fallback happen in BE-639 once the dual-write binary is drained on prod.

## 🐾 Next steps

- BE-639: backfill rollout-window rows, enforce `NOT NULL`, add the `created_at` index, drop the JSONB keys and the `COALESCE` fallback.
- **Deploy ordering:** this migration must run only after BE-635 binaries are fully drained on prod (older composite-insert binaries can't write the new columns).

## 🛡 What tests cover this?

- Transpile tests pinning the `COALESCE` fallback in the auth filter and the `created_at` sort SQL.
- DB roundtrip test asserting the columns are populated consistently with the provenance and that reads fall back correctly when the columns are NULL.

## ❓ How to test this?

1. Run the migration against a populated DB and confirm the three columns backfill from `provenance`.
2. Query entities filtered by creator / sorted by "Created"; confirm results match the pre-change ordering.


libs/@hashintel/ds-components/package.json
```diff
@@ -47,7 +47,7 @@
     "build": "npm-run-all build:lib build:buildinfo",
     "build:buildinfo": "panda ship --config panda.config.ts --outfile dist/panda.buildinfo.json",
     "build:ladle": "ladle build",
-    "build:lib": "tsup --config tsup.config.ts",
+    "build:lib": "NODE_OPTIONS=--max-old-space-size=8192 tsup --config tsup.config.ts",
```
TimDiekmann: Proper fix landed already
```suggestion
    "build:lib": "tsup --config tsup.config.ts",
```

thehabbos007 (approved):

PR #8991, opened 2026-07-09
BE-639: Make the denormalized provenance columns authoritative

TimDiekmann:

Finalizes the entity provenance denormalization started in #8979: the `created_by_id` and creation-timestamp columns become the single source of truth. The columns become `NOT NULL` and indexed, writes no longer duplicate the values into the JSONB, and creator/creation-time lookups no longer touch the JSONB — including in HashQL and the REST filter API. Removing the now-redundant JSONB keys from existing rows is deferred to [BE-687](https://linear.app/hash/issue/BE-687) so binaries from the previous release can keep reading during the deploy gap.

## 🔗 Related links

- [BE-639](https://linear.app/hash/issue/BE-639) _(internal)_
- [BE-687](https://linear.app/hash/issue/BE-687) _(internal)_ — follow-up that strips the JSONB keys
- #8979 (BE-638, first phase)

## 🚫 Blocked by

- [x] #8979 is merged **and deployed to prod** (the V55 backfill covers rows written during the deploy gap; running both migrations in one release would break inserts from pre-BE-638 binaries)

## 🔍 What does this change?

- Migration `V55` backfills any rows still missing column values, sets `NOT NULL`, and indexes the creation timestamps. The redundant JSONB keys are **not** removed here: binaries from the previous release deserialize them as required fields, so stripping before the binary rollover would fail every entity read in the deploy gap. [BE-687](https://linear.app/hash/issue/BE-687) removes them in a later release, once no such binaries remain. Fresh installs (`v009`) create the final state directly.
- The `COALESCE` fallback reads from #8979 are removed — filters, summaries, and sorting read the columns directly.
- Storage provenance representations are split from the interface types: `SqlEntityProvenance`/`SqlEntityEditionProvenance` mirror the column/JSONB layout, with exhaustive-destructure `From` impls so a new interface field is a compile error until it is routed to a storage side. All write paths (create, update, snapshot restore) and the read assembly route through them.
- `InferredEntityProvenance` is flattened into `EntityProvenance`. The wire format is unchanged (the field was already `#[serde(flatten)]`ed); the manual utoipa schema and the tsify patch move along, and the OpenAPI spec is regenerated with an identical generated TS client.
- HashQL exposes `metadata.provenance.created_by_id` / `created_at_transaction_time` / `created_at_decision_time` / `edition_created_by_id` (and `metadata.read_only`) as first-class, filterable paths that compile to column references, with timestamps as epoch-ms consistent with the temporal bounds.
- The REST filter grammar gains `createdById`, `editionCreatedById`, `createdAtTransactionTime`, and `createdAtDecisionTime` tokens; the two in-repo callers of the former JSONB spellings are migrated.

This PR:

The changes in this PR:

The changes in this PR:

## 🐾 Next steps

- Resolve the remaining `provenance.inferred` blob path in HashQL into individual optional paths (`firstNonDraftCreatedAt*`, `deletedBy*`)

## 🛡 What tests cover this?

- `created_by_columns.rs`: column round-trip through the read path, creator summaries, and a two-actor test pinning that entity creator, edition creator, and the two creation timestamps are independent columns
- HashQL: SQL snapshot tests proving column pushdown for creator and creation-time filters, an end-to-end filter run against a real database, and re-blessed whole-entity output snapshots
- `type-system`: serde round-trips for the flattened `EntityProvenance` and a schema-drift test asserting the manual OpenAPI schema matches the serialized shape
- Filter-token deserialization tests and doctests for the new query paths
- Full suites: postgres-store, graph integration, hashql (incl. compiletest), and the HTTP test suite against a live server

## ❓ How to test this?

1. Run the migrations against a database on the #8979 state — the backfill fills any column gaps while the JSONB keys stay in place
2. `cargo nextest run -p hash-graph-postgres-store --all-features`
3. From `tests/graph/http/`: `sh test.sh` against a running graph — `friendship.http` exercises the new `createdById` filter token


libs/@local/hashql/core/src/module/std_lib/graph/types/knowledge/entity.rs
```diff
@@ -149,31 +149,90 @@ pub mod types {
...
+        deps: Option<CreatedByIdDependencies>,
+    ) -> TypeId {
+        let CreatedByIdDependencies { uuid } = deps.unwrap_or_else(|| CreatedByIdDependencies {
+            uuid: std_lib::core::uuid::types::uuid(ty),
+        });
+
+        ty.opaque(sym::path::CreatedById, uuid)
+    }
+
+    // newtype CreatedAtTransactionTime = Timestamp;
```
indietyp, line 169: shouldn't this be `TransactionTime` aka

`newtype CreatedAtTransaction = TransactionTime<Timestamp>`

TimDiekmann: Yes, correct!
```suggestion
    // newtype CreatedAtTransactionTime = TransactionTime<Timestamp>;
```

libs/@local/hashql/core/src/module/std_lib/graph/types/knowledge/entity.rs
```diff
@@ -149,31 +149,90 @@ pub mod types {
...
+    #[must_use]
+    pub fn created_at_transaction_time(ty: &TypeBuilder<'_, '_>) -> TypeId {
+        ty.opaque(
+            sym::path::CreatedAtTransactionTime,
+            std_lib::graph::temporal::types::timestamp(ty),
+        )
+    }
+
+    // newtype CreatedAtDecisionTime = Timestamp;
```
TimDiekmann, line 178:
```suggestion
    // newtype CreatedAtDecisionTime = DecisionTime<Timestamp>;
```

libs/@local/graph/postgres-store/postgres_migrations/V55__entity_provenance_columns_authoritative.sql
```diff
@@ -0,0 +1,30 @@
...
+        created_at_decision_time = COALESCE(
+            created_at_decision_time, (provenance ->> 'createdAtDecisionTime')::timestamptz
+        )
+    WHERE created_by_id IS NULL
+       OR created_at_transaction_time IS NULL
+       OR created_at_decision_time IS NULL;
+
+UPDATE entity_editions
+    SET created_by_id = (provenance ->> 'createdById')::uuid
+    WHERE created_by_id IS NULL;
```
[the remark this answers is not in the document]
TimDiekmann: This is intentional. JSONB will be stripped in a follow-up.

libs/@local/graph/postgres-store/postgres_migrations/V55__entity_provenance_columns_authoritative.sql
```diff
@@ -0,0 +1,30 @@
...
+    ALTER COLUMN created_at_transaction_time SET NOT NULL,
+    ALTER COLUMN created_at_decision_time SET NOT NULL;
+
+ALTER TABLE entity_editions
+    ALTER COLUMN created_by_id SET NOT NULL;
+
+CREATE INDEX entity_ids_created_at_transaction_time
+    ON entity_ids (created_at_transaction_time);
+CREATE INDEX entity_ids_created_at_decision_time
+    ON entity_ids (created_at_decision_time);
```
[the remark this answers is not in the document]
TimDiekmann: This is intentional. JSONB will be stripped in a follow-up.

indietyp (approved):

PR #9031, opened 2026-07-15
H-6699: Add Petrinaut Optimizer Dockerfile

kube:

Package the Petrinaut optimizer from [#8989](https://github.com/hashintel/hash/pull/8989) as a lightweight service image.

## 🔗 Related links

- [Petrinaut Python optimizer API](https://github.com/hashintel/hash/pull/8989)

## 🚫 Blocked by

- [ ] [#8989](https://github.com/hashintel/hash/pull/8989)

## 🔍 What does this change?

- Adds a multi-stage Dockerfile that builds the Petrinaut CLI and focuses its production dependencies.
- Installs the locked Python dependencies with `uv`, without retaining `uv` in the runtime image.
- Runs Uvicorn as a non-root user on `0.0.0.0:4004`.
- Uses Python's standard library for the container healthcheck, so `curl`, `wget`, and a sidecar are not required.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `docker build -f apps/petrinaut-opt/docker/Dockerfile -t petrinaut-opt:split-check .`

## ❓ How to test this?

Build the image from the repository root with the command above.

## 📹 Demo

Not applicable; this PR only adds the service image definition.

### DIFF
--- a/apps/petrinaut-opt/docker/Dockerfile
+++ b/apps/petrinaut-opt/docker/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 # Build from the repository root:
-# docker build -f apps/petrinaut-opt/Dockerfile .
+# docker build -f apps/petrinaut-opt/docker/Dockerfile .
 
 ARG NODE_VERSION=22
 ARG PYTHON_VERSION=3.13
--- a/apps/petrinaut-opt/docker/Dockerfile
+++ b/apps/petrinaut-opt/docker/Dockerfile
@@ -91,9 +91,7 @@ ENV PATH="/opt/venv/bin:${PATH}" \
     PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1 \
     NODE_OPTIONS="--permission --allow-fs-read=/opt/petrinaut-cli --allow-fs-read=/usr/local/bin/petrinaut --disable-proto=throw --max-old-space-size=768 --no-addons --no-global-search-paths" \
-    PETRINAUT_CLI_NODE_OPTIONS="--permission --allow-fs-read=/opt/petrinaut-cli --allow-fs-read=/usr/local/bin/petrinaut --disable-proto=throw --max-old-space-size=768 --no-addons --no-global-search-paths" \
-    UVICORN_HOST=0.0.0.0 \
-    UVICORN_PORT=4004
+    PETRINAUT_CLI_NODE_OPTIONS="--permission --allow-fs-read=/opt/petrinaut-cli --allow-fs-read=/usr/local/bin/petrinaut --disable-proto=throw --max-old-space-size=768 --no-addons --no-global-search-paths"
 
 WORKDIR /app
 
@@ -118,6 +116,6 @@ USER petrinaut
 EXPOSE 4004
 
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=5 \
-    CMD ["python", "-c", "import os, urllib.request; urllib.request.urlopen(f'http://localhost:{os.environ.get(\"UVICORN_PORT\", \"4004\")}/status', timeout=4).close()"]
+    CMD ["python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:4004/status', timeout=4).close()"]
 
-ENTRYPOINT ["python", "-m", "uvicorn", "src.optimization_api:app"]
+CMD ["python", "-m", "uvicorn", "src.optimization_api:app", "--host", "0.0.0.0", "--port", "4004"]
--- a/apps/petrinaut-opt/docker/Dockerfile
+++ b/apps/petrinaut-opt/docker/Dockerfile
@@ -118,6 +118,6 @@ USER petrinaut
 EXPOSE 4004
 
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=5 \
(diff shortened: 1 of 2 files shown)


apps/petrinaut-opt/docker/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+      -type f \( -name "*.d.ts" -o -name "*.map" \) -delete
+
+
+FROM scratch AS cli-runtime
+
+COPY --from=cli-builder /repo/node_modules /node_modules
+COPY --from=cli-builder /repo/libs/@hashintel/petrinaut-core/package.json /libs/@hashintel/petrinaut-core/package.json
+COPY --from=cli-builder /repo/libs/@hashintel/petrinaut-core/dist /libs/@hashintel/petrinaut-core/dist
+COPY --from=cli-builder /repo/libs/@hashintel/petrinaut-cli/package.json /libs/@hashintel/petrinaut-cli/package.json
+COPY --from=cli-builder /repo/libs/@hashintel/petrinaut-cli/dist /libs/@hashintel/petrinaut-cli/dist
```
TimDiekmann, line 61: Nothing to worry about now, but it might be worth thinking about a bundler so we get this even smaller. (I'm sure there is a linear ticket for that somewhere, at least we talked about it quite a few times).

TimDiekmann (commented):
I was mainly looking at the Dockerfile, which looks good to me. I hope the remaining stuff is a rebase-thing 😅

apps/petrinaut-opt/docker/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+    ln -s /opt/petrinaut-cli/libs/@hashintel/petrinaut-cli/dist/cli.js /usr/local/bin/petrinaut && \
+    groupadd --system petrinaut && \
+    useradd --system --gid petrinaut --create-home petrinaut
+
+USER petrinaut
+
+EXPOSE 4004
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=5 \
+    CMD ["python", "-c", "import os, urllib.request; urllib.request.urlopen(f'http://127.0.0.1:{os.environ.get(\"UVICORN_PORT\", \"4004\")}/status', timeout=4).close()"]
```
YannisZa: Should this reference `os.environ.get(\"UVICORN_HOST\",\"localhost\")` too?

apps/petrinaut-opt/docker/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+
+CMD ["python", "-m", "uvicorn", "src.optimization_api:app"]
```
YannisZa: [OPTIONAL]: Do you want to append `["--host", [UVICORN_HOST], "--port", [UVICORN_PORT]]` just in case the local env vars are not properly read?

YannisZa (commented):
Should this dockerfile be moved to `apps/petrinaut-opt/docker/Dockerfile`?

TimDiekmann:
> Should this dockerfile be moved to apps/petrinaut-opt/docker/Dockerfile?

It would be consistent with the rest, yes. 
image

apps/petrinaut-opt/docker/Dockerfile
kube: It is actually bundled already.

`@hashintel/petrinaut-core` is an external dependency because of some issues with worker resolution when bundled.

We'll iterate on that. (Improvement to be made on the way workers are provided/intantiated)

apps/petrinaut-opt/docker/Dockerfile
kube: I'm using `localhost` here directly, as suggested by @TimDiekmann. Except if I'm mistaken.

apps/petrinaut-opt/docker/Dockerfile
kube: Seems to be default in Uvicorn?

kube: https://uvicorn.dev/settings/#configuration-methods

apps/petrinaut-opt/docker/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
```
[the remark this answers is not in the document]
kube: https://uvicorn.dev/settings/#configuration-methods

apps/petrinaut-opt/docker/Dockerfile
TimDiekmann: `UVICORN_HOST` is `0.0.0.0`, this is the address where it listens. but you want to connect to the address, so you either use `localhost` or `127.0.0.1` (`UVICORN_HOST` uses IPv4, so that's fine), Stick to `localhost` please.

kube: `localhost` is used in the latest version.

https://github.com/hashintel/hash/pull/9031/changes#diff-320e2402151f16d33192592149fb4016e5c2ec771108113f16162e08aab63bcaR121

apps/petrinaut-opt/docker/Dockerfile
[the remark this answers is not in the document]
TimDiekmann: I think `uvicorn.run()` doesn't pick it up, in this case it's working. But in case, someone changes the entry point to `uvicorn.run()` at some point it silently fails. To be sure, I'd actually would do
```Dockerfile
CMD ["python", "-m", "uvicorn", "src.optimization_api:app", "--host", "0.0.0.0", "--port", "4004"]
```

and removing the UVICORN env-vars.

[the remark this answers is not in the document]
TimDiekmann: Either way works, we should just agree on one route (and I don't mind either way)

PR #9033, opened 2026-07-15
BE-641: Extend the SQL AST along gram.y and make invalid SQL unrepresentable

TimDiekmann:

Extends the SQL statement AST in `hash-graph-postgres-store` so that every Postgres statement we need is expressible in typed Rust, modelled strictly along Postgres' grammar (`gram.y`): at most the grammar per node — invalid SQL is unrepresentable — with completeness added on demand. On top of the structural work, the `SelectCompiler` no longer uses the AST as mutable working state: it keeps its own bookkeeping and assembles the statement in `compile()`, which is the groundwork for the BE-598 query-shape changes.

Stacked on #9020 (module restructure); only the commits of this branch are new.

## 🔗 Related links

- [BE-641](https://linear.app/hash/issue/BE-641) _(internal)_ — the ticket driving this work
- [BE-699](https://linear.app/hash/issue/BE-699) _(internal)_ — follow-up: verify/generate the hand-maintained schema mirror
- #9020 — base PR (module restructure)

## 🚫 Blocked by

- [ ] #9020

## 🔍 What does this change?

**Grammar-faithful AST (per node: at most gram.y):**

- `WITH`: CTEs with plain [`TableName`]s, `MATERIALIZED`/`NOT MATERIALIZED` hints, optional column lists, and `WITH RECURSIVE`; the clause holds a `NonEmptyVec<CommonTableExpression>`
- The SELECT three-way split along gram.y: `SimpleSelect` (`simple_select`), `SelectClause` (set-operation tree: `UNION`/`INTERSECT`/`EXCEPT [ALL]`, operands parenthesized so Postgres re-parses the exact tree), `SelectStatement` (`select_no_parens`: `WITH`/`ORDER BY`/`LIMIT`/`OFFSET`) — clause placement rules are now type-structural (no `ORDER BY` on a bare set-operation operand)
- `GROUP BY [ALL|DISTINCT] grouping_element [, …]` with `GroupingElement::{Empty, Expressions}` (ROLLUP/CUBE/GROUPING SETS fenced off), `HAVING`, hardened `WindowDefinition` (`PARTITION BY` only, non-empty), `TABLESAMPLE` with finite `SamplePercentage`
- `WHERE` is `Option<Expression>` (the grammar's `[ WHERE condition ]`); `ORDER BY` uses AST-own `SortBy`/`SortDirection`/`NullsOrder` (gram.y `sortby`), decoupling the AST from store types
- `NonEmptyVec<T>` as the list-invariant helper (grow-only, no shrink API, `Deref<[T]>`); target lists, CTE lists, `DISTINCT ON`, `PARTITION BY`, and sort lists are non-empty by construction
- Fixed a latent clause-order bug: `ORDER BY` was emitted **before** `GROUP BY` (never hit because no statement set both)

**Builder composition:**

- bon builders throughout with `#[builder(derive(Into))]` conversion chains: builders flow into `impl Into` slots, `.build()` only remains at the outermost edge; `NonEmptyVec` members use plain `#[builder(into)]` setters (single element and ready-made vec convert infallibly, a raw `Vec` parses visibly via `NonEmptyVec::try_from`)

**Compiler decomposition:**

- `SelectCompiler` keeps `selects`/`from`/`quantifier`/`with`/`conditions`/`cursor`/`sort_by`/`limit` as its own fields and assembles the statement in `compile()`; join bookkeeping lives in mirrored `CompiledJoin` records instead of reading the statement tree back, and `Table::aliased_name` is the single owner of alias-name mangling
- The keyset cursor is built as an `Expression` tree; `nulls: None` now mirrors Postgres' actual default (ASC → `NULLS LAST`, DESC → `NULLS FIRST`) — previously unhinted ascending sorts over nullable keys silently lost their `NULL` rows during pagination. Provably `NOT NULL` sort keys (verified whitelist `Column::NON_NULL_COLUMNS`) skip the null-handling arms, keeping the hot default-sort predicates bare
- Silent failure modes made loud: empty `Identifier`s panic at construction, parameters with pending conversions return `PendingParameterConversion` instead of `unimplemented!`, and a second embeddings distance filter is rejected via an explicit flag (previously it silently reinstalled the subquery with mismatched parameters)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The emitted SQL changed shape: top-level `WHERE` conjuncts are now individually parenthesized and the cursor continuation is an expression tree. Semantically identical (validated by executing the hashql E2E suite and the integration tests against real Postgres), but `pg_stat_statements` will mint new query IDs — dashboards comparing against old fingerprints reset.
- The cursor port fixes two latent invalid-SQL edge cases of the old string-based `WHERE` builder (empty conditions combined with a cursor produced `WHERE  AND (…)` / a dangling `WHERE`).
- Documented grammar subsets remain (fenced in the doc comments): `LIMIT`/`OFFSET` as expressions (`LIMIT $1`), `FETCH`, locking clauses, `VALUES`/`TABLE`/parenthesized set-operation operands, the statement-level `WINDOW` clause, ROLLUP/CUBE/GROUPING SETS, `ORDER BY … USING op`, `SEARCH`/`CYCLE`.
- `FromItem` still allows a `USING` join with an empty column list (degrades to `ON TRUE`, documented); the structural fix is bundled with the BE-528 `column_definitions` work, which touches the same grammar production.

## 🐾 Next steps

- Subquery predicates (`EXISTS`, `IN (SELECT …)`) as the last planned AST block
- BE-528 bundle: `column_definitions` for typed JSON + `AliasClause` pairing / `USING` non-emptiness
- BE-699: complete `Column::nullable()` or generate the schema mirror from migrations
- BE-598: the fetch-keys-then-hydrate query shape on top of the decomposed compiler

## 🛡 What tests cover this?

- 164 unit tests in `hash-graph-postgres-store` (golden SQL for every compiler shape, all `cursor_condition` branches — previously untested —, per-clause transpile tests, `NonEmptyVec` invariants)
- `tests/schema.rs`: verifies `Column::NON_NULL_COLUMNS` against `information_schema` of the live database
- 98 hashql E2E tests (execute the emitted SQL against Postgres via testcontainers; 4 snapshots regenerated, diffs are pure added parentheses)
- 24 sorting/pagination integration tests (nullable keys, `NULLS FIRST`/`LAST` mixes, cursor loops against real data)

## ❓ How to test this?

1. `cargo nextest run --package hash-graph-postgres-store --lib`
2. `cargo nextest run --package hashql-eval` (needs Docker for testcontainers)
3. With a migrated local database: `cargo nextest run --package hash-graph-postgres-store --test schema` and `cargo nextest run -p hash-graph-integration --test postgres -E 'test(sorting)'`

## 📹 Demo

```rust
WithClause::builder()
    .recursive(true)
    .common_table_expressions(
        CommonTableExpression::builder().name("traversal").statement(
            SelectClause::from(base_case).union_all(recursive_case),
        ),
    )
    .build()
```

```sql
WITH RECURSIVE "traversal" AS (SELECT * FROM "data_types"
UNION ALL
SELECT * FROM "property_types")
```


libs/@local/graph/postgres-store/src/store/postgres/query/ast/non_empty.rs
```diff
@@ -0,0 +1,144 @@
...
+/// [`bon`] builder members of this type use the plain generated setter with `#[builder(into)]`:
+/// a single element and a ready-made `NonEmptyVec` convert infallibly, so setters never return
+/// [`Result`]. A [`Vec`] is parsed visibly at the call site via `NonEmptyVec::try_from` — the
+/// one place where emptiness can fail.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct NonEmptyVec<T>(Vec<T>);
+
+#[derive(Debug, PartialEq, Eq, derive_more::Display)]
+#[display("the vector must not be empty")]
+pub struct EmptyVec;
```
thehabbos007, line 24: Wouldn't we call this a `EmptyVecError`?

thehabbos007 (commented):
Looks good to me, only one comment

libs/@local/graph/postgres-store/src/store/postgres/query/ast/non_empty.rs
TimDiekmann: Yes, I think we should rename it. Would you mind if I add this in the follow-up (BE-705 which is already stacked on this)?

thehabbos007: Sure, do that

thehabbos007 (approved):

PR #9048, opened 2026-07-17
BE-702: Compile sorted entity reads as keys-first statements

TimDiekmann:

Sorted and limited entity reads now compile as a keys-first statement: a narrow key query carries the filters, cursor, and sort, and the wide projection is hydrated only for the rows that survive the `LIMIT`. This keeps the planner's ability to serve the ordering from an index while it sorts and spools far narrower rows, and it is the structural groundwork for optimizing the entity read path against known query shapes.

Grounded in the created-at sort incident: the pathological plan was reproduced locally against a 700k-entity seed and the shape trade-offs were measured with result-set-equivalence checks on every variant (see the Slack thread for the discussion).

## 🔗 Related links

- [BE-702](https://linear.app/hash/issue/BE-702) _(internal)_ — the driving ticket
- [BE-598](https://linear.app/hash/issue/BE-598) _(internal)_ — the investigation this implements
- [Slack: entities-page performance & sort keys](https://hashintel.slack.com/archives/C022217GAHF/p1784220011668439) _(internal)_ — incident context, measurements, and the follow-up direction
- [Slack: entities table loading behaviour](https://hashintel.slack.com/archives/C022217GAHF/p1783687778077499) _(internal)_

## 🚫 Blocked by

- [x] #9033

## 🔍 What does this change?

- `StatementShape` on the `SelectCompiler`: `SinglePass` (default, unchanged) | `KeysFirst` | `FencedKeysFirst`. The keys-first shapes compile into a key CTE (filters + cursor + sort columns), a page CTE (`ORDER BY` + `LIMIT`), and a hydration `SELECT` with all key-table references rewritten; the fenced variant adds `AS MATERIALIZED` to shield the key query from the outer ordering demand.
- Join classification by use (filters/cursor/sort keys, closed transitively over the join conditions) splits the collected joins into the key query's and the hydration's share; the `FROM` tree is now assembled in `compile()` from that bookkeeping.
- Statements that cannot be split degrade to `SinglePass` with a logged `ShapeFallback` reason; the effective shape is recorded on the `compile` span as `statement.shape`. Both the shape's materialization and its span value live on `StatementShape`, so a shape cannot be recorded under a fence it was not compiled with.
- The guards that reject a split are total rather than reasoned per side: any right or full outer join anywhere disqualifies the statement, and colliding key-column output names are rejected instead of surfacing as an ambiguous column reference at execution time.
- Pre-order `Expression::visit`/`visit_mut` walkers on the AST returning `ControlFlow`, so the classification stops at the first subquery it cannot see through instead of flagging one and checking after the walk. Used by the classification and the reference rewriting.
- `read_entities_impl` opts into `KeysFirst` for sorted entity reads; embedding-distance queries stay on `SinglePass` (`TODO(BE-618)`).
- A note on `AsClient for Object` documents that bypassing the pool's statement cache keeps every query on a custom plan — the shape strategy relies on parameter-aware planning.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- `FencedKeysFirst` has no production caller yet; it is exercised by the compile tests and reserved as an escalation shape.
- Statement texts for sorted entity reads change shape, so `pg_stat_statements` fingerprints reset for those queries.

## 🐾 Next steps

- Distinguish include- from exclude-type filters on the entity read path: inclusion lists are GIN-indexable and estimable, and the entities page can derive them from the type summary it already fetches (see the Slack thread).
- [BE-709](https://linear.app/hash/issue/BE-709) _(internal)_ — `Visitor`/`VisitorMut` traits over the whole SQL AST, which is what the expression walkers here should eventually become. Statement-level traversal is what unlocks parameter offsetting for statement composition, alias rewriting as a single pass, and seeing into subqueries well enough to drop `ShapeFallback::OpaqueExpression`; expression-level traversal on its own reaches none of them.

## 🛡 What tests cover this?

- Golden compile tests for both keys-first shapes: sort keys across joins, cursor continuation inside the key CTE, embedding-distance queries, carried CTEs, and filter-joined tables staying out of the key projection
- Fallback tests asserting the specific `ShapeFallback` reason plus byte-equality with the single-pass output, covering a generating join on either side of the split and a key-column output-name collision. Neither is reachable through a query path today, so both plant the offending join type or column name directly — the guards are fences against the surrounding structure changing.
- The full `hash-graph-integration` postgres suite (including all sorting/pagination tests) against a migrated database with `KeysFirst` active
- Result-set-equivalence checks (row-set fingerprints) of every measured variant against a 700k-entity seed

## ❓ How to test this?

1. `cargo nextest run --package hash-graph-postgres-store --lib`
2. With a migrated local database: `cargo nextest run --package hash-graph-integration --test postgres`

## 📹 Demo

```sql
WITH "roots" AS (SELECT "entity_ids_0_1_0"."created_at_transaction_time", "entity_temporal_metadata_0_0_0"."entity_uuid", "entity_temporal_metadata_0_0_0"."entity_edition_id"
FROM "entity_temporal_metadata" AS "entity_temporal_metadata_0_0_0"
INNER JOIN "entity_ids" AS "entity_ids_0_1_0" ON ...
WHERE ...),
"limited" AS (SELECT * FROM "roots" ORDER BY "created_at_transaction_time" ASC LIMIT 500)
SELECT DISTINCT ON("limited"."created_at_transaction_time", ...) ..., "entity_editions_1_1_0"."properties"
FROM "limited"
INNER JOIN "entity_editions" AS "entity_editions_1_1_0" ON "entity_editions_1_1_0"."entity_edition_id" = "limited"."entity_edition_id"
ORDER BY "limited"."created_at_transaction_time" ASC LIMIT 500
```

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
@@ -48,6 +48,96 @@ pub enum Function {
     Now,
 }
 
+/// Direct-child traversal for [`Expression::visit`] and [`Expression::visit_mut`].
+///
+/// [`Expression::visit`]: super::Expression::visit
+/// [`Expression::visit_mut`]: super::Expression::visit_mut
+impl Function {
+    pub(super) fn for_each_child(&self, visitor: &mut dyn FnMut(&Expression)) {
+        match self {
+            Self::Min(expr)
+            | Self::Max(expr)
+            | Self::JsonAgg(expr)
+            | Self::JsonExtractText(expr)
+            | Self::JsonExtractAsText(expr, _)
+            | Self::JsonScalar(expr)
+            | Self::ToJson(expr)
+            | Self::Lower(expr)
+            | Self::Upper(expr)
+            | Self::LowerInc(expr)
+            | Self::UpperInc(expr)
+            | Self::LowerInf(expr)
+            | Self::UpperInf(expr)
+            | Self::ExtractEpochMs(expr) => visitor(expr),
+            Self::JsonContains(lhs, rhs)
+            | Self::JsonPathQueryFirst(lhs, rhs)
+            | Self::Coalesce(lhs, rhs) => {
+                visitor(lhs);
+                visitor(rhs);
+            }
+            Self::JsonExtractPath(exprs)
+            | Self::JsonBuildArray(exprs)
+            | Self::Unnest(exprs)
+            | Self::ArrayLiteral {
+                elements: exprs, ..
+            } => {
+                for expr in exprs {
+                    visitor(expr);
+                }
+            }
+            Self::JsonBuildObject(pairs) => {
+                for (key, value) in pairs {
+                    visitor(key);
+                    visitor(value);
+                }
+            }
+            Self::Now => {}
+        }
+    }
+
+    pub(super) fn for_each_child_mut(&mut self, visitor: &mut dyn FnMut(&mut Expression)) {
+        match self {
+            Self::Min(expr)
+            | Self::Max(expr)
+            | Self::JsonAgg(expr)
+            | Self::JsonExtractText(expr)
+            | Self::JsonExtractAsText(expr, _)
+            | Self::JsonScalar(expr)
+            | Self::ToJson(expr)
+            | Self::Lower(expr)
+            | Self::Upper(expr)
+            | Self::LowerInc(expr)
+            | Self::UpperInc(expr)
+            | Self::LowerInf(expr)
+            | Self::UpperInf(expr)
+            | Self::ExtractEpochMs(expr) => visitor(expr),
+            Self::JsonContains(lhs, rhs)
+            | Self::JsonPathQueryFirst(lhs, rhs)
+            | Self::Coalesce(lhs, rhs) => {
+                visitor(lhs);
+                visitor(rhs);
+            }
+            Self::JsonExtractPath(exprs)
+            | Self::JsonBuildArray(exprs)
+            | Self::Unnest(exprs)
+            | Self::ArrayLiteral {
+                elements: exprs, ..
+            } => {
+                for expr in exprs {
+                    visitor(expr);
+                }
+            }
+            Self::JsonBuildObject(pairs) => {
+                for (key, value) in pairs {
+                    visitor(key);
+                    visitor(value);
+                }
+            }
+            Self::Now => {}
+        }
+    }
+}
+
 impl Transpile for Function {
     #[expect(
         clippy::too_many_lines,
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
@@ -1,4 +1,7 @@
-use core::fmt::{self, Write as _};
+use core::{
+    fmt::{self, Write as _},
+    ops::ControlFlow,
+};
 
 use hash_graph_store::filter::PathToken;
 
@@ -53,7 +56,10 @@ pub enum Function {
 /// [`Expression::visit`]: super::Expression::visit
 /// [`Expression::visit_mut`]: super::Expression::visit_mut
 impl Function {
-    pub(super) fn for_each_child(&self, visitor: &mut dyn FnMut(&Expression)) {
+    pub(super) fn for_each_child<B>(
+        &self,
+        visitor: &mut impl FnMut(&Expression) -> ControlFlow<B>,
+    ) -> ControlFlow<B> {
         match self {
             Self::Min(expr)
             | Self::Max(expr)
@@ -72,8 +78,8 @@ impl Function {
             Self::JsonContains(lhs, rhs)
             | Self::JsonPathQueryFirst(lhs, rhs)
             | Self::Coalesce(lhs, rhs) => {
-                visitor(lhs);
-                visitor(rhs);
+                visitor(lhs)?;
+                visitor(rhs)
             }
             Self::JsonExtractPath(exprs)
             | Self::JsonBuildArray(exprs)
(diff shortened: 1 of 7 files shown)


libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
```diff
@@ -48,6 +48,96 @@ pub enum Function {
     Now,
 }
 
+/// Direct-child traversal for [`Expression::visit`] and [`Expression::visit_mut`].
+///
+/// [`Expression::visit`]: super::Expression::visit
+/// [`Expression::visit_mut`]: super::Expression::visit_mut
+impl Function {
+    pub(super) fn for_each_child(&self, visitor: &mut dyn FnMut(&Expression)) {
```
indietyp: Two questions:
1. Why `dyn FnMut()`? Why not just `&mut impl FnMut()`
2. Wouldn't a trait-based approach work better here? Inside of HashQL, I use the visitor pattern extensively, using e.g. libs/@local/hashql/mir/src/visit/mut.rs; the big benefit: The caller decides when to stop, how to stop, and how to transform, if pre- or post-step. Generally more applicable.

I would really like us to formalise this properly, we've seen that the same shape here has been done inside of `Filter` many times handrolled, so I am advocating to us just doing it properly the first time.

libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/mod.rs
```diff
@@ -407,6 +407,127 @@ impl Expression {
...
+            Self::StartsWith(lhs, rhs)
+            | Self::EndsWith(lhs, rhs)
+            | Self::ContainsSegment(lhs, rhs) => {
+                visitor(lhs);
+                visitor(rhs);
+            }
+        }
+    }
+
+    fn for_each_child_mut(&mut self, visitor: &mut dyn FnMut(&mut Self)) {
```
indietyp: Why is this `&mut FnMut` anyway? if https://doc.rust-lang.org/stable/std/ops/trait.FnMut.html#impl-FnMut%3CA%3E-for-%26mut+F exists?

libs/@local/graph/postgres-store/src/store/postgres/query/compile/mod.rs
```diff
@@ -488,33 +700,406 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
...
+                .keys_first_or_fallback("fenced_keys_first", Some(Materialization::Materialized)),
+        };
+        (statement.transpile_to_string(), &self.artifacts.parameters)
+    }
+
+    /// Compiles the keys-first split, degrading to the single-pass layout with a logged
+    /// reason when the statement does not qualify.
+    fn keys_first_or_fallback(
+        &self,
+        shape: &'static str,
```
indietyp: Why is this a static string and not just `StatementShape`? Especially considering that it's just `self.shape`. Could remove the argument, and just record via `Display` instead.

libs/@local/graph/postgres-store/src/store/postgres/query/compile/mod.rs
```diff
@@ -488,33 +700,406 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
...
+        let mut in_key_query = vec![false; self.artifacts.joins.len()];
+        for (slot, join) in in_key_query.iter_mut().zip(&self.artifacts.joins).rev() {
+            if required.contains(&join.table.aliased_name(join.alias)) {
+                *slot = true;
+                for condition in &join.conditions {
+                    collect_referenced_tables(condition, &mut required, &mut opaque);
+                }
+            }
+        }
+        if opaque {
```
indietyp: In combination with the visitor pattern outlined before, you could move this opaque to the top, stop traversing on opaque, and then just early return. No?

libs/@local/graph/postgres-store/src/store/postgres/query/compile/mod.rs
```diff
@@ -488,33 +700,406 @@ impl<'p, 'q: 'p, R: PostgresRecord> SelectCompiler<'p, 'q, R> {
...
+        let mut key_joins = Vec::new();
+        let mut hydrated_joins = Vec::new();
+        for (join, in_key_query) in self.artifacts.joins.iter().zip(&in_key_query) {
+            if *in_key_query {
+                key_joins.push(join);
+            } else {
+                hydrated_joins.push(join);
+            }
+        }
+        if key_joins.iter().any(|join| join.to_many) {
```
indietyp, line 931: The guard only checks `hydrated_joins`, but not `key_join`. While not hittable today, this is a latent bug waiting, as it depends on a fact that is established and checked nowhere. That key query joins are never `RightOuter | FullOuter` may be true now, but there is no guard, no contract here, that establishes that this will never be true for key joins.

libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
TimDiekmann: The vtables are basically free here (very small), but the generated code is not due to monomorphization. With a trait, you would effectively call it with `F`, then `&mut F`, then `&mut &mut F` etc, which implies even more generated code bloat. I'll explore the visitor pattern nevertheless.

libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/mod.rs
TimDiekmann: See above

libs/@local/graph/postgres-store/src/store/postgres/query/ast/expression/function.rs
indietyp: Isn't that premature optimisation, though? I would say that, arguably, using `impl FnMut` here is usually the more idiomatic approach. Hence my suggestion.

TimDiekmann: It might be, yes. For now I changed it to `impl` as it was only a four-line change. I looked at the `Visitor` pattern here to properly move this into a trait, but this would explode the scope of this PR. I created a dedicated ticket to properly implement a `Visitor` trait (this code does not need to look at `Expression`, which makes it easier). See BE-709

indietyp (approved):

PR #9070, opened 2026-07-20
feat: allow map_err like syntax on change_context

khanhtranngoccva:

- Add `Report::current_context_mut` to access the internal error mutably. This is required if the caller needs to modify or extract non-Clone fields as part of error recovery.
- Add `ResultExt::change_context_adaptive` to allow `map_err` like syntax. This helps with cases of error propagation where the underlying error needs to be surfaced to the domain level in the current form or in another form.

## 🔍 What does this change?

- Added `Report::current_context_mut`
- Added `ResultExt::change_context_adaptive`

This PR:

The changes in this PR:

  - _Provide more detail here_

The changes in this PR:

## ⚠️ Known issues
- Compiler jank related to non-lexical lifetimes prevents embedding the report in case of a panic.

## 🐾 Next steps
- Re-introduce the report embedding in the `unreachable!` panic in `Report::current_context_mut`

## 🛡 What tests cover this?

- Test cases in `test_change_context.rs` and `test_conversion.rs` have been modified to cover these two functions.

## ❓ How to test this?
- Running cargo test is sufficient, as these tests are fuzzed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2655,7 +2655,7 @@ dependencies = [
 
 [[package]]
 name = "error-stack"
-version = "0.8.0"
+version = "0.8.1"
 dependencies = [
  "ansi-to-html",
  "anyhow",
@@ -2668,6 +2668,7 @@ dependencies = [
  "insta",
  "owo-colors",
  "pin-project-lite",
+ "rand 0.10.1",
  "regex",
  "rustc_version",
  "serde",
--- a/libs/error-stack/Cargo.toml
+++ b/libs/error-stack/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name          = "error-stack"
-version       = "0.8.0"
+version       = "0.8.1"
 authors       = { workspace = true }
 edition       = "2021"
 rust-version  = "1.83.0"
@@ -42,6 +42,7 @@ futures-util       = { workspace = true }
 glob               = { workspace = true }
 insta              = { workspace = true, features = ["filters", "ron"] }
 owo-colors         = { workspace = true }
+rand               = { workspace = true, features = ["thread_rng"] }
 regex              = { workspace = true }
 serde              = { workspace = true, features = ["derive"] }
 supports-color     = { workspace = true }
--- a/libs/error-stack/src/report.rs
+++ b/libs/error-stack/src/report.rs
@@ -465,8 +465,8 @@ impl<C> Report<C> {
     where
         C: Send + Sync + 'static,
     {
-        if let Some(r) = self.downcast_mut::<C>() {
-            return r;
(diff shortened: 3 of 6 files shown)


indietyp:
Thank you for the PR!

After careful consideration, we have decided not to merge it.

`change_context_adaptive` goes against the API surface and intuitions developed across the Rust landscape. As you pointed out, this is a map-like operator, and map-like operators by convention take `self`; making `change_context_adaptive` conform would turn it into a `map_err` in sheep's clothing.

We found that `current_context_mut` is the wrong tool because users can already do this through the public API; more importantly, it sets the wrong expectation. We do not want to encourage context modification, and adding the helper would signal that this is intended behaviour, when it isn't. We see context as largely immutable: you seldom need to remove context, only add to it, and adding information is better done through attachments. This gives us a provenance tree; any modification outside this chain is invisible to it. Removal is worse still: by design it swaps out or discards information, meaning we change the material reality of the underlying application context, something we might not be privy to. It would leave a report that is valid at runtime yet informationally incomplete, which makes misuse easy to achieve. This is why the underlying utilities are public, yet we don't expose helper functions like these: possible by design, but deliberately not frictionless. We're happy to add an example outlining how to do this with the existing API, or to review one if you'd like to contribute it.

khanhtranngoccva:
@indietyp 

- I intentionally made `change_context_adaptive` to act like `map_err`, but this variant of `map_err` should automatically return an `error-stack` augmented `Result` type, while using `map_err` as-is unintentionally discards the entire report stack, leaving only the topmost error (and the user may end up asking where the rest of the error tree is). 
  - This corresponds with the DDD patterns where users may want to reuse the information of existing errors to map to parent layers (such that errors reflect the documented use cases). For example, an `InvalidInput` or `NotFound` error in the lower layer can be intentionally reduced to `NotFound` on the upper layer to be returned to an example public API. I think this API may be worth reconsidering.
  - Even then, `change_context_adaptive` does not exactly look like a `map_err`, because `change_context_adaptive` must keep ownership of the report to chain it. The closure instead receives a mutable reference to the report and may call `current_context` from there.

- `current_context_mut` aims to make non-trivial error recovery a part of documentation. For example, I am currently writing an API for a protected file wrapper, and if the wrapper fails to load, the original file descriptor is recovered and the callers may view it in ciphertext form. In this example, I would like to make the `File` a part of the statically typed error struct, documenting that the users may want to extract the file out. It appears that I figured out a middle ground though - the non-trivial `File` can be wrapped by a `Mutex`, where an API function can swap out the `File` by calling a method `Self::take_file(&self)` and trivially acquiring the lock on the fast path. In this kind of approach, the users must explicitly implement non-trivial recovery methods in the corresponding error structs.

PR #9080, opened 2026-07-21
FE-1238: Send the type universe as an include list on the entities page

TimDiekmann:

The default view of the `/entities` page filters types only by exclusion (`ignoreNoisySystemTypesFilter`), which compiles to NOT-array-overlap chains Postgres cannot estimate — the root of the multi-second sorted-read incidents on the entities page. The page already knows the visible-type universe from the `summarizeEntities` call that feeds its filter pills, so it now sends that universe as an include-type clause the planner can index and estimate. Measured against a 700k-entity seed, the incident-shaped query drops from 5.5s to sub-millisecond with the include clause in place.

## 🔗 Related links

- [FE-1238](https://linear.app/hash/issue/FE-1238) _(internal)_ — the driving ticket
- [BE-702](https://linear.app/hash/issue/BE-702) _(internal)_ — the graph-side keys-first statement shape this complements
- [BE-705](https://linear.app/hash/issue/BE-705) _(internal)_ — follow-up: dedicated entities-page endpoint
- [Slack: entities-page performance & sort keys](https://hashintel.slack.com/archives/C022217GAHF/p1784220011668439) _(internal)_

## 🔍 What does this change?

- `buildEntitiesFilter` accepts the type universe and sends it as an `any(equal(type.versionedUrl, …))` clause when no explicit type clause exists. The noisy-system-type exclusions stay on top as the semantic recheck: a multi-type entity can carry a noisy type alongside a universe type, so the include list alone would not be result-equivalent.
- `useAvailableTypes` exposes the universe (plus an error and a refetch handle). Its own summarize query keeps the exclusion-only shape — it defines the universe and has no sort+limit, so the pathological plan shape does not apply to it.
- `useEntitiesVisualizerData` holds both the subgraph query and the count summarize back until the universe is available, reports `loading` accordingly, and guards the exposed `refetch` from punching through the gate (Apollo's refetch ignores `skip`).
- A failed universe query surfaces as an error state with a retry button instead of an endless spinner.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- A type created between the summary and the page query stays invisible until the summary refetches. This matches what the filter UI shows anyway; `TODO(BE-705)` removes the window by running both in one transaction server-side.
- Cold loads serialize the summarize and the page query (one extra hop) — the price for never sending the unestimable filter shape.

## 🐾 Next steps

- [BE-705](https://linear.app/hash/issue/BE-705): a dedicated Graph endpoint runs the summary and the page query in one transaction, replacing this client-side coordination.

## 🛡 What tests cover this?

- The entities-visualizer area has no automated test coverage; verified via type checks, lint, and manual testing of the page (default view, type selection, pinned types, pagination).

## ❓ How to test this?

1. Open `/entities` and check the network tab: `queryEntitySubgraph` now fires after the `summarizeEntities` response and its filter carries an `any(equal(type.versionedUrl, …))` clause alongside the exclusions.
2. Filter, sort, and paginate as usual — behavior is unchanged.
3. Error path: block the summarize request (devtools request blocking) and reload — the page shows an error state with a working "Try again" button instead of an endless spinner.

## 📹 Demo

Filter shape of the default view, before → after:

```jsonc
// before: exclusions only — unestimable
{ "all": [ …, { "notEqual": [{ "path": ["type", "baseUrl"] }, …] }, ×12 ] }

// after: include universe first, exclusions kept as recheck
{ "all": [ …,
  { "any": [ { "equal": [{ "path": ["type", "versionedUrl"] }, …] }, … ] },
  { "notEqual": [{ "path": ["type", "baseUrl"] }, …] }, ×12
] }
```

### DIFF
--- a/apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-visualizer-data.tsx
+++ b/apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-visualizer-data.tsx
@@ -1,5 +1,5 @@
 import { useQuery } from "@apollo/client";
-import { useMemo } from "react";
+import { useCallback, useMemo } from "react";
 
 import { getLatestEntityVertices, getRoots } from "@blockprotocol/graph/stdlib";
 import {
@@ -53,7 +53,15 @@ export type EntitiesVisualizerData = Partial<
    * Note that if is hasCachedContent is true, data for the given query is available before loading is complete.
    * The cached content will be replaced automatically and the value updated when the network request completes.
    */
-  refetch: () => Promise<ApolloQueryResult<QueryEntitySubgraphQuery>>;
+  /**
+   * Refetches the subgraph query only — the type universe keeps its last
+   * snapshot, so entities of types that appeared after it loaded stay hidden
+   * until the universe query itself refetches. Resolves to `undefined` without
+   * querying while the type universe is still awaited.
+   */
+  refetch: () => Promise<
+    ApolloQueryResult<QueryEntitySubgraphQuery> | undefined
+  >;
   subgraph?: Subgraph<EntityRootType<HashEntity>>;
   tableData: EntitiesTableData | null;
   totalResultCount: number | null;
@@ -70,6 +78,19 @@ export const useEntitiesVisualizerData = (params: {
   internalWebs: { webId: WebId }[];
   limit?: number;
   sort?: EntityQuerySortingRecord;
+  /**
+   * The type universe from `useAvailableTypes` — `null` while the summary is in
+   * flight, and permanently for pinned types, which never fetch it. The default
+   * (no type selection) view sends it as an include-type clause and holds its
+   * queries back until it arrives — see {@link buildEntitiesFilter}.
+   */
+  typeUniverse: VersionedUrl[] | null;
+  /**
+   * Set when the type universe failed to load. The gated queries stay skipped,
+   * but `loading` stops reporting `true` so the page can show an error state
+   * instead of an endless spinner.
+   */
+  typeUniverseError?: Error;
   view: VisualizerView;
 }): EntitiesVisualizerData => {
   const {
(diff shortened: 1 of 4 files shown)


apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-visualizer-data.tsx
```diff
@@ -102,10 +125,25 @@ export const useEntitiesVisualizerData = (params: {
...
+      internalWebIds,
+      entityTypeBaseUrl,
+      entityTypeIds,
+      typeUniverse,
+    ],
   );
 
+  // TODO(BE-705): a dedicated Graph endpoint will run the summary and the page
+  //  query in one transaction, replacing this client-side coordination.
+  const awaitingTypeUniverse =
```
alex-e-leon, line 141: @TimDiekmann right now there's a lot of logic in hash that depends on all the types being loaded and available. How are we planning on handling all of this when the type universe gets very large and we need to paginate or similar?

alex-e-leon (approved):
LGTM - but I'm not sure this approach will scale nicely.

apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-visualizer-data.tsx
TimDiekmann: I definitely hear the concerns and totally agree. Important to note is that the type universe fetched through summarize-entities will return only types, which actually have entities visible to the user. Regardless, I raised concerns around this endpoint in general, in particular with the counting-part-per-type. The plan is to entirely move this endpoint to the Graph, the optimization opportunities are better there as we're closed to the database. (This effectively applies to all endpoint, I want to move away from a single-entrypoint-approach (`queryEntities` and siblings) and have more dedicated endpoints.

PR #9095, opened 2026-07-24
BE-705: Move the entities table view onto the table endpoint

TimDiekmann:

Moves the entities table view off the subgraph query and onto the dedicated `queryEntitiesTable` endpoint. The table renders from flat endpoint rows, and its data flow lives in a hook decoupled from the subgraph path — the Grid and Graph views stay on the old query, whose hook goes back to serving only them.

## 🔗 Related links

- [BE-705](https://linear.app/hash/issue/BE-705) _(internal)_
- Stacked on #9094 (the Graph endpoint)

## 🔍 What does this change?

- **`useEntitiesTableQuery`** (new): fetches pages through the endpoint, accumulates rows across pages, deep-merges the closed-type maps and definitions, and processes responses in an effect — Apollo's `onCompleted` swallows exceptions in production. A `requestKey` resets the accumulation when the filters change, consumed cursors dedupe Apollo's cache-then-network double emission, and a failed page keeps prior rows on screen with its cursor unconsumed so a retry reprocesses it.
- **Error surfacing**: contract breaks reach Sentry and the error UI with their message instead of collapsing into a fixed string. A first page without a summary throws like the other contract checks (previously the filter chips loaded forever), pinned types resolving to no version surface as an error without a dead retry button, and the retry banner distinguishes a failed refresh (stale rows shown) from a failed load-more.
- **Type pills**: the table passes its endpoint summary into `useAvailableTypes` through a discriminated `SummarySource`, so the pills come from the same transaction as the page instead of a second summarize call.
- **Filter builders**: the subgraph builder applies the same kind×operator gates as the endpoint builder, so a pill can no longer read as active while the table drops its filter — pinned by a builder-agreement test.
- **Bulk actions**: archive/un-archive works on rows via a structural `ArchivableEntity` slice, so table rows qualify without carrying full entity metadata. One-sided links (an endpoint hidden by permissions) render their visible side.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The Grid and Graph views still run the subgraph query, so the old data hook and the subgraph filter builder stay alive until they move too.
- The accumulation fold inside `useEntitiesTableQuery` is not unit-tested yet — extracting it into a pure function with vitest coverage is a planned follow-up.

## 🐾 Next steps

- Move the Grid and Graph views onto dedicated queries, deleting the subgraph path and the type-universe gating
- Extract the page-accumulation fold from `useEntitiesTableQuery` and unit-test it

## 🛡 What tests cover this?

- `build-property-filter-clause.test.ts`: value coercion, incomplete and kind-mismatched filters staying inert, and the endpoint/subgraph builder agreement on null-ness
- The endpoint behavior itself is covered by the integration tests in the base PR

## ❓ How to test this?

1. Checkout the branch, run `yarn dev`, open `/entities`
2. Page through a table (scroll to load more), toggle type pills, web filters, archived, and property filters — the counts and rows should stay consistent
3. Sort by different columns and page again — continuation pages keep the sort
4. Pin a type via an entity-type page and confirm the table scopes to it

## 📹 Demo

_(screenshots to follow — UI is visually unchanged, the data path is new)_


alex-e-leon:
@TimDiekmann did a quick test, and it looks like this breaks the grid view:

Screenshot 2026-07-29 at 14 20 37

TimDiekmann:
Thanks, @alex-e-leon. Where did you test this? Because the Backend is not deployed yet, it's based on https://github.com/hashintel/hash/pull/9094.

alex-e-leon:
> Thanks, @alex-e-leon. Where did you test this? Because the Backend is not deployed yet, it's based on #9094.

locally on this branch

TimDiekmann:
Thanks, I'll take a look once the base branch is merged! 🙂

apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
```diff
@@ -279,6 +339,14 @@ export const EntitiesVisualizer: FunctionComponent<{
     view,
   });
 
+  useEffect(() => {
+    setTableSummary(tableQuery.summary);
+  }, [tableQuery.summary]);
+
+  useEffect(() => {
+    setTableSummaryError(tableQuery.error);
+  }, [tableQuery.error]);
```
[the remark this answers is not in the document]
TimDiekmann: Claude, please just shut up, thank you.

apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
```diff
@@ -633,14 +821,34 @@ export const EntitiesVisualizer: FunctionComponent<{
...
-            size="small"
-          >
-            Try again
-          </Button>
+          {blockingError ? (
+            <Typography
+              variant="smallTextParagraphs"
+              sx={{ color: ({ palette }) => palette.gray[70] }}
+            >
+              {blockingError.message}
```
alex-e-leon, line 829: @TimDiekmann - just checking does the BE actually return error messages intended for users to read? Or are they internal/dev error messages

apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-table-query.tsx
```diff
@@ -0,0 +1,542 @@
...
+
+/**
+ * The pages accumulated for one set of request filters. A response for
+ * different filters resets the accumulation, so stale pages never mix into a
+ * new sequence.
+ *
+ * Everything here belongs to its sequence and goes when the sequence does —
+ * unlike the types the pages resolved, which {@link ResolvedTypes} keeps.
+ */
+type AccumulatedPages = {
```
alex-e-leon, line 78: Why are we holding + managing our own accumulated pages? These should already be cached by apollo client normally. If there's a valid reason for this (like unstable cursors) no need to fix in this PR - but I'm a little surprised we're doing this.

alex-e-leon (approved):
LGTM - left 2 small comments, not blocking though.

apps/hash-frontend/src/pages/shared/entities-visualizer.tsx
TimDiekmann: The responses itself are harmless, though, not very relevant to the user. Either, it's a frontend contract issue (such as that the response doesn't carry a summary), that's simply a bug on our side (and we should get an error in the logs/Sentry in this case anyway). The other errors are actual errors when calling the Node API, I don't think it's helpful to the user to see these messages.

apps/hash-frontend/src/pages/shared/entities-visualizer/use-entities-table-query.tsx
TimDiekmann: I think you're right and the cursor is stable (it even captures the time).

PR #9121, opened 2026-07-29
SRE-870: Removes preview from stable rust components, fixing mise warning

alex-e-leon:

See title.

This PR:

The changes in this PR:

### DIFF
--- a/.github/actions/install-tools/install-rust.sh
+++ b/.github/actions/install-tools/install-rust.sh
@@ -84,13 +84,7 @@ for component in $COMPONENTS; do
   for mirror in "${MIRRORS[@]}"; do
     if install_component "$component" "$mirror"; then
       # Verify component was installed
-      verify_name="$component"
-      case "$component" in
-        "llvm-tools-preview") verify_name="llvm-tools" ;;
-        "rustc-codegen-cranelift-preview") verify_name="rustc-codegen-cranelift" ;;
-      esac
-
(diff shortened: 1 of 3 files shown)


TimDiekmann (changes requested):
We should update `.github/actions/install-tools/` as well

alex-e-leon:
> We should update `.github/actions/install-tools/` as well

@TimDiekmann done

.github/actions/install-tools/install-rust.sh
```diff
@@ -86,8 +86,8 @@ for component in $COMPONENTS; do
       # Verify component was installed
       verify_name="$component"
       case "$component" in
-        "llvm-tools-preview") verify_name="llvm-tools" ;;
-        "rustc-codegen-cranelift-preview") verify_name="rustc-codegen-cranelift" ;;
+        "llvm-tools") verify_name="llvm-tools" ;;
+        "rustc-codegen-cranelift") verify_name="rustc-codegen-cranelift" ;;
```
TimDiekmann: This is now a 1:1 mapping, I guess we should just inline `"^$component"` where we used `"^$verify_name"` below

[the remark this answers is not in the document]
TimDiekmann: first 😂

alex-e-leon: Ok should be fixed now

TimDiekmann (approved):

PR #9130, opened 2026-07-31
FE-1207: add chip to ds

alex-e-leon:

Adds a chip component to the DS

This PR:

The changes in this PR:


lunelson (approved):
Looks good, my only nitpick is that I believe the remove/dismiss box for chips and tabs should always be on the left or rather at block-start, never the right (block-end). It's widely done on the right (maybe a bad convention inherited from microsoft windows) but once you need to remove several chips of different lengths in a row, in a flex-flow ltr layout, it's clear why this should be at block-start

alex-e-leon:
> Looks good, my only nitpick is that I believe the remove/dismiss box for chips and tabs should always be on the left or rather at block-start, never the right (block-end). It's widely done on the right (maybe a bad convention inherited from microsoft windows) but once you need to remove several chips of different lengths in a row, in a flex-flow ltr layout, it's clear why this should be at block-start

I agree with you, but I also think it reads better/the intention of the chips are clearer when remove is at the end. 

Personally adding an X at the beginning would confuse me into thinking that the chip represents the item not being present in the list.

Instead I think a better pattern is a) in combo boxes allowing the whole lozenge to be clicked to remove it (instead of just the button) and b) allowing users to remove items with backspace, which solve the issue but keep readability better.

PR #9132, opened 2026-07-31
SRE-854: Serve the health probe on GET /health across the services

TimDiekmann:

Probe paths differed per service — `/health` on the graph binaries and the Temporal workers, `/health-check` on the Node API, nothing at all on petrinaut-opt — so every consumer (ALB target groups, container healthchecks, the Alloy span-drop map) carried per-service special cases. Every service now answers `GET /health` the same way.

The old paths stay in place, so the probe definitions in internal-infra can switch independently of this deploy. Removing them is SRE-884.

## 🔗 Related links

- [SRE-854](https://linear.app/hash/issue/SRE-854/unify-service-health-endpoints-on-get-health) _(internal)_
- [SRE-884](https://linear.app/hash/issue/SRE-884/remove-the-legacy-health-endpoints) _(internal)_ — removes the old paths once internal-infra has switched

## 🚫 Blocked by

- [ ] #9131 — this branch stacks on it, because both touch `hash-backend-utils/opentelemetry.ts`

## 🔍 What does this change?

- Adds `libs/@local/graph/api/src/rest/probe.rs`, serving `GET /health` with `{"status":"pass"}` as `application/health+json`. That is the only member `draft-inadarei-api-health-check` requires, which leaves room to grow the document later without changing the contract now.
- Uses that router in the Graph API, the Graph Admin API and Atlas. The type fetcher answers the same body inline, because it cannot depend on `hash-graph-api` — that crate depends on it.
- Excludes probes from tracing in the two places that create HTTP server spans: `HttpTracingLayer` returns a disabled span for the path, and `createHttpInstrumentation` ignores it, which covers the Node API and both Temporal workers at once. Otherwise a probe answered every few seconds per task dominates the RED metrics derived from spans.
- The Graph API's `/health` was already untraced, but only as a side effect of `openapi_only_router()` being merged after the layers. The explicit filter no longer depends on that ordering.
- Keeps probes free of dependency checks. One that reaches through to Postgres takes every task out of rotation as soon as Postgres is slow, turning a degraded dependency into an outage.
- The worker health server now also answers `HEAD`, which is what `hash-graph … --healthcheck` and `wget --spider` issue.
- Points the Atlas healthcheck subcommand at `/health` and drops a stale `/status` mention from the subcommand docs.
- petrinaut-opt gains `/health` and keeps `/status`, which is not a probe — it returns the status of every optimization run. Its generated OpenAPI spec is regenerated accordingly.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Nothing in internal-infra changes yet, so the probes still target the old paths after this merges. That is intentional — both paths answer.
- The `dropped_spans` map in internal-infra becomes dead once its probes move to `/health`, but its own validation rejects an empty map. Whoever prepares that change needs to decide between a placeholder entry and relaxing the Alloy filter.

## 🐾 Next steps

- Switch the probe definitions in internal-infra (`catalog/services/app/api.tf`, `catalog/services/ingress/defaults.tf`, `catalog/services/app/petrinaut_opt.tf`).
- SRE-884 removes the old endpoints afterwards.

## 🛡 What tests cover this?

- No new automated tests. The endpoints are constant-response routes, and the interesting property — that they emit no spans — cannot be asserted without a running collector.
- Verified by hand against a running Node API and a running Atlas stub, reading the resulting spans out of Tempo. Both directions were checked: the new path produces no span while the old path still does, so the check is not vacuously green.

## ❓ How to test this?

1. Bring up the observability profile, then start the service you want to check (`yarn workspace @apps/hash-api dev`, or `cargo run --bin hash-graph -- atlas` with `HASH_OTLP_ENDPOINT` set).
2. `curl -i localhost:5001/health` — expect `200`, `application/health+json` and `{"status":"pass"}`. `curl -I` should also answer `200`.
3. Confirm the old path still answers: `curl -i localhost:5001/health-check`.
4. In Grafana on `localhost:3001`, allow a few seconds for ingest, then confirm no trace exists for the new path while one exists for the old path.

## 📹 Demo

Not applicable — no user-facing surface.


CiaranMn (dismissed):

TimDiekmann:
@copilot resolve the merge conflicts in this pull request

CiaranMn (approved):

PR #9152, opened 2026-08-03
BE-743: Make the semantic-search candidate pool and ef_search configurable

TimDiekmann:

The semantic search's candidate pool was `limit × 4` with the factor compiled in, and `hnsw.ef_search` was derived from the pool alone. Both are now store settings, so the pool depth can be tuned against real policies and data without a rebuild.

Deriving `ef_search` from the pool also leaves the index walk shallow whenever a search asks for few results, which is where a missed neighbour is most likely. A configurable floor decouples how deep the walk goes from how much the exact re-scoring has to read.

## 🔗 Related links

- [BE-743](https://linear.app/hash/issue/BE-743) — this change
- [BE-618](https://linear.app/hash/issue/BE-618) — the search rework this builds on; recall measurements and the `ef_search` findings are recorded there

## 🚫 Blocked by

- Stacked on `t/be-618-investigate-slow-top-nav-search-query-performance`, so that one merges first.

## 🔍 What does this change?

- New `SemanticSearchSettings` on `PostgresStoreSettings`, holding the candidate multiplier (`NonZero<usize>`, default 4) and an `ef_search` floor (default 400)
- The derivation moved with it: `candidate_pool(limit)` saturates instead of wrapping, `ef_search(pool)` takes the larger of pool and floor and clamps to the range pgvector accepts. The bounds live in one named constant rather than as a literal `clamp(1, 1000)` next to the query
- Both search paths read the settings; `QUANTIZED_RANK_OVERFETCH` is gone
- Two server flags with `HASH_GRAPH_SEMANTIC_SEARCH_CANDIDATE_OVERFETCH` and `HASH_GRAPH_SEMANTIC_SEARCH_MINIMUM_EF_SEARCH`

The floor only takes effect below a limit of 100 — above that the pool is deeper than 400 and decides the walk itself. A search for a single record previously walked with `ef_search = 4`.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The defaults are unchanged from BE-618 on purpose: the factor is the measured-optimistic end from a 100k sample without filters, and the floor of 400 is a deliberate middle ground rather than pinning `ef_search` at its maximum. Both want a measurement against this code with real policies before they move, which is what making them configurable is for.

## 🐾 Next steps

- Measure factor and floor over several limits with real policies, then decide whether the factor should become a function of the limit rather than a constant
- The type search shares the multiplier but sets no `ef_search`, since `entity_type_embeddings` has no HNSW index — its pool is quantization reserve only

## 🛡 What tests cover this?

- Unit tests for the derivation: pool scaling including saturation, `ef_search` holding the whole pool, the floor applying to small pools, and both bounds of the accepted range
- The existing search tests exercise the default settings end to end

## ❓ How to test this?

1. Start the graph with `HASH_GRAPH_SEMANTIC_SEARCH_CANDIDATE_OVERFETCH=16`
2. Run a search and confirm from the query log that the candidate reads use `LIMIT limit × 16` and `hnsw.ef_search` follows
3. With a small limit, confirm `ef_search` stays at the floor rather than following the pool


TimDiekmann:
@copilot resolve the merge conflicts in this pull request

libs/@local/graph/postgres-store/src/store/postgres/mod.rs
```diff
@@ -91,6 +93,55 @@ use crate::store::error::{
...
+    ///
+    /// Deriving the search list from the candidate pool alone leaves the walk shallow for
+    /// searches asking for few results, which is where a missed neighbour is most likely.
+    pub minimum_ef_search: usize,
+}
+
+impl Default for SemanticSearchSettings {
+    fn default() -> Self {
+        Self {
+            candidate_overfetch: NonZero::new(4).expect("4 is not zero"),
```
indietyp, line 120: you could in theory do `const {NonZero::new(4).unwrap() }` here instead, to have it const constructible, but it doesn't make much difference.

indietyp (approved):

PR #9165, opened 2026-08-05
FE-1322: Generate Petrinaut architecture docs from in-code annotations

kube:

Petrinaut's architecture documentation had no mechanism keeping it true. This makes the architecture something you **declare next to the code it describes**, generates the docs from those declarations, and fails CI when a declaration stops matching the code.

The output is a **portable bundle**, not a website — the same artefact renders locally, embeds into `hash.dev/docs/petrinaut`, or gets handed to an AI agent.

```mermaid
flowchart LR
  A["@layerRoot + @role<br/><i>in doc comments and README frontmatter</i>"]
  C["content/<br/><i>authored MDX, optional</i>"]
  E{{"Extractor<br/>+ dependency-cruiser"}}
  D{{"d2"}}
  F["bundle/<br/><i>architecture.json · architecture.md<br/>pages · diagrams · components</i>"]
  G["Starlight site"]
  H["hash.dev"]
  I["AI agents"]

  A --> E
  E -->|"layer model +<br/>real import graph"| F
  E --> D
  D -->|"44 SVGs"| F
  C --> F
  F --> G
  F --> H
  F --> I
```

### 1. Two annotations describe the whole architecture

A declaration is two lines — which layer a folder forms, and what it is for:

```ts
/**
 * @layerRoot core.simulation.monte-carlo
 * @role Runs many simulations with bounded frame memory
 */
```

A folder `README.md` can declare the same thing in frontmatter, and its prose becomes that layer's page:

```yaml
---
layer: core.simulation.monte-carlo
role: Runs many simulations with bounded frame memory
---
```

**Files with no annotation inherit from the nearest declaring ancestor**, which is what keeps this proportional to the architecture rather than to the file count: **37 declarations cover 412 files**. Everything else — layer sizes, 177 dependency edges, the parent/child tree — is derived from the real TypeScript import graph.

The vocabulary is deliberately this small. Both tags are needed to place a node in the graph and label it; anything more would be a claim the generator cannot check.

### 2. Every layer gets a diagram

D2, rendered to SVG at build time. **44 diagrams**, in three kinds that each bound their node count differently:

- **Overview** (1) — the top-level layers, everything below folded into its parent.
- **Neighbourhood** (37) — one per layer: what it depends on, what depends on it. Every layer has one, including leaves, which is where readers land.
- **Drill-down** (6) — for a layer with sub-layers, its direct children.

Aggregation is honest: an edge appears because imports exist, and counts are real `fileDependencies` sums. Neighbours are capped at 12 — `core.types` has 18, so the rest becomes a dashed "+6 further layers" node carrying its 8 dependencies rather than being dropped.

### 3. Hand-written content merges into the generated tree

`content/` is optional (the system works with the directory absent) and carries the reasoning an import graph cannot express. A page names the layer it explains and nests beneath it:

```yaml
---
title: Memory model
attachTo: core.simulation
---
```

The layer's page gains a **Guides** section. Because `attachTo` decides a page's depth, links are written by name — `[text](layer:core.simulation.engine)`, `[text](doc:simulation/memory-model)` — and resolved at emit time. Unresolved targets fail the build. Authored pages may import diagram components from the bundle (`@diagrams/byte-map`).

### 4. CI enforces the structure

`lint:arch-docs` fails on: an unannotated source file, a layer id implying an undeclared ancestor, a duplicate declaration, a malformed tag, an `attachTo` or link target that does not resolve, a configured package whose language has no extractor, and any dependency violating a rule in `architecture.config.ts`.

Every check is a statement about the graph, which is the whole of what the docs assert. Four rules are enforced; the substantive one — `react` must not depend on `ui` — already held (0 imports against 235 the other way), so it locks in an existing property.

### 5. What this replaces

- `petrinaut-core/scripts/generate-dependency-diagrams.mjs` held the architecture as ~180 lines of `if (path.startsWith(...))` far from the code, with a fallback that silently mis-bucketed anything renamed. It also hard-coded 7 of `petrinaut-core`'s 10 entry points, so imports through `./ai`, `./optimization` and `./compiled-model` were absent from the diagrams entirely.
- The 3,100 lines of hand-written HTML in `docs/architecture/` are migrated to authored MDX. Doing so surfaced two stale claims: UUID support described as unimplemented (it is fully implemented), and a sandbox description thinner than the current implementation.

## 🔗 Related links

- FE-1322 (this PR)
- FE-1157 — moving Petrinaut docs to `hash.dev/docs/petrinaut`. This produces the exportable artefact that work needs; it publishes nothing itself.
- FE-1139 — the ticket that produced the HTML pages migrated here.

`@hashintel/petrinaut` and `petrinaut-core` are touched only by comments, READMEs, and removal of a private script and its `dependency-cruiser` devDependency. No runtime code, types or exports change.

The user-facing guide (`libs/@hashintel/petrinaut/docs/`) is untouched — no UI or behaviour changed. `AGENTS.md` gains a section on declaring layers and what CI enforces.

The docs build lives entirely in the task graph — no repo-root entry point, nothing shelling out across packages:

```
@apps/petrinaut-docs#{build,lint:tsc,dev} → #sync:bundle → @local/petrinaut-arch-docs#doc:architecture
```

`doc:architecture` is deliberately **not cached**: Turborepo hashes a package plus its dependencies' task *outputs*, and the annotations this reads are source comments in `petrinaut` and `petrinaut-core` — nobody's output. A cached bundle would survive an annotation change and go quietly stale, which is the rot this package exists to catch. It declares `outputs`, so consumers depend on the task rather than on the directory existing.

Removes `doc:dependency-diagram` from `petrinaut-core` and the repo-root `mise run doc:architecture` task.

## ⚠️ Known issues

- **The taxonomy is a tree, so cross-cutting concerns have no home.** Every layer has exactly one parent. `core.types` is the symptom already in the model: four distinct parents across two packages depend on it, so it is filed under `core` while behaving like a shared foundation under everything.
- **Nothing verifies `@role`.** The structure is checked against the import graph; the one-line description of each layer is prose. Declaring it beside the code makes it likelier to be corrected when that code changes, but CI does not hold it to account.
- **Richer annotations are written but unread.** `@boundary`, `@invariant`, `@entryPoint` and `@layerName` remain in the Petrinaut source and are ignored by this version. Unknown tags are skipped by design and README frontmatter is tolerant, so re-reading one is a schema change in the generator, not a rewrite of the packages. Layer names are meanwhile derived from the id's last segment, so a few read plainly (`ui` renders as "Ui").
- **Three packages are not covered.** `petrinaut-cli`, `petrinaut-website` and `petrinaut-opt` have no declarations. A TypeScript package is a config entry plus one root declaration; the Python app needs docstring extraction, which is not written — configuring a package for a language with no extractor is a hard error rather than a silent no-op.
- **Hosting the bundle asks one thing**: a React-capable MDX pipeline, for the authored pages that import diagram components. Nothing hydrates, so no client-side runtime is required. Generated pages are plain CommonMark, diagrams are static SVG, and `architecture.md` has no components at all.
- **The docs site is local-only.** Nothing deploys it yet — see next steps.
- `mise run fix:package-json` could not run locally (needs a nightly Cargo feature), so `package.json` key ordering was verified by reading the sorter's field list.

## 🐾 Next steps

- Decide where the site is deployed. `demo.petrinaut.org/docs` is feasible — the SPA has no catch-all rewrite, and an Astro `base: "/docs"` build was tested — but it needs `d2` added to the Vercel install step, and is worth settling against FE-1157 first so there is one canonical URL.
- Cover the remaining three packages.

## 🛡 What tests cover this?

48 tests in `@local/petrinaut-arch-docs`:

- `tags.test.ts` — tag grammar: multi-line continuation, duplicates, typo suggestions, that a tag named in prose is not a declaration, and that an unread annotation is ignored rather than rejected.
- `frontmatter.test.ts` — declarations, malformed YAML, half-written declarations, CRLF, unknown keys ignored.
- `extract.test.ts` — inheritance through undeclared folders, uncovered files, stable ordering.
- `check.test.ts` — each CI check in both directions: fires when broken, silent when not.
- `emit/mdx.test.ts` — link resolution at varying depths, fragments, unresolved targets.

Existing suites unaffected: 842 (`petrinaut-core`), 187 (`petrinaut`).

## ❓ How to test this?

```bash
turbo run dev --filter @apps/petrinaut-docs   # http://localhost:4321
```

Open **Architecture** for the overview, then any layer — including a leaf such as `core.clipboard` — and confirm it opens with a diagram of what it depends on and what depends on it. `core.types` exercises the neighbour cap: 12 drawn, plus a dashed "+6 further layers" node.

To watch the task graph do its job, delete the bundle first — the build regenerates it, syncs it, then renders, as three ordered tasks:

```bash
rm -rf libs/@local/petrinaut-arch-docs/bundle
turbo run build --filter @apps/petrinaut-docs
```

Confirm the checks hold, and hold in either order — formatting and generation both used to claim the bundle's files:

```bash
yarn fix:format && yarn workspace @local/petrinaut-arch-docs doc:architecture
yarn lint:format                                          # expect 0
yarn workspace @local/petrinaut-arch-docs lint:arch-docs  # expect 0
```

Then break something and confirm it is caught: change a `role:` in any layer-declaring README, point an `attachTo` at a layer that does not exist, or reference `layer:core.nonexistent` — each fails `lint:arch-docs` with the offending file named.

For the AI-facing side, read `libs/@local/petrinaut-arch-docs/bundle/architecture.md` — the whole architecture in one file.


.claude/launch.json
```diff
@@ -0,0 +1,17 @@
...
+        "workspace",
+        "@apps/petrinaut-docs",
+        "dev",
+        "--port",
+        "4321"
+      ],
+      "port": 4321
+    }
+  ]
+}
```
kube: @TimDiekmann @CiaranMn what do you think of having this `.claude/launch.json` in the repo?

apps/petrinaut-docs/LICENSE.md
```diff
@@ -0,0 +1,5 @@
+# License
```
vilkinsons, line 1: Thee docs need switching from AGPLv3 to Apache 2.0/MIT dual-license.

apps/petrinaut-docs/astro.config.mjs
```diff
@@ -0,0 +1,165 @@
...
+    // pages stay plain Markdown and never need this.
+    react(),
+    starlight({
+      title: "Architecture Docs",
+      description:
+        "How the Petrinaut packages fit together — generated from annotations in the source.",
+      // The helmet carries the Petrinaut identity, so the title beside it names
+      // only what this site is. `replacesTitle: false` keeps both.
+      logo: {
+        src: "./src/assets/petrinaut-helmet.png",
```
vilkinsons, line 144: Half a megabyte!

libs/@hashintel/petrinaut-core/src/hir/README.md
```diff
@@ -1,3 +1,18 @@
+---
+layer: core.hir
+name: HIR compiler
+role: Lowers user-authored TypeScript to a source-spanned IR, then typechecks, lints and emits it
+seams:
```
vilkinsons: Can we use a term other than `seams`? Feels very AI.

lunelson:
Nice initiative, I've been thinking about doing something similar.  One thing you might consider adding here is [drift](https://github.com/fiberplane/drift).

Current checks cover facts (ancestors, exports, dep rules, links), but not whether a given sentence is still true; whereas `drift` binds a markdown file to a source file or AST symbol, so you can use `drift check` to trace when your doc goes (potentially) stale. 

The docs stay pure markdown(bindings live in `drift.lock`, not in frontmatter) so it doesn't collide with `attachTo` or the emit pipeline. 

It might also be a good answer to the `petrinaut-opt` item under Known Issues, since drift parses Python natively

Adoption, roughly, according to claude:

- Bind each `content/**.mdx` page to the handful of symbols it actually describes, e.g. `drift link libs/@local/petrinaut-arch-docs/content/simulation/memory-model.mdx <frame buffer allocation site>#<symbol>`. Roughly 15–20 bindings total.
- Add `drift check --changed` to the existing `lint:arch-docs` CI job; it exits 1 on stale.
- Skip its broken-link checking — `emit/mdx.ts` already resolves links, and doubling up would just produce two error formats for one problem.
- Don't let bindings substitute for the generator's coverage check; drift is blind to any file nobody bound, so the two checks are doing different jobs.

kube:
> Nice initiative, I've been thinking about doing something similar. One thing you might consider adding here is [drift](https://github.com/fiberplane/drift).
> 
> Current checks cover facts (ancestors, exports, dep rules, links), but not whether a given sentence is still true; whereas `drift` binds a markdown file to a source file or AST symbol, so you can use `drift check` to trace when your doc goes (potentially) stale.
> 
> The docs stay pure markdown(bindings live in `drift.lock`, not in frontmatter) so it doesn't collide with `attachTo` or the emit pipeline.
> 
> It might also be a good answer to the `petrinaut-opt` item under Known Issues, since drift parses Python natively
> 
> Adoption, roughly, according to claude:
> 
> * Bind each `content/**.mdx` page to the handful of symbols it actually describes, e.g. `drift link libs/@local/petrinaut-arch-docs/content/simulation/memory-model.mdx <frame buffer allocation site>#<symbol>`. Roughly 15–20 bindings total.
> * Add `drift check --changed` to the existing `lint:arch-docs` CI job; it exits 1 on stale.
> * Skip its broken-link checking — `emit/mdx.ts` already resolves links, and doubling up would just produce two error formats for one problem.
> * Don't let bindings substitute for the generator's coverage check; drift is blind to any file nobody bound, so the two checks are doing different jobs.

Thanks @lunelson, it actually seems to be a good thing to add on top.

For this first version I'll keep things simple so we can already provide some value with docs, see graph of dependencies etc...

But definitely a good tool to add in a next step.

kube:
I close this PR as it is now split in three:
- https://github.com/hashintel/hash/pull/9204
- https://github.com/hashintel/hash/pull/9205
- https://github.com/hashintel/hash/pull/9206

PR #9167, opened 2026-08-05
SRE-924: Mint the release token from a Vault-held key

TimDiekmann:

The release job read a GitHub App's private key out of the legacy Vault and
minted an installation token with no repository scope, using an App that also
carries `workflows: write`. It now signs as a dedicated release App whose key
never leaves Vault, and the token it gets is scoped to this repository.

## 🔗 Related links

- hashintel/.github#95 — the composite action this uses
- hashintel/internal-infra#321 — the transit mount, key and role _(internal)_

## 🚫 Blocked by

- [ ] hashintel/.github#95, because the action is pinned to a commit on that branch

## 🔍 What does this change?

- `release.yml` drops the `vault-action` secret read and `create-github-app-token`
  in favour of `hashintel/.github/.github/actions/github-app-token`, which has
  Vault sign the App JWT
- the job now declares `environment: main`, which is what resolves the production
  Vault address, and puts the environment claim in the OIDC token so the Vault
  role can bind it
- authenticates as the `hash-release` App (four permissions) rather than
  `hash-worker` (nine)

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The action is pinned to an unmerged commit of hashintel/.github#95, marked in
place with `# unmerged — re-pin to main`. It has to be re-pinned once that PR
merges, and this PR must not merge before then.

## 🐾 Next steps

Once a release run is green, the App key can be revoked in the App settings.

## 🛡 What tests cover this?

- None automated. The JWT assembly was verified against the GitHub API, and the
  imported key was verified byte-for-byte against its PEM.

## ❓ How to test this?

1. Merge hashintel/.github#95 and re-pin the action to main
2. Push to `main`
3. Confirm the release job obtains a token and changesets runs; a mismatched role,
   claim or key fails visibly at the Vault login or the App JWT exchange

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,8 +25,8 @@ jobs:
           app-id: "4497272"
           transit-key: github-app-hash-release
           repository: ${{ github.repository }}
-          cf-access-client-id: ${{ vars.CF_ACCESS_CLIENT_ID }}
-          cf-access-client-secret: ${{ secrets.CF_ACCESS_CLIENT_SECRET }}
+          cf-access-client-id: ${{ vars.CF_ACCESS_PROD_CLIENT_ID }}
+          cf-access-client-secret: ${{ secrets.CF_ACCESS_PROD_CLIENT_SECRET }}
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,30 +11,21 @@ jobs:
   release:
     runs-on: ubuntu-latest
     if: github.repository == 'hashintel/hash'
+    environment: main
     permissions:
       id-token: write
 
     steps:
-      - name: Authenticate Vault
-        id: secrets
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -11,30 +11,22 @@ jobs:
...
         with:
-          app-id: ${{ steps.secrets.outputs.GITHUB_WORKER_APP_ID }}
-          private-key: ${{ steps.secrets.outputs.GITHUB_WORKER_APP_PRIVATE_KEY }}
+          vault-address: ${{ vars.VAULT_ADDR }}
+          vault-role: ci-hash-release
+          app-id: "4497272"
+          transit-key: github-app-hash-release
+          repository: ${{ github.repository }}
+          cf-access-client-id: ${{ vars.CF_ACCESS_CLIENT_ID }}
+          cf-access-client-secret: ${{ secrets.CF_ACCESS_CLIENT_SECRET }}
```
TimDiekmann:
```suggestion
          cf-access-client-id: ${{ vars.CF_ACCESS_PROD_CLIENT_ID }}
          cf-access-client-secret: ${{ secrets.CF_ACCESS_PROD_CLIENT_SECRET }}
```

CiaranMn (approved):

PR #9178, opened 2026-08-08
FE-1341: Give experiment compute backends one interface

kube:

Adds `@hashintel/petrinaut-core/experiments`: one interface both compute backends satisfy, and a registry that picks between them. `ExperimentsProvider` hardcodes the backend choice today; #9179 registers a backend against this interface instead of rewiring the provider. App behaviour is unchanged.

This PR sits above #9177 in stack #9342, and #9179 builds on it.

## 🔗 Related links

- [FE-1341](https://linear.app/hash/issue/FE-1341/make-experiment-compute-backends-swappable-behind-one-interface) _(internal)_: this PR.
- [FE-1340](https://linear.app/hash/issue/FE-1340/run-experiments-on-a-webgpu-compute-backend) _(internal)_: the second backend.
- [FE-1429](https://linear.app/hash/issue/FE-1429/run-trial-replicates-through-the-shared-experiment-backend) _(internal)_: a third consumer, the Petrinaut CLI's seeded optimization trials, adopts this interface for parallelism.

## 🚫 Blocked by

- [ ] #9177

## 🔍 What does this change?

Both paths already produce a `MonteCarloExperiment`, and consumers drive one without branching. That half is untouched. This PR adds four types:

- `ExperimentRequest` describes what to compute, as serializable data. Worker factories, the ODE method, signals, and callbacks describe how to compute it, and belong to backend construction or per-call options.
- `ExperimentAssessment` is a backend's answer to "can you run this?", as a value. Each blocker carries a `code`, an optional `itemId`, and an `origin`.
- `ExperimentBackend.assess` checks the net without acquiring a device or worker pool, then returns an `instantiate()` closure that reuses that work.
- `selectExperimentBackend` walks registrations in preference order and records each refusal.

A blocker's `origin` tells a UI what to do: `model` means edit the net, `configuration` means edit the experiment, `environment` means hide the option, and `capacity` means retry or use fewer runs. Each registration's `load` is deferred, so a heavy backend stays out of a bundle that never uses it.

Adds `content/experiments/backend-selection.mdx` and the `core.experiments` architecture layer.

**Review fixes**

- `selectExperimentBackend` treats a thrown `assess()` or `instantiate()` as that backend's refusal and moves to the next candidate instead of abandoning the walk. Abort errors are re-thrown, and the walk stops before assessing a candidate when the signal is already aborted, so a cancelled experiment cannot fall through and start on the fallback backend.
- The worker-pool backend re-throws `AbortError` instead of reporting the cancellation as a configuration blocker.
- The provider imports `WORKER_POOL_BACKEND_ID` instead of duplicating the literal.
- The module docstring moved above the imports, and the `load()` doc no longer claims a promise cache that does not exist.
- A patch changeset covers the new `@hashintel/petrinaut-core/experiments` entry point.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- One backend is registered here, so `assess` never refuses and the blocker types stay unexercised until #9179.
- Worker pools and `GPUDevice` are still acquired per experiment. Reuse depends on the worker resetting frame and metric state; a worker that kept state would produce plausible-looking wrong results.

## 🐾 Next steps

Supply backends through a React context, so an embedder can register its own.

## 🛡 What tests cover this?

`select-experiment-backend.test.ts` runs 9 cases covering preference order, deferred loading, refusal recording, instantiation failure, blocker ordering, unavailable backends, module load failure, and per-requirement request building.

## ❓ How to test this?

Run an experiment. Behaviour is unchanged.

### DIFF
--- a/libs/@hashintel/petrinaut-core/package.json
+++ b/libs/@hashintel/petrinaut-core/package.json
@@ -26,6 +26,10 @@
       "types": "./dist/index.d.d.ts",
       "import": "./dist/index.js"
     },
+    "./experiments": {
+      "types": "./dist/experiments.d.d.ts",
+      "import": "./dist/experiments.js"
+    },
     "./ai": {
       "types": "./dist/ai.d.d.ts",
       "import": "./dist/ai.js"
--- a/libs/@hashintel/petrinaut-core/package.json
+++ b/libs/@hashintel/petrinaut-core/package.json
@@ -26,6 +26,10 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
+    "./experiments": {
+      "types": "./dist/experiments.d.d.ts",
+      "import": "./dist/experiments.js"
+    },
     "./ai": {
       "types": "./dist/ai.d.ts",
       "import": "./dist/ai.js"
--- a/.changeset/experiments-backend-registry.md
+++ b/.changeset/experiments-backend-registry.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut-core": patch
+---
+
+Add the `@hashintel/petrinaut-core/experiments` entry point: an `ExperimentBackend` interface, a worker-pool implementation of it, and `selectExperimentBackend`, which walks backends in preference order and records every refusal.
--- a/libs/@hashintel/petrinaut-core/src/experiments.ts
+++ b/libs/@hashintel/petrinaut-core/src/experiments.ts
@@ -1,20 +1,16 @@
 /**
- * Swappable experiment backends.
+ * The contract both compute backends satisfy, and the registry that picks one.
  *
- * The runtime half of swappability already existed and is unchanged: both the
- * worker-pool and WebGPU paths produce a `MonteCarloExperiment`, and consumers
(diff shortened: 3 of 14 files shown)


YannisZa (dismissed):

YannisZa (approved):

TimDiekmann (approved):

libs/@hashintel/petrinaut-core/package.json
```diff
@@ -26,6 +26,10 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
+    "./experiments": {
+      "types": "./dist/experiments.d.d.ts",
```
TimDiekmann, line 30: Is the double-d expected?

kube: Thanks for spotting that @TimDiekmann.

There were some bugs with the DTS generation at some point, fixed in https://github.com/hashintel/hash/commit/29946836d4c16cac0b62cb578cfd26c195fd8873, and this is a relicate.

I'll fix that in the next stacked PR, because otherwise I'll end up rebasing everything again, and the PR is already in queue.

PR #9179, opened 2026-08-08
FE-1340: Add an experimental WebGPU compute backend for experiments

kube:
> [!IMPORTANT]
> **Experimental**
> Behind the **WebGPU** and **Compilation output** feature flags.

## Summary

Before this PR, experiments run every Monte Carlo run on the CPU engine, which compiles the net's lowered HIR to buffer programs. There is no GPU path.

This PR adds a WebGPU compute backend for experiments. One invocation per run steps the whole net on the device, with WGSL generated from the same lowered HIR. Metrics reduce on-device into per-frame histograms. Only a compact per-run summary is read back. Backend is opt-in per experiment and falls back to the CPU for nets it cannot run.

## Links

- Ticket [FE-1340](https://linear.app/hash/issue/FE-1340)
- Docs [GPU backend](https://petrinaut-docs-git-cf-fe-1340-webgpu-experiment-backend.stage.hash.ai/architecture/core/webgpu/gpu-backend)
- Docs [Simulation performance › 8. GPU / WebGPU](https://petrinaut-docs-git-cf-fe-1340-webgpu-experiment-backend.stage.hash.ai/architecture/core/simulation/performance#8-gpu--webgpu)
- Docs [User guide](https://github.com/hashintel/hash/blob/cf/fe-1340-webgpu-experiment-backend/libs/@hashintel/petrinaut/docs/README.md)
- Docs [Compilation output](https://github.com/hashintel/hash/blob/cf/fe-1340-webgpu-experiment-backend/libs/@hashintel/petrinaut/docs/compilation-output.md)
- Docs [Experiments](https://github.com/hashintel/hash/blob/cf/fe-1340-webgpu-experiment-backend/libs/@hashintel/petrinaut/docs/experiments.md)

#### Backend

- Backend reports whether it can run a net
  > Check goes through the backend interface. A declined net runs on the CPU with the reason recorded.
- Typed places must declare a capacity
  > Buffer sizes must be known before dispatch.
- Transitions consume at most two typed tokens per place
  > Weight 2 scans every pair by combinatorial unranking and preserves the CPU's firing order.
- Tokens carry no `string` or `uuid` attributes
  > Both need more than the 32 bits WGSL offers.
- Metrics measure place token counts without time aggregation
  > On-device reduction is a histogram.
- Run state stays on the device
  > Shader writes place counts and status to a summary buffer of ~16 bytes per run, which the host decodes.
  > Reading full state back needed host-visible memory equal to the state, capping a 3112-byte-per-run net at ~689k runs.
- Device limits are requested explicitly
  > `requestDevice()` without `requiredLimits` returns the WebGPU floor of 128 MiB per storage binding.
  > An Apple metal-3 adapter reports 4096 MiB.
- Lost GPU device fails the experiment with the loss reason
  > Before, loss surfaced as a truncated "Complete".
  > Driver resets and browser reclaims count as loss.
  > `dispose()` destroys the device intentionally and is filtered out.

#### UI and packaging

- Per-experiment **Run on GPU** switch
- **Compilation** panel behind the **Compilation output** setting
- `./webgpu` entry in `@hashintel/petrinaut-core`
- `core.webgpu` architecture layer

#### Review fixes

- Dead benchmark page, one-shot runner, and HIR duplicate deleted
  > `benchmarks/webgpu-vs-cpu.html` hardcoded absolute `/@fs` paths from one machine's checkout.
  > Unused `runGpuMonteCarloExperiment` module went with its test.
  > `compile-net-shader.test.ts` builds its HIR through `compileHirArtifacts` and `hirFromArtifacts`, the production path, which left `lower-net-hir.ts` a dead duplicate.
- WGSL reserved-word list deleted
  > Mangling prefixes every identifier, so the list had no effect.
- `./webgpu` entry exports only what the app consumes
  > Backend factory, compilation report, and metric-spec check. Three stale entry-point rationale comments corrected.
- `./webgpu` and `./experiments` `types` point at the emitted `.d.ts` files
  > Previous DTS toolchain wrote `.d.d.ts` names.
- GPU profile separates declared capacity from allocated token count
  > Each place's declared token capacity travels apart from the token count allocated for it in the buffer. A capacity on an uncoloured place constrains the shader the same way it constrains the CPU engine.
- Histogram decode truncates trailing sample-free frames
  > Cancelling mid-experiment, or a net whose runs all end early, no longer appends a null-valued tail across the remaining timeline.

## Known issues

- Results are not seed-identical to the CPU, because WebGPU cannot reproduce the CPU generator
  > Trajectories differ and distributions agree within 0.5% on SIR.
  > Dynamics use RK4, more accurate than the CPU's Euler.
  > Backend used is recorded on the experiment.
- GPU steps every run to max time, while the CPU stops a run when it cannot fire
  > Results match; the reported frame count does not. A deadlocked run skips its work and its histogram contribution, so this affects reporting only.
- Nothing validates the generated WGSL in CI
  > Tests assert on emitted text, which is how a same-scope redeclaration reached a user.
  > `web-naga` as wasm or Dawn via `webgpu` with a null backend both run headlessly in vitest.
- Run state is one storage buffer, so `maxStorageBufferBindingSize` bounds the run count
  > Integer attributes store unsigned values, and expression metrics cannot run on-device. Both cases are refused, so results cannot silently diverge.

## Next steps

- Add WGSL validation to the test suite
- Chunk large dispatches
- Add a token-reducing metric kind that gives expression-style metrics a GPU equivalent

## Test coverage

- `compile-net-shader.test.ts`:
  > Shader generation, the summary ABI, and a naga-calibrated redeclaration check across three ODE methods.
- `eligibility.test.ts`, `compilation-report.test.ts`, `emit-wgsl.test.ts`:
  > Refusal rules, the report, and code emission.
- `pair-selection.test.ts`:
  > Pair ordering against the engine's own enumerator.
- `runner.test.ts`:
  > Buffer limits, chunked seeding, and device limits.
- Verified on a real adapter:
  > 4096 SIR runs dispatch with no validation or uncaptured errors and decode correctly.

## How to test

- Open [Petrinaut preview](https://petrinaut-git-cf-fe-1340-webgpu-experiment-backend.stage.hash.ai) on Vercel
- Viewport controls > Settings > Simulation > WebGPU
- Load example
- Simulate > Experiments > Create
- Switch on Run on GPU, Run
- Expect GPU badge in the summary and distributions matching a CPU run of the same configuration
- Load example > Satellites without capacities
- Expect Run on GPU disabled, naming the place that needs a capacity
- Viewport controls > Settings > Simulation > Compilation output
- Expect Compilation tab in the bottom panel


libs/@local/petrinaut-arch-docs/content/simulation/gpu-backend.mdx
TimDiekmann, line 1: As we already have `@local/petrinaut-optimizer-client` I was wondering, if we want to group these into a `@local/petrinaut/{optimizer-client,arch-docs}`

libs/@hashintel/petrinaut-core/src/webgpu/emit-wgsl.ts
```diff
@@ -0,0 +1,696 @@
...
+  #constantIndex(expr: HirExpr, env: ReadonlyMap<string, WgslValue>): number {
+    if (expr.kind === "numberLit") {
+      return expr.value;
+    }
+    // The JS emitter folds constants before emitting; anything still dynamic
+    // here cannot index a compile-time tuple.
+    const value = this.emit(expr, env);
+    const literal = /^-?\d+(?:\.0)?$/u.exec(
+      value.kind === "f32" ? value.code : "",
+    );
```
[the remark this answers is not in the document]
kube: I read this as a false positive, and I have left the thread open rather than dismissing the alert myself.

The pattern is `/^-?\d+(?:\.0)?$/u`. It is anchored at both ends, has no nested quantifier, and its one repetition `\d+` cannot overlap with `(?:\.0)?`, since that alternative starts with a literal `.`. There is a single way to match any input, so it runs in linear time; I measured no growth on long digit runs.

The input is also not external. `value.code` is WGSL that this emitter generated from HIR two lines earlier, so nothing user-supplied reaches the match.

Happy to change it if you would rather not carry the alert: replacing the regex with a `Number.parseFloat` round-trip check would satisfy the rule without one.

libs/@hashintel/petrinaut-core/package.json
```diff
@@ -70,13 +70,18 @@
       "types": "./dist/workers/simulation.d.ts",
       "import": "./dist/workers/simulation.js"
     },
+    "./webgpu": {
+      "types": "./dist/webgpu.d.d.ts",
```
TimDiekmann:
```suggestion
      "types": "./dist/webgpu.d.ts",
```

libs/@hashintel/petrinaut-core/package.json
```diff
@@ -70,13 +70,18 @@
...
+    },
     "./package.json": "./package.json"
   },
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
-    "build": "vite build",
+    "build": "vite build && yarn check:browser-safe-entries",
+    "check:browser-safe-entries": "node scripts/check-browser-safe-entries.mjs",
```
TimDiekmann: The `check:` field sounds like a lint to me, not a build step

libs/@hashintel/petrinaut-core/package.json
TimDiekmann: Fine to ignore if that's covered in the next step as well

libs/@hashintel/petrinaut-core/src/webgpu/emit-wgsl.ts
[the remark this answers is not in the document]
kube: False positive. The pattern `/^-?\d+(?:\.0)?$/u` is anchored at both ends with one non-nested repetition, and the optional group starts with a literal `.` that `\d` cannot match, so no input is ambiguous and a rejecting input backtracks linearly. The subject is WGSL the emitter produced on the line above, bounded by the emitted expression, not request input. Recommending the alert be dismissed as a false positive.

libs/@hashintel/petrinaut-core/src/webgpu/runner.ts
```diff
@@ -0,0 +1,666 @@
...
+  bytesPerRun: number;
+  runCount: number;
+}): string {
+  const gib = (bytes: number) => (bytes / 1024 ** 3).toFixed(2);
+  // Readback doubles it: the state lives on the device and again in a mappable
+  // buffer, and the mappable one is what runs out.
+  return `The GPU could not allocate memory for ${runCount} runs: ${bytesPerRun} bytes per run is ${gib(
+    stateBytes,
+  )} GiB of run state, and reading it back needs that much again in host-visible memory. Try fewer runs, or lower the token capacities that set the per-run size. (${message.trim()})`;
+}
```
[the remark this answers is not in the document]
kube: Fixed in this PR's stack, at the top layer where `describeAllocationFailure` now lives (`runner/device.ts`): the message and its doc comment attribute the failure to the device-side run-state buffer alone, since readback is only a compact per-run summary and the histogram.

libs/@hashintel/petrinaut-core/src/webgpu/gpu-experiment-handle.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  // Uncoloured places carry a plain count; a typed place's initial marking is an
+  // array of token records whose length is the count the shader needs.
+  const placeCounts = backend.profile.places.map((place) => {
+    const marking = config.initialMarking[place.id];
+    if (typeof marking === "number") {
+      return marking;
+    }
+    return Array.isArray(marking) ? marking.length : 0;
+  });
```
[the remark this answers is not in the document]
kube: Already fixed higher in the stack: `FE-1553: Seed typed places' initial token attributes on the GPU` adds `placeTokenOffsets`/`placeTokenWords` to the seeding path so token words are staged per run, and a later commit refuses a marking larger than the place's capacity.

libs/@hashintel/petrinaut-core/package.json
kube: Applied: every subpath export now points at its emitted declaration file, so `./webgpu` reads `"types": "./dist/webgpu.d.ts"` alongside the other twelve subpaths.

libs/@hashintel/petrinaut-core/package.json
kube: Agreed that it is not a lint: it walks `dist`, so it has to run after `vite build`. The separate `check:` script is gone and the build script calls `node scripts/check-browser-safe-entries.mjs` directly.

libs/@hashintel/petrinaut-core/src/webgpu/compile-net-shader.ts
```diff
@@ -0,0 +1,1154 @@
...
+            for (const { wordOffset, valueExpr } of tokenWrites) {
+              push(`          state[out + ${wordOffset}u] = ${valueExpr};`);
+            }
+            push(`        }`);
+          }
+        }
+        push(
+          `        pending[${index}u] = pending[${index}u] + ${arc.weight};`,
+        );
+      }
```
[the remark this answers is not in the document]
kube: Fixed in this PR's stack, at the top layer where the firing code lives (`compile-net-shader/transition-firing.ts`): the compaction sweep after a typed consume now runs to `counts[p] + max(0, pending[p])`, so tokens produced earlier in the frame shift down with the survivors and the end-of-frame fold leaves a contiguous live range. A test asserts the emitted bound for a net whose typed place is both produced into and consumed from in one frame.

libs/@hashintel/petrinaut-core/package.json
TimDiekmann: The description of `check-browser-safe-entries.mjs` sounds very much like a lint, not a build step. It's not required to run for building. It requires `build` to be run before, yes, but this is what we have `dependsOn` for. I appreciate that we currently don't have a lint section for that, and thus, calling it after `build` is sensible, but we should
- create an issue to properly call this in a lint workflow, not as part of a build workflow (if downstream packages need `petrinaut-core`, they don't have to call this script, they can rely on the output of `vite build`)
- Call it `lint:browser-safe-entries` rather than introducing a new term `check`.

kube: Sorry, Claude answered the first time.

It's actually a check on the output to verify that there is no server-side dependency in the output for browser.

We could do that before build, but before the tree-shaking (only done by build), it would result in a lot of false positives.

TimDiekmann: we can figure that out at some other point, thanks for creating the follow-ups!
(And if you want to prevent Claude from posting comments you can configure the auto-mode-classifier to detect that, it works quite well. I have that + at least one mentioning in CLAUDE.md + probably a few memory entries 😄 )

YannisZa (approved):
Cannot toggle on GPU in simulation experiment setup but this might be because there are no GPU resources allocated to the vercel deployment?

TimDiekmann (approved):

PR #9189, opened 2026-08-10
FE-1350: fix popover point positioning

alex-e-leon:

- Fixes popover positioning + tracking when position "point" is set
- Adds autofocus: "never" prop to inputs + buttons now that ark-ui + zag have implemented it (https://github.com/chakra-ui/zag/issues/3161#event-28334757079)
- Prevents close buttons + tooltips from being autofocused in dialogs
- Removes custom code in popover to handle focus out properly now that zag.js have fixed the issue
- Splits Tooltip + BaseTooltip to allow for full tooltip UI customization.

This PR:

The changes in this PR:


lunelson (approved):
Looks good, although I don't quite get the "center" and "outside" position options

alex-e-leon:
> Looks good, although I don't quite get the "center" and "outside" position options

Those 2 story examples are poorly labeled/named. The 2 prop options are actually:

```
// The preferred position of the popover - depending on the viewport, trigger and content another position may be chosen for better fit
position?: Position;

// Instead of positioning around the trigger, position from a specific point inside or outside the trigger where 0,0 is the top left of the trigger element
positionFromPoint?: { x: number; y: number };
```

And the center and outside examples are just test cases for positionFromPoint.

positionFromPoint was specifically added for attaching a popover to an element inside a canvas (which has no dom node to connect to), which I needed in graph-viz, and almost all the tables in hash-app currently are virtualized in canvas too.

alex-e-leon:
Also, in case I haven't said this before - it's worth noting that ladle examples are actually the UI tests for the DS rather than documentation, the goal being to render out specific edge cases for visual testing + regressions.

It also happens to have a lot of crossover with documenting the various features + props though, which is great while we don't have proper generated documentation yet - but it can mean that you'll sometimes see examples which are more for handling edge cases than real expected usage.

PR #9193, opened 2026-08-10
FE-1351: Use new banner component in ds-components

alex-e-leon:

Updates a few areas to use the new banner component.
@kube - there were other areas where you show errors that I think are not suited to the notice component, since the error UI is more customized (which is ok/good!), so I've been conservative with adding it in. Let me know what you think about the spots where I've added it. Happy to undo it/remove it.

Before after screenshots of the changes below:

Before:
import-error-og
After:
import-error-new

Before:
simulation-diagnostic-og
After:
simulation-diagnostic-new

Before:
scenario-compilation-og
After:
Screenshot 2026-08-10 at 16 59 08

This PR:

The changes in this PR:


kube (approved):
Looks better! Thanks!

PR #9197, opened 2026-08-11
SRE-927: Migrate same-repo action references to the $/ syntax

TimDiekmann:
## Purpose

Migrates same-repo action references to the `$/` self-repository syntax and wires up the centralized actionlint preflight (hashintel/.github#100–#102), which lints all workflow files on every PR.

## Related links

- [SRE-927](https://linear.app/hash/issue/SRE-927) _(internal)_

## What does this change?

- `uses: ./.github/actions/…` → `$/.github/actions/…` in 7 workflows and 2 composite actions (nested references)
- Fixes the actionlint/shellcheck findings the new check surfaces: quoted `$GITHUB_OUTPUT`/`$GITHUB_ENV`/`$GITHUB_STEP_SUMMARY` redirects, backticks in echoed help text replaced with double quotes, `$(…)` instead of legacy backticks in `canary-release.yml`, grouped summary redirects in `lint.yml`, a dead `TEST_TASKS` assignment removed in `test.yml`, and two unicode apostrophes replaced
- `warm-up-vercel.yml`: cron `*/3` → `*/5` — GitHub schedules at a 5-minute minimum, so the previous value was never honored
- The `pull_request` preflights are consolidated into one `preflight.yml`: Todo comments plus the new Lint job calling the centralized `preflight-actionlint.yml` (check name: `Preflight / Lint / Actions`). The `pull_request_target` file (Dependencies, Labeler, PR title) moves to `preflight-target.yml`; check names stay unchanged.

## Known issues

- `bench.yml` base-leg semantics change with `$/`: the base-bench steps previously resolved composite actions from the base checkout; `$/` always resolves them at the running commit, so both legs now use the PR's action versions. This matches the behavior the other repos converged on and removes a base/head skew, but base runs with changed actions are no longer byte-identical to their original runs.
- `release.yml` keeps its two `./` references for now — #9192 restructures that file and both changes would conflict. Migrate after #9192 merges.

## How to test this?

- Every job on this PR resolves its composite actions via `$/`
- The `Preflight / Lint / Actions` check lints this repo's workflows with only the three syntax-gap ignores (`$/`, `concurrency.queue`; SRE-941)


thehabbos007 (approved):
The big one 🪨

PR #9208, opened 2026-08-12
FE-1381: use ds button group in petrinaut

alex-e-leon:

Use button group in petrinaut. I've left many button collections with their custom flex styles instead of switching to button group which is as intended - button group does not need to be used all the time.
 
This PR:

The changes in this PR:


kube (approved):
:+1:

PR #9211, opened 2026-08-12
FE-850: use ds form field in petrinaut

alex-e-leon:

- Switch Petrinaut to use Form.Field, Form.Row and Form.Section components
- Adds inline layout to Form.Field
- Adds align prop to HelpTooltip

This PR:

The changes in this PR:


libs/@hashintel/petrinaut/src/ui/components/section.tsx
```diff
@@ -150,6 +145,10 @@ const fillHeightContentStyle = css({
   minHeight: "[0]",
 });
 
+const indentedContentStyle = css({
+  paddingLeft: "2",
+});
+
```
kube: That's the only thing I'm gonna ask you to remove for now.

Sections should be indented by default.
The Petrinaut UI depends on it in the Figma files, as you can see in:

https://www.figma.com/design/EuokCTrNYWhEMBQ7MmGrwJ/Petrinaut?node-id=3015-9366&t=AyqQwb33cfkgoCsS-4

Image

If you want to remove the indentation, you'll have to check with the designers and product.

kube (changes requested):
Just a change request on the indentation of `Form.Section` component.

I'm also not sure about naming everything `Form.Section`, where sometimes it's not actually not a Form. But we can come back to that later.

kube (approved):

kube (approved):

PR #9219, opened 2026-08-14
BE-760: Verify Kratos sessions in the Graph API

TimDiekmann:

Sessions issued by Ory Kratos now authenticate requests at the Graph REST API. The Kratos-backed provider is the first real credential path through the authentication pipeline introduced in #9218: it verifies the session against the whoami endpoint, resolves the identity's provisioned Graph actor, and only accepts existing user actors.

## 🔗 Related links

- [BE-760: Verify Kratos sessions in the Graph API](https://linear.app/hash/issue/BE-760/verify-kratos-sessions-in-the-graph-api) _(internal)_

## 🚫 Blocked by

- #9218 — this PR is stacked on its branch

## 🔍 What does this change?

- `hash-graph-authentication` gains `kratos::KratosSessionProvider`: recognizes the `X-Session-Token` header and the `ory_kratos_session` cookie, verifies the session against the Kratos whoami endpoint, requires an explicitly active session, reads the actor from the identity's `metadata_public.graph_actor_id`, and accepts only an existing user actor. Cookies are matched byte-level by name, so unrelated (including non-ASCII) cookies neither trigger verification nor mask the session cookie
- `actor::StorePoolActorResolver` resolves actor IDs against the principal store behind the `ResolveActor` trait, keeping the provider independent of the concrete store types
- the REST server wires the Kratos provider as its authentication provider; `HASH_KRATOS_PUBLIC_URL` is required to run the server. The flag is optional at parse time so `server --healthcheck` works without it — the same pattern the admin server uses for its external-service URLs
- authentication failures now return client-safe messages; identifiers remain in the server-side logs (addresses the review comment on #9218 — the affected error variants only become reachable with this provider). Rejections caused by broken actor provisioning (missing, unknown, or non-user actor behind a verified session) log at warn level so they are visible in production logs
- the whoami HTTP client never follows redirects (a redirect would forward the session token to the redirect target) and attaches the provider response body to all failure reports
- the compose graph service sets `HASH_KRATOS_PUBLIC_URL`
- new httpyac suite `kratos-session.http` runs the real wire flow against Kratos: provision an identity, log in, authenticate with the session, and verify the rejection paths

This PR:

The changes in this PR:

The changes in this PR:

  - the new `hash-graph-authentication` → `hash-graph-authorization`/`hash-graph-store` edges are carried by the generated `package.json` dependencies (`sync:turborepo`); no `turbo.json` changes are needed

## ⚠️ Known issues

- Deployments outside compose must set `HASH_KRATOS_PUBLIC_URL` (with network reach to Kratos' public API) before this rolls out — the server refuses to start without it. `server --healthcheck` parses and probes without the variable.
- Verified sessions are not cached yet — every request verifies against Kratos and the principal store. Tracked in BE-755 _(internal)_; lands before the frontend switches to direct Graph sessions.
- The httpyac suite covers the token flow; the cookie flow is covered by unit tests and was verified manually (the browser CSRF dance adds no verification value for the Graph side).
- Kratos identities accumulate across local test runs because `reset-database` only wipes the Graph database; the suite uses unique e-mail addresses per run. Cleaning Kratos/Hydra in the reset command is tracked in BE-762 _(internal)_.
- The session's `expires_at` is not tracked by the Graph; Kratos' own per-request whoami validation is authoritative.

## 🐾 Next steps

- service credentials for internal services (BE-714) _(internal)_
- Cloudflare Access JWTs as a provider in the same pipeline (BE-761) _(internal)_
- session verification cache (BE-755) _(internal)_

## 🛡 What tests cover this?

- `kratos.rs` wire tests against an in-test fake Kratos that only serves sessions when the expected credential is forwarded: active-flag handling, provisioning, actor-type checks, status-code mapping (including redirects and timeouts), cookie parsing next to non-ASCII cookies
- the middleware and orchestration tests from the parent PR run unchanged against the new provider surface
- `kratos-session.http` end-to-end against a real Kratos (runs in CI via the compose `dev` profile): session token authenticates and is subject to user policies, invalid token is rejected as an invalid session, a rejected session never falls back to the actor-ID header, unprovisioned and unknown-actor identities are rejected with the client-safe messages

## ❓ How to test this?

1. Start the external services (`docker compose --profile dev up` in `infra/compose`) and the Graph (`yarn start:graph` or `cargo run --bin hash-graph --all-features -- server`)
2. `cd tests/graph/http && yarn reset-database -o none && yarn httpyac send --all tests/kratos-session.http`
3. Full suite: `cd tests/graph/http && sh test.sh`

## 🐒 Demo

_N/A — backend authentication path; see the httpyac suite output for the wire-level flow._


apps/hash-graph/src/subcommand/server.rs
```diff
@@ -167,6 +167,45 @@ pub struct CompilerConfig {
...
+        let kratos_public_url = self.kratos_public_url.ok_or_else(|| {
+            Report::new(GraphError).attach(
+                "--kratos-public-url (HASH_KRATOS_PUBLIC_URL) is required when running the server",
+            )
+        })?;
+
+        Ok(KratosSessionConfig {
+            kratos_public_url,
+            http_timeout: Duration::from_secs(self.session_http_timeout_secs),
+        })
```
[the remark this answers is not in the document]
TimDiekmann: IMO it's fine to fail a server startup if you pass a `mailto` URL

libs/@local/graph/authentication/src/kratos.rs
```diff
@@ -0,0 +1,761 @@
...
+
+    None
+}
+
+/// Formats a provider response for report attachments, truncating long bodies.
+fn provider_response(status: reqwest::StatusCode, body: Result<String, reqwest::Error>) -> String {
+    let Ok(body) = body else {
+        return format!("provider response ({status}): <body unavailable>");
+    };
+    let mut snippet: String = body.chars().take(512).collect();
```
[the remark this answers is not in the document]
TimDiekmann: If someone DoS os on the Kratos URL, we have _**far**_ bigger issues than an OOM in the graph.

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -167,6 +167,50 @@ pub struct CompilerConfig {
...
+    // server refuses to start without it.
+    #[clap(long, env = "HASH_KRATOS_PUBLIC_URL")]
+    pub kratos_public_url: Option<Url>,
+
+    /// HTTP client timeout for session verification requests (in seconds).
+    #[clap(
+        long,
+        env = "HASH_GRAPH_SESSION_HTTP_TIMEOUT",
+        default_value_t = 10,
+        value_parser = clap::value_parser!(u64).range(1..)
```
indietyp, line 188: TIL this is a thing. Neat!

libs/@local/graph/authentication/src/kratos.rs
```diff
@@ -0,0 +1,761 @@
...
+    if let Some(token) = headers.get(SESSION_TOKEN_HEADER) {
+        return Some(
+            token
+                .to_str()
+                .map(SessionCredential::Token)
+                .change_context(AuthenticationError::MalformedCredential),
+        );
+    }
+
+    for cookie_header in headers.get_all(header::COOKIE) {
```
indietyp, line 52: Instead of hand-rolling this, would it maybe make sense to just use https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/index.html?

libs/@local/graph/authentication/src/kratos.rs
indietyp: non-blocking: Wouldn't something like this avoid the double-char scan while still keeping the 512-character limit?

```rs
let mut chars = body.chars();
let mut snippet: String = chars.by_ref().take(512).collect();
if !chars.as_str().is_empty() {
	snippet.push('\u{2026}');
}
```

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -167,6 +167,50 @@ pub struct CompilerConfig {
...
+    )]
+    pub session_http_timeout_secs: u64,
+}
+
+impl KratosSessionAuthConfig {
+    /// Converts the CLI configuration into the session provider configuration.
+    fn into_provider_config(self) -> Result<KratosSessionConfig, Report<GraphError>> {
```
indietyp, line 195: Non-blocking: correct me if I am wrong, but aren't we delaying configuration validation here? Everywhere else, we validate before we start anything; the temporal client, graph, etc. would have connected before this could fail validation and tear down the server.

libs/@local/graph/authentication/src/kratos.rs
TimDiekmann: I added `cookie`, which is agnostic.

libs/@local/graph/authentication/src/kratos.rs
TimDiekmann: Fixed

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: fixed

indietyp (approved):

PR #9228, opened 2026-08-16
FE-1270: Extract Python bindings for the Petrinaut CLI into @local/petrinaut-python

kube:

`apps/petrinaut-opt` carried its own Petrinaut CLI client, private to that service and covering two of the CLI's five protocol methods. This PR extracts the client into a package any Python code can use, and removes the service's knowledge of the CLI.

Bottom of stack #9269; FE-1467 (#9261) sits above.

Most of the diff is a move: `petrinaut_client.py` → `session.py`.

## 🔗 Related links

- [FE-1270](https://linear.app/hash/issue/FE-1270/create-python-bindings-to-petrinaut-core) *(internal)*: this PR

## 🔍 What does this change?

**New package `libs/@local/petrinaut-python`** (import name `petrinaut`), internal to the monorepo:

- `PetrinautSession` runs one model per CLI process and exposes `healthz()`, `metadata()`, `run()`, and `request()`.
- `OptimizationSession` drives a study over a manifest it treats as opaque JSON: `describe()`, `evaluate()`, `objective()`.
- Process handling is unchanged from the old client: scrubbed child environment, bounded reads with deadlines, process-group shutdown. Errors split into `PetrinautRunError` (recoverable) and `PetrinautClientError`/`PetrinautProtocolError` (session closes).

**`apps/petrinaut-opt` depends on the bindings and nothing else.** It no longer names the CLI in code, tests, docs, configuration, or the image:

- The bindings declare `@hashintel/petrinaut-cli` as a runtime dependency, so the Docker image provisions the bindings' closure (`turbo prune @local/petrinaut-python`) and links the closure's `petrinaut` executable into `/usr/local/bin`. The image ships built output only.
- `uv sync --no-editable` builds the bindings into the virtualenv, so the runner copies the virtualenv alone.
- The Node-options passthrough is renamed `PETRINAUT_CHILD_NODE_OPTIONS`; the image's own `NODE_OPTIONS` never reached the child and is removed.
- FE-1414 (#9230) adds the architecture check that enforces this: `optimizer → python-bindings` with no edge to `cli`.

**Review fixes.** `from_model_file` and `from_model` are staticmethods rather than classmethods: they always build a base `PetrinautSession`, and a classmethod read as a subclass constructor on `OptimizationSession` while returning the base class. `PetrinautRunError`'s docstring now also names the non-finite-objective case, and a doubled sentence in the README's development instructions is fixed.

Two follow-up commits address the remaining review threads:

- `describe_optimization()` is renamed to `describe()`: the old name repeated the class name, its siblings `evaluate()`/`objective()` carry no prefix, and `describe()` maps 1:1 to the protocol's `optimization.describe`. The previous name remains as an alias.
- Caller-input bugs no longer raise `PetrinautClientError`, whose contract says the session's process or transport is gone: `params` or a bootstrap payload that will not serialize to JSON raise `TypeError`, and a payload over the 8 MiB line cap raises `ValueError` — all before anything is written, leaving the session untouched (the request id is only consumed after a successful encode). Raised by Cursor and in review.

This PR:

- [x] modifies a workspace but **not** a publishable library. `@local/petrinaut-python` is private.

The changes in this PR:

- [x] require changes to docs which **are made** as part of this PR: the package README and both consumers' READMEs. In-app user-guide behaviour is unchanged.

The changes in this PR:

- [x] affected the execution graph, and the `turbo.json`'s have been updated to reflect this: the bindings' `test:unit` depends on the CLI's `build`, and `petrinaut-opt`'s codegen inputs include the bindings' `*.py` sources.

## 🛡 What tests cover this?

- Bindings: every test from the deleted `test_petrinaut_client.py`, plus end-to-end tests that spawn the built CLI for a run and an optimization study.
- `apps/petrinaut-opt`: 75 tests pass against the package.
- `Build petrinaut-opt (arm64)` in CI builds the rewired image.

## ❓ How to test this?

1. `turbo run test:unit --filter @local/petrinaut-python --filter @apps/petrinaut-opt`
2. `uv run python -c "from petrinaut import OptimizationSession, PetrinautSession"`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -140,7 +140,7 @@ jobs:
               push:       ["ecr", "ghcr"],
               dockerfile: "apps/petrinaut-opt/docker/Dockerfile",
               context:    ".",
-              paths:      ["apps/petrinaut-opt", "libs/@hashintel/petrinaut-cli", "libs/@hashintel/petrinaut-core"],
+              paths:      ["apps/petrinaut-opt", "libs/@hashintel/petrinaut-cli", "libs/@hashintel/petrinaut-core", "libs/@local/petrinaut-python"],
               ecs: [{ service: "petrinaut-opt", cluster: "h-stage-euc1-app", service_name: "h-stage-euc1-app-petrinaut-opt" }]
             },
             {
--- a/apps/petrinaut-opt/README.md
+++ b/apps/petrinaut-opt/README.md
@@ -105,7 +105,7 @@ bindings](../../libs/@local/petrinaut-python/README.md), created with the
 manifest as opaque JSON. This service starts the session, describes the study,
 evaluates one trial at a time, and closes the session.
 
-`describe_optimization()` returns the direction, the study settings, and the
+`describe()` returns the direction, the study settings, and the
 flat parameters that are not fixed, each one a descriptor such as:
 
 ```json
--- a/apps/petrinaut-opt/README.md
+++ b/apps/petrinaut-opt/README.md
@@ -1,9 +1,10 @@
 # Petrinaut optimization
 
 This service uses Optuna to optimize the flat, non-fixed parameters of one
-Petrinaut scenario. It keeps Yannis's Server-Sent Events API and study
-lifecycle, while delegating all Petrinaut-specific interpretation to
-`petrinaut-cli`.
+Petrinaut scenario. It owns the Server-Sent Events API and the study lifecycle,
+and delegates every Petrinaut-specific concern to the
+[`petrinaut` Python bindings](../../libs/@local/petrinaut-python/README.md).
+Which process the bindings run, and how, is theirs to decide.
 
 The Python service treats the optimization manifest as opaque JSON. It does not
 read Petrinaut models, scenario bindings, metrics, or the Petrinaut type system.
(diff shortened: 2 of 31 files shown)


YannisZa (dismissed):
Many thanks @kube! Does this PR imply that the `petrinaut_client.py` is now obsolete?

YannisZa:
I think its worth addressing the cursor issue here too. Sorry forgot to mention it earlier!

kube:
@YannisZa 

> Does this PR imply that the `petrinaut_client.py` is now obsolete?

Yes, `petrinaut_client.py` is now called `session.py` and is part of the Python bindings package.

`petrinaut-opt` does not have to deal with `petrinaut-cli` anymore, only consume the Python library.

YannisZa:
> Yes, `petrinaut_client.py` is now called `session.py` and is part of the Python bindings package.

This naming is a bit confusing.  `optimization.py` is also a session but for consuming optimisation manifests. My understanding is that these scripts are still clients/wrappers for `petrinaut-cli`, right?

To add to these, why do we need a `petrinaut` directory within `@local/petrinaut-python/src` ?

kube:
> To add to these, why do we need a `petrinaut` directory within `@local/petrinaut-python/src` ?

This seems to be a pretty established convention, and is what `uv init` does by default:

https://docs.astral.sh/uv/guides/projects/

kube:
> This naming is a bit confusing. optimization.py is also a session but for consuming optimisation manifests. My understanding is that these scripts are still clients/wrappers for petrinaut-cli, right?

Yes, I'm gonna rename that, and they are still binding/wrappers for `petrinaut-cli` calls, but the idea being that you do not have to care about it when using it.
It could in the future communicate with/use `petrinaut-core` differently.

vilkinsons (approved):

YannisZa (approved):

PR #9233, opened 2026-08-17
BE-714: Gate the actor-ID header behind a service credential

TimDiekmann:

Closes the last unauthenticated path into the Graph REST API: the `X-Authenticated-User-Actor-Id` header is only honored next to a service secret. Internal services (Node API, Temporal workers) authenticate themselves before naming the actor they act for, and the bootstrap routes require the same secret instead of being reachable by anyone on the network.

## 🔗 Related links

- [BE-714: Define Node API authentication to the Graph](https://linear.app/hash/issue/BE-714/define-node-api-authentication-to-the-graph) _(internal)_

## 🚫 Blocked by

- #9219 — this PR is stacked on its branch

## 🔍 What does this change?

- `AuthenticationProvider` returns `ControlFlow<Authentication>`: `Continue(())` means the request carries no credential the provider handles, `Break(Verified | Rejected)` is terminal. Chaining is the `?` operator, so a rejected credential structurally cannot fall through to another provider
- `Authentication::Verified` carries the authorization crate's `AuthenticatedActor` — Kratos verifies to a typed `Id`, service delegation passes the claimed actor through as `Uuid`
- `delegation::ServiceDelegationProvider` recognizes the pair of `X-HASH-Service-Secret` and the actor-ID header as one credential. The secret is compared in constant time. An actor-ID header without the secret is rejected, never ignored
- the bare actor-ID header fallback in `resolve_request_actor` is deleted; the new rejections log at warn level, since only internal services send these credentials and a mismatch is a deployment problem
- bootstrap routes (`/policies/seed`, `/actors/machine/identifier/system/{identifier}`) require the service secret and lose their unauthenticated allowlist entries
- the server requires a non-empty `HASH_GRAPH_SERVICE_SECRET` to start (parse-optional so `server --healthcheck` works without it), and an empty configured secret never matches — a set-but-empty variable cannot silently disable the credential
- bootstrap routes admit only requests carrying no credential at all: an actively rejected credential (malformed actor header, stale session cookie) stays rejected there too
- the Node API and both Temporal workers send the secret as an axios default header from the same variable; the config structs mask it (and the OpenAI key) in `Debug` output
- compose sets the secret on the graph service and all four senders; the httpyac suite injects it via a `.httpyac.config.js` hook, and `service-delegation.http` pins the rejection paths end-to-end
- drive-by: `hide_env_values` on the five sensitive clap env args (Postgres password ×2, OpenAI key, Mailchimp key, Sentry DSN) — `--help` no longer prints their values

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Deploy order matters: the senders roll out first (the current Graph ignores the extra header), then the Graph — it refuses to start without the secret. The secret must exist in Secrets Manager and the task definitions before rollout; that is the next internal-infra PR.
- One static secret shared by all senders; rotating it requires restarting every consumer in lockstep. A dual-secret rotation window is tracked in BE-763 _(internal)_.
- Delegated actors are passed through as claimed rather than validated against the principal store: the secret authenticates the sender, the Node API legitimately delegates to the public actor for anonymous requests, and a per-request store lookup would tax the hottest path before the BE-755 _(internal)_ cache exists.
- The admin server keeps its bare actor-ID header middleware (behind Cloudflare Access) until BE-761 _(internal)_ moves it onto the shared pipeline.
- The custom header is transitional by design and superseded by Hydra `client_credentials` (BE-728 _(internal)_).

## 🐾 Next steps

- provision the secret in staging/production (internal-infra)
- Cloudflare Access JWTs as a provider and admin-server unification (BE-761) _(internal)_
- OAuth2 `client_credentials` for service identity (BE-728) _(internal)_

## 🛡 What tests cover this?

- `delegation.rs` unit tests: the full header matrix — pair verifies, wrong secret, actor header without secret (the no-fallback case), secret without actor, malformed actor UUID
- `provider.rs` chain tests: falls through on `Continue`, stops at the first `Verified` or `Rejected`
- middleware tests: bootstrap routes with/without/with-wrong secret, bare actor headers, verified passthrough
- `service-delegation.http` end-to-end rejection paths with client-safe message asserts; the entire HTTP suite now runs through the delegation path via the injected secret

## ❓ How to test this?

1. Start the external services and the Graph (`yarn start:graph` or `cargo run --bin hash-graph --all-features -- server`)
2. `curl -i http://127.0.0.1:4000/actors/machine/identifier/system/h -H "X-HASH-Service-Secret: hash-svc-local-dev-secret"` → 200; without the header → 401
3. Any request with `X-Authenticated-User-Actor-Id` but without the secret → 401 `the actor-ID header requires the service credential`
4. Full suite: `cd tests/graph/http && sh test.sh`

## 🐒 Demo

_N/A — backend authentication path; see the httpyac suite output for the wire-level flow._


tests/graph/http/tests/service-delegation.http
```diff
@@ -0,0 +1,64 @@
...
+> {%
+  client.test("status", function() {
+    client.assert(response.status === 401, "Response status is not 401");
+  });
+%}
+
+### Malformed actor-ID header is rejected with the valid secret
+GET http://127.0.0.1:4000/actors/machine/identifier/h
+Content-Type: application/json
+X-HASH-Service-Secret: hash-svc-local-dev-secret
```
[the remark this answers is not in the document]
TimDiekmann: It's a test file

libs/@local/graph/api/src/rest/auth.rs
```diff
@@ -82,22 +83,29 @@ fn store_outcome(request: &mut Request, outcome: AuthenticationOutcome) {
...
     mut request: Request,
     next: Next,
 ) -> Response
 where
     P: AuthenticationProvider,
 {
+    let bootstrap = is_bootstrap_route(request.uri().path());
+    if bootstrap && !presents_service_secret(request.headers(), &service_secret) {
+        return rejection(&AuthenticationError::MissingServiceSecret).into_response();
+    }
```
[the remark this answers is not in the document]
TimDiekmann: Fixed in #9259

libs/@local/graph/api/src/rest/auth.rs
```diff
@@ -221,6 +234,14 @@ mod tests {
             .expect("the request should build")
     }
 
+    fn request_with_secret(uri: &str, secret: &str) -> Request<Body> {
+        Request::builder()
+            .uri(uri)
+            .header("X-HASH-Service-Secret", secret)
```
indietyp: Can we _please_ not introduce another `X-` prefixed header, like that time we introduced `X-Authorized-User-Id` (where I was against the `X-` prefix as well). These have been deprecated and discouraged since 2012: https://www.rfc-editor.org/rfc/rfc6648.html#section-3

> Creators of new parameters **SHOULD NOT** prefix their parameter names with `"X-"` or similar constructs.

This also strikes me as actually being more `Authorization` header-adjacent, which has the semantics that we want and benefits from automatic secret redaction as a well-known header:

https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.2

```
Authorization = credentials
credentials = auth-scheme [ 1*SP ( token68 / #auth-param ) ]
```

TimDiekmann: Changed to `Authorization: HASH-Service XXX`

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -238,7 +238,8 @@ pub struct ServerConfig {
     ///
     /// If not set, the entity and entity-type search endpoints cannot resolve a `semanticString`;
     /// callers must provide a precomputed `embedding` instead.
-    #[clap(long, env = "HASH_GRAPH_OPENAI_API_KEY")]
+    #[clap(long, env = "HASH_GRAPH_OPENAI_API_KEY", hide_env_values = true)]
+    #[debug("***")]
     pub openai_api_key: Option<String>,
```
indietyp, line 243: non-blocking: I like to use zeroize or https://docs.rs/secrecy/latest/secrecy/, they make it so that it's not a footgun to maybe expose them.

apps/hash-graph/src/subcommand/server.rs
```diff
@@ -521,6 +533,19 @@ where
         )?;
     }
 
+    // HTTP strips surrounding whitespace from header values, so the senders' secret arrives
+    // trimmed. Trimming here keeps both sides comparing the same bytes.
+    let service_secret = config
```
indietyp, line 538: Non-blocking: doesn't this need to move up so we don't partially start services before we finish validation? Or better yet: move that validation into the type system with a `Secret` type that has the `Debug` impl and validation. That way, we could re-use it consistently across all our keys.

indietyp (approved):
Some minor, non-blocking potential improvements; otherwise, LGTM.

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: I'd propose we globally introduce that, agreed that we want that or a derivation of it!

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: Actually already fixed in https://github.com/hashintel/hash/pull/9235 🚀

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: Tracked under [BE-791](https://linear.app/hash/issue/BE-791/carry-credentials-in-a-secret-type-instead-of-string)

PR #9234, opened 2026-08-18
BE-765: Tighten the authentication provider types and split the Kratos module

TimDiekmann:

The Graph's authentication providers carry types and a module layout from when there was only one provider. The next provider (Cloudflare Access, BE-761) touches the same places, so this lands the cleanup on its own to keep both changes reviewable apart.

No new capability, and no change to which credentials authenticate — with one deliberate exception noted under Known issues.

## 🔗 Related links

- [BE-765](https://linear.app/hash/issue/BE-765/tighten-the-authentication-provider-types-and-split-the-kratos-module) _(internal)_

## 🚫 Blocked by

- [ ] #9233

## 🔍 What does this change?

- `AuthenticationProvider` breaks with a `Result` rather than a dedicated two-variant enum. `Continue` still means "no credential of mine here"; both outcomes still stop the chain, so a rejected credential never falls through to another provider. Each provider's tail collapses from a match to a `map`.
- `resolve_user_actor` returns the `UserId` it already checked for, instead of the wider `ActorId`, and moves to `actor.rs` — it holds no Kratos knowledge.
- The Kratos module splits into `mod` and `session`. The HTTP status handling, previously written out inline, becomes `read_response_body` in `mod.rs`, which makes it testable without spawning a server. The session provider keeps its 401/403 check ahead of that call, so an expired session still reports as an invalid session rather than as a provider fault.
- Near-identical negative tests merge into `rstest` cases, and the shared test helpers move next to the types they exercise.
- `StaticAuthenticationProvider` sits behind a `test-utils` feature, so the fake provider can no longer be constructed in a production build.
- The client-message tests now iterate every error variant through a helper whose exhaustive match refuses to compile when a variant is added — including the check that no client message leaks an identifier.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

The identity metadata deserializes `graph_actor_id` as a `UserId`, which validates the RFC 4122 version and variant. A value that fails that check now aborts the whoami parse and is reported as an invalid provider response, where it previously parsed and failed later at actor resolution as an unauthenticated request. The two differ in status code and log level.

Actor IDs are generated as version 4 UUIDs and the nil UUID is explicitly still accepted, so this only reaches records provisioned by hand. Keeping the stricter type is deliberate: the same struct is what the Graph will write once identity provisioning moves over. If the error class turns out to matter operationally, it is a one-line type change.

## 🐾 Next steps

- BE-761 stacks on this branch and adds the Cloudflare Access provider, the Kratos identity lookup, and the unified admin API.
- [BE-764](https://linear.app/hash/issue/BE-764/model-the-public-actor-as-an-actor-not-as-the-nil-uuid) _(internal)_ covers what this PR deliberately leaves alone: the nil UUID as the public actor, and carrying the actor type through to the endpoints instead of collapsing it to a bare UUID.

## 🛡 What tests cover this?

Existing coverage is preserved; the suite grew rather than shrank.

- The status ladder is now covered directly in `kratos/mod.rs`, one case per rung, without a fake server.
- The session provider keeps its end-to-end cases, including a rate-limited provider — that one guards the ordering of the 401/403 check, since widening it to every client error would tell callers their session is bad while Kratos is merely throttling.
- The chain tests still assert that a rejected credential never falls through, and that a valid session beats a delegated actor.
- The new client-message tests cover every error variant and are wired so a future variant cannot silently skip them.

## ❓ How to test this?

2. `cargo clippy --package hash-graph-authentication --package hash-graph-api --all-features --all-targets`
3. To confirm the exhaustive guard: add a variant to `AuthenticationError` and observe that the crate stops compiling until it is handled and listed.


libs/@local/graph/authentication/src/request.rs
```diff
@@ -289,4 +287,104 @@ mod tests {
             "a request without credentials should fail authentication"
         );
     }
+
+    /// One instance of every error variant.
+    ///
+    /// A variant added to the enum stops compiling here until it is listed, so the tests below
+    /// cannot silently skip it.
+    fn every_error(identity_id: &str, actor_id: ActorEntityUuid) -> Vec<AuthenticationError> {
```
indietyp: Non-blocking, but I still wanted to share this neat trick! Another way to write this would be:

```
// make sure that we have all the errors
let errors: [_; core::mem::variant_count::<AuthenticationError>()] = [];
let discriminants: HashSet<_> = errors.iter().map(core::mem::discriminant).collect();
assert_eq!(errors.len(), discriminants.len())
```

(it also makes sure that there are no duplicates)

libs/@local/graph/authentication/src/request.rs
```diff
@@ -289,4 +287,104 @@ mod tests {
...
+    }
+
+    /// No client message may carry an identifier, whichever variant it comes from.
+    #[test]
+    fn no_client_message_carries_an_identifier() {
+        let identity_id = "d290f1ee-6c54-4b01-90e6-d701748f0851";
+        let actor_id = ActorEntityUuid::new(Uuid::new_v4());
+
+        for error in every_error(identity_id, actor_id) {
+            let message = error.client_message();
```
indietyp: Isn't this tautological, as this is already stated in the method itself? `client_message` is `&'static str`, so without leaking, this wouldn't be possible either way, no?

indietyp (dismissed):

libs/@local/graph/authentication/src/request.rs
TimDiekmann: Thanks, fixed in https://github.com/hashintel/hash/pull/9234/changes/4af4a9511aa252f7f001d2d012f532fb4113c85b

libs/@local/graph/authentication/src/request.rs
TimDiekmann: Yes, removed in https://github.com/hashintel/hash/pull/9234/changes/4af4a9511aa252f7f001d2d012f532fb4113c85b

indietyp (approved):

PR #9235, opened 2026-08-18
BE-761: Authenticate Cloudflare Access JWTs in the Graph API

TimDiekmann:

Adds Cloudflare Access as a third credential alongside the Kratos session and the service delegation, and replaces the admin API's own authentication scheme with the shared providers.

The two ports do **not** authenticate identically: the admin API deliberately runs without the session provider. Its handlers erase entities and delete users while authorizing nothing beyond "some actor", so operators reach it through Access and internal services through the shared secret — an end-user session is not a credential there.

This is what lets an operator reach the admin API through Cloudflare Access with their own identity, rather than through a header the server has to trust.

## 🔗 Related links

- [BE-761](https://linear.app/hash/issue/BE-761/authenticate-cloudflare-access-jwts-in-the-graph-api) _(internal)_

## 🚫 Blocked by

- [ ] #9234

## 🔍 What does this change?

- A Cloudflare Access provider that recognizes **only** `Cf-Access-Jwt-Assertion`. `Authorization: Bearer` stays reserved for the OAuth work in BE-728, so a Bearer token is not silently accepted here.
- The token's `email` claim resolves to an actor through the Kratos admin API. A credentials-identifier lookup also returns identities whose matching address is still **unverified**, so requiring `verified` is what stops someone registering an address they do not control and authenticating as its owner. The unsigned `Cf-Access-Authenticated-User-Email` header is never consulted.
- The admin API drops `AdminActorId` and its JWT extractors, and runs the operator chain — Access JWT, then service delegation. `/health` stays open; everything else authenticates, and a Kratos session does not.
- The bulk destructive endpoints (`/snapshot`, `/accounts`, `/data-types`, `/property-types`, `/entity-types`) move behind the `unsafe-dev-endpoints` cargo feature, replacing the runtime flag. The Dockerfile builds without it, so a released image cannot serve them at all; local and CI builds enable it.
- JWT validation moves from the API crate into the authentication crate, next to the providers that use it. It now requires `aud` and `iss` rather than only comparing them when present, and validates `nbf`. `jsonwebtoken` requires just `exp` by default, so a token omitting `aud` — or carrying it with the wrong type — verified before this; `aud` is what scopes a token to this application, since every application of an Access team is signed by the same keys.
- A JWKS entry that cannot be turned into a verification key is reported as a provider fault instead of an invalid token, so a malformed key set surfaces as an error rather than answering every request with 401 at debug level.
- A failed JWKS fetch is remembered for the cooldown. The cooldown previously bounded nothing until the first fetch had succeeded, so during an outage every request mounted its own fetch and queued behind the fetch mutex.
- The looked-up email no longer reaches error reports. It travels in a query string, and a `reqwest` error renders the URL it failed on, which put the address into error-level logs and Sentry on any Kratos fault.
- The three JWT timing flags reject `0`, which silently disabled the refresh cooldown or made every fetch time out.
- A non-HTTP `--kratos-admin-url` is rejected at startup instead of aborting the process inside the resolver.
- Callers that reach the admin API present the service credential: the httpyac suite, the Playwright helper, and the backend integration helper.

This PR:

The changes in this PR:

  - The [Graph Admin API] runbook describes the previous JWT-or-header scheme and the runtime flag. Its access instructions and the endpoint availability table need updating once this lands.

The changes in this PR:

## ⚠️ Known issues

The JWT validator accepts a single audience. An Access application with more than one AUD tag would need the configuration widened to a list — a small change, but not one this PR makes since nothing needs it yet.

The admin routes carry no OpenAPI annotations, so they stay absent from the generated spec as before. Worth folding into BE-757 rather than here.

Removing the `Authorization: Bearer` fallback is a behaviour change for anyone who reached the admin API that way. Nothing in the repo does, but a hand-rolled client might.

A Kratos session no longer authenticates on the admin API. Anything that reached it with a session must present the service credential instead; the callers in this repo already do.

The Access path does not read the identity's Kratos `state`, while the session path requires an active session. Deactivating an identity would therefore not revoke Access-based entry. Latent today, since users are deleted rather than deactivated — tracked as a follow-up rather than fixed here, because it needs a decision on whether `state` gates authentication or only the session.

## 🐾 Next steps

- [BE-764](https://linear.app/hash/issue/BE-764/model-the-public-actor-as-an-actor-not-as-the-nil-uuid) _(internal)_ — model the public actor properly and drop the provider-side actor lookup, which the policy layer already performs.
- BE-763 covers rotating the service secret.

## 🛡 What tests cover this?

- Unit tests for the Access provider mint tokens against a fake JWKS: algorithm confusion (an HS256 token signed with an attacker-chosen secret), a signature transplanted from another valid token, expiry, wrong audience, wrong issuer, missing and unknown key IDs, and a missing email claim.
- The JWKS cache is covered by counting fetches at the fake endpoint, which pins the refresh cooldown that stops a crafted `kid` from driving one fetch per request.
- Contract tests in `tests/kratos.rs` run the identity lookup against the **live Kratos** of the compose stack, covering a verified address, an unverified one, an identity without a Graph actor, an unknown address, and case-insensitive matching. These exist because the unit tests serve their responses from a hand-written fake, which only holds as long as the fake matches Kratos.
- The httpyac suite covers the admin authentication boundary end to end: a **valid** Kratos session is rejected there, an actor header without the secret is rejected, a wrong secret is rejected, and `/health` stays open.
- Required-claim coverage asserts that an omitted or wrongly-typed `aud`/`iss` fails, not only a wrong one — the mutation that removes the requirement turns those cases red.

## ❓ How to test this?

2. With the compose stack up: `yarn workspace @rust/hash-graph-authentication test:integration` for the Kratos contract tests
3. From `tests/graph/http/`: `sh test.sh`
4. Confirm the feature gate: build without `--all-features` and check that `DELETE /data-types` on the admin port answers 404 while `POST /users/delete` still answers 401


apps/hash-graph/src/subcommand/admin_server.rs
```diff
@@ -102,7 +104,8 @@ pub struct JwtConfig {
     #[clap(
         long = "jwt-jwks-cache-ttl",
         env = "HASH_GRAPH_JWT_JWKS_CACHE_TTL",
-        default_value_t = 600
+        default_value_t = 600,
+        value_parser = clap::value_parser!(u64).range(1..)
     )]
     pub jwks_cache_ttl_secs: u64,
```
indietyp: I don't mind this either way, but wouldn't `NonZero<u64>` be more idiomatic here? (I guess `--help` would be worse)

apps/hash-graph/src/subcommand/admin_server.rs
```diff
@@ -187,28 +193,71 @@ pub struct AdminConfig {
...
+                .ok_or_else(|| {
+                    Report::new(GraphError).attach(
+                        "--kratos-admin-url (HASH_KRATOS_ADMIN_URL) is required for Cloudflare \
+                         Access authentication",
+                    )
+                })?;
+            // The resolver extends this URL with a path, which a cannot-be-a-base URL such as
+            // `mailto:` does not support. Rejecting it here turns a process abort into a startup
+            // error.
+            if !matches!(kratos_admin_url.scheme(), "http" | "https") {
```
indietyp, line 228: Does it make sense to ensure it's only HTTPS in prod, never HTTP? Especially with JWKS, where HTTPS is required per the spec.

apps/hash-graph/src/subcommand/admin_server.rs
```diff
@@ -187,28 +193,71 @@ pub struct AdminConfig {
...
+                return Err(Report::new(GraphError).attach(
+                    "--kratos-admin-url (HASH_KRATOS_ADMIN_URL) must be an http or https URL",
+                ));
+            }
+            Ok(Some(CloudflareAccessConfig {
+                jwt: JwtValidatorConfig {
+                    jwks_url,
```
indietyp, line 235: I don't think we validate the scheme here for jwks, do we?

libs/@local/graph/authentication/src/jwt.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    /// Returns the cached JWKS or fetches a fresh copy.
+    ///
+    /// When `force_refresh` is `true`, fetches unless the cache was refreshed within the cooldown
+    /// period (to prevent denial-of-service via crafted `kid` values).
+    ///
+    /// The fetch mutex ensures only one outbound JWKS request is in-flight at a time. Concurrent
+    /// callers wait for the single fetch to complete and then re-check the cache.
+    async fn get_jwks(&self, force_refresh: bool) -> Result<JwkSet, Report<JwtError>> {
```
indietyp: Could we make this an enum? I was very confused about what `get_jwks(false)` was supposed to mean when reading the code above.

libs/@local/graph/authentication/src/cloudflare.rs
```diff
@@ -0,0 +1,607 @@
...
+                JwtError::UnusableKey { .. } => {
+                    report.change_context(AuthenticationError::InvalidProviderResponse)
+                }
+                JwtError::Validation | JwtError::MissingKeyId | JwtError::UnknownKeyId { .. } => {
+                    report.change_context(AuthenticationError::InvalidAccessToken)
+                }
+            }
+        })?;
+
+        let Some(email) = claims.email else {
```
indietyp, line 87: Correct me if I am wrong, but isn't there a potential authorisation problem here? By dropping the `sub` claim here and using only the email, we open a potential attack vector that exploits a race condition.

```
t0  Bob owns b@example.com
    Cloudflare issues token T:
       sub   = bob-generation
       email = b@example.com
       exp   = t0 + 5 minutes

t1  Bob is fired
    Bob's Kratos identity is deleted
    Token T is not revoked in Cloudflare

t2  Bernd is provisioned
    Bernd now owns b@example.com
    Bernd's Kratos identity points to Bernd's Graph actor

t3  Bob replays the unchanged token T
```

indietyp (changes requested):

apps/hash-graph/src/subcommand/admin_server.rs
TimDiekmann: No, kratos runs in the same VPC, so it would actually break.

apps/hash-graph/src/subcommand/admin_server.rs
TimDiekmann: No, we would simply fail the fetch, which is IMO a fine behavior.
The check like for Kratos does not apply here because that exists as the resolver extends the URL with a path (a cannot-be-a-base URL like `mailto:` would abort there).

libs/@local/graph/authentication/src/cloudflare.rs
TimDiekmann: This implies, that an email is used by another person. Who does this? I don't see a strong attack vector. To _properly_ solve this, we'd need to map actor -> sub. We can use `sub`, yes, but I would defer that to another ticket (this is mainly moving the already existing JWT auth (`libs/@local/graph/api/src/rest/jwt.rs`) into the correct form).
Note, that in prod we require our own jwks here and we don't re-use emails.

apps/hash-graph/src/subcommand/admin_server.rs
indietyp: But then we validate the URL for everything else; why do we fail late here but fail early everywhere else? (e.g., Kratos just above). I still don't fully understand how the JWKS authority differs in that regard.

libs/@local/graph/authentication/src/cloudflare.rs
indietyp: The issue isn't that we expect someone to do this intentionally; it is that we rely on operator convention to avoid it. We discard a stable subject in favour of an email, a mutable property that resolves to whichever user happens to have access to it. The example above showed the issue from a malicious perspective; another valid one is when someone accidentally assigns the wrong email and then needs to rename it. Same "exploit", different lens. I don't think authorisation should depend on human judgement and avoiding mistakes.

Additionally, using our own JWKS doesn't change the problem. Signature validation establishes that a trusted issuer made the token’s `sub -> email` assertion and that the assertion has not been modified. We establish that the mapping of `sub -> email` was true at `t0`; it makes no promises during request time `t1`. That's ostensibly our job.

Is the replay window narrow? Yes. Is it real? Yes.

From OIDC, directly from paragraph 5.7:
> The sub (subject) and iss (issuer) Claims from the ID Token, used together, are the only Claims that an RP can rely upon as a stable identifier for the End-User, since the sub Claim MUST be locally unique and never reassigned within the Issuer for a particular End-User, as described in [Section 2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken). Therefore, the only guaranteed unique identifier for a given End-User is the combination of the iss Claim and the sub Claim.

TimDiekmann: The fix isn't a straight email -> sub replacement, though, as we don't know `sub` before the first login. I didn't say we shouldn't do that, it's just not a simple drive-by (and as mentioned above, this was largely porting over the old implementation). I added BE-800.

libs/@local/graph/authentication/src/cloudflare.rs
```diff
@@ -0,0 +1,607 @@
...
+            return Err(Report::new(AuthenticationError::InvalidAccessToken)
+                .attach("the token carries no email claim"));
+        };
+
+        self.email_resolver.resolve_email_actor(&email).await
```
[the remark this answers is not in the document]
TimDiekmann: BE-800

apps/hash-graph/src/subcommand/admin_server.rs
TimDiekmann: I added an enforced `https` check in https://github.com/hashintel/hash/pull/9235/changes/4fce11fc8041086731448582be6a9c3eb496c368

libs/@local/graph/authentication/src/jwt.rs
TimDiekmann: done: https://github.com/hashintel/hash/pull/9235/changes/4fce11fc8041086731448582be6a9c3eb496c368

apps/hash-graph/src/subcommand/admin_server.rs
TimDiekmann: I don't know why the `--help` text would change, no reason to not use `NonZero<u64>` I guess: https://github.com/hashintel/hash/pull/9235/changes/4fce11fc8041086731448582be6a9c3eb496c368

indietyp (approved):

PR #9238, opened 2026-08-18
FE-1202: add filter component to ds

alex-e-leon:

- Adds Filter component do DS
- Also fixes panda css selector specificity bug across multiple components where panda applies specificity based on property name over selector specificity causing some css to not be applied.
- Adds missing type-fest dependency

This PR:

The changes in this PR:


lunelson:
Some of the conditions are self-contained, like "power of 10", and should, as with booleans, not render an input:

CleanShot 2026-08-18 at 18 01 22@2x

alex-e-leon:
> Some of the conditions are self-contained, like "power of 10", and should, as with booleans, not render an input:
> 
> CleanShot 2026-08-18 at 18 01 22@2x

Good shout, I'll rename it - the intention of this example was to test that "step: 10" worked, and I came up with a badly named example.

alex-e-leon:
> Some of the conditions are self-contained, like "power of 10", and should, as with booleans, not render an input:
> 
> CleanShot 2026-08-18 at 18 01 22@2x

@lunelson I've renamed those examples so they should be more clear what they are testing. Let me know if they make more sense now.

lunelson (approved):
On its own terms, it seems good; TBH I'm not sure how they work though. Personally I would like to see them proven out in a real filtering scenario if possible

PR #9251, opened 2026-08-19
FE-1454: add sort menu to ds

alex-e-leon:

Add SortMenu component to DS

This PR:

The changes in this PR:


lunelson (dismissed):
Looks nice! Small bug: the direction of the icon in the trigger vs the direction in the menu, are inverted on this demo:

Image

alex-e-leon:
> Looks nice! Small bug: the direction of the icon in the trigger vs the direction in the menu, are inverted on this demo:
> 
> Image

The logic was actually correct, the issue is that the ascending A-Z sort icon points down (because z is at the bottom) but an ascending arrow points up - can see how if not treated carefully though this could definitely trip users up. I've switched the example to remove the confusion though

lunelson (approved):

PR #9256, opened 2026-08-19
BE-770: Use the Kratos user-ID lookup on the admin delete-user path

TimDiekmann:

Deleting a user by email resolved the address through a JSONB scan over all entity editions, extracting every `email` property as an array. One entity carrying a scalar `email` — an org invitation, for example — could fail the whole query with `cannot extract elements from a scalar`, because Postgres does not guarantee predicate evaluation order even though the query joined `user_actor`.

Kratos already owns the email-to-identity mapping, and the Graph provisions each identity with its actor (`graph_actor_id` in `metadata_public`). The delete-user endpoint now resolves the email there, so the shape of stored `email` properties no longer matters for deletion.

## 🔗 Related links

- [BE-770](https://linear.app/hash/issue/BE-770/use-the-kratos-user-id-lookup-on-the-admin-delete-user-path) _(internal)_

## 🚫 Blocked by

- [ ] #9235

## 🔍 What does this change?

- `KratosIdentityProvider` gains `find_user_by_email`: the identity is looked up by credentials identifier and the actor is read from its `metadata_public.graph_actor_id`. Unlike the authentication resolver, an **unverified** address resolves — deletion must reach accounts that never completed verification.
- An address held by several identities fails with an explicit error instead of picking one — that state means an account has claimed an address it may not own, so the operator falls back to deletion by user ID. An identity without a provisioned actor is reported as such rather than as a missing user.
- The resolved Kratos identity ID travels into `delete_user` instead of being read back from the graph. Deletion order stays graph-then-Kratos, so the partial state a non-fatal Kratos failure leaves behind — entities purged, identity still alive — remains findable and deletable by email.
- `get_user_id_by_email` and its JSONB scan are removed from `AccountStore`, the Postgres store, and the type-fetcher passthrough.
- `MetadataPublic` is exported from the authentication crate, so the metadata contract keeps a single Rust definition.

This PR:

The changes in this PR:

  - The [Graph Admin API] runbook's delete-user section: deleting by email now requires the identity to exist in Kratos and carry a provisioned actor, and an ambiguous address is rejected. Same page #9235 already flags for updating.

The changes in this PR:

## ⚠️ Known issues

- A user whose Kratos identity is already gone while their graph entities remain is deletable by user ID only; the email lookup no longer sees them.

## 🐾 Next steps

- Update the [Graph Admin API] runbook once this lands.

## 🛡 What tests cover this?

- New unit tests for the Kratos lookup against a fake admin API: resolution of a provisioned identity, unknown email, unprovisioned identity, and an address held by several identities.
- A new integration test creates an org invitation — whose `email` property is a single string — through the real invite resolver and deletes a user by email against the admin server; this failed with `cannot extract elements from a scalar` before.
- The existing integration tests that delete users by email exercise the new path.

## ❓ How to test this?

The new integration test in `user.test.ts` automates the scenario: an org invitation exists, and deleting a user by email through the admin server succeeds.

[Graph Admin API]: https://www.notion.so/hashintel/Graph-Admin-API-31a3c81fe02480f792c9d7bedfdc49db

### DIFF
--- a/libs/@local/graph/api/src/identity_provider.rs
+++ b/libs/@local/graph/api/src/identity_provider.rs
@@ -21,6 +21,18 @@ pub(crate) enum EmailLookupError {
 struct AdminIdentity {
     id: String,
     metadata_public: Option<MetadataPublic>,
+    #[serde(default)]
+    traits: IdentityTraits,
+}
+
+/// The identity traits the Graph reads.
+///
+/// An identity created outside the registration flow may carry no traits at all, so the addresses
+/// default to none rather than failing the read.
+#[derive(Default, Deserialize)]
+struct IdentityTraits {
+    #[serde(default)]
+    emails: Vec<String>,
 }
 
 /// A Kratos identity resolved to its provisioned Graph actor.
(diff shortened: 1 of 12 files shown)


CiaranMn (dismissed):

CiaranMn (dismissed):

libs/@local/graph/api/src/identity_provider.rs
```diff
@@ -1,19 +1,164 @@
...
+    pub(crate) async fn find_user_by_email(
+        &self,
+        email: &str,
+    ) -> Result<Option<ResolvedUser>, Report<EmailLookupError>> {
+        let mut url = self.admin_url.clone();
+        url.path_segments_mut()
+            .expect("admin URL is not a cannot-be-a-base URL")
+            .extend(["admin", "identities"]);
+        url.query_pairs_mut()
+            .append_pair("credentials_identifier", email);
```
indietyp: This needs to check whether the value is non-empty before proceeding; Kratos treats an empty query as if the filter were absent. See: https://github.com/ory/kratos/blob/9d7085948039ffb8960160d4979f71527b5cf4d5/identity/handler.go#L212-L216

libs/@local/graph/api/src/identity_provider.rs
```diff
@@ -36,15 +181,207 @@ impl IdentityProvider for KratosIdentityProvider {
...
+            .get(url)
+            .send()
+            .await
+            .change_context(IdentityProviderError::LookupFailed)?;
+
+        // An identity deleted between resolution and this read holds no addresses.
+        let status = response.status();
+        if status == reqwest::StatusCode::NOT_FOUND {
+            tracing::info!(%identity_id, "Kratos identity no longer exists");
+            return Ok(Vec::new());
```
indietyp: There's a subtle regression here: if nothing is found, subscriptions_deleted becomes true (as we return Ok(Vec::new) for emails), even though no cleanup ever ran.  Meaning that Mailchimp is never informed. This means we can end up in a state where Mailchimp information is unlinked from any user information HASH retains, and a user cannot be unsubscribed except through manual intervention (which would be hard to track down - a user is no longer available in the system, yet they continue to receive notifications), since we have no way to delete emails from it once they're deleted from Kratos.

indietyp (approved):

PR #9259, opened 2026-08-20
BE-764: Model the public actor as an anonymous caller instead of the nil UUID

TimDiekmann:

Requests without credentials now resolve to an anonymous caller instead of impersonating the nil UUID. Whether a route serves anonymous callers is expressed in the types: authentication chains resolve to `ActorId` (actor required) or `Option<ActorId>` (anonymous allowed), and the stores take `Option<ActorId>` only on the read paths that serve public data.

## 🔗 Related links

- [BE-764](https://linear.app/hash/issue/BE-764) _(internal)_

## 🚫 Blocked by

- [ ] #9256

## 🔍 What does this change?

- `AuthenticationProvider` is generic over a sealed `Caller` (`ActorId` or `Option<ActorId>`); a chain without a recognized credential resolves through `Caller::anonymous()`, so the REST chain serves anonymous callers while the admin chain rejects them
- `ServiceDelegationProvider` treats the service secret as the credential: the delegated actor is resolved against the principal store, the nil UUID reads as "acting for nobody", and a verified secret without a delegated actor fails as the new `MissingDelegatedActor`
- An actor-ID header without the secret no longer yields 401 — the request resolves as anonymous, since the header alone was never a credential
- Admin router split: `/entities/delete` and `/users/delete` authenticate through the chain and require an actor, while the `unsafe-dev-endpoints` group is gated by the service secret alone
- Store boundary: reads that serve public data take `Option<ActorId>`; everything else (writes, search, validate/diff/cluster, policy CRUD) requires an actor
- Benches, seeds, the AI worker, and tests act as the system machine — fetched through the secret-gated bootstrap route — instead of the nil UUID
- Removes the unused `isEntityPublic` GraphQL field, resolver, and query

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- `authentication_middleware` takes the service secret only for the bootstrap routes, so the parameter is unused on the admin chain. Kept for a uniform middleware signature — splitting it buys nothing until the bootstrap routes move onto the admin API (BE-771).

## 🐾 Next steps

- BE-771: move the bootstrap routes onto the admin API

## 🛡 What tests cover this?

- Unit tests on the `Caller`/provider chain, the delegation flow, and the middleware (bootstrap admission, secret gate, extractor behaviour)
- Postgres integration tests for public-actor reads (permit/deny pair, empty permission maps in subgraphs)
- The httpyac suite pins the delegation rejection paths and an anonymous entity query end to end

## ❓ How to test this?

2. From `tests/graph/http/`: `yarn reset-database -o none && yarn httpyac send --all tests/service-delegation.http -o none`

### DIFF
--- a/libs/@local/graph/api/src/rest/entity_type.rs
+++ b/libs/@local/graph/api/src/rest/entity_type.rs
@@ -8,7 +8,6 @@ use axum::{
     routing::{post, put},
 };
 use error_stack::{Report, ResultExt as _};
-use hash_graph_authorization::policies::principal::actor::AuthenticatedActor;
 use hash_graph_embeddings::OpenAiEmbeddingClient;
 use hash_graph_postgres_store::{
     ontology::patch_id_and_parse,
@@ -167,7 +166,7 @@ where
         .acquire(temporal_client.0)
         .await
         .map_err(report_to_response)?
-        .has_permission_for_entity_types(AuthenticatedActor::from(actor), params)
+        .has_permission_for_entity_types(actor, params)
         .await
         .map(Json)
         .map_err(report_to_response)
@@ -497,7 +496,7 @@ where
     )
 )]
 async fn query_entity_types<S>(
-    AuthenticatedActorId(actor_id): AuthenticatedActorId,
+    actor_id: Option<AuthenticatedActorId>,
     store_pool: Extension<Arc<S>>,
     temporal_client: Extension<Option<Arc<TemporalClient>>>,
     Extension(api_config): Extension<ApiConfig>,
@@ -507,6 +506,7 @@ async fn query_entity_types<S>(
 where
     S: StorePool + Send + Sync,
 {
+    let actor_id = actor_id.map(|AuthenticatedActorId(actor_id)| actor_id);
     if let Some(query_logger) = &mut query_logger {
         query_logger.capture(actor_id, OpenApiQuery::GetEntityTypes(&request));
     }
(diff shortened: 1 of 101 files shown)


libs/@local/graph/authentication/src/delegation.rs
```diff
@@ -29,69 +31,115 @@ pub fn presents_service_secret(headers: &HeaderMap, secret: &str) -> bool {
...
+
+        let actor_uuid = match actor_id_from_header(headers) {
+            Ok(actor_uuid) => actor_uuid,
+            Err(error) => return ControlFlow::Break(Err(Report::new(error))),
         };
 
-        core::future::ready(decision)
+        if Uuid::from(actor_uuid).is_nil() {
+            return ControlFlow::Break(Ok(None));
+        }
```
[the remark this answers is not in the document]
TimDiekmann: That's on purpose, we currently don't require an anonymous actor to do a search.

libs/@local/graph/api/src/rest/entity_type.rs
thehabbos007, line 615: Should this also have the `actor_id: Option<AuthenticatedActorId> ` treatment like the rest? and similarly for the rest down to the store? 

I say this because `searchEntityTypes` which uses this endpoint is "any user" in gql, so anonymous callers could call this right?

thehabbos007 (commented):
Looks good to me, sorry for the long wait time for the review. I have one comment, local testing turned out well.

libs/@local/graph/api/src/rest/entity_type.rs
TimDiekmann: I'd argue not to make this public, no. We currently don't have a way for logged-out users to search entity types by semantic strings (you need to be logged in), and lifting it is very simple.

thehabbos007: Very fair not to make search public, i agree. But i think either it is wrongly documented in the server GQL setup, or we are missing an auth guard for it, but it isn't blocking merging this in

thehabbos007 (dismissed):

indietyp (dismissed):

indietyp (approved):

PR #9262, opened 2026-08-21
FE-1468: Generate pydantic models from the CLI's protocol schemas

kube:

The bindings returned `dict[str, Any]`, and `apps/petrinaut-opt` read the study description field by field with nothing checking it; a protocol change first showed up as a failing study. This PR makes the protocol's response shapes one definition every language consumes: Zod schemas in `petrinaut-core`, published by the CLI as JSON Schema, generated into pydantic models for Python.

The repository already generates types across this boundary in the other direction (pydantic → `openapi.json` → `@local/petrinaut-optimizer-client`). This is that pipeline reversed, with the same regenerate-and-diff guard.

Sits on FE-1467 (#9261) in stack #9269; FE-1412 (#9229) is above.

## 🔗 Related links

- [FE-1468](https://linear.app/hash/issue/FE-1468/python-generate-pydantic-models-from-the-clis-protocol-schemas) *(internal)*: this PR

## 🔍 What does this change?

- **`petrinaut-core`**: the response shapes become Zod schemas (`petrinautOptimizationDescribeParameterSchema`, `…DescribeResultSchema`, `…ReplicateSchema`, `…EvaluateResultSchema`). The existing types are now `z.infer` of them and are unchanged; 880 core and 39 CLI tests pass untouched. Patch changeset.
- **`petrinaut-cli`**: a `codegen` task writes the checked-in `schemas/optimization-protocol.schema.json`, piped through oxfmt so formatting and regeneration agree. CI regenerates and diffs it.
- **`petrinaut-python`**: `codegen` runs `datamodel-codegen` over that document into `src/petrinaut/models.py`. `describe()` returns `OptimizationDescribeResult` and `evaluate()` returns `OptimizationEvaluateResult`, validated in JSON-strict mode: a string where a number belongs is a contract bug and raises `PetrinautProtocolError`, while enums still parse from their wire strings. `py.typed` is added — without it no checker trusted the package's annotations. pydantic becomes the package's first runtime dependency.
- **`petrinaut-opt`**: `_parse_description` keeps only the semantic rules a schema cannot express (bound ordering, log-scale domains, duplicates, study limits); the shape validation moves to the models. pydantic is declared directly instead of arriving through FastAPI.
- **Both Python packages move to `typeCheckingMode = "strict"`.** After the typed rewrite, strict reported 7 errors (41 on the old code); all fixed.

**Review fixes.** The describe response schema declares what the domain already guarantees: an int parameter's `minimum`/`maximum`/`step` are integers, and `study.seedsPerTrial` is bounded to the CLI's 1–100 range, so the generated models carry those constraints. `PetrinautSession`'s docstring states the validation seam: model-protocol methods return the CLI's JSON verbatim as dicts, optimization methods return schema-validated models. The README example reads `full.replicates` by attribute, matching the model. The Docker image builds only the bindings' dependency closure (`--filter '@local/petrinaut-python^...'`): the bindings themselves have nothing to build in the image, and their codegen needs uv, which the Node-only builder stage lacks — this is what fixed the stack's Docker build.

This PR:

- [x] modifies an existing publishable library, and I have added a changeset. Patch: additive schema exports, derived types unchanged.

The changes in this PR:

- [x] require changes to docs which **are made** as part of this PR: the bindings README documents the models and the codegen guard.

The changes in this PR:

- [x] affected the execution graph, and the `turbo.json`'s have been updated to reflect this: `petrinaut-cli#codegen` (outputs `schemas/**`) and `petrinaut-python#codegen` (outputs `models.py`, depends on the CLI's codegen).

## 🛡 What tests cover this?

- Core 880, CLI 39: unchanged, proving the type derivation is behaviour-compatible.
- Bindings 24, including: a bool, null, or string objective is a schema violation that closes the session; an `Infinity` that Python's lax JSON parser admits stays a recoverable `PetrinautRunError`. The two e2e tests drive the built CLI through the typed methods.
- Service 75. basedpyright strict and ruff pass in both packages. Regenerating twice produces no diff.

## ❓ How to test this?

1. `turbo run codegen --filter @local/petrinaut-python` then `git diff --exit-code`
2. Change a response schema in `petrinaut-core/src/optimization.ts` and rerun step 1: the diff is what Python consumers see.


YannisZa:
@kube looks great too - thanks! I will approve once we fix the cursor-raised issue.

YannisZa (approved):

PR #9281, opened 2026-08-21
BE-773: Drop the nil-UUID sentinel from the created-by filter

TimDiekmann:

The property-protection filter compiled a request without an actor to a nil-UUID parameter, the last place in the query layer that read the nil UUID as the public actor. The comparison now says what it means: nothing equals an actor that is not there.

## 🔗 Related links

- [BE-773](https://linear.app/hash/issue/BE-773) _(internal)_

## 🚫 Blocked by

-  #9279

## 🔍 What does this change?

- `FilterExpression::from_cell_filter_expression` returns [`None`] for the actor expression when the request carries no actor, instead of substituting the nil UUID
- `Filter::for_property_filter` decides per comparison what a missing actor means: an equality matches no record, an inequality matches every record, and an `IN` matches no record
- Covers the actor-present branch, which had no test

Behaviour is unchanged: an empty `All` transpiles to `TRUE` and an empty `Any` to `FALSE`, which is what `uuid != nil` and `uuid = nil` evaluated to for every real row.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The remaining `Uuid::nil()` uses in `entity/table.rs` are cursor and fixture values in tests, not actor sentinels, so they stay.

## 🐾 Next steps

- BE-758: rate limiting, ahead of the versioned public API surface

## 🛡 What tests cover this?

- The property-protection transform cases, whose fixtures now pin the anonymous case as a tautology rather than a nil comparison
- A new case pinning that a request with an actor compares against that actor's UUID

## ❓ How to test this?

1. `cargo nextest run --package hash-graph-store --all-features`
2. `cargo clippy --all-features --package hash-graph-store --all-targets`

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/query/compile/tests.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/query/compile/tests.rs
@@ -2118,9 +2118,8 @@ mod property_masking {
             &compiler,
             r#"
             SELECT ("entity_editions_0_1_0"."properties" - (CASE WHEN
-                ("entity_temporal_metadata_0_0_0"."entity_uuid" != $1)
-                AND ("entity_edition_cache_0_1_0"."base_urls" @> ARRAY[$2]::text[])
-                THEN ARRAY[$3]::text[]
+                ("entity_edition_cache_0_1_0"."base_urls" @> ARRAY[$1]::text[])
+                THEN ARRAY[$2]::text[]
                 ELSE ARRAY[]::text[] END))
             FROM "entity_temporal_metadata" AS "entity_temporal_metadata_0_0_0"
             INNER JOIN "entity_editions" AS "entity_editions_0_1_0"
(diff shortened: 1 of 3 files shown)


thehabbos007:
Looks good to me, just curious about how the sentinel value used to work before. What was the reason for `uuid_not_actor`? Was it to just have the parameter filled in all types of queries?

TimDiekmann:
> What was the reason for uuid_not_actor? Was it to just have the parameter filled in all types of queries?

Kind-of, yes. It previously generated a filter if the creator of a type was not the current actor, which does not make sense to have for the `nil` sentinel as no public user was ever able to create something in the DB in the first place. With the `Filter` now being compiled conditionally, the weird test-helper is not needed anymore.

thehabbos007 (dismissed):
I'll look towards the root of the stack now, but this one was short and easy to review,, thanks!

indietyp (dismissed):

indietyp (dismissed):

indietyp (dismissed):

indietyp (approved):

PR #9294, opened 2026-08-24
FE-1471: update petrinaut to use ds segmented control

alex-e-leon:

Updates petrinaut to use the new DS component.

@kube The segmented control in the figma is quite different from your original implementation (gaps, colors/subtlety, etc), and TBH I think in some respects your design was better. In particular gaps between the buttons and the rail in the figma IMO were not great, so I've adjusted the designs to be somewhere in the middle of both designs. The biggest remaining UI difference from your original version is figma version is more square, which I actually don't mind - I think it fits better into the rest of the UI language.

But lmk your feedback - happy to get any suggestions/improvements before bringing in. And if this doesn't fix the alignment issue you mentioned let me know.

This PR:

The changes in this PR:


kube (commented):
We still have the same issue where the thumb is not aligned properly inside the slider.

Screenshot 2026-08-24 at 18 20 28

alex-e-leon:
> We still have the same issue where the thumb is not aligned properly inside the slider.
> 
> Screenshot 2026-08-24 at 18 20 28

Great to hear thats a bug: Fixed

kube (approved):
Perfect! Thank you!

PR #9299, opened 2026-08-24
BE-784: hgraph: Statement parameters travel as iterators to the Postgres client

indietyp:

Statement parameters reach the Postgres client as iterators. `SelectCompiler::compile` returns an `ExactSizeIterator` over the bound parameters instead of a slice, and the pool grows iterator-taking query paths (`query_params_iter`, `GenericClientIter`), so the fourteen call sites that used to collect parameters into vectors now hand their iterators straight through.

This layer carries the signature change alone, so the two compiler PRs above it (#9300, #9301) stay reviewable as behavior changes rather than plumbing.

Review focus: the client needs the parameter count before the wire message goes out, so each call site keeps the iterator's length in step with the placeholder count in the statement text. The adapted tests in `compile/tests.rs` pin the new return type.

## 🔍 What does this change?

- `pool.rs`: `query_params_iter` and `GenericClientIter`, the iterator-shaped client plumbing.
- `SelectCompiler::compile` returns `impl ExactSizeIterator` over the parameters, and the affected tests adapt.
- Fourteen call sites across `ontology/`, `crud.rs`, `traversal_context.rs`, and `knowledge/entity/` pass iterators through instead of collecting.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- The existing postgres-store lib suite covers this, plus the workspace battery: `cargo check`, clippy at zero warnings, `fmt` clean.

## ❓ How to test this?

```bash
cargo nextest run -p hash-graph-postgres-store --lib
```

### DIFF
--- a/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/search.rs
+++ b/libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/search.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 
 use error_stack::{Report, ResultExt as _};
+use futures::TryStreamExt as _;
 use hash_graph_authorization::policies::{MergePolicies, PolicyComponents, action::ActionName};
 use hash_graph_store::{
     entity::{
@@ -272,9 +273,9 @@ where
         compiler.set_limit(candidate_pool);
 
         let (statement, parameters) = compiler.compile();
-        Ok(self
-            .as_client()
-            .query(&statement, parameters)
+
+        self.as_client()
+            .query_raw(&statement, parameters)
             .instrument(tracing::info_span!(
                 "SELECT",
                 otel.kind = "client",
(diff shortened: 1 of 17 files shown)


libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/search.rs
```diff
@@ -284,13 +285,16 @@ where
...
-            .map(|row| EntityId {
-                web_id: row.get(0),
-                entity_uuid: row.get(1),
-                draft_id: row.get(2),
+            .and_then(|row| {
+                core::future::ready(Ok(EntityId {
+                    web_id: row.get(0),
+                    entity_uuid: row.get(1),
+                    draft_id: row.get(2),
+                }))
```
TimDiekmann:
```suggestion
            .map_ok(|row| EntityId {
                web_id: row.get(0),
                entity_uuid: row.get(1),
                draft_id: row.get(2),
```

TimDiekmann (commented):
Just one minor thing

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/search.rs
indietyp: TIL that's a thing.

TimDiekmann: I knew it was, but I had to checkout the code to actually find the correct function 😂

TimDiekmann (dismissed):

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/search.rs
TimDiekmann: (You can imagine how much I messed around with streams in the Postgres implementation 😁)

TimDiekmann (approved):

PR #9303, opened 2026-08-24
BE-781: hgraph: Entity event feed with deletion tombstones

indietyp:

Entities gain deletion tombstones and an event feed. A deletion now writes who deleted and when (transaction and decision time) into `entity_ids`, with a check constraint holding the tombstone total or absent. The new feed module serves ordered change events polled from current rows in two arms: `changed` for present rows and `ended` for decision-closed ones. Consumers that mirror entity state can now hear about deletions, which the store previously forgot.

The migrations pair as in #9302: `V59` (deletion provenance columns), `V60` (feed indexes), and `V61` (feed statistics) alter deployed databases, and the same schema enters the fresh-database `v009` here. The `V60`/`V61` objects exist because the feed's predicates are expressions, which carry no column statistics. The migration comments state what each object fixes and the measured plan bounds.

Review focus: feed completeness at the poll-window edges (an event exactly on a boundary) and across the split between the two arms. The `tests/deletion` suite exercises the crossings.

## 🔍 What does this change?

- `graph-migrations/v009__entities/up.sql`: tombstone columns with their totality constraint and partial index, plus the feed indexes and extended statistics.
- `postgres_migrations/V59`, `V60`, `V61`: the same schema for deployed databases.
- `knowledge/entity/feed.rs`: the feed. `snapshot/entity/channel.rs` carries the deletion fields through snapshots.
- `delete.rs`: tombstone writes. `provenance.rs`: the deletion provenance column.
- `rows.rs`, `table.rs`, `insert.rs`: the `Deleted*` row vocabulary and fixtures.
- `Cargo.toml`: `pin-project-lite` for the feed stream plumbing.
- `tests/deletion/`: a new integration suite (drafts, erase, feed, links, purge).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `tests/deletion/` under the integration harness, which provisions its own scratch database.
- The existing postgres-store lib suite covers this, plus the workspace battery: `cargo check` including all test targets, clippy at zero warnings, `fmt` clean.

## ❓ How to test this?

```bash
cargo nextest run -p hash-graph-postgres-store --lib
turbo run test:integration --filter @rust/hash-graph-postgres-store
```

### DIFF
--- a/libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
+++ b/libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
@@ -119,13 +119,8 @@ CREATE INDEX entity_temporal_metadata_closed_transaction_start
 ON entity_temporal_metadata (lower(transaction_time))
 WHERE draft_id IS NULL AND upper_inf(transaction_time) AND NOT upper_inf(decision_time);
 
--- The feed predicates are expressions, which carry no column statistics, so without these
--- objects the planner sweeps every present row for the ended arm. The mcv object prices the
--- predicate combination both arms share, which bounds the ended arm's anti-join by the
--- candidate count. The expression object gives the transaction-time start a histogram, which
--- fixes the changed arm's cardinality. The bounded plan wins to roughly twenty thousand
--- candidates per poll window. A burst the statistics have not caught up with stretches it to
--- about four times the sweep at a hundred thousand candidates.
+-- Give the planner a histogram for the transaction-time cutoff and multivariate statistics for
+-- the expression predicates shared by the feed arms.
 CREATE STATISTICS entity_temporal_metadata_transaction_start_stats
 ON lower(transaction_time) FROM entity_temporal_metadata;
 CREATE STATISTICS entity_temporal_metadata_feed_stats (MCV)
--- a/libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
+++ b/libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
@@ -6,13 +6,28 @@ CREATE TABLE entity_ids (
     created_by_id UUID NOT NULL,
     created_at_transaction_time TIMESTAMP WITH TIME ZONE NOT NULL,
     created_at_decision_time TIMESTAMP WITH TIME ZONE NOT NULL,
-    PRIMARY KEY (web_id, entity_uuid)
+    deleted_by_id UUID,
+    deleted_at_transaction_time TIMESTAMP WITH TIME ZONE,
+    deleted_at_decision_time TIMESTAMP WITH TIME ZONE,
+    PRIMARY KEY (web_id, entity_uuid),
+
+    -- The deletion tombstone is total or absent: either all three deleted_* columns are set, or
+    -- none is.
+    CONSTRAINT entity_ids_deletion_tombstone_total CHECK (
+        ((deleted_by_id IS NULL) = (deleted_at_transaction_time IS NULL))
+        AND ((deleted_by_id IS NULL) = (deleted_at_decision_time IS NULL))
+    )
 );
 
 CREATE INDEX entity_ids_created_at_transaction_time
 ON entity_ids (created_at_transaction_time);
 CREATE INDEX entity_ids_created_at_decision_time
 ON entity_ids (created_at_decision_time);
+
+CREATE INDEX entity_ids_deleted_at_transaction_time
+ON entity_ids (deleted_at_transaction_time)
+INCLUDE (web_id, entity_uuid)
+WHERE deleted_at_transaction_time IS NOT NULL;
 
 CREATE TABLE entity_drafts (
     web_id UUID NOT NULL,
@@ -93,6 +108,29 @@ ON entity_temporal_metadata
 USING gist (web_id, entity_uuid, transaction_time, decision_time);
 CREATE INDEX entity_temporal_metadata_edition_id_idx
 ON entity_temporal_metadata (entity_edition_id);
+
+-- The update feed polls current rows by their transaction-time start. Every predicate it uses
+-- is a function over a column, so only these partial expression indexes can serve it, one per
+-- feed arm: present rows for `changed`, decision-closed current rows for `ended`.
+CREATE INDEX entity_temporal_metadata_present_transaction_start
+ON entity_temporal_metadata (lower(transaction_time))
+WHERE draft_id IS NULL AND upper_inf(transaction_time) AND upper_inf(decision_time);
+CREATE INDEX entity_temporal_metadata_closed_transaction_start
+ON entity_temporal_metadata (lower(transaction_time))
+WHERE draft_id IS NULL AND upper_inf(transaction_time) AND NOT upper_inf(decision_time);
+
+-- The feed predicates are expressions, which carry no column statistics, so without these
+-- objects the planner sweeps every present row for the ended arm. The mcv object prices the
+-- predicate combination both arms share, which bounds the ended arm's anti-join by the
+-- candidate count. The expression object gives the transaction-time start a histogram, which
+-- fixes the changed arm's cardinality. The bounded plan wins to roughly twenty thousand
+-- candidates per poll window. A burst the statistics have not caught up with stretches it to
+-- about four times the sweep at a hundred thousand candidates.
+CREATE STATISTICS entity_temporal_metadata_transaction_start_stats
+ON lower(transaction_time) FROM entity_temporal_metadata;
+CREATE STATISTICS entity_temporal_metadata_feed_stats (MCV)
+ON draft_id, upper_inf(transaction_time), upper_inf(decision_time)
+FROM entity_temporal_metadata;
 
 CREATE TABLE entity_is_of_type (
     entity_edition_id UUID NOT NULL REFERENCES entity_editions,
--- a/libs/@local/graph/postgres-store/postgres_migrations/V61__entity_feed_statistics.sql
+++ b/libs/@local/graph/postgres-store/postgres_migrations/V61__entity_feed_statistics.sql
@@ -1,10 +1,5 @@
--- The feed predicates are expressions, which carry no column statistics, so without these
--- objects the planner sweeps every present row for the ended arm. The mcv object prices the
--- predicate combination both arms share, which bounds the ended arm's anti-join by the
--- candidate count. The expression object gives the transaction-time start a histogram, which
--- fixes the changed arm's cardinality. The bounded plan wins to roughly twenty thousand
--- candidates per poll window. A burst the statistics have not caught up with stretches it to
--- about four times the sweep at a hundred thousand candidates.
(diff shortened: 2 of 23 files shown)


libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
```diff
@@ -94,6 +109,29 @@ USING gist (web_id, entity_uuid, transaction_time, decision_time);
...
+-- predicate combination both arms share, which bounds the ended arm's anti-join by the
+-- candidate count. The expression object gives the transaction-time start a histogram, which
+-- fixes the changed arm's cardinality. The bounded plan wins to roughly twenty thousand
+-- candidates per poll window. A burst the statistics have not caught up with stretches it to
+-- about four times the sweep at a hundred thousand candidates.
+CREATE STATISTICS entity_temporal_metadata_transaction_start_stats
+ON lower(transaction_time) FROM entity_temporal_metadata;
+CREATE STATISTICS entity_temporal_metadata_feed_stats (MCV)
+ON draft_id, upper_inf(transaction_time), upper_inf(decision_time)
+FROM entity_temporal_metadata;
```
TimDiekmann, line 128: Let's keep these comments more concise, in particular we should not state numbers in migration files. It reads a bit like this comes from an actual run.

libs/@local/graph/postgres-store/postgres_migrations/V61__entity_feed_statistics.sql
```diff
@@ -0,0 +1,17 @@
+-- The feed predicates are expressions, which carry no column statistics, so without these
+-- objects the planner sweeps every present row for the ended arm. The mcv object prices the
+-- predicate combination both arms share, which bounds the ended arm's anti-join by the
+-- candidate count. The expression object gives the transaction-time start a histogram, which
+-- fixes the changed arm's cardinality. The bounded plan wins to roughly twenty thousand
+-- candidates per poll window. A burst the statistics have not caught up with stretches it to
+-- about four times the sweep at a hundred thousand candidates.
```
TimDiekmann: Same as above

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
```diff
@@ -0,0 +1,353 @@
...
+       lower(transaction_time) AS occurred_at,
+       'updated' AS reason,
+       NULL::UUID AS deleted_by_id,
+       NULL::TIMESTAMPTZ AS deleted_at_decision_time
+FROM entity_temporal_metadata
+JOIN entity_editions USING (entity_edition_id)
+WHERE draft_id IS NULL
+  AND upper_inf(transaction_time)
+  AND upper_inf(decision_time)
+  AND lower(transaction_time) > $1
```
TimDiekmann: I don't think we can enforce it (and I'm sure I did the same mistakes inline as well) but SQLFluff would force us to write e.g. `entity_temporal_metadata.web_id` instead of `web_id` due to the `JOIN`. Not sure if it's worth changing, but I see the point. At least we should be consistent with the above where we rename `entity_temporal_metadata` to `closed` (or whatever we'll end up with).

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
```diff
@@ -0,0 +1,353 @@
...
+-- incoming link entities.
+SELECT closed.web_id,
+       closed.entity_uuid,
+       NULL::UUID AS entity_edition_id,
+       NULL::BOOLEAN AS archived,
+       max(lower(closed.transaction_time)) AS occurred_at,
+       'ended' AS reason,
+       NULL::UUID AS deleted_by_id,
+       NULL::TIMESTAMPTZ AS deleted_at_decision_time
+FROM entity_temporal_metadata AS closed
```
TimDiekmann: Closed is a bit misleading here as we use the term `closed` mostly when talking about types and inheritance/multi-entity-types. We could come up with another term which is more suited to temporal bounds, such as `historical` or something I guess. Let's be consistent with all queries here.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
```diff
@@ -0,0 +1,353 @@
...
+        let mut this = self.project();
+        loop {
+            match this.state.as_mut().project() {
+                FeedStateProj::Opening { query } => match ready!(query.poll(cx)) {
+                    Ok(rows) => this.state.set(FeedState::Streaming { rows }),
+                    Err(error) => {
+                        return Poll::Ready(Some(Err(
+                            Report::new(error).change_context(QueryError)
+                        )));
+                    }
```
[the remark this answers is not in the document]
TimDiekmann: Isn't it even forbidden to re-poll a completed future?

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
```diff
@@ -0,0 +1,353 @@
...
+
+-- The entity was purged: its editions and temporal rows are gone, and the tombstone on its
+-- entity_ids row is the only readable record that it existed and who deleted it. Erase
+-- removes the entity_ids row itself and leaves no tombstone, so an erased entity never
+-- appears here.
+SELECT web_id,
+       entity_uuid,
+       NULL::UUID AS entity_edition_id,
+       NULL::BOOLEAN AS archived,
+       deleted_at_transaction_time AS occurred_at,
```
TimDiekmann, line 126: Remind me: `deleted_at_transaction_time` Is _only_ set when actually erase the entity (i.e. not when archiving it), right? (i.e. through the admin API).

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
```diff
@@ -0,0 +1,353 @@
...
+       NULL::UUID AS entity_edition_id,
+       NULL::BOOLEAN AS archived,
+       max(lower(closed.transaction_time)) AS occurred_at,
+       'ended' AS reason,
+       NULL::UUID AS deleted_by_id,
+       NULL::TIMESTAMPTZ AS deleted_at_decision_time
+FROM entity_temporal_metadata AS closed
+WHERE closed.draft_id IS NULL
+  AND upper_inf(closed.transaction_time)
+  AND NOT upper_inf(closed.decision_time)
```
TimDiekmann: Tricky!
We may want to extend the doc above slightly to explain why we only need one variant. Presumably, if we'd had `valid_time`, this would be the `NOT upper_inf` and the other two would be `upper_inf`?

TimDiekmann (changes requested):
Looks good, only have a few minor suggestions.

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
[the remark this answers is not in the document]
indietyp: Stream:
```
Once a stream has finished (returned Ready(None) from poll_next), calling its poll_next method again may panic, block forever, or cause other kinds of problems; the Stream trait places no requirements on the effects of such a call. However, as the poll_next method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the stream’s state.
```

Future:
```
Once a future has completed (returned Ready from poll), calling its poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call. However, as the poll method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the future’s state.
```

_so in theory_ what I am doing here is fine, but it doesn't cost much to make it be a `FusedStream` instead (and it is more correct, so will be doing that)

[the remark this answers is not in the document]
TimDiekmann: Oh, yeah, futures vs streams 🙃  Thanks!

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
indietyp: Yep, correct - when we set purge (`POST /entities/delete` with `scope: "purge"`) or during user deletion.

indietyp: (there is also `erase` which removes the entity outright, but I don't think we use that anywhere)

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
indietyp: I'll add the comment; that's correct. The reason is that we can make an implicit assumption here: if the decision time changed (i.e., it was superseded by the current one) but the transaction time is still infinite, we have a new entity. With valid time, we'd need an additional query to see whether the decision time or valid time changed. Here we get it for basically free™.

indietyp: done in [47e5bc3](https://github.com/hashintel/hash/pull/9303/commits/47e5bc3444433bf09aeff2c15504412fdc77db99)

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
indietyp: done in [47e5bc3](https://github.com/hashintel/hash/pull/9303/commits/47e5bc3444433bf09aeff2c15504412fdc77db99)

libs/@local/graph/postgres-store/src/store/postgres/knowledge/entity/feed.rs
indietyp: done in [47e5bc3](https://github.com/hashintel/hash/pull/9303/commits/47e5bc3444433bf09aeff2c15504412fdc77db99)

libs/@local/graph/postgres-store/postgres_migrations/V61__entity_feed_statistics.sql
indietyp: done in [47e5bc3](https://github.com/hashintel/hash/pull/9303/commits/47e5bc3444433bf09aeff2c15504412fdc77db99)

libs/@local/graph/migrations/graph-migrations/v009__entities/up.sql
indietyp: done in [47e5bc3](https://github.com/hashintel/hash/pull/9303/commits/47e5bc3444433bf09aeff2c15504412fdc77db99)

TimDiekmann (dismissed):

TimDiekmann (approved):

PR #146, opened 2022-01-11
Change the log output to `stderr` instead of `stdout`

TimDiekmann:
## 🌟 What is the purpose of this PR?

- Change the log output to `stderr` instead of `stdout`
- Change default log format to `full` instead of `pretty` (see #145 for demos)

## 🛡 Tests
- ✅ Manual Tests

PR #9305, opened 2026-08-24
BE-782: Fit 2D semantic maps over graph entities and serve them as tiles

indietyp:

The new `hash-graph-atlas` crate fits 2D semantic maps over the entity embeddings stored in the graph, blending semantic similarity (what entities mean) with relational structure (how they connect), and distills each fitted map into a small encoder that places new entities on the existing map without refitting. Serving reads the fitted artifacts from disk and answers tile queries over them.

The crate arrives whole and inert. Nothing links it until #9306 wires the CLI, the API proxy, and the compose service, so this layer reviews as a self-contained library plus its own `cli` module.

Review guidance for 627 files: line-reading is the wrong mode. The contract is at the boundaries, and they are small. The `cli` module is the only public one (everything else is `pub(crate)`), the on-disk artifact formats live in the `file` module with SHA-256 content identity in `integrity`, and the wire format is documented in `wire.md`. The module docs top-down from `lib.rs` give the architecture: the SALT pipeline sits on domain-independent foundation modules (`math`, `random`, `bitset`, `identity`, `morton`, `runs`, `progress`, `file`).

Review focus: determinism and cross-machine identity of the artifacts. x86-64 builds now assume the x86-64-v3 baseline (AVX2, FMA, BMI2) because the math kernels pin fused-multiply-add semantics, and the crate verifies the baseline at process entry so that a mismatched machine fails diagnosably. That `.cargo/config.toml` change applies workspace-wide and is the one hunk here whose effect reaches beyond the crate.

---

For an authorisation-focused review, one wants to take a look at (in order of building up on each other):
- src/api/visibility.rs - how we authenticate and retrieve the information. This will undoubtedly change once the Kratos changes are in / will need to be retrofitted.
- src/serve/authorization.rs - this is the `Atlas-Authority` token that we send over the wire to authenticate
- src/api/authorization.rs - is the bridge between the API and the actual serve layer

The part to get right here is the cache semantics, as well as the cascading effect: we cannot backfill tiles, because that would trivially expose information about which tiles actually have occupancy. Everything revolves around that fact. The module enforcing it is `serve/schedule/`: a restricted response delivers from its own first-occupant cascade over the visible rows alone, so a hidden row contributes to no output at all rather than leaving a hole.
The fitting process (so everything in `salt/`) is authorisation-blind. It only does Postgres queries (these are in the `postgres/` folders respectively). The serve layer applies authorisation, supplied via the API layer.

We can model it through the following diagram:

```mermaid
flowchart TD
    G["Gateway authenticates the session,<br/>states the actor in X-Authenticated-User-Actor-Id"] --> R{Route}

    R -->|manifest| P["Presented: reads the token if any, never rejects<br/>(api/authorization.rs)"]
    P --> MJ["Judge the generation, bind the filter,<br/>seal Scope: actor, filter digest, delivery cut<br/>XChaCha20-Poly1305 under a per-generation HKDF key"]
    MJ --> T[Atlas-Authority header back to the client]

    R -->|data route| AD["admit (api/authorization.rs)"]
    AD --> Q{"tag valid? actor matches the header?<br/>issued_at inside the window? epoch held?"}
    Q -->|any refusal| U["uniform 401, oracle-free:<br/>client re-fetches the manifest"]
    Q -->|ok| S["sealed Scope enters the visibility cache<br/>key = generation + actor + filter digest<br/>soft 8 min refresh-behind, hard 10 min ceiling<br/>(hard = revocation lag = token-age bound)"]
    S -->|hit| V[VisibilityProof]
    S -->|miss, one resolve per scope| C["compile policy + caller filter<br/>(serve/hydrate/compile.rs)"]
    C --> E{"admits_every_row?<br/>no caller filter AND policy filter = empty Filter::All"}
    E -->|yes| F["full_visibility: every row of every domain"]
    E -->|no| M["from_masks: fail-closed masks"]
    F --> V
    M --> V
    V --> W["assembly answers under the proof:<br/>every corpus-bearing response is masked"]
    W --> D["delivery from a per-scope schedule at cut k<br/>(serve/schedule/): built over visible rows alone,<br/>hidden rows contribute to nothing - no tile backfill"]
```

The crate layout is roughly:

```
- api: layer between HTTP and the actual serving (`serve`)
- bitset: domain-typed membership sets and persistable bit frames (the generic in-memory dense forms already come from `hashql_core`)
- cli: cli components, a superset of what either entry point exposes (the standalone binary lacks serve, the graph binary lacks dump/reports), to be reworked into a single binary endpoint
- dataset: ways to get the data out of the graph, and also to adapt to other systems (used for e.g. memory based testing)
- file: file formats (dumps and archives) that atlas uses
- identity: identity information, such as ids used throughout the system
- integrity: security relevant information, such as hashing infrastructure, secrets, etc.
- math: what the name says - strongly typed math primitives with a SIMD first approach
- morton: implementation of [z-order keys](https://en.wikipedia.org/wiki/Z-order_curve)
- postgres: well postgres
- random: random utilities, primarily adapters / additional functions not exposed by rand
- runs: fencepost-compressed per-key item lists, supplementing sprs
- salt: ML
- serve: taking what was generated by salt, and turning it in something that is consumed through api
- allocator: live-byte accounting at the allocator boundary, to be potentially subsumed into hashql_core
- bench: benchmarking infrastructure + entrypoint
- offload: rayon infrastructure to bridge async + sync
- device: chooses which device to run on training and inference
- progress: progress support for ML
```

## 🚫 Blocked by

- [x] #9469, the middleware layer that was to sit below this branch in the stack. `gh stack modify` refuses an insert while a stack holds merged branches whose local refs git has pruned and for which the tool never recorded a head ([gh-stack #146](https://github.com/github/gh-stack/issues/146), open since 2026-06-25, unfixed on `main` at v0.1.0), so it is a single branch cut from `main` and merges ahead of this one.

## 🔍 What does this change?

- `libs/@local/graph/atlas/`: the crate (614 files), its docs, benches, and tests.
- `.cargo/config.toml`: the x86-64-v3 baseline for x86-64 targets, workspace-wide.
- `.clippy.toml`: `core::fmt::LowerExp` joins `allow-renamed-params-for`.
- `.dockerignore`: excludes `wire.md`.
- `.github/actions/prune-repository/prune.py`: learns the atlas edge (inert until the frontend merges).
- Workspace `Cargo.toml`, `Cargo.lock`, `yarn.lock`: crate and package registration.
- Dependency diagrams regenerated.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Serving is unreachable at this layer: no binary links the crate until #9306.

## 🛡 What tests cover this?

- The crate's unit tests and doctests cover the fit and serve paths.
- The workspace battery: `cargo check`, clippy at zero warnings across targets and features, `fmt` clean.

## ❓ How to test this?

```bash
cargo nextest run -p hash-graph-atlas --lib
cargo test -p hash-graph-atlas --doc
```


TimDiekmann (changes requested):
Could you please give instructions on how to review this?

libs/@local/graph/atlas/package.json
```diff
@@ -0,0 +1,31 @@
...
...it_slice_array_starts_as_empty_frames bitset::tests::dense_bit_slice_array_indexes_independent_frames bitset::tests::dense_bit_slice_array_round_trips_through_its_region bitset::tests::dense_bit_slice_array_of_no_frames bitset::tests::dense_bit_slice_array_refuses_misshapen_regions bitset::tests::dense_bit_slice_array_zerocopy_doors_carry_the_region_invariant allocator::tests:: dsquare::tests::block_height_invariance dsquare::tests::the_solution_reproduces_the_right_hand_side dsquare::tests::both_dots_reduce_equal_inputs_to_identical_bits dsquare::tests::the_row_dot_and_dvecn_dot_reduce_equal_bytes_to_identical_bits dsquare::tests::every_row_is_aligned_for_simd dsquare::tests::writes_through_rows_land_at_their_offsets dsquare::tests::the_zero_matrix_reads_zero_everywhere dsquare::tests::the_empty_matrix_factors_and_solves dsquare::tests::a_row_past_the_order_panics dsquare::tests::a_mismatched_right_hand_side_panics dsquare::tests::matrix_drop_returns_the_buffer_to_its_allocator dsquare::tests::factor_drop_returns_the_moved_buffer_to_its_allocator runs::tests::structure_view scalar::tests::try_from_bytes morton::tests::try_from_ref authorization::tests::archived_actor_entity_uuid",
```
TimDiekmann: It would be good if this would be more maintainable. This way, it's very easy to miss.

libs/@local/graph/atlas/src/api/visibility.rs
```diff
@@ -0,0 +1,550 @@
...
+impl FromRequestParts<AppState> for Actor {
+    type Rejection = Problem<'static>;
+
+    fn from_request_parts(
+        parts: &mut Parts,
+        _state: &AppState,
+    ) -> impl Future<Output = Result<Self, Self::Rejection>> + Send {
+        future::ready(actor(parts).map(Self))
+    }
+}
```
TimDiekmann: This reinvents the existing Authentication setup, please use already existing authentication we have in `hash-graph-api`, so we don't have multiple locations where we define how we get the `AuthenticatedActor`.
The auth-setup itself lives in `hash-graph-authentication`. We may need to move some middleware from the api-crate to the authN-crate to re-use it here. This should reduce the amount of work being required.

libs/@local/graph/atlas/src/api/visibility.rs
```diff
@@ -0,0 +1,550 @@
...
+//!
+//! A [`VisibilityProof`] masks every corpus-bearing response, and [`Visibility`] is that proof for
+//! one request. Extraction admits the presented token through [`authorization::admit`], then
+//! resolves the scope the token seals (actor and filter digest, the visibility key's own fields)
+//! through the store, held in the visibility cache for its reuse window. A data route that resolves
+//! visibility without an admitted token is unrepresentable, because the key's identity comes out of
+//! the sealed scope.
+//!
+//! The `X-Authenticated-User-Actor-Id` header names the actor, and that header is the trust
+//! boundary the graph's REST API stands on. The gateway authenticates the session and states the
```
TimDiekmann: This is not true (anymore): https://github.com/hashintel/hash/pull/9233

libs/@local/graph/atlas/src/api/visibility.rs
```diff
@@ -0,0 +1,550 @@
...
+        compile::{ProofError, visibility_proof},
+    },
+    schedule::ViewSchedule,
+};
+
+/// The header naming the authenticated actor.
+///
+/// Parity with the graph's REST API, which accepts this name, fixes the spelling. Headers this
+/// crate introduces carry no prefix.
+const ACTOR_HEADER: &str = "X-Authenticated-User-Actor-Id";
```
TimDiekmann: See above

libs/@local/graph/atlas/src/api/visibility.rs
indietyp: Yes, I understand, but I deliberately didn't want to depend on `hash-graph-api` since it doesn't make sense to do so just for the middleware. I tracked all of this in BE-805; I'm happy to do it as part of this PR or as a follow-up. Still, I didn't want to make this already massive PR bigger by moving things into, e.g. `hash-graph-authentication`, especially considering that before yesterday the way I was doing this was basically just 10 lines of code that were the same between them (which is no longer true), so it's just unfortunate timing^^. Several adjustments are needed here, including removing the hash-api proxy, moving to kratos, and moving the middleware into `hash-graph-authentication`, which I didn't want to do as part of this. I can prepare this under this PR and then adjust this PR, making it reviewable (for each), or you can take it up if you'd like. This should then also make the authentication wiring straightforward.

libs/@local/graph/atlas/src/cli/serve.rs
```diff
@@ -0,0 +1,461 @@
...
+        visibility: VisibilityLimits,
+        ensure: Option<EmbeddingEnsure>,
+        RequestFacilities {
+            provider,
+            service_secret,
+            rate_limit,
+        }: RequestFacilities<P>,
+    ) -> Result<Router, ServeError>
+    where
+        P: AuthenticationProvider<ActorId> + 'static,
```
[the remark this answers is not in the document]
indietyp: intentional

.cargo/config.toml
```diff
@@ -1,6 +1,12 @@
 [registries.crates-io]
 protocol = "sparse"
 
+# x86-64 builds assume the x86-64-v3 baseline (AVX2, FMA, BMI2): the math kernels pin fused
+# multiply-add semantics and take the hardware lowering from it. `hash-graph-atlas` verifies the
+# baseline at process entry, turning a mismatched machine into a diagnosable failure.
+[target.'cfg(target_arch = "x86_64")']
+rustflags = ["-Ctarget-cpu=x86-64-v3"]
```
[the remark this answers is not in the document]
indietyp: explanation added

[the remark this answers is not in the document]
indietyp: explanation added

libs/@local/graph/atlas/package.json
indietyp: fixed in a630dd736d

libs/@local/graph/atlas/src/api/visibility.rs
indietyp: fixed in multiple places (w/ #9469) in addition to 2acb14a8f3, 0f7e70ffd6

libs/@local/graph/atlas/src/api/visibility.rs
indietyp: fixed in multiple places (w/ #9469) in addition to 2acb14a8f3, 0f7e70ffd6

libs/@local/graph/atlas/src/api/visibility.rs
indietyp: fixed in multiple places (w/ #9469) in addition to 2acb14a8f3, 0f7e70ffd6

libs/@local/graph/atlas/benches/projector_backend.rs
```diff
@@ -0,0 +1,240 @@
...
+                || (),
+                |()| black_box(stepper.refresh(black_box(&batch))),
+                BatchSize::PerIteration,
+            );
+        });
+
+        // These decomposition phases record autodiff graphs no backward consumes. Orphan graphs pin
+        // device buffers until reclamation, which a sampling loop outruns on a pooled asynchronous
+        // device, so the decomposition stays a synchronous-backend instrument. `refresh` and `step`
+        // are the production motions every backend measures.
```
vilkinsons: Not a huge fan of the intelligibility/writing style of these comments. Would like us to clean these up (in post, please).

TimDiekmann (dismissed):

TimDiekmann (approved):

libs/@local/graph/atlas/benches/projector_backend.rs
indietyp: tracked in [BE-822](https://linear.app/hash/issue/BE-822/atlas-clean-up-clarity-of-documentation)

vilkinsons (approved):

PR #9309, opened 2026-08-24
FE-855: Prevent extending types without create permission in UI

alex-e-leon:

Hide extending a type when user does not have create permission for that type in UI. Note that there is another ticket for the BE to prevent this behaviour as well: https://linear.app/hash/issue/BE-786/validate-instantiate-permissions-when-extending-types

This PR:

The changes in this PR:

### DIFF
--- a/apps/hash-api/src/graph/ontology/primitive/entity-type.ts
+++ b/apps/hash-api/src/graph/ontology/primitive/entity-type.ts
@@ -72,6 +72,51 @@ export const checkPermissionsOnEntityType: ImpureGraphFunction<
 };
 
 /**
+ * Check the requesting user's permissions on multiple entity types at once.
+ *
+ * Uses the bulk `hasPermissionForEntityTypes` check per action, so the number of
+ * graph calls is fixed regardless of how many entity types are passed.
+ */
+export const checkPermissionsOnEntityTypes: ImpureGraphFunction<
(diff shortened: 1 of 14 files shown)


vilkinsons:
@CiaranMn I've functionally tested this. Can you confirm happy with the approach in code?

CiaranMn (commented):
The code looks good for what it does, although to note that it is still possible for users to add these types as a parent to an existing type via`inheritance-row.tsx` in `EntityTypeEditor`.

Preventing this would require one of:
1. updating the entity type options provided by `definition-tab.tsx` to request the user's permissions on _all_ entity types and filter out the ones that they cannot instantiate – ideally this would involve updating `checkPermissionsOnEntityType` / `checkUserPermissionsOnEntityTypeResolver` to actually accept multiple type ids, as the Graph API endpoint ultimately relied on accepts multiple type ids (and we wouldn't want to fire off hundreds of requests to check this). Or, add a new bulk GraphQL query for this to limit the changes elsewhere which would be implied by making it multiple everywhere.
2. requesting permission at the point that a user tries to add a type as a parent, and rejecting and showing a modal when the user cannot instantiate entities of the type.

(1) is better UX (don't even give the user the option to do something they cannot), but puts another API request in the way of loading the options menu.

CiaranMn (approved):

PR #9316, opened 2026-08-25
BE-758: Add rate limiting to the Graph API

TimDiekmann:

The Graph enforced no rate limits. Direct API exposure and anonymous access need per-principal limits in the Graph itself, so self-hosted deployments get them without ingress infrastructure. These are proper-use limits — DoS protection stays at the edge.

Two keyed GCRA limiters. An **address gate** ahead of the authentication middleware, so an invented credential is throttled before it reaches the uncached Kratos verification, and a **principal limiter** behind it, keyed by actor for authenticated callers and by address for anonymous ones.

Ships in **observe mode**: over-budget requests are served and recorded, so the budgets can be sized against real traffic before anyone is turned away.

## 🔗 Related links

- BE-758
- https://github.com/hashintel/internal-infra/pull/338 — sets `HASH_GRAPH_CLIENT_IP_SOURCE` and the mode _(internal)_

## 🚫 Blocked by

- [ ] https://github.com/hashintel/internal-infra/pull/338 — without it the limiter falls back to the connection address, and the Service Connect ingress proxy makes that the same peer for every request, collapsing all callers into one budget

## 🔍 What does this change?

- `rest::rate_limit` with the two middlewares, the configuration, and address resolution
- IPv6 addresses group by /64, and a v4-mapped address shares the v4 address's budget — a `BucketKey` newtype makes that the only way an address reaches a limiter, since one /64 allocation would otherwise hold 2^64 budgets and store entries
- The client address is read from a configured source (connection, `X-Forwarded-For`, `CF-Connecting-IP`), because behind a proxy the connection's peer is the proxy
- A denial answers `429` with `Retry-After` and nothing else. The `RateLimit-*` fields are an IETF draft that has since replaced them with a single structured field, so an admitted request carries no rate-limit field at all
- Callers presenting the verified service secret pass both middlewares unchecked
- `attach_request_middlewares` holds the layer stack: the gate is a `layer` and so also covers the 404 fallback, authentication and the principal limiter are `route_layer`s and are not reached by an unmatched path, and the health probe is merged afterwards so it carries none of them
- Keys are evicted on a 60s sweep, which also emits one aggregate log line per interval — store sizes always, the interval's counters at warn when it saw a denial, a fallback or an unresolvable address
- Limiter state lives in process memory, so enforcement is per instance: N instances admit up to N times the configured rates, and a rolling release starts every budget over

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- **No instrumentation.** The limiter ships with logs only; metrics are tracked in Linear as BE-774. The sweep's interval line carries the denial counters and store sizes, which is enough to size the budgets, but not enough to alert on. Adding metrics later needs no change to the decision path.
- **Per-instance enforcement.** A deployment with N instances admits up to N times the configured rates, and a rolling release starts every budget over. Shared state was considered and rejected: it buys precision these limits do not need, at the cost of a dependency on the request path that would have to fail open anyway.
- **One flat anonymous budget.** Per-endpoint tiers and cost weighting are decided with the public `/v1` route shapes in BE-757, so a generous type read and an expensive query currently draw on the same budget.
- **Store growth is bounded by the sweep interval**, not by a key cap. Eviction is the only mechanism releasing keys, so a burst of distinct addresses is retained until the next sweep. Tuning that interval trades memory against log volume; a cap would decouple them.

## 🐾 Next steps

Tracked in Linear: BE-774 (instrumentation), BE-757 (per-endpoint tiers), BE-791 (a `Secret` type for credential config), BE-795 (carry the authentication `Report` to the rejection), BE-796 (resolve an invalid session as anonymous rather than rejecting it).

## 🛡 What tests cover this?

New behavioural tests in `rest::rate_limit::tests` and `rest::rate_limit::address::tests`, chosen so that removing the code they cover fails a test rather than passing quietly. They pin, among others: the traversal order and that an exhausted gate answers without the identity provider being consulted; that an enforced denial never runs the inner stack; that the gate covers unmatched paths while a router merged afterwards does not; the `/64` and v4-mapped bucketing; that a wrong service secret bypasses nothing; that a missing authentication or gate extension answers 500 rather than skipping a budget; that one request produces exactly one count when its address cannot be resolved; that observe is the default and serves without any rate-limit field; and that the sweep releases every store and the spawned maintenance task drives it.

`attach_request_middlewares` is driven directly, so what each route carries is read off the production assembly rather than a copy of it.

## ❓ How to test this?

1. Check out the branch and start the Graph: `cargo run --bin hash-graph --all-features -- server`
2. The startup line reports the effective mode and every quota
3. `curl -i localhost:4000/entities/query -X POST -d '{}' -H 'content-type: application/json'` past the anonymous burst — the request is served, and the sweep reports the denial within a minute
4. Restart with `HASH_GRAPH_RATE_LIMIT_MODE=enforce HASH_GRAPH_RATE_LIMIT_GATE_BURST=2` and repeat: the third request answers `429` with `Retry-After`
5. `curl -i localhost:4000/health` stays available past any budget, and `hash-graph server --healthcheck` still exits 0

### DIFF
--- a/libs/@local/graph/api/src/rest/mod.rs
+++ b/libs/@local/graph/api/src/rest/mod.rs
@@ -640,7 +640,7 @@ where
         .fold(Router::new(), Router::merge)
         .merge(hashql::HashQlResource::routes())
         .fallback(|| {
-            tracing::error!("404: Not found");
+            tracing::debug!("404: Not found");
             async { StatusCode::NOT_FOUND }
         });
 
--- a/libs/@local/graph/api/src/rest/mod.rs
+++ b/libs/@local/graph/api/src/rest/mod.rs
@@ -16,6 +16,7 @@ pub mod admin;
 pub mod auth;
 pub mod http_tracing_layer;
 pub mod probe;
+pub mod rate_limit;
 
 pub mod hashql;
 mod json;
@@ -38,6 +39,7 @@ use axum::{
 use error_stack::{Report, ResultExt as _};
 use futures::{SinkExt as _, channel::mpsc::Sender};
 use hash_codec::numeric::Real;
+use hash_graph_authentication::provider::{AuthenticationProvider, Caller};
 use hash_graph_authorization::policies::store::{PolicyStore, PrincipalStore};
 use hash_graph_embeddings::{EmbeddingError, EmbeddingGenerator as _, OpenAiEmbeddingClient};
 use hash_graph_postgres_store::store::{PostgresStorePool, error::VersionedUrlAlreadyExists};
@@ -522,6 +524,7 @@ where
     pub session_auth: auth::KratosSessionConfig,
     pub cloudflare_access: Option<auth::CloudflareAccessConfig>,
     pub service_secret: String,
+    pub rate_limit: rate_limit::RateLimitConfig,
     pub compiler: Arc<hashql::CompilerContext>,
     pub clustering: Arc<ClusteringContext>,
     /// Whether to serve an interactive rendering of the `OpenAPI` specification.
@@ -530,8 +533,8 @@ where
     pub serve_api_reference: bool,
 }
 
-/// A [`Router`] that only serves the `OpenAPI` specification (JSON, and necessary subschemas) for
-/// the REST API.
+/// A [`Router`] serving the `OpenAPI` specification (JSON, and necessary subschemas) for the
+/// REST API.
 ///
 /// The specification is served at `/openapi.json`. It references its subschemas by relative path
 /// (`./models/…`), so `/models/{path}` has to stay a sibling of the specification for those
@@ -547,7 +550,6 @@ pub fn openapi_only_router(serve_api_reference: bool) -> Router {
         serve_api_reference.then(|| Html(Scalar::new(open_api_doc.clone()).to_html()));
 
     let mut router = Router::new()
-        .merge(probe::router())
         .route("/openapi.json", get(|| async { Json(open_api_doc) }))
         .route("/models/{*path}", get(serve_static_schema));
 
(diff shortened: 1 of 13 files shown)


libs/@local/graph/api/src/rest/rate_limit/config.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Sustained requests per hour each actor may send, counted across all its source addresses.
+    #[cfg_attr(
+        feature = "clap",
+        clap(
+            long,
+            env = "HASH_GRAPH_RATE_LIMIT_ACTOR_PER_HOUR",
+            default_value = "6000"
+        )
+    )]
+    pub rate_limit_actor_per_hour: NonZeroU32,
```
thehabbos007, line 123: Sensible defaults for all of the values in this file, just thinking about whether or not we should have a separate allowlist for internal systems talking to the graph. Integrations would likely hit a high rate across all actors (i know this is for a single actor) but since it is also IP based, increasing quota for IPs, or some other form of identification would probably be worthwhile

libs/@local/graph/api/src/rest/mod.rs
```diff
@@ -572,55 +631,50 @@ where
     ));
     let service_secret: Arc<str> = Arc::from(dependencies.service_secret);
 
+    let rate_limiters = Arc::new(rate_limit::RateLimiters::new(&dependencies.rate_limit));
+    rate_limit::spawn_maintenance(Arc::downgrade(&rate_limiters));
```
thehabbos007, line 635: Ah, so since we downgrade here, the maintenance task could end up being drop'd in its loop. I was confused about that part, but it makes sense because of this here

libs/@local/graph/api/src/rest/mod.rs
```diff
@@ -572,55 +631,50 @@ where
...
+    .layer(Extension(dependencies.clustering));
 
     if let Some(query_logger) = dependencies.query_logger {
         router = router.layer(Extension(query_logger));
     }
 
-    router.merge(openapi_only_router(dependencies.serve_api_reference))
+    // The health probe is merged after the layers, so it is served outside the middlewares and
+    // carries no budget.
+    router.merge(probe::router())
```
thehabbos007, line 677: nice

thehabbos007 (dismissed):
Looks good to me :)

libs/@local/graph/api/src/rest/rate_limit/config.rs
TimDiekmann: I think we can lift this restriction for a few endpoints through a specific authentication route, such as certain special service tokens. Regardless, also integrations should try to take rate limits into account and defer to bulk-operations over single mutations (I appreciate this is not possible everywhere right now). I added some IP whitelisting previously, but decided to defer it a bit. A typical whitelist would be the loopback or the VPC CIDR I guess.
Though, we cannot lift this for endpoints coming from the frontend as every origin could trivially be spoofed.

TimDiekmann: I chose to not add too many configs, yet, as configuring it all through env-vars is really hard. I'm going to shortly add a config-crate, where things can be much more granular.

libs/@local/graph/api/src/rest/mod.rs
TimDiekmann: I had quite a few thoughts on this actually. I don't think any external people need `/health` in the first place. But default rate limits would be hit as every request from the ALB would be anonymous (no secrets). This is where I added IP filters in the first place, though, as `/health` doesn't really do anything, I left this unbudgeted 🙂

indietyp (dismissed):

indietyp (approved):

PR #9352, opened 2026-08-26
FE-1510: Render the Petrinaut canvas centered in a single initial render

kube:

Opening a net painted the canvas twice: once at the origin at zoom 1, then again once a `fitView` effect centered it. The canvas now renders centered on the net from its very first frame. petrinaut-core owns node dimensions, net bounds and zoom limits, and the canvas derives its initial viewport and zoom floor from them during render — no fit effects, no manual memoization, React Compiler throughout.

https://github.com/user-attachments/assets/fb85d35e-fe40-4563-98c9-11a7a26bcd38

## 🔗 Related links

- [FE-1510](https://linear.app/hash/issue/FE-1510/render-the-petrinaut-canvas-centered-in-a-single-initial-render)

## 🔍 What does this change?

- petrinaut-core `layout/` gains the render dimension constants (moved from the UI), `getComponentInstanceHeight` (port-based height), and a new `geometry.ts` with `getBoundsOfCenteredBoxes`, `getMinZoomForBounds` and `ZOOM_PADDING`. `layoutNodeDimensions` is now derived (per-axis max) instead of hand-copied.
- `SDCPNView` measures its container (`useContainerSize`: layout effect + ResizeObserver, null until the element has area; the first real size lands pre-paint, later sizes only once they have settled for 100ms, so resizes and panel animations re-render once instead of per frame) and only then mounts `<ReactFlowProvider key={petriNetId}>` with the new `SDCPNCanvas`. The initial viewport is computed in a lazy `useState` from core bounds and React Flow's pure `getViewportForBounds`, and passed as `defaultViewport` — the first render is the final one. Switching nets re-centers via the keyed remount, replacing the `fitView`-on-id-change effect.
- `minZoom` is derived during render: the core candidate, clamped by a `useStore` selector that yields the current zoom only while it is below the floor, then debounced with a new `useDebouncedValue`. The lodash `useDebounceCallback` and `useResizeObserver` utilities are deleted; the `onInit`/instance-state/`onViewportChange` plumbing is gone.
- Node CSS no longer duplicates sizes: React Flow sizes the node wrapper from the declared width/height and the cards fill it. This corrects two silent drifts (compact cards declared 48px but rendered 50px; classic component instances declared 120px but rendered 96px) and makes component instances actually grow with their port count.
- `ui/lib/viewport.ts` is now pure: bounds come from core instead of `instance.getNodesBounds`, and `getViewportRect` takes a size instead of a DOM element. The recenter-on-panel-open hook uses `useReactFlow` and the measured container size.
- Auto-layout on import uses `layoutNodeDimensions`, so its output no longer depends on the compact/classic setting (the drift the core docs warned about).

This PR:

The changes in this PR:

No user-guide page describes initial centering or the zoom floor; behaviour is the same, minus the initial jump.

The changes in this PR:

## ⚠️ Known issues

Intentional behaviour deltas, all making declared values match what was actually rendered:

- Compact node dimension constant is 50px (was 48; the card always rendered ~50).
- Classic component-instance dimension is 96px (was 120; the CSS always rendered 96), so derived layout dimensions for instances are 96 — auto-layout spacing around instances tightens slightly.
- Component instances with more than 2 ports now actually grow, spreading their port handles.

## 🐾 Next steps

- `libs/@hashintel/petrinaut/docs/visual-settings.md` mentioned a "fit-to-view" button the viewport controls do not have — pre-existing drift, fixed in #9353 (FE-1511).

## 🛡 What tests cover this?

- New: `layout/geometry.test.ts` and `layout/dimensions.test.ts` in petrinaut-core (bounds, min-zoom formula, port heights, derived layout dims).
- Extended: `ui/lib/viewport.test.ts` covers `getInitialViewport` (centering, zoom cap, large-net fit) and its fixtures moved to the center-origin convention the runtime actually uses.

## ❓ How to test this?

1. `yarn workspace @hashintel/petrinaut dev`, open the `Petrinaut → Handle Spike With Sir` story.
2. The net appears centered immediately — no jump after load, no React Flow container warnings in the console.
3. Zoom out: the floor stops the net at a readable fraction of the viewport. Toggle compact/classic in the viewport settings: positions hold and node DOM sizes match the declared dimensions exactly.

## 📹 Demo

Screenshots pending.


.changeset/canvas-single-render.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashintel/petrinaut-core": patch
+"@hashintel/petrinaut": patch
+---
+
+The canvas renders centered on the net from its first frame, instead of jumping there after a first paint at the origin. petrinaut-core now owns node dimensions, net bounds and zoom limits; component instances grow with their port count; auto-layout on import no longer depends on the compact/classic setting.
```
CiaranMn: This could do with splitting up to better describe the actual changes to each package, rather than lumping them into one.

YannisZa (dismissed):

.changeset/canvas-single-render.md
kube: Fixed

YannisZa (approved):

PR #9362, opened 2026-08-26
FE-1500: Add example models and the read-only example page

kube:

Publishes seven Petrinaut models as website examples: canonical `/examples/<slug>` pages rendering the full Petrinaut component read-only, with the scenario, subnet, and canvas selection encoded in the URL. Examples are embed and deep-link content, so there is no index page: `/examples` redirects to `/`. Stacked on the controlled-navigation PR.

## 🔗 Related links

- [FE-1500](https://linear.app/hash/issue/FE-1500/view-only-iframe-embed-of-a-petrinaut-net)

## 🔍 What does this change?

- Adds the model sources under `src/examples/models`. The precompiled HIR runtime artifacts are NOT committed: `scripts/generate-example-artifacts.ts` writes them to the gitignored `src/examples/generated/` at build time. A new `examples:generate` Turbo task feeds `build`, `lint:tsc`, and `test:unit`, so local builds, CI, and Vercel deploys (which build through Turbo) all generate them; `yarn dev` generates before starting Vite.
- Adds the catalog: metadata (titles and safe parameter bounds for scenario sliders), a cached model loader, and a runtime-artifact loader. The metadata module stays free of Petrinaut runtime imports so the oEmbed Vercel function can validate URLs without bundling models.
- Adds the canonical `/examples/$slug` route rendering `<Petrinaut readonly>` from an in-memory read-only document handle, and an `/examples` redirect to `/`. (The `review` presentation profile that hides — rather than disables — authoring controls arrives later in the stack, in #9425.)
- Encodes the URL contract in `example-search.ts`: scenario, subnet, and canvas selection. The module is deliberately free of React and of the editor, so the canonical page, the embed page (#9427), and the oEmbed function (#9363) all speak one contract instead of re-deriving it. Validation doubles as normalization, equality is defined as "the canonical query strings agree", and the selection vocabulary comes from `@hashintel/petrinaut-core/selection`.
- Per-key search validation runs through zod (already a dependency): string fallbacks are `.optional().catch(undefined)` schemas, and the `"type:id"` selection encoding is a bidirectional `z.codec`.
- `navigation-search.ts` keeps only the projection onto Petrinaut's navigation state. Everything the URL does not carry (mode, Simulate section, drawers) lives in page state through the `useSharedSearchNavigation` hook, which keeps the full location in memory and mirrors only its shared projection to the URL, so controls driving non-shared fields never silently snap back.
- The codec laws are property-tested with fast-check: decoding never throws on generated garbage, `encode(decode(s))` is a fixed point (what the embed redirect relies on), and every decodable state survives encode then decode.
- Review fixes: a rejected example-chunk import evicts the loader cache so a transient failure or stale deploy is retryable without a hard reload; the URL codec round-trips all four editor modes (notebook and actual no longer snap back to edit); the routed demo prunes abandoned empty nets again (a Back entry to a pruned net falls back to the most recent one); the generation script iterates the canonical `exampleSlugs` list instead of its own copy.
- Review fixes (second round): restated `env: ["TEST_COVERAGE"]` on the `test:unit` override, which replaces rather than merges with the root task; `sharedSearchKeys` is module-private, as nothing imported it; hoisted `useSearch` out of a JSX prop position; added a test asserting the truck-fleet rewrite initialises every attribute of the `Truck` and `Conditions` types, so a field added to either fails here instead of simulating with a silent default; dropped the `/examples*` rewrites, now covered by the SPA fallback in #9359.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- New: `catalog.test.ts` (every catalog entry loads, every slug has a catalog entry, every scenario has generated HIR, one cached load per example), `normalize-example.test.ts` (the truck-fleet rewrite fires on the real model; other slugs pass through), `example-search.test.ts`, `example-search.property.test.ts` (fast-check contract laws over generated search inputs: decoding never throws, validation is idempotent, selection survives encode then decode, the canonical string re-decodes to the same location), `navigation-search.test.ts` (state projection), `use-shared-search-navigation.test.tsx` (in-memory fields survive URL merges; only shared changes write the URL), `readonly-example-handle.test.ts`.
- CI: generation runs before every build, typecheck, and test task via the Turbo graph, and fails the task when a model or scenario no longer compiles.

## ❓ How to test this?

1. Checkout the branch and run `yarn workspace @apps/petrinaut-website dev`.
2. Open `/examples/gases-1-pn`, select items, switch subnets, and pick a scenario.
3. Confirm the URL tracks those steps and Back/Forward walk the same states, that mode switching works without touching the URL, and that editing is not possible.


apps/petrinaut-website/src/routes/index.tsx
```diff
@@ -1,7 +1,82 @@
...
+    next: LocalDemoSearch,
+    history: "push" | "replace",
+  ) => {
+    const current = latestSearchRef.current;
+    if (localDemoSearchesMatch(current, next)) {
+      return;
+    }
+    latestSearchRef.current = next;
+    void navigate({ replace: history === "replace", search: next });
+  };
```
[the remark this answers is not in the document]
kube: Fixed as suggested: the demo route now passes search updater functions to `navigate` (matching the example routes), so every navigation applies to the router's own latest search and the optimistic ref is gone. Duplicate suppression for repeated identical navigations lives in the navigation provider; the route tests now assert updater composition and idempotence. Landed in the "route Petrinaut app locations" commit (branch tip 447e7bcf3c).

apps/petrinaut-website/src/main/app/local-storage-demo/local-storage-demo-app.tsx
```diff
@@ -142,11 +163,40 @@ export const LocalStorageDemoApp = () => {
...
+  }, [activeHandle, onNetNavigate, setStoredSDCPNs, storedSDCPNs]);
+
+  useEffect(() => {
+    if (onNetNavigate && initialNet && requestedNetId !== initialNet.id) {
+      onNetNavigate(initialNet.id, {
+        history: "replace",
+        preserveNavigation: requestedNetId === undefined,
+      });
+    }
+  }, [initialNet, onNetNavigate, requestedNetId]);
```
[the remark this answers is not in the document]
kube: Confirmed and fixed: the unknown-net normalization effect now also requires `currentNetId === initialNet.id` — it only fires while the app still shows the net the URL resolved to. A cross-net switch remounts the app via the route's `key`, so the guard holds at every fresh URL resolution, and the in-flight window after pruning the just-left net can no longer replace-navigate to an unrelated recent net. Regression test added (three stored nets; switching away from a pruned empty net triggers exactly one navigation, to the chosen target). Branch tip d16db5d8a3.

apps/petrinaut-website/turbo.json
```diff
@@ -2,15 +2,27 @@
...
       "outputs": ["src/routeTree.gen.ts"]
     },
+    "examples:generate": {
+      "dependsOn": ["^build"],
+      "inputs": [
+        "scripts/generate-example-artifacts.ts",
+        "src/examples/models/**",
+        "src/examples/normalize-example.ts"
+      ],
+      "outputs": ["src/examples/generated/**"]
```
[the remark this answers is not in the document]
kube: Confirmed and fixed: `src/examples/catalog-metadata.ts` is now listed in the `examples:generate` task's `inputs`, so editing the slug list invalidates the cached generate result. Branch tip 9fac113186.

apps/petrinaut-website/turbo.json
```diff
@@ -2,15 +2,28 @@
...
+    },
     "lint:tsc": {
-      "dependsOn": ["codegen", "^build"]
+      "dependsOn": ["codegen", "examples:generate", "^build"]
+    },
+    "test:unit": {
+      "dependsOn": ["codegen", "examples:generate", "^build"]
```
TimDiekmann: `codegen` and `examples:generate` already depends on `^build`

kube: Yes, but is it really a problem though?
`codegen` and `examples:generate` could later not depend on `^build`.
But `test:unit` will always require a build`.

Isn't Turbo supposed to optimize that?
(even if we don't have proper caching for now)

TimDiekmann: No, it's not problematic, just noticed this. And I guess with the [upcoming changes](https://hashintel.slack.com/archives/C02TWBTT3ED/p1788115972338189) this will be changed anyway (I was tagged due to `turbo.json` being changed, a rule we just added today)

apps/petrinaut-website/turbo.json
```diff
@@ -2,15 +2,28 @@
   "extends": ["//"],
   "tasks": {
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["examples:generate", "^build"],
       "outputs": ["dist/**"],
       "cache": false
```
TimDiekmann: May I ask you to drive-by this?

```suggestion
      "outputs": ["dist/**"]
```

TimDiekmann: No worries if this interrupts your stack, I need to touch the files anyway. Just resolve this thread then 🙂

kube: Applied the drive-by: `cache: false` is gone and the `build` task keeps `outputs: ["dist/**"]`. Caching stays correct up the stack because `examples:generate` declares its inputs and outputs, so its hash feeds `build`'s cache key.

apps/petrinaut-website/src/examples/full-example-page.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export const FullExamplePage = ({
+  example,
+  onSearchChange,
+  search,
+}: FullExamplePageProps) => {
+  const handle = getReadonlyExampleHandle(example);
+  const navigation = useSharedSearchNavigation(search, onSearchChange);
+
```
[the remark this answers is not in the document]
kube: Confirmed and fixed: the example route now remounts the page per example (`key={example.catalog.slug}` on `FullExamplePage`), so in-memory mode, scenario, and selection cannot leak across a client-side transition or Back/Forward between examples; the fresh mount re-derives state from the new URL. The embed route got the same key.

apps/petrinaut-website/turbo.json
TimDiekmann: Nice, thanks!

apps/petrinaut-website/.tanstack/tmp/507d46af-61db48957dccd63d09a772feda53370f
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const rootRouteChildren: RootRouteChildren = {
+  IndexRoute: IndexRoute,
+  BrunchRoute: BrunchRoute,
+  OptimizationRoute: OptimizationRoute,
+};
+export const routeTree = rootRouteImport
+  ._addFileChildren(rootRouteChildren)
+  ._addFileTypes<FileRouteTypes>();
```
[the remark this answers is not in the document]
kube: Both parts fixed: the committed `.tanstack/tmp/…` snapshot is removed from the commit that added it, and `.tanstack/` is now ignored. The ignore went in the root `.gitignore` next to the other generated-output patterns rather than the app's, since the directory is router-plugin scratch that any app using the plugin would produce.

CiaranMn (dismissed):

CiaranMn (dismissed):

CiaranMn (approved):

PR #9386, opened 2026-08-27
FE-1526: Enumerate weighted-arc token combinations lazily

kube:
## Summary

Before this PR, `indexCombinations` materialised every k-combination of an input arc's tokens before the first one was used. A transition with a weight-`w` coloured input arc over `n` tokens therefore allocated `C(n, w)` arrays per evaluation per frame. Measured cost was 13.4 ms per run-frame at 400 tokens for a never-firing weight-2 arc.

Generator now enumerates lazily with no per-combination allocation. Order is unchanged, so every seeded trajectory is identical. Measured on `benchmarks/coloured-enumeration.mjs`, extended with an always-fires self-loop case, at 400 tokens:

| Case | Before | After |
| --- | --- | --- |
| Never fires (examines all 79 800 combinations) | 13 771 µs | 2 463 µs (5.6×) |
| Always fires (examines one) | 3 858 µs | 12.6 µs (305×) |

A firing transition's cost is now flat in the token count. Never-firing case keeps its `C(n, w)` lambda evaluations, since each combination's rate must be computed, at ~31 ns per combination instead of ~170 ns.

## Links

- Ticket [FE-1526](https://linear.app/hash/issue/FE-1526)
- Docs [Simulation performance](https://petrinaut-docs-git-cf-fe-1526-enumerate-weighted-arc-tok-148020.stage.hash.ai/architecture/core/simulation/performance)

## Changes

- `enumerate-weighted-markings.ts` enumerates lazily with an in-place successor per arc
  > Odometer over arcs advances a lexicographic successor in place; nothing is allocated per combination.
  > Yielded marking is a reused buffer.
  > Module doc comment states the contract and both engines already copy on accept.
- Eager `enumerateWeightedMarkingIndices` is reimplemented over the lazy generator
  > One source of truth for the order.
- Enumeration order is unchanged
  > Engine fires the first passing combination, so an order change would diverge every seeded trajectory.
  > `webgpu/pair-selection.ts` reproduces the same order by unranking.

## Test coverage

- `enumerate-weighted-markings.test.ts`:
  > New test pins the sequence against the previous algorithm across a case matrix, keeping that algorithm in the test as the ordering oracle. Another test pins the buffer-reuse contract.
- `webgpu/pair-selection.test.ts`:
  > GPU pair-order tests pass unchanged. Their collection helper now copies per yield, as the contract requires.

## How to test

- `turbo run build test:unit --filter '@hashintel/petrinaut-core'`
- `node libs/@hashintel/petrinaut-core/benchmarks/coloured-enumeration.mjs`
- Compare never-fires and always-fires rows at 400 tokens with table above
- Run any seeded simulation with weighted coloured input arc on this branch and its base
- Expect identical trajectories

### DIFF
--- a/.changeset/lazy-weighted-arc-enumeration.md
+++ b/.changeset/lazy-weighted-arc-enumeration.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut-core": patch
+---
+
+Weighted-arc token combinations now enumerate lazily, in place, in the same lexicographic order. Evaluating a transition with a weight-2 coloured input arc no longer materialises every `C(n, 2)` combination per frame: measured, a firing transition at 400 tokens in the place drops from 3.86 ms to 12.6 µs per run-frame, and a never-firing one from 13.8 ms to 2.5 ms. Trajectories are unchanged for every seed.
--- a/.changeset/lazy-weighted-arc-enumeration.md
+++ b/.changeset/lazy-weighted-arc-enumeration.md
@@ -2,4 +2,4 @@
 "@hashintel/petrinaut-core": patch
 ---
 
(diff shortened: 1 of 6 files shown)


YannisZa (dismissed):

TimDiekmann (dismissed):
re-approval of pre-rebase

TimDiekmann (approved):

TimDiekmann (approved):

PR #9387, opened 2026-08-27
FE-1530: Restyle Petrinaut scrollbars

kube:

Petrinaut's scroll containers showed platform-default scrollbars: a gray track and a wide thumb, always visible on Windows and on macOS whenever a mouse is connected.

Every native scrollbar under `.petrinaut-root` now shows a rounded 6px thumb centered in a 10px transparent gutter. The thumb is hidden until the pointer is over its scroll container, fades in and out over 200ms, and darkens while hovered or dragged. Firefox falls back to thin standard scrollbars in the same colors, with the same hover reveal. Portalled content (drawers, dialogs, popovers, select lists) gets the same treatment because it mounts inside `.petrinaut-root`.

## 🔗 Related links

- [FE-1530: Restyle Petrinaut scrollbars](https://linear.app/hash/issue/FE-1530/restyle-petrinaut-scrollbars) _(internal)_
- [H-3062: Improve our scrollbars app-wide](https://linear.app/hash/issue/H-3062/improve-our-scrollbars-app-wide) _(internal)_ — this PR covers Petrinaut; app-wide scrollbars stay tracked there.

## 🔍 What does this change?

- The Petrinaut Panda preset (`libs/@hashintel/petrinaut/src/panda-preset.ts`) adds `.petrinaut-root`-scoped `globalCss`: `::-webkit-scrollbar-*` rules for Chromium and Safari, and `scrollbar-width: thin` plus `scrollbar-color` for Firefox behind `@supports (-moz-appearance: none)`. The rules land in Panda's `base` layer, so component utilities still override them. Both build pipelines emit them: `hash-frontend`'s Panda config imports the preset, and Petrinaut's own stylesheet compiles from it.
- Scrollbar pseudo-elements do not run CSS transitions, so the thumb color routes through `--petrinaut-scrollbar-thumb`, a registered custom property (`@property`, declared via the preset's `globalVars`) that the scroll container animates and the thumb paints each frame.
- Chromium and Safari ignore `::-webkit-scrollbar-*` styling on any element whose `scrollbar-width` is not `auto`. The declarations that did this are removed: the sidebar's `scrollbarWidth: "[thin]"` in `vertical-sub-views-container.tsx`, and the `@hashintel/ds-components` overlay-body (drawer, dialog, popover), selectable-list, and textarea recipes.
- Both packages get a patch changeset.

This PR:

The changes in this PR:

  - the Petrinaut user guide does not mention scrollbars and contains no screenshots

The changes in this PR:

## ⚠️ Known issues

- On macOS with "Show scroll bars: When scrolling", scrollbars previously overlaid content and took no layout space. The styled scrollbars reserve a 10px gutter whether or not the thumb is shown, which keeps panel layout identical across platforms.
- The reveal is pointer-driven: scrolling with the keyboard does not show the thumb.

## 🛡 What tests cover this?

- No new automated tests. `panda.config.shared.test.ts` checks the preset's shape against Panda's `Preset` type, which now covers the `globalCss` and `globalVars` additions.

## ❓ How to test this?

1. Run `yarn workspace @hashintel/petrinaut dev` and open the `Petrinaut` → `Handle Spike With Sir` story, or load an example from the burger menu.
2. Move the pointer over the sidebar's Nodes list: the thumb fades in over 200ms and fades out when the pointer leaves. Drag it: it darkens.
3. Switch to Simulate and open "Create scenario" at a small window height: the drawer body scrolls with the same thumb. This is the `ds-components` half of the change.
4. Open a select with a long list and a textarea with overflowing content: same treatment.


alex-e-leon:
@kube - I like the new scrollbar look, I think it's a big improvement, but I think there are still a couple of gaps with this PR.

1) I think the new scrollbar styles should be applied in the DS directly instead of just in petrinaut, so that the UI stays consistent and so that changes to the DS get tested/reviewed with the new scrollbars (for ex keeping spacing balanced)

2) We should bring back the thin scrollbars in textarea + selectable list (but in the new custom styles). They should still have a fallback to thin for firefox etc where scrollbar styling isn't supported.

3) IMO we should still continue to respect users who have selected to always show scrollbars. It's not the default in any OS anymore as far as I'm aware, so if this setting is on it's because a user has explicitly decided they prefer to see scrollbars. There's no css for checking this, but we can do a quick js measurement  and then apply a global class on :root element, and then if the root class is present just remove the hide on hover css.

Also happy to take this PR over and apply my own recommendations if you're busy

kube:
@alex-e-leon 
> 1. I think the new scrollbar styles should be applied in the DS directly instead of just in petrinaut, so that the UI stays consistent and so that changes to the DS get tested/reviewed with the new scrollbars (for ex keeping spacing balanced)
> 2. We should bring back the thin scrollbars in textarea + selectable list (but in the new custom styles). They should still have a fallback to thin for firefox etc where scrollbar styling isn't supported.
> 3. IMO we should still continue to respect users who have selected to always show scrollbars. It's not the default in any OS anymore as far as I'm aware, so if this setting is on it's because a user has explicitly decided they prefer to see scrollbars. There's no css for checking this, but we can do a quick js measurement  and then apply a global class on :root element, and then if the root class is present just remove the hide on hover css.

Agree for 1 and 2. (2 could come in a second step though, it's up to you)

I do not agree on 3 though. I ended up having these ugly scrollbars appearing, just because I had a mouse plugged, and the browser or OS decided by itself to show them.

Modern UIs all have their own scrollbars, and we should not be designing for our grandmas.

What most apps do nowadays is that they suggest scrollability with a shadow/gradient/blur.

alex-e-leon:
@kube The only situation where a scrollbar always shows by default is on OSX if a third party mouse with a scroll wheel is plugged in (it doesn't happen for trackpads or with a magic mouse), and can easily be changed by a system setting. in every other common case (on windows, most linux distros + osx with a magic mouse or no mouse) the default setting is for scrollbars to auto-hide.

But the ability to always have scrollbars showing is important for accessability - some users need this.

And doesn't it annoy you on other websites as well? Why don't you change your system setting?

alex-e-leon:
@kube - I've updated with my suggested changes, so now would be good to get your review in reverse. Here's a summary of the changes, including a few new changes that came out of testing:

- moved the scrollbar logic + ui into the DS applied to everything under whatever scope you set in pandas preset
- check the users scrollbar settings and honor always showing scrollbars if the user has set it, but use the new custom ui for scrollbars if they do instead of the os/browser default
- re-implement thin scrollbars in the new scrollbar ui for textarea + selectable list
- handle fallbacks for firefox + IOS (which doesn't support custom scrollbars) + workaround OSX safari scroll rendering bugs
- for users with auto-hide scroll on: darken the scrollbar on scroll as well as hover, otherwise the scrollbars never show if a user scrolls with the keyboard

I've tested it safari/firefox/chrome on OSX, and safari on IOS and all behave nicely now - if I had a windows box I'd test there too but I don't. LMK if you have any feedback or find any other issues.

alex-e-leon:
Superceded by https://github.com/hashintel/hash/pull/9815

PR #9413, opened 2026-08-28
BE-774: Instrument the Graph API rate limiter

TimDiekmann:

BE-758 shipped the Graph API rate limiter with logs as its only signal, so its decisions are not measurable: the budgets are estimates, observe mode reports what it would deny only per request in the log stream, and the paths that pass a request unchecked aggregate nowhere. This PR instruments the limiter and the authentication middleware with OpenTelemetry metrics, and fixes the registration gap that made every metric a silent no-op: `TelemetryRegistry` built an `SdkMeterProvider` but never called `set_meter_provider`.

## 🔗 Related links

- BE-774
- Builds on #9402 (BE-796) and #9400 (BE-793)

## 🚫 Blocked by

- [ ] #9400
- [ ] #9402

## 🔍 What does this change?

- `hash-telemetry`: `init_global` registers the meter provider globally; the scoped `init` deliberately never does (it would replace the registration and shut the provider down on drop — the setup the Graph benches run). The opaque guard became a public `#[must_use]` `Telemetry` type with `meter(scope)`; without an OTLP endpoint it holds a reader-less provider and warns once that every measurement is discarded.
- `hash-graph`: `block_on` takes the subcommand as a closure over the initialized `Telemetry`, so no instrument can be bound before the telemetry exists. The embedded admin server gets its own meter scope, keeping its rejections a separate series from the public API's, as running it standalone would.
- `hash-middleware` rate limiter, on the meter the state is built with:
  - `hash.rate_limit.decisions` {limiter, outcome=allowed|denied|would_deny} — configured budgets can be compared against real traffic before enforcement is switched on
  - `hash.rate_limit.unchecked` {stage, reason} and `hash.rate_limit.address_fallbacks` {reason} — the "has the limiter become a no-op" question
  - `hash.rate_limit.misconfigurations` {reason} — the internal-error paths where a route is wired without the middleware the principal limiter builds on
  - `hash.rate_limit.tracked_keys` gauge (read from the stores directly), `hash.rate_limit.maintenance_runs`, and `hash.rate_limit.evicted_keys` — a lost maintenance task shows as a flat run count next to a climbing key count
  - `hash.rate_limit.denial_wait` histogram (seconds, explicit boundaries) — separates a burst-edge denial from an exhausted budget
  - `hash.rate_limit.mode` gauge — keeps a `denied` alert honest when the mode flips to observe
- `hash-middleware` authentication: `hash.authentication.rejections` {`http.response.status_code`, `fault_domain`} counted at the single rejection-response builder; `ResolvedAuthentication` carries the metrics handle so the extractor's rejections count too — on the REST chain, the missing-credentials 401 arises almost only there.
- `RateLimiters::start(config, meter) -> Arc<Self>` replaces `new` + `spawn_maintenance`: the gauge registration and the maintenance task cannot be forgotten or doubled. The label vocabulary moved into enums (`Outcome`, `Stage`, …), so a typo cannot silently fork a time series.
- The `Tally` struct and its 60-second interval report are removed: they double-booked the same counters at lower resolution. The per-event debug logs stay — they carry the key and header detail too wide for a metric label.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Without an OTLP collector there is no aggregated signal any more, since the interval report is gone; the startup warning names the gap, and the per-event debug logs remain. Accepted because production deployments run a collector, and re-adding a log aggregate would reintroduce the double bookkeeping.
- The two extractor paths answering 500 on a route wired without the authentication middleware count nowhere: the metrics handle lives inside the extension that is missing in exactly that case. Stated on `AuthenticationMetrics`; the three equivalent paths in the rate limiter do count.

## 🐾 Next steps

- BE-755 decides the credential-verification cache on the data these metrics produce.

## 🛡 What tests cover this?

- `hash-middleware` unit tests assert the instruments through an in-memory exporter: every decision outcome and limiter, both unchecked reasons at both stages, all three address-fallback reasons, all three misconfiguration reasons, the wait-time histogram's unit, the mode gauge, eviction counts, and the exact label key set of the decisions series.
- An exhaustive table test drives every `AuthenticationError` variant through the rejection counter and stops compiling when a variant is added.
- A drop test pins that the gauge callback and the maintenance task hold the limiter state weakly.
- `hash-telemetry` gains its first test: a scoped `init` must leave a globally registered meter provider recording.

## ❓ How to test this?

1. Start the graph with `HASH_OTLP_ENDPOINT` pointing at a collector (the compose stack provides one).
2. Send requests with and without credentials, and past a budget's burst.
3. Confirm the `hash.rate_limit.*` and `hash.authentication.rejections` series arrive with the labels above.

## 📹 Demo

Not applicable — observability change without UI.


apps/hash-graph/src/subcommand/server.rs
```diff
@@ -514,13 +516,18 @@ pub(crate) struct AuthenticationSetup {
     pub service_secret: String,
 }
 
+#[expect(
```
indietyp, line 519: Maybe we can move these into a struct instead - just a bag of types? Because 8 is a lot to keep track of, even if they're positional arguments with distinct types. It makes calling the function harder than it needs to be.

TimDiekmann: We should do that, but I think that's worth a dedicated PR, we can sweep all CLI invocations then and use a similar interface.

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -34,58 +39,104 @@ use self::{
...
-fn rejection(report: &Report<AuthenticationError>) -> Response {
+fn rejection(metrics: &AuthenticationMetrics, report: &Report<AuthenticationError>) -> Response {
     let error = report.current_context();
+    metrics.record_rejection(error);
     error_response(error.status_code(), error.client_message().to_owned())
 }
 
 /// Reports a rejection and converts it into the response returned to the client.
-fn logged_rejection(error: AuthenticationError) -> Response {
+fn logged_rejection(metrics: &AuthenticationMetrics, error: AuthenticationError) -> Response {
```
indietyp, line 113: Non-blocking: would it make sense to make this an inherent method on `AuthenticationMetrics` (and `rejection`), so it's easier to call / make the intent is clearer? (or I guess actually on `AuthenticationError::rejection(&metrics)` would make more sense.

apps/hash-graph/src/subcommand/server.rs
TimDiekmann: I created BE-810

indietyp (dismissed):

libs/@local/middleware/src/authentication/mod.rs
TimDiekmann: Will do in the follow up, good flag!

hashdotai (commented):
Re-approval due to GitHub bug

PR #9434, opened 2026-08-29
FE-1549: Rotate sweep parameter draws by a seed-derived shift

kube:

Range sweeps drew each run's parameter values from a bare per-axis Halton sequence: every experiment explored the identical value sequence regardless of its seed, and deterministic low-discrepancy points are not exchangeable random samples. This rotates each axis's draws by a seed-derived shift (Cranley–Patterson), making them unbiased over the seed while keeping the low-discrepancy convergence, the even early coverage, and prefix stability. Stacked on #9419.

## 🔗 Related links

- [FE-1549](https://linear.app/hash/issue/FE-1549/randomize-sweep-parameter-draws-with-a-seed-derived-shift) _(internal)_
- Base PR: #9419

## 🔍 What does this change?

- `sweepRunFraction` takes the experiment seed and rotates the radical-inverse fraction by a per-axis shift hashed from the seed and the axis index (32-bit `Math.imul` hash), modulo 1.
- `sweepRangeRuns` threads the seed through; the sweep session passes its experiment seed. A draw depends only on the seed, the axis, and the global run index, so ladder batches still extend the same sequence and the selection cache stays valid.
- Tests: prefix stability, unit-interval spread across seeds, per-axis distinctness, exact reproduction for one seed, distinct sequences across seeds.
- Docs: the user-guide sweep section and the parameter-sweeps architecture page describe the seed-shifted draws.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `parameter-grid.test.ts` (`sweepRunFraction` suite, extended) and the existing `sweep-session.test.ts` range-draw assertions, which now exercise the seeded path.

## ❓ How to test this?

1. Run `yarn workspace @hashintel/petrinaut dev` and open "Simulate / SweepNavigator / Real compute on CPU".
2. Select a range: the streamed distribution covers the interval from the first rung, as before.
3. Create two sweeps with different seeds over the same range: their per-run draws differ; re-running the same seed reproduces them.

### DIFF
--- a/.changeset/seeded-sweep-draws.md
+++ b/.changeset/seeded-sweep-draws.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+Range sweeps rotate each axis's low-discrepancy parameter draws by a seed-derived shift (Cranley–Patterson), so draws are unbiased over the seed and experiments with different seeds explore different value sequences — prefix stability and the selection cache are unaffected.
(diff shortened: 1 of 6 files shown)


kube:
GitHub auto-marked this merged when a stack-link mishap put its base on the branch above it, whose rebase then contained this head. Superseded by #9435, which sits in the correct stack position (#9419 ← #9435 ← #9421).

PR #9437, opened 2026-08-29
BE-802: Load programmatic defaults in hash-config

TimDiekmann:

`hash-config` gets its first working slice. A binary builds its configuration type from programmatic defaults with `Loader::new().with_defaults(…).load::<Config>()`, which returns the deserialized configuration. Figment does the merging and appears in no public signature.

The error path is the substantive part of this change. Figment's own messages inline the value they rejected, so a bad `store.port` reads `invalid type: found string "hunter2", expected u16`. A configuration value can be a database password or an API key, which makes such an error unsafe to put in a log. Here the error names the key and the shape that was expected, and never the value.

## 🔗 Related links

- BE-802 _(internal)_
- H-4170 _(internal)_ — the parent, tracking the remaining slices
- #9333 — BE-801, the scaffold this builds on

## 🚫 Blocked by

Nothing. BE-801 is merged.

## 🔍 What does this change?

- `Loader::with_defaults` accepts any `impl Serialize` that serializes to a map and merges it into the default layer. Repeated calls apply in order: later scalars and arrays replace earlier ones, and maps merge recursively. BE-802 calls this method `defaults`; it is `with_defaults` here to match the builder convention the rest of the workspace uses, including the layered builder in `libs/@local/telemetry/src/lib.rs`.
- `Loader::load` deserializes the merged values into the caller's type and returns `Report<LoadError>` on failure.
- Errors carry the key path. Figment supplies it on its error type and the diagnostic discarded it, so every failure read as "something is a string" with no way to tell which of a hundred settings was meant. A missing field joins with its parent, so it reads `missing field \`store.port\`` rather than `missing field \`port\``.
- Errors carry Serde's `expected` text. `Visitor::expecting` receives no value, so it cannot quote one.
- Configuration keys count as schema and are reported, map keys and array indices included. Scalars, enum variant names, lengths and hand-written `serde::de::Error::custom` messages count as values and are dropped.
- A private provider wraps `Serialized` so a diagnostic names the layer `defaults` instead of the Rust type it was handed, and so keys render dotted instead of prefixed with the profile they were found under.
- `Loader` implements `Debug` by hand and lists the layer names. Figment's derived `Debug` renders the merged values, which would put configuration values into any log line that formats a loader.
- The crate is split into `lib.rs` for the `Loader`, `error.rs` for the error type and the diagnostic translation, and `defaults.rs` for the provider. Each later source is a provider plus a `Loader` method and gets a module of its own.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

Serde stops at the first error within a single document, so a configuration with three mistakes reports them one run at a time. More than one diagnostic appears only when several sources each fail to produce data at all. This is Serde's behaviour rather than something the loader chooses, and reporting every mistake in one pass would mean writing a deserializer of our own.

`LoadError` has a single variant, so a caller can render a report but cannot branch on what went wrong; the categories live in a private type. That is enough while there is one failure mode. Adding variants is a breaking change for callers, and doing it now would mean guessing at the joints before the file, environment and override layers exist.

`unknown field` only fires for configuration types that set `#[serde(deny_unknown_fields)]`. Without it, Serde ignores an unrecognised key and the field takes its default, so a typo passes silently. Closing that properly needs the schema walk, which can compare a document against the schema whether or not the attribute is set.

## 🐾 Next steps

The remaining slices under H-4170: file sources with profiles and the directory cascade, the environment layer, `-C` overrides with the clap integration, the schema walk, and the `config` subcommand. `set_default(key, value)` for seeding a single subtree belongs to the defaults layer and is not in this PR.

## 🛡 What tests cover this?

- `libs/@local/config/tests/defaults.rs` covers the merge contract and one case per diagnostic category, each asserting both halves of the contract: the key appears, the value does not.
- `libs/@local/config/examples/defaults.rs` is the runnable example and carries two tests of its own. `test = true` on the `[[example]]` target is what makes them run, and `doc-scrape-examples = true` puts the example on the `Loader::new`, `with_defaults` and `load` doc pages.
- The doctest on `Loader`.

## ❓ How to test this?

1. Check out the branch.
2. `cargo nextest run --all-features --package hash-config`
3. `cargo run --all-features --package hash-config --example defaults`

## 📹 Demo

The example composes two default layers and then sets the port to a password:

```
localhost:6543 serving ["api", "health"]

the configuration could not be loaded
├╴at libs/@local/config/examples/defaults.rs:55:10
╰╴invalid type: found string, expected u16 at `store.port` in `defaults` (libs/@local/config/examples/defaults.rs:54:10)
```


libs/@local/config/src/error.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn from(kind: FigmentKind) -> Self {
+        match kind {
+            // A custom message is written by hand and may quote the value it rejected.
+            FigmentKind::Message(_) => Self::Opaque,
+            FigmentKind::InvalidType(actual, expected) => Self::InvalidType {
+                actual: actual_kind(&actual),
+                expected: expected.into_boxed_str(),
+            },
+            FigmentKind::InvalidValue(actual, expected) => Self::InvalidValue {
+                actual: actual.actual_kind(&actual),
```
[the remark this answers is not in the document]
TimDiekmann:
```suggestion
                actual: actual_kind(&actual),
```

libs/@local/config/src/error.rs
indietyp, line 1: Not saying we should do this now, but we do have the data for it - would be extremely cool to use annotate-snippets to show the exact location with the error.

indietyp (approved):

libs/@local/config/src/error.rs
TimDiekmann: Yeah, absolutely!

PR #9445, opened 2026-08-29
SRE-986: Dissolve the graph type-defs crate into the graph API

TimDiekmann:

`libs/@local/graph/type-defs` generated Rust from two TypeScript files by shelling out to `yarn tsx` and `quicktype` from `build.rs`, so building the Graph needed a working yarn install. Of that output only `status_payloads.rs` was included at all, and of its three structs only `ErrorInfo` was ever constructed.

Its only consumers are `rest/json.rs` and `rest/entity_type.rs`, and `rest/status.rs` already owns the conversion into responses. The crate moves there and disappears.

Builds on SRE-983. Last crate directory carrying a hand-maintained TypeScript surface, which is what SRE-984 was waiting on.

## 🔍 What does this change?

- `ErrorInfo` is hand-written in `rest/status.rs`, keeping its field names and its `HASH Graph` domain default. `RequestInfo` and `ResourceInfo` are gone — nothing ever constructed them
- The `StatusPayloadInfo` wrapper is gone with them, so the error body changes from `{"contents":[{"Error":{…}}]}` to `{"contents":[{…}]}`. That is what `json_schemas/status.json` already described; the externally tagged enum was the deviation. The schemas narrow to the payload that remains
- `openapi.json` itself does not change — it references `./models/status.json`, and the generator emits no client model for it
- The backend integration test builds its `GraphStatus` from `@local/status`, which is where the crate's TypeScript re-exported it from anyway

## 🐾 Next steps

SRE-984. With this merged no crate directory carries a JavaScript surface, so the remaining `package.json` files are pure generated wiring.

`metadata` stays on `ErrorInfo` for now: five call sites put the serde error text, the rejected entity type ID and the colliding base URL there, and none of that appears in `message`. H-3009 wants to add report information to it. Replacing it is a separate change.

## 🛡 What tests cover this?

The backend integration tests exercise the admin endpoints whose responses carry this status shape.

## ❓ How to test this?

1. `cargo clippy --all-features -p hash-graph-api`
2. `turbo run codegen:generate-openapi-specs --filter '@rust/hash-graph-api' && git diff --exit-code libs/@local/graph/api/openapi/openapi.json`
3. `mise run sync:turborepo && git diff --exit-code '**/package.json'`


libs/@local/graph/type-defs/build.rs
```diff
@@ -1,67 +0,0 @@
...
-    let type_defs_path = crate_dir
-        .join(TYPE_DEFS_PATH)
-        .canonicalize()
-        .expect("Failed to find HASH Graph type-defs path");
-
-    println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed={}", codegen_package_path.display());
-    println!("cargo:rerun-if-changed={}", type_defs_path.display());
-
-    if !std::process::Command::new("yarn")
```
indietyp, line 39: I am so happy that this is gonna go 😊 rust-analyzer always had trouble finding yarn

indietyp (approved):

libs/@local/graph/type-defs/build.rs
TimDiekmann: It was the most painful thing we had in the rust toolchain, yes.

vilkinsons (approved):

PR #9469, opened 2026-08-31
BE-809: hgraph: type middleware authentication rejections by their error

indietyp:

`hash-middleware` runs authentication and rate limiting as `async fn`s behind `axum::middleware::from_fn`. A mount site captures its state in a closure and clones an `Arc` per layer per request, and a middleware that rejects a request renders the `Response` itself, so the reason a request failed is unavailable to any outer layer.

After this change each one is a `tower::Layer` with a service of its own. State lives in the service, so construction at a mount site is a struct literal. A rejection travels as a typed value, because the service response type is `Result<S::Response, Rejection>` and each rejection renders through `IntoResponse` at the router's edge.

Reporting a failure also stops depending on the function that handles it. The report state lives on `AuthenticationError`, so `AuthenticationError::log` claims a latch and returns early once some site has claimed it. `resolve_request_actor` reports where the provider rejects, which includes the verified rejection that degrades to anonymous and reaches no render at all, and `AuthenticationRejection::into_response` reports where a rejection reaches the client. One log line per failure follows from those two sites rather than from a convention each middleware kept, and the level follows the error's fault domain.

The alternative was to keep `from_fn` and pass the rejection out through a request extension. An extension leaves the response type erased and the per-request `Arc` clones in place, so an outer layer has nothing to match on and the typing would have been a convention rather than a signature.

What to verify:

- The rejection belongs to the service signature, so an outer layer matches a value instead of inspecting a rendered response.
- Each service passes `poll_ready` through to the service it wraps, and a rejecting `call` returns without touching that service at all.
- Layer order holds at both mount sites, with the ip budget outermost and the principal budget behind authentication.
- The crate-level doctest in `src/lib.rs` states the construction contract.

## 🔍 What does this change?

Each service names its own future through `impl_trait_in_assoc_type` rather than boxing it. `AuthenticationError` becomes a struct over `AuthenticationErrorKind` with a const constructor per kind and `Deref` to the kind, which is what lets the call sites in `hash-graph-authentication` construct a failure by name.

## 🛡 What tests cover this?

The existing `hash-middleware` unit tests cover the layers through their routers. The rewrite updated construction in `src/rate_limit/tests.rs` and moved no assertion, so the same expectations hold against the new types. The crate-level doctest exercises a router built from the layer literals, and `libs/@local/graph/authentication/tests/kratos.rs` follows the new constructors.

## ❓ How to test this?

1. `cargo nextest run -p hash-middleware` and `cargo test --doc -p hash-middleware`.
2. Read `src/authentication/mod.rs` from `AuthenticationRejection` outward, then `src/rate_limit/mod.rs`. Both follow one shape, so the first explains the second.
3. Against a running graph API, send a request with no credential, one with a malformed credential, and one over the principal budget. Confirm the status code and client message of each, and one log line per rejected request.

This PR:

The changes in this PR:

The changes in this PR:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/libs/@local/graph/api/src/rest/mod.rs
+++ b/libs/@local/graph/api/src/rest/mod.rs
@@ -71,7 +71,13 @@ use hash_graph_temporal_versioning::{
 };
 use hash_graph_type_fetcher::TypeFetcher;
 use hash_graph_types::Embedding;
-use hash_middleware::authentication::provider::{AuthenticationProvider, Caller};
+use hash_middleware::{
+    authentication::{
+        AuthenticationLayer,
+        provider::{AuthenticationProvider, Caller},
+    },
+    rate_limit::{IpGateLayer, PrincipalLimitLayer},
+};
 use hash_status::Status;
 use hash_temporal_client::TemporalClient;
 use include_dir::{Dir, include_dir};
@@ -581,40 +587,25 @@ fn attach_request_middlewares<P, C>(
 ) -> Router
 where
     P: AuthenticationProvider<C> + Send + Sync + 'static,
-    C: Caller + Send + 'static,
+    C: Caller + Send + Sync + 'static,
 {
-    let auth_secret = Arc::clone(&service_secret);
-    let gate_secret = Arc::clone(&service_secret);
-    let gate_limiters = Arc::clone(&rate_limiters);
-
     routes
-        .route_layer(axum::middleware::from_fn(move |request, next| {
-            rate_limit::principal_limit_middleware(
-                Arc::clone(&rate_limiters),
-                Arc::clone(&service_secret),
-                request,
-                next,
-            )
-        }))
-        .route_layer(axum::middleware::from_fn(move |request, next| {
-            auth::authentication_middleware::<_, C>(
-                Arc::clone(&provider),
-                Arc::clone(&auth_secret),
-                Arc::clone(&authentication_metrics),
-                auth::is_bootstrap_route,
-                request,
-                next,
-            )
-        }))
+        .route_layer(PrincipalLimitLayer {
+            limiters: Arc::clone(&rate_limiters),
+            service_secret: Arc::clone(&service_secret),
+        })
+        .route_layer(AuthenticationLayer {
+            provider,
+            service_secret: Arc::clone(&service_secret),
+            metrics: authentication_metrics,
+            bootstrap_route: auth::is_bootstrap_route,
+            caller: core::marker::PhantomData::<C>,
+        })
         .merge(unauthenticated)
-        .layer(axum::middleware::from_fn(move |request, next| {
-            rate_limit::ip_gate_middleware(
-                Arc::clone(&gate_limiters),
-                Arc::clone(&gate_secret),
-                request,
-                next,
-            )
-        }))
+        .layer(IpGateLayer {
+            limiters: rate_limiters,
+            service_secret,
+        })
 }
 
 /// A [`Router`] that serves all of the REST API routes, and the `OpenAPI` specification.
--- a/libs/@local/graph/api/src/rest/mod.rs
+++ b/libs/@local/graph/api/src/rest/mod.rs
@@ -594,12 +594,12 @@ where
             limiters: Arc::clone(&rate_limiters),
             service_secret: Arc::clone(&service_secret),
         })
-        .route_layer(AuthenticationLayer {
+        .route_layer(AuthenticationLayer::<_, C> {
             provider,
             service_secret: Arc::clone(&service_secret),
             metrics: authentication_metrics,
             bootstrap_route: auth::is_bootstrap_route,
-            caller: core::marker::PhantomData::<C>,
+            caller: core::marker::PhantomData,
         })
         .merge(unauthenticated)
         .layer(IpGateLayer {
--- a/libs/@local/middleware/Cargo.toml
+++ b/libs/@local/middleware/Cargo.toml
@@ -25,6 +25,7 @@ clap = { workspace = true, optional = true, features = ["derive"] }
 # Private third-party dependencies
 bytes                              = { workspace = true }
 derive_more                        = { workspace = true, features = ["display", "error"] }
+futures.workspace                  = true
 governor                           = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true, features = ["semconv_experimental"] }
 serde                              = { workspace = true, features = ["derive"] }
--- a/libs/@local/middleware/Cargo.toml
+++ b/libs/@local/middleware/Cargo.toml
@@ -23,7 +23,6 @@ tower-service = { workspace = true, public = true }
 clap = { workspace = true, optional = true, features = ["derive"] }
 
 # Private third-party dependencies
-bytes                              = { workspace = true }
 derive_more                        = { workspace = true, features = ["display", "error"] }
 futures                            = { workspace = true }
 governor                           = { workspace = true }
--- a/libs/@local/middleware/Cargo.toml
+++ b/libs/@local/middleware/Cargo.toml
@@ -25,7 +25,7 @@ clap = { workspace = true, optional = true, features = ["derive"] }
 # Private third-party dependencies
 bytes                              = { workspace = true }
 derive_more                        = { workspace = true, features = ["display", "error"] }
-futures.workspace                  = true
+futures                            = { workspace = true }
 governor                           = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true, features = ["semconv_experimental"] }
 serde                              = { workspace = true, features = ["derive"] }
--- a/libs/@local/middleware/src/authentication/mod.rs
+++ b/libs/@local/middleware/src/authentication/mod.rs
@@ -113,6 +113,7 @@ impl AuthenticationMetrics {
 ///
 /// A rejection records itself when it drops, whether or not a response was rendered from it,
 /// so no holder logs or counts one.
+#[derive(Clone)]
 pub enum AuthenticationRejection {
     /// The credentials did not resolve to a caller the route admits.
     Authentication {
--- a/libs/@local/middleware/src/authentication/mod.rs
+++ b/libs/@local/middleware/src/authentication/mod.rs
@@ -7,16 +7,18 @@
 //! chain's [`Caller`] type: a chain over [`ActorId`] rejects it, a chain over `Option<ActorId>`
 //! verifies it as anonymous.
 //!
-//! Bootstrap routes — the paths the service names through the layer's predicate — require the
-//! service secret regardless of the chain and pass even when no actor resolves. Routes that take
-//! no actor at all are gated by [`ServiceSecretLayer`] instead.
+//! Bootstrap routes, the paths the service names through the layer's predicate, require the
+//! service secret regardless of the chain. With the secret presented they tolerate exactly one
+//! failure: a verified service credential that names no delegated actor. Every other rejection
+//! still fails the request. Routes that take no actor at all use [`ServiceSecretLayer`]
+//! instead.
 
 pub mod provider;
 pub mod request;
 pub mod service_secret;
 
 use alloc::sync::Arc;
-use core::{convert::Infallible, future, marker::PhantomData, task};
+use core::{future, marker::PhantomData, task};
 
 use axum::{
     extract::{FromRequestParts, OptionalFromRequestParts},
@@ -39,11 +41,30 @@ use self::{
 };
 use crate::response::error_response;
 
-/// Instruments recording authentication rejections.
+/// How a request proceeded although its credential resolution failed.
+#[derive(Copy, Clone)]
+enum Degradation {
+    /// A verified rejection resolved as anonymous on a chain serving anonymous callers.
+    Anonymous,
+    /// A bootstrap route admitted the service credential without its delegated actor.
+    Bootstrap,
+}
+
+impl Degradation {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::Anonymous => "anonymous",
+            Self::Bootstrap => "bootstrap",
+        }
+    }
+}
+
+/// Instruments recording credential failures.
 ///
 /// A route wired without the middleware answers with an internal error that is not counted here.
 pub struct AuthenticationMetrics {
     rejections: Counter<u64>,
+    degradations: Counter<u64>,
 }
 
 impl AuthenticationMetrics {
@@ -54,6 +75,11 @@ impl AuthenticationMetrics {
             rejections: meter
                 .u64_counter("hash.authentication.rejections")
                 .with_description("Requests rejected for their credentials")
+                .with_unit("{request}")
+                .build(),
+            degradations: meter
+                .u64_counter("hash.authentication.degradations")
+                .with_description("Requests served although their credential failed")
                 .with_unit("{request}")
                 .build(),
         }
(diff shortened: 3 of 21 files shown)


libs/@local/middleware/src/authentication/request.rs
```diff
@@ -191,21 +194,187 @@ impl AuthenticationError {
...
+        Self::new(AuthenticationErrorKind::StoreError)
+    }
+
+    /// Logs the report at the level its fault domain assigns.
+    ///
+    /// Service faults log as errors, operator faults as warnings, and caller faults at debug: a
+    /// caller-repairable rejection must not hand anonymous traffic the log volume. Each error
+    /// logs once: a later call on a report whose error already logged does nothing, so every
+    /// layer that sees a rejection may call this without duplicating the record.
+    pub fn log(report: &Report<Self>) {
```
TimDiekmann: This was a crate-private method before (`fn log_rejection`). Why is it needed to be exposed? If crate-external callers call this, it modifies the internal state and may omit later logging statements.

```suggestion
    /// Logs the report at the level its fault domain assigns.
    ///
    /// Service faults log as errors, operator faults as warnings, and caller faults at debug: a
    /// caller-repairable rejection must not hand anonymous traffic the log volume. Each error
    /// logs once: a later call on a report whose error already logged does nothing, so every
    /// layer that sees a rejection may call this without duplicating the record.
    pub(crate) fn log(report: &Report<Self>) {
```

TimDiekmann: Also, I found `log_rejection` clearer.

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -1,30 +1,30 @@
...
 //! chain's [`Caller`] type: a chain over [`ActorId`] rejects it, a chain over `Option<ActorId>`
 //! verifies it as anonymous.
 //!
-//! Bootstrap routes — the paths the service names through the middleware's predicate — require
-//! the service secret regardless of the chain and pass even when no actor resolves. Routes that
-//! take no
-//! actor at all are gated by [`service_secret_middleware`] instead.
+//! Bootstrap routes — the paths the service names through the layer's predicate — require the
+//! service secret regardless of the chain and pass even when no actor resolves. Routes that take
+//! no actor at all are gated by [`ServiceSecretLayer`] instead.
```
TimDiekmann: That's not what the code is doing, it tolerates only `MissingDelegatedActor`, all other rejections still fail. I know it was preexisting, but should be fixed I think. (we have the same wording below)

libs/@local/graph/api/src/rest/mod.rs
```diff
@@ -581,40 +587,25 @@ fn attach_request_middlewares<P, C>(
...
-                Arc::clone(&gate_limiters),
-                Arc::clone(&gate_secret),
-                request,
-                next,
-            )
-        }))
+        .layer(IpGateLayer {
+            limiters: rate_limiters,
+            service_secret,
+        })
```
TimDiekmann, line 608: I may miss something here. Each `.layer()` call wraps it into the `BoxCloneSyncService`, and the rejection is already a `Response` by the time the layer sees it. Where is the typed rejection?

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -72,6 +73,55 @@ impl AuthenticationMetrics {
     }
 }
 
+/// The response a request that failed authentication is answered with.
+///
+/// The failure is logged and counted when the rejection drops, whether or not a response was
+/// rendered from it.
```
TimDiekmann: Not true for the two paths that never build a rejection, they log but don't count.

libs/@local/middleware/src/rate_limit/mod.rs
```diff
@@ -555,49 +604,88 @@ where
...
+            Ok(())
+            | Err(Denial {
+                mode: RateLimitMode::Observe,
+                retry_after: _,
+            }) => Either::Right(self.inner.call(req).map_ok(Ok)),
+            Err(Denial {
+                mode: RateLimitMode::Enforce,
+                retry_after,
+            }) => Either::Left(future::ready(Ok(Err(TooManyRequests { retry_after })))),
+        }
```
TimDiekmann: This was `Denial::apply` before, now it's copied. What was the reason to inline it? Can't we return `Option<TooManyRequests>` or something like this instead?

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -240,64 +307,145 @@ fn store_outcome(
...
+            metrics: Arc::clone(&self.metrics),
+            bootstrap_route: self.bootstrap_route,
+        }
+    }
+}
+
+impl<B, P, C, S> tower::Service<http::Request<B>> for AuthenticationService<P, C, S>
 where
-    P: AuthenticationProvider<C>,
+    S: tower::Service<http::Request<B>, Error = Infallible> + Clone,
```
TimDiekmann: Whyd oes it need to be `Infallible`?

libs/@local/middleware/src/rate_limit/mod.rs
```diff
@@ -655,129 +741,142 @@ pub async fn ip_gate_middleware(
...
-    }
-}
+        let actor = match resolved.outcome() {
+            Ok(actor) => *actor,
+            Err(error) => {
+                // The authentication layer stores an error only as `MissingDelegatedActor` on a
+                // bootstrap route, and those requests present the service secret and passed above.
+                // Reaching this means the middleware order broke, so the report's attachments are
+                // the only account of what the provider saw — `Display` would print the first
+                // context and drop them.
```
TimDiekmann: This comments easily go out of sync if we add another tolerated error variant

libs/@local/middleware/src/lib.rs
TimDiekmann:
```suggestion
//! extends [`AuthenticationErrorKind`], a new caller type the sealed [`Caller`] — both are changes to
```

libs/@local/middleware/src/authentication/request.rs
```diff
@@ -191,21 +194,187 @@ impl AuthenticationError {
...
+}
+
+impl core::ops::Deref for AuthenticationError {
+    type Target = AuthenticationErrorKind;
+
+    fn deref(&self) -> &Self::Target {
+        &self.kind
+    }
+}
+
```
TimDiekmann: `AuthenticationError` is not a pointer

```suggestion
```

libs/@local/middleware/src/authentication/request.rs
```diff
@@ -191,21 +194,187 @@ impl AuthenticationError {
...
+impl AuthenticationError {
+    /// Creates the error for `kind`, not yet logged.
+    #[must_use]
+    pub const fn new(kind: AuthenticationErrorKind) -> Self {
+        Self {
+            kind,
+            logged: Atomic::<bool>::new(false),
+        }
+    }
```
TimDiekmann, line 222: Do we need both, `new` and `from`?

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -240,64 +307,145 @@ fn store_outcome(
...
+    /// The secret bootstrap routes require.
+    pub service_secret: Arc<str>,
+    /// The instruments rejections are counted on.
+    pub metrics: Arc<AuthenticationMetrics>,
+    /// Names the bootstrap routes by path.
+    pub bootstrap_route: fn(&str) -> bool,
+
+    /// Selects the caller type the chain resolves to: [`ActorId`] rejects uncredentialed
+    /// requests, `Option<ActorId>` serves them anonymously.
+    pub caller: PhantomData<C>,
```
TimDiekmann: This enforces `C: Sync`
```suggestion
    pub caller: PhantomData<fn() -> C>,
```

libs/@local/middleware/Cargo.toml
```diff
@@ -25,6 +25,7 @@ clap = { workspace = true, optional = true, features = ["derive"] }
 # Private third-party dependencies
 bytes                              = { workspace = true }
```
TimDiekmann: I think we can drop `bytes` here now.

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -72,6 +73,55 @@ impl AuthenticationMetrics {
...
+        }
+    }
+}
+
+impl IntoResponse for AuthenticationRejection {
+    fn into_response(self) -> Response {
+        match &self {
+            Self::Authentication { report, .. } => {
+                let error = report.current_context();
+                error_response(error.status_code(), error.client_message().to_owned())
```
TimDiekmann: `error.client_message()` is a `&'static str` and we expect `Cow`. It appears we don't need `Cow` at all here.

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -72,6 +73,55 @@ impl AuthenticationMetrics {
...
+    fn drop(&mut self) {
+        match self {
+            Self::Authentication { report, metrics } => {
+                AuthenticationError::log(report);
+                metrics.record_rejection(report.current_context());
+            }
+            Self::Misconfigured => {
+                tracing::error!(
+                    "`AuthenticatedActorId` extracted on a route without authentication middleware"
+                );
```
TimDiekmann, line 157: We should probably add the metrics here

libs/@local/middleware/src/authentication/request.rs
```diff
@@ -40,7 +45,7 @@ impl FaultDomain {
 
 /// Errors that can occur while authenticating a request.
 #[derive(Debug, Clone, derive_more::Display)]
-pub enum AuthenticationError {
+pub enum AuthenticationErrorKind {
```
TimDiekmann: (preexisting): we should not derive `Display` but use `client_message()` instead

TimDiekmann (changes requested):
A few pre-existing issues I spottet, but also a few new.

What is the exact motivation to do this PR now? I see that we avoid an `Arc` for rate limits, but the middleware in `layer` is still dynamic. Mainly asking because this highly conflicts with my current work.

libs/@local/middleware/src/authentication/mod.rs
indietyp: I had `fn() -> C` before, but this version allows us to say that `caller: PhantomData::<C>`, which I found to be neat, happy to change it tho.

libs/@local/middleware/src/authentication/request.rs
indietyp: I first did this, but then reverted it again - reason being that this means that e.g. the actor id is not present in the report, you had a test for exactly that, so reverted it.

libs/@local/middleware/src/authentication/mod.rs
indietyp: This can't have a metric, because in the extractor, we have no way to actually get the metrics (it'd be another extension that could also be misconfigured), but I added metrics for degradation in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9) - which I think was the other thing you were pointing at?

libs/@local/middleware/src/authentication/mod.rs
indietyp: done in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/Cargo.toml
indietyp: done in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/authentication/mod.rs
indietyp: done in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/rate_limit/mod.rs
indietyp: done in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/authentication/mod.rs
indietyp: it doesn't, it's residue from a previous version, where I thought that the error needed to be more composable - it did not - thankfully - removed in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/rate_limit/mod.rs
indietyp: reworded in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/authentication/mod.rs
indietyp: reworded in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9) to try to say it's only for the built rejection type.

libs/@local/graph/api/src/rest/mod.rs
indietyp: (discussed in greater detail offline) Typed the rejection on the extractor, but on the layer itself, to allow for better testing and use outside of Axum. Axum itself only sees `Response` once the layer is applied.

libs/@local/middleware/src/authentication/mod.rs
indietyp: reworded in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/authentication/request.rs
indietyp: renamed and rescoped in [84308b3](https://github.com/hashintel/hash/pull/9469/commits/84308b37fc0376536456b0b5c7b0e83eef5efbe9)

libs/@local/middleware/src/authentication/request.rs
indietyp: I removed `From`; I had both because I didn't want to enable `const_convert` for the crate, and `new` is const-constructible.

TimDiekmann: Where do we need a `const` new btw?

libs/@local/middleware/src/authentication/request.rs
TimDiekmann: Ah, yes, I think I had something similar 😄

TimDiekmann (dismissed):

libs/@local/middleware/src/authentication/request.rs
indietyp: There is no depper hidden meaning, I just like to have especially constructors always have a way to create them via const, because it cascades. I've had many times where I want to create e.g. a const or static of something, and then I need to make all the constructors all the way down const as well, which is a pain, especially considering that we already try to constify the methods on them where available

TimDiekmann (dismissed):

libs/@local/middleware/src/authentication/mod.rs
```diff
@@ -240,64 +347,146 @@ fn store_outcome(
...
+    P: AuthenticationProvider<C>,
 {
-    let bootstrap = bootstrap_route(request.uri().path());
-    if bootstrap
-        && let Some(rejection) =
-            service_secret_rejection(request.headers(), &service_secret, &metrics)
-    {
-        return rejection;
+    type Error = S::Error;
+    type Response = Result<S::Response, AuthenticationRejection>;
```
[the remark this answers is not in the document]
indietyp: tracked in BE-820

TimDiekmann (approved):

TimDiekmann (approved):

PR #9475, opened 2026-08-31
FE-1552: Reuse sweep workers and batch surface sampling

kube:

Filling the sweep drawer's view spawned 166 web workers and every slider commit spawned 16–24 more, each one re-fetching and re-evaluating the engine module before it could simulate. This PR pools the workers (a lease resets and returns its worker instead of terminating it) and samples the contour surface in batched chunks — one experiment for 24 cells instead of one per cell — so the same view now fills with 20 workers, a slider commit spawns 0, and switching the shown surface metric repaints from cached samples instead of re-simulating. Stacked on #9474.

## 🔗 Related links

- Base PR: #9474

## 🔍 What does this change?

- `@hashintel/petrinaut-core` gains `createReusableWorkerFactory`: wraps any `WorkerFactory` with an idle pool and hands out leases. A lease's `terminate()` detaches its listeners, posts `cancel` (the protocol re-initializes on the next `init` by design), and returns the worker to the pool — only once every acknowledgement the worker still owes has arrived (the lease counts cancels out and acks in; the runtime's dispose path sends its own cancel right before terminating), so no stale `cancelled` reply can land on the next lease's listeners and settle its shard. A worker whose script errored is never re-pooled.
- The factory distinguishes `drain()` (flush idle and mid-reset workers; the pool keeps operating) from `dispose()` (shut the pool: later releases terminate their workers even though their acks arrive after). The experiments provider holds one pool per mount with `maxIdle` sized to the sharded batch plus the background lanes, routes all four worker-creation sites through it, drains it when the injected factory changes, and disposes it on unmount.
- The sweep session gains `sampleCellsBatch`: many surface cells become one experiment by laying each cell's parameter values out as per-run draws (`runsPerCell` runs per cell) and grouping the per-run metric values back into index-aligned per-cell means. Every cell pins the same explicit per-run seeds the per-cell ladder's first batch derives, so a cell's value does not depend on which chunk sampled it and matches the navigator's own runs. These batches route to the CPU lane because only the CPU workers report per-run values.
- One batch shares one compiled initial marking, so before batching a chunk the provider compiles each cell's initial state (at the actual cell values) and batches only when all of them equal the first cell's; a chunk whose marking the swept parameters shape — or with a cell that does not compile — keeps the per-cell path via the same `sampleSurfaceCells` contract.
- The surface component samples its grid in coarse-to-fine 24-cell chunks and caches every metric's value per cell, so changing the shown metric repaints without re-simulating. A refused chunk is a hole to fill later, not the end of the walk, and a chunk resolving after the walk restarted cannot merge stale values into the new grid (the grid state carries its walk key).
- Updates the user guide's surface paragraph and the `sweep-surface` / `worker-sharding` architecture pages.

Measured on the `Bench / SweepDrawer` CPU story (baseline → this PR): workers to fill the view 166 → 20, workers per slider commit 16–24 → 0, commit→first-chart-frame median 171 ms → 150 ms, 11×11 surface fill completes in ≤4 s (previously the same view left the surface empty on terminating nets and took ~15 s to sample serially). The GPU story fills the same surface with 3 workers.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `reusable-worker-factory.test.ts` (new, 12 tests): reuse after release, listener detachment, the owed-acks gate (a worker owing two acks is not pooled on the first; acks already observed are not owed; non-cancelled traffic is ignored), quarantine of a silent worker, `maxIdle` overflow, drain of idle and mid-reset workers, dispose semantics (later releases and late acks terminate), double-terminate safety.
- `sweep-session.test.ts`: three new `sampleCells` tests pin the draw layout, the pinned per-cell seeds, the index-aligned regrouping into means, and the null entry for a cell with no finished runs.
- `provider.test.tsx`: a new test pins that one worker serves two consecutive experiments (reset and re-leased, then terminated when the provider unmounts); the message-sequence pins follow the pooled contract.

## ❓ How to test this?

2. Watch the surface fill: all 121 cells sample within a few seconds, coarse shape first.
3. Switch the surface metric: the plot repaints instantly from cache.
4. Drag and release a parameter slider: charts restream without worker churn (observable via the browser task manager or a `Worker` constructor hook).

## 📹 Demo

### DIFF
--- a/.changeset/sweep-compute-reuse.md
+++ b/.changeset/sweep-compute-reuse.md
@@ -1,0 +1,6 @@
+---
+"@hashintel/petrinaut-core": patch
+"@hashintel/petrinaut": patch
+---
+
+Reuse simulation workers across sweep batches instead of spawning and terminating a pool per batch, and sample the sweep surface in batched chunks (one experiment for many cells when the swept parameters do not shape the initial marking). Switching the surface metric now re-reads the cached samples instead of re-simulating.
--- a/libs/@hashintel/petrinaut-core/src/experiments.ts
+++ b/libs/@hashintel/petrinaut-core/src/experiments.ts
@@ -40,3 +40,7 @@ export {
   WORKER_POOL_BACKEND_ID,
   type WorkerPoolExperimentBackendOptions,
 } from "./experiments/worker-pool-experiment-backend";
+export {
+  createReusableWorkerFactory,
+  type ReusableWorkerFactory,
+} from "./simulation/runtime/reusable-worker-factory";
--- a/libs/@hashintel/petrinaut-core/src/simulation/runtime/reusable-worker-factory.test.ts
+++ b/libs/@hashintel/petrinaut-core/src/simulation/runtime/reusable-worker-factory.test.ts
@@ -1,0 +1,255 @@
+import { describe, expect, it } from "vitest";
+
+import { createReusableWorkerFactory } from "./reusable-worker-factory";
+
+import type { WorkerLike } from "../../environment";
+
+type Listener = (event: { data: unknown }) => void;
+
+/**
+ * A worker double that records messages and dispatches by event type the way
+ * a real worker does. With `autoAck` (the default) it answers `cancel` with
+ * `cancelled` synchronously, like the in-process worker; `autoAck: false`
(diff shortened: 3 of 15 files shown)


kube:

PR #9476, opened 2026-08-31
FE-1552: Prioritize the selected point and fill the surface in quad-tree levels

kube:

The metric charts and the contour surface competed for workers from the first moment, and the surface swept its grid in one fixed order. Now the navigator's selected point always streams first — surface sampling starts only after the current selection's first frames, and re-waits after every slider move — and the surface fills in quad-tree levels (corners, then each level splitting every region in two per axis) with several chunks in flight, so each completed level is a full coarse picture. Stacked on #9475.

## 🔗 Related links

- Base PR: #9475

## 🔍 What does this change?

- The sweep session gains `whenSelectionStreamed()`: a gate that opens once the current selection has streamed its first frames (or has cached runs, or failed, so nothing hangs) and re-arms on every selection change. The provider awaits it before dispatching any surface chunk, so the selected point outranks secondary sampling at all times.
- `quadTreeLevels(nx, ny)` groups the grid into refinement levels: level 0 is the four corners, each further level is the `i / 2^level` lattice rounded onto the grid, minus the cells earlier levels produced. On the 11×11 surface grid the levels are 4, 5, 16, 56, and 40 cells.
- The surface samples level-aligned chunks (split at 24 cells within a level) with three chunks in flight, pulled from one queue by parallel lanes — the coarse picture completes level by level while chunk setup and compute overlap.
- Updates the user guide's surface paragraph and the `sweep-surface` architecture page.

Measured on the `Bench / SweepDrawer` CPU story: the navigator's runs stream at 927 ms after page load, the first surface samples land after them at 1041 ms, the fill progresses 4 → 25 → 81 → 121 cells, and the full 11×11 surface completes at 1.6 s (previously ~4 s with sequential chunks).

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `sweep-session.test.ts`: four new `whenSelectionStreamed` tests (resolves on first stream; re-arms on selection change; resolves immediately for a cached selection; resolves on dispose).
- `contour-grid.test.ts`: three new `quadTreeLevels` tests (corners first and the level-1 cross; levels disjoint and covering the grid; gradual levels on non-power-of-two grids).
- Existing surface and provider suites cover the sampling paths.

## ❓ How to test this?

2. Confirm the metric charts start streaming before any surface cell appears.
3. Watch the surface fill corners-first, sharpening level by level rather than sweeping.
4. Drag a parameter slider: the charts restream first, then the surface resumes.

## 📹 Demo


kube:

PR #9477, opened 2026-08-31
FE-1552: Pipeline ladder rungs and shard surface chunks adaptively

kube:

The sweep view computed one batch at a time in two places: the navigator's ladder waited for each rung to finish before starting the next, and surface chunks were fixed at one worker each even when the rest of the pool sat idle. This PR pipelines the ladder, lets surface chunks shard across whatever the foreground isn't using, and paints surface cells as each shard completes instead of at chunk boundaries. Stacked on #9476.

## 🔗 Related links

- Base PR: #9476

## 🔍 What does this change?

- Surface chunk batches adapt their width to foreground pressure: the session stamps each chunk request with whether its refine ladder is computing (`foregroundActive`), and the provider keeps a narrow (one-worker) and a wide (sharded) CPU lane — while the ladder computes on the CPU pool a chunk takes one worker, and once the ladder idles or the experiment computes on the GPU the same chunk shards across the freed pool.
- The ladder pipelines its rungs: run ranges are disjoint and draws and seeds are prefix-stable, so as soon as a rung streams its first frames the next rung starts alongside it — at most two in flight. Folds stay strictly ordered (a rung folds only after every earlier rung folded; a stopped rung discards its started successor so a gap can never enter the cache), and both live rungs' frames merge into one published view.
- `sampleCells` accepts a partial listener: CPU workers report per-run values as each shard completes, so a sharded chunk streams its cells in slices, and the per-cell fallback reports each resolved cell the same way. The surface merges partials through the same walk-key-guarded path as final results.
- The worker pool's `maxIdle` covers the pipelined working set (two sharded foreground batches plus the surface lanes); a smaller pool terminated the overflow on every slider commit and respawned it a moment later.
- Updates the `parameter-sweeps` and `sweep-surface` architecture pages.

Measured on the `Bench / SweepDrawer` CPU story (previous PR → this one): navigator reaches its full 1000 runs at 1312 ms after page load (was 1488 ms, with the runs counter climbing smoothly through rung boundaries instead of plateauing), surface full fill unchanged at ~1.6 s, worker churn per slider commit 0 → 1.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- On the GPU backend the pipelined ladder can hold two GPU experiments concurrently, doubling GPU buffer residency for the overlap window. Bounded and not observed to fail; flagged for review.

## 🛡 What tests cover this?

- `sweep-session.test.ts` (26 tests): new pins for the `foregroundActive` hint (flips with the ladder's computing state), pipelining (successor starts on first stream with the derived seed; out-of-order completions fold in order; a selection change aborts both live rungs), and partial streaming (per-cell means firm up shard by shard, nulls for cells with no runs yet).
- Existing contour-grid, provider, and factory suites cover the surrounding paths.

## ❓ How to test this?

2. Watch the run counter climb smoothly through the ladder instead of pausing at each rung.
3. Watch the surface: after the navigator's runs finish, cells land in visibly larger waves (sharded chunks streaming shard by shard).
4. On `→ Gpu`, the surface fills at full width immediately since the navigator uses no CPU workers.

## 📹 Demo

### DIFF
--- a/.changeset/parallel-orchestration.md
+++ b/.changeset/parallel-orchestration.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+Sweep compute parallelism: surface chunk batches shard across the CPU pool whenever the navigator's ladder is idle or computing on the GPU; the ladder pipelines its rungs (the next rung starts once the current one streams); and surface cells paint as each shard completes instead of waiting for the whole chunk.
--- a/libs/@hashintel/petrinaut/src/react/experiments/context.ts
+++ b/libs/@hashintel/petrinaut/src/react/experiments/context.ts
@@ -203,6 +203,9 @@ export type ExperimentsContextValue = {
     experimentId: string,
     positions: readonly Readonly<Record<string, number>>[],
     runsPerCell: number,
+    onPartial?: (
+      cells: readonly (Readonly<Record<string, number>> | null)[],
+    ) => void,
   ) => Promise<readonly (Readonly<Record<string, number>> | null)[] | null>;
   /**
    * Computes one metric sample against an arbitrary net snapshot, on the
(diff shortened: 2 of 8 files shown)


kube:

PR #9479, opened 2026-08-31
BE-755: Add session cache to Graph authentication

TimDiekmann:

Caches verified sessions in the Graph's authentication, so repeated requests with the same credential do not verify against Kratos and query the principal store each time. Entries expire on a hard TTL — the upper bound on how long a revoked session keeps authenticating.

Stacked on #9469.

## 🔗 Related links

- BE-755
- BE-812

## 🚫 Blocked by

- [ ] #9469

## 🔍 What does this change?

- The provider chain hands rejections out as `Arc<Report<AuthenticationError>>`: one verification can answer several concurrent requests. The rejection count latches per request, the log per error, so N concurrent rejections count N times and log once.
- The Kratos session provider caches verified sessions in a `moka` cache: bounded capacity, TTL eviction, one verification per credential in flight. Keys are HMAC-SHA-256 digests of the credential under a key drawn per process, so session secrets are not kept in memory and a digest cannot be enumerated back to its credential, and the token and cookie forms hash separately.
- Cache lookups are metered by outcome (`hit`, `miss`, `failed`) and the entry count is gauged.
- TTL and capacity are CLI-configurable. A TTL of 0 disables the cache.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- Rejected sessions are not cached: invented credentials bypass any cache and are the rate limiter's job. Negative caching stays out until production metrics argue for it.
- User deletion does not invalidate cache entries: the admin server can run as its own process and cannot reach this in-process cache, so the TTL is the accepted exposure window for deleted users and revoked sessions alike.

## 🐾 Next steps

- BE-812: refresh-ahead for cached sessions, gated on production metrics.
- Eviction and capacity instruments for the cache, once production shows whether the bound is ever under pressure.

## 🛡 What tests cover this?

- Session-provider tests for cache behaviour: repeated requests verify once, rejections are never cached, token and cookie credentials cache separately, expiry re-verifies, and the lookup metrics record hits and misses.
- The existing middleware, graph-authentication, and graph-api suites.

## ❓ How to test this?

1. Checkout the branch
2. `cargo nextest run --package hash-middleware --package hash-graph-authentication --package hash-graph-api --all-features`

## 📹 Demo

Not applicable — no user-facing change.


.config/mise/tasks/test/unit.sh
```diff
@@ -85,4 +86,8 @@ if [[ $COVERAGE == "true" || ${TEST_COVERAGE:-false} == 'true' || ${TEST_COVERAG
 fi
 
 cargo hack "${HACK_ARGUMENTS[@]}" nextest run "${NEXTEST_ARGUMENTS[@]}" "${ARGUMENTS[@]:-}"
-cargo test "${COMMON_ARGUMENTS[@]}" --all-features --doc
+
+# Doc-tests need a library target, which binary-only crates do not have
+if jq -e --arg crate "$CRATE" '.packages[] | select(.name == $crate) | .targets | any(.kind | contains(["lib"]))' <<< "$METADATA" > /dev/null; then
+    cargo test "${COMMON_ARGUMENTS[@]}" --all-features --doc
```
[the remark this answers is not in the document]
TimDiekmann: That's a duplicate of the above

libs/@local/graph/authentication/src/kratos/session.rs
```diff
@@ -29,6 +36,32 @@ enum SessionCredential<'h> {
     Cookie(&'h str),
 }
 
+/// The cache key of a session credential: SHA-256 over its form byte and value.
+///
+/// The cache holds these digests, so session secrets do not sit in memory for an entry's
+/// lifetime. Hashing keeps the digest uninvertible only because Kratos session credentials are
+/// high-entropy — a low-entropy credential must not be keyed through this. Without [`Debug`],
+/// so the digest of a live credential cannot reach logs.
```
indietyp: I have pondered over this quite a bit, because getting this wrong would be severe. I think the implementation is practically safe, but the documentation currently overstates the property it provides.

We are basically working with two separate properties:

* collision resistance: A ≠ B => cache_key(A) ≠ cache_key(B) (is `cache_key` injective?)
* preimage resistance: invert(cache_key(A)) = computationally infeasible or impossible (aka: possession of a cache key should not reveal the credential)

First, regarding the high/low entropy claim (the shorter one), that is well-founded, but it's about preimage resistance, not collision resistance. Unkeyed SHA-256 (which is used here) is deterministic and unkeyed, which means that a key could be recovered through enumeration; low entropy has nothing to do with collision resistance.

Essentially, the session key presented here represents two universes that we equate to the same session key: the session token and the cookie itself. Which is why I'll go through them separately.

Kratos session tokens are contained inside a token universe of $62^{32} \\approx 2^{190.5}$, because this is smaller than the SHA-256 output space of $2^{256}$; it is mathematically possible that an injective mapping between the two exists - but we should not expect that to be the case. However, we can only speculate, verifying only through exhaustive enumeration (which is computationally infeasible) that such a mapping exists - that's the whole argument anyway - if we're able to enumerate the whole token universe to guarantee that the output space is injective, we'd be doing our own enumeration attack and creating a lookup of SHA to session token.

If we only had to worry about the Kratos session token universe, many things here would be easier, as 190.5 bits is enough to make use of functions that allow for an injective mapping; the complicating factor here is the section type that we cram into the same hash: the session cookie. To understand the cookie's entropy, we must first understand its makeup. A Kratos cookie is created through the following code:

```
cookie.Values["session_token"] = session.Token
cookie.Values["expires_at"] = session.ExpiresAt.UTC().Format(time.RFC3339Nano)
cookie.Values["nonce"] = randx.MustString(8, randx.Alpha)
```

Gorilla `securecookie` then encrypts the serialization using AES-CTR, which preserves entropy, the issue we face: the state space we're in for the cookie is larger than $2^256$ for any cookie, that means that an injective mapping *is not possible by definition* (it doesn't say anything about the feasibility of such a collision), but it is possible, and has nothing to do with entropy, but rather just with the state space that is used.

TL;DR: I think that SHA-256 is a defensible design choice here with the session cookie kinda forcing our hand here, but I think that we could strengthen it against offline attacks by using HMAC-SHA-256 instead of unkeyed SHA-256, and I think that instead of just having `&str` as type, I think `SessionCredential` would benefit from some newtypes that make sure that their source provenance is known, so that we can guarantee the claims this code relies on, as otherwise it's eas(ier) to get wrong.

indietyp (dismissed):

indietyp (dismissed):

indietyp (dismissed):

indietyp (dismissed):

.config/mise/tasks/test/unit.sh
```diff
@@ -14,14 +14,18 @@ COVERAGE=$usage_coverage
...
+
+if ! jq -e --arg crate "$CRATE" '.packages | any(.name == $crate)' <<< "$METADATA" > /dev/null; then
+    echo "Error: '$PACKAGE' is not a crate in the Cargo workspace"
     exit 1
 fi
 
-# Remove the package namespace from the package to get the crate name
-CRATE=${PACKAGE#*@rust/}
+# Doc-tests need a library target, which binary-only crates do not have
+HAS_LIB=$(jq --arg crate "$CRATE" '.packages[] | select(.name == $crate) | .targets | any(.kind | contains(["lib"]))' <<< "$METADATA")
```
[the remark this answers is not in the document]
TimDiekmann: Fixed in https://github.com/hashintel/hash/pull/9479/commits/9eca6a6bfa19b5a92251d1b75ccf655dd40594f6

libs/@local/graph/authentication/src/kratos/session.rs
```diff
@@ -678,4 +1050,513 @@ mod tests {
...
+        let waiter = tokio::spawn({
+            let provider = Arc::clone(&provider);
+            async move {
+                let authentication: ControlFlow<Result<ActorId, _>> =
+                    provider.authenticate(&session_token_header()).await;
+                authentication
+            }
+        });
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        release.notify_one();
```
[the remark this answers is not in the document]
TimDiekmann: Fixed in https://github.com/hashintel/hash/pull/9479/commits/35ce090d0324b2a667a2e5c080221647f7f2d9a9

libs/@local/middleware/src/authentication/request.rs
```diff
@@ -206,18 +207,15 @@ pub struct AuthenticationError {
     kind: AuthenticationErrorKind,
     /// Whether the error has been logged.
     logged: Atomic<bool>,
-    /// Whether the error has been counted as a rejection.
```
indietyp, line 209: Out of curosity: why did you move this out from the error to be inside of the rejection?

TimDiekmann: Because the cache, all concurrent requests for the same actor will share a single Report. If `recorded` would live in the error, it would count as a single rejection.

indietyp: ah gotcha makes complete sense, thank you!

indietyp (approved):

PR #9486, opened 2026-09-01
FE-1552: Make the surface a drag control and show parallel batches in the summary

kube:

The sweep drawer's feedback loop gets three upgrades: the contour surface becomes a control you can drag (live crosshair and value readout, commit on release, with the navigator's position marked), the summary shows every simulation computing in parallel as a compact toggleable list instead of one inherited single-run progress bar, and the summary itself tightens into a stat strip. Stacked on #9478.

## 🔗 Related links

- Base PR: #9478

## 🔍 What does this change?

- The contour surface navigates by drag as well as click: press shows a crosshair (CSS overlay positioned in percentages, so pointer moves never repaint the field raster) with the parameter values under the pointer in the caption; release collapses both shown axes there. An orange ring marks where the navigator currently sits (nearest sampled cell, correct on unevenly sub-sampled integer axes). Drags track one armed pointer id — a second finger is ignored, `touch-action: pan-y` leaves vertical swipes to scroll the drawer (the browser's pointercancel aborts without committing), and `lostpointercapture` cleans up when something swallows the release. The optimization surface inherits the same interaction.
- The sweep session keeps a registry of every computing batch — the selection's ladder rungs, surface chunks, cell refinements — as a `batches` store (progress ticks throttled to 100 ms), exposed on the record as `sweepBatches`. The summary renders it as a "N computing" chip that expands into compact rows: kind, its own progress bar, run counts, selection batches first and tinted as the priority work.
- The progress bar's meaning changes from one batch's simulated time (which sawed to 100% once per ladder rung) to the selected combination's runs over the run budget; the top-bar Active-experiments popover is aligned to the same meaning, so the two bars agree.
- The summary grid becomes a single wrapping stat strip with hairline dividers (dividers clip at row starts, so wrapped rows begin flush) and a status dot (blue running, green complete, red error).
- Updates the user guide's experiments page and the `sweep-surface` architecture page.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `sweep-session.test.ts`: a new test pins the batch registry (selection-first ordering, the pipelined successor joining, batches dropping as they finish, dispose clearing).
- Existing contour-surface, provider, and drawer suites cover the surrounding paths; the drag interaction and the activity list were verified by driving the `Bench / SweepDrawer` story with Playwright (drag readout, commit on release, chip expansion mid-compute).

## ❓ How to test this?

2. While it computes, click the "N computing" chip: the list shows the selection batch first, then surface chunks, each with its own progress.
3. Press and drag on the surface: a crosshair follows with the parameter values in the caption; release and the navigator jumps there, charts restreaming.
4. Note the orange ring marking the committed position, and the summary's status dot and stat strip.

## 📹 Demo

### DIFF
--- a/.changeset/sweep-drawer-ux.md
+++ b/.changeset/sweep-drawer-ux.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+The sweep surface navigates by drag as well as click (live crosshair and value readout, commit on release) and marks the navigator's current position; the experiment summary shows a compact toggleable list of every batch computing in parallel, with the progress bar tracking the selected combination's runs instead of one batch's simulated time.
--- a/libs/@hashintel/petrinaut/docs/experiments.md
+++ b/libs/@hashintel/petrinaut/docs/experiments.md
@@ -68,7 +68,7 @@ Every selection uses the same seed sequence (common random numbers), and a run's
 
 #### The surface view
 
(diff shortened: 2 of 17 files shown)


kube:
Consolidated at Chris's request: this PR's commits now live in #9478 (same commits, rebased onto main). Closing; the branch stays in place for reference.

PR #9489, opened 2026-09-01
FE-1552: Open GPU runs with a preview tile and reuse calibration across batches

kube:

GPU experiments deliver their first usable picture much sooner. A streamed run now opens with a small preview tile — a full-timeline picture from 128 runs before the bulk computes — and one GPU backend (device, compiled shader, learned calibration) is reused across a session's batches instead of being rebuilt and re-probed per ladder rung. On the bench story the full 1,000-run GPU ladder drops from ~1.6 s to ~0.9 s. Stacked on #9486.

## 🔗 Related links

- Base PR: #9486

## 🔍 What does this change?

- `planTiles` opens a streamed experiment with a `GPU_PREVIEW_TILE_RUNS` (128) tile before the full-size tiles. Every tile re-delivers the whole time axis cumulatively, so the preview puts a statistically usable picture on screen while the bulk still computes — the GPU's version of the ladder's first rung. Seeds derive from absolute run indices, so the split changes nothing about the results; no preview when nobody streams or the run count is within 2× of the preview size.
- `gpu-backend-cache` holds one GPU setup per experiment backend, keyed by what shapes the shader: net/artifact identity, baked parameter values, metric set, dt, integration method, initial marking. Values of per-run-buffered parameters are deliberately not in the key, so range batches across different selections share one backend — and its calibration (histogram windows, derived capacities, the shader compiled at them) carries over instead of being re-probed. A stale calibration heals through the existing escape/overflow re-runs, which also update the cache.
- Cached backends are leased: pipelined rungs hold the same entry concurrently, and a displaced or unsupported entry destroys its device only once the last lease ends. Without a cache (CLI, tests) the handle behaves exactly as before.
- The calibration itself is also keyed per marking (`calibrationKey`) inside a shared backend, so marking-shaping sweeps keep separate calibrations per marking.
- Updates the `gpu-capacity-calibration` architecture page and the user guide's calibration paragraph.

Measured on `Bench / SweepDrawer → Gpu` (base branch → this PR): full 1,000-run ladder 1,512–1,582 ms → 937 ms after page load; slider commits 217–255 ms → 216–225 ms to first new runs; GPU command encoders per commit 75–79 → 68–79. No page errors; results unchanged (same seeds, same trajectories).

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

- The preview tile is invisible on light models (all tiles finish inside one 100 ms publish window); it pays on heavy models where a full tile takes visible time.

## 🛡 What tests cover this?

- `runner.test.ts`: four new `planTiles` tests (preview then full tiles, exact coverage with and without preview, the skip conditions).
- `gpu-backend-cache.test.ts` (new, 9 tests): reuse and single build per key, concurrent acquires sharing an in-flight build, lease counting across eviction, evicting releases, unsupported builds never cached, uncached releases destroying their own device; plus two `gpuBackendSetupKey` tests (buffered values excluded, everything else keyed).
- `metric-windows.test.ts`: two new `calibrationKey` tests (marking + metric set, token-word boundaries).

## ❓ How to test this?

1. Open `Bench / SweepDrawer → Gpu` in Storybook (`yarn workspace @hashintel/petrinaut dev:storybook`) with a WebGPU-enabled browser.
2. Watch the run counter: the full ladder completes in about a second (it re-probed and recompiled per rung before).
3. Move a parameter slider: new selections stream without a per-batch setup pause.

## 📹 Demo


kube:

PR #9490, opened 2026-09-01
FE-1552: Paint heatmap bins by extent and ease streamed updates in

kube:

Distribution heatmaps had two visible faults while streaming: frames binned on different lattices (the host's exact initial frame next to GPU frames binned at a stride) drew as alternating dark and empty rows, and every cumulative re-delivery — a GPU tile, a pipelined rung folding in, a calibration re-run — snapped the whole picture, which read as flashing. Bins are now painted across the rows their own width covers, and each streamed update eases into the displayed picture over a few publishes. Stacked on #9489.

## 🔗 Related links

- Base PR: #9489

## 🔍 What does this change?

- `buildHeatmapDensityGrid` spreads each bin's count over the rows its extent covers (`[value − w/2, value + w/2]`, weighted by overlap) instead of splatting it as a point onto its two nearest rows. `w` is the frame's own smallest bin gap (the chart-wide gap for single-bin frames), so a stride-2 bin over a 1-unit lattice fills both rows evenly. The half bin reaching past the outermost row centers is not drawn; renormalizing it into the edge rows made them darker than the interior.
- `blendHeatmapGrids` mixes the new grid with the one on screen (weight 0.5 per publish — an eighth of the old picture remains after three publishes at the session's 100 ms cadence). Shared columns blend, appended columns show as they are, and a moved lattice is bridged by sampling the previous grid at each new row's value. The plugin keeps the displayed grid and resets it when the frame set empties, so a new selection starts from its own data (the plot's crossfade overlay bridges that transition, as before).
- Adds `Simulate / ExperimentMetricTimeline → Mixed bin lattices`, a story with an exact frame 0 and stride-2 frames that reproduces the stripes on the base branch and renders continuously here.
- Updates the user guide's heatmap paragraph.

Measured on the `Bench / SweepDrawer` stories after a slider commit, sampling the already-drawn left 40% of the first chart every 75 ms (frontier columns excluded, so appended progress does not count as flicker): mean per-sample change 5.7 → 4.2 on CPU, 14.4 → 11.2 on GPU. The largest single jump is unchanged — that is the first data after the selection change, which the crossfade overlay covers on its own canvas.

This PR:

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

- `distribution-heatmap.test.ts`: two new splat tests (a stride-2 frame beside an exact frame fills every spanned row evenly; a bin on the lattice stays in its own row) and four `blendHeatmapGrids` tests (no previous grid, shared vs appended columns, a moved lattice resampled by value, convergence over repeated publishes). The five pre-existing grid tests still pass unchanged.

## ❓ How to test this?

1. Open `Simulate / ExperimentMetricTimeline → Mixed bin lattices` in Storybook (`yarn workspace @hashintel/petrinaut dev:storybook`): the band is continuous, with no alternating rows.
2. Open `Bench / SweepDrawer → Cpu`, wait for the fill, then drag a parameter slider: the charts restream and successive updates ease in rather than snapping.

## 📹 Demo

### DIFF
--- a/.changeset/heatmap-smoothing.md
+++ b/.changeset/heatmap-smoothing.md
@@ -1,0 +1,6 @@
+---
+"@hashintel/petrinaut": patch
+"@hashintel/petrinaut-core": patch
+---
+
+Distribution metric frames carry the extent of their bins (`binExtent`, set by width binning and GPU histogram windows), and distribution heatmaps paint each bin across the rows that extent covers, so frames binned at different strides no longer alias into alternating dark and empty rows. Streamed heatmap updates ease into the picture and settle instead of snapping, so re-deliveries no longer flash.
--- a/libs/@hashintel/petrinaut-core/src/index.ts
+++ b/libs/@hashintel/petrinaut-core/src/index.ts
@@ -278,6 +278,7 @@ export type {
   MonteCarloUserDefinedMetricAggregation,
   MonteCarloUserDefinedMetricConfig,
   MonteCarloUserDefinedDistributionMetricFrame,
+  MonteCarloUserDefinedMetricBinExtent,
   MonteCarloUserDefinedMetricDistributionBin,
   MonteCarloUserDefinedMetricFrame,
   MonteCarloUserDefinedMetricMeasureInput,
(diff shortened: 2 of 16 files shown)


kube:
Consolidated at Chris's request: this PR's commits now live in #9478 (same commits, rebased onto main). Closing; the branch stays in place for reference.

PR #9505, opened 2026-09-02
FE-681: add multiselect to ds

alex-e-leon:

- Adds multiple prop to select to enable multi select
- Adds searcheable prop to select

This PR:

The changes in this PR:


lunelson (approved):
I'm wondering only if the `value` should be sorted in the same order as the `option`s. It's possible to create different `value`s which are logically identical but will appear different depending on the order in which they were clicked

alex-e-leon:
> I'm wondering only if the `value` should be sorted in the same order as the `option`s. It's possible to create different `value`s which are logically identical but will appear different depending on the order in which they were clicked

Its not a bad idea, and I almost went ahead and implemented it, but after thinking about it I think it's better not to have this.

In practice it's not really an issue - it won't create any issues with memoization, since in order to change the order, a user would need to unselect and re-select options which would already trigger onChange with the inbetween states and therefore break any memoization in the interim.

And in cases where the order is important the consuming developer is likely already sorting, and in practice its often not a straight alphabetical sort, so we would be double sorting in some scenarios.

But the actual reasons I decided it's better not to implement it are
a) either the selected UI would annoyingly sort for users (so when a user selected an item it could be added to the front) or the data + ui would be out of sync
b) it also prevents a consumer from knowing the order a user selected options, and in some situations the order may be important. Leaving it unsorted gives developers more control on how its used.

PR #9535, opened 2026-09-04
SRE-1023: Skip playwright tests unless in merge queue

CiaranMn:

Playwright tests are often the bottleneck in required CI actions at ~20 minutes, and are required to get into the merge queue, and then again in the merge queue. 

Ideally we would get the time down (most of it is dependency builds which can be cached), and will be doing that, but as a temporary fix this PR skips the Playwright tests until actually in the merge queue.

The downside is that feedback if tests actually fail will only come when a PR is kicked out of the merge queue, but this will be the minority of cases.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -225,19 +225,31 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
+    env:
+      # Playwright is only executed in the merge queue. Pull requests and pushes
+      # to main still spawn this matrix cell so the check reports success, but
+      # skip the suite itself.
+      RUN_TESTS: ${{ matrix.name != '@tests/hash-playwright' || github.event_name == 'merge_group' }}
     steps:
       - name: Show selection reason
         env:
           MATRIX: ${{ toJSON(matrix) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -225,35 +225,50 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
+    env:
+      # Playwright is only executed in the merge queue. Pull requests and pushes
+      # to main still spawn this matrix cell so the check reports success, but
+      # skip the suite itself.
+      RUN_TESTS: ${{ matrix.name != '@tests/hash-playwright' || github.event_name == 'merge_group' }}
```
TimDiekmann: This way it still spawns the job. This means, it will show green on the PR, although it never ran. The better way would be to remove it from the matrix `Determine changed packages` step.

TimDiekmann (changes requested):
Not a fan of this, but fine to do for now, But we should disable it properly and not through skipping the steps.

TimDiekmann (approved):

TimDiekmann:
It works:
image

PR #9553, opened 2026-09-06
BE-826: Load explicit TOML files in hash-config

TimDiekmann:

Allow HASH applications to load required TOML files on top of programmatic defaults. Files have predictable precedence, and failures identify the source without including rejected values or TOML source excerpts.

## 🔗 Related links

- [BE-826: Load explicit TOML files in hash-config](https://linear.app/hash/issue/BE-826/load-explicit-toml-files-in-hash-config) _(internal)_
- [H-4170: Introduce config file for HASH](https://linear.app/hash/issue/H-4170/introduce-config-file-for-hash) _(internal)_

## 🚫 Blocked by

None.

## 🔍 What does this change?

- Add repeatable `Loader::with_toml_file(path)`, with file reads deferred until `load`.
- Apply every file above all defaults, regardless of builder-call order. Later files replace scalars and arrays; maps merge recursively.
- Add typed read and parse errors. Read failures retain their I/O cause; malformed TOML reports its path and position without retaining the parser error or source excerpt.
- Derive `LoadError` display messages and document the API with an executable file example.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This slice accepts required TOML files only. Profiles, discovery, optional files, other formats, environment variables, CLI integration, and a public provenance API are outside its scope.

## 🐾 Next steps

Choose the next configuration source or inspection feature as a separate, small change.

## 🛡 What tests cover this?

- Eleven file integration tests cover deferred reads, precedence, merging, correction of an earlier mistyped value, empty file layers, missing files, directories, invalid UTF-8, malformed TOML, invalid values, and missing required fields.
- Six Insta snapshots check complete error reports alongside typed error assertions and checks that rejected values are absent before snapshot normalization.
- `yarn workspace @rust/hash-config test:unit` passes all 30 tests, including snapshots, and both executable doctests.
- Clippy with all features/targets and warnings denied, Rustdoc generation, and `git diff --check` pass.

## ❓ How to test this?

1. Run `yarn workspace @rust/hash-config test:unit`.
2. Run `cargo clippy --package hash-config --all-features --all-targets --no-deps -- -D warnings`.

## 📹 Demo

Given `hash-graph.toml`:

```toml
host = "localhost"
```

```rust
#[derive(serde::Deserialize)]
struct Config {
    host: String,
    port: u16,
}

let config = hash_config::Loader::new()
    .with_defaults(serde_json::json!({ "port": 5432 }))
    .with_toml_file("hash-graph.toml")
    .load::<Config>()?;

assert_eq!(config.host, "localhost");
assert_eq!(config.port, 5432);
```

### DIFF
--- a/libs/@local/config/src/file.rs
+++ b/libs/@local/config/src/file.rs
@@ -1,0 +1,52 @@
+use std::{fs, path::PathBuf};
+
+use error_stack::{Report, ResultExt as _};
+use figment::{
+    Metadata, Profile, Provider, Source,
+    error::Error as FigmentError,
+    value::{Dict, Map},
+};
+
+use crate::LoadError;
+
+/// A required TOML file contributing one configuration layer.
+pub(crate) struct File {
+    path: PathBuf,
+    values: Dict,
+}
+
+impl File {
+    pub(crate) fn read(path: PathBuf) -> Result<Self, Report<LoadError>> {
+        let text = fs::read_to_string(&path)
+            .change_context(LoadError::ReadFile)
+            .attach_with(|| path.display().to_string())?;
+        let values = toml::from_str(&text).map_err(|error: toml::de::Error| {
+            // TOML errors retain the source document and may quote its values. Keep only the
+            // position so neither the report nor its underlying frames can expose that input.
+            let mut report = Report::new(LoadError::ParseFile).attach(path.display().to_string());
+            if let Some(prefix) = error.span().and_then(|span| text.get(..span.start)) {
+                let line = prefix.bytes().filter(|byte| *byte == b'\n').count() + 1;
+                let column = prefix
+                    .rsplit('\n')
+                    .next()
+                    .map_or(1, |line| line.chars().count() + 1);
+                report = report.attach(format!("at line {line}, column {column}"));
+            }
+            report
+        })?;
+
+        Ok(Self { path, values })
+    }
+}
+
+impl Provider for File {
+    fn metadata(&self) -> Metadata {
+        Metadata::named("file")
+            .source(Source::File(self.path.clone()))
+            .interpolater(|_profile, keys| keys.join("."))
+    }
+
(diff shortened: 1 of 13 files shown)


libs/@local/config/src/file.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::LoadError;
+
+/// A required TOML file contributing one configuration layer.
+pub(crate) struct File {
+    path: PathBuf,
+    values: Dict,
+}
+
+impl File {
+    pub(crate) fn read(path: PathBuf) -> Result<Self, Report<LoadError>> {
```
indietyp, line 19: Have you thought about moving to `camino` instead of using `std::path`? I have been using it in, e.g., atlas (and hashql compiletest), and cargo uses it too, and it makes your life _so much easier_. It's a `PathBuf` but guaranteed utf-8, derefs to `Path`, etc, and also implements `Display`, etc.

TimDiekmann: What would `camino` provide us here? We essentially forward the path to `fs::read_to_string`, which already takes an std-`PathBuf`. The only UTF-8 relevant code we use here is `.attach()` with a `path.display().to_string()` which properly does printing a string without erroring.

indietyp: I always feel like it's a cumulative effect, many small things that make it easier but the reason *why* I suggested it is precisely the attach. We usually try to avoid bare attaches (especially in cases of potential re-use), as we also capture the semantic reason which we lose here. Outside of eg FilePath newtype attaching the actual type instead of prematurely rendering on attach would already help a lot.

TimDiekmann: I don't see the argument for `Utf8PathBuf` here, but an argument for using a newtype. I chose the (probably better) third route: put it on the error itself. The path is known at construction time, so it should live next to it.

indietyp (approved):

PR #9570, opened 2026-09-07
FE-1633: Give every dev server a port of its own

kube:
> [!IMPORTANT]
> Dev tooling, nothing user-facing.

## Summary

Before this PR, two pairs of dev servers wanted the same port. Petrinaut docs and the Brunch chat server both defaulted to 4321, and the chat server binds it with `strictPort` as one half of its contract with the panel, so whichever started second lost. Both Storybooks asked for 6006. Refractive's Storybook and ds-components' Ladle pinned their ports in the command and the config, so neither could take the port a preview task assigned it.

Every dev server now binds a port of its own, and a variable moves any of them: `PORT` for the single servers, `BRUNCH_CHAT_PORT` and `BRUNCH_PANEL_PORT` for the pair. `AGENTS.md` carries the map, so the next server added has somewhere to look before picking a number.

## Links

- Ticket [FE-1633](https://linear.app/hash/issue/FE-1633)

## Changes

- Petrinaut docs default to 4322
  > 4321 stays free for the Brunch chat server, which cannot move: its port and the panel's are a contract between the pair, declared `strictPort` at both ends.
- Refractive Storybook defaults to 6007
  > 6006 belongs to the Petrinaut Storybook. The two are separate component sets and get run together.
- Refractive Storybook and ds-components Ladle read `PORT`
  > Both hardcoded a port, so a preview that assigned another one got a server listening elsewhere.
  > `${PORT:-6007}` and `Number(process.env.PORT ?? 61000)` match how the other dev tasks already read it.
- Brunch listen ports come from `BRUNCH_CHAT_PORT` and `BRUNCH_PANEL_PORT`
  > Both were literals, so the pair could only run on 4321 and 4915.
  > `BRUNCH_CHAT_ORIGIN` moved what the panel assumed about the chat server, never what either bound.
- Brunch keeps `strictPort` on both servers
  > Drift is what the pair cannot survive: the panel proxies `/agents/chat` to a literal chat origin, and a panel on an unexpected port serves the editor where nobody is looking.
  > Naming the ports makes an explicit move possible without making an accidental one silent.
- Turbo forwards the port variables to the dev tasks
  > Strict environment mode forwards only the variables a task's `turbo.json` lists, so `turbo run dev --filter` stayed on the defaults.
  > Every dev task that reads `PORT` now declares it, `BRUNCH_CHAT_PORT` reaches both Brunch tasks and `BRUNCH_PANEL_PORT` the panel's.
- `.claude/launch.json` follows the docs to 4322
  > The three preview tasks now name three ports no other server wants.
- `AGENTS.md` gains the port tables
  > One lists every dev server and its port, with the ports the Brunch contract and the Optimizer container fix.
  > A second lists the ports the compose stack and the graph take, with the variable that moves each.

## Test coverage

- Manual, `@apps/petrinaut-docs`:
  > Astro binds 4322 through the preview task and leaves 4321 unbound.
- `local-dev-origins.test.ts`:
  > An overridden port moves the listener, the chat origin and the proxy target together.
  > A value that is not a usable port falls back, and `strictPort` survives the override.
  > Both Brunch dev tasks list their port variables in `turbo.json`.
- Manual, yarn script shell:
  > `${PORT:-6007}` expands to 6007 unset and to the assigned port when set.
- Existing lint and typecheck suites for `@apps/petrinaut-docs`, `@hashintel/refractive` and `@hashintel/ds-components`.

## How to test

- `yarn workspace @apps/petrinaut-docs dev`
  > Expect Astro on 4322
- `lsof -nP -iTCP:4321 -sTCP:LISTEN`
  > Expect no listener
- `yarn workspace @apps/brunch-agent dev` alongside it
  > Expect chat server on 4321, no strictPort failure
- `yarn workspace @hashintel/refractive dev`
  > Expect Storybook on 6007
- `PORT=61234 yarn workspace @hashintel/ds-components dev:ladle`
  > Expect Ladle on 61234
- `PORT=61234 yarn exec turbo run dev:ladle --filter @hashintel/ds-components`
  > Expect Ladle on 61234 through Turbo as well
- `BRUNCH_CHAT_PORT=4331 BRUNCH_PANEL_PORT=4925 yarn dev:brunch`
  > Expect chat on 4331, panel on 4925, chat reachable from the panel

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,12 +55,38 @@ Every dev server binds a port of its own, so any set of them can run at once:
 | 4321  | `@apps/brunch-agent` chat -- `strictPort`, paired with the panel |
 | 4322  | `@apps/petrinaut-docs`                                           |
 | 4915  | `@apps/brunch-agent` panel -- `strictPort`                       |
+| 5001  | `@apps/hash-api`                                                 |
 | 5173  | `@apps/petrinaut-website`                                        |
 | 6006  | `@hashintel/petrinaut` Storybook                                 |
 | 6007  | `@hashintel/refractive` Storybook                                |
 | 61000 | `@hashintel/ds-components` Ladle                                 |
 
-`PORT` overrides the default for every one of them except the two Brunch servers, which take `BRUNCH_CHAT_PORT` and `BRUNCH_PANEL_PORT` instead: one variable each, because the pair binds two ports and the panel proxies to whatever the chat variable names. Set them on `yarn dev:brunch`, which passes its environment to both servers. Turbo runs a dev task in strict environment mode and forwards only the variables its `turbo.json` lists in `env` or `passThroughEnv`, so a dev server that reads a port variable declares it there. `.claude/launch.json` names the servers a Claude Code session can preview; keep its `port` in step with the table when a default moves.
+`PORT` overrides the default for every one of them except the Optimizer service, which the dev script publishes at 4004 and compose moves with `PETRINAUT_OPT_PORT`, and the two Brunch servers, which take `BRUNCH_CHAT_PORT` and `BRUNCH_PANEL_PORT` instead: one variable each, because the pair binds two ports and the panel proxies to whatever the chat variable names. Set them on `yarn dev:brunch`, which passes its environment to both servers. Turbo runs a dev task in strict environment mode and forwards only the variables its `turbo.json` lists in `env` or `passThroughEnv`, so a dev server that reads a port variable declares it there. `.claude/launch.json` names the servers a Claude Code session can preview; keep its `port` in step with the table when a default moves.
+
+The compose stack and the graph take these, so a new dev server stays off them as well. Where a variable moves the port, the table names it:
+
+| Port       | Service                     | Moved by                       |
+| ---------- | --------------------------- | ------------------------------ |
+| 1025       | Mailslurper SMTP            |                                |
+| 3001       | Grafana                     |                                |
+| 3100       | Temporal UI                 | `HASH_TEMPORAL_UI_PORT`        |
+| 4000       | Graph API HTTP              | `HASH_GRAPH_HTTP_PORT`         |
+| 4001       | Graph admin API             | `HASH_GRAPH_ADMIN_PORT`        |
+| 4002       | Graph HaRPC                 | `HASH_GRAPH_RPC_PORT`          |
+| 4003       | Graph Atlas                 | `HASH_GRAPH_ATLAS_PORT`        |
+| 4040       | Pyroscope                   |                                |
+| 4317       | OpenTelemetry collector     |                                |
+| 4433, 4434 | Kratos public and admin API |                                |
+| 4436, 4437 | Mailslurper web and API     |                                |
+| 4444, 4445 | Hydra public and admin API  |                                |
+| 4455       | Type fetcher                | `HASH_GRAPH_TYPE_FETCHER_PORT` |
+| 5432       | Postgres                    | `POSTGRES_PORT`                |
+| 6379       | Redis                       |                                |
+| 7233       | Temporal server             | `HASH_TEMPORAL_SERVER_PORT`    |
+| 8200       | Vault                       | `HASH_VAULT_PORT`              |
+| 9000, 9001 | MinIO API and console       |                                |
+
+The Brunch agent's `start` and `start:test` scripts and its production image listen on 3002, where the integration tests expect it; the dev pair uses 4321 and 4915.
 
 ### Starting Services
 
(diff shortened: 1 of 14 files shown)


AGENTS.md
```diff
@@ -44,6 +44,24 @@ Package-specific standing instructions live in that package’s `AGENTS.md`. On-
...
 - Frontend only: `yarn dev:frontend`
 
+### Dev server ports
+
+Every dev server binds a port of its own, so any set of them can run at once:
+
+| Port  | Server                                                           |
+| ----- | ---------------------------------------------------------------- |
+| 3000  | `@apps/hash-frontend`                                            |
+| 4004  | Petrinaut Optimizer service (`--with-optimizer-service`)         |
```
TimDiekmann, line 54: Could you extend it with other ports please?
- 3001: Grafana
- 3100: Temporal UI
- 3002: Brunch agent
- 4000: Graph API HTTP
- 4001: Graph Admin API
- 4002: Graph Api HaRPC
- 4003: Graph Atlas
- 4455: Type Fetcher
- 5432: Postgres
- 6379: Redis
- 8200: Vault

kostandinang (dismissed):
The new port variables aren't declared in the Turborepo config, so under strict env mode (the default here) they're stripped before the scripts read them:

- `libs/@hashintel/refractive/turbo.json`: `PORT` missing from `dev`
- `libs/@hashintel/ds-components/turbo.json`: `PORT` missing from `dev` and `dev:ladle`
- `apps/brunch-agent/turbo.json`: `BRUNCH_CHAT_PORT` missing from `dev`, `BRUNCH_PANEL_PORT` from `petrinaut:dev` (also noted by Bugbot)

The documented `yarn workspace …` / `yarn dev:brunch` paths bypass Turbo so they work, but `turbo run dev --filter …` (as AGENTS.md suggests) won't. `@apps/petrinaut-docs` already handles this with `"env": ["PORT"]`.

apps/brunch-agent/src/http/local-origins.ts
```diff
@@ -1,14 +1,34 @@
...
   host: "127.0.0.1",
-  port: 4321,
+  port: listenPort("BRUNCH_CHAT_PORT", 4321),
   strictPort: true,
 } as const;
 
 export const localPanelListen = {
   host: "127.0.0.1",
-  port: 4915,
+  port: listenPort("BRUNCH_PANEL_PORT", 4915),
```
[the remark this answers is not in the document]
kube: Fixed in 22cf2f3016: `dev` now lists `BRUNCH_CHAT_PORT`, and `petrinaut:dev` lists both variables, since the panel derives its proxy target from the chat port. `local-dev-origins.test.ts` pins both allowlists, and a Turbo dry run shows the variables under `passthrough` for each task.

AGENTS.md
kube: Added in dc9565bccf as a second table under the dev servers, with the variable that moves each port where one exists. It also carries Kratos 4433/4434 and Hydra 4444/4445 from the dev profile, and hash-api 5001 joins the dev-server table. 3002 is noted as the port of the Brunch agent's `start`/`start:test` scripts and production image rather than listed as a dev server.

kube:
@kostandinang fixed in 22cf2f3016: every dev task that reads a port variable now declares it in `passThroughEnv`. That is refractive, ds-components and the two Brunch tasks you listed, plus petrinaut, petrinaut-website, hash-api and hash-frontend, which had the same gap under the table's `PORT` claim. `petrinaut:dev` also gets `BRUNCH_CHAT_PORT`, since the panel derives its proxy target from the chat port. Verified per task with `turbo run <task> --filter <pkg> --dry-run=json`.

kostandinang (approved):

AGENTS.md
```diff
@@ -44,6 +44,50 @@ Package-specific standing instructions live in that package’s `AGENTS.md`. On-
 - Backend only: `yarn dev:backend` or `yarn dev:backend:api`
 - Frontend only: `yarn dev:frontend`
 
+### Dev server ports
```
indietyp, line 47: The graph itself should live on the dev server side, not in the compose stack, and maybe, longer term, we should move this into the dev docs of some kind? Seems useful outside of the AGENTS.md (I also don't know if I am a huge fan of adding this block _to every system prompt_; it seems more useful as a skill that teaches how to start a dev server, which ones exist, etc.

indietyp (approved):
Infra OK

AGENTS.md
kube: Yeah would definitely be better to have a dedicated .md for that, and just reference it from AGENTS.md in case it's needed.

Given I got two approvals I'll delay this other change, but definitely something we can clean.

My goal was just to be able to work with multiple threads in parallel without all dev servers colliding.

AGENTS.md
TimDiekmann: Great, thanks for double checking!

PR #9575, opened 2026-09-07
FE-1704: Give Codex the dev servers Claude Code already has

kube:
> [!IMPORTANT]
> Field names in this file are inferred, not documented. See Known issues.

## Summary

`.claude/launch.json` names the dev servers a Claude Code session can start and preview. Codex reads its configuration from `.codex` at the repository root, so it has none of them, and anyone working there retypes each server into a terminal.

`.codex/environment.toml` declares the same servers as actions, plus the lint and typecheck a change has to pass, and a setup script that runs the repository's own getting-started steps when Codex creates a worktree.

## Links

- Ticket [FE-1704](https://linear.app/hash/issue/FE-1704)

## Changes

- Four actions, matching the servers `.claude/launch.json` already names
  > Petrinaut Storybook, the website, the architecture docs through turbo, and the lint and typecheck trio.
  > Commands are the ones the two files share, so a server behaves the same whichever tool starts it.
- Setup script installs the pinned toolchains and the workspace
  > `mise trust && mise install`, then `yarn install`, the steps the README gives for a fresh clone.
- No ports in the file
  > An action runs a command in the terminal rather than being handed a port, so each server picks its own and `AGENTS.md` stays the one map.
- Configuration sits in `.codex/environment.toml`, beside `.codex/config.toml` rather than inside it
  > A wrong guess in a file of its own is ignored. The same guess inside `config.toml` would sit in the file the CLI loads for a trusted project.

## Known issues

- The schema is inferred
  > The [documentation](https://learn.chatgpt.com/docs/environments/local-environment) describes a setup script with per-platform overrides and actions carrying a name, a command and an icon, and says the configuration lives in `.codex`. It names neither the file nor its fields, and says the app's settings pane writes it.
  > Adding one action in the Codex app will produce the real file. Reconcile this one with it, keeping the commands.
- Icon names are placeholders
  > `book`, `browser`, `document` and `check` are guesses at whatever set the app offers.

## Test coverage

- Manual, TOML parse:
  > The file parses, and yields the setup script and the four actions.
- Manual, command existence:
  > Every command names a script that exists: three workspace `dev` scripts and the three root lint scripts.
- `yarn lint:format`:
  > Passes with the file in place.

## How to test

- Open the repository in the Codex app
- Settings > local environment
  > Expect the four actions and the setup script, if the inferred field names are right
- Add one action by hand in the same pane
  > Expect the app to write its own file, which names the real schema
- Run any action from the top bar
  > Expect the same server a `.claude/launch.json` task starts

### DIFF
--- a/.codex/environment.toml
+++ b/.codex/environment.toml
@@ -12,17 +12,17 @@ yarn install
 # The launcher selects an available port and exposes it as PORT, matching the
 # autoPort behavior of .claude/launch.json.
 [[actions]]
-command = "node .codex/run-with-auto-port.mjs yarn workspace @hashintel/petrinaut dev"
+command = "node .codex/run-with-auto-port.mjs 6006 yarn workspace @hashintel/petrinaut dev"
 icon    = "run"
 name    = "Petrinaut Storybook"
 
 [[actions]]
-command = "node .codex/run-with-auto-port.mjs yarn workspace @apps/petrinaut-website dev"
+command = "node .codex/run-with-auto-port.mjs 5173 yarn workspace @apps/petrinaut-website dev"
 icon    = "run"
 name    = "Petrinaut website"
 
 [[actions]]
-command = "node .codex/run-with-auto-port.mjs yarn exec turbo run dev --filter @apps/petrinaut-docs"
+command = "node .codex/run-with-auto-port.mjs 4322 yarn exec turbo run dev --filter @apps/petrinaut-docs"
 icon    = "run"
 name    = "Petrinaut architecture docs"
 
--- a/.codex/environment.toml
+++ b/.codex/environment.toml
@@ -1,50 +1,32 @@
-# Codex local environment for this repository: what to run when a worktree is
-# created, and the dev servers worth one click from the top bar.
-#
-# The same servers as `.claude/launch.json`, started by the same commands.
-# Ports are not repeated here, because an action runs a command in the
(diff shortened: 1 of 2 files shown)


.codex/environment.toml
```diff
@@ -0,0 +1,32 @@
+# Common HASH actions for Codex worktrees.
+name    = "HASH"
+version = 1
+
+# Run the repository's setup steps whenever Codex creates a worktree.
+[setup]
+script = """
+mise trust && mise install
+yarn install
```
TimDiekmann, line 9: Just a small warning here: Codex. happily creates work trees and running `yarn install` (+ potentially `cargo`) might result in quite high disk usage.

.codex/environment.toml
```diff
@@ -0,0 +1,32 @@
...
+icon    = "run"
+name    = "Petrinaut website"
+
+[[actions]]
+command = "node .codex/run-with-auto-port.mjs 4322 yarn exec turbo run dev --filter @apps/petrinaut-docs"
+icon    = "run"
+name    = "Petrinaut architecture docs"
+
+[[actions]]
+command = "yarn lint:tsc && yarn lint:eslint && yarn lint:format"
```
TimDiekmann, line 30: `yarn fix:eslint` and `yarn fix:format` might even fix the errors if auto-fixable

PR #9589, opened 2026-09-08
FE-1631: Fix Node ESM imports in Petrinaut Vercel Functions

kostandinang:

Petrinaut's Voice and oEmbed Vercel Functions currently fail before their handlers run because the emitted ESM retains extensionless relative imports. Native Node ESM resolution does not append `.js`, so Vercel reports `ERR_MODULE_NOT_FOUND` and the Voice config endpoint returns 500.

This repairs the complete local runtime import graphs and adds a dedicated NodeNext TypeScript project that recursively checks every Petrinaut Vercel Function entrypoint using native Node ESM resolution rules.

## 🔗 Related links

- [FE-1631](https://linear.app/hash/issue/FE-1631/fix-node-esm-imports-in-petrinaut-vercel-functions) _(internal)_
- [PR #9564](https://github.com/hashintel/hash/pull/9564), where the deployed Voice failure was observed
- [Node.js ESM mandatory file extensions](https://nodejs.org/docs/latest-v22.x/api/esm.html#mandatory-file-extensions)

## 🚫 Blocked by

- [x] Nothing

## 🔍 What does this change?

- Adds `.js` to relative imports used by the oEmbed, Voice config, and Voice Realtime call runtime graphs.
- Adds `api/tsconfig.json` with NodeNext resolution and `verbatimModuleSyntax`, recursively covering current and future Vercel Function entrypoints.
- Runs both the browser and API TypeScript projects from `lint:tsc`, while limiting ambient Node types to the API project.

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

This restores function loading only. Petrinaut's current policy still disables OpenAI Voice in production; enabling production Voice requires the separate authentication/security work tracked by FE-1622.

## 🐾 Next steps

- Smoke-test `/api/voice/config`, `/api/voice/realtime-call`, and `/api/oembed` on the Vercel preview.

## 🛡 What tests cover this?

- Dedicated NodeNext TypeScript checking for all Vercel Function entrypoints and their imported local modules.
- Existing Voice config, Voice Realtime call, and oEmbed endpoint unit tests.
- Complete `@apps/petrinaut-website` unit-test suite.
- Dependency-aware production build, TypeScript checks, and Oxlint.

## ❓ How to test this?

1. Run `mise exec -- turbo run build --filter='@apps/petrinaut-website'`.
2. Run `mise exec -- yarn workspace @apps/petrinaut-website test:unit`.
3. Run `mise exec -- yarn workspace @apps/petrinaut-website lint:tsc`.
4. On the Vercel preview, confirm Voice config returns JSON, an invalid Realtime request returns a controlled 4xx, and a valid oEmbed request returns JSON rather than `FUNCTION_INVOCATION_FAILED`.

## 📹 Demo

Not applicable: this fixes serverless module loading and has no visual change.


kostandinang:
Preview smoke test passed on [the Petrinaut deployment](https://petrinaut-git-kostandin-fe-1631-fix-node-esm-imports-in-4637a2.stage.hash.ai):

- `GET /api/voice/config` → `200` with `{"available":true,"connectionTimeoutMs":15000}`
- `GET /api/voice/realtime-call` → controlled `405` with `Allow: POST`, request ID, and server timing
- Valid `GET /api/oembed` → `200` with the expected JSON payload

None of the routes produced `FUNCTION_INVOCATION_FAILED` or `ERR_MODULE_NOT_FOUND`.

apps/petrinaut-website/src/server/vercel-function-entrypoints.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import { describe, expect, it } from "vitest";
+
+const appRoot = fileURLToPath(new URL("../..", import.meta.url));
+const entrypointPaths = [
+  "api/chat.ts",
+  "api/oembed.ts",
+  "api/voice/config.ts",
+  "api/voice/realtime-call.ts",
+] as const;
```
kube: A grep in the folder here would be better, given Vercel automatically discovers these endpoints, and we want this to save us for new endpoints in the future too.

But the better solution here would be to create a second `api/tsconfig.json`, in which we would have something like:

```json
"compilerOptions": {
  "module": "NodeNext",
  "moduleResolution": "NodeNext",
  "noEmit": true,
  "verbatimModuleSyntax": true
}
```

And then run it during "lint:tsc":

```sh
tsgo --noEmit && \
tsgo --noEmit --project api/tsconfig.json
```

kube (commented):
Thanks for discovering that.

Please check my comment on adding another tsconfig.json rather than creating unit tests to verify compilation.

apps/petrinaut-website/package.json
```diff
@@ -42,6 +42,7 @@
     "@fast-check/vitest": "0.4.1",
     "@tanstack/router-generator": "1.167.32",
     "@tanstack/router-plugin": "1.168.34",
+    "@types/node": "22.18.13",
```
kube, line 45: Please make sure `@types/node` is not imported/loaded outside of `api/` folder.

Check my previous comment about creating a dedicated `tsconfig.json` for `api/` folder.

kostandinang: Scoped `@types/node` to `api/tsconfig.json`. The root project now sets `types: []` and no longer includes `api`, so browser/source checking does not load Node ambient types.

apps/petrinaut-website/src/server/vercel-function-entrypoints.test.ts
kostandinang: Replaced the bespoke entrypoint test with `api/tsconfig.json` using NodeNext resolution and `verbatimModuleSyntax`. `lint:tsc` now checks both projects, and the API project recursively includes current and future endpoints.

apps/petrinaut-website/package.json
kube: Perfect thank you!

kube (approved):
Thank you!
This will also solve the oEmbed issue I had to investigate. (certainly)

apps/petrinaut-website/tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "resolveJsonModule": true,
     "noEmit": true,
     "skipLibCheck": true,
-    "isolatedModules": true
+    "isolatedModules": true,
+    "types": []
```
kube, line 17: I don't remember what this does with `@types/react` or `@types/react-dom`.

(Particularly when writing JSX in a file that does not import anything from React)

If it does not break anything, it's good enough.

kostandinang: Confirmed this keeps Node types limited to the API while React types continue to load through the JSX runtime.

PR #9594, opened 2026-09-09
FE-1611/FE-1636: Replace the Preview's view-only badge with a full-view link, and stop its text being selectable

kube:
## Summary

Before this PR, the top-right corner of an embedded net carried a "VIEW ONLY" badge. It told the reader what they could not do, in the one corner of the frame with room for something they can do, and the embed offered no way through to the full model. Dragging across the embed also highlighted labels and panel text, which a viewer never wants: a drag there pans the canvas or draws a selection box.

`PetrinautPreview` takes a `fullViewUrl` and renders it in the badge's place as a **Full view** link that opens in a new tab. The website's embed passes the model's own page, carrying the scenario and the selected item, so the full page opens where the reader was. Text in the preview is no longer selectable, and a field that still takes typing keeps its own selection.

https://github.com/user-attachments/assets/1a3ca685-2cd2-4b10-bce8-d5d5f63eacf7

## Links

- Ticket [FE-1611](https://linear.app/hash/issue/FE-1611)
- Ticket [FE-1636](https://linear.app/hash/issue/FE-1636)
- Docs [Embedded Preview](https://github.com/hashintel/hash/blob/claude/fe-1611-embed-open-in-full-view/libs/@hashintel/petrinaut/docs/preview.md)

#### Preview

- New `fullViewUrl` prop, rendered as a **Full view** link in the header
  > Absent prop renders nothing, so a host that has no full-size page keeps the header it has.
  > The link is an anchor with `target="_blank"`, so an embed on someone else's page never navigates the page around it.
- "VIEW ONLY" badge removed
- Text in the embed is not selectable
  > `user-select: none` on the embed root, with an enabled `input` or `textarea` put back to `text` so a field that still takes typing stays usable.
  > Blink ignores `user-select` inside a form control, so in Chromium a field's text selects either way; the rule governs the engines that honour it.
- Label hides under 480px, icon and tooltip carry the link
  > Matches how the Quick Simulation label already folds. An explicit `aria-label` names the link at every width.

#### Website

- Embed page passes the model's full page as `fullViewUrl`
  > Root-relative, so it resolves against the site inside a host's iframe.
  > The reader's canonical search string rides along, so scenario and selection survive the jump.

## Test coverage

- Existing `@hashintel/petrinaut` and `@apps/petrinaut-website` unit suites
- Verified in the running embed:
  > Link resolves to `/examples/<slug>?scenario=<id>`, opens the full page on that scenario, and the embed root computes `user-select: none`.

## How to test

- Open [SIR embed](https://petrinaut-git-claude-fe-1611-embed-open-in-full-view.stage.hash.ai/embed/examples/sir-epidemic-model) on Vercel
  > Expect **Full view** in the top-right, no "VIEW ONLY" badge
- Drag across the canvas and across the net's labels
  > Expect no text highlight
- Select a place, then Quick Simulation > pick another scenario
- Click **Full view**
  > Expect a new tab on the full page, same scenario, same place selected
- Narrow the window under 480px
  > Expect the link to keep its icon and drop the word

### DIFF
--- a/.changeset/preview-full-view-link.md
+++ b/.changeset/preview-full-view-link.md
@@ -2,4 +2,4 @@
 "@hashintel/petrinaut": patch
 ---
 
-`PetrinautPreview` takes a `fullViewUrl`, and offers it in the header as a **Full view** link that opens in a new tab. The link replaces the "View only" badge, and text in the preview is no longer selectable.
+`PetrinautPreview` takes a `fullViewUrl`, and offers it in the header as a **Full view** link that opens in a new tab. The link replaces the "View only" badge, and text in the preview is no longer selectable, apart from a field that still takes typing.
--- a/.changeset/preview-full-view-link.md
+++ b/.changeset/preview-full-view-link.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+`PetrinautPreview` takes a `fullViewUrl`, and offers it in the header as a **Full view** link that opens in a new tab. The link replaces the "View only" badge, and text in the preview is no longer selectable.
--- a/.changeset/rounded-preview-playback-bar.md
+++ b/.changeset/rounded-preview-playback-bar.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+The Preview's playback bar has rounded corners, matching the zoom controls and the minimap that float over the same canvas.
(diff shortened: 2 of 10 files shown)


kube:
Confirmed and fixed in `49da2aa568`.

Reproduced the block before changing anything: a frame sandboxed `allow-scripts allow-same-origin` swallows the `target="_blank"` navigation, and the same frame with `allow-popups` opens the tab. So the link was dead in exactly the third-party embeds it exists for.

The oEmbed iframe now carries `allow-popups allow-popups-to-escape-sandbox`. The second flag matters as you say: without it the full page would inherit the embed's sandbox rather than running as itself. The endpoint's expected-HTML tests and the README's sandbox paragraph move with it.

drdma (approved):

PR #9596, opened 2026-09-09
FE-1637: Show read-only property values as text instead of disabled inputs

kube:
## Summary

Before this PR, a read-only net still rendered every property value in the control that edits it, greyed out. A name arrived as a disabled text box, a description as a disabled textarea, and the field labels dimmed along with them, so the values a reader came for were the lowest-contrast text on the panel. Two fields in the arc panel already sidestepped this and printed their value as text.

The properties panel now renders read-only values the way those two arc fields do: the value as text, at full contrast, under a label that no longer greys out. This reaches the embedded Preview, the website's full example view, and the editor whenever the net is read-only, which includes Simulate mode and any live simulation.

A place's properties on a read-only net, before and after

## Links

- Ticket [FE-1637](https://linear.app/hash/issue/FE-1637)
- Docs [Embedded Preview](https://github.com/hashintel/hash/blob/claude/fe-1637-readonly-values-as-text/libs/@hashintel/petrinaut/docs/preview.md)

#### Shared

- New `PropertyValue`, wrapping one field's editing control
  > Renders its children on an editable net and the value as text on a read-only one, reading the same `useIsReadOnly` every panel already consults.
  > A missing or blank value renders a caller-chosen stand-in, defaulting to "None".
- `DraftFieldInput` and the description field render through it
  > That covers the name and description of transitions, types, parameters, differential equations, places, and component instances in one place.
- Read-only labels keep full contrast
  > `Form.Field` no longer receives `disabled` when the only reason was the net being read-only. A read-only field is not a disabled one.

#### Panels

- Place name, accepted token type, arc type, arc weight, differential-equation type, and component-instance name read as text
  > Arc type labels now come from one map shared with the select's options, so the two cannot name a type differently.
- Toggles and checkboxes are unchanged
  > A checkbox already shows its value by being ticked, so it reads correctly without a text form.

## Known issues

- Complex editors keep their controls on a read-only net
  > The type-element list, the initial-state grid, and the arc tables are structures rather than single values, and each needs its own read-only presentation.

## Test coverage

- `property-value.test.tsx`:
  > Control on an editable net, text on a read-only one, and the stand-in for undefined, null, and blank values.
- Existing `@hashintel/petrinaut` unit suite
- Verified in the running embed:
  > Selecting a place leaves no text input in the panel, and name, description, and token type render as text.

## How to test

- Open [SIR embed](https://petrinaut-git-claude-fe-1637-readonly-values-as-text.stage.hash.ai/embed/examples/sir-epidemic-model) on Vercel
- Select the Infected place
  > Expect name and description as text, labels at full contrast, no input boxes
- Select an arc
  > Expect Type and Weight as text
- Open [SIR full view](https://petrinaut-git-claude-fe-1637-readonly-values-as-text.stage.hash.ai/examples/sir-epidemic-model) and select a transition
  > Expect name and description as text
- Switch to Simulate in the full view and select a place
  > Expect the same text presentation, since the net is read-only there

### DIFF
--- a/.changeset/readonly-property-values-as-text.md
+++ b/.changeset/readonly-property-values-as-text.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+The properties panel shows the values of a read-only net as text instead of as disabled inputs. Names, descriptions, and single-value fields read at full contrast, and their labels no longer grey out.
--- a/.changeset/readonly-property-values-as-text.md
+++ b/.changeset/readonly-property-values-as-text.md
@@ -2,4 +2,4 @@
 "@hashintel/petrinaut": patch
 ---
 
-The properties panel shows the values of a read-only net as text instead of as disabled inputs. Names, descriptions, and single-value fields read at full contrast, and their labels no longer grey out.
+The properties panel shows a read-only net's names, descriptions, and the single-value fields of places, arcs, differential equations and component instances as text instead of as disabled inputs. They read at full contrast, and their labels no longer grey out. Toggles, checkboxes and the list editors keep their controls.
--- a/libs/@hashintel/petrinaut/docs/preview.md
+++ b/libs/@hashintel/petrinaut/docs/preview.md
@@ -21,8 +21,9 @@ looking at.
 
 Select a place, transition, arc, or other supported item to inspect it. The
 preview presents the same property information as Petrinaut's inspector,
-without source editing or other authoring controls. Values you cannot change
-are shown as text rather than as filled-in fields. On a wide embed the
+without source editing or other authoring controls. A field's name, description and
+type are shown as text rather than as filled-in fields; the remaining controls
+are still rendered, and simply do nothing. On a wide embed the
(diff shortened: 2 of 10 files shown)


kube:
All four addressed in `20b84dbf08`.

**Description height after editing resumes** (also raised by Bugbot): real. The sizing `useLayoutEffect` keyed on `sourceId` alone, and the textarea is now unmounted while the net is read-only, so returning to Edit with the same item selected remounted it unmeasured. `isReadOnly` joins the dependencies.

**Stale draft error beside the canonical value**: real, and it reached three places. `DraftFieldInput` now passes errors only while editable, and the place and component-instance panels hide their own error nodes the same way. The draft itself is untouched, so the error returns with it.

**`trim()` eating authored whitespace**: right, and it defeated the point of `pre-wrap`. The trim now only decides whether the value is blank; the text renders as authored.

**Docs broader than the behaviour**: correct. The sentence in `preview.md` and the changeset now name what is converted -- names, descriptions, and the single-value fields -- and say the remaining controls stay as controls.

drdma (approved):

PR #9598, opened 2026-09-09
FE-1638: Show elapsed and total time in the playback readout

kube:
## Summary

Before this PR, the playback readout led with the frame index and count, over the elapsed time in a smaller, lighter size. A frame index is an artefact of how a run is stored: nothing a reader can act on, and on a long run it is four digits of noise wider than the numbers that matter. The screenshot on the ticket shows "4539 / 4611" over "45.380s".

Readout now shows elapsed time over total time and nothing else, in both the editor's toolbar and the embedded Preview's bar. Times print at the precision the run's own time step carries, so a run stepped at 0.01 reads 3.07 over 3.70s rather than padding three decimals it does not have. The scrubber beside it sizes itself to the space the bar has, in place of a fixed 300px.

https://github.com/user-attachments/assets/7cd031f7-3e7d-483d-9bc5-bf8fab44fbea

## Links

- Ticket [FE-1638](https://linear.app/hash/issue/FE-1638)
- Docs [Simulation](https://github.com/hashintel/hash/blob/claude/fe-1638-playback-time-readout/libs/@hashintel/petrinaut/docs/simulation.md)
- Docs [Embedded Preview](https://github.com/hashintel/hash/blob/claude/fe-1638-playback-time-readout/libs/@hashintel/petrinaut/docs/preview.md)

## Changes

- Playback readout is elapsed time over total time, stacked over two lines
  > Elapsed carries the emphasis, the total sits under it in lighter grey.
  > The frame index, the frame count, and the "Frame" label are gone.
- Scrubber sizes itself to the space the bar has
  > `clamp(140px, 24vw, 420px)` with `flex: 1 1 auto`: `clamp` tracks the viewport continuously, so a resize slides its width instead of stepping at a breakpoint, and flex takes any slack the row has left over.
  > Inside an embed `vw` is the iframe's own width, so a host resizing its frame resizes the scrubber with it.
- New `playback-time.ts` derives and formats both numbers
  > `playbackTimes` puts the first frame at t=0, so a run of n frames ends at (n - 1) steps.
  > `formatPlaybackTimes` prints the decimals the step itself carries, found by asking which fixed-point rendering round-trips, and capped at milliseconds.
  > Bucketing by magnitude instead printed "1.8" for a true 1.75 on a run stepped at 0.25.
- Readout carries an accessible label
  > `Elapsed 3.07 of 3.70s`, since the visible text is two spans of digits.
- Fixed width holds the controls either side still
  > 58px, 54px on the compact embed bar, with tabular figures, so nothing shifts as the numbers count up.

## Test coverage

- `playback-time.test.ts`:
  > Frame-to-time mapping at the start and end of a run, the empty run, and the printed precision for each step size.
  > Steps of 0.25 and 0.025 pin the case that magnitude bucketing rounded into a time the run never visited.
- Existing `@hashintel/petrinaut` unit suite
- Verified in both surfaces:
  > The embed bar reads `3.07 / 3.70s` and the editor toolbar `3.45 / 4.00s`, with no "Frame" text left in either page.

## How to test

- Open [SIR embed](https://petrinaut-git-claude-fe-1638-playback-time-readout.stage.hash.ai/embed/examples/sir-epidemic-model) on Vercel
- Press Play
  > Expect elapsed over total time, no frame numbers
- Drag the scrubber
  > Expect the elapsed number to follow the pointer, the total to hold
- Open [SIR full view](https://petrinaut-git-claude-fe-1638-playback-time-readout.stage.hash.ai/examples/sir-epidemic-model) and press Play in the bottom toolbar
  > Expect the same readout
- Narrow the window while the run plays
  > Expect the scrubber to shrink with it, the readout to hold its width
- Narrow the window until the toolbar folds
  > Expect Play to stay, the readout to fold away with the scrubber

### DIFF
--- a/.changeset/playback-time-readout.md
+++ b/.changeset/playback-time-readout.md
@@ -2,4 +2,4 @@
 "@hashintel/petrinaut": patch
 ---
 
-The playback readout shows elapsed and total simulated time instead of the frame index and frame count, in both the editor's toolbar and the embedded Preview's bar. Times print at the precision the run's time step carries.
+The playback readout shows elapsed and total simulated time instead of the frame index and frame count, in both the editor's toolbar and the embedded Preview's bar. Times print at the precision the run's time step carries, stacked over two lines. The scrubber sizes itself to the space the bar has instead of a fixed width.
--- a/.changeset/playback-time-readout.md
+++ b/.changeset/playback-time-readout.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+The playback readout shows elapsed and total simulated time instead of the frame index and frame count, in both the editor's toolbar and the embedded Preview's bar. Times print at the precision the run's time step carries.
--- a/libs/@hashintel/petrinaut/docs/preview.md
+++ b/libs/@hashintel/petrinaut/docs/preview.md
@@ -41,7 +41,7 @@ The selected scenario's initial marking appears on the shared canvas before a
 run starts. Press **Play** in the compact bar at the bottom to start the run;
 the preview never starts it automatically. The same bar lets you pause, reset,
 choose from the playback speeds allowed by the embed, and scrub through the
-frames that have been produced.
+run. It reads out the elapsed time and the run's total time.
 
 As soon as frames arrive, the compact bar expands upward to show a small
 timeline. The timeline follows playback, lets you hover to inspect a series,
--- a/libs/@hashintel/petrinaut/docs/simulation.md
+++ b/libs/@hashintel/petrinaut/docs/simulation.md
@@ -90,16 +90,17 @@ A transition blocked only because its output place is at [capacity](drawing-a-ne
 
 The bottom toolbar provides playback controls:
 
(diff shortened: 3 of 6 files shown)


kube:
Both addressed in `c280ea3339`.

**`flex` cannot take slack in the editor bar**: correct, and the comment claimed otherwise. `CollapsibleGroup`'s content is `width: max-content`, so there is no free space to grow into there. The declaration stays because the embed's expanded bar is a different case: its row is `width: max-content` with `min-width: 100%`, so the bar's full width does create slack and the scrubber takes it. The comment now says which bar it applies to. Sizing in the editor comes from `clamp` alone, which is what the viewport-resize demo shows.

**`aria-label` on a generic div**: taken. Chrome does expose the name today -- the tree reports `role: generic, name: "Elapsed ...", ignored: false` -- but naming a generic is prohibited and other engines are free to drop it. The readout now carries `role="timer"`, which is nameable and, being live-off by default, does not announce on every frame.

drdma (approved):

PR #9498, opened 2026-09-02
SRE-1007: Make the Brunch agent transport boundary check fire in CI, and fix the violation

kostandinang:
## 🌟 What is the purpose of this PR?

Restore the Brunch transport boundary and make future violations fail at the earliest relevant CI gates. The Petrinaut website now consumes core-owned client-tool contracts directly, while the AI SDK transport remains independent of Brunch core, bindings, plugins, and substrates.

This proves that manifest violations fail during `yarn install` and that changes to every file scanned by the core architecture suite affect its Turbo task. It does not replace the source-level architecture checks or change runtime behavior.

## 🔗 Related links

- [SRE-1007](https://linear.app/hash/issue/SRE-1007/make-the-brunch-agent-transport-boundary-check-fire-in-ci-and-fix-the) _(internal)_
- [Failing CI job](https://github.com/hashintel/hash/actions/runs/33612937924/job/100192372248)
- [Dependency-introducing PR #9356](https://github.com/hashintel/hash/pull/9356)

## 🚫 Blocked by

None.

## 🔍 What does this change?

- Moves browser-safe client-tool constants, inferred types, and parsers to the existing `@hashintel/brunch-agent/client-tools` core subpath.
- Removes the transport's core dependency and client-tools re-export; website consumers now declare and import core directly.
- Declares the Brunch context and app scan roots as inputs to core's `test:unit` task, excluding generated and dependency directories.
- Adds matching Yarn and Oxlint boundary enforcement and enables constraint checks during installation.
- Updates the lockfile and prune graph fixture to represent the corrected dependency direction.

<details>
<summary>🏗️ Agent notes</summary>

### Mission contract

- **Imperative:** Repair the transport dependency direction and prevent the architecture check from being silently skipped when any governed package or app file changes.
- **Throughline:** `apps/petrinaut-website` reads browser contracts from Brunch core; `transport-aisdk` depends only on wire-level libraries. Package installation checks the manifest boundary, while Turbo selects the core architecture suite from every tree it scans.
- **Proof:** The full core suite and transport suite pass; affected website consumers pass tests, lint, typecheck, and build; a deliberately reintroduced transport-to-core dependency fails both `yarn constraints` and `yarn install`; Turbo's resolved inputs include the transport manifest and app source while excluding generated directories.
- **Constraints:** Preserve the existing host-facing tool contract and runtime behavior. Keep private package topology intact. Do not replace broader source-level architecture checks with manifest-only constraints.
[...]

PR #9625, opened 2026-09-09
SRE-1010: Move Brunch architecture checks out of core unit tests

lunelson:
-------------------------
- **ln/sre-1010-move-brunch-checks** :point_left:
- main

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

Remove Brunch architecture source walkers from unit tests and put each surviving contract at its native enforcement point. Authored import boundaries remain in package-local Oxlint configuration, manifest boundaries remain in Yarn constraints, and behavioral tests prove emitted behavior.

This removes the reverse `libs → apps` task relationship introduced by the SRE-1007 hotfix without replacing it with another cross-workspace scanner. Core and app unit-test tasks now own only files and behavior inside their packages.

## 🔗 Related links

- [SRE-1010](https://linear.app/hash/issue/SRE-1010/move-brunch-architecture-checks-out-of-core-unit-tests) _(internal)_
- [SRE-1007](https://linear.app/hash/issue/SRE-1007/make-the-brunch-agent-transport-boundary-check-fire-in-ci-and-fix-the) _(internal)_
- [PR #9498](https://github.com/hashintel/hash/pull/9498)
- [PR #9498 ownership review](https://github.com/hashintel/hash/pull/9498#discussion_r3913908431)

## 🚫 Blocked by

None.

## 🔍 What does this change?

- Deletes family-wide Vitest checks that scanned source files and mirrored package metadata, import rules, schema allowlists, filenames, task scripts, and reviewer-maintained inventories.
- Removes the temporary architecture-test workspace and the subsequent repo-chores checker/CI task; no parallel static-policy surface remains.
- Deletes the app-local source walker. Agent registration remains covered by the emitted-bundle test, now using `ChatAgent.agentName` directly instead of rediscovering agents with a filesystem regex.
- Keeps existing package-local Oxlint import restrictions and the Yarn transport-boundary constraint as the native static enforcement.
- Moves the Linear graph utility into core, where its tests use static imports and never skip based on checkout shape.
- Deletes the open-gaps stderr ledger and its shape tests, which tracked review state through Vitest rather than proving behavior.
- Removes app, sibling-package, and context-root inputs from core unit tests and removes the corresponding prune exceptions.

<details>
<summary>🏗️ Agent notes</summary>

### Inductive diagnosis

The review comments were a sample of one larger defect pattern: static policy had accumulated as filesystem-scanning unit tests. Those tests duplicated source code, package manifests, linter rules, file layout, and review decisions, then required custom ignore behavior and cross-workspace task inputs to remain runnable.

The rework applies one ownership rule:

```text
authored import policy → package-local Oxlint
manifest policy → Yarn constraints
emitted/runtime behavior → behavioral tests
source or review mirrors with no independent oracle → delete
```

### Mission contract

**Imperative.** Remove static-policy source walkers from Brunch unit tests and let existing enforcement mechanisms own the contracts they can actually prove.

**Proof.**

1. Core has no app, sibling-package, or context-root unit-test inputs and a core prune adds no downstream Brunch workspaces.
2. Package-local Oxlint and Yarn constraints pass; their existing policies remain unchanged.
3. The app build and emitted-bundle test prove the chat agent is registered under its source-of-truth identity.
4. The Linear graph tests use a normal static import from core and run unconditionally.
5. Searches find no architecture checker workspace, repo-chores checker, custom source walker, open-gap test ledger, or skip-on-file-existence guard.

**Constraints.**

- Do not duplicate Oxlint or Yarn policy in tests or repo-chores.
- Keep behavioral tests that can fail while source and manifests remain unchanged.
- Do not replace hard-coded mirrors with a new checker abstraction.

</details>

This PR:

The changes in this PR:

The changes in this PR:

## ⚠️ Known issues

None known.

## 🐾 Next steps

The reviewer identified similar pre-existing tautological assertions in core naming and authored-skill tests. They are non-blocking for SRE-1010 and are intentionally left for a separate product-test cleanup.

## 🛡 What tests cover this?

- `@hashintel/brunch-agent`: 10 files and 94 tests; build, lint, and typecheck pass.
- `@apps/brunch-agent`: 23 files and 166 tests; build, lint, and typecheck pass.
- Emitted-bundle suite: 10 behavioral assertions pass.
- Linear graph suite: 10 tests pass through a static core import.
- Prune repository fixtures: 8 tests pass.
- Yarn constraints, immutable install, task-dependency generation, and repository formatting pass.

## ❓ How to test this?

1. Run `yarn install --immutable && yarn constraints`.
2. Run `yarn workspace @hashintel/brunch-agent build`, `lint:tsc`, `lint:eslint`, and `test:unit`.
3. Run `yarn workspace @apps/brunch-agent build`, `lint:tsc`, `lint:eslint`, and `test:unit`.
4. Run `python3 .github/actions/prune-repository/prune_test.py`.
5. Inspect `yarn exec turbo run test:unit --filter @hashintel/brunch-agent --dry=json` and confirm no app, sibling-package, or context-root Brunch inputs.

## 📹 Demo

Not applicable — this changes static-policy ownership and CI task selection only.


TimDiekmann (changes requested):
Could you add a description and a title to the PR please?

TimDiekmann (dismissed):
Infra ✅

tests/brunch-agent-architecture/turbo.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": ["//"],
+  "tasks": {
```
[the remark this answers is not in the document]
lunelson: Fixed in 378b2970d1 with a repository-wide Yarn constraint. It derives the governed Brunch workspaces from the full workspace graph and requires each as a `workspace:*` devDependency of `@tests/brunch-agent-architecture`, before affected selection or pruning. I verified the negative case by removing one dependency and observing `yarn constraints` fail with the missing field.

apps/brunch-agent/test/architecture/app-workspace.ts
```diff
@@ -0,0 +1,63 @@
...
+        continue;
+      }
+      const path = join(current, entry);
+      if (statSync(path).isDirectory()) {
+        walk(path);
+      } else if (SOURCE_EXTENSIONS.test(entry)) {
+        found.push({
+          path,
+          relPath: relative(APP_ROOT, path).replaceAll("\\", "/"),
+          text: readFileSync(path, "utf8"),
```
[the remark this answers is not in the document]
lunelson: Fixed in 378b2970d1. The scanner now consumes `Dirent` values from `readdirSync(..., { withFileTypes: true })` and uses `entry.isDirectory()`, removing the separate path-based `statSync` check before reading.

.github/actions/prune-repository/prune.py
```diff
@@ -44,39 +44,22 @@
...
-    # context root alongside their plugin task dependencies.
+    # Core's Linear graph tests exercise the context-root script.
     "@hashintel/brunch-agent": [
         ".config/oxlint/brunch",
-        "libs/@hashintel/brunch-agent/AGENTS.md",
-        "libs/@hashintel/brunch-agent/CONTEXT.md",
-        "libs/@hashintel/brunch-agent/docs",
-        "libs/@hashintel/brunch-agent/evaluations",
-        "libs/@hashintel/brunch-agent/scripts",
+        "libs/@hashintel/brunch-agent/scripts/linear-project-graph.ts",
```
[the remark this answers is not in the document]
lunelson: Fixed in 378b2970d1. `contextRootPresent` now checks the exact copied file, `scripts/linear-project-graph.ts`. The focused suite runs all 10 tests locally.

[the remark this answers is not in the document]
lunelson: Fixed in 378b2970d1. `contextRootPresent` now checks the exact copied file, `scripts/linear-project-graph.ts`, and the focused Linear graph suite runs all 10 tests.

indietyp (commented):
I don't think this should be a test at all, and moving won't resolve it really. Conventions are typically enforced via linting (and can therefore also be checked in CI) instead of unit tests. I thought we agreed in the previous PR to - if we were to touch this - move this over to repochores (or oxlint?) instead of having unit tests for it.

lunelson:
@indietyp You’re right. Re-reading the prior thread, I explicitly agreed that `repo-chores` plus a dedicated CI step was the better long-term owner; the later SRE-1010 wording allowed a dedicated architecture workspace, but following that alternative did not honor the more specific review agreement. I’ll rework this PR so the family-wide repository conventions are owned by `repo-chores`, while app-local checks remain with the app and package-local Yarn/Oxlint rules remain with their packages.

lunelson:
@indietyp Reworked in e9f0007ab6: the family-wide checks now run as `@local/repo-chores#lint:brunch-architecture`, with a dedicated step in the Global lint job. The temporary Vitest architecture workspace and its dependency constraint are removed. App-only checks remain app-local as SRE-1010 specifies, and the package-local Yarn/Oxlint enforcement is unchanged. I also verified that representative app and package source changes each invalidate the repo-chores task, and that generated directories are excluded from its inputs.

.github/actions/prune-repository/prune_test.py
```diff
@@ -15,6 +15,9 @@
 CORE = "@hashintel/brunch-agent"
 TRANSPORT = "@hashintel/brunch-agent-transport-aisdk"
 APP = "@apps/brunch-agent"
+ARCHITECTURE = "@tests/brunch-agent-architecture"
```
indietyp: I believe this is unused now.

apps/brunch-agent/test/architecture/app-boundaries.test.ts
indietyp, line 1: I strongly oppose having this kind of thing here; this is a test that lists the source files. That's not a test. This is a lint stage masquerading as a unit test. Specifically, everything in here can be expressed inside existing infrastructure through [oxlint](https://oxc.rs/docs/guide/usage/linter/js-plugins.html). It would have multiple benefits; for one, you actually parse and have access to an AST and therefore don't need to rely on regular expressions, and it would integrate with of all our existing infrastructure.

apps/brunch-agent/test/architecture/app-workspace.ts
```diff
@@ -0,0 +1,63 @@
...
+).replace(/[/\\]$/u, "");
+
+export interface SourceFile {
+  readonly path: string;
+  readonly relPath: string;
+  readonly text: string;
+}
+
+const SOURCE_EXTENSIONS = /\.(ts|tsx|mts|mjs|js|jsx)$/u;
+const SKIP_DIRECTORIES = new Set([
```
indietyp: This feels like a partially incomplete `.gitignore` to me (something established linters would natively support, such that you can say: gitignore + no tests.

apps/brunch-agent/test/architecture/app-boundaries.test.ts
indietyp, line 1: non-blocking for this PR: looking closer, some of these tests also seem to be extremely tautological in nature:
- https://github.com/hashintel/hash/blob/e9f0007ab616572cd87819d37bd68362714f16de/libs/%40hashintel/brunch-agent/packages/core/test/elicitation-skill.test.ts#L8-L17 seems to just restate source code?
- https://github.com/hashintel/hash/blob/e9f0007ab616572cd87819d37bd68362714f16de/libs/%40hashintel/brunch-agent/packages/core/test/naming.test.ts#L84 Shouldn't this rather be encoded in the type system?
- https://github.com/hashintel/hash/blob/e9f0007ab616572cd87819d37bd68362714f16de/libs/%40hashintel/brunch-agent/packages/core/test/naming.test.ts#L62 I don't understand what this test is trying to achieve. You would typically want to use: https://vitest.dev/guide/testing-types.html

libs/@hashintel/brunch-agent/packages/core/test/architecture/context-root.ts
```diff
@@ -11,6 +11,6 @@ export const CONTEXT_ROOT = fileURLToPath(
  * Pruned jobs copy these non-workspace paths explicitly. Skip the affected
  * tests instead of failing opaquely if that CI contract drifts.
  */
-export const contextRootPresent =
-  existsSync(join(CONTEXT_ROOT, "docs")) &&
-  existsSync(join(CONTEXT_ROOT, "scripts"));
+export const contextRootPresent = existsSync(
```
indietyp: If tests start out with a skip depending on the existence of a file, then it's not a test - it's a lint.

This is also evident in https://github.com/hashintel/hash/blob/e9f0007ab616572cd87819d37bd68362714f16de/libs/%40hashintel/brunch-agent/packages/core/test/architecture/linear-project-graph.test.ts#L118, which seems a tad insane to me. 

This seems dubious at best: https://github.com/hashintel/hash/blob/e9f0007ab616572cd87819d37bd68362714f16de/libs/%40hashintel/brunch-agent/packages/core/test/architecture/open-gaps.test.ts#L30. This looks like it's trying to achieve what the excluded linear-project-graph is, but by using stderr on vitest instead?

libs/@local/repo-chores/node/scripts/check-brunch-architecture/workspace.ts
```diff
@@ -11,30 +11,18 @@
...
-  existsSync(join(CONTEXT_ROOT, "scripts"));
-const PACKAGES_ROOT = join(CONTEXT_ROOT, "packages");
-const APP_ROOT = join(HASH_ROOT, "apps/brunch-agent");
+export const HASH_ROOT = monorepoRootDirPath;
+export const CONTEXT_ROOT = nodePath.join(
+  HASH_ROOT,
+  "libs/@hashintel/brunch-agent",
+);
+const PACKAGES_ROOT = nodePath.join(CONTEXT_ROOT, "packages");
+const APP_ROOT = nodePath.join(HASH_ROOT, "apps/brunch-agent");
```
indietyp: It seems to me like the repo-chore should be a command that is invoked and targets a package, instead of hard-coding different roots.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
```diff
@@ -0,0 +1,476 @@
...
+  "joi",
+  "ajv",
+  "superstruct",
+  "arktype",
+  "io-ts",
+  "runtypes",
+  "@sinclair/typebox",
+  "@standard-schema/spec",
+  "@standard-schema/utils",
+];
```
indietyp: This seems unnecessary; we already have lints that disallow undeclared packages, and installing one of these would be better enforced via `yarn constraints`.

Essentially, there are two cases here:
1. transitive dependencies that one depends on: oxlint and eslint both catch these
2. a new dependency to be added: in light of the supply chain, we should minimise dependencies, so I would expect that the inclusion of a new dependency would first have a "background check", at which point it should be clear from the review that an unwanted schema library has entered the library.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
```diff
@@ -0,0 +1,476 @@
...
+
+    for (const dependency of runtimeDependencies(core)) {
+      assertCondition(
+        !/^@hashintel\/brunch-agent-(binding|plugin)-/u.test(dependency),
+        "core direction",
+        `${core.relPath} depends on ${dependency}`,
+      );
+    }
+
+    assertEqual(
```
indietyp: incomplete review of the code here, but it seems very tautological to me to have a test that asserts what is inside of a package.json.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
```diff
@@ -0,0 +1,476 @@
...
+      workspacePackage.relPath,
+    );
+  }
+
+  const core = packages.find(({ name }) => name === corePackageName);
+  assertCondition(
+    core !== undefined,
+    "core workspace",
+    `${corePackageName} is missing`,
+  );
```
indietyp: What is this trying to achieve?

.github/actions/prune-repository/prune_test.py
lunelson: Agreed — this fixture still modeled the temporary architecture workspace after the checker moved to repo-chores. Removed the test and its now-unused constants in 7c3e9cd633; the remaining 8 prune tests pass.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
```diff
@@ -0,0 +1,476 @@
...
+      `${filePath} has no review rationale`,
+    );
+  }
+
+  return violations;
+};
+
+const violations = checkBrunchArchitecture();
+
+if (violations.length > 0) {
```
indietyp: This all seems extremely convoluted and not really like a repo chore (at least not like this). It seems to me that most of the assertions present here basically fall into four camps:
1. already present through other mechanisms
2. better expressed through existing tooling
3. tautological and/or testing things that should be part of a review, not hard-coded.
4. Cryptic messaging: to a bystander, it is not clear what a "substrate" is, and why this specific substrate now needs to be "hermetic".

Some of these may be repo-chores, yes, but much of this is better expressed through existing tooling instead of starting another surface to maintain.

indietyp (changes requested):

apps/brunch-agent/test/architecture/app-workspace.ts
lunelson: Agreed. The custom filesystem walker is deleted rather than expanded into another partial ignore implementation. No replacement source scanner remains.

libs/@local/repo-chores/node/scripts/check-brunch-architecture/workspace.ts
lunelson: Agreed on the reusable-command shape, but after classifying the assertions there was no sufficient repo-chores command left: the checks were either already owned by Oxlint/Yarn/build behavior or were hard-coded mirrors. The command and root scanner are deleted in 9d8ae32.

libs/@hashintel/brunch-agent/packages/core/test/architecture/context-root.ts
lunelson: Addressed in 9d8ae32. The Linear graph utility now lives in core, its tests use static imports and never skip based on checkout shape, and the duplicated local interface declarations are gone. The open-gaps stderr ledger and its shape tests are deleted.

apps/brunch-agent/test/architecture/app-boundaries.test.ts
lunelson: Agreed that these are evidence of the same broader tautological-test pattern. This PR removes that pattern from the architecture-check ownership path; the named elicitation/naming tests predate SRE-1010 and are left for a separate non-blocking cleanup so this change does not broaden into product semantics.

lunelson: Agreed. The source-listing app architecture test and its regex walker are deleted in 9d8ae32. Forbidden imports remain with the existing package-local Oxlint configuration; the only retained agent-registration check inspects the emitted bundle, because that is observable build behavior rather than static source policy.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
lunelson: Agreed. The schema dependency/import inventory is deleted with the repo-chores checker. Existing undeclared-import lint, Yarn constraints, and dependency review remain the native enforcement points.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
lunelson: Agreed with the four-way classification. I treated the comments as a sample of a larger defect pattern rather than patching individual assertions: the repo-chores checker, custom scanners, hard-coded review inventory, manifest mirrors, schema allowlist, and associated CI task are all deleted in 9d8ae32. Existing Oxlint/Yarn enforcement remains, and only behavioral tests survive.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
lunelson: This attempted to constrain which core files could name the selected runtime, but it duplicated package-local import policy with a hard-coded file allowlist. The whole assertion is deleted; Oxlint remains the authored-import owner.

libs/@local/repo-chores/node/scripts/check-brunch-architecture.ts
lunelson: Agreed. The package.json mirror assertions and the entire bespoke checker are deleted in 9d8ae32; no replacement metadata snapshot was added.

libs/@hashintel/petrinaut/src/ui/views/Editor/panels/ai-assistant-panel.tsx
indietyp, line 1: I am confused, these file edits don't seem to correspond to the mentioned ticket (SRE-1010). Why is that?

lunelson: You are right: those Petrinaut edits were unrelated. Git Town had synchronized this branch against a stale local `main` worktree and replayed the already-merged H-6800 patch onto the PR. I rebuilt the ten SRE-1010 commits on current `origin/main` and force-pushed; the PR diff now contains only the intended Brunch architecture/prune changes.

indietyp (approved):

PR #9634, opened 2026-09-09
FE-1645: Show the prepared-fixture selector only in Brunch demo mode

kube:
## Summary

Before this PR, the demo site rendered a "Prepared Brunch fixtures" panel beside the editor whenever the Brunch AI was configured and the crew-reservation fixture was not the current net. The panel sat 16px from the top at a stacking order below Petrinaut's 64px top bar, so only its rounded bottom edge showed, as a stray box behind the model name in every mode. Its one link opens the fixture through the `brunch-fixture` query, the entry point of the Mission 6 two-tab witness.

The panel is now a demo affordance that is never displayed by default. The palette command **Toggle Brunch demo mode** flips a persisted user setting, `brunchDemoMode`, that the settings dialog does not show; while it is on, the selector sits centred below the top bar, clear of the side panels, like the fixture status banner that appears once the fixture is current. To read that setting outside the editor, the demo shell mounts Petrinaut's `UserSettingsProvider` above `Petrinaut`, and the editor reuses that instance.

**Before**

https://github.com/user-attachments/assets/1a754aa2-04b1-4996-92db-14d31441da71

**After**

https://github.com/user-attachments/assets/56f6715b-5027-4bb4-87ef-7595ecb0729e

## Links

- Ticket [FE-1645](https://linear.app/hash/issue/FE-1645)

#### Petrinaut

- `UserSettingsProvider` reuses an ancestor provider
  > A host mounts it above `Petrinaut` and reads or changes the persisted settings from its own components; the editor's own instance then renders its children, so one state owns the storage key.
- `UserSettingsProvider`, `UserSettingsContext` and their types are exported from `@hashintel/petrinaut/react`
- `brunchDemoMode` joins the user settings
  > Off by default, persisted with the others, no control in the settings dialog.

#### Demo site

- **Toggle Brunch demo mode** joins the palette's Demo commands
  > Listed only on a Brunch-configured build; it flips `brunchDemoMode`.
- Prepared-fixture selector shows only in demo mode
  > Fixed 80px from the top and centred, the same placement as the fixture status banner, so nothing of the demo shell sits behind the top bar.

## Test coverage

- `user-settings-provider.test.tsx`:
  > Brunch demo mode starts off and toggles; a nested provider shares the ancestor's state.
- `local-storage-demo-app.test.tsx`:
  > The selector is absent by default, appears fixed at 80px after the palette command, and disappears when the command runs again.
- `prepared-fixture-banner.test.tsx`:
  > The selector's label, link and placement.

## How to test

- Open [Petrinaut preview](https://petrinaut-git-claude-fe-1645-remove-prepared-fixture-selector.stage.hash.ai) on Vercel
- Menu > Load example > any model
  > Expect no box peeking under the top bar behind the model name, in Edit and in Simulate
- Press ⌘K > Toggle Brunch demo mode
  > Expect the "Prepared Brunch fixtures" panel centred below the top bar
- Reload
  > Expect the panel still shown
- Click "Open the labelled crew-reservation fixture"
  > Expect the "Test-authored prepared fixture" banner in the same place, and no selector
- Press ⌘K > Toggle Brunch demo mode
  > Expect the selector gone once the fixture is left

### DIFF
--- a/.changeset/user-settings-provider-for-hosts.md
+++ b/.changeset/user-settings-provider-for-hosts.md
@@ -1,0 +1,5 @@
+---
+"@hashintel/petrinaut": patch
+---
+
+Hosts can mount `UserSettingsProvider` from `@hashintel/petrinaut/react` above `Petrinaut` to read and change the persisted user settings from their own components; the editor reuses that provider instead of creating its own.
--- a/apps/petrinaut-website/src/main/app/local-storage-demo/local-storage-demo-app.test.tsx
+++ b/apps/petrinaut-website/src/main/app/local-storage-demo/local-storage-demo-app.test.tsx
@@ -584,9 +584,6 @@ describe("local storage demo prepared fixture", () => {
     );
 
     expect(
-      document.querySelector('[aria-label="Prepared fixture selector"]'),
-    ).toBeNull();
-    expect(
       document.querySelector('[aria-label="Prepared fixture status"]'),
     ).toBeNull();
     const aiAssistant = editorProps.current?.aiAssistant as
--- a/apps/petrinaut-website/src/main/app/local-storage-demo/local-storage-demo-app.test.tsx
+++ b/apps/petrinaut-website/src/main/app/local-storage-demo/local-storage-demo-app.test.tsx
@@ -1,7 +1,14 @@
 /**
  * @vitest-environment jsdom
  */
-import { act, cleanup, render, waitFor } from "@testing-library/react";
+import {
+  act,
+  cleanup,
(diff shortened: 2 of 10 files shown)


kostandinang (approved):
Not a blocker: `/docs/ai-assistant.md` may need to reflect that the fixture is now shown via Toggle Brunch demo mode.

lunelson (approved):
Thanks for catching this. As far as I can tell, it doesn't conflict with anything in our other branches, so it should be fine for you to merge this first.

PR #9636, opened 2026-09-09
BE-742: Keep the messages Postgres sends outside a statement's results

TimDiekmann:

Postgres reports warnings and notices on the same wire as query results, but `tokio_postgres` delivers them on a side channel that only `Connection::poll_message` exposes. Awaiting the connection, which is what a pool does by default, discards every one of them. Server warnings have therefore been invisible for as long as the graph has had a pool: `SET LOCAL` outside a transaction, for one, is only ever reported as a warning.

The graph now owns its pool's connection manager. Each connection runs a task that reads that channel, so what the server says becomes a tracing event under `hash_graph_postgres_store::server` at the level of its severity, with the SQLSTATE, detail, hint and context as fields. A termination the server initiates arrives on the same channel as a fatal error response and is recorded the same way, and a connection whose recording task has stopped is not handed out again.

On the way, the store gets a transaction API of its own instead of implementing the migration runner's traits, and `deadpool-postgres` goes.

## 🔗 Related links

- [BE-742](https://linear.app/hash/issue/BE-742) _(internal)_
- Parent: [BE-708](https://linear.app/hash/issue/BE-708) _(internal)_, plan capture over the same channel, which this PR deliberately leaves out
- Follow-up: [BE-703](https://linear.app/hash/issue/BE-703) _(internal)_, pool timeouts

## 🚫 Blocked by

- Nothing.

## 🔍 What does this change?

- **Store transaction API** (first commit): `PostgresStore::transaction`, the builder's `isolation_level`/`read_only`/`deferrable`, and `commit`/`rollback` are inherent methods. The store no longer implements `hash_graph_migrations::{Context, Transaction, TransactionBuilder}`, and nine test and bench files stop importing those traits to open a transaction.
- **Migrations crate**: the runner never configured a transaction, so `TransactionBuilder` folds into `Context::transaction` and `IsolationLevel` moves to `hash-graph-postgres-store` next to `TransactionOptions`. `hash-graph-postgres-store` no longer depends on `hash-graph-migrations`; `package.json`, `docs/task-dependencies.json` and `yarn.lock` are regenerated accordingly.
- **`connection.rs`** (new): `ManagedConnection` pairs a `tokio_postgres::Client` with the task driving `poll_message`; `ConnectionManager` implements `deadpool::managed::Manager` and rejects a connection on recycle when its client is closed or its task has finished. `ConnectionError` replaces `deadpool_postgres::PoolError` as `StorePool::Error`.
- **`pool.rs`**: `PooledConnection` is a newtype over deadpool's `Object`, which keeps `deadpool` a private dependency. `PostgresStorePool::new` never connected and is no longer `async`; the callers drop their `.await`.
- **`deadpool-postgres`** is removed. Only its `Config` was ever used: the statement cache is opt-in and was never opted into, and its transaction wrappers exist to offer that same cache.
- The `SET LOCAL` sites in the store already run inside transactions, so the new warning does not fire on a hot path.

This PR:

The changes in this PR:

The changes in this PR:

  - `hash-graph-postgres-store` loses its edge to `hash-graph-migrations`; `mise run sync:turborepo` regenerated `package.json` and `docs/task-dependencies.json`, no `turbo.json` needed a change

## ⚠️ Known issues

- **Pool timeouts stay unbounded**, as before: deadpool's defaults are all `None`, matching the explicit `None`s the old configuration set. `ConnectionError::Unavailable` therefore covers a wait timeout and a missing runtime with one variant, and no pool error carries a `StatusCode`, so a saturated pool would answer 500 rather than 503. Unreachable until a timeout is configured; tracked in BE-703, where the variant split and the status code belong.
- **Server messages carry no request trace.** The recording task runs outside any span on purpose, because a connection outlives the request that made the pool grow. Every event carries the `connection` id as a field of its own, so it survives any filter level, and that id joins it to the acquisition that created the connection, which is logged within that request's span. Attributing a message to the statement that provoked it needs a collector, which is BE-708's subject.
- **Session state is not reset when a connection returns to the pool**, unchanged from `deadpool-postgres`'s `RecyclingMethod::Fast`. The store sets session parameters only with `SET LOCAL` inside a transaction, and a `SET LOCAL` outside one now shows up as a warning, so a reset round trip per checkout is not worth its cost.

## 🐾 Next steps

- BE-708: plan capture via `auto_explain` notices over this channel, with a collector per transaction.
- BE-741: turning plan capture on in a deployment without a redeploy.
- BE-703: pool timeouts, with the error split and the status code above.

## 🛡 What tests cover this?

- New, in `libs/@local/graph/postgres-store/tests/connection/main.rs`, each installing a `tracing` layer that records what reaches `hash_graph_postgres_store::server`:
  - `connection_records_server_warnings`: `SET LOCAL` outside a transaction is recorded at WARN with severity, SQLSTATE `25P01` and the connection id on the event.
  - `connection_records_server_notices`: `RAISE NOTICE … USING DETAIL, HINT` is recorded at INFO with detail, hint and context, so the level mapping is pinned on more than one arm and every field the server sends is asserted.
  - `connection_records_notifications`: `LISTEN` and `NOTIFY` on one connection record the notification at INFO with its channel.
  - `connection_records_server_termination`: `pg_terminate_backend` from a second connection is recorded at ERROR with `FATAL` and `57P01`.
  - `pool_replaces_terminated_connection`: a pool of one connection does not hand a terminated backend out again.
  - `pool_reports_unreachable_database`: the pool connects lazily and reports a missing database as `ConnectionError::Connect` with the driver's error in the report.
- Existing, in `tests/graph/integration/postgres/transaction.rs`: options composition, defaults and savepoint nesting through the store's transaction API.
- Every other integration test reaches the database through the new pool.

## ❓ How to test this?

1. Start the compose stack and run `cargo nextest run -p hash-graph-postgres-store --all-features --test connection`.
2. Run the graph with `RUST_LOG=hash_graph_postgres_store::server=info` and, from `psql`, `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE usename = 'graph' AND state = 'idle' LIMIT 1;`.
3. Confirm the graph logs an ERROR event with `severity=FATAL` and `code=57P01`, and that the next request succeeds on a fresh connection.

## 📹 Demo

No UI. The events the tests assert on are what an operator sees in the graph's log.


libs/@local/graph/postgres-store/src/store/postgres/connection.rs
```diff
@@ -0,0 +1,332 @@
...
+            PoolError::Backend(report) | PoolError::PostCreateHook(HookError::Backend(report)) => {
+                report
+            }
+            PoolError::Timeout(TimeoutType::Wait) => Report::new(Self::Unavailable)
+                .attach("the pool timed out waiting for a connection to become available"),
+            PoolError::Timeout(TimeoutType::Create) => Report::new(Self::Unavailable)
+                .attach("the pool timed out establishing a connection"),
+            PoolError::Timeout(TimeoutType::Recycle) => {
+                Report::new(Self::Unavailable).attach("the pool timed out recycling a connection")
+            }
```
indietyp: shouldn't these attaches be enum variants w/ display? / just a normal change context over the timeout type?

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+        |severity| severity.to_string(),
+    );
+    let code = notice.code().code();
+    let message = notice.message();
+
+    // `detail`, `hint` and `context` hold what to do about the message and which function raised
+    // it.
+    macro_rules! event {
+        ($level:ident) => {
+            tracing::$level!(
```
indietyp, line 135: I believe this is the same as `tracing::event!(target: ..., Level::WARN, ...)`, in which case the macro here is not needed, and you can just derive the level from the severity, and attach it that way. ([https://docs.rs/tracing/latest/tracing/macro.event.html](<https://docs.rs/tracing/latest/tracing/macro.event.html>))

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Establishes a connection with the TLS setup erased.
+///
+/// A [`tokio_postgres::Connection`] names the TLS stream type it was built from, so it cannot cross
+/// an erased boundary. Driving it starts here, where that type is still known, and only the
+/// [`ManagedConnection`] comes back.
+trait Connect: Send + Sync {
+    fn connect(
+        &self,
+        id: u64,
+    ) -> BoxFuture<'static, Result<ManagedConnection, Report<ConnectionError>>>;
```
indietyp: why does `Connect` require a `BoxFuture` here, can't this just use: RPTIT? I guess it's to allow the connection manager to stay without generics, but what is the reason that the connection manager shouldn't be generic over the connection method?

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
```diff
@@ -0,0 +1,332 @@
...
+/// its messages carry `id` rather than a request's trace, and `id` is what joins them to the
+/// acquisition that created the connection.
+async fn drive(
+    id: u64,
+    mut messages: impl Stream<Item = Result<AsyncMessage, tokio_postgres::Error>> + Unpin,
+) {
+    while let Some(message) = messages.next().await {
+        match message {
+            Ok(message) => record(id, message),
+            // `Connection::poll_message` documents an error as terminal and closes its request
```
indietyp: shouldn't this be handled in your stream by implementing `FusedStream`? (and requiring it here?) I also believe you can get rid of the `Unpin` requirement, you could just `let mut messages = pin!(messages)` as the first statement.

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
```diff
@@ -0,0 +1,332 @@
...
+    ) -> BoxFuture<'static, Result<ManagedConnection, Report<ConnectionError>>> {
+        let config = self.config.clone();
+        let tls = self.tls.clone();
+
+        Box::pin(async move {
+            let (client, mut connection) = config
+                .connect(tls)
+                .await
+                .change_context(ConnectionError::Connect)?;
+            let messages = stream::poll_fn(move |context| connection.poll_message(context));
```
indietyp: I would honestly just have a `MessageStream` newtype that polls, then implements `FusedStream` and terminates/drops on the first err (to encode the invariants that `poll_message` requires. Said type could then also have the `drive` (or `run`) as an inherent method, with e.g. `spawn`.

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+/// A pooled connection that records the server's messages instead of dropping them.
+///
+/// The client and the task recording its messages are one unit. Dropping the connection drops the
+/// client, which lets `tokio_postgres` close the connection gracefully with a `Terminate`; the
+/// recording task then sees the message stream end and finishes on its own.
+#[derive(Debug)]
+pub struct ManagedConnection {
+    id: u64,
+    client: Client,
+    driver: JoinHandle<()>,
```
indietyp, line 196: I guess the flow here is (and why it isn't an [`AbortOnDropHandle`](<https://docs.rs/tokio-util/latest/tokio_util/task/struct.AbortOnDropHandle.html>)) that `Drop` will terminate the connection, which means that the driver will yield `None`, which in turn means that it winds down automatically.

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Postgres planning every execution against its actual parameter values.
+    #[must_use]
+    pub const fn client(&self) -> &Client {
+        &self.client
+    }
+
+    /// The client statements are issued on, borrowed exclusively.
+    pub const fn client_mut(&mut self) -> &mut Client {
+        &mut self.client
+    }
```
indietyp, line 212: Why not impl `AsRef`, `AsMut`?

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+}
+
+impl deadpool::managed::Manager for ConnectionManager {
+    type Error = Report<ConnectionError>;
+    type Type = ManagedConnection;
+
+    async fn create(&self) -> Result<Self::Type, Self::Error> {
+        let id = self.next_id.fetch_add(1, Ordering::Relaxed);
```
indietyp: considering that we control the id, would it make sense to create a newtype for this? that way we can't just accidentally launder ids.

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
TimDiekmann: It is, however, it requires a const value:

> attempt to use a non-constant value in a constant non-constant value (rustc [E0435](https://doc.rust-lang.org/error_codes/E0435.html))

TimDiekmann: Image

indietyp: I see now, it's because it creates a static using it: [`static __CALLSITE:`](https://docs.rs/tracing/latest/src/tracing/macros.rs.html#619), so there's no easy way for us to do this.

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
TimDiekmann: I removed the `Report` return from the manager, so we can properly use the `PoolError` as `impl Error` here and fixed the concern.

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
TimDiekmann: I originally had it used as `dyn Connect` further below and this would have required the associated type to be bubbled up. However, I just decided to make it generic over `Tls` so this is gone.

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
TimDiekmann: Added it through a `MessageStream`

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
TimDiekmann: Probably the best solution, yes. Fixed.

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
TimDiekmann: Mainly to be explicit here, that we return the `Client`, not anything else from the connection. I can add them if you want.

libs/@local/graph/postgres-store/src/store/postgres/connection.rs
TimDiekmann: Added a `ConnectionId`

libs/@local/graph/postgres-store/src/store/postgres/connection/mod.rs
indietyp: No, that makes total sense, thank you!

indietyp (dismissed):

indietyp (approved):

PR #9646, opened 2026-09-10
FE-1648: add select option to filter

alex-e-leon:

Adds a new "select" option to <Filter> that allows adding single or multi-selects as filter input types. Also tidies up various ui select/select adjacent components. Specifically:
- Add selection summary options to Select + better overflow handling inside the input for multiSelect
- Improves keyboard handling for Selects, overriding ark-ui's default focus-trapping while selects are open so that users can tab out of selects, maps some other Select keyboard shortcuts to be closer to APG guidelines, and adds better keyboard handling inside Filter
- Updates some Select + Input widths to correctly handle min/max + default widths at all container sizes/cases, including flex container variants.
- Improve alignment of header/footer/search/custom items inside SelectableList UI
- Simplified api pattern for searchable/clearable/etc updated across various components.

This PR:

The changes in this PR:


lunelson (approved):
Looks really nice 👍🏼

